PR #85, opened 2023-08-28
Issue management documentation

sdhoyt:
# Change Summary

## Overview
This update adds documentation for how requirements/issues will be tracked and managed in Github. My goal with this is to give a quick general overview of requirement mapping, lifecycle, then go into details of how this will be managed in Github.

The main things I'm hoping for people to review are:
1. What do you think about this plan for managing requirements in Github? Do people think this will be helpful? Does the strategy make sense? Any suggested changes?
2. Is the documentation clear enough to understand? Do you feel from reading this you understand how requirements and issues in Github (including their labels, status, tasks, comments) should evolve throughout development and test? Is there too much information?  Too little information?  Any parts that seem complicated? (If so, it's probably due to a bad explanation on my part)
3. Does anyone have any suggestions for making the images easier to see? The diagrams are large and can currently only be clearly viewed by clicking the image and opening it in another tab.

## New Dependencies
None

## New Files
- requirement-and-issue-tracking.rst
   - documentation for requirement and issue tracking
- new images added to _static

## Deleted Files
None

## Updated Files
- style-guide.rst
   - added requirement-and-issue-tracking to table of contents

## Testing
Built the documentation on my branch: https://imap-processing-fork.readthedocs.io/en/issue-docs/

PR #92, opened 2023-08-30
DOC: Update sphinx references

greglucas:
# Change Summary

As I was going through #85 I noticed that there were unrelated doc build warnings: https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/6015856727/job/16318618164#step:5:43
so this goes through and fixes up the minor warnings.

There was a duplicate explicit named reference, so change those to implicit names with "__" for the links.
The poetry-link reference needed to be swapped and go within the brackets.

## New Dependencies

I updated the pre-commit versions with `pre-commit autoupdate` and then committed that file. There was a version mismatch in the `poetry.lock` file (pre-commit saying it was 1.5.0 that it wanted, but the lock file having 1.5.1). I am not sure how that happened, but perhaps because someone made the lock file through their local poetry venv rather than the pre-commit venv? I wonder if there is a way to make sure those are consistent or that we can guarantee the consistency somehow there...

Now, of course, we updated to 1.6.0 within pre-commit and it is OK with the 1.5.1 creation path! I think there was an extra YAML dependency added in with version conflicts, so apparently now the poetry.lock check is happy 🤷


maxinelasp:
The lock file may be inconsistent because the precommit uses the `--no-update` flag. It presumably saw that the version was up to date and didn't update again because it wasn't necessary. That's my guess!

maxinelasp (approved):

greglucas:
> The lock file may be inconsistent because the precommit uses the --no-update flag. It presumably saw that the version was up to date and didn't update again because it wasn't necessary. That's my guess!

Yeah, I think it does have to do with the way we are updating things locally vs pre-commit.

My guess is that it is because I don't have poetry in my main environment and I rely on the specific poetry version within pre-commit's environment (separate from my work environment). But, you all have poetry available locally so if you manually run `poetry lock` that will use the _local_ poetry which may not be pinned to the exact same version as the pre-commit poetry version. (That is why there was a "this was generated with 1.5.1" within the poetry lock file when pre-commit is pinned to 1.5.0

maxinelasp:
I looked into keeping poetry at one version, but I couldn't find a way to specify the _poetry_ version in pyproject.toml (turns out googling "poetry specify version" does not get you helpful results, and I don't know what else to search for, lol)

bourque (approved):
LGTM

PR #95, opened 2023-09-05
DRAFT - SWE l1a algorithm

tech3371:
# Change Summary

## Overview
This code is still in work. I wanted to open it in case anyone is working on similar task like mine and like to take a look at what I am doing.

## New Files
l1a_swe.py contains all of l1a algorithm.

## Updated Files
- pyproject.toml
   - It includes libraries that I needed locally to test but could be take away before final PR
- swe_packet_definition.xml
   - updated SWE packet definition to match their latest submission

## Testing
None


imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+    # It's like putting a puzzle together. One full data table in the
+    # the algorithm document is of this shape ( 24, 30, 7). Once we
+    # have populated data table with all four spins, we can reshape
+    # it to what subsequent algorithm needs which is (24, 7, 30).
+    spin_one_indexes = [[1,  5, 9, 13, 17,  21], [23, 19, 15, 11, 7, 3]]
+    spin_two_indexes = [[2,  6, 10, 14, 18,  22], [24, 20, 16, 12, 8, 4]]
+    spin_three_indexes = [[3,  7, 11, 15, 19,  23], [21, 17, 13, 9, 5, 1]]
+    spin_four_indexes = [[4,  8, 12, 16, 20,  24], [22, 18, 14, 10, 6, 2]]
+
+    match spin_number:
```
greglucas, line 85: This is Python 3.10+, so I don't think we can use this quite yet if we want to keep supporting Python 3.9.

This could also be written as a dictionary lookup if you'd prefer:

```python
step_indices = {
    0: [[1,  5, 9, 13, 17,  21], [23, 19, 15, 11, 7, 3]],
    1: [[2,  6, 10, 14, 18,  22], [24, 20, 16, 12, 8, 4]],
    2: [[3,  7, 11, 15, 19,  23], [21, 17, 13, 9, 5, 1]],
    3: [[4,  8, 12, 16, 20,  24], [22, 18, 14, 10, 6, 2]]
}
```
Then access with `step_indices[0]`

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
+import bitstring
+import numpy as np
+import xarray as xr
+
+from imap_processing import packet_definition_directory
+from imap_processing.swe import decom_swe
+
+
+def uncompress(cem_count):
```
greglucas, line 9: I'm not sure how often this is being used below, but you might want to consider adding a cache to this since it seems to be a direct lookup table calculation. Or just pre-computing the entire lookup table ahead of time:

```suggestion
@functools.cache
def uncompress(cem_count):
```

or since you know there is a maximum of 256 values that seems reasonable to store up front if you're going to be hitting it a lot.

`decompression_table = [uncompress(i) for i in range(256)]`

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+    -------
+    int
+        uncompressed count. Eg. 40959
+    """
+    # index is the first four bits of input data
+    # multi is the last four bits of input data
+    index = cem_count // 16
+    multi = cem_count % 16
+
+    # uncompression formula
```
greglucas, line 27: I'd suggest linking to explicit pages/locations in the algorithm documents in these formulas too. I think that will help us later going back to debug things and cross-reference check where values came from.

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+        return 2 ** (index + 3)
+    elif index == 7:
+        return 768
+    elif index == 8:
+        return 1024
+    elif 9 <= index <= 15:
+        return 2**index + 1024
+    else:
+        print("\nError! Switch for index failed in function N.")
+        return 0
```
greglucas, line 48:
```suggestion
        raise ValueError("Invalid base, must be between 0 and 15.")
```

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+        # read raw data
+        binary_data = decom_data[total_packets].data["SCIENCE_DATA"].raw_value
+        # read raw data as binary array using bitstring
+        bit_array = bitstring.ConstBitStream(bin=binary_data)
+        # chunk binary into 1260 units each with 8-bits
+        byte_data = bit_array.readlist(["uint:8"] * 1260)
+
+        # Uncompress counts. Uncompressed data is a list of 1260
+        # where 1260 = 15 seconds x 7 CEMs x 12 energy steps
+        uncompressed_data = [uncompress(i) for i in byte_data]
```
greglucas, line 135: Looks like you are calling this 1260 times on each packet, so I'd definitely suggest caching / creating the lookup table above.

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # TODO: get metadata of each data and stores. Find out how to combine four
+        # packet's metadata.
+
+        # read raw data
+        binary_data = decom_data[total_packets].data["SCIENCE_DATA"].raw_value
+        # read raw data as binary array using bitstring
+        bit_array = bitstring.ConstBitStream(bin=binary_data)
+        # chunk binary into 1260 units each with 8-bits
+        byte_data = bit_array.readlist(["uint:8"] * 1260)
```
greglucas, line 131: Alternatively if you wanted to go straight to a numpy array and avoid bitstrings since you are dealing with full bytes.
```suggestion
        binary_data = decom_data[total_packets].data["SCIENCE_DATA"].raw_value
        # read raw data into a uint8 numpy array
        byte_data = np.frombuffer(binary_data, dtype=np.uint8)
```

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+        # Reset column index for each spin.
+        column_index = 0
+        # Get indexes for each spin.
+        indexes = get_step_indexes(spin_number)
+
+        for every_second in range(15):
+            # go through first six steps of first half second and populate data in the right index
+            # of one_full_cycle array.
+            for every_step, index_number in enumerate(indexes[0]):
+                one_full_cycle[index_number-1, column_index] = reshaped_data[every_second, every_step]
```
greglucas, line 148: Wow, they didn't make this easy on you! Nice work decoding all this.

I _think_ you might be able to get rid of the outer `every_second` loop, because `column_index` is incrementing with `every_second`?

```suggestion
            for every_step, index_number in enumerate(indexes[0]):
                one_full_cycle[index_number-1, ::2] = reshaped_data[:, every_step]
                one_full_cycle[index_number-1, 1::2] = reshaped_data[:, 6 + every_step]
```

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+                        value.raw_value
+                        if value.derived_value is None
+                        else value.derived_value
+                    )
+            print(one_full_cycle[:4, 0, :])
+            xarray_data = xr.DataArray(
+                list(one_full_cycle), dims=["energy_step", "cem_counts", "angle"],
+            )
+            dataset = xr.Dataset({"data": xarray_data})
+            dataset.to_netcdf(f"swe_l1a_{total_packets}.nc")
```
greglucas, line 178: :( I wish we could use netcdf too! I don't think we should keep this as a dependency, unfortunately. One thought would be to add it to testing dependencies.

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+        24 rows --> 24 energy steps
+        7 column --> 7 CEMs value
+        30 element --> 30 spin angles
+    """
+    packet_file = f"{packet_definition_directory}/../swe/tests/science_block_20221116_163611Z_idle.bin"
+    decom_data = decom_swe.decom_packets(packet_file)
+
+    total_packets = 0
+    one_full_cycle = np.zeros((24, 30, 7))
+    while total_packets < len(decom_data):
```
greglucas, line 115:
```suggestion
    for packet in decom_data:
```

imap_processing/swe/l1a_swe.py
```diff
@@ -0,0 +1,183 @@
...
+            xarray_data = xr.DataArray(
+                list(one_full_cycle), dims=["energy_step", "cem_counts", "angle"],
+            )
+            dataset = xr.Dataset({"data": xarray_data})
+            dataset.to_netcdf(f"swe_l1a_{total_packets}.nc")
+
+        total_packets += 1
+
+
+swe_l1a()
```
greglucas, line 183:
```suggestion
if __name__ == "__main__":
    swe_l1a()
```

greglucas (commented):
I like all the comments throughout! Very helpful to follow what is going on.

I only have pretty minor comments/suggestions for optimizations, so nothing critical at all. I appreciate the TODOs in the code as well knowing that we'll need to come back to those areas.

imap_processing/swe/l1a_swe.py
tech3371: will do

imap_processing/swe/l1a_swe.py
tech3371: noted

imap_processing/swe/l1a_swe.py
tech3371: That's a great idea! I always forget about using dictionary.

imap_processing/swe/l1a_swe.py
tech3371: I had that originally but I don't remember now why I changed it to while statement. :) I went through many iteration of implementation.

imap_processing/swe/l1a_swe.py
tech3371: good to know!

imap_processing/swe/l1a_swe.py
tech3371: Screen Shot 2023-09-06 at 8 53 54 AM
SWE lookup table looks like this.

I do agree that calculating ahead might be good but I will think about your suggestion offline to understand.

I plan to refactor code after discussion with SWE team this friday.

imap_processing/swe/l1a_swe.py
tech3371: I will try that. I was trying to think different ways and it was mind twisting. I was happy when this thing worked but was hoping for suggestion like this. I am going to ask them how they did in the past or what we have is good or not.

imap_processing/swe/l1a_swe.py
tech3371: This was mainly for testing. This line will be replaced by xarray_to_cdf call. I plan to remove it before opening the PR.

imap_processing/swe/l1a_swe.py
greglucas: If you're looking for more things to think about, I believe `take()` or `take_along_axis()` might be able to do the fancy numpy indexing for you to remove the Python loops altogether based on your index lookup tables.

https://numpy.org/doc/stable/reference/generated/numpy.take.html

I think it is very good to start with the basic for-loops here to make sure you have things working as you expect and then optimize one thing at a time, so I think you've got a good approach going right now.

imap_processing/swe/l1a_swe.py
greglucas: :+1: Sounds good. Looking at that table, is there a reason not to just hard-code those numbers into a list? I think you have some fancy 2^3 code currently, which is somewhat confusing when you're going back and forth between fixed table values. Unless the 2^3 is somewhere in the document as well.

tech3371: I was thinking that myself too - If I can take this table, hard code this base and step_size and look up using index value. I will ask them if this table will change.

imap_processing/swe/l1a_swe.py
tech3371: cool. I will take a look at that.

imap_processing/swe/l1a_swe.py
greglucas: So, I decided to run some timing comparisons to see if this would be worth it actually...

```bash
python -m timeit 'import bitstring; bitstring.ConstBitStream(bytes(1260)).readlist(["uint:8"]*1260)'
100 loops, best of 5: 2.68 msec per loop
python -m timeit 'import numpy as np; np.frombuffer(bytes(1260), dtype=np.uint8)'
500000 loops, best of 5: 668 nsec per loop
```

That is a pretty massive improvement by using numpy directly. I'm kind of surprised bitstring is that slow...

tech3371: That's impressive!

tech3371:
I am closing this draft PR because this code is going to change a lot after my discussion with SWE team this afternoon.

Issue #75, opened 2023-08-23
Xtce build follow up: make updates on closed PR #72


GFMoraga:
### Summary
This is a way to track the progress of updating the xtce files and parsing tools 

### Completed
- [x] We need to add pandas, openpyxl to poetry dependency list: **added to dependencies, and not optional dependencies.** 
- [x] Made a new directory called 'tools' under the 'imap_processing', as well as 'tools/xtce_generation' which will house the files to make xtce
- [x] Fix linting issues
- [x] Moving docstrings around to be compliant with PEP257/numpydocs
- [x] Adding said documentation to the imap_processing docs (as well as any other outward-facing docs we want to add)
- [x] Utilizing** classes, config files, or argument parsing to abstract out parameters from the code
- [x] Make the code more generalized, generally speaking
- [x] Squash merge commits
- [x] Create for loop for each row in lengthInBits column and use that to create IntegerParameterType. That way, if there are int sizes greater than 32 or unique values after that, they would get created.
- [x] Fix Matthew's code suggestion on the ".org/space/xtce"

### Requirements

The more specific the better.
  - Review closed PR #72 
  

### Related tickets

#72

GFMoraga:
@bourque I copied your comment from the PR as a starter. I will go back and edit/update.

bourque:
Thanks @GFMoraga. I am also happy to help out with some of this!

GFMoraga:
@maxinelasp can you suggest names for classes? Or do you think these are fine? I am working on this comment:

"It is best practice to have the main function be really small and call out to other methods (IMO). Ideally, this would look something like:"

df = parse_excel_file(filename)
header = generate_header()
integer_parameter_types = generate_integer_parameter_types()
codice_parameters = generate_codice_parameters()
tree = generate_xml_tree(df, header, integer_parameter_types, codice_parameters)
tree.write(...)

GFMoraga:
@tech3371 how do I do this? "We need to add pandas, openpyxl to poetry dependency list" in poety.toml?

maxinelasp:
Make sure your names are clear, with spelled out names (per the [style guide](https://imap-processing.readthedocs.io/en/latest/development/style_guide/python-coding.html)). Ideally, it should be clear what the class is without other developers needing to read any documentation or comments. If you're having trouble describing a class in 2-3 words, that is a sign the class is too complicated and may need to be split down further. 

[...]

PR #106, opened 2023-09-06
Adding CoDICE Event data + Species counts packet definitions for L0

GFMoraga:
## Overview
I made `.py` files for 5 codice science packets, and output to `packet_definitions`. I added a new directory `/packet_definitions/codice` to store codice specific packet definitions. 

## Goal
Break down the repetitive nature of the `.py` files. 

## Feedback required
Please help me with insight on how to break down the python code. 
Issue #75 "Utilizing classes, config files, or argument parsing to abstract out parameters from the code"
-argparse
-configure (json)
-classes

## New Files

new file:   L0/hi_pha.py
new file:   L0/lo_pha.py
new file:   L0/nsw_species_counts.py
new file:   L0/omni_species_counts.py
new file:   L0/sw_species_counts.py

new file:   ../packet_definitions/codice/hi_pha.xml
new file:   ../packet_definitions/codice/lo_nsw_species_counts.xml
new file:   ../packet_definitions/codice/lo_pha.xml
new file:   ../packet_definitions/codice/lo_sw_species_counts.xml
new file:   ../packet_definitions/codice/omni_species_counts.xml

## Updated Files

- make_xtce.py
   - changed the ouput to `example_xtce.xml`


imap_processing/codice/L0/lo_pha.py
```diff
@@ -0,0 +1,220 @@
...
+    - Within `TelemetryMetaData`, the `ParameterTypeSet` element is created
+    to define parameter types.
+    - Similarly, the `ParameterSet` element is added to `TelemetryMetaData`
+    to define specific parameters.
+    These elements and their organization help organize and standardize
+    data representation in XML format.
+    """
+
+    # Create the TelemetryMetaData element
+    telemetry_metadata = Et.SubElement(
```
maxinelasp: This looks like a fair amount of duplicated code - could you pull out parts of this into methods, so you can reuse it in all these processing steps? Maybe by passing in the unique names as a dictionary?

maxinelasp: From running `hi_pha` and `lo_nsw` in diffchecker, it looks like they are almost identical. Definitely pull that into a function so you can reuse it.

maxinelasp (changes requested):
This has a lot of duplicated code that I'd like to see consolidated into one method. However, that code itself looks great, and this will definitely be a handy reference for other instruments!

imap_processing/codice/L0/hi_pha.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+
+    # Create the ParameterTypeSet element
+    parameter_type_set = Et.SubElement(telemetry_metadata, "xtce:ParameterTypeSet")
+
+    # Create the ParameterSet element
+    parameter_set = Et.SubElement(telemetry_metadata, "xtce:ParameterSet")
+
+    # Extract unique values from the 'lengthInBits' column
+    unique_lengths = pkt["lengthInBits"].unique()
```
GFMoraga: @laspsandoval I saw you code and wanted to show you this 86-114. It finds the BYTE dataType and doesn't print all UINT#'s like my previous code.

GFMoraga:
@maxinelasp look at the class I made in `tools/xtce_generation` called `telemetry_generator.py` 
I only updated `sw_species_counts.py` so lmk what you think of that set up!

imap_processing/packet_definitions/codice/lo_sw_species_counts.xml
```diff
@@ -133,24 +133,45 @@
 						<xtce:Comparison parameterRef="PKT_APID" value="1156" useCalibratedValue="false" />
 					</xtce:RestrictionCriteria>
 				</xtce:BaseContainer>
+				<xtce:EntryList />
+			</xtce:SequenceContainer>
+		</xtce:ContainerSet>
+		<xtce:ContainerSet>
+			<xtce:SequenceContainer name="CCSDSPacket">
 				<xtce:EntryList>
-					<xtce:ParameterRefEntry parameterRef="Spin_Period" />
```
maxinelasp: It looks like this did change the XTCE output - is this still what you want the XTCE to look like?

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
+import xml.etree.ElementTree as Et
+
+import pandas as pd
+
+from tools.xtce_generation.ccsds_header_xtce_generator import CCSDSParameters
+
+
+class TelemetryGenerator:
```
maxinelasp, line 9: Nice, this looks great!!

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+        comparison = Et.SubElement(restriction_criteria, "xtce:Comparison")
+        comparison.attrib["parameterRef"] = "PKT_APID"
+        comparison.attrib["value"] = apid
+        comparison.attrib["useCalibratedValue"] = "false"
+
+        Et.SubElement(codice_science_container, "xtce:EntryList")
+
+        return telemetry_metadata
+
+    def add_codice_science_parameters(self, codice_science_container, pkt):
```
maxinelasp: Is this codice specific? It looks pretty generic to me. If you move this to "science_parameters" and "science_container", then this whole class becomes non-instrument specific. 

That being said, maybe this is exclusively codice, in which case you can leave it as is :)

maxinelasp (commented):
This looks a lot better, nice job! I'll approve when you have updated the other generation files to use this shared class.

imap_processing/codice/L0/sw_species_counts.py
```diff
@@ -1,220 +1,30 @@
...
-    # Create the XML tree
-    tree = Et.ElementTree(root)
-    Et.indent(tree, space="\t", level=0)
 
-    # Save the XML document to a file in the current working directory
-    # imap_processing/packet_definitions/codice/
-    output_xml_path = "../../packet_definitions/codice/lo_sw_species_counts.xml"
-    tree.write(output_xml_path, encoding="utf-8", xml_declaration=True)
+if __name__ == "__main__":
+    main()
```
maxinelasp, line 30: It makes me happy to see this file go from 200 lines of code to 30 :)

tools/xtce_generation/telemetry_generator.py
tech3371: I agree. Really nice work!

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+        pkt.fillna("", inplace=True)
+
+        # Create the root element and add namespaces
+        root = Et.Element("{http://www.omg.org/space}SpaceSystem")
+        root.attrib["name"] = self.packet_name
+
+        # Create the Header element with attributes 'date', 'version', and 'author'
+        # Versioning is used to keep track of changes to the XML file.
+        header = Et.SubElement(root, "{http://www.omg.org/space}Header")
+        header.attrib["date"] = "2023-09"
```
tech3371: Should we get date from the day the XTCE was created using python `datetime`?

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+        header.attrib["version"] = "1.0"
+        header.attrib["author"] = "IMAP SDC"
+
+        # Create the TelemetryMetaData element
+        telemetry_metadata = Et.SubElement(
+            root, "{http://www.omg.org/space}TelemetryMetaData"
```
tech3371, line 61: Should we assigned `http://www.omg.org/space` to a variable because if this link changes, we can easily replace it in one place.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+        - parameter_type_set: The ParameterTypeSet element.
+        - parameter_set: The ParameterSet element.
+        - telemetry_metadata: The TelemetryMetaData element.
+        """
+        # Register the XML namespace
+        Et.register_namespace("xtce", "http://www.omg.org/space")
+
+        # Load data from Excel file
+        xls = pd.ExcelFile(self.path_to_excel_file)
+        pkt = xls.parse(self.packet_name)
```
maxinelasp, line 40: Looks like you're generating `pkt` here using `self.path_to_excel_file` and `self.packet_name`, but you take in `pkt` as a variable to your `init` function (line 9). I'd suggest removing `self.path_to_excel_file` and `self.packet_name` from the class, and just using `self.pkt`.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+        Parameters:
+        - parameter_type_set: The ParameterTypeSet element where parameter types are.
+        - unique_lengths: Unique values from the 'lengthInBits' column.
+        - data_type: The data type for the "Data" mnemonic.
+        - sci_byte: The BYTE number of lengthInBits in the "Data" mnemonic
+
+        Returns:
+        - parameter_type_set: The updated ParameterTypeSet element.
+        """
+        for size in unique_lengths:
```
tech3371, line 135: Can we add comment that this for loop is going through all uint data type and creating xtce:IntegerParameterType for those uint?

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+                parameter_type_set, "xtce:IntegerParameterType"
+            )
+            parameter_type.attrib["name"] = f"uint{size}"
+            parameter_type.attrib["signed"] = "false"
+
+            encoding = Et.SubElement(parameter_type, "xtce:IntegerDataEncoding")
+            encoding.attrib["sizeInBits"] = str(size)
+            encoding.attrib["encoding"] = "unsigned"
+
+        if data_type == "BYTE":
```
tech3371: Here, it seems like user need to input science_byte sizeInBits if packet has science data of datatype BYTE. Is that right? can we raise error if user forgot to input science_byte?

tools/xtce_generation/telemetry_generator.py
maxinelasp: Or, since `pkt` is optional, you could make the path and packet name optional as well, and in `__init__` assign `self.pkt` to the provided packet dataframe if it exists, and otherwise, generate/parse it from the excel file and packet name. 


```
if pkt:
    self.pkt = pkt
else:
   if self.path_to_excel_file and self.packet_name:
        xls = pd.ExcelFile(self.path_to_excel_file)
        self.pkt = xls.parse(self.packet_name)
   else:
        raise TypeError("Error creating TelemetryGenerator: either a packet dataframe OR a packet name and excel file path must be provided!")
```

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+            telemetry_xml_root,
+            parameter_type_set,
+            parameter_set,
+            telemetry_metadata,
+        ) = self.create_telemetry_xml()
+        unique_lengths, data_type = self.extract_data_info()
+        parameter_type_set = self.create_parameter_types(
+            parameter_type_set, unique_lengths, data_type, self.sci_byte
+        )
+        parameter_set = self.create_ccsds_packet_parameters(
```
tech3371, line 330: I am having little trouble following this. We use `parameter_set` often and it's confusing me little bit. Here, is it creating CCSDS base container or just the entryList of CCSDS containerset?

tech3371: After looking at `self.create_container_set()`, I think I understand. You are creating entryList for CCSDS containerSet at this line. And then creating CCSDS containerset in `create_container_set()`.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+
+            description = Et.SubElement(parameter, "xtce:LongDescription")
+            description.text = parameter_data["description"]
+
+        return parameter_set
+
+    def create_container_set(self, telemetry_metadata, ccsds_parameters, apid):
+        """
+        Create XML elements for ContainerSet, CCSDSPacket SequenceContainer,
+        and CoDICESciencePacket SequenceContainer.
```
tech3371: should we remove CoDICE specific comment because I think this could be use for others as well, right?

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+            parameter_ref_entry = Et.SubElement(
+                ccsds_packet_entry_list, "xtce:ParameterRefEntry"
+            )
+            parameter_ref_entry.attrib["parameterRef"] = parameter_data["name"]
+
+        # Create CoDICESciencePacket SequenceContainer
+        codice_science_container = Et.SubElement(
+            container_set, "xtce:SequenceContainer"
+        )
+        codice_science_container.attrib["name"] = "CoDICESciencePacket"
```
tech3371: If we take in packet_name as an input and then assign it to this name, then we can use this to create sequence container for any data packet.

tools/xtce_generation/telemetry_generator.py
tech3371: No. It looks like this code could be reuse for other instrument as well. This function is creating EntryList of a SequenceContainer using data from row 10 and onwards from excel sheet. This means we can reuse it for others as well.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+            if index < 9:
+                continue  # Skip rows before row 10
+
+            parameter = Et.SubElement(
+                parameter_set, "{http://www.omg.org/space}Parameter"
+            )
+            parameter.attrib["name"] = row["mnemonic"]
+            parameter_type_ref = ""
+
+            if row["mnemonic"] == "Data":
```
tech3371: If we want to make this useable for other instrument, how can we generalized this. This could be called differently in different packet. Eg. Science data could be called `SCIENCE_DATA` or `DE_TOF` and so on.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,294 @@
...
+        telemetry_metadata = self.create_container_set(
+            telemetry_metadata, CCSDSParameters().parameters, self.apid
+        )
+        codice_science_container = self.create_container_set(
+            telemetry_metadata, CCSDSParameters().parameters, self.apid
+        )
+        codice_science_container = self.add_codice_science_parameters(
+            codice_science_container, self.pkt
+        )
+        parameter_set = self.create_parameters_from_dataframe(parameter_set, self.pkt)
```
tech3371: To summarize what's happening in these lines, 
1. Creating Parameter definition for CCSDS parameters, first 10 rows from excel sheet.
2. Create CCSDS container set with CCSDS parameters as EntryList. Then create empty containerset for a packet and use CCSDS containerset as base container.
3. Populate above empty containerset with EntryList using data from row 10 and onwards from excel sheet.
4. Create remaining parameter for row 10 and onwards from excel sheet.

If this is the logic, then it is non-instrument specific. I feel like we can generalized it to be non-instrument specific if we take out CoDICE specific names.

imap_processing/codice/L0/sw_species_counts.py
tech3371: will other files code will be reduce to this also?

tools/xtce_generation/telemetry_generator.py
GFMoraga: I did take out codice specific I think

tools/xtce_generation/telemetry_generator.py
maxinelasp: I just used this code to generate a GLOWS file - still need to make sure it all works, but this was the only place it got hung up! Once I fixed this line (the GLOWS mnemonic is different - BIN0, BIN01, etc ) it all ran successfully, which is great!

GFMoraga: I juist had to do this for my updated packets with"Data" and 'Event_Data',

tools/xtce_generation/telemetry_generator.py
GFMoraga: Yes! We could add an error.

maxinelasp (approved):
Awesome! I had a few suggestions, but they're minor, so you can take them or leave them. Like I mentioned, I ran this on the GLOWS packet with very minimal issues - which is great! You've certainly saved me a lot of time, thanks!

tools/xtce_generation/telemetry_generator.py
GFMoraga: @maxinelasp @tech3371 I tried to figure this out, but I think I would need guidance on how to elevate it work on the mnemonic. Would it be as simple as assigning it?

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,341 @@
...
+            # Create ArrayParameterType if parameter_type_ref is "BYTE"
+            if parameter_type_ref == "BYTE":
+                array_parameter_type = Et.SubElement(
+                    parameter_type_set, "xtce:ArrayParameterType"
+                )
+                array_parameter_type.attrib["name"] = parameter_type_ref
+                array_parameter_type.attrib["signed"] = "false"
+
+                encoding = Et.SubElement(
+                    array_parameter_type, "xtce:IntegerDataEncoding"
```
greglucas: I think this should be `BinaryDataEncoding`? @tech3371 had this in a slack thread: https://lasp.slack.com/archives/C05CRQL2LJF/p1693503000638779?thread_ts=1693501674.878519&cid=C05CRQL2LJF

```
xtce:BinaryParameterType name="sciData">
                <xtce:UnitSet/>
                <xtce:BinaryDataEncoding bitOrder="mostSignificantBitFirst">
                    <xtce:SizeInBits>
                        <xtce:FixedValue>10080</xtce:FixedValue>
                    </xtce:SizeInBits>
                </xtce:BinaryDataEncoding>
            </xtce:BinaryParameterType>
```

i.e. you don't want one massive integer out, you want a binary field of 0s and 1s

GFMoraga:
Yay! I want to make it perfect, but I hear Greg saying "we will always update code"...

tools/xtce_generation/telemetry_generator.py
GFMoraga: Would this work on all instruments? This a generalized way?

greglucas: I am saying that BYTE is not an Integer parameter type, and it is just a field we know nothing about, so we want to leave it as a binary stream and do the processing on it later. So I think this is more general and is what instruments are expecting.

I think it would just be updating a few of the above variable settings you have from "Integer" to "Binary", not using those fixed values.

tools/xtce_generation/telemetry_generator.py
GFMoraga: Done.

tools/xtce_generation/telemetry_generator.py
GFMoraga: yes

tools/xtce_generation/telemetry_generator.py
GFMoraga: done

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,341 @@
...
+        Parameters:
+        - pkt: The DataFrame containing packet data.
+
+        Returns:
+        - unique_lengths: Unique values from the 'lengthInBits' column.
+        - data_type_data: The data type for the "Data" mnemonic
+        - data_type_event_data: The data type for the "Event_Data" mnemonic
+        """
+        # Extract unique values from the 'lengthInBits' column
+        unique_lengths = self.pkt["lengthInBits"].unique()
```
maxinelasp, line 86:
```suggestion
        unique_lengths = self.pkt["lengthInBits"].unique().astype(int)
```
I noticed some issues with it producing types like `uint8.0`, stemming from the fact that this value is stored as a float instead of an int. So anywhere that this "length in bits" column is, it should be an int.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -0,0 +1,341 @@
...
+                parameter_set, "{http://www.omg.org/space}Parameter"
+            )
+            parameter.attrib["name"] = row["mnemonic"]
+            parameter_type_ref = ""
+
+            # Use BYTE type for "Data" and "Event_Data" mnemonics
+            if row["mnemonic"] in ["Data", "Event_Data"]:
+                parameter_type_ref = "BYTE"
+            else:
+                parameter_type_ref = f"uint{row['lengthInBits']}"  # Use UINT for others
```
maxinelasp, line 294:
```suggestion
                parameter_type_ref = f"uint{int(row['lengthInBits'])}"  # Use UINT for others
```

Same as above. If it would be possible to get this value from the `unique_lengths` variable above, it's better to only parse through the dataframe once exactly because of these type issues. But, it doesn't matter for this review - just wanted to note that for the future.

PR #128, opened 2023-09-07
L0 Decom for Ultra

### LINKED ISSUE #65: Ultra: L0 Parsing
### Summary

_No response_

### Requirements

The more specific the better.
  - 8 APIDs required (45/90):
  - Single Events: 0x380, 0x3c0
  - Auxiliary Data: 0x370, 0x3b0
  - Image Rates: 0x371, 0x3b1
  - TOF Images: 0x373, 0x3b3

The location of these
IMAPCATS-1 : https://lasp.colorado.edu/flares/browse/IMAPCATS-1

TLM_U90FM_2023-08-16-1.xls
* U90_AUXILIARY, 0x3B0, 944
* U90_IMAGE_RATES, 0x3B1, 945
* U90_IMG_ENA_PHXTOF_HI_ANG, 0x3B3, 947
* U90_IMG_RAW_EVENTS, 0x3C0, 960

TLM_U45FM_2023-08-22.xls
- U45_IMG_RAW_EVENTS, 0x380, 896
- U45_AUXILLARY, 0x370, 880
- U45_IMAGE_RATES, 0x371, 881
- U45_IMG_ENA_PHXTOF_HI_ANG, 0x373, 883

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
This is still a work in progress! Right now this code decoms two apids for Ultra: 880 and 881. 

TODOs:
- For now the xtces are kept separate, but will be combined later
- Hard-coded values will be changed so they aren't hard-coded
- Code and test data for at least two more apids will be available for the next PR

## New Files
- conftest.py: tells pytest where data_path_fixtures.py is
- data_path_fixtures.py: clean way to organize the paths for pytest
- decom_ultra.py: fast-decom and log decom based on what was described in Ultra Algorithm Doc.
- test_apid_880.py and test_apid_881.py: tests that decom is being performed properly
- test_data contains CCSDS and csv files describing what went into CCSDS
- packet_definitions directory includes the xtce's created so far 

## Deleted Files
None

## Updated Files
- decom.py: added a function to check if we are reading off the end of the data.

## Testing
- Please only check P_U45_AUXILIARY.xml and P_U45_IMAGE_RATES.xml. The others haven't been tested. Note these xml's will be combined later. I just keep them separate as I create the decom code because it is easier to troubleshoot.


imap_processing/ultra/tests/unit/test_decom_apid_880.py
GFMoraga, line 1: These unit tests look great! I have not tryed them yet, but the functions are straight forward!

imap_processing/ultra/tests/unit/test_decom_apid_881.py
```diff
@@ -0,0 +1,42 @@
...
+    for time in decom_ultra.time.values:
+        assert df.loc[time].SID == decom_ultra['sid_data'].sel(time=time)
+        assert df.loc[time].Spin == decom_ultra['spin_data'].sel(time=time)
+        assert df.loc[time].AbortFlag == decom_ultra['abortflag_data'].sel(time=time)
+        assert df.loc[time].StartDelay == decom_ultra['startdelay_data'].sel(time=time)
+
+        arr1 = ast.literal_eval(df.loc[time].Counts)
+        arr2 = decom_ultra['fastdata_00'].sel(time=time).data
+
+        np.testing.assert_array_equal(arr1, arr2.item()[0:48])
```
laspsandoval: Just a note: The test data contains 48 values in each array. Our resulting data had ~52-60 values. It is unclear why that is and I have it as a TODO to figure it out.

imap_processing/decom.py
```diff
@@ -1,5 +1,37 @@
+# Standard
+import logging
+
+from bitstring import ReadError
+
+# Installed
```
greglucas: I still am not a huge fan of these comments in every file, but if you're going to include them then I think this should be above `bitstring` since that is also an extra and not from the standard library.

imap_processing/decom.py
```diff
@@ -1,5 +1,37 @@
...
 from space_packet_parser import parser, xtcedef
 
+logging.basicConfig(level=logging.INFO)
```
greglucas: I don't think you need this because you are only logging ERROR below and this will override if anyone else wants something different.

imap_processing/decom.py
```diff
@@ -1,5 +1,37 @@
...
+
+
+def safe_packet_generator(packet_parser, binary_data):
```
greglucas: I'm a little confused by this. I think this will actually keep going because you are catching the error and just logging it, but not re-raising.

More generally, I'm curious which behavior we actually want here? If we get a bad packet do we want to log and continue, possibly creating a file that is missing data and our log files are the only thing telling us that. Or do we want to raise and not produce a data file at all, so someone would have to come in and update the bad data.

imap_processing/decom.py
```diff
@@ -21,5 +53,7 @@ def decom_packets(packet_file: str, xtce_packet_definition: str):
     packet_parser = parser.PacketParser(packet_definition)
 
     with open(packet_file, "rb") as binary_data:
-        packet_generator = packet_parser.generator(binary_data)
-        return list(packet_generator)
+        packets = [packet for packet in safe_packet_generator(
+            packet_parser, binary_data)]
```
greglucas:
```suggestion
        return list(safe_packet_generator(packet_parser, binary_data))
```

imap_processing/swe/tests/test_decom.py
```diff
@@ -1,3 +1,5 @@
+import os
```
greglucas: We should stick with pathlib throughout the library for consistency.
```suggestion
import pathlib
```

imap_processing/swe/tests/test_decom.py
```diff
@@ -6,7 +8,9 @@
 @pytest.fixture(scope="session")
 def decom_test_data():
     """Read test data from file"""
-    packet_file = "imap_processing/swe/tests/science_block_20221116_163611Z_idle.bin"
+    current_directory = os.path.dirname(os.path.abspath(__file__))
+    packet_file = os.path.join(
+        current_directory, "science_block_20221116_163611Z_idle.bin")
```
greglucas:
```suggestion
    packet_file = pathlib.Path(__file__).parent / "science_block_20221116_163611Z_idle.bin"
```

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
+# Standard
+import logging
+
+import numpy as np
+
+# Installed
```
greglucas: numpy is also installed.

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+#TODO: This is a work in progress. Hard-coded values will be removed.
+def read_n_bits(binary: str, n: int, current_position: int):
+    """
+    Extract a specified number of bits from a binary string,
+    starting from a given position.
+
+    Parameters
+    ----------
+    binary : str
+        The binary string from which bits will be read.
```
greglucas:
```suggestion
        The string of binary data from which bits will be read.
```

"Binary string" has a special meaning in python `b"test"`, is that what this is? Or is it just a string of 0s and 1s?

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+    Returns
+    -------
+    value : int
+        The integer representation of the read bits or None if the end of the
+        string is reached before reading 'n' bits.
+    current_position + n
+        - The updated position in the binary string after reading the bits.
+    """
+
+    # Ensure we don't read past the end
```
greglucas: raise IndexError to indicate that we read past the end? Otherwise this is sort of hidden and we may not know we implemented something wrong.

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+    if current_position + n > len(binary):
+        return None, current_position
+
+    value = int(binary[current_position:current_position + n], 2)
+    return value, current_position + n
+
+def log_decompression(value: int) -> int:
+
+    """
+    Perform logarithmic decompression on a 16-bit integer.
```
greglucas: It might be helpful to put a reference to the exact page/line number you're getting this decompression equation from for future info.

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+        A binary string containing the compressed data.
+
+    Returns
+    -------
+    list
+        A list of decompressed values.
+    """
+    current_position = 0
+    decompressed_values = []
+    width_bit = 5  # The bit width that describes the width of data in the block
```
greglucas: The width is 5 bits, but your decompression is based on 16-bit integers.

Does this mean there is really 5 16-bit integers. Or there are 16 5-bit integers? Your for-loop below with range(16) makes me think you are reading 16 5-bit integers which doesn't seem quite right to me...

imap_processing/ultra/tests/conftest.py
```diff
@@ -0,0 +1,5 @@
+"""Pytest fixtures"""
+
+pytest_plugins = [
```
greglucas: Why do we need the plugins, can't you just put the fixtures directly in this file?

imap_processing/ultra/tests/plugins/data_path_fixtures.py
```diff
@@ -0,0 +1,62 @@
...
+import pytest
+
+from imap_processing import packet_definition_directory
+
+
+@pytest.fixture()
+def test_data_path():
+    """Returns the Path to the test_data directory"""
+    return Path(sys.modules[__name__.split(
+        '.')[0]].__file__).parent / 'ultra'/ 'tests' / 'test_data'
```
greglucas:
```suggestion
    return Path(__file__).parent.parent / 'ultra'/ 'tests' / 'test_data'
```
But note you'd only need one parent if you move this over to conftest.

imap_processing/ultra/tests/plugins/data_path_fixtures.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    return test_data_path / 'L0' / \
+        'Ultra45_EM_SwRI_Cal_Run7_ThetaScan_20220530T225054.CCSDS'
+
+@pytest.fixture()
+def xtce_image_rates_path():
+    """Returns the xtce image rates directory.
+    """
+    return f"{packet_definition_directory}ultra" \
+           f"/P_U45_IMAGE_RATES.xml"
```
greglucas: These are strings, do you want paths?

```suggestion
    return Path(packet_definition_directory) / "ultra" / "P_U45_IMAGE_RATES.xml"
```

imap_processing/ultra/tests/unit/test_decom_apid_880.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def test_ultra_apid_880_length(decom_test_data):
+    """Test if total packets in data file is correct"""
+    total_packets = 23
+    assert len(decom_test_data) == total_packets
+
+
+def test_enumerated_apid_880(decom_test_data):
+    """Test if enumerated values derived correctly"""
```
greglucas, line 23: This is really just testing the XML definitions. We don't know that any of these paths are actually being hit because they are guarded by if-blocks.

I am not sure how much value there is here versus actually asserting against a _specific_ packet that you know about.

```python
packet = decom_test_data[0]

assert packet.data["SPINPERIODVALID"].raw_value == 0
assert packet.data["SPINPERIODVALID"].derived_value == 'INVALID'

File: imap_processing/ultra/tests/unit/test_decom_apid_880.py [hunk 17]
```diff
@@ -0,0 +1,80 @@
...
+
+
+def test_ultra_apid_880(decom_test_data, xtce_aux_test_path):
+    """Test values for apid 880"""
+
+    df = pd.read_csv(xtce_aux_test_path, index_col='MET')
+
+    for packet in decom_test_data:
+
+        time = packet.data['SHCOARSE'].derived_value
```
greglucas, line 51: Get the sub dataframe so you don't have to do the lookup every time below? `subdf = df.loc[time]`

imap_processing/ultra/tests/unit/test_decom_apid_881.py
```diff
@@ -0,0 +1,42 @@
...
+    """Test if total packets in data file is correct"""
+    total_packets = 22
+    assert len(decom_test_data) == total_packets
+
+def test_ultra_apid_881(decom_ultra, xtce_image_rates_test_path):
+    """Test values for apid 881"""
+
+    df = pd.read_csv(xtce_image_rates_test_path, index_col='MET')
+
+    for time in decom_ultra.time.values:
```
greglucas: Do you need to iterate through time, or can you assert the arrays are the same?
```
assert_array_equal(df["SID"], decom_ultra["sid_data"])
```

If you do need to iterate, I'd suggest doing the selection only once and then re-using that selection. there is a lot of duplicated `[time]` and `sel(time=time)`.

imap_processing/ultra/tests/unit/test_decom_apid_881.py
```diff
@@ -0,0 +1,42 @@
...
+        assert df.loc[time].SID == decom_ultra['sid_data'].sel(time=time)
+        assert df.loc[time].Spin == decom_ultra['spin_data'].sel(time=time)
+        assert df.loc[time].AbortFlag == decom_ultra['abortflag_data'].sel(time=time)
+        assert df.loc[time].StartDelay == decom_ultra['startdelay_data'].sel(time=time)
+
+        arr1 = ast.literal_eval(df.loc[time].Counts)
```
greglucas: Why do we need `ast.literal_eval` here? Is Counts a string that looks like `"[1, 2, 3]"` and you're trying to turn that into a list for comparison? Is there a way to load that within pandas instead?

pyproject.toml
```diff
@@ -29,6 +29,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
+xarray = '>=2023.0.0'
```
greglucas: xarray bring in pandas as a dependency, but if you are explicitly importing pandas it'd be good to add that to the optional test dependencies as well.

greglucas (commented):
Looking good! Lots of minor comments, but the overall structure looks good to me.

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+        dimension.
+    """
+
+    packets = decom.decom_packets(packet_file, xtce)
+
+    met_data, sid_data, spin_data, abortflag_data, startdelay_data, fastdata_00_data = \
+        [], [], [], [], [], []
+
+    for packet in packets:
+        if packet.header['PKT_APID'].derived_value == 881:
```
sdhoyt: Just a suggestion, but I created enums for the APIDs to make the code a little more readable. So instead having the number in the code, I have: 

```python
if packet.header['PKT_APID'].derived_value == LO_APID.ILO_RAW_DE.value:
```

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+
+# Installed
+import xarray as xr
+
+# Local
+from imap_processing import decom
+
+logging.basicConfig(level=logging.INFO)
+#TODO: This is a work in progress. Hard-coded values will be removed.
+def read_n_bits(binary: str, n: int, current_position: int):
```
sdhoyt: Is this being used to handle variable length fields?

imap_processing/ultra/decom_ultra.py
sdhoyt, line 139: For Lo I created an enum class with the APID numbers rather than putting the numbers in the code. So it'll look like:
```python
if packet.header['PKT_APID'].derived_value == LO_APID.ILO_RAW_DE.value:
```
Not saying you should definitely do this, but I think it makes the code a little more readable.

imap_processing/ultra/tests/unit/test_decom_apid_880.py
```diff
@@ -0,0 +1,80 @@
...
+    total_packets = 23
+    assert len(decom_test_data) == total_packets
+
+
+def test_enumerated_apid_880(decom_test_data):
+    """Test if enumerated values derived correctly"""
+
+    for packet in decom_test_data:
+
+        if packet.data["SPINPERIODVALID"].raw_value == 0:
```
sdhoyt: Should these be split up into separate tests? It's hard to tell from looking at the test if all of these if statements are being hit from the packets in the loop.

sdhoyt (commented):
Great work! I'm excited to steal a lot of this lol

imap_processing/packet_definitions/ultra/P_U45_IMG_ENA_PHXTOF_HI_ANG.xml
```diff
@@ -0,0 +1,204 @@
+<?xml version='1.0' encoding='utf-8'?>
+<xtce:SpaceSystem xmlns:xtce="http://www.omg.org/space" nsmap="{'xtce': 'http://www.omg.org/space'}" name="P_U45_IMG_ENA_PHXTOF_HI_ANG">
+	<xtce:Header date="2023-08-23" version="1.0" author="Laura Sandoval" />
```
tech3371: I think we decided to use "IMAP SDC" as author for all packet definitions.

imap_processing/ultra/decom_ultra.py
tech3371: I didn't think of that. I agree it's more readable. If we want to do it this way, I will need to change mine too.

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+    -------
+    xr.Dataset
+        A dataset containing the decoded data fields with 'time' as the coordinating
+        dimension.
+    """
+
+    packets = decom.decom_packets(packet_file, xtce)
+
+    met_data, sid_data, spin_data, abortflag_data, startdelay_data, fastdata_00_data = \
+        [], [], [], [], [], []
```
tech3371:
```suggestion
    met_data, sid_data, spin_data, abortflag_data, startdelay_data, fastdata_00_data = []
```

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+    list
+        A list of decompressed values.
+    """
+    current_position = 0
+    decompressed_values = []
+    width_bit = 5  # The bit width that describes the width of data in the block
+
+    while current_position < len(binary):
+        # Read the width of the block
+        width, current_position = read_n_bits(binary, width_bit, current_position)
```
tech3371: I read in the `read_n_bits` doc string that it's an int value of read bits. Why is that int value is used as width? Just curious why.

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+            spin_data.append(packet.data['SPIN'].derived_value)
+            abortflag_data.append(packet.data['ABORTFLAG'].derived_value)
+            startdelay_data.append(packet.data['STARTDELAY'].derived_value)
+            decompressed_data = decompress_binary(packet.data['FASTDATA_00'].raw_value)
+            fastdata_00_data.append(decompressed_data)
+
+    fastdata_00_array = np.array(fastdata_00_data, dtype=object)
+
+    ds = xr.Dataset({
+        'sid_data': ('time', sid_data),
```
tech3371: Can we spell this out sid as science_id? That will help what we are storing here for those who didn't look at packet definition.

imap_processing/ultra/tests/plugins/data_path_fixtures.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.fixture()
+def xtce_aux_path():
+    """Returns the xtce auxilliary directory.
+    """
+    return f"{packet_definition_directory}ultra/P_U45_AUXILIARY.xml"
```
tech3371: since we decided to user this path, `imap_processing/{instrument}/packet_definitions/` for packet definition, this will change now, right?

imap_processing/ultra/tests/unit/test_decom_apid_880.py
tech3371: I agree. We can use first packet from `decom_test_data` and use that data in assert. If enums are correct in one packet, it should be correct in others as well.

imap_processing/ultra/tests/unit/test_decom_apid_880.py
```diff
@@ -0,0 +1,80 @@
...
+            assert packet.data["SPINPHASEVALID"].derived_value == 'VALID'
+        if packet.data["SPINPERIODSOURCE"].raw_value == 0:
+            assert packet.data["SPINPERIODSOURCE"].derived_value == 'SAFING'
+        elif packet.data["SPINPERIODSOURCE"].raw_value == 1:
+            assert packet.data["SPINPERIODSOURCE"].derived_value == 'NOMINAL'
+        if packet.data["CATBEDHEATERFLAG"].raw_value == 0:
+            assert packet.data["CATBEDHEATERFLAG"].derived_value == 'UNFLAGGED'
+
+
+def test_mode_apid_880(decom_test_data):
```
tech3371: what packet is `apid_880`? Can we use the packet name instead of the appid? Eg. `test_mode_in_image_rate_packet`

imap_processing/ultra/tests/unit/test_decom_apid_880.py
```diff
@@ -0,0 +1,80 @@
...
+        elif packet.data["IMCENB"].raw_value == 1:
+            assert packet.data["IMCENB"].derived_value == 'MODE1'
+        if packet.data["LEFTDEFLECTIONCHARGE"].raw_value == 0:
+            assert packet.data["LEFTDEFLECTIONCHARGE"].derived_value == 'MODE0'
+        elif packet.data["LEFTDEFLECTIONCHARGE"].raw_value == 1:
+            assert packet.data["LEFTDEFLECTIONCHARGE"].derived_value == 'MODE1'
+        if packet.data["RIGHTDEFLECTIONCHARGE"].raw_value == 0:
+            assert packet.data["RIGHTDEFLECTIONCHARGE"].derived_value == 'MODE0'
+        elif packet.data["RIGHTDEFLECTIONCHARGE"].raw_value == 1:
+            assert packet.data["RIGHTDEFLECTIONCHARGE"].derived_value == 'MODE1'
```
tech3371: Same thing as above. We can run assert on first data only.

imap_processing/ultra/tests/unit/test_decom_apid_880.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def test_ultra_apid_880(decom_test_data, xtce_aux_test_path):
+    """Test values for apid 880"""
```
tech3371:
```suggestion
    """This function reads validation data and checks that decom data matches validation data for auxiliary packet"""
```

imap_processing/ultra/tests/unit/test_decom_apid_880.py
```diff
@@ -0,0 +1,80 @@
...
+            assert packet.data["LEFTDEFLECTIONCHARGE"].derived_value == 'MODE0'
+        elif packet.data["LEFTDEFLECTIONCHARGE"].raw_value == 1:
+            assert packet.data["LEFTDEFLECTIONCHARGE"].derived_value == 'MODE1'
+        if packet.data["RIGHTDEFLECTIONCHARGE"].raw_value == 0:
+            assert packet.data["RIGHTDEFLECTIONCHARGE"].derived_value == 'MODE0'
+        elif packet.data["RIGHTDEFLECTIONCHARGE"].raw_value == 1:
+            assert packet.data["RIGHTDEFLECTIONCHARGE"].derived_value == 'MODE1'
+
+
+def test_ultra_apid_880(decom_test_data, xtce_aux_test_path):
```
tech3371:
```suggestion
def test_ultra_auxilary_packet(decom_test_data, xtce_aux_test_path):
```

imap_processing/ultra/tests/unit/test_decom_apid_881.py
```diff
@@ -0,0 +1,42 @@
...
+    data_packet_list = decom.decom_packets(ccsds_path, xtce_image_rates_path)
+    return data_packet_list
+
+@pytest.fixture()
+def decom_ultra(ccsds_path, xtce_image_rates_path):
+    """Data for decom_ultra"""
+    data_packet_list = decom_ultra_packets(ccsds_path,  xtce_image_rates_path)
+    return data_packet_list
+
+def test_ultra_apid_881_length(decom_test_data):
```
tech3371:
```suggestion
def test_ultra_image_rates_length(decom_test_data):
```

tech3371 (commented):
I had some comments. Nice work!

imap_processing/ultra/decom_ultra.py
bryan-harter: It looks to me like the initial 5 bits of these 1's and 0's always describe the width of the next 16 values.  So the first 5 bits could hold the value of "10".  So then a total of 16*10 bits are read in, and converted to 16 numbers. And after those values are read, it starts over again with 5 bits describing the next 16 numbers.  

I'm guessing log_decompression() is supposed to change behavior based on the width, but right now it's only set to decompress if the width is set to 16.  Perhaps all of the initial data sent to us had a 16 bit compression?

Correct me if I'm wrong Laura, that's just my interpretation of what appears to be going on here!  Will there be other widths to decompress?

imap_processing/ultra/decom_ultra.py
bryan-harter: Does that actually work? I just tried it in a python terminal and it said `not enough values to unpack`

imap_processing/ultra/decom_ultra.py
```diff
@@ -0,0 +1,160 @@
...
+        'spin_data': ('time', spin_data),
+        'abortflag_data': ('time', abortflag_data),
+        'startdelay_data': ('time', startdelay_data),
+        'fastdata_00': ('time', fastdata_00_array)
+    }, coords={
+        'time': met_data,
```
bryan-harter: Minor, but when it comes time to convert this to a CDF, the time dimension should be 'epoch'

imap_processing/decom.py
laspsandoval: I'm trying to think of a better way to do it. The reason I did it that way was that my packets were all the same length (848 bits), but the very last packet had incomplete data. I assume this is when the simulator was turned off. So it had fewer bits than expected. I created safe_packet_generator to handle that last packet.

Do you think there would be real (not simulated) packets that are cutoff like that during normal processing? 

I am changing it to this, but not certain if that solves the problem. Thoughts?

    try:
        yield from packet_parser.generator(binary_data)
    except ReadError as e:
        if "Reading off the end of the data." in str(e):
            logging.warning(f"Acceptable read error encountered: {e}")
        else:
            logging.error(f"Unexpected read error: {e}")
            raise

greglucas: I didn't realize ReadError is actually from bitstring and not from space_packet_parser. I feel like space_packet_parser should actually not yield this packet if it isn't the right size to be able to read all of the data?

> Do you think there would be real (not simulated) packets that are cutoff like that during normal processing?
No, which is why I am somewhat worried that this could hide the fact that we got bad packets down during the mission by just logging and continuing on rather than raising and actually raising alarms on our end.
I see why you need to do this in this situation since you presumably got incomplete data though.

Some thoughts:
1) Did you verify with Ultra that the final packet is the wrong length? Maybe there is something else we need to be aware of and this caught something that went wrong on their end.
2) Can you strip that binary file down to be the right length and remove that final packet from the original file? Then you wouldn't need to do this dance.

imap_processing/ultra/decom_ultra.py
laspsandoval: It looks to me like the initial 5 bits of these 1's and 0's always describe the width of the next 16 values. So the first 5 bits could hold the value of "10". So then a total of 16*10 bits are read in, and converted to 16 numbers. And after those values are read, it starts over again with 5 bits describing the next 16 numbers.
**(For what is known as "Fast Decompression": Yes you've got it! Other apids will have other widths so that is something I intend to change in the next PR for other apids.)**

I'm guessing log_decompression() is supposed to change behavior based on the width, but right now it's only set to decompress if the width is set to 16. Perhaps all of the initial data sent to us had a 16 bit compression?
**You are right. I only did this on data that had 16 bit log decompression. Other apids have 10 bit log decompression, so this will need to be updated in the next PR**

imap_processing/decom.py
laspsandoval: I've emailed them (CC'd you). We'll see what they say.

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -0,0 +1,192 @@
...
+    current_position = 0
+    decompressed_values = []
+
+    while current_position < len(binary):
+        # Read the width of the block
+        width, current_position = read_n_bits(binary, width_bit, current_position)
+        # If width is 0 or None, we don't have enough bits left
+        if width is None or len(decompressed_values) >= \
+                UltraParams.ULTRA_IMG_RATES.value.len_array:
+            print('hi')
```
tech3371, line 127: is that a debug print statement? I do that too :)

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -0,0 +1,192 @@
...
+    array_data = np.array(fastdata_00)
+
+    ds = xr.Dataset({
+        'science_id': ('epoch', science_id),
+        'spin_data': ('epoch', spin_data),
+        'abortflag_data': ('epoch', abortflag_data),
+        'startdelay_data': ('epoch', startdelay_data),
+        'fastdata_00': (('epoch', 'index'), array_data)
+    }, coords={
+        'epoch': met_data,
```
tech3371, line 189: @bryan-harter Does it matter is it's lower case or not? Asking for my future self.

imap_processing/ultra/tests/unit/data_path_fixtures.py
```diff
@@ -0,0 +1,67 @@
...
+import pytest
+
+
+@pytest.fixture()
+def ccsds_path():
+    """Returns the ccsds directory.
+    """
+    return Path(sys.modules[__name__.split(
+        '.')[0]].__file__).parent / 'ultra' / 'tests' / 'test_data' / 'l0' / \
+        'Ultra45_EM_SwRI_Cal_Run7_ThetaScan_20220530T225054.CCSDS'
```
tech3371, line 14:
```suggestion
    filename = "Ultra45_EM_SwRI_Cal_Run7_ThetaScan_20220530T225054.CCSDS" #  to avoid line too long error
    return f"{imap_module_directory}/ultra/tests/test_data/l0/{filename}"
```

imap_processing/ultra/tests/unit/data_path_fixtures.py
```diff
@@ -0,0 +1,67 @@
+"""Pytest plugin module for test data paths"""
+import sys
+from pathlib import Path
+
+import pytest
+
```
tech3371, line 6:
```suggestion
 from imap_processing import imap_module_directory
```
To use in later path functions

imap_processing/ultra/tests/unit/data_path_fixtures.py
```diff
@@ -0,0 +1,67 @@
...
+        'Ultra45_EM_SwRI_Cal_Run7_ThetaScan_20220530T225054.CCSDS'
+
+
+@pytest.fixture()
+def xtce_image_rates_path():
+    """Returns the xtce image rates directory.
+    """
+    return Path(sys.modules[__name__.split(
+        '.')[0]].__file__).parent / 'ultra' / 'packet_definitions' \
+        / "P_U45_IMAGE_RATES.xml"
```
tech3371, line 23:
```suggestion
    filename = "P_U45_IMAGE_RATES.xml" #  to avoid line too long error
    return f"{imap_module_directory}/ultra/tests/test_data/l0/{filename}"
```

imap_processing/ultra/tests/unit/data_path_fixtures.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@pytest.fixture()
+def xtce_aux_path():
+    """Returns the xtce auxilliary directory.
+    """
+    return Path(sys.modules[__name__.split(
+        '.')[0]].__file__).parent / 'ultra' / 'packet_definitions' \
+        / "P_U45_AUXILIARY.xml"
```
tech3371, line 32:
```suggestion
    filename = "P_U45_AUXILIARY.xml" #  to avoid line too long error
    return f"{imap_module_directory}/ultra/tests/test_data/l0/{filename}"
```

imap_processing/ultra/tests/unit/data_path_fixtures.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@pytest.fixture()
+def xtce_image_raw_events_path():
+    """Returns the xtce image raw events directory.
+    """
+    return Path(sys.modules[__name__.split(
+        '.')[0]].__file__).parent / 'ultra' / 'packet_definitions' \
+        / "P_U45_IMG_RAW_EVENTS.xml"
```
tech3371, line 41:
```suggestion
    filename = "P_U45_IMG_RAW_EVENTS.xml" #  to avoid line too long error
    return f"{imap_module_directory}/ultra/tests/test_data/l0/{filename}"
```

imap_processing/ultra/tests/unit/data_path_fixtures.py
```diff
@@ -0,0 +1,67 @@
...
+def xtce_image_rates_test_path():
+    """Returns the xtce image rates test data directory.
+    """
+    filename = (
+        "ultra45_raw_sc_ultraimgrates_Ultra45_EM_SwRI_Cal_Run7_ThetaScan_"
+        "20220530T225054.csv"
+    )
+    return Path(sys.modules[__name__.split(
+        '.')[0]].__file__).parent / 'ultra' / 'tests' / \
+        'test_data' / 'l0' / filename
```
tech3371, line 54:
```suggestion
    return f"{imap_module_directory}/ultra/tests/test_data/l0/{filename}"
```

imap_processing/ultra/tests/unit/data_path_fixtures.py
```diff
@@ -0,0 +1,67 @@
...
+def xtce_aux_test_path():
+    """Returns the xtce auxiliary test data directory.
+    """
+    filename = (
+        "ultra45_raw_sc_auxdata_Ultra45_EM_SwRI_Cal_Run7_ThetaScan_"
+        "20220530T225054.csv"
+    )
+    return Path(sys.modules[__name__.split(
+        '.')[0]].__file__).parent / 'ultra' / 'tests' / \
+        'test_data' / 'l0' / filename
```
tech3371, line 67:
```suggestion
    return f"{imap_module_directory}/ultra/tests/test_data/l0/{filename}"
```

tech3371 (approved):
I had path code suggestions if you like to change it. Besides that it looks good to me!

greglucas:
This was failing checks when you merged. Make sure to get the green check before merging, or explain why you are merging over the failures if there is something sporadic.

I've added a commit to fix this up over in: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/171

I'd suggest installing the pre-commit hooks or IDE integrations locally so these show up right away.

bourque:
@all-contributors please add @laspsandoval for Code!

Issue #104, opened 2023-09-06
Create GLOWS XTCE Packet Definition


maxinelasp:
### Summary

Create XTCE packet definitions for use in GLOWS L1A algorithm using space packet parser. This could use the script from Gabriel for converting CSVs to XTCE. 

### Requirements

The more specific the better.
  - [ ] Create a GLOWS XTCE Packet definition

### Related tickets

- #97

maxinelasp:
#175

PR #145, opened 2023-09-11
Xtce generator

maxinelasp:
# Change Summary
Updates to the XTCE generator code to make it more generic, including some API changes. 

## Overview

Codice needs two events for scientific data from telemetry. Rather than hardcoding these two event mnemonics, I changed the code to accept a dictionary. This makes it easier to overwrite methods for instrument specific changes without needing to update the writer/generator parts. 

Making this change also required updating all the codice python files to provide the events through a dictionary instead of assuming them. 

Long term, codice could benefit from a specific CodiceTelemetryGenerator class which has codice-specific code or helper functions. I have an example of this in my GLOWS code, which is still incomplete. 

I also split up some methods which people may want to override in the future. 

Please double check my docstrings, I'm not sure if I accurately represented what the mnemonics are used for. 

## New Files

- new file 1

## Updated Files

- telemetry_generator.py
   - Moved science mnemonics to dictionary 
- codice files
  - updated to match changed api

## Testing

I checked, and the output codice file matches the existing one in the repository. So I didn't change any functionality.

### DIFF
--- a/tools/xtce_generation/telemetry_generator.py
+++ b/tools/xtce_generation/telemetry_generator.py
@@ -1,8 +1,6 @@
 import xml.etree.ElementTree as Et
 from datetime import datetime
 
-import pandas as pd
-
 from tools.xtce_generation.ccsds_header_xtce_generator import CCSDSParameters
 
 
@@ -18,14 +16,6 @@ class TelemetryGenerator:
         """
         Create an XML representation of telemetry data based on input parameters.
 
-        Parameters:
-        - packet_name: The name of the packet (sheet) in the Excel file.
-        - path_to_excel_file: The path to the Excel file with packet definitions.
-        - APID: The APID of the packet used to filter the parameters.
-        - sci_byte (optional): The BYTE number of lengthInBits in the "Data" mnemonic
-          in the CoDICE Science Packet.
-        - pkt: The DataFrame containing packet data.
-
         Returns:
         - root: The root element of the generated XML tree.
         - parameter_type_set: The ParameterTypeSet element.
@@ -34,13 +24,6 @@ class TelemetryGenerator:
         """
         # Register the XML namespace
         Et.register_namespace("xtce", "http://www.omg.org/space")
-
-        # Load data from Excel file
-        xls = pd.ExcelFile(self.path_to_excel_file)
-        pkt = xls.parse(self.packet_name)
-
-        # Fill missing values with '-*-'
-        pkt.fillna("", inplace=True)
 
         # Get the current date and time
         current_date = datetime.now().strftime("%Y-%m")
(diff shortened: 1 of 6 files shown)


maxinelasp:
Done as part of issue #104

GFMoraga (approved):
This looks great. Thank you for making these changes for more flexible use!

tools/xtce_generation/telemetry_generator.py
```diff
@@ -69,53 +60,42 @@ def create_telemetry_xml(self):
...
+        return unique_lengths, data_types
 
     def create_parameter_types(
         self,
         parameter_type_set,
         unique_lengths,
-        data_type_data,
-        data_type_event_data,
-        sci_byte,
+        data_types,
```
tech3371: Can we update the doc string of this function as well?

tools/xtce_generation/telemetry_generator.py
```diff
@@ -133,21 +113,19 @@ def create_parameter_types(
...
                     "BYTE"  # Use BYTE type for both "Data" and "Event_Data"
                 )
             elif size == self.sci_byte:
-                parameter_type_ref = (
-                    data_type_data  # Use data_type_data for "Data" mnemonic
-                )
-            elif size == self.sci_byte:
+                parameter_type_ref = data_types[
+                    "Data"
+                ]  # Use data_type_data for "Data" mnemonic
```
tech3371:
```suggestion
                parameter_type_ref = data_types[
                    "Data"
                ]  # use dataType of "Data" in data_types dict
```

tools/xtce_generation/telemetry_generator.py
```diff
@@ -133,21 +113,19 @@ def create_parameter_types(
...
         """
         for size in unique_lengths:
-            parameter_type_ref = None
+            parameter_type_ref = f"uint{size}"  # Use UINT for other sizes
 
             if size == sci_byte:
                 parameter_type_ref = (
                     "BYTE"  # Use BYTE type for both "Data" and "Event_Data"
                 )
             elif size == self.sci_byte:
```
tech3371: What is difference between this function input parameter `sci_byte` and `self.sci_byte`?

tools/xtce_generation/telemetry_generator.py
```diff
@@ -302,7 +284,20 @@ def create_parameters_from_dataframe(self, parameter_set, pkt):
...
-    def generate_telemetry_xml(self, output_xml_path):
+    def generate_telemetry_xml(
+        self, output_xml_path, mnemonic_names, science_packet_name="SciencePacket"
+    ):
+        """
+        Create and output an XTCE file based on the data within the class.
+        Parameters
+        ----------
+        output_xml_path: the path for the final xml file
+        mnemonic_names: a list or single name for mnemonics for extracting data types
```
tech3371, line 282: So this is where we give in names of science data that are specific to each instrument such as `SCIENCE_DATA` or `HIST_DATA` field in packet definition?

tech3371 (commented):
Thank you for making it more generic!

tools/xtce_generation/telemetry_generator.py
maxinelasp: Good catch! I actually removed the `sci_byte` input to this function and moved it entirely to the class variable. In the future, hopefully we can move this variable, since it is codice specific (?). In glows, I override this function and don't use it.

tools/xtce_generation/telemetry_generator.py
maxinelasp: Yes, exactly!

tech3371 (approved):

tools/xtce_generation/telemetry_generator.py
GFMoraga: It was built for packets with BYTE as a dataType. So, it is optional I think I put it as.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -125,29 +91,26 @@ def create_parameter_types(
...
                 )
             elif size == self.sci_byte:
-                parameter_type_ref = (
-                    data_type_data  # Use data_type_data for "Data" mnemonic
-                )
-            elif size == self.sci_byte:
+                parameter_type_ref = data_types[
+                    "Data"
+                ]  # use dataType of "Data" in data_types dict
+            elif data_types["EventData"] and size == self.sci_byte:
```
sdhoyt, line 111: is "EventData" a CoDICE specific thing? I'm getting an error because I don't have "EventData" defined as a key

GFMoraga: Yes it is codice specific. But you can change the mnemonic. I also think Maxine might have fixed this, so when pushed, it should work..... If this helps, I have packets with "Data" OR "Event_Data" so it was my way of calling the correct one depending on what was in the row. 

 - data_type_event_data: The data type for the "Event_Data" mnemonic

maxinelasp: Ah yes, it is codice specific. In my GLOWs code, I override it: 

```    
def create_parameter_types(
        self,
        parameter_type_set,
        unique_lengths,
        data_types,
        sci_byte=None,
    ):
        """
        Create parameter types based on 'dataType' for the unique 'lengthInBits' values.
        This will loop through the unique lengths and create a ParameterType element
        for each length representing a UINT type.

        Parameters:
        - parameter_type_set: The ParameterTypeSet element where parameter types are.
        - unique_lengths: Unique values from the 'lengthInBits' column.
        - data_type_data: The data type for the "Data" mnemonic.
        - data_type_event_data: The data type for the "Event_Data" mnemonic.
        - sci_byte: The BYTE number of lengthInBits in both "Data" and "Event_Data"

        Returns:
        - parameter_type_set: The updated ParameterTypeSet element.
        """
        for size in unique_lengths:
            # All bins equal?
            # Are bins the only thing that matters? -> check glows code for this
            parameter_type_ref = data_types["BIN"]  # Use UINT for other sizes
            print(parameter_type_ref)
            parameter_type = Et.SubElement(parameter_type_set, "xtce:ParameterType")
            parameter_type.attrib["name"] = f"uint{size}"
            parameter_type.attrib["signed"] = "false"

            encoding = Et.SubElement(parameter_type, "xtce:IntegerDataEncoding")
            encoding.attrib["sizeInBits"] = str(size)
            encoding.attrib["encoding"] = "unsigned"

            # Create BinaryParameterType if parameter_type_ref is "BYTE"
            if parameter_type_ref == "BYTE":
                binary_parameter_type = Et.SubElement(
                    parameter_type_set, "xtce:BinaryParameterType"
                )
                binary_parameter_type.attrib["name"] = parameter_type_ref

                Et.SubElement(binary_parameter_type, "xtce:UnitSet")

                binary_data_encoding = Et.SubElement(
                    binary_parameter_type, "xtce:BinaryDataEncoding"
                )
                binary_data_encoding.attrib["bitOrder"] = "mostSignificantBitFirst"

                size_in_bits = Et.SubElement(binary_data_encoding, "xtce:SizeInBits")
                fixed_value = Et.SubElement(size_in_bits, "xtce:FixedValue")
                fixed_value.text = str(sci_byte)  # Set the size in bits to sci_byte

        return parameter_type_set
```

maxinelasp: Ideally, in the future, this will be something that is probably overridden in most instruments, or the more generic code above can be modified so it can be used in all cases where there are very simple parameters.

sdhoyt (approved):
Looks good!

PR #154, opened 2023-09-12
SWE science L1A algorithm

### LINKED ISSUE #100: Implement L1A algorithm for SWE
### Summary
This ticket implements L1A algorithm for all appIds for SWE instrument.

### Requirements
Packet Names:
SWE_APP_HK
SWE_CEM_RAW
SWE_SCIENCE
SWE_EVTMSG

### LINKED ISSUE #103: Uncompress counts for SWE_SCIENCE
### Summary
This ticket implement uncompression algorithm for SWE_SCIENCE data for SWE instrument.

### Requirements

The more specific the better.
  - [x] Write XTCE to decommutate data
  - [x] Unpack raw data according to algorithm document
  - [x] Write Data to xarray dataset

tech3371:
# Change Summary
This contains code for SWE science data's l1a algorithm implementation. This is implementation of this issue, #103.

## Overview
I moved files around to match folder structure we decided to use, `imap_processing/{instrument}/packet_definitions/`.

## New Files

- swe_l1a.py
   - This code read l0 file and based on appId, it starts processing science or housekeeping data and so on.
- swe_science.py
   - This code contains decompression algorithms and stores process data to xarray dataset.

## Updated Files
- swe_packet_definition.xml
   - Updated to match latest packet definition.
- decom_swe.py and test_decom.py
   - updated decom with new folder path and update import in test file.
- pyproject.toml
   - updated it to include bitstring
## Testing
- test_swe_science.py
   - Wrote basic tests for SWE science.


tech3371:
pre-commit.ci autofix

tech3371:
I really mess it up with rebase. I am gonna close this temporarily to fix rebase issue. Will reopen ASAP

tech3371:
sorry about long and duplicate commits. I messed up during rebase. I will squash commits before merging.

GFMoraga (approved):

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    for key, value in data_packet.header.items():
+        metadata_arrays[key].append(value.raw_value)
+
+    for key, value in data_packet.data.items():
+        if key != "SCIENCE_DATA":
+            metadata_arrays[key].append(value.raw_value)
+
+
+def init_metadata_array(data_packet):
```
greglucas: I think you can remove this function and use one of the two methods below to set the initial dictionary contents in the function above:
1) `metadata_arrays = collections.defaultdict(list)` https://docs.python.org/3/library/collections.html#collections.defaultdict
2) `metadata_arrays.setdefault(key, []).append(value.raw_value)` https://docs.python.org/3.11/library/stdtypes.html#dict.setdefault

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,174 @@
...
+    metadata_arrays = init_metadata_array(decom_data[0])
+
+    for data_packet in decom_data:
+        # read raw data
+        binary_data = data_packet.data["SCIENCE_DATA"].raw_value
+        # read raw data as binary array using bitstring
+        # Eg. "0b0101010111"
+        bit_array = bitstring.ConstBitStream(bin=binary_data)
+        # chunk bit array into 1260 units, each with 8-bits
+        byte_data = np.frombuffer(bit_array.bytes, dtype=np.uint8)
```
greglucas: Do you need to create the bit_array at all? Or can you just grab the bytes from binary_data directly? I don't know what the `.bytes` attribute is doing if it returns utf-8 bytes or something else, but I'm guessing there is a way to avoid bitstring here.

```suggestion
        # chunk bit array into 1260 units, each with 8-bits
        byte_data = np.frombuffer(bytes(binary_data), dtype=np.uint8)
```

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,174 @@
...
+        # Uncompress counts. Uncompressed data is a list of 1260
+        # where 1260 = 15 seconds x 7 CEMs x 12 energy steps
```
greglucas: Are we sure the below reshapes the data in the proper order? Here you have `15 x 7 x 12`, but you reshape it below as `15 x 12 x 7`.

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,174 @@
...
+        uncompressed_data = [uncompress_counts(i) for i in byte_data]
+        reshaped_data = np.array(uncompressed_data).reshape(15, 12, 7)
```
greglucas: I think I'd still stick to numpy as much as possible here, so creating a decompression table above once and re-using it for lookups here.

somewhere above where you only run it in the module once near the function definition:
```python
# We know we can only have 8 bit numbers input, so iterate over all possibilities once up front
decompression_table = np.array([uncompress_counts(i) for i in range(256)]
```

Then inside this loop you can do something like this:
```python
# Take the "byte_data" indices/counts mapping from decompression_table and then reshape the return
reshaped_data = np.take(decompression_table, byte_data).reshape(15, 12, 7)
```

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,174 @@
...
+        add_metadata_to_array(data_packet, metadata_arrays)
+
+    science_xarray = xr.DataArray(
+        science_array,
+        dims=["number_of_packets", "seconds", "energy_steps", "cem_counts"],
+    )
+
+    dataset = xr.Dataset(
+        {"SCIENCE_DATA": science_xarray},
+        coords={"met_time": metadata_arrays["SHCOARSE"]},
```
greglucas: I think we will all be using some form of "met_time", so is that we are using on all of the other instruments as well? Or do we want to call the variable "MET" which I've seen in most of the POC documents?

We should also put a short_description tag in with a label "Mission Elapsed Time" and units field "seconds since start of the mission" or something like that as well.

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+
+    # create xarray dataset for each metadata field
+    for key, value in metadata_arrays.items():
+        if key == "SHCOARSE":
+            continue
+
+        dataset[f"{key}"] = xr.DataArray(
```
greglucas: This is already a string
```suggestion
        dataset[key] = xr.DataArray(
```

greglucas (commented):
Looks good, just a few minor suggestions to move towards more numpy functions when possible, but nothing blocking and those can be made later if you want as well.

imap_processing/swe/l1a/swe_science.py
tech3371: Yeah, I tried it but it wouldn't work. If I call `bytes(binary_data)`, it expect bytes-like object. And to convert binary string to byte-like object, it looks like I need to do this
```
binary_string = "0b0110011001100001011011000110011001100101"
bytes_object = bytes(int(binary_string, 2).to_bytes((len(binary_string) + 7) // 8, byteorder='big'))
```
It looks more complicate and confusing. That's why I didn't do it. Do you know another way though?

imap_processing/swe/l1a/swe_science.py
tech3371: good catch! comment is outdate and need update.

imap_processing/swe/l1a/swe_science.py
tech3371: I meant to ask you. I will try to look at np options in the future.
I will try that.

tech3371: It took me a while to understand but now I understand it after reading up on it and trying. It's really nice and efficient!

imap_processing/swe/l1a/swe_science.py
tech3371: I am not sure on how we use this, tbh. I thought that this packet creation time could be something of use.

But I did change to have those inputs. I created xr.DataArray for `SHCOARSE` and added those attrs.

imap_processing/swe/l1a/swe_science.py
greglucas: Uggh, I forgot that it was a string :( I was thinking it was already a bytesIO object. I think you're right that is what you'd have to do. I agree it looks more complicated and confusing, but removes a potentially slow step and import. But, I think you could get rid of some of the bit padding and checks since you know the exact length in bytes already:

```python
int(s, 2).to_bytes(len(s) // 8, byteorder='big')
# Or since you know the exact length already just use it directly
int(s, 2).to_bytes(1260, byteorder='big')
```

One other question would be is `bit_array` really a string, or was it a bitstream already that the `space_packet_parser` returns to you and you could just call `binary_data.bytes` directly on that object instead?

tech3371: binary_data that `space_packet_parser` returns is a string type so I can't call `.bytes`. The example I wrote above wrong. It actually looks like this.
```
binary_string = "0000000011" # without 0b at start
```
Sorry for the confusion. But I think this suggestion makes sense.
```
int(s, 2).to_bytes(1260, byteorder='big')
```

tech3371:
```suggestion
        # read raw data
        binary_data = data_packet.data["SCIENCE_DATA"].raw_value
        # read binary string as an int and then convert it to
        # bytes.
        # Eg. "0000000011110011" --> b'\x00\xf3'
        byte_data = int(binary_data, 2).to_bytes(1260, byteorder='big')
        # convert bytes to numpy array of uint8
        raw_counts = np.frombuffer(byte_data, dtype=np.uint8)
```
@greglucas Does this change look ok?

greglucas (approved):

imap_processing/swe/l1a/swe_science.py
greglucas: Yeah. I personally think that looks better and then you can remove the explicit dependency in our requirements. But looks good either way to me!

tech3371: Perfect. I can remove import in mine but I can't remove `bitstring` from dependency because IDEX code is using it.

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,168 @@
...
+        # decompression_table and then reshape the return
+        uncompress_data = np.take(decompression_table, raw_counts).reshape(15, 12, 7)
+
+        # Save data with its metadata field to attrs and DataArray of xarray.
+        science_array.append(uncompress_data)
+        add_metadata_to_array(data_packet, metadata_arrays)
+
+    science_xarray = xr.DataArray(
+        science_array,
+        dims=["number_of_packets", "seconds", "energy_steps", "cem_counts"],
```
bryan-harter: I'm not quite sure how to change this around, but when we run the xarray->cdf, time should be in the first dimension.

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,168 @@
...
+    xarray.Dataset
+        xarray dataset with data.
+    """
+    science_array = []
+
+    metadata_arrays = collections.defaultdict(list)
+
+    # We know we can only have 8 bit numbers input, so iterate over all
+    # possibilities once up front
+    decompression_table = np.array([uncompress_counts(i) for i in range(256)])
```
bryan-harter, line 116: If the table is static and doesn't change, does it make sense to just have it live in a csv file somewhere?  I guess it's less flexible in case the table needs to be changed.  But it might help people understand what is going on a little better.

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,168 @@
...
+
+    met_time = xr.DataArray(
+        metadata_arrays["SHCOARSE"],
+        name="MET",
+        dims=["number_of_packets"],
+        attrs=dict(
+            description="Mission elapsed time",
+            units="seconds since start of the mission",
+        ),
+    )
```
bryan-harter, line 147: Is this time since launch? At some point we should convert the time to unix timestamp (seconds since 1970).  The xarray_to_cdf routine has a flag to convert unix timestamp to CDF_TT2000 time

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,168 @@
...
+        # Eg. "0000000011110011" --> b'\x00\xf3'
+        byte_data = int(binary_data, 2).to_bytes(1260, byteorder="big")
+        # convert bytes to numpy array of uint8
+        raw_counts = np.frombuffer(byte_data, dtype=np.uint8)
+
+        # Uncompress counts. Uncompressed data is a list of 1260
+        # where 1260 = 15 seconds x 12 energy steps x 7 CEMs
+        # Take the "raw_counts" indices/counts mapping from
+        # decompression_table and then reshape the return
+        uncompress_data = np.take(decompression_table, raw_counts).reshape(15, 12, 7)
```
bryan-harter, line 133: I had to look up what the np.take function does, I guess it does the same thing as the "fancy indexing" I'm used to!  I believe that `decompression_table[raw_counts]` actually returns the same thing.

imap_processing/swe/l1a/swe_science.py
bryan-harter: Actually should the first dimension be number_of_packets? It seems like "number_of_packets" could just be "met_time"? Then you don't need the intermediate dimension of "number_of_packets".  You'd also need to change the dimensions of "met_time" below to just be "met_time" (a variable is allowed to be its own dimension within xarray).

imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -0,0 +1,23 @@
...
+    packet_file : str
+        The path and filename to the L0 file to read
+    """
+    decom_data = decom_swe.decom_packets(packet_file)
+    logging.info(f"Unpacking data from {packet_file}")
+
+    # If appId is science, then the file should contain all data of science appId
+    if decom_data[0].header["PKT_APID"].raw_value == 1344:
+        logging.info("Processing science data")
+        swe_science(decom_data=decom_data)
```
bryan-harter: Should this function return the output of swe_science?

tech3371: yes. good catch. I should return the data here. Then we can think about how we go from this to CDF in the future.

imap_processing/swe/l1a/swe_science.py
tech3371: They said that it will hardly change. But do you mean, run uncompression once for 256 values and store in a csv? Then read csv and use it directly. That could work.

I will be using lookup table for l1b that could change sometimes. I was going to ask you about how to store that lookup table and I think you gave me ideas for that. Thank you!

imap_processing/swe/l1a/swe_science.py
tech3371: same! I had to lookup too.

imap_processing/swe/l1a/swe_science.py
tech3371: That's good to know. I didn't know that. I will write a suggested change based on this discussion in a new conversation.

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # create xarray dataset for each metadata field
+    for key, value in metadata_arrays.items():
+        if key == "SHCOARSE":
+            continue
+
+        dataset[key] = xr.DataArray(
+            value,
+            dims=["number_of_packets"],
+        )
```
tech3371: @bryan-harter will it look something like this when we add time to be the first dimension.
```suggestion
    met_time = xr.DataArray(
        metadata_arrays["SHCOARSE"],
        name="met_time",
        dims=["met_time"],
        attrs=dict(
            description="Mission elapsed time",
            units="seconds since start of the mission",
        ),
    )

    science_xarray = xr.DataArray(
        science_array,
        dims=["met_time", "seconds", "energy_steps", "cem_counts"],
    )

    dataset = xr.Dataset(
        {"SCIENCE_DATA": science_xarray},
        coords={"met_time": met_time},
    )

    # create xarray dataset for each metadata field
    for key, value in metadata_arrays.items():
        if key == "SHCOARSE":
            continue

        dataset[key] = xr.DataArray(
            value,
            dims=["met_time"],
        )
```

imap_processing/swe/l1a/swe_science.py
tech3371: I totally agree. I am not sure of that myself. I asked Nick Dutton and his guess is that it may be the time when we start getting first series of data packet from spacecraft or time when we start getting science data. We don't know for sure though.

imap_processing/swe/l1a/swe_science.py
```diff
@@ -0,0 +1,168 @@
...
+    # possibilities once up front
+    decompression_table = np.array([uncompress_counts(i) for i in range(256)])
+
+    for data_packet in decom_data:
+        # read raw data
+        binary_data = data_packet.data["SCIENCE_DATA"].raw_value
+        # read binary string to an int and then convert it to
+        # bytes. This is to convert the string to bytes.
+        # Eg. "0000000011110011" --> b'\x00\xf3'
+        byte_data = int(binary_data, 2).to_bytes(1260, byteorder="big")
```
sdhoyt, line 125: Where are you getting the 1260 byte length from?

sdhoyt (approved):
Great job!

imap_processing/swe/l1a/swe_science.py
greglucas: That is a good question of when we want to convert time. Because you can't get to unix timestamp from MET without using SPICE. So I think we'll need @laspsandoval's routines to handle that conversion.

Can we convert to Unix time right away in L1a? Or do we need to carry MET around in every file for the SPICE queries? We can carry around both MET and Unix time presumably, just call them different variable names.

imap_processing/swe/l1a/swe_science.py
tech3371: 1260 is from 1260 = 15 seconds x 12 energy steps x 7 CEMs. I had comment about it before but removed from recent changes. I will add it back. Thank you!

tech3371:
I have addressed all of feedback and created issue for time conversion question. Will revisit in future PR. Going to squash commits and merge this PR.

PR #155, opened 2023-09-13
Xtce generator updates

tech3371:
# Change Summary

## Overview
Refactored XTCE generator code to work with different dataType and remove needing to specify specific mnemonic.

## New Files
- Create `xtce_generator_template.py` that we can use to generate any number of XTCE for any instrument.

## Deleted Files
- Deleted files from `imap_processing/codice/L0` folder which are being replaced by `xtce_generator_template.py`

## Updated Files
- moved xml files from `imap_processing/packet_definitions` into their instrument folder
- tools/xtce_generation/ccsds_header_xtce_generator.py. Updated `unint` to `UINT`
- tools/xtce_generation/telemetry_generator.py
   - main refactor in this PR happened in this file


tools/xtce_generation/telemetry_generator.py
```diff
@@ -188,131 +181,112 @@ def create_container_set(
...
         Returns:
         - parameter_set: The updated ParameterSet element.
         """
-        # Process rows from 10 until the last available row in the DataFrame
-        for index, row in pkt.iterrows():
-            if index < 9:
+        # Process rows from SHCOARSE until the last available row in the DataFrame
+        for index, row in self.pkt.iterrows():
+            if index < 8:
                 continue  # Skip rows before row 10
```
maxinelasp: Comment needs updating

tools/xtce_generation/xtce_generator_template.py
```diff
@@ -0,0 +1,50 @@
+from pathlib import Path
+
+from tools.xtce_generation.telemetry_generator import TelemetryGenerator
+
+
+def main():
```
maxinelasp, line 6: Nice! This example is great!

tools/xtce_generation/telemetry_generator.py
```diff
@@ -1,16 +1,20 @@
...
     def __init__(self, packet_name, path_to_excel_file, apid, sci_byte=None, pkt=None):
         self.packet_name = packet_name
-        self.path_to_excel_file = path_to_excel_file
         self.apid = apid
         self.sci_byte = sci_byte
-        self.pkt = pkt
+
+        # Read excel sheet
+        xls = pd.ExcelFile(path_to_excel_file)
+        self.pkt = xls.parse(packet_name)
```
maxinelasp: If someone wants to edit the packet definitions before calling TelemetryGenerator, then moving this parse step inside the class makes it harder to do that. For example, the GLOWS sheet contains a few extra columns which are not part of the packet - I removed them in Pandas before calling TelemetryGenerator. 

I think this is fine, because you can always edit the excel sheet, but I do want to call out that loss of flexibility by doing the parsing inside the class. Depending on how much you need to change the excel document, it could add some extra effort for processing new versions of the sheet.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -188,131 +181,112 @@ def create_container_set(
...
 
-        return science_container
+        return telemetry_metadata
 
-    def create_parameters_from_dataframe(self, parameter_set, pkt):
+    def create_remaining_parameters(self, parameter_set):
         """
-        Create XML elements for parameters based on DataFrame rows starting from row 10.
+        Create XML elements for parameters based on DataFrame rows starting
+        from SHCOARSE.
```
maxinelasp: Is SHCOARSE from the CCSDS header? I don't see that name in GLOWS excel document, but there is a "CCSDS packet header coarse time" which might be what that corresponds to?

maxinelasp (commented):
This looks great! I will be testing locally with the GLOWS packet to see how the output looks and how it works. Thanks for the updates! I'll approve once I test it out locally, but just had a few comments first :)

tools/xtce_generation/telemetry_generator.py
tech3371: yes!

tools/xtce_generation/telemetry_generator.py
tech3371: `SHCOARSE` is also know as `MET`.

tools/xtce_generation/telemetry_generator.py
tech3371: Can you give an example? 

Removing column should not effect it. We mainly use these column in generator - `lengthInBits`, `dataType`, `mnemonic`, `longDescription`, `shortDescription`. If those column contains, it should still work.

tools/xtce_generation/telemetry_generator.py
maxinelasp: Ah yes, that's what it's called. Can you update the comment to make that clear?

tools/xtce_generation/telemetry_generator.py
maxinelasp: The GLOWS sheet has a column which sums up all the bits - presumably as a check for the size of the packet. If I wanted to remove that programmatically, I wouldn't be able to, because the excel file is parsed inside this class instead of the class having the packet passed in as a dataframe. 

```
xls = pd.ExcelFile(excel_file)
    packet_df = xls.parse(packet_name)

    # Clean up packet
    packet_df.drop(
        labels=packet_df.columns[packet_df.columns.str.contains("Unnamed")],
        axis=1,
        inplace=True,
    )
    packet_df.dropna(subset="packetName", inplace=True)

    generator = GlowsTelemetryGenerator(packet_name, excel_file, apid, pkt=packet_df)
```

maxinelasp: it's hard to explain, and barely matters 😅 but this bit of code just chops off the unnamed columns and cleans up the sheet.

imap_processing/codice/packet_definitions/P_COD_HI_OMNI_SPECIES_COUNTS.xml
```diff
@@ -0,0 +1,179 @@
...
+					<xtce:ParameterRefEntry parameterRef="Bias_Gain_Mode" />
+					<xtce:ParameterRefEntry parameterRef="Suspect" />
+					<xtce:ParameterRefEntry parameterRef="Compression" />
+					<xtce:ParameterRefEntry parameterRef="Byte_Count" />
+					<xtce:ParameterRefEntry parameterRef="Data" />
+				</xtce:EntryList>
+			</xtce:SequenceContainer>
+		</xtce:ContainerSet>
+	</xtce:TelemetryMetaData>
+</xtce:SpaceSystem>
```
GFMoraga, line 179: This looks great!

tools/xtce_generation/xtce_generator_template.py
```diff
@@ -0,0 +1,50 @@
...
+    for packet_name, app_id in packets.items():
+        print(packet_name)
+        telemetry_generator = TelemetryGenerator(
+            packet_name=packet_name, path_to_excel_file=path_to_excel_file, apid=app_id
+        )
+        telemetry_generator.generate_telemetry_xml(
+            f"{packet_definition_path}/{packet_name}.xml", packet_name
+        )
+
+
```
GFMoraga, line 48: This looks great, but I will test it here in a bit!

GFMoraga:
I am assuming that we will build documentation for how to use this?

tools/xtce_generation/telemetry_generator.py
tech3371: @maxinelasp If I change it to this, will it help?
```suggestion
         if pkt is None:
            # Read excel sheet
            xls = pd.ExcelFile(path_to_excel_file)
            self.pkt = xls.parse(packet_name)
        else:
            self.pkt = pkt
```

tools/xtce_generation/telemetry_generator.py
tech3371: for sure! I updated it in last commit.

tools/xtce_generation/xtce_generator_template.py
tech3371: In here, you can add other definitions you need to convert by adding packet name and appId to the `packet` dict.

tools/xtce_generation/telemetry_generator.py
maxinelasp: Yes!

tech3371: cool. I will make that change.

maxinelasp (approved):
Works great! Thanks for this. Can you squash your commits before merging?

tech3371:
pre-commit.ci autofix

GFMoraga (approved):
Thanks for updating this and having an impromptu wg! It looks great!

PR #158, opened 2023-09-13
Lo  pkt definitions for housekeeping, sectored, angular,and priority counts

GFMoraga:
# Change Summary
I added the rest of codice lo packets in xtce format. Also deleted some files that where not being used. changed directory to not be uppercase

## Overview
Added codice packet definitions. 

## New Files

- `xtce_generator_codice.py`
   - the xtce generator for ONLY codice packets

- 7 packet definitions in xtce format

## Deleted Files

- deleted `decom_python_example.py`
   - not required anymore due to new decom files
- deleted `example_xtce.xml`
   - not needed because we have `xtce_generator_template.py`


tools/xtce_generation/xtce_generator_codice.py
```diff
@@ -0,0 +1,48 @@
...
+
+from tools.xtce_generation.telemetry_generator import TelemetryGenerator
+
+
+def main():
+    """This function can be used by any instrument to generate XTCE
+    for certain number of packets. Change values where TODO is
+    """
+
+    # TODO: change instrument name
```
tech3371: since this won't change, we can delete this TODO

tools/xtce_generation/xtce_generator_codice.py
```diff
@@ -0,0 +1,48 @@
...
+    instrument_name = "codice"
+    current_directory = Path(__file__).parent
+    module_path = f"{current_directory}/../../imap_processing"
+    packet_definition_path = f"{module_path}/{instrument_name}/packet_definitions"
+    # TODO: Copy packet definition to tools/xtce_generation/ folder
+    path_to_excel_file = (
```
tech3371: I would avoid committing local path.
I would do something like this
```
path_to_excel_file = f"{current_directory}/TLM_COD_20230629-110638(update).xlsx"
```
all someone else would need to do is copy excel file to `tools/xtce_generation/` and update excel filename as needed if it's different.

tech3371 (approved):
besides that one comment about local path, it looks good to me.

PR #159, opened 2023-09-14
Xtce IntegerParameterType fixes

tech3371:
# Change Summary
This fixes issue of using wrong parameter data types.
## Overview
Before, when we have `uint`, we were using `ParameterType` when it should be `IntegerParameterType`. Fixed that in this PR. Also made minor updates to template so that we can duplicate and modify easily.

## Updated Files
- tools/xtce_generation/telemetry_generator.py
   - It has fixes for data type
- tools/xtce_generation/xtce_generator_template.py
   - modify to make it easy to duplicate and modify

### DIFF
--- a/tools/xtce_generation/telemetry_generator.py
+++ b/tools/xtce_generation/telemetry_generator.py
@@ -106,7 +106,9 @@ class TelemetryGenerator:
         """
         for parameter_type_ref_name, size in unique_lengths.items():
             if "UINT" in parameter_type_ref_name:
-                parameter_type = Et.SubElement(parameter_type_set, "xtce:IntegerParameterType")
+                parameter_type = Et.SubElement(
+                    parameter_type_set, "xtce:IntegerParameterType"
+                )
                 parameter_type.attrib["name"] = parameter_type_ref_name
                 parameter_type.attrib["signed"] = "false"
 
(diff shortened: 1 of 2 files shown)


maxinelasp (approved):

tools/xtce_generation/telemetry_generator.py
```diff
@@ -298,4 +298,4 @@ def generate_telemetry_xml(self, output_xml_path, container_name):
         Et.indent(tree, space="\t", level=0)
 
         # Use the provided output_xml_path
-        tree.write(output_xml_path, encoding="utf-8", xml_declaration=True)
+        tree.write(output_xml_path, encoding="UTF-8", xml_declaration=True)
```
greglucas: Did this need to be changed? The standard library examples use lowercase.
https://docs.python.org/3/howto/unicode.html#reading-and-writing-unicode-data

tools/xtce_generation/telemetry_generator.py
```diff
@@ -246,9 +246,9 @@ def create_remaining_parameters(self, parameter_set):
 
             description = Et.SubElement(parameter, "xtce:LongDescription")
 
-            if row["longDescription"] is None and row["shortDescription"] is None:
+            if row.get("longDescription") is None and row.get("shortDescription") is None:
```
greglucas: This branching reads a little odd to me with all the checks. Perhaps something like this, or is it too confusing at first glance?

```python
description.text = row.get("shortDescription") or row.get("longDescription") or ""
```
Which I think will check them in the order you want.

greglucas (approved):
Two minor comments that are just there if you want but not necessary.

Edit: @tech3371 we should also fix the pre-commit black formatting check.

tools/xtce_generation/telemetry_generator.py
tech3371: oh ok. I will change it back to lowecase. I changed it thinking it may be causing issue but in reality it was the IntegerParameterType that was causing issue.

tools/xtce_generation/telemetry_generator.py
tech3371: Yeah, this is much more readable and cleaner. will add that. thanks!

PR #162, opened 2023-09-18
Extracting and adding new data to the IDEX L1

### LINKED ISSUE #194: Create Level 1 processing code for IDEX

bryan-harter:
# Change Summary

## Overview
* Added 30 or so new fields for information that's needed to parse the trigger
* Removed the logging function, all info that was logged is now going into the CDF
* Converting the FPGA time to unix time 

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files

- imap_processing/idex/idex_packet_parser.py
   - Adding 30 new fields from the header packet into the CDF
   - Removed the logging statement, instead just logging the trigger information
   - Converting the FPGA epoch time to a unix timestamp 
   -  updated the variable name from "data" to "l1_data" in the PacketParser class
- imap_processing/idex/tests/test_decom.py
  - Updated the unit test to reflect the new changes 
- pyproject.toml
  - Ignoring a new ruff rule for functions that are too long.  I'm not sure how to break up the extraction of the 30 new variables further, it seems to me they all belong in the same function.   

## Testing
Updating the unit test to reflect the new changes, no new tests needed though


bryan-harter:
Oh yeah one controversial thing I did was remove the ruff warning for functions that are too long.  I agree functions should be kept short, but extracting like 30 variables from pretty much right next to each other in the header seems like they all belong together.  I could just break it up into smaller functions, but I'm building that big dictionary that I can just loop through later, I kinda feel like a singular function would be better to keep track of the contents of that dictionary better.

bryan-harter:
Ehhh you know what, instead of doing all this stuff with bit shifting and masking, I'm going to leave that to space_packet_parser and modify the xml template.  I'll push some changes later today to modify the XML, and hopefully make the python code shorter

imap_processing/idex/idex_packet_parser.py
```diff
@@ -90,16 +92,36 @@ def __init__(self, packet_file: str):
...
-    HIGH_SAMPLE_RATE = 1 / 260  # nanoseconds per sample
-    LOW_SAMPLE_RATE = 1 / 4.0625  # nanoseconds per sample
+    HIGH_SAMPLE_RATE = 1 / 260  # microseconds per sample
+    LOW_SAMPLE_RATE = 1 / 4.0625  # microseconds per sample
+
+    NUMBER_SAMPLES_PER_LOW_SAMPLE_BLOCK = (
+        8  # The number of samples in a "block" of low sample data
+    )
+    NUMBER_SAMPLES_PER_HIGH_SAMPLE_BLOCK = (
+        512  # The number of samples in a "block" of high sample data
```
tech3371, line 121: is "block" like 8-bit chunks?

imap_processing/idex/idex_packet_parser.py
```diff
@@ -133,35 +161,335 @@ def __init__(self, header_packet):
...
+        trigger_notes_dict[
+            "tof_high_trigger_num_max_1_2"
+        ] = """Maximum number of samples between pulse 1 and 2 for TOF High double
+               pulse triggering"""
+        trigger_dict["tof_high_trigger_num_min_1_2"] = (
+            tofh_trigger_info_1 >> 21 & self.ELEVEN_BIT_MASK
+        )
+        trigger_notes_dict[
+            "tof_high_trigger_num_min_1_2"
+        ] = """Minimum number of samples between pulse 1 and 2 for TOF High double
```
tech3371, line 209: I see now. `.shortDescription` would have nice in this file. It's very informative.

imap_processing/idex/idex_packet_parser.py
```diff
@@ -133,35 +161,335 @@ def __init__(self, header_packet):
...
+        trigger_dict["detector_current"] = (
+            rejection_voltage_detector_current >> 10 & self.TWELVE_BIT_MASK
+        )
+        trigger_notes_dict[
+            "detector_current"
+        ] = "Last measurement in raw dN for processor board signal: Detector Current"
+
+        return trigger_dict, trigger_notes_dict
 
     def _calc_impact_time(self, packet):
```
tech3371, line 421: could I use this function in my PR to convert epoch time to unix time? or can we create a utils folder under `imap_processing` and move this function there since this seems like a common function that we all could benefit from?

Every packet that we get has time in epoch. It's called SHCOARSE or MET.

imap_processing/idex/idex_packet_parser.py
bryan-harter: Yeah it would certainly shorten the length of this function!

imap_processing/idex/idex_packet_parser.py
bryan-harter: There are 8 samples in a block, where a sample is just one unit of data collection.  Some sensors in the instrument take samples every 1/4.0625 microseconds, so a block of this low sample data is about 2 microseconds.  The only time that a block of data matters is in a function below, where we calculate how many samples of data are included before the before the dust impact triggered the instrument. 

For whatever reason, instead of the number of samples before triggering, the header has information about the number of blocks before triggering.

imap_processing/idex/idex_packet_parser.py
bryan-harter: Yes that sounds good! I guess it all must be in seconds before 2012?

tech3371: I will do that in future PR because I don't want extend both PRs. It should be small change.

tech3371 (approved):

imap_processing/idex/idex_packet_parser.py
```diff
@@ -27,20 +28,20 @@ class PacketParser:
 
     Attributes
     ----------
-        data (xarray.Dataset): An object containing all of the relevant data in the file
+        l1_data (xarray.Dataset): An object containing all of the relevant L1 data
```
greglucas: Do you plan on having other data fields around on this object? If not, I liked your generic `data` from before better so it doesn't say `l1_data.l1_data` repeated.

imap_processing/idex/idex_packet_parser.py
greglucas: Maybe add some of that great description as a docstring here?

imap_processing/idex/idex_packet_parser.py
```diff
@@ -90,13 +92,27 @@ def __init__(self, packet_file: str):
...
+        8  # The number of samples in a "block" of low sample data
+    )
+    NUMBER_SAMPLES_PER_HIGH_SAMPLE_BLOCK = (
+        512  # The number of samples in a "block" of high sample data
+    )
+
+    # Bit masks, spelled out for readability
     FOUR_BIT_MASK = 0b1111
     SIX_BIT_MASK = 0b111111
     TEN_BIT_MASK = 0b1111111111
```
greglucas: Remove all of these now?

Although it looks like you still have one TEN_BIT_MASK left that you might be able to add to the XTCE as well?

imap_processing/idex/idex_packet_parser.py
```diff
@@ -125,6 +142,11 @@ def __init__(self, header_packet):
             self.high_sample_trigger_time,
         ) = self._calc_sample_trigger_times(header_packet)
 
+        self.trigger_values_dict, self.trigger_notes_dict = self._get_trigger_dicts(
+            header_packet
+        )
+        logging.debug(f"{self.trigger_values_dict}")  # Log values here in case of error
```
greglucas:
```suggestion
        logging.debug(f"trigger_values_dict:\n{self.trigger_values_dict}")  # Log values here in case of error
```
Something to indicate what this is logging otherwise it might be tough to tie the dictionary to the area of interest.

imap_processing/idex/idex_packet_parser.py
```diff
@@ -133,35 +155,318 @@ def __init__(self, header_packet):
...
+        trigger_notes_dict[
+            "detector_current"
+        ] = "Last measurement in raw dN for processor board signal: Detector Current"
+
+        return trigger_dict, trigger_notes_dict
 
     def _calc_impact_time(self, packet):
         """
-        This calculates the number of seconds since Jan 1 2012
+        This calculates the unix timestamp from the FPGA header information
```
greglucas: Nothing you can do here, but I do find this odd on their part... I would have thought they wanted to keep track of microseconds since MET or some spacecraft pulse. Otherwise, it seems like this value could drift over time since there is no way of synchronizing it with a "true" time on the ground (unless they are doing this every contact, but that still doesn't seem ideal because it could have some significant drift over 3-day periods if you want accurate timings).

imap_processing/idex/idex_packet_parser.py
```diff
@@ -133,35 +155,318 @@ def __init__(self, header_packet):
...
+        We are given the number of seconds since Jan 1 2012, we need seconds since 1970.
+        Parameters
+        ----------
+            packet: The IDEX FPGA header packet
```
greglucas:
```suggestion
            packet : `space_packet_parser.ParsedPacket`
                The IDEX FPGA header packet
```
or something like that to reference whatever you're passing in as the type.

imap_processing/idex/idex_packet_parser.py
```diff
@@ -133,35 +155,318 @@ def __init__(self, header_packet):
...
+        Returns
+        -------
+            float
+                The unix timestamp
         """
-        # Number of seconds here
-        seconds_since_epoch = packet.data["SHCOARSE"].derived_value
+        # Number of seconds just Jan 1 2012 here
```
greglucas:
```suggestion
        # Number of seconds since Jan 1 2012
```

imap_processing/idex/idex_packet_parser.py
greglucas: Ahh, I think I see now. You are actually referencing SHCOARSE below, which is the MET time. So that is what I was thinking for synchronizing with pulses on the S/C.

I would update the below to be: `seconds_since_launch` or `mission_elapsed_time` so it is clear. I _think_ that they have arbitrarily chosen Jan 1 2012 as their launch epoch, but that won't actually be the case for us. It is just their testing frame for now.

imap_processing/idex/idex_packet_parser.py
```diff
@@ -133,35 +155,318 @@ def __init__(self, header_packet):
...
+        return (
+            unix_time_2012 + seconds_since_2012
+        )  # Return seconds between 1970 and 2012, and 2012 to present
 
     def _calc_sample_trigger_times(self, packet):
         """
         Calculate the high sample and low sample trigger times from a header packet
+        Parameters
+        ----------
+            packet : The IDEX FPGA header packet info
```
greglucas:
```suggestion
            packet : ...
                The IDEX FPGA header packet info
```

imap_processing/idex/idex_packet_parser.py
```diff
@@ -133,35 +155,318 @@ def __init__(self, header_packet):
...
+
+        Returns
+        --------
+            int
+                The actual trigger time for the low sample rate in microseconds
+            int
+                The actual trigger time for the high sample rate in microseconds
```
greglucas:
```suggestion
            (int, int)
                The actual trigger times for the low and high sample rate in microseconds
```

pyproject.toml
```diff
@@ -66,4 +67,4 @@ target-version = "py39"
 select = ["B", "E", "F", "I", "N", "S", "W", "PL", "PT", "UP", "RUF"]
 # Ignore import sorting for now until lines_after_imports is respected
 # by ruff and we can replace isort
-ignore = ["D203", "D212", "PLR0913", "PLR2004", "S101"]
+ignore = ["D203", "D212", "PLR0913", "PLR2004", "S101", "PLR0915"]
```
greglucas: Do you still need this after your refactor to XTCE?

Perhaps we can add it to `per-file-ignores` for just this one file instead if you do need it?
https://docs.astral.sh/ruff/settings/#per-file-ignores

greglucas (commented):
Could you add idex to the list of instruments being documented too?

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/b1a7b65be84de48c1a748155250ad2c4f3642910/docs/source/reference/index.rst?plain=1#L12

imap_processing/idex/idex_packet_parser.py
bryan-harter: Ahh ok, yeah that wasn't clear to me either.  I thought it was weird they chose seconds since 2012.

bryan-harter:
Alright I believe I addressed all comments! I'm going to merge just so I can get the CDF review up, but if there were any outstanding issues that I missed I can make tweaks in the next IDEX review

PR #164, opened 2023-09-21
initial hit l1a decom

### LINKED ISSUE #166: Create initial structure for HIT L1A processing algorithms
Create the initial structure for unpacking the HIT L0 data for all APIDs into L1A data

### LINKED ISSUE #170: Generate HIT XTCE definitions

sdhoyt:
# Change Summary

## Overview
This is an initial pass at the HIT L1A decommutation. Not everything is here and testing was very limited because the data I have are packets from the HIT SIM that only contain SHCOARSE. This initial code does the following:

1. unpacks the data
2. If there are multiple instances of an APID in a packet, it groups them all together and sorts them by time
3. Creates an Xarray dataset (I haven't fully tested this because I don't have anything besides SHCOARSE in the data) with an Epoch dimension
4. Returns a dictionary with a Dataset for each APID

I'm still not sure the best way to organize all of this, but I wanted to get something out there for people to start giving feedback on. One thing I particularly struggled with is variable naming because we get a CCSDS packet, but that packet contains multiple APIDs that we also call packets(?). Please let me know if anyone has any better naming ideas.

This updates also adds all of the HIT XTCE files.

## New Dependencies
None

## New Files
- HIT XTCE Files
   - HIT_AUT = 1250  # Autonomy
   - HIT_HSKP = 1251  # Housekeeping
   - HIT_SCIENCE = 1252  # Science
   - HIT_IALRT = 1253  # I-ALiRT
   - HIT_MEMDUMP = 1255  # Memory dump

- hit_apid.py
   - Enums for the HIT APIDs
- hit_l1a_decom.py
   - decommutation code for creating L1A datasets
 
## Deleted Files
None

## Updated Files
None

## Testing
- test_hit_decom.py
   - unit tests for hit_l1a_decom. This is currently very limited until I get more data.


imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,131 @@
...
+
+from imap_processing import decom
+from imap_processing.hit.l0.hit_apid import HitAPID
+
+logging.basicConfig(level=logging.INFO)
+
+
+def decom_hit_packets(packet_file: str, xtce: str):
+    """
+    Unpack and decode IMAP-Lo packets using CCSDS format and XTCE packet definitions.
```
sdhoyt: Need to replace IMAP-Lo with HIT

imap_processing/hit/l0/hit_apid.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class HitAPID(Enum):
```
greglucas:
```suggestion
class HitAPID(IntEnum):
```
Probably not a huge deal, but if you do want to verify against numbers later it might help to be explicit that these are ints?

```python
assert HitAPID.HIT_AUT == 1250
```

imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,131 @@
+import logging
+
+import xarray as xr
+
+from imap_processing import decom
+from imap_processing.hit.l0.hit_apid import HitAPID
```
greglucas: Do you want to just include the APID Enum directly in this file since this will presumably be the only place they are used?

imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    # TODO: XTCE Files need to be combined
+    logging.info(f"Unpacking {packet_file} using xtce definitions in {xtce}")
+    packets = decom.decom_packets(packet_file, xtce)
+    logging.info(f"{packet_file} unpacked")
+
+    unpacked_data = {}
+    for apid in [id.value for id in HitAPID]:
+        # TODO: if science packet, do decompression
+        apid_name = HitAPID(apid).name
```
greglucas:
```suggestion
    for apid_name, apid in [(id.name, id.value) for id in HitAPID]:
        # TODO: if science packet, do decompression
```

imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,131 @@
...
+    packets: list
+        List of all the unpacked data from decom.decom_packets()
+    apid: int
+        APID number for the data you want to group together
+    Returns
+    -------
+    list
+        list containing a dictionary with data for each instance the specified APID.
+        For example, if a CCSDS packet has N instances of APID X, this list
+        will contain N dictionaries all containing data for APID X.
```
greglucas: It looks like you are creating a list of dicts here, then later going to  dict of lists. Would you want to start with the dict of lists here?

For SWE we are using defaultdict to create an empty list that can be appended to straight away: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/b1a7b65be84de48c1a748155250ad2c4f3642910/imap_processing/swe/l1a/swe_science.py#L112

I'll add the inline version of what I think it could be below as a suggestion.

imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,131 @@
...
+        if packet.header["PKT_APID"].derived_value == apid:
+            data_dict = {}
+            # get a list of all the APID fields
+            packet_fields = list(packet.data.keys())
+            for field in packet_fields:
+                # put the value of the field in a dictionary
+                data_dict[field] = packet.data[field].derived_value
+            # put the data dictionary in a list
+            all_data.append(data_dict)
+    return all_data
```
greglucas:
```suggestion
    data_dict = collections.defaultdict(list)
    for packet in packets:
        if packet.header["PKT_APID"].derived_value == apid:
            # get a list of all the APID fields
            for field in packet.data:
                # append the value of the field from the packet in the dictionary
                data_dict[field].append(packet.data[field].derived_value)
    return data_dict
```

imap_processing/hit/tests/test_hit_decom.py
```diff
@@ -0,0 +1,43 @@
...
+    assert len(decom_test_data) == total_datasets
+
+
+def test_dataset_dims_length(decom_test_data, raw_packet_data):
+    """Test if the time dimension length in the dataset is correct"""
+    num_packet_times = 0
+    for packet in raw_packet_data:
+        if packet.header["PKT_APID"].derived_value == HitAPID.HIT_SCIENCE.value:
+            # Each APID will contain one time
+            num_packet_times += 1
```
greglucas: If you go to an `IntEnum` I think you can do the comparison directly then.
```suggestion
    # Verify the number of science packets in the data
    sum(1 for packet in raw_packet_data if packet.header["PKT_APID"].derived_value == HitAPID.HIT_SCIENCE)
```

Also, if you know that explicit number I would add that too? So we aren't testing 0 == 0 here on accident.

greglucas:
Oh, one more thing, could you add this to the documentation as well by doing an autodoc similar to SWE under the instruments and adding it to the Instruments TOC tree.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/docs/source/reference/swe.rst

imap_processing/hit/l0/hit_l1a_decom.py
sdhoyt: I did the list of dictionaries first because that was a way to sort all the data by time and keep it all grouped together. So if I have packets with arrays of events, I can sort the dictionaries by time and those arrays will also get sorted by time, then when each field gets put into its own list it's already in the correct time order. I couldn't think of a simple way to do that by putting all of the fields into their own list to start.

imap_processing/hit/l0/hit_apid.py
sdhoyt: Ah cool, I didn't know about IntEnum

imap_processing/hit/l0/hit_l1a_decom.py
sdhoyt: Yeah, I'll move it into the file

imap_processing/hit/l0/hit_l1a_decom.py
sdhoyt: Agreed, I'll update to this

imap_processing/hit/l0/hit_l1a_decom.py
greglucas: Ahh, I was worried about sorting by time as well earlier when reviewing IDEX code, but it has been confirmed that our packets will come to us in sorted order. So we are guaranteed a sorted-by-time packet list. But, I agree that doesn't seem great to rely on... I guess another way would be to sort the initial list right away by time. Something like this: `packets = sorted(packets, key=lambda packet: packet.SHCOARSE)`

sdhoyt: Oh duh, yeah, it'll still be the right order if I sort when the APIDs are still mixed together. Thanks! And yeah, the Lo documentation implied that I can't trust that they will be in the correct time order, so I figured that was the case with all instruments. That's good to hear they will be, but I'll still sort just in case.

greglucas: > the Lo documentation implied that I can't trust that they will be in the correct time order

They may not be in the correct order when first received at the MOC, but our POC folks do sort them when we query their databases. So it is sort of an internal reason we get them sorted.

Or, does this mean that Lo's packet1 may contain earlier information than Lo's packet0 even if the MET is ordered...

sdhoyt: It's just saying the MET needs to be ordered. The documentation says, "The processing software will identify all available packets for a given type which fall within the pointing period and will time sort them and drop the individual fields into arrays"

imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    # TODO: XTCE Files need to be combined
+    logging.info(f"Unpacking {packet_file} using xtce definitions in {xtce}")
+    packets = decom.decom_packets(packet_file, xtce)
+    logging.info(f"{packet_file} unpacked")
+
+    # sort all the packets in the list by their spacecraft time
+    sorted_packets = sorted(packets, key=lambda x: x.data["SHCOARSE"].derived_value)
+
+    unpacked_data = defaultdict(list)
```
greglucas:
```suggestion
    # Store data for each apid
    # unpacked_data = {apid0: {var0: [item0, item1, ...], var1: [item0, item1, ...]}, ...}
    unpacked_data = {}
```
I think this one should just be a dictionary? So unpacked data will be a dict of dicts now I believe.

imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,110 @@
...
+        A dictionary containing xr.Dataset for each APID. each dataset in the
+        dictionary will be converted to a CDF.
+    """
+    dataset_dict = defaultdict(list)
+    # create one dataset for each APID in the data
+    for apid, data_dict in data.items():
+        # if data for the APID exists, create the dataset
+        if data_dict != {}:
+            epoch = xr.DataArray(
+                name="Epoch", data=data_dict["SHCOARSE"], dims=("Epoch")
```
greglucas: We should get @bryan-harter 's comment on how to name this.

I think this should be `MET` here since it is the mission elapsed time and then we won't confuse it with the reserved epoch name. Then later, I think we need to convert to an EPOCH which I believe needs to be all capitalized, but I also think it has a specific meaning and needs to be offset from a specific date for CDFs.

I don't know the details here though, so hopefully Bryan can give us some guidance.

imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+            dataset = xr.Dataset(data_vars={}, coords={"Epoch": epoch})
+            data_dict.pop("SHCOARSE")
```
greglucas: If you need to remove it, just pop() above when you are making the previous DataArray?

imap_processing/hit/l0/hit_l1a_decom.py
```diff
@@ -0,0 +1,116 @@
...
+            dataset_dict[apid] = dataset.assign(**data_dict)
+
+    return dataset_dict
+
+
+def group_apid_data(packets, apid):
+    """
+    Creates a list of dictionaries containing data for each instance
+    of a specified APID
```
greglucas:
```suggestion
    Creates a dictionary of lists containing data for each APID
```

docs/source/reference/hit.rst
```diff
@@ -0,0 +1,16 @@
+.. _hit:
+
+HIT (High-energy Ion Telescope)
+==========================
+
+.. currentmodule:: imap_processing.hit.l0
+
+This is the SWE Instrument module, which contains the code for processing
+data from the SWE instrument.
```
greglucas:
```suggestion
This is the HIT Instrument module, which contains the code for processing
data from the HIT instrument.
```

docs/source/reference/hit.rst
```diff
@@ -0,0 +1,16 @@
+.. _hit:
+
+HIT (High-energy Ion Telescope)
+==========================
+
+.. currentmodule:: imap_processing.hit.l0
```
greglucas:
```suggestion
.. currentmodule:: imap_processing.hit
```
Do we want to make this be the `hit` base module, and then automodule/autosummary it so that it picks up your future levels as well?

Or I think you could put in l0.hit_l1a_decom which might be more explicit here that you're in the l0 code path as well.

greglucas (approved):
Nice, I think this is a great start! A few minor corrections after the last update since I had you move things around I don't think the documentation got updated with it.

docs/source/reference/hit.rst
sdhoyt: ` l0.hit_l1a_decom` worked

tech3371 (approved):

docs/source/reference/index.rst
```diff
@@ -16,6 +16,7 @@ Instruments
    :maxdepth: 1
 
    swe
+   hit
```
bourque: Real nit-picky here but maybe we should set the precedent here for keeping this list in alphabetical order?

bourque (approved):
This looks great!

docs/source/reference/index.rst
sdhoyt: That's a good point. I'll update that

imap_processing/hit/l0/hit_l1a_decom.py
laspsandoval: Looks good. Nice job!

laspsandoval (approved):
Nice work.

imap_processing/hit/l0/hit_l1a_decom.py
sdhoyt: Merging with this comment unresolved. This will be fixed later by issue #269 .

PR #174, opened 2023-09-27
Ruff docs

greglucas:
# Change Summary

NOTE: This is a proposal. Do we want to follow all of these conventions? Are some of these too annoying and will make too much noise and therefore we should ignore them? I went through and made the changes I thought made sense for now and they didn't seem _too_ bad, and it is better to start these checks sooner rather than later IMO.

## Overview

Add numpydoc conventions to our continuous integration setup to
start enforcing documentation standards.

Have chosen to ignore module-level and package-level docstrings
for now. We can add those rules later if we want. We also ignore
these for the tools and tests directory, which can be added later
as well if we think they are needed.

## Updated Files
Many, I updated the docstrings to meet the ruff enforced standards.

## Testing
Ignored docstrings in our tests directory.

### DIFF
--- a/imap_processing/codice/decom.py
+++ b/imap_processing/codice/decom.py
@@ -1,4 +1,6 @@
-""" This is an example of how to use the 'space_packet_parser' module to parse with a
+"""Decommutate CODICE L0 data.
+
+This is an example of how to use the 'space_packet_parser' module to parse with a
 XTCE file specifically for the CODICE L0 data. This is a modified version of the
 example found in the 'space_packet_parser' module documentation.
 This is the start of CODICE L0 data processing.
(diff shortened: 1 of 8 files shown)


maxinelasp (approved):
I think the changes you had to make show that we are already using this basic style in our documentation. I think it's a good thing to add to the pre commit.

bourque (approved):
I like all of these changes, and I am for enforcing these moving forward.

I think I would also be good with module-level enforcement too.

sdhoyt (approved):
I agree. I think it would be good to add this to the checks.

greglucas:
Since we seem to have a consensus on this. I'm going to merge because there are going to be minor merge conflicts coming up with all the PRs and I'd like to push those on to everyone else so you all can update the docstrings in your own words and I don't have to infer what you want!

Issue #105, opened 2023-09-06
Create GLOWS L1A Algorithm


maxinelasp:
### Summary

Create an algorithm for GLOWS for generating level 1a CDF files from packets using XTCE formatting

A lot of this algorithm is already included in the GLOWS-provided [Python code](https://lasp.colorado.edu/galaxy/display/IMAP/GLOWS)

### Requirements

The more specific the better.
  - [x] Parse packets using space_packet_parser
  - [x] Create XArray following the standard format for CDFs (#77 )
  - [x] Output first draft CDF files for GLOWS L1A
  - [x] Create a basic unit test for checking if packet decommutation works 

### Related tickets

#104 , #97

PR #175, opened 2023-09-27
Glows Decom work

maxinelasp:
# Change Summary
Adding files for GLOWS L0 decom work, issue #104 and part of #105.

## Overview
I created a data class for both direct event and histogram data in glows. This code will decommute the test packet provided by the GLOWS team, and return two lists of the two main science packets needed for processing. 

This review includes both the separate XML files for both direct events and histograms, and a combined XML file for both of them. The final code uses only the combined file. These files were almost entirely generated from the xtce_generator. I am planning on going back and adding both the binary data type, which I manually changed for the histogram data field, and code for creating multi-APID xml files. 

## New Files

- glows/l0/glows_decom
   - Provides a method for decoding GLOWS packets into histogram and direct events data
-  glows/l0/glows_l0_data
   - Provides data classes for GLOWS level 0 data, including methods for modifying that data (I expect to need some additional methods for processing before outputting to CDF files, so although they're overkill now, this is future-planning)
-  packet definitions
   - 3 packet defintions files: one each for histogram and direct events data, and one which combines them
- test_glows_decom
   - tests for glows_decom.py
- xtce_generation/xtce_generator_glows
   - code for automatically generating the separate histogram and direct events XTCE files  

## Updated Files

- xtce_generator.py
   - Updating to fix a typo in the unsigned integer generation

## Testing

I added some general decom unit tests. This uses a smaller piece of the whole packet file, since the entire thing is very large. 

In addition, I was able to confirm the histogram data matches the expected value from the packets using a plot provided to me by the GLOWS team:

Expected result: 

RMlj7rjNti7BZpxm

My results: 
histogram_test_data_plot


imap_processing/glows/l0/glows_decom.py
```diff
@@ -0,0 +1,59 @@
...
+                # Do something with the packet data
+                if packet.header["PKT_APID"].derived_value == hist_apid:
+                    hist_l0 = GlowsHistL0(packet)
+                    histdata.append(hist_l0)
+
+                if packet.header["PKT_APID"].derived_value == de_apid:
+                    de_l0 = GlowsDeL0(packet)
+                    dedata.append(de_l0)
+
+        except ReadError as e:
```
maxinelasp: This is similar to the safe_packet_generator that Laura created in her Ultra L0 PR: #128. It does occur with the smaller packet that I created, because I ended up cutting it off in the middle of the last packet.

imap_processing/glows/packet_definitions/GLX_COMBINED.xml
```diff
@@ -0,0 +1,209 @@
...
+			<xtce:Parameter name="ELVAR" parameterTypeRef="UINT16">
+				<xtce:LongDescription>Variance of event-impulse length</xtce:LongDescription>
+			</xtce:Parameter>
+			<xtce:Parameter name="EVENTS" parameterTypeRef="UINT32">
+				<xtce:LongDescription>Number of events</xtce:LongDescription>
+			</xtce:Parameter>
+			<xtce:Parameter name="HISTOGRAM_DATA" parameterTypeRef="BYTE28800">
+				<xtce:LongDescription>Histogram Counts</xtce:LongDescription>
+			</xtce:Parameter>
+			<!--Direct events sequence containers-->
```
maxinelasp, line 141: There is a name collision in the histogram and direct events packets - they both have a parameter named `SEC`. I could update the name (`DE_SEC`) but opted to just use the same `SEC` definition (line 81) because they were the same size and represent the same thing. This may come up again with other combined XTCE files.

imap_processing/glows/l0/glows_decom.py
```diff
@@ -0,0 +1,58 @@
...
+        Path to data packet path with filename
+
+    Returns
+    -------
+    data : tuple[list[GlowsHistL0], list[GlowsDeL0]]
+        A tuple with two pieces: one list of the GLOWS histogram data, in GlowsHistL0
+        instances, and one list of the GLOWS direct event data, in GlowsDeL0 instance
+    """
+    hist_apid = 1480
+    de_apid = 1481
```
greglucas: HIT and Ultra are using Enum/IntEnum for these, do you want to do the same here?

imap_processing/glows/l0/glows_decom.py
greglucas: > It does occur with the smaller packet that I created, because I ended up cutting it off in the middle of the last packet

I think this was covering up a true error in Ultra, so I'd be hesitant to put this in here. Are you able to trim the packet to the exact binary length you want? Maybe also put the command you did to trim the file in a comment at the top of this file or somewhere so that we can remember how to recreate the data here if we get a new test data dump.

imap_processing/glows/l0/glows_decom.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Define paths
+    xtce_document = Path(
+        f"{imap_module_directory}/glows/packet_definitions/GLX_COMBINED.xml"
+    )
+
+    hist_packet_definition = xtcedef.XtcePacketDefinition(xtce_document)
```
greglucas: Just a general packet definition now since you've combined them? You label the `hist_`, but then also return `de_`, so the naming is just a bit confusing.

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -0,0 +1,153 @@
...
+    data: dict
+        The data from the packet
+
+    Methods
+    -------
+    process_packet(packet)
+        Process the given packet into the data dictionary
+
+    _convert_histogram_data(binary_hist_data)
+        Process the binary histogram data into 8-bit bins, in a list.
```
greglucas: These should get documented automatically because you have the docstrings associated with those methods below I believe.

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -0,0 +1,153 @@
...
+            Data dictionary created from the packet, using L0_KEYS
+        """
+        data = {}
+        for key in self.L0_KEYS:
+            if key != "HISTOGRAM_DATA":
+                data[key] = packet.data[key].derived_value
+            else:
+                data[key] = self._convert_histogram_data(packet.data[key].raw_value)
+            if data[key] is None:
+                raise ValueError(f"Missing data for {key} in {packet.data}")
```
greglucas: Is this meaning that the packet definition isn't filled out properly? The derived_value returning None seems a bit concerning. I figured it would have been converted to int/float according to your definition in the xml.

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -0,0 +1,78 @@
...
+    """Read test data from file"""
+    current_directory = Path(__file__).parent
+    packet_path = f"{current_directory}/glows_test_packet_20230920_v00.pkts"
+    data_packet_list = glows_decom.decom_packets(packet_path)
+    return data_packet_list
+
+
+def test_glows_decom_count(decom_test_data):
+    expected_hist_len = 61
+    expected_de_len = 311
```
greglucas: Is this the number of histogram packets in the packet file? A quick comment or a change of variable name to `expected_num_hist_packets` or something like that would help I think.

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -0,0 +1,78 @@
...
+    assert len(decom_test_data[0]) == expected_hist_len
+
+    # Direct events data
+    assert len(decom_test_data[1]) == expected_de_len
+
+
+def test_glows_hist_data(decom_test_data):
+    for hist in decom_test_data[0]:
+        for key in hist.data:
+            assert hist.data[key] is not None
```
greglucas: This would raise from your actual code if this is the case right?

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -0,0 +1,78 @@
...
+        assert len(hist.data["HISTOGRAM_DATA"]) == 3599
+        assert min(hist.data["HISTOGRAM_DATA"]) > 10
+        assert max(hist.data["HISTOGRAM_DATA"]) < 170
```
greglucas: You made that awesome-looking plot, do you have the data output from that chart which you could turn into an array to assert against directly rather than the min/max?

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -0,0 +1,78 @@
...
+        "HVAVG": 2007,
+        "HVVAR": 0,
+        "SPAVG": 46875,
+        "SPVAR": 0,
+        "ELAVG": 6,
+        "ELVAR": 0,
+        "EVENTS": 95978,
+    }
+    for key in expected_data.keys():
+        assert decom_test_data[0][0].data[key] == expected_data[key]
```
greglucas: Can you do a direct comparison? Or is this to avoid potential extra keys in the actual test data?

```suggestion
    assert decom_test_data[0][0].data == expected_data
```

imap_processing/glows/packet_definitions/GLX_COMBINED.xml
GFMoraga: I think you highlight a great observation here. Something to keep in mind for sure. The xtce looks great!

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -0,0 +1,153 @@
...
+        if len(histograms) != 3599:
+            raise ValueError(
+                f"Histogram packet is lacking bins. Expected a count of 3599, "
+                f"actually received {len(histograms)}"
+            )
+
+        return histograms
+
+
+class GlowsDeL0:
```
greglucas: I always struggle with naming initialisms. Do you want it to be `DE` to signify that they are two words?

I'm not sure you need Glows here because you are in the glows module already and that might save some typing, an alternative is something like this to spell it out more explicitly: `L0DirectEvent`?

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -0,0 +1,153 @@
+class GlowsHistL0:
```
greglucas: Do you want to turn these into true DataClasses and keep the items accessed as attributes on the class `class.key` rather than via the `class.data[key]` dict lookup?
https://docs.python.org/3/library/dataclasses.html

greglucas (commented):
Nice, this looks like a great setup. Can you also add glows to the documentation list?

imap_processing/glows/l0/glows_decom.py
maxinelasp: You're right, good catch! That was leftover from when this _was_ a hist-only packet parser.

imap_processing/glows/l0/glows_decom.py
maxinelasp: Yeah, I'll take a look at the packet and see if I can actually get a properly trimmed packet. I was using the code that the GLOWS team provided, which does byte by byte reading, but apparently failed to actually chop it at the right place. I'll try again, just wanted to get this PR up mostly

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: Yeah, that actually does make sense. Not sure why I didn't do that in the first place... but I will update it now!

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: So I should remove the whole `methods` section? I looked up some numpy pydocs guide and it said to include this.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: It's not supposed to happen, that's why it throws an error. I came across it once when I had a bad XTCE so I figured I should throw up an alert just in case rather than having some other random piece fail due to missing data.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: I went back and forth on the `DE` and ultimately left it like this because the three capitals in a row just looked weird. But that's not a strong preference, and it is two words... 

I do like `L0DirectEvent`, so maybe I will just change it to that. Thanks for the suggestion!

imap_processing/glows/tests/test_glows_decom.py
maxinelasp: Yes, it is. I can update it to be more clear

imap_processing/glows/tests/test_glows_decom.py
maxinelasp: good point, I'll take the assert out

imap_processing/glows/tests/test_glows_decom.py
maxinelasp: No, I don't. Ultimately, the plan I have discussed with the GLOWS team is to have some kind of statistical test for checking the output - so I didn't go too overkill with the testing now, because I plan to go back and add that once they provide that data.

imap_processing/glows/tests/test_glows_decom.py
maxinelasp: This skips the histogram data, which is a long array and already checked in another test.

imap_processing/glows/packet_definitions/GLX_COMBINED.xml
tech3371: It's ok to use same parameter if they do same thing and same size but referred in different packets. I have similar thing in SWE.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -105,16 +105,21 @@ def create_parameter_types(self, parameter_type_set, unique_lengths):
         - parameter_type_set: The updated ParameterTypeSet element.
         """
         for parameter_type_ref_name, size in unique_lengths.items():
-            if "UINT" in parameter_type_ref_name:
+            if "UINT" or "INT" in parameter_type_ref_name:
```
tech3371:
```suggestion
            if "UINT" in parameter_type_ref_name:
```

tools/xtce_generation/telemetry_generator.py
```diff
@@ -105,16 +105,21 @@ def create_parameter_types(self, parameter_type_set, unique_lengths):
...
                 parameter_type = Et.SubElement(
                     parameter_type_set, "xtce:IntegerParameterType"
                 )
                 parameter_type.attrib["name"] = parameter_type_ref_name
-                parameter_type.attrib["signed"] = "false"
+                parameter_type.attrib["signed"] = (
+                    "false" if "UINT" in parameter_type_ref_name else "true"
+                )
```
tech3371:
```suggestion
                parameter_type.attrib["signed"] = false
```

tools/xtce_generation/telemetry_generator.py
```diff
@@ -105,16 +105,21 @@ def create_parameter_types(self, parameter_type_set, unique_lengths):
...
 
                 encoding = Et.SubElement(parameter_type, "xtce:IntegerDataEncoding")
                 encoding.attrib["sizeInBits"] = str(size)
-                encoding.attrib["encoding"] = "unsigned"
+                encoding.attrib["encoding"] = (
+                    "unsigned" if "UINT" in parameter_type_ref_name else "signed"
+                )
+
```
tech3371:
```suggestion
                encoding.attrib["encoding"] = "unsigned"

```

tools/xtce_generation/telemetry_generator.py
```diff
@@ -105,16 +105,21 @@ def create_parameter_types(self, parameter_type_set, unique_lengths):
...
             elif "SINT" in parameter_type_ref_name:
```
tech3371:
```suggestion
            elif "SINT" or "INT" in parameter_type_ref_name:
```

imap_processing/glows/l0/glows_decom.py
```diff
@@ -0,0 +1,58 @@
...
+
+    histdata = []
+    dedata = []
+
+    with open(packet_file_path, "rb") as binary_data:
+        try:
+            hist_packets = histparser.generator(
+                binary_data,
+                buffer_read_size_bytes=5790778,
+            )
```
tech3371: We do lot of these in `imap_processing/decom.py`. We could update that function to take optional parameter `buffer_read_size_bytes` in case other needs to use it in the future.

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -0,0 +1,153 @@
...
+        "HISTOGRAM_DATA",
+    )
+
+    def __repr__(self):
+        return str(self.data)
+
+    def __init__(self, packet):
+        self.data = self.process_packet(packet)
+
+    def process_packet(self, packet) -> dict:
```
tech3371: What is the reason for this function? It seems to be doing duplicate work that may not be needed. What I mean is `packet` already has keys and value of HIST packet. You can reference it like this
Eg.
```
for key, value in packet.data.items():
        if key != "HISTOGRAM_DATA":
           # get key and value here
           print(key, value.derived_value)
```

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -0,0 +1,153 @@
...
+        for key in self.L0_KEYS:
+            if key != "HISTOGRAM_DATA":
+                data[key] = packet.data[key].derived_value
+            else:
+                data[key] = self._convert_histogram_data(packet.data[key].raw_value)
+            if data[key] is None:
+                raise ValueError(f"Missing data for {key} in {packet.data}")
+        return data
+
+    def _convert_histogram_data(self, binary_hist_data: str) -> list[int]:
```
tech3371, line 87: can we replace this function with this line? I did this in SWE. SWE also needed to get data in 8-bit chunks too.
```
binary_hist_data = packet.data["HISTOGRAM_DATA"].raw_value
byte_data = int(binary_hist_data, 2).to_bytes(3599, byteorder="big")
# convert bytes to numpy array of uint8
raw_counts = np.frombuffer(byte_data, dtype=np.uint8)
```

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -0,0 +1,153 @@
...
+        ----------
+        packet
+            Direct event packet from space_packet_parser
+        Returns
+        -------
+        data: dict
+            Data dictionary created from the packet, using L0_KEYS
+        """
+        data = {}
+        for key in self.L0_KEYS:
```
tech3371: same comment as above

imap_processing/glows/l0/glows_decom.py
maxinelasp: I didn't use that method because I wanted to use the generator directly and produce the data types as needed, rather than reading and converting the entire packet, converting that to a list, then looping through the list _again_ to process it for glows.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: This is a great point. My one concern is, the expected list of keys does act as an additional check on the data - if one of the expected keys is missing, then the program raises an error. Is that something we would ever expect to happen? Or is that already checked by space_packet_parser?

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: No, that doesn't work. I get an overflow error from trying to convert the binary_hist_data to an int.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: I went ahead and made the change, but I'm still curious about the answer to this question.

imap_processing/glows/l0/decom_glows.py
```diff
@@ -0,0 +1,77 @@
...
+from pathlib import Path
+
+from bitstring import ReadError
+from space_packet_parser import parser, xtcedef
+
+from imap_processing import imap_module_directory
+from imap_processing.glows.l0.glows_l0_data import DirectEventL0, HistogramL0
+
+
+class GlowsParams(Enum):
```
greglucas, line 11:
```suggestion
class GlowsParams(IntEnum):
```
so then below you can do `if x == GlowsParams.HIST_APID` instead of having to get into the `.value`

tools/xtce_generation/telemetry_generator.py
```diff
@@ -110,12 +110,13 @@ def create_parameter_types(self, parameter_type_set, unique_lengths):
                     parameter_type_set, "xtce:IntegerParameterType"
                 )
                 parameter_type.attrib["name"] = parameter_type_ref_name
-                parameter_type.attrib["signed"] = "false"
+                parameter_type.attrib["signed"] = false
```
greglucas:
```suggestion
                parameter_type.attrib["signed"] = False
```
Is false defined above to be a string? If so it is confusing. Or do you want this to be the boolean and need the caps?

greglucas (approved):
Approving modulo fixing the `false` error.

tools/xtce_generation/telemetry_generator.py
maxinelasp: It should be a string, I'll update that.

sdhoyt (approved):
Looks great to me!

imap_processing/glows/l0/glows_decom.py
tech3371: I see

imap_processing/glows/l0/glows_l0_data.py
tech3371: I see. I see your point. I think that's where validation data comes to help. The way you validated with the plot. If the packet definition didn't have a key, validation data wouldn't match. But if we track that in here too, then it would add more things to track. That's my thought :)

imap_processing/glows/l0/glows_l0_data.py
tech3371: bummer. ok

tech3371 (approved):

imap_processing/glows/packet_definitions/P_GLX_TMSCHIST.xml
```diff
@@ -0,0 +1,190 @@
...
+			<xtce:IntegerParameterType name="UINT32" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="32" encoding="unsigned" />
+			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="INT32" signed="true">
+				<xtce:IntegerDataEncoding sizeInBits="32" encoding="signed" />
+			</xtce:IntegerParameterType>
+			<xtce:BinaryParameterType name="BYTE28800">
+				<xtce:BinaryDataEncoding bitOrder="mostSignificantBitFirst">
+					<xtce:SizeInBits>
+						<xtce:FixedValue>28800</xtce:FixedValue>
```
laspsandoval, line 39: Wondering if this might change and not be a FixedValue.

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -0,0 +1,156 @@
...
+    STARTID : int
+        Histogram Start ID
+    ENDID : int
+        Histogram End ID
+    FLAGS : int
+        Histogram flags
+    SWVER : int
+        Version of SW used for generation
+    SEC : int
+        Block start time (IMAP), seconds<
```
laspsandoval: minor typo <

imap_processing/glows/packet_definitions/P_GLX_TMSCHIST.xml
maxinelasp: Good, point, as far as I am aware, it is a fixed value. I have gotten some test data and this successfully decoms it. It's a standard size for the number of histogram buckets, so I think it will stay the same, but I will double-check when I meet with them next.

laspsandoval:
Maybe it is my imagination, but the Expected Result and your result look ever so slightly different.  I do wonder if the Fixed value parameter in the xtce may have something to do with it. Just for the sake of not missing anything have you considered adding more unit tests?

All the code looks good.

imap_processing/glows/l0/glows_decom.py
laspsandoval: Agree with Greg. Glad he caught that for Ultra.

maxinelasp:
> Maybe it is my imagination, but the Expected Result and your result look ever so slightly different. I do wonder if the Fixed value parameter in the xtce may have something to do with it. Just for the sake of not missing anything have you considered adding more unit tests?

Like I mentioned as a reply to one of the other comments, the GLOWS team and I have discussed having some statistical tests for this data, so that's my ultimate plan to test the data. I think the graph provided to me by the team is the data that went into the simulator, so I would expect some minor differences. I'll definitely talk about testing this process with them when I meet with them later this month. 

Since the histogram value (with the FixedValue) should always have the same number of buckets, I think a FixedValue is the right way to go there. Unless part of the packet is missing, there should always be the same bucket count. I will confirm this with GLOWS.

imap_processing/glows/packet_definitions/P_GLX_TMSCHIST.xml
laspsandoval: With Ultra there were some nuances that I didn't figure out until I was given the Flight Software Specifications Table. Even after that we had to walk through their L0 decom code in C for a couple of hours and only then it became clear what was going on. My suggested questions for them:

1. Should the decom data be exactly the same as the simulator data (I do think that it should be)
2. Should this be a variable length packet?
3. Do they take only the first x values from the returned array? Ultra had hard-coded [0:48] into their C code to select only the first 48 values, which I never would have known about if I hadn't been walked through what they were doing.

I might just be paranoid from my own experience so maybe its all exactly as it should be :) I just recommend that you double check.

maxinelasp: Definitely will! Thank you for those suggested questions, that's helpful for understanding your concerns and what I should look into.

greglucas: I agree that this is a good thing to discuss with them, but I would tend to agree that this is almost certainly a fixed value. I think this comes down to differences between whether instruments detect individual events, or if they are integrating over some bins. So, I think the event-based instruments will accumulate individual events and therefore we don't know how long those packets will be until we get them. But, for the other instruments, like GLOWS, they will be integrating over some specific bin and just adding up those counts within those bins, so the number of bins is fixed and they just report the number of events/bin, not individual events. That is my guess at the reason for the fixed vs variable-length differences between instruments.

maxinelasp:
I'm merging this with the code coverage check failing, because the uncovered lines do not need testing (pretty print methods in the two data classes)

PR #176, opened 2023-09-28
Housekeeping data testing and validation for CoDICE l0

GFMoraga:
# Change Summary
This is the first pass at l0 decom for codice- specifically NHK, which is the raw housekeeping data. This data has been validated. Testing is done in this PR and it validates against a csv we recieved from the instrument team at swRi. 
 
## Overview

Updated the decom code in `test_decom.py` 
Added testing code that validates SHCOARSE


imap_processing/codice/l0/decom_nhk.py
```diff
@@ -0,0 +1,43 @@
...
+        return list(packet_generator)
+
+
+# Decompose packets
+decomposed_packets = decom_packets(packet_file, xtce_document)
+
+# Print decomposed packets
+for packet in decomposed_packets:
+    if packet.header["PKT_APID"].raw_value == 1136:
+        print(packet.data)
```
tech3371: can you move that into a test file and a test function.

imap_processing/codice/l0/housekeeping_data.bin
tech3371, line 1: can we move test data into `tests/` folder?

imap_processing/codice/l0/decom_nhk.py
tech3371, line 1: do you want to name this file `decom_codice` to be consistent with other instrument?

imap_processing/codice/l0/decom_nhk.py
```diff
@@ -0,0 +1,43 @@
+from pathlib import Path
+
+from space_packet_parser import parser, xtcedef
+
+from imap_processing import imap_module_directory
+
+# Define paths
+packet_file = Path("housekeeping_data.bin")
+xtce_document = Path(f"{imap_module_directory}/codice/packet_definitions/P_COD_NHK.xml")
```
tech3371: This would go in test file I was mentioning earlier and will be pass as input now.

imap_processing/codice/l0/decom_nhk.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def decom_packets(packet_file: str, xtce_packet_definition: str):
```
greglucas: I don't think you need this function if you import the shared decom_packets function?

```python
from imap_processing.decom import decom_packets
```

imap_processing/codice/l0/decom_nhk.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    packet_definition = xtcedef.XtcePacketDefinition(xtce_packet_definition)
+    packet_parser = parser.PacketParser(packet_definition)
+
+    with open(packet_file, "rb") as binary_data:
+        packet_generator = packet_parser.generator(binary_data)
+        return list(packet_generator)
+
+
+# Decompose packets
```
greglucas:
```suggestion
# Decommutated packets
```
To be consistent everywhere we should use `decommutated`. I think we are a bit inconsistent currently, so something to look into elsewhere for us too.

imap_processing/codice/l0/decom_nhk.py
greglucas: and assert what you expect the data to be here. See @tech3371 's SWE PR with the most recently updated data she received.

bourque:
I added an initial pass at some unit tests (basically took what @tech3371 has for SWE). We will have to think about what to add to validate CoDICE housekeeping data specifically, but this adds in the basic testing structure for now.

imap_processing/codice/tests/test_decom.py
```diff
@@ -0,0 +1,73 @@
...
+    # First way to get header data
+    ccsds_header = decom_test_data[0].header
+    assert all(key in ccsds_header.keys() for key in ccsds_header_keys)
+
+    # Second way to get header data
+    ccsds_header = decom_test_data[0][0]
+    assert all(key in ccsds_header.keys() for key in ccsds_header_keys)
+
+
+def test_ways_to_get_data(decom_test_data):
```
greglucas, line 88: This is just a note that we should discuss this as a group on what we want to test for each of the instruments because I think a few of us are duplicating testing the CCSDS header, which I think is covered by space_packet_parser, but we should verify this.

Additionally, this is sort of testing our dependencies and saying `space_packet_parser` needs to be able to retrieve data in both of these ways. Not necessarily codice data/classes you've defined must be that way.

greglucas: Discussed this on a call today. It sounds like the team would like these tests to be in a single location and tested once rather than each of the instruments needing to remember to add these tests. So I think if you could move this to the `test_decom.py` file or somewhere similar and think of how to do this for the general case that would be the way to go here.

bourque: This work will be addressed in #184

GFMoraga:
This is up for review to be merged. This is l0 decom and validation testing for the raw housekeeping bin, against the csv. There are no engineering conversions yet because that will be done in l1b.

imap_processing/codice/l0/decom_codice.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
+
+from imap_processing import imap_module_directory
+from imap_processing.decom import decom_packets
+
+if __name__ == "__main__":
+    # Define paths
+    packet_file = Path("housekeeping_data.bin")
+    xtce_document = Path(
+        f"{imap_module_directory}/codice/packet_definitions/P_COD_NHK.xml"
```
bourque: Does this decom work on the new files in the `data/` directory? If so, maybe we should use those since they come from the 'official' CoDICE testing data

imap_processing/codice/tests/test_decom.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Read in the CSV file (perhaps to a pandas dataframe?)
+    validation_file = Path(
+        f"{imap_module_directory}/codice/tests/data/"
+        f"idle_export_raw.COD_NHK_20230822_122700.csv"
+    )
+    validation_data = pd.read_csv(validation_file)
+    # Remove the timestamp column and data
+    if "timestamp" in validation_data.columns:
+        validation_data = validation_data.drop(columns=["timestamp"])
```
bourque: Nice, glad this was as easy thing to do!

imap_processing/codice/tests/test_decom.py
```diff
@@ -0,0 +1,104 @@
...
+    for key, value in first_data.data.items():
+        if key == "SHCOARSE":
+            # Compare SHCOARSE value
+            assert value.raw_value == validation_row.name
+            continue
+
+        # Check if the key is present in validation_row
+        if key not in validation_row.index:
+            print(f"Key '{key}' not found in validation_row. Skipping comparison.")
+            continue
```
bourque: So this is suggesting that there are key/value pairs in the housekeeping packet that are not in the validation file? Would this then mean that the validation data are missing some items? 

If that is the case, this is probably something we want to ask Joey and/or Greg Dunn about. I would think we would ideally want a one-to-one mapping?

imap_processing/codice/l0/decom_codice.py
greglucas: I don't think we want this guarded in a `__name__` block. This is a module, so you can import the code in other locations (i.e. the tests)

imap_processing/codice/l0/decom_codice.py
```diff
@@ -0,0 +1,19 @@
...
+    )
+
+    # Decommutated packets
+    decomposed_packets = decom_packets(packet_file, xtce_document)
+
+    # Print decommutated packets
+    for packet in decomposed_packets:
+        if packet.header["PKT_APID"].raw_value == 1136:
+            print(packet.data)
```
greglucas: Do you want to return a list of the packets or append these together so you can access them externally?
```python
return [packet for packet in decommutated_packets if packet.header["PKT_APID"].raw_value == 1136]
```

imap_processing/codice/tests/test_decom.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+    xtce_document = Path(
+        f"{imap_module_directory}/codice/packet_definitions/P_COD_NHK.xml"
+    )
+    data_packet_list = decom_packets(packet_file, xtce_document)
+    data_packet_list = [
+        packet
+        for packet in data_packet_list
+        if packet.header["PKT_APID"].raw_value == 1136
+    ]
```
greglucas: This is the same code that you've got in you `__name__` block above I think. Could you somehow abstract that logic out, then import that function here and use it to generate your list of packets?

imap_processing/codice/tests/test_decom.py
greglucas: Do you need to drop the timestamp? Or can you just ignore it later in your processing if it isn't used there.

Also, I think you can do this all in one line and ignore the errors if that column doesn't exist.
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html

```python
validation_data.drop(columns=["timestamp"], inplace=True, errors="ignore")
```

imap_processing/codice/tests/test_decom.py
greglucas: Probably shouldn't print in tests. You should either fail or pass.

Also, I think this would be easier to parse as an `if-elif-else` clause rather than using continues.

imap_processing/codice/l0/decom_codice.py
tech3371: You could do similar to what I have done for SWE and SWAPI.
```suggestion
from imap_processing import decom, imap_module_directory


def decom_packets(packet_file: str):
    """Decom CoDICE data packets using CoDICE packet definition.

    Parameters
    ----------
    packet_file : str
        Path to data packet path with filename.

    Returns
    -------
    List
        List of all the unpacked data
    """
    # TODO: how to call other packet's definition? May be, you could pass another optional input that passes apid and based on apid, you can grab correct packet definition?
    xtce_document = (
       f"{imap_module_directory}/codice/packet_definitions/P_COD_NHK.xml"
    )
    return decom.decom_packets(packet_file, xtce_document)
```

imap_processing/codice/tests/test_decom.py
```diff
@@ -0,0 +1,104 @@
+"""Tests the decommutation process for CoDICE CCSDS Packets"""
+
+from pathlib import Path
+
+import pandas as pd
+import pytest
+
+from imap_processing import imap_module_directory
+from imap_processing.decom import decom_packets
+
```
tech3371, line 10:
```suggestion
from imap_processing.codice import decom_codice

```

imap_processing/codice/tests/test_decom.py
tech3371:
```suggestion
    data_packet_list = decom_codice.decom_packets(packet_file)

imap_processing/codice/tests/test_decom.py
tech3371: I agree with others. If key is not in validation data, then packet definitions may have changed. It should fail if it didn't find the key.

GFMoraga:
I had to change a few things that where outside the general change summary. I had to update the `conf.py` due to an external issue. Names where changed and data was swapped.

imap_processing/codice/l0/decom_codice.py
GFMoraga: Decided not to do this this PR but will be looking into it on the next housekeeping PR.

bourque (approved):
Nice work here @GFMoraga!

PR #179, opened 2023-09-29
L1 cdf creation

### LINKED ISSUE #195: Create CDF files for IDEX Level 1

bryan-harter:
# Change Summary

## Overview
Adding the ability to create CDF files from the L1 xarray data for IDEX

## New Dependencies
cdflib 

## New Files
- global_cdf_attrs.py
   - A place to store all common CDF attribute dictionaries
- idex_cdf_attrs.py
  - A place to store all IDEX specific CDF attribute dictionaries 

## Updated Files
- idex_packet_parser.py
   - Added the function "write_l1_cdf".  When called it generates a l1 CDF file from the L1 data

## Testing
- test_l1_cdfs.py
  - Unit tests for the CDF file creation


bourque:
@all-contributors please add @bryan-harter for Code!

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+
+global_base = {
+    "Project": ["STSP Cluster>Solar Terrestrial Science Programmes, Cluster"],
```
greglucas: Why do we have Cluster here?

I would have thought it'd be "STP>Solar Terrestrial Probes"
https://science.nasa.gov/heliophysics/programs/solar-terrestrial-probes

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,29 @@
...
+    "Source_name": ["IMAP>Interstellar Mapping and Acceleration Probe"],
+    "Discipline": ["Solar Physics>Heliospheric Physics"],
+    "Descriptor": ["IDEX>Interstellar Dust Experiment"],
```
greglucas: This should only be in the idex group I think?

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,29 @@
...
+    "PI_name": ["Dave McComas"],
```
greglucas:
```suggestion
    "PI_name": ["David J. McComas"],
```
I assume we should use full qualified names?

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,29 @@
...
+    "PI_affiliation": ["Princeton"],
```
greglucas: Princeton University?

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,29 @@
...
+    "Instrument_type": ["Particles (space)"],
+    "Mission_group": ["IMAP"],
```
greglucas: We put IMAP here, but we never define IMAP, do you need to say what IMAP stands for somewhere?

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,29 @@
...
+}
+
+epoch_attrs = {
+    "CATDESC": "Default time",
+    "FIELDNAM": "Epoch",
+    "FILLVAL": np.array([-9223372036854775808]),
```
greglucas:
```suggestion
    "FILLVAL": np.iinfo(np.int64).min,
```
Why are these an array, should they be just a single value? Also might help to say it is the minimum possible int64?

Same for the other values below this.

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -0,0 +1,148 @@
+import numpy as np
+
+from .. import common_cdf_attrs
```
greglucas:
```suggestion
from imap_processing import common_cdf_attrs
```

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Set IDEX software version here for now
+software_version = "01"
```
greglucas:
```suggestion
__version__ = "01"
```
Do we want to follow more standard Python version naming schemes here?

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Global Attributes
+idex_l1_global_attrs = (
+    {
+        "Data_type": ["l1"],
+        "Data_version": [software_version],
+        "TEXT": [
+            """Greg Newcomb, the flight software (FSW) engineer for the Interstellar
```
greglucas: This is going into every file? This seems more like a test/verification note.

imap_processing/idex/idex_packet_parser.py
```diff
@@ -3,10 +3,14 @@
 import bitstring
 import numpy as np
 import xarray as xr
+from cdflib.xarray import xarray_to_cdf
 from space_packet_parser import parser, xtcedef
 
 from imap_processing import imap_module_directory
 
+from .. import common_cdf_attrs
+from . import idex_cdf_attrs
```
greglucas:
```suggestion
from imap_processing import common_cdf_attrs
import idex_cdf_attrs
```

imap_processing/idex/idex_packet_parser.py
```diff
@@ -89,6 +93,48 @@ def __init__(self, packet_file: str):
 
         self.data = xr.concat(processed_dust_impact_list, dim="Epoch")
 
+    def write_l1_cdf(self, description: str = "", version: str = "01"):
```
greglucas: Is this version different from the software version? Would help to clarify what this version is based upon.

imap_processing/idex/idex_packet_parser.py
```diff
@@ -89,6 +93,48 @@ def __init__(self, packet_file: str):
...
+        # Convert the xarray object to a CDF!
+
+        try:
+            xarray_to_cdf(
+                self.data, filename, terminate_on_warning=True
+            )  # Terminate if not ISTP compliant
+        except Exception as e:
+            logging.error("Error writing a CDF file! Aborting....")
+            logging.error(str(e))
+            return
```
greglucas: Why not just let this raise? I don't think we need this in a try/except. You say terminate, but then you just return from the exception instead of re-raising.

imap_processing/idex/tests/test_l1_cdfs.py
```diff
@@ -0,0 +1,79 @@
...
+@pytest.fixture(scope="session")
+def decom_test_data():
+    return PacketParser("imap_processing/idex/tests/imap_idex_l0_20230725_v01-00.pkts")
+
+
+def test_idex_cdf_file(decom_test_data):
+    # Verify that a CDF file can be created with no errors thrown by xarray_to_cdf
+    file_name = decom_test_data.write_l1_cdf(version="01", description="")
+    assert file_name == "imap_idx_l1_20250724_0000_p_v01.cdf"
+    os.remove(file_name)
```
greglucas: If a failure were to happen this would leave files around in the current directory. Should your writer take a directory path as well for where to put the file?
https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

```python
with tempfile.TemporaryDirectory() as tmpdir:
    ...
```
which will clean up for you properly.

This also applies to all the other tests as well.

imap_processing/idex/tests/test_l1_cdfs.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_bad_cdf_attributes(decom_test_data):
+    # Deliberately mess up the attributes to verify that no CDF is created
+    copy_to_modify = copy.deepcopy(decom_test_data)
```
greglucas: Rather than needing to deepcopy this, we should just update the fixture above to not be `session`, so it will create a new data file every single time if you need a fresh one.

greglucas (commented):
Overall looks like a good way to handle the writing to me.

imap_processing/idex/idex_packet_parser.py
bryan-harter: It isn't really based on anything at the moment, it just writes the version number to the end of the file name.  It should probably be based on the IDEX software version, but I didn't think we had reached any conclusions yet about how software versions corresponded to file versions, so I left it open-ended for now.

imap_processing/idex/idex_packet_parser.py
bryan-harter: I'll change that. I initially just figured if we wanted to write multiple IDEX files we didn't want the whole program to crash.  But we can just have the try/except located outside of whatever calls this function in that case.

imap_processing/idex/tests/test_l1_cdfs.py
bryan-harter: Added

imap_processing/idex/tests/test_l1_cdfs.py
bryan-harter: changed

bryan-harter (commented):
I apparently need to submit a review for my comments to go through

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,31 @@
+import numpy as np
+
+global_base = {
+    "Project": ["STP>Solar-Terrestrial Physics"],
+    "Source_name": ["IMAP>Interstellar Mapping and Acceleration Probe"],
+    "Discipline": ["Solar Physics>Heliospheric Physics"],
+    "PI_name": ["Dr. David J. McComas"],
+    "PI_affiliation": [
+        "Princeton Plasma Physics Laboratory",
+        "100 Stellarator Road, Princeton, NJ 08540",
```
maxinelasp: I wish I lived on Stellarator road...

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -0,0 +1,149 @@
...
+        "Logical_source_description": ["IMAP Mission IDEX Instrument Level-1 Data.  "],
+    }
+    | common_cdf_attrs.global_base
+)
+
+idex_l2_global_attrs = (
+    {
+        "Data_type": ["l2"],
+        "Data_version": [__version__],
+        "TEXT": [
```
maxinelasp: Could we move the common attributes into one variable? Or have some kind of `base_dict` which you can then expand?

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -0,0 +1,149 @@
...
+
+# L1 variables base dictionaries (these are not complete)
+l1_data_base = {
+    "DEPEND_0": "Epoch",
+    "DISPLAY_TYPE": "spectrogram",
+    "FILLVAL": np.array([-1.0e31], dtype=np.float32),
+    "FORMAT": "E12.2",
+    "UNITS": "dN",
+    "VALIDMIN": np.array([-1.0e31], dtype=np.float32),
+    "VALIDMAX": np.array([1.0e31], dtype=np.float32),
```
maxinelasp: Similarly, can all these magic numbers get moved to a constant somewhere and maybe commented on why those values were chosen?

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -0,0 +1,149 @@
...
+        ],
+        "Mission_group": ["IMAP"],
+        "Logical_source": ["imap_idx_l2"],
+        "Logical_file_id": ["FILL ME IN AT FILE CREATION"],
+        "Logical_source_description": ["IMAP Mission IDEX Instrument Level-2 Data"],
+    }
+    | common_cdf_attrs.global_base
+)
+
+# L1 variables base dictionaries (these are not complete)
```
maxinelasp: What does it take for these to be complete? Is it complete enough for others to use as an example (because it's only missing IDEX specific stuff) or are there other reasons?

I don't mind if you merge them incomplete btw, just curious what is left.

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -0,0 +1,149 @@
...
+    "CATDESC": "TOF_High",
+    "FIELDNAM": "TOF_High",
+    "VAR_NOTES": "This is the high gain channel of the time-of-flight signal.",
+} | l1_tof_base
+
+tof_mid_attrs = {
+    "CATDESC": "TOF_Mid",
+    "FIELDNAM": "TOF_Mid",
+    "VAR_NOTES": "This is the mid gain channel of the time-of-flight signal.",
+} | l1_tof_base
```
maxinelasp, line 129: I didn't know you could use `|` like this. Good new info!

imap_processing/common_cdf_attrs.py
bryan-harter: Fixed, and also added "Dr." because thats how the one IBEX CDF file I found had it

imap_processing/common_cdf_attrs.py
bryan-harter: Fixed

imap_processing/common_cdf_attrs.py
bryan-harter: Fixed

imap_processing/common_cdf_attrs.py
bryan-harter: Yeah I added more detailed here, including an address

maxinelasp (commented):
I apparently didn't submit this two days ago. If the comments are outdated, just ignore them

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -0,0 +1,167 @@
+import numpy as np
+
+from imap_processing import common_cdf_attrs
+
+# Set IDEX software version here for now
+__version__ = "1"
```
tech3371:
```suggestion
from imap_processing import common_cdf_attrs
from imap_processing.idex import version
```
Then in `imap_processing/idex/__init__.py`
```
version = "01"
```
I feel like we will use this variable a lot in different places. This way, it makes it easy to import anywhere. What do you think?

imap_processing/idex/idex_cdf_attrs.py
tech3371: I was thinking that as I was reading above code!

imap_processing/idex/idex_packet_parser.py
```diff
@@ -89,6 +91,45 @@ def __init__(self, packet_file: str):
...
+        -------
+            str
+                The name of the file created
+        """
+        # Determine the start date of the data in the file,
+        # based on the time of the first dust impact
+        file_start_date = self.data["Epoch"][0].data
+        date_string = np.datetime_as_string(file_start_date, unit="D").replace("-", "")
+        if description:  # Just append an underscore to the description
+            description = f"_{description}"
```
tech3371: To avoid double underscore
```suggestion
            description = description if description.startswith("_") else f"_{description}"
```

imap_processing/idex/idex_packet_parser.py
```diff
@@ -89,6 +91,45 @@ def __init__(self, packet_file: str):
...
+        # Set file name for IDEX L1 based on the date string above and the version
+        filename = f"imap_idx_l1_{date_string}_0000{description}_p_v{version}.cdf"
+        filename_and_path = os.path.join(directory, filename)
+        # Set the global attributes in the xarray
+        self.data.attrs = idex_cdf_attrs.idex_l1_global_attrs
+        # The Logical_file_id is always the name of the file without the extension
```
tech3371: good to know

imap_processing/idex/tests/test_l1_cdfs.py
```diff
@@ -0,0 +1,91 @@
...
+
+def test_descriptor_in_file_name(decom_test_data):
+    # Deliberately mess up the data to verify no CDF is created
+    with tempfile.TemporaryDirectory() as tmpdir:
+        file_name = decom_test_data.write_l1_cdf(
+            version="01", description="impact-lab-test001", directory=tmpdir
+        )
+        assert file_name == os.path.join(
+            tmpdir, "imap_idx_l1_20250724_0000_impact-lab-test001_p_v01.cdf"
+        )
```
tech3371: How did this check that there is no CDF file? out of curiosity

tech3371 (commented):
Looks really good and complicated! 

Can you write a guidance of how to write instrument specific attrs dictionary. Yours had lot of good information and I wouldn't know where to get those informations. Does those instrument specific attrs stores information about xarray.DataArray's dimension?

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,31 @@
+import numpy as np
+
+global_base = {
+    "Project": ["STP>Solar-Terrestrial Physics"],
```
greglucas: I find it odd that these are all lists. Is that what it is supposed to be or are they supposed to be single entities?

`Project` is singular, so should it be `Projects` as the lookup to indicate that these need to be lists? Will `PI_affiliation` below be two entries, or a single concatenated string? That reads as two entries currently, but I think they are related with name/address.

imap_processing/common_cdf_attrs.py
```diff
@@ -0,0 +1,31 @@
...
+        "100 Stellarator Road, Princeton, NJ 08540",
+    ],
+    "Instrument_type": ["Particles (space)"],
+    "Mission_group": ["IMAP>Interstellar Mapping and Acceleration Probe"],
+}
+
+epoch_attrs = {
+    "CATDESC": "Default time",
+    "FIELDNAM": "Epoch",
+    "FILLVAL": np.iinfo(np.int64).min,
```
greglucas: I feel like a lot of my comments are getting lost when I say to update them here and also below, you got here, but not the subsequent ones where this same pattern would be helpful (removing magic numbers as @maxinelasp pointed out as well). Would it help to comment in every place? I find that to be a lot of noise, but I'm happy to do it too if it'd be helpful.

imap_processing/idex/tests/test_l1_cdfs.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_idex_cdf_file(decom_test_data):
+    # Verify that a CDF file can be created with no errors thrown by xarray_to_cdf
+    with tempfile.TemporaryDirectory() as tmpdir:
+        file_name = decom_test_data.write_l1_cdf(
+            version="01", description="", directory=tmpdir
+        )
+        assert file_name == os.path.join(tmpdir, "imap_idx_l1_20250724_0000_p_v01.cdf")
+        os.remove(file_name)
```
greglucas: you don't need to remove anything, that is the nice thing about the context manager here.
```suggestion
```

imap_processing/idex/tests/test_l1_cdfs.py
```diff
@@ -0,0 +1,91 @@
...
+    return PacketParser("imap_processing/idex/tests/imap_idex_l0_20230725_v01-00.pkts")
+
+
+def test_idex_cdf_file(decom_test_data):
+    # Verify that a CDF file can be created with no errors thrown by xarray_to_cdf
+    with tempfile.TemporaryDirectory() as tmpdir:
+        file_name = decom_test_data.write_l1_cdf(
+            version="01", description="", directory=tmpdir
+        )
+        assert file_name == os.path.join(tmpdir, "imap_idx_l1_20250724_0000_p_v01.cdf")
```
greglucas: This is a bit cryptic how you know that this is the proper filename. Should we add some of these identifiers to the `write_l1_cdf` routine so that routine can take in dates, software versions? Or another possibility is to make the string here be composed from those attributes `imap_idx_{level}_{date}_{version?}_{identifer?}_{version?}`
```suggestion
        assert file_name == str(tmp_path / "imap_idx_l1_20250724_0000_p_v01.cdf")
        assert Path(file_name).exists()
```
We should also assert that the file exists?

imap_processing/idex/tests/test_l1_cdfs.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.fixture()
+def decom_test_data():
+    return PacketParser("imap_processing/idex/tests/imap_idex_l0_20230725_v01-00.pkts")
+
+
+def test_idex_cdf_file(decom_test_data):
+    # Verify that a CDF file can be created with no errors thrown by xarray_to_cdf
+    with tempfile.TemporaryDirectory() as tmpdir:
```
greglucas: I just realized you can actually use the `tmpdir` or `tmp_path` fixture so we don't even need to do this.

```suggestion
def test_idex_cdf_file(decom_test_data, tmp_path):
    # Verify that a CDF file can be created with no errors thrown by xarray_to_cdf
```

imap_processing/idex/tests/test_l1_cdfs.py
greglucas: Agreed, this should be something like: `assert not file_name.exists()`, but also do we want it to just continue on or should we raise an error/emit a warning if nothing gets created?

imap_processing/idex/tests/test_l1_cdfs.py
```diff
@@ -0,0 +1,91 @@
...
+            tmpdir, "imap_idx_l1_20250724_0000_impact-lab-test001_p_v01.cdf"
+        )
+        os.remove(file_name)
+
+
+def test_idex_tof_high_data_from_cdf(decom_test_data):
+    # Verify that a sample of the data is correct inside the CDF file
+    # impact_14_tof_high_data.txt has been verified correct by the IDEX team
+    with open("imap_processing/idex/tests/impact_14_tof_high_data.txt") as f:
+        data = np.array([int(line.rstrip("\n")) for line in f])
```
greglucas: rstrip() should remove the trailing newline by default?
```suggestion
        data = np.array([int(line.rstrip()) for line in f])
```

imap_processing/common_cdf_attrs.py
greglucas, line 1: I don't love this name. What about just calling it `imap_processing/cdf_utils.py` or something like that? Do we want an entire `cdf` module with additional utilities? `cdf/global_attributes.py`, `cdf/utils.py`, all nested under a single directory.

greglucas (commented):
I don't quite get why the other comments are getting buried in the main thread and not showing up in the code review area for me... I see there are a lot of other comments, but I only see them under the conversation tab. I'll try and look into this...

imap_processing/idex/idex_cdf_attrs.py
bryan-harter: They're not "complete" in that they can't used as variable attributes as it is, they need more info and get completed by the dictionaries below it

imap_processing/common_cdf_attrs.py
bryan-harter: From cdflib's perspective it doesn't matter if they are lists or not, cdflib will insert them into a list if they aren't already. This just saves cdflib a step. 

Each member of the list is an "entry" of the attribute. Attributes can have multiple entries. The SPDF asked me to put the address as a separate entry of "PI_Affiliation".  

"Project" is the ISTP standard attribute, and their tools will complain if given the attribute name "Projects", even if there are multiple projects the data comes from.

bryan-harter:
Ok I think I finally responded to all comments, and the code now makes files that are PDS compliant.  Summary of the major changes since a few weeks ago:

- Renamed "cdf_common_attrs.py" to "cdf_utils.py"
- Added a function "write_cdf()" to the cdf_utils, which writes a CDF file based on the global attributes of the xarray object
- Cleaned up the unit tests 
- Commented or replaced some of the magic numbers
- Modified quite a few of the values in the attribute dictionaries based on feedback from the SPDF

I also made a couple tweaks to cdflib.xarray_to_cdf, so the required cdflib version is now 1.2.3

maxinelasp (approved):

imap_processing/cdf_utils.py
```diff
@@ -0,0 +1,102 @@
...
+    "VAR_TYPE": "support_data",
+    "SCALETYP": "linear",
+    "MONOTON": "INCREASE",
+    "TIME_BASE": "J2000",
+    "TIME_SCALE": "Terrestrial Time",
+    "REFERENCE_POSITION": "Rotating Earth Geoid",
+}
+
+
+def write_cdf(data: xr.Dataset, description: str = "", directory: str = ""):
```
tech3371, line 46: I am writing similar function in my working branch to make it more generic that we all can use. Later, I can combine this and mine into one and test it for boths.

imap_processing/cdf_utils.py
```diff
@@ -0,0 +1,102 @@
...
+        + f"_v{data.attrs['Data_version']}.cdf"
+    )
+    filename_and_path = os.path.join(directory, filename)
+
+    # Insert the final attribute:
+    # The Logical_file_id is always the name of the file without the extension
+    data.attrs["Logical_file_id"] = filename.split(".")[0]
+
+    # Convert the xarray object to a CDF
+    xarray_to_cdf(
```
tech3371, line 95: Did you notice that it doesn't overwrite existing file with new changes? I noticed that when I was testing it locally. I had to delete old files and run again to see new changes.

tech3371 (approved):

PR #184, opened 2023-10-03
Test CCSDS header across all instrument packet definitions

bourque:
# Change Summary

## Overview
This PR adds a new folder for storing tests that can be run on all instruments, and a new test that tests the existence and proper formatting of the CCSDS header in all packet definition XML files.

## New Dependencies
None

## New Files

- `imap_processing/tests/test_decom.py`
   - Tests for the decommutation process, applied to all instruments

## Deleted Files

None

## Updated Files

None

## Testing

This test is currently failing on the `idex_packet_defintiion.xml` file (which currently doesn't contain a properly formatted header) and a handful of others (due to inconsistency with considering `SHCOARSE` as part of the header or not).


bourque:
Some open questions for discussion:

1. Is `SHCOARSE` to be considered part of the header or not? Should we be consistent on how this is applied across all packet definition files?
2. Is this a sound method for testing the header (i.e. reading in the file and comparing strings)? Or should we try to actually decom each file and check if the header is in the results?
3. Are there any other generalized tests we can add to this file?

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,61 @@
+"""Tests aspects of the decommutation process that is generic to all instruments"""
+
+import glob
+
+import pytest
+
+from imap_processing import imap_module_directory
+
+xtce_documents = glob.glob(f"{imap_module_directory}/*/packet_definitions/*.xml")
```
greglucas:
```suggestion
xtce_documents = imap_module_directory.glob("*/packet_definitions/*.xml")
```
imap_module_directory is a Path object so I think you can get rid of the glob import and use it directly on the Path instead.

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,61 @@
...
+xtce_documents = [
+    item for item in xtce_documents if "idex_packet_definition.xml" not in item
+]  # Temporary until IDEX packet definition is fixed
```
greglucas: Use `pytext.mark.xfail()` inside your test instead?

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.mark.parametrize("xtce_document", xtce_documents)
+def test_ccsds_header(xtce_document):
+    """Test if the XTCE document contains the proper CCSDS header information"""
+
+    print(f"\nTesting {xtce_document}")
```
greglucas:
```suggestion
```
This is already in the really long parameterization name of the tests.
`imap_processing/tests/test_decom.py::test_ccsds_header[/home/runner/work/imap_processing/imap_processing/imap_processing/ultra/packet_definitions/P_U45_IMAGE_RATES.xml]`
You might be able to trim that somehow to make a nicer `id=filename` somehow, but not sure how much effort that would be.

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # Define what is expected in the XTCE document
```
greglucas, line 24: Reference here for what space packet parser is checking for I believe.
https://github.com/medley56/space_packet_parser/blob/bef50e6bfd262fdef4727381c5dafea6488571c2/space_packet_parser/parser.py#L20-L27

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,61 @@
...
+    header_keys = [
+        "VERSION",
+        "TYPE",
+        "SEC_HDR_FLG",
+        "PKT_APID",
+        "SEG_FLGS",
```
greglucas:
```suggestion
        "SEQ_FLGS",
```
Do we have a typo in all of these?

greglucas (commented):
I think this looks like a nice approach and is also a good way to make sure we are consistent as a team too!

Those are all really good questions that I don't know the answer to :)

imap_processing/tests/test_decom.py
bourque: Oh cool! I didn't realize there was a `glob` method on `Path` objects.

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.mark.parametrize("xtce_document", xtce_documents)
+def test_ccsds_header(xtce_document):
+    """Test if the XTCE document contains the proper CCSDS header information"""
+
+    print(f"\nTesting {xtce_document}")
+
+    # Define what is expected in the XTCE document
+    header_keys = [
```
GFMoraga, line 25: I checked on these and I think they are all correct, except "SEG_FLGS". This is from space_packet_parser: 

CCSDS_HEADER_DEFINITION = [
    CcsdsPacketHeaderElement('VERSION', 'uint:3'),
    CcsdsPacketHeaderElement('TYPE', 'uint:1'),
    CcsdsPacketHeaderElement('SEC_HDR_FLG', 'uint:1'),
    CcsdsPacketHeaderElement('PKT_APID', 'uint:11'),
    CcsdsPacketHeaderElement('SEQ_FLGS', 'uint:2'),
    CcsdsPacketHeaderElement('SRC_SEQ_CTR', 'uint:14'),
    CcsdsPacketHeaderElement('PKT_LEN', 'uint:16')

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,61 @@
...
+    ]
+    header_container_set = (
+        "<xtce:ContainerSet>"
+        '    <xtce:SequenceContainer name="CCSDSPacket">'
+        "            <xtce:EntryList>"
+        '                    <xtce:ParameterRefEntry parameterRef="VERSION" />'
+        '                    <xtce:ParameterRefEntry parameterRef="TYPE" />'
+        '                    <xtce:ParameterRefEntry parameterRef="SEC_HDR_FLG" />'
+        '                    <xtce:ParameterRefEntry parameterRef="PKT_APID" />'
+        '                    <xtce:ParameterRefEntry parameterRef="SEG_FLGS" />'
```
GFMoraga: Need to fix typo here. Also We need to go in and change the `ccsds_header_xtce_generator.py`

GFMoraga (commented):
This is great! Need to change the typo on a broader scale but that can be done in another PR

imap_processing/tests/test_decom.py
maxinelasp: so does that mean SHCOARSE or MET are not defined in space_packet_parser? I did notice that in the excel file, SHCOARSE is called a header, but in the code, it is in the data section.

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,61 @@
...
+    header_keys = [
+        "VERSION",
+        "TYPE",
+        "SEC_HDR_FLG",
+        "PKT_APID",
+        "SEG_FLGS",
+        "SRC_SEQ_CTR",
+        "PKT_LEN",
+    ]
+    header_container_set = (
```
maxinelasp: Related to above, header_container_set does include SHCOARSE while the above code doesn't. Do we want those to match? Maybe we can add a check for MET _or_ SHCOARSE?

maxinelasp (approved):
Looks good as a first pass! Definitely a few outstanding questions but I don't think any that would prevent a merge.

imap_processing/tests/test_decom.py
bourque: Probably not really necessary to have anything printed here. It was just handy while I was developing this.

imap_processing/tests/test_decom.py
bourque: It turns out we do have a bunch of packet definition files that have this typo, as @GFMoraga pointed out to me. He is going to fix the files in a separate PR.

imap_processing/tests/test_decom.py
bourque: Yes, these should match, and I will update accordingly depending on if we decide that `SHCOARSE` is part of the header or not.

I am inclined to say it doesn't belong in the header, because that is also how [`space_packet_parser` defines it](https://github.com/medley56/space_packet_parser/blob/bef50e6bfd262fdef4727381c5dafea6488571c2/space_packet_parser/parser.py#L20-L27).

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,74 @@
...
+    with open(xtce_document) as f:
+        document = f.read()
+
+    # Check that each header key is defined as a Parameter
+    for header_key in header_keys:
+        assert (
+            f'<xtce:Parameter name="{header_key[0]}" parameterTypeRef="{header_key[1]}"'
+            in document
+            or f'<xtce:Parameter name="{header_key[0]}" parameterTypeRef="{header_key[1].lower()}"'
+            in document
```
tech3371: Were some of the header lower case?

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,74 @@
...
+# Need to remove IDEX from the list in order to append it again with proper marking
+xtce_documents = [item for item in xtce_documents if item != idex_document]
+xtce_documents.append(
+    pytest.param(
+        idex_document,
+        marks=pytest.mark.xfail(
+            reason="Packet Definition does not include properly formatted CCSDS Header"
+        ),
+    )
+)
```
tech3371: can we create function like `decom_test_data()` and move all these into it? just curious.

imap_processing/tests/test_decom.py
bourque: This was resolved by #185

imap_processing/tests/test_decom.py
bourque: Thanks @GFMoraga! This was resolved by #185

bourque:
Ok, I believe this is good to go and the tests should pass once SWE, Ultra, and GLOWS update their XTCE files with the fixes implemented in #185

maxinelasp:
Updated GLOWS XTCE to pass these tests in #244

imap_processing/tests/test_decom.py
greglucas: Turning this into a fixture does seem like a good idea. https://docs.pytest.org/en/7.3.x/how-to/parametrize.html#

imap_processing/tests/test_decom.py
bourque: Yes, in the Ultra packet definitions, the `parameterTypeRef` were defined in lower case (e.g. `uint3` instead of `UINT3`)

imap_processing/tests/test_decom.py
bourque: I've been trying to think of a creative solution to make this work with a fixture that returns a list and then uses said list as a parameterize. It appears that it is not straightforward (https://stackoverflow.com/questions/52764279/pytest-how-to-parametrize-a-test-with-a-list-that-is-returned-from-a-fixture), but perhaps I am missing something.

I could use a fixture and then iterate over the list within the test function, but that reduces the test module down to one test, which makes it slightly more difficult to see how many XML files are failing.

greglucas: Hm... maybe not worth it then. But, let me give you another potential idea.

Setting up a fixture, and yielding items. But, here you only want to mark one of the items in the list that is being yielded? So I think you can modify the param being returned somehow / add the xfail mark in some way. Here is some pseudo-code of what I think might work.

```python
@pytest.fixture(params=full_list_of_items)
def xtce_documents(request):
    if request.param.contains("idex"):
        yield pytest.param(request.param, mark=pytest.xfail)
    yield request.param
```

But again, if you don't think this is worth it and would take too much effort, then I'd say don't worry about it here. It'd just be a nice to have, not a requirement.

bourque: Thanks for the suggestion! Implementing something like this helped clean up the code a bit.

bourque:
pre-commit.ci autofix

bourque:
I think we still need to update the packet definitions for SWE and Ultra with the changes from #185 to get these tests to pass.

@tech3371 @laspsandoval Could you update those files for SWE and Ultra (respectively)? Or if it is straightforward, I am happy to do it as part of this PR if you could share the Excel spreadsheet.

imap_processing/swe/packet_definitions/swe_packet_definition.xml
```diff
@@ -61,47 +60,42 @@
...
-                <xtce:UnitSet/>
                 <xtce:IntegerDataEncoding sizeInBits="1" encoding="unsigned"/>
                 <xtce:EnumerationList>
                     <xtce:Enumeration label="STAR_TRACKER" value="0"/>
                     <xtce:Enumeration label="SUN_SENSOR" value="1"/>
                 </xtce:EnumerationList>
             </xtce:EnumeratedParameterType>
             <xtce:EnumeratedParameterType name="ACTIVE_ENUM" signed="false">
-                <xtce:UnitSet/>
+
```
greglucas: Reason for the extra line here? Did our robot friends decide to mess with us 😂

imap_processing/tests/test_decom.py
```diff
@@ -0,0 +1,66 @@
...
+import pytest
+
+from imap_processing import imap_module_directory
+
+# Scrape the repo for all packet definition XML files
+xtce_document_list = imap_module_directory.glob("*/packet_definitions/*.xml")
+
+
+@pytest.fixture(params=xtce_document_list, scope="session")
+def xtce_document(request):
```
greglucas, line 12: Nice work figuring this out! I think this approach makes sense.

greglucas (commented):
Just an FYI, but you can always xfail Ultra for now and update it later the same way you're doing with idex I think? Or do you want to update idex as well and remove all xfails altogether? Just giving you an out to get your tests passing so you can merge this in and then let others update and remove xfails later so the onus isn't on you to do that work.

imap_processing/swe/packet_definitions/swe_packet_definition.xml
bourque: Haha I think so? I don't recall putting that there! I will remove it.

imap_processing/ultra/packet_definitions/P_U45_AUXILIARY.xml
```diff
@@ -358,14 +358,14 @@
 		</xtce:ParameterSet>
 		<!-- End metadata -->
 		<xtce:ContainerSet>
-			<xtce:SequenceContainer name="CCSDSPacket" abstract="true">
```
bourque, line 361: @laspsandoval Do you know what `abstract="true"` does here? I had to remove it in order to get these tests to pass, and the other instrument's packet definitions don't have it. I am thinking that it is not important?

bourque: I now see that removing this caused `test_decom_apid_88*.py` to fail, so I guess it is important afterall!

bourque: I ended up just making my header tests a bit more generalized to allow for this `abstract` parameter.

bourque:
> Just an FYI, but you can always xfail Ultra for now and update it later the same way you're doing with idex I think? Or do you want to update idex as well and remove all xfails altogether? Just giving you an out to get your tests passing so you can merge this in and then let others update and remove xfails later so the onus isn't on you to do that work.

It was an easy fix to get the Ultra packet definitions to pass, so I just did that. I think I will leave IDEX as `xfail` for now -- that one will require more work.

PR #244, opened 2023-10-05
Updating GLOWS XTCE files with generator bugfix

maxinelasp:
# Change Summary
Updated with new generator code from PR #185 and checked against #184 to confirm the new files pass those tests.

## Overview

Reran xtce_generator_glows and recreated combined file. This also created the new long and short descriptions for the parameters.

This also updated SHCOARSE to MET (the base excel file now has MET), so I also updated the tests.

I also made a minor change to telemetry_generator.py, as I noticed an extra line in the BinaryParameterType that wasn't right. It doesn't seem to affect decom at all as it was just a random end tag.


tools/xtce_generation/telemetry_generator.py
```diff
@@ -132,8 +132,6 @@ def create_parameter_types(self, parameter_type_set, unique_lengths):
                 )
                 binary_parameter_type.attrib["name"] = parameter_type_ref_name
 
-                Et.SubElement(binary_parameter_type, "xtce:UnitSet")
```
maxinelasp: I don't think this is necessary? It just created one line for me:
`<xtce:UnitSet />` and it isn't in any of the other type parameter generator flows. So I just removed it. Looks like it dates back to the original pass of this file, so I suspect it was just something extra that fell through the cracks when we added the byte/sint/uint flow.

greglucas (approved):
Looks good to me. One question I have is if we want everyone to be updating their own instrument's files by rerunning the spreadsheets, or if we want to set up a script to do that automatically for us anytime the xtce_generator gets updated. I can see arguments both ways...

maxinelasp:
Probably we want everyone doing it manually. I still have a few manual steps (in the combined sheet) and I'm not sure if it's worth making everyone update the generator for ALL use cases (for example, variable length fields are also not created with the generator)

maxinelasp:
I'm open to the idea of creating an automatic run, but we would need to put a fair amount more work into the generator before that's feasible.

tools/xtce_generation/telemetry_generator.py
tech3371: good catch!

tech3371 (approved):

tech3371:
Regarding Greg's question, I think we should not auto run of xtce generator code. Some instruments needs manual changes such as creating ENUM based on what is recorded in description(That's how I did before). But I recently found out from Yaxi that they record all ENUM values in `States` table of packet definition excel sheet. If we can add that feature to the xtce generator code, it would be helpful. This will solve another manual steps. But there is still variable packet length still.

greglucas:
I didn't realize there were several manual steps, what you both say makes total sense to me then. One thing I'll say is try to add comments about what manual steps you are taking to the python file as well so someone else in the future knows what needs to be done if they do update it. (Maybe you already have this info, I just haven't been following the xtce udpates as closely)

maxinelasp:
>I didn't realize there were several manual steps, what you both say makes total sense to me then. One thing I'll say is try to add comments about what manual steps you are taking to the python file as well so someone else in the future knows what needs to be done if they do update it. (Maybe you already have this info, I just haven't been following the xtce udpates as closely)

Sure, I'm happy to do that. Which python file do you mean? The generator? I can add a comment at the top of the class to say what the generator _won't_ do. In this specific case, the only changes I made were in the combined class, but I can also add a note there to say it was manually created from the automatically generated files.

greglucas:
Maybe here in just the GLOWS section? https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/b0d52a544ed05178cf5062045c1951614e0e7066/tools/xtce_generation/xtce_generator_glows.py#L7C18-L7C18
It sounds like every instrument would probably need their own comments because the manual steps may be slightly different?

Just so that you have a list of steps you can follow if in a year you need to come back to this.

PR #245, opened 2023-10-06
MNT: Updating settings for codecov

greglucas:
# Change Summary

## Overview

- Run on dev branch so that we update the reference target for the base commit
- Change to 90% patch target
- Limit some pyproject.toml settings to per-file

## New Files
codecov.yml for codecoverage settings


.github/workflows/test.yml
```diff
@@ -3,7 +3,7 @@ name: Tests
 on:
```
maxinelasp, line 3: I have noticed codecov will count test files towards the total percentage of code coverage, which seems wrong. Is there some way to fix that? https://app.codecov.io/gh/IMAP-Science-Operations-Center/imap_processing?search=&displayType=list (you can see `codice/tests/test_bin_data.py` has a warning)

maxinelasp (approved):

tech3371 (commented):
look good! A question, is codecov test going to fail from now onwards? This PR test coverage was well above 90% but it still failed.

.github/workflows/test.yml
greglucas: That is actually the one place we want 100% code coverage IMO! Otherwise, we have tests that are never run is what that is telling us.

The reason for this is because there is an `__main__` block that doesn't get hit when not run from the command line. We should remove that / change it over to an actual test function.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/b0d52a544ed05178cf5062045c1951614e0e7066/imap_processing/codice/tests/test_bin_data.py#L42

greglucas:
> A question, is codecov test going to fail from now onwards? This PR test coverage was well above 90% but it still failed.

I think the default was expecting about 95 or maybe even 100 on the patches? This should update it to pass with anything above 90%. We should at least see how this goes and we can update numbers later if there is still noise.

tech3371 (approved):

greglucas:
Merging to help out subsequent PRs and try to get code coverage back to pointing at the latest commit on dev.

PR #246, opened 2023-10-10
I-ALiRT: L0 parsing

laspsandoval:
# Change Summary

## Overview
Decom I-ALiRT packet data and create xarray. 

## New Dependencies
None

## New Files
- decom_ialirt.py
   - Decom I-ALiRT packet data and create xarray. 

## Testing
- test_decom_ialirt.py
   - Tests that decom was successfully performed. 

## Note:
No test data was provided so far, so further tests will be added at a later date.


laspsandoval:
I intend to add more tests once I have test data. Just FYI.

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+from space_packet_parser import parser, xtcedef
+
+logging.basicConfig(level=logging.ERROR)
+logger = logging.getLogger(__name__)
+
+
+class PacketLength(Enum):
+    """Class that represents properties of the IALiRT packet."""
+
+    EXPECTED_LENGTH = 1464
```
greglucas: Do we need an Enum for this, or just call it `IALIRT_PACKET_LENGTH = 1464` as a global variable?

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+        List of all the unpacked data
+    """
+    packet_definition = xtcedef.XtcePacketDefinition(xtce_packet_definition)
+    packet_parser = parser.PacketParser(packet_definition)
+
+    packets = []
+
+    with open(packet_file) as file:
+        for line_number, line in enumerate(file, 1):
+            if not line.startswith("#"):
```
greglucas: I think this manipulation should be in the test files. We aren't going to actually get the raw text string during flight, we'll get CCSDS binary files at that point. So this isn't really necessary in the actual processing steps, but sort of a necessary requirement for you to test with currently.

I think that'd be something like reading in the current text file and then writing to a temporary binary file instead of making the bitstring. Then, pass that file to your reader here.

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+                    binary_values += bin_value_padded
+
+                # Check the length of binary_values
+                if len(binary_values) != PacketLength.EXPECTED_LENGTH.value:
+                    error_message = f"Error on line {line_number}: " \
+                                    f"Length of binary_values (" \
+                                    f"{len(binary_values)}) does not equal " \
+                                    f"{PacketLength.EXPECTED_LENGTH.value}."
+                    logger.error(error_message)
+                    raise ValueError(error_message)
```
greglucas: I don't think we need to log and raise? I think we can just raise the ValueError here and that will come out in the logs for free because it will fail.

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+    -------
+    xr.Dataset
+        A dataset containing the decoded data fields with 'time' as the coordinating
+        dimension.
+    """
+    try:
+        packets = decom_packets(packet_file, xtce)
+    except Exception as e:
+        logger.error(f"Error during packet decomposition: {str(e)}")
+        return
```
greglucas: I'd just let your previous ValueError bubble up through here and avoid the nested try/excepts

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if not packets:
+        logger.warning(f"No packets found in {packet_file}.")
```
greglucas: Good use of the warning here. I think you could actually test that this code path gets hit too by passing in an empty file?

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+        return
+
+    logger.info(f"Decomposed {len(packets)} packets from {packet_file}.")
```
greglucas:
```suggestion
    logger.info(f"Decommutated {len(packets)} packets from {packet_file}.")
```

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+
+    # List of instruments and their corresponding MET keys
+    instruments = ['SC', 'HIT', 'MAG', 'COD_LO', 'COD_HI', 'SWE', 'SWAPI']
+    instrument_coords = ['SC_SCLK_SEC', 'HIT_SC_TICK', 'MAG_ACQ', 'COD_LO_ACQ',
+                         'COD_HI_ACQ', 'SWE_ACQ_SEC', 'SWAPI_ACQ']
+
+    # Create a dictionary mapping each instrument to its time-dimension key
+    time_keys = dict(zip(instruments, instrument_coords))
```
greglucas: Do you want to just create the dictionary right away. Then you could get your other lists with `list(time_keys.keys())` or `values()` respectively if you need the entries.
```python
time_keys = {"SC": "SC_SCLK_SEC", "HIT": "HIT_SC_TICK", ...}
```

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+
+    # Initialize storage dictionary
+    data_storage = {inst: {} for inst in instruments}
```
greglucas:
```suggestion
    data_storage = {inst: collections.defaultdict(list) for inst in instruments}
```
Then you can just append directly to the array below I think.

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+
+    for packet in packets:
+        for key, value in packet.data.items():
+            for inst in instruments:
+                if key.startswith(inst):
+                    if key not in data_storage[inst]:
+                        data_storage[inst][key] = []
+                    data_storage[inst][key].append(value.derived_value)
+                    break
+            else:
```
greglucas: I don't think the break is needed? The `else` is currently with the for-loop, but seems to belong with the `if-block` indented once more.
```suggestion
                else:
```

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+                logger.warning(f"Unexpected key '{key}' found in packet data.")
+
+    logger.info("Generating datasets for each instrument.")
```
greglucas, line 81:
```suggestion
    logger.debug("Generating datasets for each instrument.")
```

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+                    bin_value = bin(int_value)[2:]
+
+                    # Make sure each binary string is 8 bits long
+                    bin_value_padded = bin_value.zfill(8)
+
+                    # Append the padded binary string to the final string
+                    binary_values += bin_value_padded
+
+                # Check the length of binary_values
+                if len(binary_values) != PacketLength.EXPECTED_LENGTH.value:
```
greglucas: Could add a few bytes to one of your test rows and create a new temporary file that you can then hit this code path with I think.

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,142 @@
...
+    # Generate xarray dataset for each instrument and spacecraft
+    datasets = {}
+    for inst in instruments:
+        dataset_dict = {key: (time_keys[inst], value)
+                        for key, value in data_storage[inst].items() if
+                        key != time_keys[inst]}
+        datasets[inst] = xr.Dataset(dataset_dict, coords={
+            time_keys[inst]: data_storage[inst][time_keys[inst]]})
+
+    logger.info(f"Generated datasets for {len(datasets)} instruments.")
```
greglucas: I think this will be the same regardless of how many you actually processed because you define the datasets[inst] based on all your initial instruments you expect. Suggest moving to debug or removing.

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,91 @@
...
+
+from imap_processing.ialirt.l0.decom_ialirt import decom_packets, generate_xarray
+
+
+@pytest.fixture(scope="session")
+def packet_path():
+    """Returns the raw packet directory."""
+    return Path(sys.modules[__name__.split(
+        '.')[0]].__file__).parent / 'ialirt' / 'tests' / 'test_data' / 'l0' / \
+        'IALiRT Raw Packet Telemetry.txt'
```
greglucas: Do we need to look at the module name, or can we just look at the file path itself? Also below.
```suggestion
    return Path(__file__).parent.parent / 'test_data' / 'l0' / \
        'IALiRT Raw Packet Telemetry.txt'
```
Or you could use the variable that comes with the package: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/65050d2f5905d78708f2deafa73c9d7e9f2ff970/imap_processing/__init__.py#L10

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,91 @@
...
+    """Data for decom_ultra"""
+    data_packet_list = generate_xarray(packet_path, xtce_ialirt_path)
+    return data_packet_list
+
+
+@pytest.fixture(scope="session")
+def decom_test_data(packet_path, xtce_ialirt_path):
+    """Read test data from file"""
+    data_packet_list = decom_packets(packet_path, xtce_ialirt_path)
+    return data_packet_list
```
greglucas: These two seem really similar to me and I didn't quite know the different due to the variable namings both being list at first glance.

I _think_ if you move the `decom_packets` code you wrote into a test function here instead, then you could use `imap_processing.decom.decom_packets()` to do your packet list creation inside of `generate_xarray()` and not need both of these fixtures.

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,91 @@
...
+        assert packet.data["SC_ATTITUDE"].derived_value == 'SUNSENSOR'
+        assert packet.data["SC_CATBEDHEATERFLAG"].derived_value == 'ON'
+        assert packet.data["SC_AUTONOMY"].derived_value == 'OPERATIONAL'
+        assert packet.data["HIT_STATUS"].derived_value == 'OFF-NOMINAL'
+        assert packet.data["SWE_NOM_FLAG"].derived_value == 'OFF-NOMINAL'
+        assert packet.data["SWE_OPS_FLAG"].derived_value == 'NON-HVSCI'
+
+
+def test_decom_instruments(decom_data, decom_test_data):
+    """This function checks that all instrument parameters are accounted for."""
```
greglucas, line 104: I'm not sure I fully follow all of this test. I think it'd help to break this down even further and actually assert on individual instruments, possibly parameterizing this over instruments if that would help.

But, I think it'd be good to say some things about each instrument individually and spell it out explicitly here. `assert len(decom_data["HIT"]["data"]) == expected HIT data size`

or, I don't think it would be unreasonable to also make an xarray dataset manually with `np.zeros(proper_shape)` and set the values you expect here so that you are asserting against a known zero array of the proper shape without having to decipher the loops within the test.

imap_processing/ialirt/packet_definitions/ialirt.xml
```diff
@@ -0,0 +1,805 @@
...
+			<!-- MAG -->
+			<xtce:Parameter name="MAG_ACQ" parameterTypeRef="uint48">
+				<xtce:LongDescription>MAG Acquisition Time</xtce:LongDescription>
+			</xtce:Parameter>
+			<xtce:Parameter name="MAG_STATUS" parameterTypeRef="uint24">
+				<xtce:LongDescription>MAG Status</xtce:LongDescription>
+			</xtce:Parameter>
+			<xtce:Parameter name="MAG_DATA" parameterTypeRef="uint24">
+				<xtce:LongDescription>MAG Data</xtce:LongDescription>
+			</xtce:Parameter>
```
greglucas, line 431: There are more bits in the MAG Acquisition Time than the data?

I think the data fields might need to be BinaryParameterType / BYTE so you can unpack the bits once you get more info later.

imap_processing/ialirt/packet_definitions/ialirt.xml
greglucas, line 1: Did you make this by hand, or did you concatenate all of the spreadsheets you had together to use the xtce creator tool? It'd be good to document how you made this and what reference files you used so we can remember in the future how this was made.

Really nice work gathering all of this info!

greglucas (commented):
Overall I think this looks good. One thing that would help is adding an example of what the xarray dataset should look like. Something like printing `ds.head()` and putting that into a comment in the code so that it is easy for someone not as familiar with this to quickly grasp how you expect things to look.

laspsandoval:
I am going to look at this once more on Monday. So please don't review again yet.

laspsandoval:
pre-commit.ci autofix

imap_processing/ialirt/packet_definitions/ialirt.xml
laspsandoval: This was from hand using I-ALiRT Packet Definitions: https://lasp.colorado.edu/galaxy/x/44nKCQ

greglucas: Nice, thanks for adding that info to the file too!

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -0,0 +1,99 @@
...
+        SC_HIT_STATUS (SC_SCLK_SEC) int64 1 0 1 0 1
+
+    This example shows a dataset with 'SC_SCLK_SEC' as the coordinate
+    and two data variables 'SC_MAG_STATUS' and 'SC_HIT_STATUS'.
+    """
+    try:
+        packets = decom_packets(packet_file, xtce)
+    except Exception as e:
+        logger.error(f"Error during packet decomposition: {e!s}")
+        return
```
greglucas:
```suggestion
    packets = decom_packets(packet_file, xtce)
```
I think we should actually fail here, so no need for a try/except.

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,197 @@
...
+        imap_module_directory
+        / "ialirt"
+        / "tests"
+        / "test_data"
+        / "l0"
+        / "IALiRT Raw Packet Telemetry.txt"
+    )
+
+    with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+        binary_file_path = Path(tmp_file.name)
```
greglucas: We want to delete the temporary file on clean-up, so we don't want this to be `delete=False`.

I think you can get what you are after here with the `tmp_path` fixture from pytest directly, and it will be cleaned up for you after the yield is finished. (meaning you don't need the unlink later either.

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    Creates an empty binary file, representative of a scenario
+    in which no packet data is available.
+    """
+    with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+        binary_file_path = Path(tmp_file.name)
+
+    yield binary_file_path
+
+    binary_file_path.unlink()
```
greglucas: I'd remove this entire block and put the relevant tmp_path usage inside the test below since you're only using it once it isn't really a fixture.
```suggestion
```

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,197 @@
...
+def test_length(decom_packets_data):
+    """Test if total packets in data file is correct"""
+    total_packets = 32
+    assert len(decom_packets_data) == total_packets
+
+
+def test_binary_value_length():
+    """
+    Validates the length of binary data converted
+    from a text file containing hexadecimal packet data.
```
greglucas: This function is sort of recreating what you're already doing in the fixture above. You could just add the assert up there if you're trying to test that file.

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,197 @@
...
+                assert len(binary_data) * 8 == IALIRT_PACKET_LENGTH
+
+
+def test_generate_xarray_empty_file(empty_binary_packet_path, xtce_ialirt_path, caplog):
+    """Test that an error is logged if an empty file is passed to generate_xarray."""
+
+    result = generate_xarray(empty_binary_packet_path, xtce_ialirt_path)
+
+    assert "Error during packet decomposition" in caplog.text
+    assert result is None
```
greglucas:
```suggestion
def test_generate_xarray_empty_file(empty_binary_packet_path, xtce_ialirt_path, caplog):
    """Test that an error is logged if an empty file is passed to generate_xarray."""
    with pytest.raises(ValueError):
        generate_xarray(empty_binary_packet_path, xtce_ialirt_path)
```
If you raise instead of log, we can assert that this does raise as expected here like this.

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,197 @@
...
+    # Assert that the length of parameters is the same
+    # between packet data and generated xarray
+    assert sum(instrument_lengths.values()) == len(
+        list(decom_packets_data[0].data.keys())
+    )
+
+    # Check that all the dimensions are the correct length
+    # example:len(xarray_data['HIT'].coords['HIT_SC_TICK'].values) == 32
+    for instrument in instruments:
+        dimension_name = next(iter(xarray_data[instrument].dims))
```
greglucas: Do you need to get all dimension names? Right now I think you're just getting the first one.
```suggestion
        for dimension_name in xarray_data[instrument].dims:
```

imap_processing/ialirt/packet_definitions/ialirt.xml
laspsandoval: I'm going off of the spreadsheet being tracked on IMAPCATS-4:
Screen Shot 2023-11-07 at 3 50 27 PM

greglucas: Interesting! That just seems odd to me, but you're definitely right. One thing I did just notice though is that it looks like the I-ALiRT spreadsheets have MET coming down, which is 32 bits instead of ACQ_TM which is 48 bits. Is there another mismatch there that we need to update in the I-ALiRT spreadsheet as well?

@maxinelasp did you have any conversations with MAG about their I-ALiRT data formatting?

imap_processing/ialirt/tests/unit/test_decom_ialirt.py
```diff
@@ -0,0 +1,157 @@
...
+                # Discard the first value since it is only a counter
+                hex_values = line.strip().split(";")[1:]
+                # Convert hex to binary
+                binary_data = bytearray.fromhex("".join(hex_values))
+                binary_file.write(binary_data)
+                assert len(binary_data) * 8 == IALIRT_PACKET_LENGTH
+
+    yield binary_file_path
+
+    return binary_file_path
```
greglucas, line 47: Do we need both of these? I'm not sure which one would be needed to make sure the `tmp_path` doesn't get cleaned up too soon.

greglucas (approved):
Nice! Looks great.

imap_processing/ialirt/packet_definitions/ialirt.xml
laspsandoval: Yes. I had to piece together different spreadsheets to get the xml:
1. I-ALiRT Packet Definitions 2022_10_25.xlsx (old)
2. IALiRT SC Packet Content.xlsx (recent)
3. And then what was in IMAPCATS

So you are correct. The definitions from I-ALiRT Packet Definitions 2022_10_25.xlsx need to be updated.

PR #253, opened 2023-10-11
SWE L1B implementation and updates to L1A

### LINKED ISSUE #98: [L5] Produce L1A CDF formatted data product for SWE
### Summary of the L5 requirement

The SDC needs to produce L1A CDF formatted data products for the SWE instrument. The data products for L1A can be found in the IMAP-Lo algorithm document here: https://lasp.colorado.edu/galaxy/display/IMAP/IMAP+Algorithm+Document+from+Instrument+Teams

### Parent requirement

SOC-SDC-L4-29 | SDC Execution of L1 Processing Software | The SDC shall be capable of executing L1 processing software to produce L1 data products within seven (7) days of receipt of all requisite data inputs. | Requirement | The SDC produces L1 data products via Instrument Team supplied algorithms and SDC produced processing code. | SOC-L3-57 | Test | SIT-3: L0 -> L1 processing | SIT-3
-- | -- | -- | -- | -- | -- | -- | -- | --

### Tasks

- [x] Create Galaxy Page for L1A algorithms, https://lasp.colorado.edu/galaxy/display/IMAP/SWE+L1A. Discuss it with instrument team.

Produce L1A data for each of these packets:
- [x] #103
- [x] #102
- [x] #101
- [x] #100

### LINKED ISSUE #101: Create L1A CDF file for SWE
### Summary
This ticket creates L1A cdf file for all appIds for SWE instrument.

### Requirements
Packet Names:
SWE_APP_HK
SWE_CEM_RAW
SWE_SCIENCE
SWE_EVTMSG

### LINKED ISSUE #111: [L5] Produce L1B CDF formatted data product for SWE
### Summary of the L5 requirement

The SDC needs to produce L1B CDF formatted data products for the SWE instrument. The data products for L1B can be found in the IMAP-Lo algorithm document here: https://lasp.colorado.edu/galaxy/display/IMAP/IMAP+Algorithm+Document+from+Instrument+Teams

### Parent requirement

SOC-SDC-L4-29 | SDC Execution of L1 Processing Software | The SDC shall be capable of executing L1 processing software to produce L1 data products within seven (7) days of receipt of all requisite data inputs. | Requirement | The SDC produces L1 data products via Instrument Team supplied algorithms and SDC produced processing code. | SOC-L3-57 | Test | SIT-3: L0 -> L1 processing | SIT-3
-- | -- | -- | -- | -- | -- | -- | -- | --

### Tasks

- [ ] Create Galaxy Page for L1A algorithms, https://lasp.colorado.edu/galaxy/display/IMAP/SWE+L1B. Discuss it with instrument team.
- [x] #115

tech3371:
# Change Summary
Implements SWE L1B algorithm and updates L1A
## Overview
These are the changes in this PR    
    - added logics to populate data based on ESA lookup table number
    - refactored code to store l1a data as 180x7 instead of 15x12x7 array.
    - added new event test data and moved tests data into folders.
    - updated l1a code to be ISTP compliant and to be able to create cdf file
    - added new test data and update test as needed
    - implemented SWE l1b algorithm for both science and other data types
    - creates cdf file for SWE l1b data product
    - added deadtime correction and place holder for in-flight calibration. converted counts to rates
    - added rest of l1b algorithm to convert raw data to engineering unit.

## New Files
- imap_processing/swe/l1b/
   - Added code files to process data to l1b
- imap_processing/swe/tests/
   - Added new tests for l1b and new l1a test data

## Deleted Files

- deleted file 1

## Updated Files
- imap_processing/swe/l1a/
   - updated l1a code to be able to create CDF file

## Testing
- imap_processing/swe/tests/
   - Added new tests for l1b and new l1a and l1b test data


imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -1,10 +1,90 @@
...
+
+    Returns
+    -------
+    xr.dataset
+        dataset with all metadata field data in xr.DataArray
+    """
+    metadata_arrays = collections.defaultdict(list)
+
+    for data_packet in packets:
+        add_metadata_to_array(data_packet, metadata_arrays)
```
greglucas: mutating the arguments is potentially ambiguous here. I'd return the metadata_arrays back out of the function to avoid relying on the implicit mutation of the inputs.

```python
metadata_arrays = add_metadata_to_array(data_packet, metadata_arrays)
```

imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -13,15 +93,23 @@ def swe_l1a(packet_file: str):
...
-    logging.info(f"Unpacking data from {packet_file}")
-
-    # If appId is science, then the file should contain all data of science appId
-    if decom_data[0].header["PKT_APID"].raw_value == 1344:
-        logging.info("Processing science data")
-        return swe_science(decom_data=decom_data)
-    else:
-        return decom_data
+    # group data by appId
+    grouped_data = group_by_apid(packets)
```
greglucas, line 35: I think you should be able to test this with the current data you've got? Just assert that the group is in sorted order or something else quick.

imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -13,15 +93,23 @@ def swe_l1a(packet_file: str):
...
+
+    for apid in grouped_data.keys():
+        # If appId is science, then the file should contain all data of science appId
+        if apid == SWEAPID.SWE_SCIENCE.value:
+            # sort data by acquisition time
+            sorted_packets = sorted(
+                grouped_data[apid], key=lambda x: x.data["ACQ_START_COARSE"].raw_value
+            )
+            logging.info("Processing science data")
```
greglucas:
```suggestion
            logging.debug("Processing science data for [%s] packets", len(sorted_packets))
```
Add some more information to the log to help debug later? Also make it a debug log since we don't need it every time and likely only when we are trying to investigate something.

imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -13,15 +93,23 @@ def swe_l1a(packet_file: str):
...
+            swe_science(decom_data=sorted_packets)
+            # TODO: write data to CDF.
+        else:
+            # If it's not science, we unpack, organize and save it as a dataset.
+            create_dataset(packets=grouped_data[apid])
```
greglucas: Do you need to access the returned dataset? ds = create_dataset(...)

imap_processing/swe/l1a/swe_science.py
```diff
@@ -133,18 +133,18 @@ def swe_science(decom_data):
         # where 1260 = 15 seconds x 12 energy steps x 7 CEMs
         # Take the "raw_counts" indices/counts mapping from
         # decompression_table and then reshape the return
-        uncompress_data = np.take(decompression_table, raw_counts).reshape(15, 12, 7)
-        raw_counts = raw_counts.reshape(15, 12, 7)
+        uncompress_data = np.take(decompression_table, raw_counts).reshape(180, 7)
+        raw_counts = raw_counts.reshape(180, 7)
```
greglucas: Update the comment above to indicate this is 180 now? Leave the 15 x 12 information though since that still seems useful.

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -0,0 +1,74 @@
...
+        # table and then reverse them because the np.polyval is
+        # expecting coefficient in decending order
+        coeff_values = metadata_list.loc[
+            metadata_list["mnemonic"] == metadata_field
+        ].values[0][6:][::-1]
+
+        # Convert the raw value to engineering unit
+        dataset[metadata_field].data = np.polyval(
+            coeff_values, dataset[metadata_field].data
+        )
```
greglucas: Use the `Polynomial` class instead because polyval is discouraged. It looks like it is also in increasing order for the coefficients.

https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.html#numpy.polynomial.polynomial.Polynomial

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # Look up all metadata fields for the packet name
+    metadata_list = eu_conversion_table.loc[
+        eu_conversion_table["packetName"] == packet_name.name
+    ]
+
+    # for each metadata field, convert raw value to engineering unit
+    for field in metadata_list.index:
+        metadata_field = metadata_list.loc[field]["mnemonic"]
+        # On this line, we are getting the coefficients from the
```
greglucas: Great comments here. I'd even add more like:

Data from the tables are the coefficients of polynomials looking like: p0 + p1*x + p2*x**2 which are provided by the instrument teams in calibration files.

imap_processing/swe/tests/test_swe_l1a.py
```diff
@@ -0,0 +1,47 @@
...
+def test_total_packets(decom_test_data):
+    assert len(decom_test_data) == 23
+
+
+def test_group_by_apid(decom_test_data):
+    grouped_data = group_by_apid(decom_test_data)
+
+    # check total dataset for swe science
+    total_science_data = grouped_data[SWEAPID.SWE_SCIENCE.value]
+    # print(total_science_data)
```
greglucas:
```suggestion
```

imap_processing/swe/tests/test_swe_l1a_event_msg.py
```diff
@@ -0,0 +1,49 @@
...
+    """Read test data from file"""
+    test_folder_path = "swe/tests/l0_data"
+    event_file_list = list(
+        imap_module_directory.glob(f"{test_folder_path}/*SWE_EVTMSG_packet.bin")
+    )
+    event_data = []
+    for packet_file in event_file_list:
+        event_data.append(decom_swe.decom_packets(packet_file)[0])
+
+    return event_data
```
greglucas, line 19: Is this comment accurate? I wasn't entirely clear why you were doing this in just code at first.
```suggestion
    # Return a list of the first packet from every file
    return [decom_swe.decom_packets(packet_file)[0] for packet_file in event_file_list]
```

imap_processing/swe/tests/test_swe_l1a_science.py
```diff
@@ -18,10 +18,10 @@ def decom_test_data():
     # Packet 4 has spin 3's data
     # reorder to match the order of spin in the data
```
greglucas, line 19: Do we need to be able to account for this re-ordering in code and make sure we can handle this ourselves? Or this is a one-off thing due to the way they ran their simulator?

imap_processing/swe/tests/test_swe_l1b.py
```diff
@@ -0,0 +1,92 @@
...
+
+    science_eu_field_list = [
+        "SPIN_PHASE",
+        "SPIN_PERIOD",
+        "THRESHOLD_DAC",
+    ]
+
+    # Test EU values for science data
+    for field in science_eu_field_list:
+        assert round(science_l1b[field].data[1], 5) == round(validation_data[field], 5)
```
greglucas, line 109: Would we be able to use numpys testing routines for this at all instead of rounding things yourself, and possibly removing the loop too then.
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose

imap_processing/swe/tests/test_swe_l1b_science.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_swe_l1b(l1a_test_data):
+    swe_l1b_science(l1a_test_data)
+
+
+def test_get_full_cycle_data_indices():
+    q = np.array([0, 1, 2, 0, 1, 2, 3, 2, 3, 0, 2, 3, 0, 1, 2, 3, 2, 3, 1, 0])
+    filtered_q = get_indices_of_full_cycles(q)
+    assert np.array_equal(filtered_q, np.array([3, 4, 5, 6, 12, 13, 14, 15]))
```
greglucas: Here you could use `np.testing.assert_array_equal()` instead of needing to assert yourself here.

imap_processing/swe/tests/test_swe_l1a_event_msg.py
tech3371: no. I got event data in multiple files. And each file contains one event data only. In this for loop, I am going through all event data file and reading those one data and return as one for testing.

imap_processing/swe/tests/test_swe_l1a_science.py
tech3371: This is a one-off thing

tech3371:
pre-commit.ci autofix

imap_processing/swe/l1a/swe_science.py
```diff
@@ -130,50 +118,80 @@ def swe_science(decom_data):
...
+    epoch_time = xr.DataArray(
         metadata_arrays["SHCOARSE"],
-        name="met_time",
-        dims=["met_time"],
-        attrs=dict(
-            description="Mission elapsed time",
-            units="seconds since start of the mission",
-        ),
+        name="Epoch",
+        dims=["Epoch"],
```
GFMoraga, line 138: Should this be "EPOCH" in caps?

imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -13,15 +21,51 @@ def swe_l1a(packet_file: str):
...
-        logging.info("Processing science data")
-        return swe_science(decom_data=decom_data)
-    else:
-        return decom_data
+    # group data by appId
+    grouped_data = group_by_apid(packets)
+
+    for apid in grouped_data.keys():
+        # If appId is science, then the file should contain all data of science appId
+        if apid == SWEAPID.SWE_SCIENCE.value:
```
greglucas:
```suggestion
        if apid == SWEAPID.SWE_SCIENCE:
```
since you're using an IntEnum I don't think you need the value here and below.

imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -1,10 +1,18 @@
...
-from imap_processing.swe.l0 import decom_swe
+from imap_processing.swe import __version__
 from imap_processing.swe.l1a.swe_science import swe_science
+from imap_processing.swe.utils.swe_utils import (
+    SWEAPID,
+    create_dataset,
+    get_descriptor,
+)
+from imap_processing.utils import group_by_apid, sort_by_time
+from imap_processing.write_to_cdf import write_to_cdf
```
greglucas: Can you change this to import from `imap_processing.cdf_utils` now that it is in? Then you could remove this portion and avoid the duplication now instead of having to fix it later.

GFMoraga (approved):
I would maybe double check the `dims` and make sure your time variables don't have to be CAPS as well.  You are most likely doing the correct format. I was just wondering about if the time needs to be CAPS if SHCOARSE is?
 This is an example:
`epoch_time = xr.DataArray(
        metadata_arrays["SHCOARSE"],
        name="Epoch",
        dims=["Epoch"],
        attrs=cdf_utils.epoch_attrs,
    )`

imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -13,15 +21,51 @@ def swe_l1a(packet_file: str):
...
+            )
+        else:
+            return write_to_cdf(
+                data,
+                "swe",
+                "l1a",
+                version=__version__,
+                description=get_descriptor(apid),
+                directory=current_dir,
+            )
```
greglucas: Put the only difference into an if statement and return after to shorten the implementation?

```suggestion
      return write_to_cdf(
          data,
          "swe",
          "l1a",
          version=__version__,
          mode=f"{data['APP_MODE'].data[0]}" if apid == SWEAPID.SWE_APP_HK else "",
          description=get_descriptor(apid),
          directory=current_dir,
      )
```

imap_processing/swe/l1a/swe_science.py
tech3371: I had the same question at one point :) After looking at Bryan's code and ISTP, it seems like ISTP likes it `Epoch`.

imap_processing/swe/l1a/swe_l1a.py
tech3371: I tried to do that. It seems to gives me back this error. Do you know way around it?
```
E           AttributeError: 'NoneType' object has no attribute 'setdefault'
```

imap_processing/swe/l1a/swe_l1a.py
tech3371: I think there is test for this function. Did you like me to add specific test in there?

imap_processing/swe/l1a/swe_l1a.py
greglucas: I think you need to return the `metadata_arrays` out of the function.

imap_processing/swe/tests/test_swe_l1b.py
tech3371: btw, I looked into this few days ago. I couldn't get it to work without complicating logics to remove the loop. The reason was that, I only check float value matches for only these three variables. If we remember correctly, I had to look up these values from both `science_l1b` and `validation_data`, then match them using `np`. It was looking unreadable and since it's just these three variables, I thought for loop might be ok.
```
science_eu_field_list = [
        "SPIN_PHASE",
        "SPIN_PERIOD",
        "THRESHOLD_DAC",
    ]
```

imap_processing/swe/tests/test_swe_l1b.py
```diff
@@ -0,0 +1,167 @@
...
+    ]
+    hk_l1b_filepath = swe_l1b(hk_data[0])
+
+    # process evtmsg data to l1b
+    evtmsg_data = [
+        ds for ds in cdf_data if ds["PKT_APID"].data[0] == SWEAPID.SWE_EVTMSG.value
+    ]
+    evtmsg_l1b_filepath = swe_l1b(evtmsg_data[0])
+
+    assert os.path.basename(sci_l1b_filepath) == "imap_swe_l1b_20230927_sci_v01.cdf"
```
greglucas: Are these `Path` objects? if so, you can just get the name. If not, maybe you can make the return be a path object instead of a string path?
```suggestion
    assert sci_l1b_filepath.name == "imap_swe_l1b_20230927_sci_v01.cdf"
```

imap_processing/swe/l1a/swe_l1a.py
greglucas: I think this comment was before you added the test and codecov was failing. Resolving this now, nothing needed here.

imap_processing/swe/l1a/swe_l1a.py
tech3371: interesting. I didn't know that.

imap_processing/swe/tests/test_swe_l1b.py
tech3371: It looks like it's not `Path` object but a string.

imap_processing/swe/l1a/swe_l1a.py
tech3371: I started working on it but it introduced more issues. I will revisit this on Monday.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,485 @@
...
+    Deadtime correction involves mathematically adjusting the measured count
+    rates to compensate for this deadtime effect. This correction is crucial
+    when dealing with high-intensity sources or particle fluxes, as the deadtime
+    can significantly affect the accuracy of the measurements.
+
+    Deadtime correction is important to ensure accurate measurements and data
+    analysis in fields where event detection rates are high and where every
+    detected event is critical for understanding physical processes.
+
+
```
bourque: Some extra lines here

imap_processing/swe/l1a/swe_science.py
```diff
@@ -130,50 +118,81 @@ def swe_science(decom_data):
...
+        attrs=cdf_utils.epoch_attrs,
+    )
+
+    # TODO: refactor this better once we develop class inheritance
+    int_attrs = swe_cdf_attrs.int_attrs
+    int_attrs["CATDESC"] = int_attrs["FIELDNAM"] = int_attrs["LABLAXIS"] = "Energy"
+    int_attrs["VALIDMAX"] = np.int64(180)
```
bryan-harter: I don't think it hurts anything to have VALIDMIN and VALIDMAX here, but they aren't actually required if these numbers don't change over time

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,485 @@
...
+    the recorded count rate appears to be lower than the actual count rate.
+
+    Deadtime correction involves mathematically adjusting the measured count
+    rates to compensate for this deadtime effect. This correction is crucial
+    when dealing with high-intensity sources or particle fluxes, as the deadtime
+    can significantly affect the accuracy of the measurements.
+
+    Deadtime correction is important to ensure accurate measurements and data
+    analysis in fields where event detection rates are high and where every
+    detected event is critical for understanding physical processes.
```
bourque: I like how you included this detailed description right in the code!

imap_processing/swe/l1a/swe_science.py
```diff
@@ -130,50 +118,81 @@ def swe_science(decom_data):
...
-            units="seconds since start of the mission",
-        ),
+        name="Epoch",
+        dims=["Epoch"],
+        attrs=cdf_utils.epoch_attrs,
+    )
+
+    # TODO: refactor this better once we develop class inheritance
+    int_attrs = swe_cdf_attrs.int_attrs
+    int_attrs["CATDESC"] = int_attrs["FIELDNAM"] = int_attrs["LABLAXIS"] = "Energy"
```
bryan-harter: SPDF is definitely gonna want something a bit more descriptive, roughly 80 characters for CATDESC, 30 for FIELDNAM.  I think LABLAXIS is fine with just "energy".

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,485 @@
...
+
+
+
+    Parameters
+    ----------
+    counts : np.array
+        counts data before deadtime corrections
+    tsampl : int
```
bourque: This parameter name is a bit confusing to me. Is it supposed to represent "time sample"?. Perhaps call it `time_sample`? or `acq_duration`? or just `duration`?

imap_processing/swe/l1a/swe_science.py
```diff
@@ -130,50 +118,81 @@ def swe_science(decom_data):
...
     dataset["RAW_SCIENCE_DATA"] = raw_science_xarray
 
     # create xarray dataset for each metadata field
     for key, value in metadata_arrays.items():
         if key == "SHCOARSE":
             continue
+        int_attrs["CATDESC"] = int_attrs["FIELDNAM"] = int_attrs["LABLAXIS"] = key
+        # get int32's max since most of metadata is under 32-bits
+        int_attrs["VALIDMAX"] = np.iinfo(np.int32).max
+        int_attrs["DEPEND_O"] = "Epoch"
```
bryan-harter: Should be a DEPEND_0 (zero), not DEPEND_O (the letter)

imap_processing/swe/swe_cdf_attrs.py
```diff
@@ -0,0 +1,227 @@
...
+# number, int and float.
+# We can create one attrs dict for both
+# with their default values.
+
+# For SWE, max range for uncompressed counts is
+# 65536. Therefore, I initialized it with that
+# value.
+int_attrs = {
+    "CATDESC": None,
+    "DISPLAY_TYPE": "spectrogram",
```
bryan-harter: The only variables that should have a DISPLAY_TYPE are the data variables.  Support_data doesn't need it, and SPDF might comment on it.

imap_processing/utils.py
```diff
@@ -0,0 +1,94 @@
+import collections
```
bourque, line 2: It would be good to add a module docstring here, especially since it will likely be used by the whole team

bourque (approved):
Awesome work here @tech3371! I just had a few comments, no blockers. Thanks for writing all of this; it will also be very helpful for CoDICE decom and testing.

imap_processing/swe/l1a/swe_science.py
tech3371: Totally agree. I plan to add more descriptive description in future PR. This PR already had lots of changes. I will create issue for that upcoming PR though.

imap_processing/swe/l1a/swe_science.py
tech3371: oh ok. good to know

imap_processing/swe/l1a/swe_science.py
tech3371: good catch!

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: I like `acq_duration`

imap_processing/swe/swe_cdf_attrs.py
tech3371: Good to know. This will probably go away when I rebase with Maxine's changes.

imap_processing/utils.py
tech3371: Sure thing

imap_processing/swe/l1a/swe_l1a.py
tech3371: I think I worked out most of issue that day and forgot to update here(because I was sick) but I will double check when I rebase with new changes from Maxine's.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+
+def read_lookup_table(table_index_value: int):
+    """Read lookup table from file.
+
+    Parameters
+    ----------
+    table_index_value : int
+        ESA table index number
+    """
+    lookup_table_filepath = f"{imap_module_directory}/swe/l1b/swe_esa_lookup_table.csv"
```
greglucas: Not critical, but you could keep it as a Path object here instead of string.
```suggestion
    lookup_table_filepath = imap_module_directory / "swe/l1b/swe_esa_lookup_table.csv"
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+    lookup_table = pd.read_csv(
+        lookup_table_filepath,
+        index_col="e_step",
+    )
+
+    if table_index_value == 0:
+        return lookup_table.loc[lookup_table["table_index"] == 0]
+
+    if table_index_value == 1:
```
```suggestion
        return lookup_table.loc[lookup_table["table_index"] == 0]
    elif table_index_value == 1:
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+        return lookup_table.loc[lookup_table["table_index"] == 1]
+
+    raise ValueError("Error: Invalid table index value")
+
+
+def deadtime_correction(counts: np.array, acq_duration: int):
+    """Calculate deadtime correction.
+
+    Deadtime correction is a technique used in various fields, including
+    nuclear physics,radiation detection, and particle counting, to compensate
```
```suggestion
    nuclear physics, radiation detection, and particle counting, to compensate
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+        counts data
+    acq_duration : int
+        Acquisition duration. acq_duration is in millieseconds
+
+    Returns
+    -------
+    np.array
+        Count rates array in seconds
+    """
+    return np.divide(data, acq_duration)
```
greglucas: Out of curiosity is there a reason to use `np.divide(a, b)` instead of `a / b`?

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+        # 720 came from 24 x 30. full_cycle_data array has (24, 30)
+        # dimension.
+        esa_step_number = 0
+        # column_index goes from 0 to 29 range where
+        # 30 came from 30 column in full_cycle_data array
+        column_index = -1
+
+        # Go through four quarter cycle data packets
+        for index in range(4):
+            uncompressed_counts = l1a_data["SCIENCE_DATA"].data[packet_index + index]
```
greglucas: I think we decided on decompressed for consistency.
```suggestion
            decompressed_counts = l1a_data["SCIENCE_DATA"].data[packet_index + index]
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+        # esa_step_number goes from 0 to 719 range where
+        # 720 came from 24 x 30. full_cycle_data array has (24, 30)
+        # dimension.
+        esa_step_number = 0
+        # column_index goes from 0 to 29 range where
+        # 30 came from 30 column in full_cycle_data array
+        column_index = -1
+
+        # Go through four quarter cycle data packets
+        for index in range(4):
```
greglucas, line 224: If it turns out this is a bottleneck in the future (no need to update now), these nested for-loops could possibly be refactored out into fancy indexing. I'm sure it wouldn't be super easy to do though either.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+    xr.dataset
+        L1A dataset with filtered metadata.
+    """
+    for key, value in l1a_data.items():
+        l1a_data[key] = value.data[full_cycle_data_indices]
+    return l1a_data
+
+
+def swe_l1b_science(l1a_data):
+    """In this function, we cap.
```
greglucas: Looks like the description wasn't finished.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+    l1a_data : xr.dataset
+        L1A dataset
+
+    Returns
+    -------
+    xr.dataset
+        L1A dataset with filtered metadata.
+    """
+    for key, value in l1a_data.items():
+        l1a_data[key] = value.data[full_cycle_data_indices]
```
greglucas, line 332: I think you may be able to avoid needing to do any copying in the other function by returning a new dataset here?

You may also be able to index into the entire dataset at once instead of needing to do a for-loop over the individual variables.
```python
return ds.loc[{"dim0": full_cycle_data_indices}]
```
https://docs.xarray.dev/en/latest/user-guide/indexing.html#dataset-indexing

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+    energy = xr.DataArray(
+        np.arange(24),
+        name="Energy",
+        dims=["Energy"],
+        attrs=dataclasses.replace(
+            swe_cdf_attrs.int_base,
+            catdesc="Energy",
+            fieldname="Energy",
+            label_axis="Energy",
+            units="keV",
```
greglucas: Is this the actual energy, or is it the integer index of the energy bin? You put `arange()` in the data field, which makes me think it may be an index and therefore unitless? Same for angle below.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -0,0 +1,509 @@
...
+    rates = xr.DataArray(
+        np.arange(7, dtype=np.float64),
+        name="Rates",
+        dims=["Rates"],
+        attrs=dataclasses.replace(
+            swe_cdf_attrs.float_base,
+            catdesc="Rates",
+            fieldname="Rates",
+            label_axis="Rates",
+            units="float",
```
```suggestion
            units="counts/s",
```

imap_processing/swe/tests/l1a_test_data/imap_swe_l1a_20230927_lveng_hk_v01.cdf
greglucas, line 1: These cdf files shouldn't be committed to the repo I don't think? It seems like they are leaking out here and we might need to put these in a temporary data test fixture to automatically get cleaned up.

Edit: I just read farther down and I think these are actually the l1a files that go into l1b? I'm hesitant to start committing all intermediate files because I feel like these are going to change and the repository and installed package size will really grow fast... At the same time, you want to be able to run the l1b tests without requiring l1a to run first.

Thoughts on this?

imap_processing/swe/tests/test_swe_l1a.py
```diff
@@ -0,0 +1,55 @@
...
+    # check total dataset for event message
+    total_event_message_data = grouped_data[SWEAPID.SWE_EVTMSG]
+    assert len(total_event_message_data) == 15
+
+
+def test_cdf_creation(decom_test_data):
+    grouped_data = group_by_apid(decom_test_data)
+    sci_cdf_filepath = swe_l1a(grouped_data[SWEAPID.SWE_SCIENCE])
+    hk_cdf_filepath = swe_l1a(grouped_data[SWEAPID.SWE_APP_HK])
+    assert os.path.basename(sci_cdf_filepath) == "imap_swe_l1a_sci_20230927_v01.cdf"
```
greglucas, line 56: If you return a Path instance from the functions you can use name in the comparison
```suggestion
    assert sci_cdf_filepath.name == "imap_swe_l1a_sci_20230927_v01.cdf"
```

imap_processing/swe/tests/test_swe_l1a.py
```diff
@@ -0,0 +1,55 @@
...
+    # check total dataset for housekeeping
+    grouped_data[SWEAPID.SWE_APP_HK]
+    assert len(total_cem_raw_data) == 2
+
+    # check total dataset for event message
+    total_event_message_data = grouped_data[SWEAPID.SWE_EVTMSG]
+    assert len(total_event_message_data) == 15
+
+
+def test_cdf_creation(decom_test_data):
```
greglucas: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html
```suggestion
def test_cdf_creation(decom_test_data, tmp_path):
```
Then maybe you can call `write_cdf()` with this temporary directory?

imap_processing/utils.py
```diff
@@ -0,0 +1,108 @@
...
+1. sort_by_time function
+    Sorts data packets list by time. It takes packets list
+    and time key as input. Then sorts packets by time key input
+2. group_by_apid function
+    Group packets by apid. It takes packets list as input.
+3. convert_raw_to_eu function
+    Convert raw data to engineering unit. It read conversion
+    table from the given input path and reads current
+    packet name's metadata fields. Then it converts raw data
+    to engineering unit.
```
greglucas: The functions are listed below with docstrings explaining in more detail what they do so I'd remove the descriptions from the top-level here. I think these should automatically get picked up by the documentation and then we won't get out of sync and need to put information in multiple places.
```suggestion
Common functions that every instrument can use.
```

greglucas (commented):
Don't be afraid of `Path` objects :) There is a lot of `f"{imap_processing}/path/to/file"` that could be `imap_processing / "path/to/file"` so you stay in the pathlib classes.

The dataset copy / indexing is something worth looking into IMO. I think we should try to return a new view into the subset dataset instead of overwriting the current one if possible.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: no particular reason. I feel like I heard in the past that using np.divide is suggested.

greglucas: The only reason I'd possibly suggest avoiding it is because of some of the deferral mechanisms for other array libraries. i.e. if you do xarray.Dataset / np.ma.MaskedArray, using `np.divide()` can force those to numpy-arrays rather than letting the dataset and masked array choose whether to return a Dataset with metadata still attached instead. No concern with what you have here, but I would actually suggest the opposite in a more general sense.

tech3371: That's good to know. ty!

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: I changed it to this. Does that look ok?
```
energy = xr.DataArray(
        np.arange(180),
        name="Energy",
        dims=["Energy"],
        attrs=dataclasses.replace(
            swe_cdf_attrs.int_base,
            catdesc="Energy's index value in the lookup table",
            fieldname="Energy Bins",
            label_axis="Energy Bins",
            units="int",
        ).output(),
    )
```

imap_processing/swe/tests/l1a_test_data/imap_swe_l1a_20230927_lveng_hk_v01.cdf
tech3371: I agree. I was having doubts about that. cool. I will remove it. At least, I got practice on how to read cds files.

imap_processing/swe/tests/test_swe_l1a.py
tech3371: I will try that.

tech3371: changed to use tmp_path

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: I have some questions about that. Can I revisit this after thanksgiving break? I like to merge in this PR as soon as possible so that others can use common utils from this PR. I appreciate all the suggestions. I learned a lot in this PR!

greglucas: Yes, that is no problem to address it later.

imap_processing/swe/l1b/swe_l1b_science.py
greglucas: I don't think "int" is right, it should probably be unitless if that is allowed. Just an empty string maybe.

greglucas: Alternatively, could you actually pass in `decommutation_table` instead of `np.arange(180)` so you got the actual energy values here?

greglucas (approved):
Because you added and removed files, please squash-merge when merging this in.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: I will leave empty string for now because those values in `decommutation_table` could be different based on look table id. I will keep this mind as we go forward.

PR #264, opened 2023-10-12
Glows L0 changes to data class to accomodate L1A needs

maxinelasp:
# Change Summary
Updated a few things in the GLOWS L0 code in the process of L1A processing

## Overview

- Added a CCSDS dataclass 
- Updated glows dataclasses to explicitly call out the attributes 
- added a few additional attributes in L0

## New Dependencies

None
## New Files

- ccsds/ccsds_data.py
   - dataclass describing common CCSDS headers

## Updated Files

- glows_l0_data
   - Added attributes for each class explicitly
   - Changed the way the data gets inserted into the data class
- glows_decom
   - knock-on changes from updating the data class

## Testing

I moved the histogram processing into a different dataclass to make L0 truly the raw packet data. This means the tests have shrunk in scope, with the deleted functionality getting added into the next L1A PR.

## Related Tickets:
- #175


maxinelasp:
As a bonus, this will also fix all the code coverage issues because I added more random lines of code in my dataclasses and I get the pretty-print function for free with `@dataclass`! ... as a whole though, this PR does reduce the testing of the packet data.

imap_processing/ccsds/ccsds_data.py
```diff
@@ -0,0 +1,47 @@
...
+        attributes = [field.name for field in fields(self)]
+
+        for key, item in packet_header.items():
+            value = (
+                item.derived_value if item.derived_value is not None else item.raw_value
+            )
+            if key in attributes:
+                setattr(self, key, value)
+            else:
+                raise KeyError(
```
greglucas, line 44: Could pass in a bad header key in your test?

```python
def test_bad_header():
    with pytest.raises(KeyError, match="Did not find matching"):
        CcsdsData({"bad_data": Mock()})
```

imap_processing/glows/l0/decom_glows.py
```diff
@@ -50,6 +50,11 @@ def decom_packets(
     histdata = []
     dedata = []
 
+    filename = packet_file_path.split("/")[-1]
```
greglucas: This will fail our Windows friends. Use pathlib?
```suggestion
    filename = Path(packet_file_path).name
```

imap_processing/glows/l0/decom_glows.py
```diff
@@ -50,6 +50,11 @@ def decom_packets(
...
+
+    # TODO: get this version from an offical source
```
greglucas: Do you want to store this in `glows/__init__.py` so you can `from imap_processing.glows import version` and get it globally for all the subsequent files and not have to pass it around in your function calls?

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -58,63 +79,70 @@ class HistogramL0:
...
-            raise ValueError(
-                f"Histogram packet is lacking bins. Expected a count of 3599, "
-                f"actually received {len(histograms)}"
-            )
+        super().__init__(software_version, packet_file_name, CcsdsData(packet.header))
 
-        return histograms
+        attributes = [field.name for field in fields(self)]
 
+        # For each item in packet, assign it to the matching attribute in the class.
```
greglucas, line 132: Would it make sense to subclass your CcsdsData dataclass for these so you don't have to rewrite this block of code each time? Or make an abstract base class that all of them can utilize and you just extend a few fields on all of them?

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -5,7 +5,7 @@
 
 from imap_processing.glows.l0 import decom_glows
 
-logging.basicConfig(level=logging.ERROR)
+# logging.basicConfig(level=logging.ERROR)
```
greglucas: Just remove?

greglucas (approved):
I like the dataclass approach here. I've got a few minor comments, but looks good to me.

greglucas:
Also an FYI that you have linting and doc failures, probably both pretty easy fixes: https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/6497938474/job/17648150672?pr=264#step:5:101

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -1,4 +1,27 @@
...
+from imap_processing.ccsds.ccsds_data import CcsdsData
+
+
+@dataclass
+class GlowsL0:
+    """Data structure for common values across histogram and direct events data.
+
+    Attributes
+    ----------
+    ground_sw_version: str
```
tech3371: What is `ground_sw_version`? Is the the version we are setting for each instrument in imap_processing or is it provided from somewhere else?

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -1,4 +1,27 @@
...
+    packet_file_name: str
+    ccsds_header: CcsdsData
+
+
+@dataclass
+class HistogramL0(GlowsL0):
     """Data structure for storing GLOWS histogram packet data.
 
     Parameters
```
tech3371: `__init__` function's `Parameters` has updated doc strings. Should we remove this or keep it?

imap_processing/glows/l0/glows_l0_data.py
tech3371: I saw in `decom_glows.py` that it's a version we are creating. should we rename it to something like `glows_sw_version`? `gound_` sounds like something MOC or flight software would use. That's why I was wondering that.

tech3371 (approved):
I just had couple minor comments. Otherwise, it looks good to me!

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: I considered it, but in the end decided that since CcsdsData is a different structure that isn't really a parent structure of the Level 0 data, it is preferable to keep it as a separate class. I may end up making a base class for all these dataclass structures if they end up having a lot of similar code, so I could be talked into doing that now if you feel strongly about it.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: Ah yeah, that makes sense. I named it that to match with the GLOWS algorithm document, but I understand your confusion. This is a glows-specific version inside a glows-specific dataclass, would it be acceptable to update the documentation to be more clear? Maybe to say something like "The software version of GLOWS SDC data processing code"

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: Good catch, I removed the `Parameters` section from the class and left it in the `__init__`. Hopefully that's the correct structure, I think that's correct according to our autodocs.

Issue #256, opened 2023-10-11
Create common script for selecting instrument/level processing containers


sdhoyt:
### Summary

command line utility for triggering process and data level. This can be done with the docker template issue

### Requirements

The more specific the better.
  - [ ]
  - [ ]

bourque:
This is done via #266

PR #266, opened 2023-10-17
Instrument processing command line utility

bourque:
# Change Summary

## Overview
This PR adds a script to be used on the command line to process a specific instrument and data level. Currently the methods within just print a message, but eventually these should contain code to perform any necessary steps to run the processing.

Relevant issue: #256 

## New Files

- `imap_processing/run_processing.py`
   - The script that can be run on the command line by passing in parameters for `instrument` and data `level`

## Updated Files

- `imap_processing/__init__.py`
   - Added list of `instruments` and list of `processing_levels`, used to validate the passed arguments


bourque:
@tech3371 @greglucas @laspsandoval Could you provide an initial review on this? I just want to make sure that I am on the right track here, and if I have designed this appropriately. Thanks!

imap_processing/__init__.py
```diff
@@ -8,3 +8,29 @@
...
+    "codice": ["l0", "l1a", "l1b", "l2"],
+    "glows": ["l0", "l1a", "l1b", "l2"],
+    "hi": ["l0", "l1a", "l1b", "l1c", "l2"],
+    "hit": ["l0", "l1a", "l1b", "l2"],
+    "idex": ["l0", "l1a", "l1b", "l2"],
+    "lo": ["l0", "l1a", "l1b", "l1c", "l2"],
+    "mag": ["l0", "l1a", "l1b", "l1c", "l2"],
+    "swapi": ["l0", "l1", "l2"],
+    "swe": ["l0", "l1a", "l1b", "l2"],
+    "ultra": ["l0", "l1a", "l1b", "l2"],
```
bourque: I still need to confirm some of these

imap_processing/run_processing.py
```diff
@@ -0,0 +1,122 @@
...
+    ----------
+    args : argparse.Namespace
+        An object containing the parsed arguments and their values
+    """
+    assert (
+        args.instrument in instruments
+    ), f"{args.instrument} is not a supported instrument"
+    assert (
+        args.level in processing_levels[args.instrument]
+    ), f"{args.level} is not a supported data level"
```
greglucas: Do we want to raise instead? I think you can also add some more context in the return message to let users know what is valid as well.
```suggestion
    if args.instrument not in instruments:
        raise ValueError(f"{args.instrument} is not in the supported instrument list: {instruments}")
    if args.level not in processing_levels[args.instrument]:
        raise ValueError(f"{args.level} is not a supported data level for the {args.instrument} instrument, valid levels are: {processing_levels[args.instrument]}")
```

imap_processing/run_processing.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class ProcessInstrument:
```
greglucas: I like this idea. Do you want to make this an abstract base class for a single instrument that then other people would override?

```python
class ProcessInstrument(abc.abc):
    def process(level):
        raise NotImplementedError


Codice(ProcessInstrument):
    # Not sure if we want this defined here, or in instrument modules?
    def process(level):
        # instruments would override this method themselves.

# lookup table to dispatch to each of the methods?

File: imap_processing/run_processing.py [hunk 4]
```diff
@@ -0,0 +1,122 @@
+"""Serves as a command line utility to run the processing for a specific
+instrument and data level.
+
+Use
+---
+    python run_processing.py <instrument> <data_level>
```
greglucas, line 10: Can you add this as an installed script as well?
```python
[project.scripts]
imap_processing = "my_package.module:function"
```

I would be explicit `imap_processing --instrument <instrument> --data-level <data_level>` Do we want data-level or just level?

imap_processing/run_processing.py
```diff
@@ -0,0 +1,122 @@
...
+        '"python run_processing swe l1a".'
+    )
+
+    instrument_help = (
+        "The instrument to process. Acceptable values are: " '"swe", and "codice".'
+    )
+    level_help = (
+        "The data level to process. Acceptable values are: "
+        '"l0", "l1a", "l1b", "l1c", and "l2".'
+    )
```
greglucas: Are you able to make this dependent on the different levels depending on what was specified as the instrument?

greglucas (commented):
:+1: I added a few ideas to consider, but overall I think this is the right direction!

imap_processing/run_processing.py
greglucas: https://packaging.python.org/en/latest/specifications/declaring-project-metadata/

imap_processing/run_processing.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+if __name__ == "__main__":
+    args = _parse_args()
+    _validate_args(args)
+
+    # Determine which function to invoke
+    process = ProcessInstrument(args.level)
+    method = getattr(process, f"process_{args.instrument}")
+    method()
```
tech3371: What are these two lines doing?

tech3371 (commented):
Yeah, it looks like you are heading in the right direction.

imap_processing/run_processing.py
```diff
@@ -0,0 +1,122 @@
+"""Serves as a command line utility to run the processing for a specific
```
tech3371: One thing I was thinking was that we may need to add something like below code block to be able to call from batch command. I don't remember the exact reason, @bryan-harter may know more about it. EMM does something like this and it has something to do with `Entrypoint` or something.
```
#!/usr/bin/env python3
```

greglucas: That is if you want to run it without specifying `python xyz.py` because it will look up your current python3 virtual environment for you and use that.

tech3371: I see. So if we specify exact python comamnd with python file with its argument, then we don't need to add that `#!/usr/bin/env python3` in the code. Is that right?

greglucas: Correct. It is still fine to add that line though if someone does want to execute it as a script, so you can do both.

tech3371: Good to know. Thank you!

imap_processing/run_processing.py
bourque: Line 121 is figuring out which method to call based on the `instrument` passed by the user, and line 122 is calling said method. I did this to avoid something like:

```python
if instrument == 'codice':
    process.process_codice()
elif instrument == 'swe':
    process.process_swe()
```

and so on

imap_processing/run_processing.py
```diff
@@ -0,0 +1,165 @@
...
+        "for a specific instrument and data level. Example usage: "
+        '"python run_processing swe l1a".'
+    )
+
+    instrument_help = f"The instrument to process. Acceptable values are: {instruments}"
+    level_help = (
+        f"The data level to process. Acceptable values are: {processing_levels}"
+    )
+
+    parser = argparse.ArgumentParser(description=description)
```
laspsandoval, line 39: This looks good. An example of the command that I was planning on using for the batch job will be:

['CodiceHi', {'l2': ['2023-10-31', '2023-06-12'], 'l3': ['2023-06-01']}, 2]

where 2 is the version number. 

And then I was planning on passing in the following environment:

                environment={
                    "OUTPUT_PATH": data_bucket.bucket_name,
                    "SECRET_NAME": db_secret_name,
                },

laspsandoval (approved):
Nice. Just update the arguments.

imap_processing/run_processing.py
greglucas: @laspsandoval , I don't think imap_processing will want the data bucket information or the DB secret because this is external to anything in our infrastructure. I think this is supposed to be going through our APIs to get the files instead.

I think the way this is written currently you'd need to kick off the container by overriding the ENTRYPOINT to be something like: `imap-processing --instrument codice-hi --level 2 --start-date XYZ --end-date XYZ` with command line arguments instead of a list/dict. Is that doable on the infrastructure side?

imap_processing/__init__.py
greglucas: Should these be in each instrument's base module information? So ultra's information would stick with ultra and they could update it themselves?

imap_processing/run_processing.py
laspsandoval: @greglucas as I understand it the algorithms will still need to query a database table, for example the universal spin table or a calibration table.

I was thinking that we had to append the s3 bucket information to the lambda url in order to use the upload/download api's. But I could be mistaken. Is that not correct?

greglucas: > @greglucas as I understand it the algorithms will still need to query a database table, for example the universal spin table or a calibration table.

I think we should avoid leaking anything database related between the processing and infrastructure. Can we make new endpoints for this instead?
`/calibration_lookup`, `/spin_table_lookup`, ... whatever else we need

> I was thinking that we had to append the s3 bucket information to the lambda url in order to use the upload/download api's. But I could be mistaken. Is that not correct?

I think the Lambda should take care of this for us? So a user should only need to know the URL, but the details of where we put that in the backend shouldn't matter to them.

`dev.imap-processing.com/upload` should map to a different bucket than `prod.imap-processing.com/upload`, but a user shouldn't need to update both the url and bucket name. I'm hoping we can handle that for them.

laspsandoval: @greglucas we could do as you suggested for the ENTRYPOINT, but to avoid duplicates I think we should do

imap-processing --instrument codice-hi --level 2 --dates [list of dates]

imap_processing/__init__.py
bryan-harter: IDEX will only have L1 and L2

imap_processing/run_processing.py
greglucas: We should talk about this at your processing trigger discussion meeting. If I am reprocessing an entire mission, do I want to put 5 x 365 dates into the command line (there may even be a limit for how long the input terminal can be), versus a start / end date? Could the processing code handle the duplicates? Yet another option is to only allow one date in. So we'd spin up 5 x 365 separate processing jobs all with one date input...

PR #268, opened 2023-11-02
Added project wide utils function and swe specific utils

tech3371:
# Change Summary
Added utils functions for project wide and SWE specific.

## New Files
- imap_processing/utils.py
- imap_processing/write_to_cdf.py

## Updated Files

- imap_processing/swe/__init__.py
   - Add version


.pre-commit-config.yaml
```diff
@@ -23,7 +23,7 @@ repos:
     rev: 'v0.0.292'
     hooks:
     - id: ruff
-      args: [--fix]
+      args: [--fix, --ignore, PLR0913]
```
greglucas: This should go in the `pyproject.toml` `per-file-ignores` block: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/10e75630bdca2637a8744356ffe3d873f7711d2b/pyproject.toml#L73

imap_processing/swe/swe_cdf_attrs.py
```diff
@@ -0,0 +1,217 @@
+"""CDF attrs explanation.
```
greglucas, line 1: This is really nice documentation! I think it'd be nice to put all of this into the cdf_utils file? Maybe move that into a separate module even: `cdf/utils.py` for organization?

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/cdf_utils.py

imap_processing/swe/swe_cdf_attrs.py
```diff
@@ -0,0 +1,217 @@
...
+            "wind. SWE is optimized to measure in situ solar wind electrons "
+            "at L1 to provide context for the ENA measurements and perform  "
+            "the in situ solar wind observations necessary to understand the"
+            "local structures that can affect acceleration and transport of "
+            "these particles. See"
+            "https://imap.princeton.edu/instruments/swe for more details."
+        )
+    ],
+    "Logical_file_id": ["FILL ME IN AT FILE CREATION"],
+} | cdf_utils.global_base
```
greglucas, line 112: Out of curiosity, will we care whether we are doing `global | swe` or `swe | global` to control what gets over-ridden in the dict?

imap_processing/swe/swe_cdf_attrs.py
```diff
@@ -0,0 +1,217 @@
...
+    "VAR_TYPE": "support_data",
+    "SCALETYP": "linear",
+}
+
+l1b_science_attrs = {
+    "CATDESC": "Science Data",
+    "DEPEND_0": "Epoch",
+    "DEPEND_1": "Energy",
+    "DEPEND_2": "Angle",
+    "DEPEND_3": "Rates",
```
greglucas, line 206: Not for you to worry about here, but an annoyance with CDF on my part... These DEPENDS seem like they should be automatically handled by specifying proper coordinates in xarray. My xarray datasets already know exactly what variables depend on which coordinates because it is a part of the spec, so that is already contained in the xarray metadata. This seems like we could abstract this out into the helper function to automatically add the DEPENDS based on the coordinates of the dataset... cc @maxinelasp

imap_processing/swe/utils/swe_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        description = "hk"
+    elif apid == SWEAPID.SWE_EVTMSG.value:
+        description = "evtmsg"
+    elif apid == SWEAPID.SWE_CEM_RAW.value:
+        description = "cemraw"
+    elif apid == SWEAPID.SWE_SCIENCE.value:
+        description = "sci"
+    else:
+        return None
+    return description
```
greglucas: Use a simple lookup table instead?
```suggestion
filename_descriptors = {SWEAPID.SWE_APP_HK: "hk", ...}
```

imap_processing/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        grouped data by apid
+    """
+    grouped_packets = collections.defaultdict(list)
+    for packet in packets:
+        apid = packet.header["PKT_APID"].raw_value
+        grouped_packets.setdefault(apid, []).append(packet)
+    return grouped_packets
+
+
+def convert_raw_to_eu(dataset: xr.Dataset, convertion_table_path, packet_name):
```
greglucas:
```suggestion
def convert_raw_to_eu(dataset: xr.Dataset, conversion_table_path, packet_name):
```
minor typo

imap_processing/write_to_cdf.py
```diff
@@ -0,0 +1,97 @@
...
+    Nick Dutton mentioned that s/c clock start epoch is
+        jan-1-2010-00:01:06.184 ET
+    We will use this for now.
+    """
+    # Get the datetime of Jan 1 2010 as the start date
+    launch_time = np.datetime64("2010-01-01T00:01:06.184")
+    return launch_time + np.timedelta64(shcoarse_time, "s")
+
+
+def write_to_cdf(
```
greglucas, line 36: This is already in cdf_utils. Can you combine this with that? Or leverage that one somehow?
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/10e75630bdca2637a8744356ffe3d873f7711d2b/imap_processing/cdf_utils.py#L46

.pre-commit-config.yaml
tech3371: Thank you! I was wondering and couldn't figure out. Could I add to this list instead with others.

```
ignore = ["D100", "D104", "PLR2004", "S101", "PLR0913"]
```

imap_processing/swe/swe_cdf_attrs.py
tech3371: Maxine is grouping all cdf utils into a folder. I could move it there after her changes are in.

imap_processing/swe/utils/swe_utils.py
tech3371: will do that.

imap_processing/utils.py
tech3371: I noticed that and changed on the other PR. Will change it here too

imap_processing/write_to_cdf.py
tech3371: Since my upcoming PR is using that, I didn't want to change it here. I created an issue to combine those two functions into one and add updates CDF attrs to have filename parameters in the dataset as we discussed on Slack. I will do that in this issue: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/270

imap_processing/swe/swe_cdf_attrs.py
tech3371: It would be nice if we can abstract this out from xr metadata.

imap_processing/swe/swe_cdf_attrs.py
tech3371: I had to think about it. I think so. I think the order matters here. If there is same variable between `global` and `swe`, the latter will replace what's in the first one. Right now, I think we kept it unique, so the order will not matter but could be something to think about if we have same variable.

.pre-commit-config.yaml
greglucas: I'd prefer not because then that applies to every file and I think it is good to leave in as a general check.

tech3371: ok. Will add to per-files-ignores

imap_processing/swe/utils/swe_utils.py
greglucas: Remove the function altogether and just use the lookup dictionary directly? The function here is just wrapping the dict lookup, so if you really wanted to do a method `()` call instead of an index `[]` call, I think you might be able to do: `get_descriptor = {...}.get`

tech3371: so if we remove function, `filename_descriptors` will be global variable in the `swe_utils.py`?

greglucas: yes, it would be a module level variable.

```python
swe_utils.filename_descriptors[1234]
# versus
swe_utils.get_descriptor(1234)
```

tech3371: ok. That makes sense now.

bourque (approved):
It looks like a lot of these changes are also included in #253? In any case, they look good!

tech3371:
all of this changes is in this PR https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/253

PR #272, opened 2023-11-06
Create CDF class based system

### LINKED ISSUE #125: Code for creating standard CDF files across instruments
### Summary

We determined that we will use XArray to generate CDF files, so we need to create a standard piece of code which contains shared requirements from ISTP that instruments can use. 

### Requirements

The more specific the better.
  - [ ] Create a standardized way to create metadata which aligns with ISTP standards for CDF files
  - [ ] Create any common attributes, variables, or utility methods which would aid in CDF file writing

### Related tickets

Solves spike #71

maxinelasp:
# Change Summary
This includes the first draft of the CDF file attribute creation system, with IDEX changes. 

## Overview
The changes include the class based system for the CDF attributes, and an example on how to use these classes with the IDEX CDF generation code. If there are other ideas for ways to use these classes, please let me know and I can see how that would work. 

These also work with the draft SWE changes from Tenzin, which is why some of the parameters are unused. 

closes #125

## New Files

- cdfutils/global_base.py
   - All the basic classes for CDF attributes
- cdfutils/constants.py
   - Constants for use in CDF attributes 

## Updated Files

- cdflib.py
   - I moved shared attributes into global_base, and moved the file to cdfutils/
- idex_cdf_attrs.py
  - I updated this code to use the new classes instead of manually defining dictionaries. In a lot of cases, the defaults worked, requiring less specification  
- other idex files
  - Generally just updating the usages of the cdf attributes

## Testing
 No new testing


maxinelasp:
I'd really like a review from @bryan-harter, since this is modifying his code and he's the CDF expert!

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+    Attributes
+    ----------
+    format: str, default="I18"
+        The format of the data, in Fortran format
+    units: str, default="int"
+        The units of the data
+    """
+
+    format: str = "I18"
+    units: str = "int"
```
bryan-harter: Maybe default to an empty string? Just because that's the default for being "unitless"

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+
+        Returns
+        -------
+        Dictionary of correctly formatted values for the attributes in the class
+        """
+        return {
+            "CATDESC": self.catdesc,
+            "DISPLAY_TYPE": self.display_type,
+            "FIELDNAM": self.fieldname,
+            "FILLVAL": GlobalConstants.INT_FILLVAL,
```
bryan-harter: return self.fill_val instead of this hard-coded int_fillval?

bryan-harter:
Overall looks good to me!

bryan-harter:
Maybe you can add more documentation about how to inherit these things in the CDF docs once that stuff is online

imap_processing/cdfutils/global_base.py
maxinelasp: Good catch, will update

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+        The scale of the axis, "SCALETYP" attribute
+    label_axis: str, default=None
+        Axis label, "LABLAXIS" attribute
+    format: str, default=None
+        The format of the data, in Fortran format
+    units: str, default=None
+        The units of the data
+    """
+
+    validmin: np.float64 | np.int64
```
greglucas:
```
    validmin: np.float64 | np.int64
E   TypeError: unsupported operand type(s) for |: 'type' and 'type'
```
I think that might only be for 3.10+, so you might need to do a typing.Union on it? or you could do a `numbers.Real` as well?

imap_processing/cdfutils/defaults.py
```diff
@@ -0,0 +1,53 @@
...
+
+    INT_FILLVAL = np.iinfo(np.int64).min
+    INT_MAXVAL = np.iinfo(np.int64).max
+    DOUBLE_FILLVAL = np.float64(-1.0e31)
+    MIN_EPOCH = -315575942816000000
+    MAX_EPOCH = 946728069183000000
+
+
+@dataclass
+class IdexConstants:
```
greglucas: Do we want this in the main `cdf` module, or do we want to put instrument specific constants with the instruments?

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
+from dataclasses import dataclass
+
+import numpy as np
+
+from imap_processing.cdfutils.defaults import GlobalConstants
+
+
+class Epoch:
```
greglucas: For these global things with static methods, what if we just made them a dictionary? So assigned them to a global constant `EPOCH = {...}` and then later you wouldn't have to reference the `.output()` call from the method, but you could just pipe this global constant in directly.

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+            "VAR_TYPE": "support_data",
+            "SCALETYP": "linear",
+            "MONOTON": "INCREASE",
+            "TIME_BASE": "J2000",
+            "TIME_SCALE": "Terrestrial Time",
+            "REFERENCE_POSITION": "Rotating Earth Geoid",
+        }
+
+
+class GlobalBase:
```
greglucas: Since you asked for opinions on naming, I'll give a quick suggestion. Add attrs, or something about cdf to this?

`GlobalAttrs`, would make it explicit that these are CDF Attributes when used elsewhere and inherited. In the IDEX namespace you are using `InstrumentBase` and things, but it is a bit unclear what is in the InstrumentBase. Another option would be to leave the base as well, `InstrumentAttrsBase`, but then it seems like it is getting long.

Overall, I think Base is fine, but just figured I throw out the Attrs to correspond to what we are using these objects for.

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+        Generate the output for the data level as a dictionary.
+
+        Returns
+        -------
+        Dictionary of correctly formatted values for the attributes in the class.
+        If the optional parameters are not defined, they are not included as attributes
+        in the output.
+        """
+        if self.depend_0 is None:
+            raise TypeError("ScienceBase requires depend_0 attribute.")
```
greglucas: Could these checks be done in `__post_init__`?
https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init__

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+
+        if self.depend_2 is not None:
+            endval["DEPEND_2"] = self.depend_2
+
+        if self.variable_purpose is not None:
+            endval["VARIABLE_PURPOSE"] = self.variable_purpose
+
+        if self.var_notes is not None:
+            endval["VAR_NOTES"] = self.var_notes
+        return endval | super().output()
```
greglucas: A slight preference to reverse these in all the output calls, so the child would replace the parent value.
```suggestion
        return super().output() | endval
```

imap_processing/cdfutils/cdf_utils.py
greglucas, line 1: From a naming perspective, I think this would be nicer to have a `cdf` module, with a `utils.py` file under that. So move this file to `imap_processing/cdf/utils.py`

greglucas (commented):
Very nice! I like this approach 👍 
Some inline suggestions, but all quite minor. One thing that would be nice is to make a `test_cdf.py` file and assert the `output()` calls work as expected and you can inherit / put keywords in and get the expected output back.

imap_processing/cdfutils/cdf_utils.py
bourque: I agree with Greg here, or alternatively renaming to `cdf_utils/cdf_utils.py` for consistency.  I don't know why but `cdfutils` bugs me. Maybe because I read it as "CD Futil[ity]" 😅

imap_processing/cdfutils/defaults.py
```diff
@@ -0,0 +1,53 @@
+from dataclasses import dataclass
```
bourque: I suggest adding a module docstring here that describes the purpose of the module, and maybe some example usage (same for `global_base.py`).

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+    @staticmethod
+    def output():
+        """
+        Return a dictionary with global base attributes.
+
+        Returns
+        -------
+        Global file attributes, including project, source_name, discipline, PI_name,
+        PI_affiliation, instrument_type, and mission_group. This should be the same
+        for all instruments.
```
bourque:
```suggestion
        dict
            Global file attributes, including project, source_name, discipline, PI_name,
            PI_affiliation, instrument_type, and mission_group. This should be the same
            for all instruments.
```

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+    text: str
+
+    def output(self):
+        """
+        Generate the output for the instrument as a dictionary.
+
+        Returns
+        -------
+        dictionary of correctly formatted values for the data_version, descriptor, text,
+         and logical_file_id, added on to the global attributes from GlobalBase
```
bourque:
```suggestion
        dict 
            dictionary of correctly formatted values for the data_version, descriptor, text,
            and logical_file_id, added on to the global attributes from GlobalBase
```

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+    instrument_base: InstrumentBase
+
+    def output(self):
+        """
+        Generate the output for the data level as a dictionary.
+
+        Returns
+        -------
+        dictionary of correctly formatted values for the attributes in the class and
+        the attributes from InstrumentBase
```
bourque:
```suggestion
        dict
            dictionary of correctly formatted values for the attributes in the class and
            the attributes from InstrumentBase
```

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+    format: str = None
+    units: str = None
+
+    def output(self):
+        """
+        Generate the output for the data level as a dictionary.
+
+        Returns
+        -------
+        Dictionary of correctly formatted values for the attributes in the class
```
bourque:
```suggestion
        dict
            Dictionary of correctly formatted values for the attributes in the class
```

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def output(self):
+        """
+        Generate the output for the data level as a dictionary.
+
+        Returns
+        -------
+        Dictionary of correctly formatted values for the attributes in the class.
+        If the optional parameters are not defined, they are not included as attributes
+        in the output.
```
bourque:
```suggestion
        dict
            Dictionary of correctly formatted values for the attributes in the class.
            If the optional parameters are not defined, they are not included as attributes
            in the output.
```

imap_processing/cdfutils/global_base.py
```diff
@@ -0,0 +1,353 @@
...
+    var_type: str = "metadata"
+    display_type: str = "no_plot"
+
+    def output(self):
+        """
+        Generate the output for the data level as a dictionary.
+
+        Returns
+        -------
+        Dictionary of correctly formatted values for the attributes in the class.
```
bourque:
```suggestion
        dict
            Dictionary of correctly formatted values for the attributes in the class.
```

bourque (approved):
Thank you for building all of this @maxinelasp and taking the time to present it to us! It looks great, I just had a few minor comments/suggestions, but nothing that would block merging.

imap_processing/cdfutils/defaults.py
GFMoraga: I agree with Matthew. If these are utils (the default), a description would be helpful.

GFMoraga (approved):
I think this all looks really great and helpful. All the changes you made make sense and I think Greg pointed out some good things to consider.

imap_processing/cdfutils/cdf_utils.py
greglucas: I sort of would be against `cdf_utils/cdf_utils.py` because I find that hard from a module import perspective to remember where you are when they are repeated. I think this would be akin to datetime which can lead to unnecessary confusion IMO: Did the top of the module `import datetime` so I need `datetime.datetime()` in the file, or did they do `from datetime import datetime` so I only need `datetime()`.

imap_processing/cdfutils/global_base.py
maxinelasp: Sure, I put them as classes in case people want to overwrite them, but I'm not sure if that's actually something that will happen (especially for some of the global attributes.) I can move some of them if they aren't going to change though

imap_processing/idex/idex_cdf_attrs.py
```diff
@@ -1,173 +1,177 @@
...
+    display_type="no_plot",
+)
+#     {
+#     "DEPEND_0": "Epoch",
+#     "FILLVAL": Constants.INT_FILLVAL,
+#     "FORMAT": "I12",
+#     "VALIDMIN": 0,  # All values are positive integers or 0 by design
+#     "VAR_TYPE": "data",
+#     "DISPLAY_TYPE": "no_plot",
+# }
```
tech3371, line 84: Do you need this for later?

tech3371 (approved):
Looks good to me! looking forward to try it in SWE.

imap_processing/idex/idex_cdf_attrs.py
maxinelasp: I wasn't sure of the answer, it should be answered probably by the IDEX developer since it's related to the specific IDEX attribute

imap_processing/cdf/global_attrs.py
```diff
@@ -0,0 +1,353 @@
...
+    def output(self):
+        """
+        Generate the output for the data level as a dictionary.
+
+        Returns
+        -------
+        dict
+            Dictionary of correctly formatted values for the attributes in the class.
+        """
+        return {
```
greglucas, line 346: Interesting, none of the current tests are using StringAttrs and calling `.output()`?

imap_processing/cdf/global_attrs.py
```diff
@@ -0,0 +1,353 @@
...
+    depend_0: str = None
+    depend_1: str = None
+    depend_2: str = None
+    variable_purpose: str = None
+    var_notes: str = None
+
+    def __post_init__(self):
+        """If depend_0 is not set, raise an error, as this attribute is required."""
+        if self.depend_0 is None:
+            raise TypeError("ScienceBase requires depend_0 attribute.")
```
greglucas, line 259: This seems like something we should make sure we are actually testing. I think adding a new module to just test these attribute classes makes some sense since they will be re-used heavily and they are somewhat non-trivial in their implementation.

```python
def test_science_attrs():
    # Must provide depend_0
    with pytest.raises(TyperError, "ScienceBase requires"):
        ScienceAttrs()
    science = ScienceAttrs(depend_0="test0", depend_1="test1", depend_2="test2").output()
    assert science == {"DEPEND_0": "test0", "DEPEND_1": "test1", "depend_2": "test2"}

greglucas (approved):
I like the new names you settled on! Looks good to me. I added a quick suggestion for a test that I think will help out with codecov. Maybe there are a few more that could be tested too like the base levels of inheritance so we sort of codify that behavior in the tests. Adding a new directory/file `imap_processing/cdf/tests/test_cdf.py` for that.

imap_processing/cdf/global_attrs.py
maxinelasp: Yes, that's leftover from some SWE code that isn't merged yet

imap_processing/cdf/global_attrs.py
maxinelasp: Sure, I can add a test on some of these classes just to catch errors like the one Bryan caught earlier, where the attribute was in the dataclass but not in the output. Since these aren't expected to have attributes removed I think that makes a certain amount of sense.

imap_processing/idex/idex_cdf_attrs.py
tech3371: It was commented out. That's why I was wondering. cool

PR #278, opened 2023-11-14
CoDICE decompression algorithms for science data

### LINKED ISSUE #249: Create module for decompressing CoDICE science data
Implement the CoDICE uncompression algorithms

bourque:
# Change Summary

## Overview
This PR implements the six possible CoDICE decompression algorithms for science data, with supporting unit tests. The algorithms are:

1. No compression
2. Lossy compression A (lookup table)
3. Lossy compression B (lookup table)
4. Lossless compression (LZMA)
5. Lossy A + Lossless
6. Lossy B + Lossless

This PR closes #249 , however there will likely need to be some tweaks down the line to get this working with the greater CoDICE L0-L1A decom algorithm

## New Dependencies
None

## New Files
- `imap_processing/codice/l0/decompress_codice.py`
   - Contains methods that implement the decompression algorithms
- `imap_processing/codice/tests/test_decompress_codice.py`
  - Parameterized unit tests for `decompress_codice.py` 

## Deleted Files
None

## Updated Files
- `imap_processing/codice/l0/__init__.py`
   - This is where I decided to store the Lossy compression lookup tables

## Testing
`test_decompress_codice.py` contains a unit test for each of the 6 algorithms using an input value of `234` (I picked this randomly), as well as an attempt to use an unsupported algorithm, which should result in a decompressed value of `None`.


imap_processing/codice/l0/decompress_codice.py
```diff
@@ -0,0 +1,111 @@
+"""Decompress CoDICE science data.
+
+For CoDICE, there are 3 forms of compression:
+    1. Table-based lossy compression A (24 -> 8-bit)
+    2. Table-based lossy compression B (24 -> 8 bit)
```
tech3371: to be consistent with below functions, should we change this or point out that it could change.
```suggestion
    1. Table-based lossy compression A (32 -> 8-bit)
    2. Table-based lossy compression B (32 -> 8 bit)
```

imap_processing/codice/l0/decompress_codice.py
```diff
@@ -0,0 +1,111 @@
...
+        elif self.algorithm == "lossyB":
+            self._apply_lossy_b(self.compressed_value)
+        elif self.algorithm == "lossless":
+            self._apply_lzma_lossless(self.compressed_value)
+        elif self.algorithm == "lossyA+lossless":
+            self._apply_lzma_lossless(self.compressed_value)
+            self._apply_lossy_a(self.decompressed_value)
+        elif self.algorithm == "lossyB+lossless":
+            self._apply_lzma_lossless(self.compressed_value)
+            self._apply_lossy_b(self.decompressed_value)
```
tech3371: Would it be helpful to create a class for these algorithm types like we did for apid. Then use its value to compare against the input value. This way, if name changes it won't effect. We can lookup by `compression` id.
```
class SWEAPID(IntEnum):
    """Create ENUM for apid.

    Parameters
    ----------
    IntEnum : IntEnum
    """

    SWE_APP_HK = 1330
    SWE_EVTMSG = 1317
    SWE_CEM_RAW = 1334
    SWE_SCIENCE = 1344
```

tech3371 (commented):
Nice work!

imap_processing/codice/l0/__init__.py
```diff
@@ -0,0 +1,515 @@
+LOSSY_A_TABLE = {
```
greglucas: Might want to put this in a separate file to make it more explicit what these are referring to? `constants.py` or `table_data.py`

greglucas: Where did you get these numbers from? It might be nice to put a reference to the specific page number of the algorithm document / sheet for easy reference in the future.

imap_processing/codice/l0/decompress_codice.py
```diff
@@ -0,0 +1,111 @@
...
+must apply lossless decompression first, then lossy decompression
+
+References
+----------
+    This information was provided via email from Greg Dunn on Oct 23, 2023
+"""
+
+import lzma
+
+from . import LOSSY_A_TABLE, LOSSY_B_TABLE
```
greglucas: explicit imports rather than relative? `from imap_processing.codice.l0 import ...`
It looks like we are doing explicit in other places already, but I don't know if there is any standard form referenced anywhere.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/ec06ec1a7bc866dac7f5600cef0ef8408aaa6618/imap_processing/swe/l1a/swe_l1a.py#L3-L4

imap_processing/codice/l0/decompress_codice.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class CodiceDecompression:
```
greglucas: Is there a reason to use a class here? This seems like a function to me: `def decompress_codice(compressed_value, algorithm) -> decompressed_value`
Basically keeping all the functions you have and just returning instead of setting attributes on self.

imap_processing/codice/l0/decompress_codice.py
```diff
@@ -0,0 +1,111 @@
...
+            self.decompressed_value = self.compressed_value
+        elif self.algorithm == "lossyA":
+            self._apply_lossy_a(self.compressed_value)
+        elif self.algorithm == "lossyB":
+            self._apply_lossy_b(self.compressed_value)
+        elif self.algorithm == "lossless":
+            self._apply_lzma_lossless(self.compressed_value)
+        elif self.algorithm == "lossyA+lossless":
+            self._apply_lzma_lossless(self.compressed_value)
+            self._apply_lossy_a(self.decompressed_value)
```
greglucas: The order of these confused me at first until I read the class docstring above indicating that these are the inverse of the original algorithm that did the compressing. Thank you for adding the great documentation here!

imap_processing/codice/l0/__init__.py
greglucas: Are these tables all sequential lists? You could just make them a list and still index into them in the same way without needing the keys() stored too. `LOSS_A_TABLE = [1, 2, 3, 4, ...]`

imap_processing/codice/tests/test_decompress_codice.py
```diff
@@ -0,0 +1,34 @@
...
+import pytest
+
+from imap_processing.codice.l0.decompress_codice import CodiceDecompression
+
+# The value of 234 compressed using LZMA
+LZMA_EXAMPLE = (
+    b"\xfd7zXZ\x00\x00\x04\xe6\xd6\xb4F\x02\x00!\x01\x16\x00\x00\x00t/\xe5\xa3"
+    b"\x01\x00\x00\xea\x00\x00\x00\x00\x15g\x1b\xad\x02.eJ\x00\x01\x19\x01\xa5,"
+    b"\x81\xcc\x1f\xb6\xf3}\x01\x00\x00\x00\x00\x04YZ"
+)
```
greglucas: Do you want to just do the compression of your value?
```suggestion
# The value of 234 compressed using LZMA
LZMA_EXAMPLE = lzma.compress((234).to_bytes())
```

imap_processing/codice/tests/test_decompress_codice.py
```diff
@@ -0,0 +1,34 @@
...
+
+# Test the algorithms using input value of 234 (picked randomly)
+TEST_DATA = [
+    (234, "no compression", 234),
+    (234, "lossyA", 221184),
+    (234, "lossyB", 1441792),
+    (LZMA_EXAMPLE, "lossless", 234),
+    (LZMA_EXAMPLE, "lossyA+lossless", 221184),
+    (LZMA_EXAMPLE, "lossyB+lossless", 1441792),
+    (None, "some_unsupported_algorithm", None),
```
greglucas: Might want to raise in this case?

greglucas (commented):
Looks great! Especially the comments and tests.

imap_processing/codice/l0/__init__.py
bourque: Good idea, I will add a note. Unfortunately these are not defined in any sort of official document -- They were given to us via email from Greg Dunn.

bourque: This could certainly work, though at least for now I prefer the dict format so I can easily see which 8 bit value corresponds to the 32 bit value.

imap_processing/codice/l0/decompress_codice.py
bourque: Good call, I prefer explicit too.

imap_processing/codice/l0/decompress_codice.py
bourque: I am not sure I follow here @tech3371. Are you saying to create a class for each algorithm? e.g.:


```python
class LossyA(value):

    decompressed_value = LOSSY_A_TABLE[value]
````

Comment (tech3371) [reply imap_processing/codice/l0/decompress_codice.py]: I was thinking something like this:
```
class CoDICECompression(IntEnum):
    """Create ENUM for CoDICE compression algorithms.

    Parameters
    ----------
    IntEnum : IntEnum
    """

    NO_COMPRESSION = 1
    LOSSY_A = 2
    LOSSY_B = 3
    LOSSLESS = 4
    .
    .
```

```suggestion
        if self.algorithm == CoDICECompression.NO_COMPRESSION:
            self.decompressed_value = self.compressed_value
        elif self.algorithm == CoDICECompression.LOSSY_A:
            self._apply_lossy_a(self.compressed_value)
        elif self.algorithm == CoDICECompression.LOSSY_B:
            self._apply_lossy_b(self.compressed_value)
        elif self.algorithm == CoDICECompression.LOSSLESS:
            self._apply_lzma_lossless(self.compressed_value)
        .
        .
```
With this changes, you can input raw int value of compression and compare like that.

imap_processing/codice/l0/decompress_codice.py
bourque: Yeah I went back and forth on this. I don't really have a good reason here to use a class, but now that I reworked this into using functions, I do think the code is a bit cleaner. So good suggestion here.

imap_processing/codice/l0/decompress_codice.py
bourque: Ah, I see now. That looks good, thanks!

tech3371 (approved):
looks great!

sdhoyt (approved):
This code was weirdly pleasant to look at lol everything looks great!

imap_processing/codice/l0/decompress_codice.py
```diff
@@ -0,0 +1,124 @@
...
+    algorithm : str
+        The algorithm to apply. Supported algorithms include 'no compression',
+        'lossyA', 'lossyB', 'lossless', 'lossyA+lossless', and 'lossyB+lossless'
+
+    Returns
+    -------
+    decompressed_value : int
+        The 24- or 32-bit decompressed value
+    """
+    if algorithm == CoDICECompression.NO_COMPRESSION:
```
maxinelasp, line 108: You could pass a function into this method instead of this if/else table. So, you could pass in `_apply_lossy_a` as `algorithm` and then this whole table could just be `algorithm(decompress)`. Not sure if that makes sense for the rest of your code, but it could be an interesting option to reduce the number of functions and get rid of that enum. 

https://stackoverflow.com/questions/1349332/python-passing-a-function-into-another-function

maxinelasp (approved):
LGTM, left one suggestion that probably isn't a good match for this piece of code, but might help with later development based on this style

greglucas (approved):
minor nit on the xfail/raises test. But looks good to me!

imap_processing/codice/l0/__init__.py
greglucas: :+1: Sounds good, just something to keep in mind in the future. My guess is you might be wanting to turn these into numpy arrays down the road and then you could send multiple compressed values at once into these to do faster lookups all at the same time rather than a Python for-loop for each individual value.

GFMoraga (approved):
Beautiful work. Everything looks great. I thought your last commit was a good catch. I think its ready to merge, and at a good spot whenever updates need to happen. (hopefully not anytime soon or too monstrous of changes)

PR #283, opened 2023-11-21
updates to IMAP-Hi and SWE's document

tech3371:
# Change Summary
Before I forget, small PR for updates in document to include hi and minor updates to SWE.

## New Files
- docs/source/reference/hi.rst
   - add hi to docs

## Updated Files

- docs/source/reference/index.rst
   - update to include hi
- docs/source/reference/swe.rst
   - updates SWE docs

### DIFF
--- a/docs/source/reference/hi.rst
+++ b/docs/source/reference/hi.rst
@@ -1,0 +1,29 @@
+.. _hi:
+
+IMAP-Hi
+==========================
+
+This is the IMAP-Hi Instrument module, which contains the code for processing
+data from the IMAP-Hi instrument.
+
+The L0 code to decommutate the CCSDS packet data can be found below.
+
+.. currentmodule:: imap_processing
+
+.. autosummary::
+    :toctree: generated/
+    :template: autosummary.rst
+    :recursive:
+
+    decom
+
+IMAP-Hi test and validation code can be found below.
+
+.. currentmodule:: imap_processing.hi
+
+.. autosummary::
+    :toctree: generated/
+    :template: autosummary.rst
+    :recursive:
+
+    tests
--- a/docs/source/reference/hi.rst
+++ b/docs/source/reference/hi.rst
@@ -5,25 +5,3 @@ IMAP-Hi
 
 This is the IMAP-Hi Instrument module, which contains the code for processing
 data from the IMAP-Hi instrument.
-
-The L0 code to decommutate the CCSDS packet data can be found below.
-
-.. currentmodule:: imap_processing
-
-.. autosummary::
-    :toctree: generated/
-    :template: autosummary.rst
-    :recursive:
-
-    decom
-
-IMAP-Hi test and validation code can be found below.
-
-.. currentmodule:: imap_processing.hi
-
-.. autosummary::
-    :toctree: generated/
-    :template: autosummary.rst
-    :recursive:
-
-    tests
--- a/docs/source/reference/swe.rst
+++ b/docs/source/reference/swe.rst
@@ -16,3 +16,23 @@ The L0 code to decommutate the CCSDS packet data can be found below.
     :recursive:
 
     l0.decom_swe
+
+The L1A code to process unpack counts per ESA step can be found below.
+
+.. autosummary::
+    :toctree: generated/
+    :template: autosummary.rst
+    :recursive:
+
+    l1a.swe_l1a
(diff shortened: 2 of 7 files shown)


docs/source/reference/hi.rst
```diff
@@ -0,0 +1,29 @@
...
+.. currentmodule:: imap_processing
+
+.. autosummary::
+    :toctree: generated/
+    :template: autosummary.rst
+    :recursive:
+
+    decom
+
+IMAP-Hi test and validation code can be found below.
```
greglucas: I personally wouldn't worry about documenting the test and validation code, probably especially with the autosummary just because we don't generally put great docstrings into the tests. Maybe just a link to specific test or validation methods that would be useful to point others to (if any)?

docs/source/reference/hi.rst
```diff
@@ -0,0 +1,29 @@
...
+
+IMAP-Hi
+==========================
+
+This is the IMAP-Hi Instrument module, which contains the code for processing
+data from the IMAP-Hi instrument.
+
+The L0 code to decommutate the CCSDS packet data can be found below.
+
+.. currentmodule:: imap_processing
```
greglucas: This should be documented in the root level `imap_processing` location, not within `hi`.

docs/source/reference/swe.rst
```diff
@@ -16,3 +16,25 @@ The L0 code to decommutate the CCSDS packet data can be found below.
...
+
+.. autosummary::
+    :toctree: generated/
+    :template: autosummary.rst
+    :recursive:
+
+    l1a.swe_l1a
+    l1a.swe_science
+
+The L1A code to process electron counts to rates and then
```
greglucas:
```suggestion
The L1B code to process electron counts to rates and then
```
Out of curiosity are we capitalizing both "L" and "B" here, or are we calling them "L1b" or "l1b"?

greglucas: I like that you broke these up with a description of what they do too!

greglucas (commented):
I think you'll need to update the references in the docstrings to be the fully qualified package name. `np.array` -> `numpy.ndarray`, `xr.Dataset` -> `xarray.Dataset`, `Path` -> `pathlib.Path`, I think you also have an extra `_type_` in there too.

docs/source/reference/hi.rst
tech3371: I didn't have anything to add for Hi at this moment besides those two. I could remove those and then we can add L1A as we develop.

docs/source/reference/swe.rst
tech3371: good catch! If I start L with then it made sense to end with capital letter as well. Otherwise, not particular about it.

docs/source/reference/hi.rst
greglucas: I'd suggest removing them for now then. I actually think a description of the instrument and what it does is just fine for this start of documentation without any need to reference code here if there is nothing to reference. Just making the placeholder is a win :)

docs/source/reference/swe.rst
greglucas: I'm not particular about which version to choose, my only preference is for consistency if we can meet that. I think filenames will be lowercase `l1b`, but maybe if we are writing in sentences we will write `L1B` with all caps like you have?

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -139,28 +139,28 @@ def calculate_calibration_factor(time):
     3. Linear interpolate between those two nearest time and get factor for input time.
 
     What this function is doing:
```
greglucas, line 142: Might need an extra linebreak here so the 1. isn't wrapped into this statement.

Also, the list above this might need to have the pipes too?
https://imap-processing--283.org.readthedocs.build/en/283/reference/generated/imap_processing.swe.l1b.swe_l1b_science.calculate_calibration_factor.html#imap_processing.swe.l1b.swe_l1b_science.calculate_calibration_factor

tech3371: fixed that

greglucas (approved):

bourque (approved):
Looks great!

bourque:
@all-contributors please add @tech3371 for documentation.

PR #286, opened 2023-11-22
Doc configuration updates

### LINKED ISSUE #271: BUG - Github CI doc generation is not working with python 3.11
## Description of the issue

When using python 3.11 in the CI doc.yml, sphinx does not pull from python's documentation. For example, a class with the super class `IntEnum` is causing an error in sphinx:

`/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/imap_processing/hit/l0/hit_l1a_decom.py:docstring of imap_processing.hit.l0.hit_l1a_decom.HitAPID:42:<autosummary>:1: WARNING: py:obj reference target not found: imap_processing.hit.l0.hit_l1a_decom.HitAPID.bit_count`

because it cannot find a target for built-in python functions in the Subclass 

This issue seems specific to the super class `IntEnum`. If using `Enum` instead the docs will build.

May be related to this issue:
https://github.com/sphinx-doc/sphinx/issues/10410

1. 2. 3.

Run doc gen using commit #dd9464c

### Code Snippet:

```python
- name: Set up Python
      uses: actions/setup-python@v4
      with:
        python-version: '3.11'
```

Docs should generate

`/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/imap_processing/hit/l0/hit_l1a_decom.py:docstring of imap_processing.hit.l0.hit_l1a_decom.HitAPID:42:<autosummary>:1: WARNING: py:obj reference target not found: imap_processing.hit.l0.hit_l1a_decom.HitAPID.bit_count`

### Additional notes

_No response_

_No response_

### Suggested fix?

_No response_

greglucas:
# Change Summary

In #271 we identified that we had to build documentation on Python 3.10 or lower due to IntEnum inheritance. Instead of pinning the lower Python version, we can instead ignore the places that cause the issues directly with numpydoc and `nitpick_ignore` options.
https://numpydoc.readthedocs.io/en/latest/install.html

The inherited methods are `to_bytes()` and the like from the `Int` class, which we don't really need/want to show up anyways. They still show up in the documentation, they just aren't clickable.

Additionally, put the `nitpicky = True` into the configuration, so that anyone building the docs gets that locally as well.

closes #271

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,32 +27,3 @@ jobs:
         # SPHINXOPTS="-W --keep-going" is used to turn warnings into errors
         # but keep going, so we don't fail on the first one.
         make -C docs html SPHINXOPTS="-W --keep-going"
-
-    - uses: actions/upload-artifact@v2
-      with:
-        name: DocumentationHTML
-        path: docs/build/html/
-
-    # Publish built docs to gh-pages branch
-    - name: Commit documentation changes
-      # push docs only when a GitHub Release is made
-      if: github.event_name == 'release' && github.event.action == 'published'
-      run: |
-        git clone https://github.com/IMAP-Science-Operations-Center/imap_processing.git --branch gh-pages --single-branch gh-pages
-        cp -r docs/build/html/* gh-pages/
-        cd gh-pages
-        git config --local user.email "action@github.com"
-        git config --local user.name "GitHub Action"
-        git add .
-        git commit -m "Update documentation" -a || true
-        # The above command will fail if no changes were present, so we ignore that.
-
-    - name: Publish docs
-      # push docs only when a GitHub Release is made
-      if: github.event_name == 'release' && github.event.action == 'published'
-
-      uses: ad-m/github-push-action@master
-      with:
-        branch: gh-pages
-        directory: gh-pages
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -24,35 +24,6 @@ jobs:
...
-
-    - name: Publish docs
-      # push docs only when a GitHub Release is made
-      if: github.event_name == 'release' && github.event.action == 'published'
-
-      uses: ad-m/github-push-action@master
-      with:
-        branch: gh-pages
-        directory: gh-pages
-        github_token: ${{ secrets.GITHUB_TOKEN }}
```
bourque, line 58: I am guessing we are removing all of this because we don't use GitHub pages now that we are using RTD?

bourque (approved):
Aside from my one comment here, this looks good to me! And it will help me with my CoDICE doc problems on #288 I think. Thank you

.github/workflows/docs.yml
greglucas: Exactly, this was so you could download and the artifacts created, but much easier to just look at the read the docs built PR pages instead now that we have that set up.

PR #287, opened 2023-11-27
Glows L1A Processing Algorithm

maxinelasp:
# Change Summary
Processing for GLOWS L1A data processing. Most of the specific byte manipulation in DirectEventsL1A comes from GLOWS example code. 

## Overview

The main changes are in `glows_l1a_data.py` where the processing from level 0 to level 1A occurs. I also made some changes in level 0, but they aren't any functionality changes and are just modifying the way the dataclasses are used. 

This doesn't include any CDF reading or writing steps for GLOWS. 

## New Dependencies

None

## New Files

- glows_l1a_data.py
   - Creates the L1A data from L0 histogram and direct events packets
- glows/utils/constants.py
  - file for classes which I expect to use across multiple data levels

## Updated Files

- glows_l0_data
   - Updated L0 dataclasses to use default dataclass init function instead of creating one from the packet. This made the dataclass more flexible and helped with later testing
- decom_glows
   - Updating to use the new init functions
- GLX_COMBINED.xml
   - Updating to properly process the direct event data, the spreadsheet didn't include the actual data attribute

## Testing

I added thorough tests for all the L1A functionality, using manual steps and the example GLOWS code to process events. I also added some additional level 0 tests to cover some gaps I found


imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -152,50 +132,47 @@ class DirectEventL0(GlowsL0):
...
+        """Convert from string to bytearray if DE_DATA is a string of ones and zeros."""
+        if isinstance(self.DE_DATA, str):
+            # Convert string output from space_packet_parser to bytearray
+            self.DE_DATA = bytearray(
+                int(self.DE_DATA, 2).to_bytes(len(self.DE_DATA) // 8, "big")
+            )
 
-    def __init__(self, packet, software_version: str, packet_file_name: str):
-        """Initialize data class with a packet of direct event data.
+    def sequence_match_check(self, de_l0):
```
maxinelasp: This is used for combining direct events in L1A

imap_processing/glows/l0/decom_glows.py
```diff
@@ -64,11 +65,28 @@ def decom_packets(
                 apid = packet.header["PKT_APID"].derived_value
                 # Do something with the packet data
                 if apid == GlowsParams.HIST_APID.value:
-                    hist_l0 = HistogramL0(packet, version, filename)
+                    values = [
+                        item.derived_value
+                        if item.derived_value is not None
+                        else item.raw_value
```
greglucas, line 73: This is a pretty common pattern that @tech3371 was using as well. Is this something we should add a helper method for? `values = utils.extract_packet_values(packet.data)` or is there some way we can add a property to the `Item.useful_value` which would do this deferral/choice for us?

imap_processing/glows/l0/decom_glows.py
```diff
@@ -64,11 +65,28 @@ def decom_packets(
...
+                        for item in packet.data.values()
+                    ]
+                    hist_l0 = HistogramL0(
+                        version, filename, CcsdsData(packet.header), *values
```
greglucas, line 77: Does this count on your HistogramL0 class attributed being defined in the same order as the XTCE definition? I'm wondering if that is pretty static or if we need to worry about a potential mismatch happening in the future and not realizing it.

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -110,37 +110,17 @@ class HistogramL0(GlowsL0):
...
-                raise KeyError(
-                    f"Did not find matching attribute in Histogram data class for "
-                    f"{key}"
+    HISTOGRAM_DATA: bytearray
+
+    def __post_init__(self):
+        """Convert HISTOGRAM_DATA attribute from string to bytearray if needed."""
+        if isinstance(self.HISTOGRAM_DATA, str):
+            # Convert string output from space_packet_parser to bytearray
+            self.HISTOGRAM_DATA = bytearray(
```
greglucas, line 121: Do you want to go straight to a numpy array from the bytes buffer similar to SWE?
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/c20b1e897d384ac669e82a6febacd9d954a2fdab/imap_processing/swe/l1a/swe_science.py#L114-L120

imap_processing/glows/l0/glows_l0_data.py
greglucas: Possibly a different name that would be more descriptive? `def within_same_sequence(self, other)`. (other might be more helpful here since that is usually how `__eq__(self, other)` appears too, but not a big deal either way)

imap_processing/glows/tests/test_glows_l1a_data.py
```diff
@@ -0,0 +1,220 @@
...
+from imap_processing.glows.l0 import decom_glows
+from imap_processing.glows.l1.glows_l1a_data import DirectEventL1A, HistogramL1A
+from imap_processing.glows.utils.constants import DirectEvent, TimeTuple
+
+
+@pytest.fixture()
+def decom_test_data():
+    """Read test data from file"""
+    current_directory = Path(__file__).parent
+    packet_path = f"{current_directory}/glows_test_packet_20230920_v00.pkts"
```
greglucas: I think this will help our friends on Windows systems
```suggestion
    packet_path = current_directory / "glows_test_packet_20230920_v00.pkts"
```

imap_processing/glows/tests/test_glows_l1a_data.py
```diff
@@ -0,0 +1,220 @@
...
+    de0 = decom_test_data[1][0]
+    de0_first = dataclasses.replace(de0, LEN=2, SEQ=0, DE_DATA=de0.DE_DATA[:100])
+    de0_second = dataclasses.replace(de0, LEN=2, SEQ=1, DE_DATA=de0.DE_DATA[100:])
+
+    expected = DirectEventL1A(de0)
+
+    de1 = DirectEventL1A(de0_first)
+    assert not de1.direct_events
+    if not de1.direct_events:
+        de1.append(de0_second)
```
greglucas, line 249: Since you are asserting right above this I think you're guaranteed to go down this path?
```suggestion
    de1.append(de0_second)
```

imap_processing/glows/tests/test_glows_l1a_data.py
```diff
@@ -0,0 +1,220 @@
...
+
+    assert de1.direct_events
+
+    # l0 attribute should not be equal, but everything else should be
```
greglucas, line 253: What is the l0 attribute? I'm guessing you mean you're comparing all other attributes here, but the comment makes me think there should also be a `assert de1.l0 != expected.l0` here that you're verifying too.

imap_processing/glows/tests/test_glows_l1a_data.py
```diff
@@ -0,0 +1,220 @@
...
+    de0 = decom_test_data[1][0]
+    de0_first = dataclasses.replace(de0, LEN=3, SEQ=0, DE_DATA=de0.DE_DATA[:100])
+    de0_second = dataclasses.replace(de0, LEN=3, SEQ=2, DE_DATA=de0.DE_DATA[100:])
+
+    expected = DirectEventL1A(de0)
+
+    de1 = DirectEventL1A(de0_first)
+    assert not de1.direct_events
+    if not de1.direct_events:
+        de1.append(de0_second)
```
greglucas, line 273:
```suggestion
    de1.append(de0_second)
```

imap_processing/glows/utils/constants.py
```diff
@@ -0,0 +1,56 @@
+from collections import namedtuple
+from dataclasses import dataclass
+
+TimeTuple = namedtuple("TimeTuple", "seconds subseconds")
+"""
+Spacecraft clock time, a float divided into seconds and subseconds
+
+.. py:attribute:: seconds
```
greglucas: I was going to comment that I think you could still put the standard numpydoc attributes in here, but that might look bad per: https://github.com/numpy/numpydoc/issues/257

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
+import struct
+from dataclasses import dataclass
+from typing import ClassVar
+
+from imap_processing.glows import version
+from imap_processing.glows.l0.glows_l0_data import DirectEventL0, HistogramL0
+from imap_processing.glows.utils.constants import DirectEvent, GlowsConstants, TimeTuple
+
+
+class DataEverySecond:
```
greglucas: Would it make sense to just have this as a global constant lookup table without needing to be a class and access the attribute?

```
DATA_EVERY_SECOND_MAPPING = {...}
```

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+        # TODO: This sanity check should probably exist in the final code. However,
+        # the emulator code does not properly set these values.
+        # if self.l0.ENDID != self.l0.SPINS:
+        #     raise ValueError(f"Inconsistency between L0 spin-numbering field ENDID "
+        #                      f"[{self.l0.ENDID}] and histogram parameter field SPINS "
+        #                      f"[{self.l0.SPINS}]")
+
+        # Create time tuples based on second and subsecond pairs
+        self.imap_start_time = TimeTuple(self.l0.SEC, self.l0.SUBSEC)
+        self.imap_end_time_offset = TimeTuple(self.l0.OFFSETSEC, self.l0.OFFSETSUBSEC)
```
greglucas: Should this be `imap_start_time_offset`, i.e. you are offsetting from the start interval right above this? Or is end time referring to end of a collection interval? Another possibility would be to remove start/end from the variables too if this is just a singular "time" and not a range.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+            self.missing_seq.extend(range(self.most_recent_seq + 1, second_l0.SEQ))
+
+        # Determine if new L0 packet matches existing L0 packet
+        match = self.l0.sequence_match_check(second_l0)
+
+        # TODO: Should this raise an error? Log? something else?
+        if not match:
+            raise ValueError(
+                f"While attempting to merge L0 packet {second_l0} "
+                f"into L1A packet {self.__repr__()}, mismatched values "
```
greglucas:
```suggestion
                f"into L1A packet {self!r}, mismatched values "
```

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+            elif (
+                marker == 0x3 and i < len(direct_events) - 2
+            ):  # 3-byte compression of timedelta
+                current_event = self._build_compressed_event(
+                    direct_events[i : i + 3], oldest_diff, current_event.timestamp
+                )
+                i += 3
+
+            else:  # wrong-marker or hitting-the-buffer-end case
+                raise Exception(
```
greglucas: Raise a `ValueError` or `IndexError` rather than a bare Exception so you can catch it in a little more detail elsewhere. Or think about making your own if that'd help too `GlowsError(Exception)`

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+                    "error: unexpected marker %d or end of "
+                    "data stream %d %d" % (marker, i, len(direct_events))
+                )
+
+            # sanity check
+            assert current_event.timestamp.subseconds < GlowsConstants.SUBSECOND_LIMIT
```
greglucas: Since we are doing check/raises elsewhere do we want to continue that here as well?
```python
if not ...
   raise ValueError("Subseconds received are less than the GLOWS subsecond limit")
```

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+            self._process_de_data()
+
+    def _process_de_data(self):
+        """
+        Process direct event bytes.
+
+        Once the packets are complete, create the status data table from the first 40
+        bytes in de_data, and the direct events from the remaining bytes.
+        """
+        self.status_data = self._generate_status_data(self.de_data[:40])
```
greglucas: Another possible idea here would be to turn `self.status_data` into a property and move the content of these functions up there into the property function definition. Fine as you have it, so just an idea to consider if you like.

greglucas (commented):
Looks great! All of my comments are minor and just for you to consider and not necessary to take.

imap_processing/glows/l0/decom_glows.py
maxinelasp: Yes, it does depend on that. Since the XTCE definition is sourced from the actual packet structure, I assumed that wouldn't be changing much. If it does change, I (or someone) would need to rearrange these attributes or explicitly define all of the parameters. However, that should be caught by the tests which check the attributes.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: That's a good point, I left it as a bytearray because this class is supposed to be the raw Level 0 data, so it's the closest type representation to the decommed data. It gets converted into a list in L1A (although it probably should be a numpy array there).

This is something I debated, I don't think it will end up mattering much where that conversion happens, so I can put it here if you think it makes more sense.

imap_processing/glows/tests/test_glows_l1a_data.py
maxinelasp: Every time 😆 why does windows do this to me?!

imap_processing/glows/l1/glows_l1a_data.py
maxinelasp: This name is to match what is given in the algorithm document. It does describe a start time and an offset. I do think `imap_time_offset` is probably a more clear name, though, thanks for pointing out the confusion

imap_processing/glows/l0/decom_glows.py
greglucas: > However, that should be caught by the tests which check the attributes.

That is a good point that you have tests which assert on specific values, so if those get bumped one further along then that should raise. Sounds good to me.

imap_processing/glows/l0/glows_l0_data.py
greglucas: I don't have a major preference, probably +0.1 on numpy. It was just that this is the first use of bytearray I think, so the consistency caught my attention. In general, I'd say if we already have a numpy dependency then it is nice to use numpy arrays, the bytearray is nice if you don't want to add numpy to your project's stack and stick with pure Python IMO.

imap_processing/glows/l0/decom_glows.py
maxinelasp: Personally I think a helper method for this is kind of overkill, especially since this is only in the decom steps. Presumably we could add something to space_packet_parser if we wanted, though

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -152,50 +132,61 @@ class DirectEventL0(GlowsL0):
...
+                int(self.DE_DATA, 2).to_bytes(len(self.DE_DATA) // 8, "big")
+            )
+
+    def sequence_match_check(self, de_l0):
+        """
+        Compare fields for L0 which should be the same for packets within one sequence.
 
-    def __init__(self, packet, software_version: str, packet_file_name: str):
-        """Initialize data class with a packet of direct event data.
+        This method compares the IMAP time (MET) and packet length (LEN) fields.
```
tech3371, line 177: What is the reason for comparing MET and LEN? Does this mean that there are other packets with same MET and LEN. If they match, then those data belong together to create one full data?

imap_processing/glows/l0/decom_glows.py
```diff
@@ -64,11 +65,28 @@ def decom_packets(
...
                 # Do something with the packet data
                 if apid == GlowsParams.HIST_APID.value:
-                    hist_l0 = HistogramL0(packet, version, filename)
+                    values = [
+                        item.derived_value
+                        if item.derived_value is not None
+                        else item.raw_value
+                        for item in packet.data.values()
+                    ]
+                    hist_l0 = HistogramL0(
```
tech3371, line 76: I think I am starting to understand the dataclass after working with CDF attrs. So here, you are creating instance of HistogramL0 to store current packet's data and then add that to the list. So the list of unpacked data is list of instances of HistogramL0. same for other apid. Just thinking out loud here.

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -110,37 +110,17 @@ class HistogramL0(GlowsL0):
...
-            )
-            if key in attributes:
-                setattr(self, key, value)
-            else:
-                raise KeyError(
-                    f"Did not find matching attribute in Histogram data class for "
-                    f"{key}"
+    HISTOGRAM_DATA: bytearray
+
+    def __post_init__(self):
```
tech3371, line 117: Does this `__post_init__` runs automatically and process unpack data? I haven't used `__post_init__` and it's good to know.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+from dataclasses import dataclass
+from typing import ClassVar
+
+from imap_processing.glows import version
+from imap_processing.glows.l0.glows_l0_data import DirectEventL0, HistogramL0
+from imap_processing.glows.utils.constants import DirectEvent, GlowsConstants, TimeTuple
+
+
+class DataEverySecond:
+    """Mapping of name to byte for dataeverysecond structure in GLOWS L1A.
```
tech3371: can you expand on dataeverysecond?

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+    compressed structure in the first 40 bytes of each direct event data field.
+
+    Attributes
+    ----------
+    mapping: dict
+        Mapping of value name to the number of bytes in the compressed data
+
+    """
+
+    mapping: ClassVar[dict] = {
```
tech3371: Is it missing one Byte? In the doc string, it say 40 bytes but I counted 39. Is that expected?

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+        "is_pps_autogenerated": 1,
+        "hv_test_in_progress": 1,
+        "pulse_test_in_progress": 1,
+        "memory_error_detected": 1,
+    }
+
+
+@dataclass
+class HistogramL1A:
+    """Data structure for GLOWS Histogram Level 1A data.
```
tech3371, line 138: btw, thank you for great documentation! Makes it easy to follow algorithm that we have not read through.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+
+@dataclass
+class DirectEventL1A:
+    """Data structure for GLOWS Histogram Level 1A data.
+
+    This includes steps for merging multiple Direct Event packets into one class,
+    so this class may span multiple packets. This is determined by the SEQ and LEN,
+    by each packet having an incremental SEQ until LEN number of packets.
```
tech3371, line 221: This answered my above question.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,411 @@
...
+        second_l0: DirectEventL0
+            Additional L0 packet to add to the DirectEventL1A class
+        """
+        # if SEQ is missing or if the sequence is out of order, do not continue.
+        if not second_l0.SEQ or second_l0.SEQ < self.most_recent_seq:
+            raise ValueError(
+                f"Sequence for direct event L1A is out of order or "
+                f"incorrect. Attempted to append sequence counter "
+                f"{second_l0.SEQ} after {self.most_recent_seq}."
+            )
```
tech3371, line 299: Will this raise cause all l1a process to stop? For example, if you have 10 packets and each 5 packets has same SEQ and LEN. You would process first 5 and say this raise happens, does it stop next 5 packets from processing?

If it doesn't stop, then I think we are ok. Otherwise, debug log or something might be helpful.

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -71,3 +72,56 @@ def test_bad_header():
...
+
+
+def test_bytearrays(decom_test_data):
+    for hist_test_data in decom_test_data[0]:
+        assert isinstance(hist_test_data.HISTOGRAM_DATA, bytearray)
+
+    for de_test_data in decom_test_data[1]:
+        assert isinstance(de_test_data.DE_DATA, bytearray)
+
+    # print(decom_test_data[0][0].HISTOGRAM_DATA[:32].hex())
```
tech3371, line 113: remove?

imap_processing/glows/tests/test_glows_l1a_data.py
```diff
@@ -0,0 +1,220 @@
...
+@pytest.fixture()
+def decom_test_data():
+    """Read test data from file"""
+    current_directory = Path(__file__).parent
+    packet_path = f"{current_directory}/glows_test_packet_20230920_v00.pkts"
+    data_packet_list = decom_glows.decom_packets(packet_path)
+    return data_packet_list
+
+
+def test_histogram_list(decom_test_data):
```
tech3371: doc string?

tech3371 (commented):
Really nice work! I had minor comments but looks great!

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: Yes, exactly. The direct events data can be split across multiple packets, so this method checks that the packets actually match. I will rename it (like Greg suggested) and update the comment to be a bit more clear.

imap_processing/glows/l0/decom_glows.py
maxinelasp: Exactly right. That way there is a check here for missing or misaligned attributes, so if for whatever reason the packet is misformed it will break right away. Then, I can pass around the dataclass instance without worrying that, for example, it's a dictionary that can have items accidentally deleted out of it. I can also easily access attributes by just typing `l0.HISTOGRAM_DATA`, so I don't need to remember or worry about ordering in a dictionary. It also tracks the expected data type, and of course, I can put any level-specific methods right into each instance, although in this case it's not super useful, and of course for L0 I don't really have many.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: It runs after the automatically generated `__init__()` function created by the `@dataclass` decorator. If I wrote a manual `__init__()` function or overrode that, it does NOT automatically run (although I can call it.)

imap_processing/glows/l1/glows_l1a_data.py
maxinelasp: Sure, I can write a better comment here. I'm not really sure of the best way to describe  it, but I will try and do better here 😆

imap_processing/glows/l1/glows_l1a_data.py
maxinelasp: Great catch, and yes, there is one byte of padding. I will explicitly note that somewhere in the code.

imap_processing/glows/l1/glows_l1a_data.py
maxinelasp: It is my understanding that this shouldn't happen, which is why I raise an error. I think this is something that should be handled by the actual level 1A processing code (which takes in a list of L0 events and produces a list of L1A events) which I haven't written yet, because I think it matches better with the CDF generation and I wanted to get this code reviewed first. 

That being said, this is something I need to ask the GLOWS team about and get a real answer. Possibly I will need to instead track the expected number of packets (which I have) and check off when I receive packets, which would make it so that packets arriving out of order isn't a problem.

imap_processing/glows/tests/test_glows_l1a_data.py
maxinelasp: Do tests need docstrings?

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: That automatic `__init()__` function just contains every attribute as a variable. There is more info on it at the top of [this page](https://docs.python.org/3/library/dataclasses.html).

imap_processing/glows/tests/test_glows_l1a_data.py
tech3371: I think one line comment about what it is testing might be helpful.

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -152,50 +132,61 @@ class DirectEventL0(GlowsL0):
...
     LEN: int
     SEQ: int
+    DE_DATA: bytearray
+
+    def __post_init__(self):
+        """Convert from string to bytearray if DE_DATA is a string of ones and zeros."""
+        if isinstance(self.DE_DATA, str):
+            # Convert string output from space_packet_parser to bytearray
+            self.DE_DATA = bytearray(
+                int(self.DE_DATA, 2).to_bytes(len(self.DE_DATA) // 8, "big")
```
sdhoyt, line 170: Is the byte order part something that GLOWS told you?

sdhoyt: Also, a few more dumb questions about this:

1. When you convert the DE_DATA string to an int, would it cause problems if the first character in the string is a 0 since it'll drop the 0 when it converts to an int?
2. Is the bytearray basically the binary string, grouped into 1 byte segments?

sdhoyt (commented):
This looks great! Just left a few questions.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: Not a dumb question! 
1. Since it's going from a byte to a binary array (the toint with a 2 indicates binary) it shouldn't cause issues. It's not going from string to int to binary, just going straight from string to binary. 
2. Yes, the bytearray is an array of bytes from the string, so if you index into it you get a byte back (usually in the form of hex)

maxinelasp: And yes, GLOWS specified big endian in their docs, although it was something I messed up at first. It also needs to match the XTCE document.

tech3371 (approved):

sdhoyt (approved):

imap_processing/glows/l0/decom_glows.py
```diff
@@ -64,11 +65,28 @@ def decom_packets(
                 apid = packet.header["PKT_APID"].derived_value
                 # Do something with the packet data
                 if apid == GlowsParams.HIST_APID.value:
-                    hist_l0 = HistogramL0(packet, version, filename)
+                    values = [
```
mstrumik, line 70: Few lines above there is `buffer_read_size_bytes=5790778` with the size 5790778B of the test input file hardcoded. My understanding is that it is only temporarily hardcoded, because generally GLOWS telemetry dumps will have variable length.

imap_processing/glows/utils/constants.py
```diff
@@ -0,0 +1,67 @@
...
+        Subseconds of clock, defined as 1/SUB_SECOND_LIMIT th of a second. Will
+        always be less than SUB_SECOND_LIMIT. If the class is initialized with a
+        subsecond value above SUB_SECOND_LIMIT, the subseconds above the limit will be
+        converted to seconds.
+    """
+
+    seconds: int
+    subseconds: int
+
+    def __post_init__(self):
```
mstrumik, line 25: My understanding is that when an instance is created then `__post_init__` is always executed and cases with `subseconds > SUB_SECOND_LIMIT` are properly handled. But someone can be tempted to do `var.subseconds = subseconds + diff` somewhere in the code after having initialized `var: TimeTuple` and then `__post_init__` method will not be executed. Presumably `@dataclass(frozen=True)` for `class TimeTuple` can solve the problem raised by me, i.e. when somebody tries to modify subseconds hastily.

imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -1,4 +1,4 @@
-from dataclasses import dataclass, fields
+from dataclasses import dataclass
 
 from imap_processing.ccsds.ccsds_data import CcsdsData
```
mstrumik, line 6: In `class GlowsL0` below there is a field  `ground_sw_version`. My understanding is that it will contain the same information as encoded in the filename in the field `vXX_YY`, where XX encodes the ground software version and YY encodes the data release. So I presume that `ground_sw_version = XX`.  It seems to be redundant (to keep it in the filename and inside the data file), but the GLOWS Team would appreciate to keep it as it is. Would it be possible also to add a field `data_release = YY` to `class GlowsL0`? We simply prefer to have this information inside the file, without the need of decoding the filename.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,468 @@
...
+    def __init__(self, level0: HistogramL0):
+        """Set the attributes based on the given L0 histogram data.
+
+        This includes generating a block header and converting the time attributes from
+        HistogramL0 into TimeTuple pairs.
+        """
+        self.l0 = level0
+        self.histograms = list(self.l0.HISTOGRAM_DATA)
+
+        self.block_header = {
```
mstrumik, line 184: Following up on my request regarding an additional `data_release` field in `class GlowsL0` in `glows_l0_data.py`, would it be possible to add such item (`data_release`) to `self.block_header` dictionary here?

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,468 @@
...
+        missing_seq. Once the last value in the sequence is reached, the data is
+        processed from raw bytes to useful information.
+
+        Parameters
+        ----------
+        second_l0: DirectEventL0
+            Additional L0 packet to add to the DirectEventL1A class
+        """
+        # if SEQ is missing or if the sequence is out of order, do not continue.
+        if not second_l0.SEQ or second_l0.SEQ < self.most_recent_seq:
```
mstrumik, line 294: As we recently discussed via email, the GLOWS Team generally prefers to save as much data as possible when the ground software detects missing packets or problems with the order of packets. Currently raising ValueError after this `if` condition means that there is no attempt of saving problematic data, but my understanding is that this work is in progress.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,468 @@
...
+    glows_time_offset: tuple[int, int]
+        GLOWS end time offset for the block, in the form (seconds, subseconds). In
+        algorithm document as "glows_end_time_offset"
+    flags: dict
+        Dictionary containing "flags_set_onboard" from L0, and "is_generated_on_ground",
+        which is set to "False" for decommed packets.
+    """
+
+    l0: HistogramL0
+    histograms: list[int]
```
mstrumik, line 166: I am not sure why `histograms` is plural, my understanding is that this variable will contain only one histogram.

imap_processing/glows/l1/glows_l1a_data.py
mstrumik: I am not sure if I understand the phrase "Mapping of name to byte".  Does it refer to `byte_attribute_mapping` dictionary in `__init__` below?

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,468 @@
...
+            # note: packet number is seq_count (per apid!) field in CCSDS header
+            "seq_count_in_pkts_file": self.l0.ccsds_header.SRC_SEQ_CTR,
+        }
+
+        # use start ID and offset to calculate the last spin ID in the block
+        self.last_spin_id = self.l0.STARTID + self.l0.ENDID
+
+        # TODO: This sanity check should probably exist in the final code. However,
+        # the emulator code does not properly set these values.
+        # if self.l0.ENDID != self.l0.SPINS:
```
mstrumik, line 197: Such sanity check `if self.l0.ENDID != self.l0.SPINS:` was suggested in the code provided by the GLOWS Team, but I realized that GLOWS is going to use its own method of determining the exact time of phase-zero transition, which can give slightly different results as compared to the method used onboard on IMAP to number the spins. Thus this sanity check can be problematic. I think e.g. `sum(self.histograms) == self.l0.EVENTS` can be checked safely at this point. Similar test to that suggested by me is already included in `tests/test_glows_l1a_data.py` in `test_histogram_list()` but executed only during tests for the first histogram.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,468 @@
...
+        self.status_data = StatusData(self.de_data[:40])
+        self.direct_events = self._generate_direct_events(self.de_data[40:])
+
+    def _generate_direct_events(self, direct_events: bytearray):
+        """Generate the list of direct events from the raw bytearray.
+
+        First, the starting timestamp is created from the first 8 bytes in the direct
+        event array. Then, the remaining events are processed based on a marker in the
+        first two bits of each section. If the marker is 0, it is uncompressed, and
+        the event is processed from the following 7 bytes. If it is 1, it is compressed
```
mstrumik, line 340: Actually `marker=2` (not 1) means two-byte compression, and `marker=3` (not 2) means three-byte compression.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -0,0 +1,468 @@
...
+        seconds = values[0]
+
+        # subsecond encoding on the least significant 21 bits
+        subseconds = values[1] & 0x1FFFFF
+
+        timestamp = TimeTuple(seconds, subseconds)
+        # first byte encodes the impulse length
+        impulse_length = (values[1] >> 24) & 0xFF
+
+        # KPLabs says it is set by FPGA and currently not used by AppSW at all
```
mstrumik, line 479: I just realized that "it" in this comment is ambiguous. My suggestion: "KPLabs says `multi_event` is set ..."

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -71,3 +72,56 @@ def test_bad_header():
     bad_data["test"].derived_value = "test"
     with pytest.raises(KeyError, match="Did not find matching"):
         CcsdsData(bad_data)
+
```
mstrumik, line 75: In the line containing `"SPINS": 7,` above, the value should be 8. After reading `SPINS` from the CCSDS packet, its value should be increased by 1. Please see the code provided by the GLOWS Team, line 43 in `toolkit/l0_data_histogram.py`.

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -11,8 +12,8 @@
 def decom_test_data():
     """Read test data from file"""
     current_directory = Path(__file__).parent
-    packet_path = f"{current_directory}/glows_test_packet_20230920_v00.pkts"
-    data_packet_list = decom_glows.decom_packets(packet_path)
+    packet_path = current_directory / "glows_test_packet_20230920_v00.pkts"
+    data_packet_list = decom_glows.decom_packets(str(packet_path))
```
mstrumik, line 16: In the lines below `expected_hist_packet_count = 61` and `expected_de_packet_count = 311` the packet counters seem to contain incorrect values. For the file `imap_l0_sci_glows_20230920_v00.pcts` provided by K. Ber it should be `expected_hist_packet_count = 505` and `expected_de_packet_count = 1088`. It looks like some packets have been omitted.

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -71,3 +72,56 @@ def test_bad_header():
     bad_data["test"].derived_value = "test"
     with pytest.raises(KeyError, match="Did not find matching"):
         CcsdsData(bad_data)
+
+
+def test_header(decom_test_data):
+    expected_hist = CcsdsData(
+        {
+            "VERSION": ParsedDataItem("VERSION", 0, unit=None),
+            "TYPE": ParsedDataItem("TYPE", 0, unit=None),
```
mstrumik, line 81: I am not sure why `TYPE` is 0 here. I think it should be 1.

imap_processing/glows/tests/test_glows_decom.py
```diff
@@ -71,3 +72,56 @@ def test_bad_header():
...
+            "PKT_LEN": ParsedDataItem("PKT_LEN", 3663, unit=None),
+        }
+    )
+
+    assert expected_hist == decom_test_data[0][0].ccsds_header
+
+    expected_de = CcsdsData(
+        {
+            "VERSION": ParsedDataItem("VERSION", 0, unit=None),
+            "TYPE": ParsedDataItem("TYPE", 0, unit=None),
```
mstrumik, line 94: Similarly to my comment on `expected_hist` above. I am not sure why `TYPE` is 0 here. I think it should be 1.

imap_processing/glows/tests/test_glows_l1a_data.py
```diff
@@ -0,0 +1,313 @@
...
+    assert hist.glows_time_offset == TimeTuple(119, 1998758)
+    assert hist.flags == {"flags_set_onboard": 64, "is_generated_on_ground": False}
+
+
+def test_direct_event_every_second(decom_test_data):
+    """Test that Level 1A direct events generate correct status data table"""
+    de0 = decom_test_data[1][0]
+
+    # Manually created dictionary from first test packet
+    data_every_second = {
```
mstrumik: Perhaps `expected_data_every_second`?

imap_processing/glows/tests/test_glows_l1a_data.py
```diff
@@ -0,0 +1,313 @@
...
+    )
+    # Required for processing
+    previous_time = TimeTuple(54232337, 1997808)
+
+    oldest_diff = first_byte & 0x3F
+
+    output = de1a._build_compressed_event(two_bytes, oldest_diff, previous_time)
+
+    final_expected = DirectEvent(
+        TimeTuple(seconds=54232338, subseconds=1988562),
```
mstrumik: It should be `TimeTuple(seconds=54232338, subseconds=1993605)` here. It looks like two last timestamps for direct events in the first packet have been omitted.

mstrumik (commented):
I am sending some comments. I think the tests for decom and l1a are quite comprehensive. However, currently they are only done for the first histogram and direct-event packets in the file `glows_test_packet_20230920_v00.pkts`. To make sure that entire file was read properly without any offsets, I would suggest doing the tests also the last histogram and direct-event packets in the input file.

imap_processing/glows/l0/decom_glows.py
maxinelasp: Yes, this is temporary, I need to look into what the proper value is for that number.

imap_processing/glows/utils/constants.py
maxinelasp: This is a great point, I will make that change! Thanks!

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: I moved the ground software version into `glows/__init__.py` so it could be imported as necessary. The version will be inside the block header, as requested in the algorithm document. The L0 data classes are only intended for internal use, so they don't store the ground software version, but that version will be inside all the CDF files, as well as in the filename. 

It is possible to add a data release field to the CDF files, but since I won't know that version until run time, I will add it in when writing the CDF files out. So, the output will have the data release as part of the `block_header` field. That will get added in another code review once I have the CDF file output.

imap_processing/glows/l1/glows_l1a_data.py
maxinelasp: See my note about `data_release` in your other comment. Short answer, yes, I can include `data_release` in the CDF files for each level.

imap_processing/glows/l1/glows_l1a_data.py
maxinelasp: Yes, this is a work in progress, I will add in the changes that came from our discussion in a later code review. That will involve re-writing how the `SEQ` field is used so we can process data if possible and add flags on any unprocessable data.

imap_processing/glows/tests/test_glows_decom.py
maxinelasp: Good catch. I shortened the packet file, since I didn't need something that large for unit testing. Therefore, the test file used here only contains part of the packet file which was provided.

imap_processing/glows/tests/test_glows_decom.py
maxinelasp: I will investigate this, thank you!

imap_processing/glows/tests/test_glows_l1a_data.py
maxinelasp: Thank you, I will check my tests. I will also add some additional tests for the last packet in the sequence, like you suggested.

imap_processing/glows/tests/test_glows_decom.py
maxinelasp: I think 0 is correct. In your `ccsds.py` file, line 30, you have the following: 
`self.data['pkt_type'] = bool(bit_seq_as_string[3])`

If I change this to `self.data['pkt_type'] = int(bit_seq_as_string[3])`, the third bit in the CCSDS header is 0. I think what happened is `bit_seq_as_string[3]` returns a one character string `"0"`. Converting this to a bool will always result in True. Instead, we should convert the string to an int and then that int to a bool: `self.data['pkt_type'] = bool(int(bit_seq_as_string[3]))`

This also aligns with the CCSDS spec, which defines a type of 0 to be telemetry, with 1 being a telecommand packet ([section 4.1.3.3.2.3](https://public.ccsds.org/Pubs/133x0b2e1.pdf))

mstrumik: Thank you for the explanation. Indeed, it was my mistake.

imap_processing/glows/l1/glows_l1a_data.py
maxinelasp: Sorry, I missed this comment @mstrumik. This documentation is outdated, thanks for catch it. I rewrote the class doc (again) to try and further explain data_every_second and how it is used.

imap_processing/glows/tests/test_glows_decom.py
maxinelasp: I reworked the test packet file so it is now small enough to put in github and run tests quickly, but also so it includes all the expected direct event and histogram packets. So, those numbers are now updated back to 505 and 1088.

imap_processing/glows/tests/test_glows_l1a_data.py
maxinelasp: This was a very good catch! I was indeed not reading the final couple of direct events because I was using the wrong number for the decom steps. I have fixed this issue, and added several additional tests to ensure I am reading all the data as expected, including testing the last direct events in each packet.

greglucas:
Looks like you need to add some docstrings now, otherwise is this good to go?
```
imap_processing/glows/l0/glows_l0_data.py:1:1: D100 Missing docstring in public module
imap_processing/glows/l1/glows_l1a.py:1:1: D100 Missing docstring in public module
imap_processing/glows/l1/glows_l1a_data.py:1:1: D100 Missing docstring in public module
imap_processing/glows/utils/constants.py:1:1: D100 Missing docstring in public module
```

maxinelasp:
Yes, otherwise good to go now! I'll add the docstrings now

bourque:
@all-contributors please add @mstrumik for review

PR #274, opened 2023-11-08
Added eu analog conversion to test_decom 

GFMoraga:
# Change Summary
This is a WIP draft. The new test does not pass right now, so will be fixing and updating. Just wanted to share the progress for the ticket #247 

## Overview

This PR will apply analog conversions to `validation_data` and test against the eu csv. 

## New Files

- idle_export_eu.COD_NHK_20230822_122700 2.csv
   - validation eu csv

## Updated Files

- `test_decom.py`
   - added analog conversions 
   - added function to call eu csv. 

## Testing

The new test function `def test_unit_conversion` will not pass! TODO: update and get feedback on how to pass this. 

closes #247

PR #288, opened 2023-11-28
Initial CoDICE L1A Processing Algorithm

bourque:
# Change Summary

## Overview
This PR includes an initial pass at CoDICE L1A processing. Currently it only supports/works on CoDICE housekeeping packets, and does not yet do any analog to EU conversions (that work will be in a future PR and is currently being done by @GFMoraga in #274). The code is heavily inspired by @tech3371's SWE processing code.

The PR also includes a re-organization of the CoDICE code as a whole.  Basically I moved all processing code to be directly under the `codice/` directory instead of having separate `l0/` and `l1a/` directories.  This organization makes most sense to me in my head, but I'm definitely open to other ideas. I also tried renaming some files so that they are a bit less repetitive.

Lastly, I added/updated some module and function docstrings, and updated the CoDICE API reference docs.

## New Dependencies
None

## New Files

- `/cdf/utils.py`
   - Added condition to set the `file_start_date` for CoDICE files
- `/codice/__init__.py`
-  `/codice/codice_cdf_attrs.py`
    - CDF attributes for CoDICE; basically just used what SWE had and replaced where appropriate 
- `/codice/codice_l1.py`
    - L1A processing code 
- `/codice/tests/test_codice_l1a.py`
    - Tests for L1A processing code 

## Deleted Files

None

## Updated Files

- `/docs/source/reference/codice.rst`
    - Added new L1A reference documentation and made changes to reflect the reorganization of files
- `/codice/codice_l0.py`
    - Moved from `imap_processing/codice/l0/decom_codice.py`
- `/codice/utils/decompress.py`
    - Moved from `imap_processing/codice/l0/decompress_codice.py `
-  `/codice/tests/test_codice_l0.py`
    - Minor formatting updates 
- `/codice/utils/codice_utils.py`
    - Added some functions to support L1A processing 

## Testing

I added unit tests for L1A processing, but currently only tests on housekeeping packets.


maxinelasp (approved):
Nice, this looks great

imap_processing/codice/codice_cdf_attrs.py
```diff
@@ -0,0 +1,116 @@
...
+
+l1b_science_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=GlobalConstants.INT_MAXVAL,
+    display_type="spectrogram",
+    depend_0="Epoch",
+    depend_1="Energy",
+    depend_2="Angle",
+    depend_3="Rates",
+    format="I12",
```
GFMoraga: Is this the correct format? Just because formatting has been case-sensitive in past work, so I wanted to highlight this default you set.

imap_processing/codice/utils/codice_utils.py
```diff
@@ -1,5 +1,57 @@
...
+import xarray as xr
+
+from imap_processing.cdf.global_attrs import ConstantCoordinates
+from imap_processing.codice import codice_cdf_attrs
+
+# Placeholder for the list of mnemonics to convert from analog to engineering units
+CONVERT_TO_EU_LIST = []
+
+
+class CODICEAPID(IntEnum):
```
GFMoraga: Will all of these be needed? Nice job!

GFMoraga (approved):
This looks great! I will have to re-base but I think all of your updates and naming conventions are great!

imap_processing/codice/utils/codice_utils.py
bourque: Probably not -- I remember Joey mentioning that in reality we wont get some of these, but it was easy enough to just put them all here for now.

imap_processing/codice/codice_cdf_attrs.py
bourque: Good question, I'm not sure 😅 . I basically just copied this from what SWE has.

bourque:
@GFMoraga Yeah, this will likely cause you some merge conflicts, but I can help resolve those since they will be my fault!

imap_processing/cdf/utils.py
```diff
@@ -63,7 +63,10 @@ def write_cdf(
     file_start_date = None
     if "idex" in data.attrs["Logical_source"]:
         file_start_date = data["Epoch"][0].data
-    elif "swe" in data.attrs["Logical_source"]:
+    elif (
```
greglucas: Is there a good reason we are inconsistent here, or can we somehow make this consistent across the instruments?
Possibly an `else:` for now though?

imap_processing/codice/codice_cdf_attrs.py
```diff
@@ -0,0 +1,116 @@
...
+    GlobalInstrumentAttrs,
+    ScienceAttrs,
+    StringAttrs,
+)
+from imap_processing.codice import __version__
+
+descriptor = "CoDICE>Compact Dual Ion Composition Experiment"
+codice_description_text = (
+    "The Compact Dual Ion Composition Experiment (CoDICE) will measure the "
+    "distributions and composition of interstellar pickup ions(PUIs), "
```
greglucas:
```suggestion
    "distributions and composition of interstellar pickup ions (PUIs), "
```

imap_processing/codice/codice_cdf_attrs.py
```diff
@@ -0,0 +1,116 @@
...
+    "particles that make it through the heliosheath into the heliosphere. "
+    "CoDICE also collects and characterizes solar wind ions including the "
+    "mass and composition of highly energized particles (called suprathermal) "
+    "from the Sun.CoDICE combines an electrostatic analyzer(ESA) with a "
```
greglucas:
```suggestion
    "from the Sun. CoDICE combines an electrostatic analyzer (ESA) with a "
```

imap_processing/codice/codice_cdf_attrs.py
```diff
@@ -0,0 +1,116 @@
...
+    "Time-Of-Flight versus Energy (TOF / E) subsystem to simultaneously  "
+    "measure the velocity, arrival direction, ionic charge state, and mass of "
+    "specific species of ions in the LISM.CoDICE also has a path for higher "
```
greglucas:
```suggestion
    "specific species of ions in the LISM. CoDICE also has a path for higher "
```

imap_processing/codice/codice_cdf_attrs.py
```diff
@@ -0,0 +1,116 @@
...
+    "energy particles to skip the ESA but still get measured by the common "
+    "TOF / E systm.These measurements are critical in determining the Local "
```
greglucas:
```suggestion
    "TOF / E system. These measurements are critical in determining the Local "
```

imap_processing/codice/codice_cdf_attrs.py
```diff
@@ -0,0 +1,116 @@
...
+    data_type="L1B->Level-1B",
+    logical_source="imap_cpdice_l1b",
+    logical_source_desc="IMAP Mission CoDICE Instrument Level-1B Data",
+    instrument_base=codice_base,
+)
+
+int_base = AttrBase(
+    validmin=0,
+    validmax=GlobalConstants.INT_MAXVAL,
+    format="I12",
```
greglucas: Out of curiosity, do we care that the maxval is a 64-bit signed integer, but we have only 12 digits of representation in the format?
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/c20b1e897d384ac669e82a6febacd9d954a2fdab/imap_processing/cdf/defaults.py#L30

imap_processing/codice/codice_l1a.py
```diff
@@ -0,0 +1,53 @@
...
+        The path to the CDF file that was created
+    """
+    # Group data by APID and sort by time
+    grouped_data = group_by_apid(packets)
+
+    for apid in grouped_data.keys():
+        if apid == CODICEAPID.COD_NHK:
+            sorted_packets = sort_by_time(grouped_data[apid], "SHCOARSE")
+            data = create_dataset(packets=sorted_packets)
+        else:
```
greglucas, line 48: Would it be helpful to log what apids you're skipping?

imap_processing/codice/tests/test_codice_l1a.py
```diff
@@ -0,0 +1,33 @@
...
+def test_codice_l1a(l0_test_data):
+    """Tests the ``codice_l1a`` function and ensured that a proper CDF file
+    was created"""
+
+    cdf_filename = codice_l1a(l0_test_data, "")
+
+    assert Path(cdf_filename).name == "imap_codice_l1a_hk_20100101_v01.cdf"
+
+    # Remove the test data
+    Path(cdf_filename).unlink()
```
greglucas: We should use the temporary test directories so this gets cleaned up for us. If your assertion fails it won't hit this unlinking and leave extra files around. You can use a fixture and then pass that to write_cdf() I think. (IDEX does something similar, but I'm not sure you need/want the full factory)

imap_processing/codice/utils/codice_utils.py
```diff
@@ -15,3 +67,77 @@ class CoDICECompression(IntEnum):
...
+        metadata_arrays.setdefault(key, []).append(value.raw_value)
+
+    for key, value in packet.data.items():
+        if key in CONVERT_TO_EU_LIST:
+            # TODO: Convert value to engineering units
+            # For now, just use the raw value
+            converted_value = value.raw_value
+            metadata_arrays.setdefault(key, []).append(converted_value)
+        else:
+            metadata_arrays.setdefault(key, []).append(value.raw_value)
```
greglucas: You might want to bring at least the `metadata_arrays.setdefault(key, []).append(...)` block outside of the if blocks since that is essentially repeated and only update your value of interest within the choice.
```suggestion
        # TODO: Convert value to engineering units
        # For now, just use the raw value
        converted_value = value.raw_value if key in CONVERT_TO_EU_LIST else value.raw_value
        metadata_arrays.setdefault(key, []).append(converted_value)
```

greglucas (commented):
Nice, looks great! Just minor comments from me, but would prefer the test cleanup one to be added if possible.

I also personally like this structure better, I find the multiple nested modules a little harder to follow. If you're working on naming I might even suggest dropping the `codice` from the filenames within your `codice` module since that is repetitive. Something like this with a flatter instrument structure. (But, that is just my personal preference, no need to take it if you disagree)

```
codice/
    cdf_attrs.py
    level0.py
    level1a.py
    decommutation.py
    utils.py
````

File: imap_processing/cdf/utils.py (hunk 3 above)
Comment (bourque) [reply imap_processing/cdf/utils.py]: Good question, I am not sure of the answer. Perhaps @tech3371 and/or @bryan-harter has some insight here?

File: imap_processing/codice/codice_l1a.py (hunk 9 above)
Comment (bourque) [reply imap_processing/codice/codice_l1a.py:48]: My next PR where I try to add support for more APIDs should address this, but for now I can add a message instead of just a `pass`

File: imap_processing/codice/tests/test_codice_l1a.py (hunk 10 above)
Comment (bourque) [reply imap_processing/codice/tests/test_codice_l1a.py]: Neat! Yes this is much cleaner.

**Comment by bourque** [comment]
I can't for the life of me figure out why the doc build is failing. There are tons of WARNINGs (which is its own problem) but I don't think those should cause the entire build to fail. I don't see any actual ERRORs.

**Comment by maxinelasp** [comment]
> I can't for the life of me figure out why the doc build is failing. There are tons of WARNINGs (which is its own problem) but I don't think those should cause the entire build to fail. I don't see any actual ERRORs.

I think warnings do cause the check to fail. You can try updating the config.py file like what I have in my currently open glows PR, and I think you will also need a `Methods` block in your class docs. I have a docfix commit on glows l1a which fixes very similar errors from my doc warning, so it might help you to look there.

**Comment by bourque** [comment]
Thanks @maxinelasp!

**Review by greglucas — APPROVED** [review]

File: imap_processing/cdf/utils.py (hunk 3 above)
Comment (tech3371) [reply imap_processing/cdf/utils.py]: I think `else:` is a good suggestion. If we can make it consistent across, it would be nice. At least the Epoch part.

File: imap_processing/codice/utils.py [hunk 12]
```diff
@@ -0,0 +1,151 @@
...
+        Updated metadata arrays with values
+    """
+    for key, value in packet.header.items():
+        metadata_arrays.setdefault(key, []).append(value.raw_value)
+
+    for key, value in packet.data.items():
+        if key in CONVERT_TO_EU_LIST:
+            # TODO: Convert value to engineering units
+            # For now, just use the raw value
+            final_value = value.raw_value
```
tech3371: concert raw to EU will be done outside of this function. These data will be stored as array in l1a process, then in l1b, you would apply conversion to array as whole.

imap_processing/codice/codice_l1a.py
```diff
@@ -0,0 +1,59 @@
...
+import space_packet_parser
+
+from imap_processing.cdf.utils import write_cdf
+from imap_processing.codice.utils import CODICEAPID, create_dataset
+from imap_processing.utils import group_by_apid, sort_by_time
+
+
+def codice_l1a(
+    packets: list[space_packet_parser.parser.Packet], cdf_directory: str
+) -> str:
```
tech3371, line 26: Perfect! I was going to suggest to include path parameter when I skimmed through your PR before.

tech3371 (commented):
Nice work!

tech3371 (approved):

PR #289, opened 2023-12-06
Mag L0/Decom work

maxinelasp:
# Change Summary
Now that I have some testing data, I was able to go back and finish the level 0 work for MAG. 

## Overview

Main focus is on the Level 0 dataclass and on the combined XTCE. Since the only difference between BURST data and NORM data is the length of the packet, I can use the same XTCE file to process both packet types (with the variable length field just changing in size depending on the packet type.)

This shared information also means that I can use the same level 0 dataclass for both types. Later processing will have them processed differently.

## New Dependencies

None

## New Files

- MAG_SCI_COMBINED.xml
   - XTCE for both MAG packets
- decom_mag.py
   - Decom raw packets into a list of L0 data
- mag_l0_data.py
    - dataclass for raw MAG packet data 

## Deleted Files

- deleted file 1

## Updated Files

- updated file 1
- updated file 2

## Testing

Testing is created for L0 using provided packet information from the MAG team (raw data mapping to expected values). Most of the testing uses those files, with some testing on a multiple-packet file I received from the POC.


imap_processing/mag/l0/decom_mag.py
```diff
@@ -0,0 +1,65 @@
...
+                        item.derived_value
+                        if item.derived_value is not None
+                        else item.raw_value
+                        for item in packet.data.values()
+                    ]
+                    print(packet.header)
+                    data_list.append(MagL0(CcsdsData(packet.header), *values))
+        except ReadError as e:
+            print(e)
+            print("This may mean reaching the end of an incomplete packet.")
```
tech3371: change to logger?

imap_processing/mag/l0/decom_mag.py
```diff
@@ -0,0 +1,65 @@
...
+
+        return data_list
+
+
+if __name__ == "__main__":
+    packet_name = "PREFLIGHT_raw_record_2023_313_16_41_02_apid1052.pkts"
+
+    output = decom_packets(packet_name)
+    print(len(output))
```
tech3371: remove?

imap_processing/mag/tests/test_mag_decom.py
```diff
@@ -0,0 +1,78 @@
...
+
+from imap_processing.mag.l0.decom_mag import decom_packets
+
+
+def test_burst_decom():
+    current_directory = Path(__file__).parent
+    burst_test_file = current_directory / "mag_single_burst.pkts"
+    print(current_directory)
+    print(burst_test_file)
+    print("Test")
```
tech3371: remove?

tools/xtce_generation/xtce_generator_mag.py
```diff
@@ -0,0 +1,44 @@
...
+
+from tools.xtce_generation.telemetry_generator import TelemetryGenerator
+
+
+def main():
+    """This function can be used by any instrument to generate XTCE
+    for certain number of packets. Change values where TODO is
+    """
+
+    # TODO: change instrument name
```
tech3371: we could remove these `TODO` since this is now MAG specific.

tech3371 (approved):
Looks good. Minor comments

imap_processing/mag/l0/decom_mag.py
```diff
@@ -0,0 +1,63 @@
...
+
+from bitstring import ReadError
+from space_packet_parser import parser, xtcedef
+
+from imap_processing import imap_module_directory
+from imap_processing.ccsds.ccsds_data import CcsdsData
+from imap_processing.mag.l0.mag_l0_data import MagL0, Mode
+
+logging.basicConfig(level=logging.ERROR)
+logger = logging.getLogger(__name__)
```
greglucas:
```suggestion
logger = logging.getLogger(__name__)
# logger.setLevel(logging.ERROR)
```
warning level is default, so we'll get the ERRORs still and I think we want them as well? So, I'd suggest either removing the level, or setting it directly on this logger itself.

imap_processing/mag/l0/decom_mag.py
```diff
@@ -0,0 +1,63 @@
...
+    packet_definition = xtcedef.XtcePacketDefinition(xtce_document)
+    mag_parser = parser.PacketParser(packet_definition)
+
+    data_list = []
+
+    with open(packet_file_path, "rb") as binary_data:
+        try:
+            mag_packets = mag_parser.generator(
+                binary_data,
+                buffer_read_size_bytes=5790778,  # Todo: what size?
```
greglucas, line 45: How much does this matter for the processing? Is it for speed to choose the right buffer size, or is it critical to set it to a good value? None of the other instruments set the buffer size, so I'm curious if we should set it in them as well.

imap_processing/mag/l0/decom_mag.py
```diff
@@ -0,0 +1,63 @@
...
+                if apid in (Mode.BURST, Mode.NORM):
+                    values = [
+                        item.derived_value
+                        if item.derived_value is not None
+                        else item.raw_value
+                        for item in packet.data.values()
+                    ]
+                    data_list.append(MagL0(CcsdsData(packet.header), *values))
+        except ReadError as e:
+            logger.warning(
```
greglucas: Do you want this as error? I don't think it will be logged currently because you set the level to ERROR above. I'll still maintain I think we should not be catching these and rather failing so it is loud when something breaks...

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class Mode(IntEnum):
+    """Enum class for MAG mode.
+
+    Attributes
+    ----------
+    BURST : int
+        APID for Burst mode data
+    NORM : int
```
greglucas: Spell out the full word? NORMAL

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -0,0 +1,106 @@
...
+    PRI_COARSETM: int
+    PRI_FNTM: int
+    SEC_COARSETM: int
+    SEC_FNTM: int
+    VECTORS: bytearray
+
+    def __post_init__(self):
+        """Convert Vectors attribute from string to bytearray if needed.
+
+        Also convert encoded "VECSEC" (vectors per second) into proper vectors per
```
greglucas, line 97: Thank you for this comment :)

imap_processing/mag/packet_definitions/MAG_SCI_COMBINED.xml
```diff
@@ -0,0 +1,235 @@
...
+					<xtce:ParameterRefEntry parameterRef="PKT_APID" />
+					<xtce:ParameterRefEntry parameterRef="SEQ_FLGS" />
+					<xtce:ParameterRefEntry parameterRef="SRC_SEQ_CTR" />
+					<xtce:ParameterRefEntry parameterRef="PKT_LEN" />
+				</xtce:EntryList>
+			</xtce:SequenceContainer>
+			<xtce:SequenceContainer name="P_MAG_SCI_NORM">
+				<xtce:BaseContainer containerRef="CCSDSPacket">
+					<xtce:RestrictionCriteria>
+						<xtce:Comparison parameterRef="PKT_APID" value="1052" useCalibratedValue="false" />
```
greglucas, line 180: FYI, If they are identical I think you can add a `comparisonList` here as well to compare against both values. Fine the way you have it though.
```python
<xtce:ComparisonList>
  <xtce:Comparison parameterRef="PKT_APID" value="1052" useCalibratedValue="false" />
  <xtce:Comparison parameterRef="PKT_APID" value="1068" useCalibratedValue="false" />
</xtce:ComparisonList>
```

greglucas (approved):
Some minor comments, but overall looks good to me!

imap_processing/mag/l0/decom_mag.py
maxinelasp: This is something I noticed when decomming GLOWS, actually. If the file is larger than the default `buffer_read_size_bytes` provided by space packet parser, then it won't read the entire packet. (I think? Unless it was some other error that I was misinterpreting) I think this is going to be something we need to set properly, though. I don't know what a good value is or how to chose one (hence the todo)

imap_processing/mag/l0/decom_mag.py
maxinelasp: If there is an incomplete packet, wouldn't we want to keep reading and not halt processing?

imap_processing/mag/packet_definitions/MAG_SCI_COMBINED.xml
maxinelasp: I tried that, and a few other ways, but I wasn't able to get it to work. IIRC the comparison list is only for if ALL the comparisons are true so it never evaluates to true... and there's some weird XTCE OR comparison structure that I tried, but that didn't work either.

imap_processing/mag/l0/decom_mag.py
maxinelasp: But, you are right that it should be logged as an ERROR at least.

imap_processing/mag/l0/decom_mag.py
greglucas: I am assuming it was a different error... I think this is just the read size which may be able to speed up if you have the memory / IO capability, but I don't think it should lead to error/no-error depending on what we set this to. (from my reading of the docs at least, which may be wrong: https://space-packet-parser.readthedocs.io/en/latest/autoapi/space_packet_parser/parser/index.html)

> buffer_read_size_bytes (int, Optional) – Number of bytes to read from e.g. a BufferedReader or socket binary data source on each read attempt. Default is 4096 bytes.

imap_processing/mag/l0/decom_mag.py
greglucas: I think this is a good question. Do we expect this to happen, and if so do we want to continue on and accept a missing point, or do we want to fail early and force us to go in and look at it to fix it. I'm not sure there is a right answer, and maybe different teams have different expectations?

I am sort of anticipating not getting incorrect packets down, so what I'm worried about is us updating a variable in the XTCE or something like that and then logging errors and not catching that we actually incorrectly updated the XTCE because we didn't raise. Either way is probably not going to happen is my guess...

imap_processing/mag/packet_definitions/MAG_SCI_COMBINED.xml
greglucas: :+1: Not worth digging into then. Your version isn't too much extra duplication.

imap_processing/mag/l0/decom_mag.py
maxinelasp: See, I read that as, the number of bytes read out of the file. If the file was longer than 4096 bytes, you would need to call `generator` multiple times.

greglucas: You could try removing it (so it'd be shorter and not get the full packet size) and see if your tests still pass? Or reach out in the xtce channel too.

imap_processing/mag/l0/decom_mag.py
```diff
@@ -0,0 +1,62 @@
+import logging
```
bourque, line 3: I've been bugging everyone to add module docstrings in their reviews so I will do the same here 😅

bourque (approved):
Nice work!

greglucas:
Are we waiting on anything here? I'm guessing you'll need to add some module-level docstrings as well once the base branch is updated as @bourque requested.

maxinelasp:
Yes, I will double check that I addressed all comments and then merge today (after adding docstrings of course 😆 )

PR #294, opened 2023-12-13
Add ruff check for module docstrings

bourque:
# Change Summary

## Overview

This PR adds an automated `ruff` check in the `pre-commit` hooks for identifying missing module docstrings. (Actually it just removes the specific check from the ignore list).

I also added module docstrings to the modules that were missing them.

## Updated Files

- `pyproject.toml`
   - Removed the module docstring check from the `ruff` ignore list
- A whole bunch of modules that were missing module docstrings


tech3371 (approved):
Thank you adding those doc strings!

PR #296, opened 2023-12-15
IMAP-Lo Science Direct Event Decompression

sdhoyt:
# Change Summary

## Overview
This PR adds a Science Direct Event data class for IMAP-Lo L0 to L1A processing. This data class includes methods to decompress the direct event data.

## New Dependencies
None

## New Files
- compression_tables.py
   - decompression tables for science direct events
   - decompression tables for science counts
- lol0.py 
   - general Lo L0 data class to be inherited by all data product classes
   - mainly used for parsing the packet data and setting attributes
- science_direct_events.py
   - data class for handling science direct event L0 data during L1A processing
   - includes methods for data decompression

## Deleted Files
None

## Updated Files
None

## Testing
- Added test_science_direct_events.py with unit tests for science direct event decompression. These unit tests should test most of the code added, but because I don't have compressed DE data, some workarounds were used to fully test the functionality. These tests will need to be updated when data becomes available. Currently, the tests will only work if the code in __init__() is commented out because there is no data to input. Not sure of a better way to handle this, but the tests will fail in the pipeline checks.

NOTE: I did not add any tests for the class lol0.py class that will be inherited by all the L0 data classes. I will be unable to test this until my next PR when I add classes for packets that I have available or when I get DE data.


imap_processing/lo/l0/compression_tables.py
```diff
@@ -0,0 +1,853 @@
...
+    247: 46763,
+    248: 48761,
+    249: 50844,
+    250: 53016,
+    251: 55282,
+    252: 57644,
+    253: 60107,
+    254: 62675,
+    255: 65353,
+}
```
maxinelasp: Is there any way to generate these numbers algorithmically? Instead of using these tables? Alternatively, could you write a function that has some kind of `return bit12_to_bit16[number] if number in bit12_to_bit16 else number` check instead of mapping one number to itself so many times?

imap_processing/lo/l0/lol0.py
```diff
@@ -0,0 +1,48 @@
...
+    ----------
+    ground_sw_version : str
+        Ground software version
+    packet_file_name : str
+        File name of the source packet
+    ccsds_header : CcsdsData
+        CCSDS header data
+    """
+
+    ground_sw_version: str
```
maxinelasp: Rather than passing it in here, it probably works better to put it in `__init__.py` and import that version when needed

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+    TIME: float
+    ENERGY: float
+    POS: float
+    case_number: int
+    tof_calculation_binary: dict
+    tof_decoder: list
+    remaining_bits: dict
+    parsed_bits: list
+
+    def __init__(self, packet, software_version: str, packet_file_name: str):
```
maxinelasp, line 89: Just an FYI that overriding the `__init__` function means you can't use some nice features of dataclasses (such as `dataclass.replace()`). If you aren't using those functions then it's fine, just wanted to mention it because I learned that after my dataclasses were all set up.

imap_processing/lo/l0/lol0.py
```diff
@@ -0,0 +1,48 @@
+"""General Lo L0 data class used for parsing data and setting attributes."""
+from dataclasses import dataclass, fields
+
+from imap_processing.ccsds.ccsds_data import CcsdsData
+
+
+@dataclass
+class LoL0:
+    """Data structure for common values across histogram and direct events data.
```
maxinelasp: Does Lo also have histogram data? Or is this leftover from GLOWS?

imap_processing/lo/l0/lol0.py
```diff
@@ -0,0 +1,48 @@
...
+    ground_sw_version : str
+        Ground software version
+    packet_file_name : str
+        File name of the source packet
+    ccsds_header : CcsdsData
+        CCSDS header data
+    """
+
+    ground_sw_version: str
+    packet_file_name: str
```
maxinelasp: Is this something Lo wants in their datafiles? Because otherwise, it's not really needed.

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+        super().__init__(software_version, packet_file_name, CcsdsData(packet.header))
+        self.parse_data(packet)
+
+    def decompress_data(self):
+        """Decompress the Lo Science Direct Events data."""
+        self._find_decompression_case()
```
maxinelasp: Might I suggest passing in pieces of `self.DATA` into these methods? eg `self._find_decompression_case(self.DATA[0:4]
self._find_tof_decoder_for_case(self.DATA[4]` 

etc.. it makes it more clear what is dependent on `self.DATA` and also lets you check that you don't have multiple methods acting on the same parts of DATA by accident. That being said, I see later you have a weird dependency for `start_bit` so it maybe wouldn't work so well in that case. 

Also, it's not clear to me, but if any of the class vars aren't expected to be used by external code (for example, if you're only using `self.case_number` to calculate TOF decoders and parsing, but it's meaningless outside of that context) then I'd consider having the method return that value and passing it in to methods that need it. If I see a class variable, I assume it's important for external use cases as well. 

That all being said, it's a nitpick, and I think these methods are clean and testable.

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+        condition for golden triples. If golden triples are below
+        a certain threshold in checksum it's considered golden, otherwise,
+        it's considered a silver triple. This is important for the compression
+        for golden triples because it's used to recover TOF1 because
+        compression scheme to save space on golden triples doesn't send
+        down TOF1 so it's recovered on the ground using the checksum
+    case_number: int
+        The compression case number for the direct event. The case number
+        determines how the bits are arranged in the compressed data.
+    tof_calculation_binary: dict
```
maxinelasp, line 56: Could you add a bit more about what the `dict` and `list`s here end up looking like? It's not clear to me after looking at where they are set.

maxinelasp: Also, you can be more specific in the typehint if that makes sense: `list[str]` or `dict[str, int]`

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+    tof_decoder: list
+    remaining_bits: dict
+    parsed_bits: list
+
+    def __init__(self, packet, software_version: str, packet_file_name: str):
+        super().__init__(software_version, packet_file_name, CcsdsData(packet.header))
+        self.parse_data(packet)
+
+    def decompress_data(self):
```
maxinelasp: Where is this method used? It's a little confusing to me that it's not called as part of `__init__`. Is there a case where you would use the data without calling `decompress_data`?

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+        # calculate the TOF values
+        tof_calculation_values = ct.tof_calculation_table[self.case_number]
+        self.tof_calculation_binary = {
+            field: bitstring.Bits(calculation_value)
+            for field, calculation_value in tof_calculation_values.items()
+        }
+
+    def _find_decompression_case(self):
+        """Find the decompression case for this DE."""
+        # The first 4 bits of the binary data are used to
```
maxinelasp: In general the documentation for these methods is good, but the paragraphs of `#` comments should just be moved into the docstring proper. I'd also recommend saying it sets a class variable if that's the case, as that's a side affect of calling the method that should be documented.

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+            decompressed_data = round(
+                sum(
+                    bit[0] * bit[1]
+                    for bit in zip(
+                        bitarray(needed_bits).tolist(), self.remaining_bits[field]
+                    )
+                ),
+                2,
+            )
+            setattr(self, field, decompressed_data)
```
maxinelasp, line 233: Does this error if `field` doesn't exist in the class? Is this overriding existing fields that were decommed (eg `self.TOF1`) or creating new ones? In general, I think `setattr` should be used with caution. It looks like `field` comes from `tof_decoder_table` and doesn't have that many values, so would it make sense to have TOF be a class with those fields and instead of using `setattr` here, you just create a new instance of that TOF class and then add it in here...? I'm not sure what the best way is, but I wanted to flag `setattr` to see if there's a more clear way (because I had to go back through several class variables to figure out what `field` would be)

imap_processing/lo/tests/test_science_direct_events.py
```diff
@@ -0,0 +1,290 @@
...
+    assert tof_decoder_true == tof_decoder_expected
+
+
+def test_read_tof_calculation_table():
+    # Arrange
+    de = ScienceDirectEvents("fake_packet", "0", "fakepacketname")
+    de.DATA = bitstring.Bits(bin="000100010101")
+    de._find_decompression_case()
+    binary_strings_expected = {
+        "TIME": bitstring.Bits(bin="0000111111111111"),
```
maxinelasp: Is this table always the same? could it be a class instead of a dict? (can you tell I don't like python dicts... haha)

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
maxinelasp: what is this file for?

maxinelasp (commented):
I had a bunch of questions but overall this looks clean, very nice!

imap_processing/lo/l0/lol0.py
```diff
@@ -0,0 +1,48 @@
...
+
+        """
+        attributes = [field.name for field in fields(self)]
+
+        # For each item in packet, assign it to the matching attribute in the class.
+        for key, item in packet.data.items():
+            value = (
+                item.derived_value if item.derived_value is not None else item.raw_value
+            )
+            print(value)
```
greglucas:
```suggestion
```

imap_processing/lo/l0/lol0.py
```diff
@@ -0,0 +1,48 @@
...
+            if key in attributes:
+                setattr(self, key, value)
+            else:
+                raise KeyError(
+                    f"Did not find matching attribute in {self.__class__} data class"
+                    f"for {key}"
+                )
```
greglucas: You could reverse this to `if key not in attributes: raise` and avoid the else block.

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
+"""L1A Science Direct Events data class."""
+from dataclasses import dataclass
+from itertools import compress
+
+import bitstring
+from bitarray import bitarray
+
+import imap_processing.lo.l0.compression_tables as ct
+from imap_processing.ccsds.ccsds_data import CcsdsData
+from imap_processing.lo.l0.lol0 import LoL0
```
greglucas: This is hard to parse with all the module levels (also the 0s and Os), and also makes me want to laugh out loud 😂 

I have a personal preference for the single level heirarchy since we don't have that many files in each level, but realize we are split right now in how we are doing that for all the instruments.

```bash
imap_processing.lo.compression_tables
imap_processing.lo.level0
```

imap_processing/lo/l0/science_direct_events.py
greglucas: I'd agree with @maxinelasp's suggestion, but possibly my preference for plain functions goes even farther where these would just be module-level functions not methods on objects and remove `self` altogether to avoid the potential spurious side-effects you mention.

```python
case = _find_decompression_case(self.DATA[0:4])
decoder = _find_tof_decoder_for_case(self.DATA[4], case)
...
```

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+            self.parsed_bits[field] = self.DATA[bit_start : bit_start + bit_length]
+            bit_start = bit_start + bit_length
+
+    def _decode_fields(self):
+        """Use the parsed data and TOF coefficients to decode the binary."""
+        for field, tof_bits in self.parsed_bits.items():
+            needed_bits = tof_bits[-len(self.remaining_bits[field]) :]
+            # Use the TOF coefficients and the bits for the current field to
+            # calculate the decompressed value. Also round to 2 because the TOF
+            # coefficients only have 2 decimals.
```
greglucas, line 222: Are the `needed_bits` and `remaining_bits` floats? Since they are named `bits`, it seems like they are integers at first glance and thus wouldn't need to be rounded until later stages.

imap_processing/lo/tests/test_science_direct_events.py
```diff
@@ -0,0 +1,290 @@
+import bitstring
+
+from imap_processing.lo.l0.science_direct_events import ScienceDirectEvents
+
+# TODO: These unit tests don't call the public function for the ScienceDirectEvents
+# class and are also only able to run if the contects of __init__ are commented out
+# because I currently don't have any compressed DE data. As a workaround, I am
+# directly setting the DATA attribute and calling and testing the private methods
+# as needed. These tests will be updated when data becomes available. When that happens
+# the need for the bitstring import will also go away.
```
greglucas: I actually think this is great! Don't remove these once you get real data. Mock data is fantastic for checking things are working as expected in a simple to comprehend way without needing to realize what binary values are in a file.

imap_processing/lo/tests/test_science_direct_events.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def test_find_decompression_case():
+    # Arrange
+    de = ScienceDirectEvents("fake_packet", "0", "fakepacketname")
+    de.DATA = bitstring.Bits(bin="000100010101")
```
greglucas: You could turn `de` into a pytest fixture since you're re-using it a lot.

Alternatively, if you make these functions without `self` you could just test the function directly without needing to modify the DATA attribute at all. You could just pass in various combinations of bitstrings to your function.

imap_processing/lo/tests/test_science_direct_events.py
```diff
@@ -0,0 +1,290 @@
...
+    case_number_expected = 1
+
+    # Act
+    de._find_decompression_case()
+    case_number_true = de.case_number
+
+    # Assert
+    assert case_number_true == case_number_expected
```
greglucas: I find the `true == xyz` hard to comprehend. IMO this is shorter and more readable.
```suggestion
    # Assert
    assert de.case_number == case_number_expected
```

imap_processing/lo/l0/compression_tables.py
```diff
@@ -0,0 +1,853 @@
+"""Compression tables for decompompressing Lo L0 data."""
+tof_case_table = {
```
greglucas: Could this just be a function call instead?
`int(bitstring, 2)`

imap_processing/lo/l0/compression_tables.py
```diff
@@ -0,0 +1,853 @@
...
+    "1010": 10,
+    "1011": 11,
+    "1100": 12,
+    "1101": 13,
+}
+
+tof_decoder_table = {
+    # First level of keys are the case numbers
+    0: {
+        1: {
```
greglucas: Here you define `[1, 0]`, then in case 1 you define `[1, 2, ..]`. Just verify that the counting is from 0 here and 1 in the next case.

imap_processing/lo/l0/compression_tables.py
```diff
@@ -0,0 +1,853 @@
...
+            "ENERGY": 3,
+            "POS": 0,
+            "TOF0": 10,
+            "TOF1": 0,
+            "TOF2": 9,
+            "TOF3": 6,
+            "CKSM": 3,
+            "TIME": 12,
```
greglucas: Would it make sense to create a namedtuple class for these?
```python
TOFData = namedtuple("TOFData", "ENERGY POS TOF0 TOF1 TOF2 TOF3 CKSM TIME")
{1: TOFData(3, 0, 10, 0, 9, 6, 3, 12)}
```

imap_processing/lo/l0/compression_tables.py
```diff
@@ -0,0 +1,853 @@
...
+        "TOF0": 0,
+        "TOF1": 0,
+        "TOF2": 11,
+        "TOF3": 0,
+        "CKSM": 0,
+        "TIME": 12,
+    },
+}
+
+tof_calculation_table = {
```
greglucas: In general, I feel like tables should be lists that can be indexed into rather than dicts since you're just adding the explicit ints here but needing a hash call. If you like to have the identifier perhaps it could be a comment `# item 0`

```python
tof_calculation_table = [{"TIME": "0xFFF", ...}, {"ITEM1"...}, ...]
```

edit: I see you did this to the one right below this already, so there is precedence :)

imap_processing/lo/l0/compression_tables.py
```diff
@@ -0,0 +1,853 @@
+"""Compression tables for decompompressing Lo L0 data."""
```
GFMoraga: I feel like this could be a class that you import to use/call.

imap_processing/lo/l0/compression_tables.py
sdhoyt: I can see if Lo can provide an algorithm for how those mappings were generated. I think I'm confused about what you mean by "instead of mapping one number to itself so many times".

As a side note, I think I should remove these  dictionaries from this PR either way because these are for the Science Count decompression. I originally started this PR thinking it would only be adding decompression tables but decided to make it a Science DE data class / decompression PR and figured I'd leave these in here, but it'd probably be better to remove them and add them in a separate Science Count data class / decompression PR. I do want to understand what you're thinking for the future though.

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Good to know, I might want to use those in the future. It looks like you switched to use `__post_init__` for that reason?

imap_processing/lo/l0/lol0.py
sdhoyt: Good catch. They do, but not here. That's there because I copied your GLOWS code.

imap_processing/lo/l0/lol0.py
sdhoyt: They're not asking for it in their data but I thought we were adding that to the cdf headers. I might be wrong about that though.

imap_processing/lo/l0/science_direct_events.py
sdhoyt: I was thinking it would be used in the decom code (that I haven't put out for PR yet). as a separate method call. So it'd be like:
```python
sci_de = ScienceDirectEvent(packet, software_version, packet_name)
sci_de.decompress_data()
```

But I guess there would be no scenario where I'd create a sci de object and not decompress, so I'll put it in the `__init__` or `__post_init__`, whatever one I end up using.

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Good points. I'll make those updates.

imap_processing/lo/l0/science_direct_events.py
sdhoyt: @maxinelasp Yeah, I was back and forth on that and figured I'd see what others thought. I will pass in and return those values instead.

@greglucas how strong is your preference for that? I definitely have a preference for organizing them as methods, but if that's a standard we want to follow, I will make that change.

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Definitely. I'll add more info for those

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Yeah that makes sense. I figured it was safe because those fields are coming from the tables and not the data. Also, the TOF values and other attributes that are being set there were never decommed; they are being pulled from the binary only so they haven't been set yet. But I like the idea of creating a TOF class, that might be the way go.

imap_processing/lo/tests/test_science_direct_events.py
sdhoyt: haha whatever is going on between you and python dictionaries seems personal. Maybe I haven't thought about it enough, but the problem is sometimes those values don't exist depending on the case number. And different case numbers have different combinations of those values that don't exist. So I'd need to have a bunch of logic in there to pick with of those values are populated based on the case number.

package-lock.json
sdhoyt: I have no idea. I assumed it was something poetry related, but now I'm realizing it was created in this PR. I definitely didn't make this on purpose lol I'll remove it.

imap_processing/lo/l0/lol0.py
sdhoyt: Ah good idea. I'll update that.

imap_processing/lo/l0/science_direct_events.py
sdhoyt:
```python
from imap_processing.lo.l0.lol0
```
In my head I've been calling this my Flo Rida import because whenever I see it I just hear the chorus to Low.

You're right though, I should make it less ridiculous.

imap_processing/lo/l0/science_direct_events.py
sdhoyt: I definitely need to clean up the naming here. `remaining_bits` are floats which doesn't make any sense. I got that name because a string of bits are used to determine which of those floats should remain, but that name doesn't work well for the resulting list here. I'll update these names to make it more clear

imap_processing/lo/tests/test_science_direct_events.py
sdhoyt: Sounds good. I can definitely continue to directly set variables for testing, but once I get data I'll be able to properly initialize the object rather than comment out `__init__`. I'll update that comment to reflect that.

imap_processing/lo/l0/compression_tables.py
sdhoyt: oh wow lol yeah...I didn't notice that. I'll remove this table and convert the binary to decimal

imap_processing/lo/l0/compression_tables.py
sdhoyt: I don't think I'm following. I have the outer dictionary as {0, 1, 2, 3, etc.} then the inner dictionary for 0 as {1, 0}. I'm not following where I define 1, 2, in case 1

imap_processing/lo/l0/compression_tables.py
sdhoyt: Yeah, I agree, that would be better.

imap_processing/lo/l0/compression_tables.py
maxinelasp: for the first 100 values, you're just mapping `n:n` for example `1:1`, `2:2`... so instead of writing them out in this dictionary, if you abstract it using some kind of `convert_bits` method, you can say `if n <= 100: return n else: return bit12_to_bit16[n]` which is more clear and reduces the lines of code here. It does _feel_ like there's some pattern here though, maybe Lo can tell you more and these dictionaries will be unneeded.

imap_processing/lo/l0/lol0.py
maxinelasp: I don't think so, but I am not sure. I have it in GLOWS because they specifically asked for it in their algorithm document.

imap_processing/lo/l0/compression_tables.py
sdhoyt: I have the dictionary with the field names so it's easier to keep track of which fields should be missing for which cases. Maxine had another comment suggesting that this could be a class, so I might try that approach. That might let me be explicit about field names, while removing the dictionary.

imap_processing/lo/l0/science_direct_events.py
maxinelasp: If you want to keep the `l0` folder, you could rename it `LoBase` or something. Or maybe we can convince everyone to rename the instrument, because it's just going to be rough with L0 and Lo 😆

imap_processing/lo/l0/science_direct_events.py
maxinelasp: Yes, and I used `*args` to turn the list of outputs from `packet.data` into the arguments for the dataclass instance.

imap_processing/lo/l0/science_direct_events.py
maxinelasp: Yeah if you're going to always apply it, I definitely vote for putting it in `__init__` or `__post_init__`

imap_processing/lo/l0/science_direct_events.py
greglucas: +0.5, non-blocking, so it is totally fine to do it this way if you prefer.

Just for clarity though, maybe another way to express what I am getting at is to say you have one class to store the initial input, and then another _different_ class to store the decompressed data, so as not to unpack it into the same object structure.

```python
def decompress_data(data: Level0Compressed) -> Level0Decompressed
    # call other functions/methods on the L0 object to get the tables you need
    return new_data_object

decompressed_data_object = decompress_data(input_raw_data)
```

maxinelasp: Personally, I prefer to keep it organized within the object, because the methods are tightly coupled with the data inside the object and are not very widely applicable outside of the dataclass.

imap_processing/lo/l0/science_direct_events.py
maxinelasp: Generally, yes, having the fields come from the tables means that this is alright in my book, but it is a red flag that someone could change those tables later and not realize that they are used for setting the fields here. So, additional documentation might be good, or just making that relationship more explicit with code design such as by making it a namedtuple or a class.

imap_processing/lo/tests/test_science_direct_events.py
maxinelasp: Yeah, makes sense to me. And I think my real problem with python dictionaries is actually because of Pythons type system. I don't mind hashtables in other languages! But they can get gnarly with ducktyping 🦆

maxinelasp:
Also, you probably want to add some kind of `lo.rst` file for documentation in this PR

imap_processing/lo/l0/compression_tables.py
```diff
@@ -0,0 +1,853 @@
...
+    "1000": 8,
+    "1001": 9,
+    "1010": 10,
+    "1011": 11,
+    "1100": 12,
+    "1101": 13,
+}
+
+tof_decoder_table = {
+    # First level of keys are the case numbers
```
tech3371: I find these comments very helpful. Can you add comments for other keys inside this case numbers. Eg. key `0` and `1` inside case number `0`

imap_processing/lo/l0/compression_tables.py
```diff
@@ -0,0 +1,853 @@
...
+    "0111": 7,
+    "1000": 8,
+    "1001": 9,
+    "1010": 10,
+    "1011": 11,
+    "1100": 12,
+    "1101": 13,
+}
+
+tof_decoder_table = {
```
tech3371: what is total number of bits pass to this table? Each decoder seems to read bits in different lengths based on case number.

imap_processing/lo/l0/compression_tables.py
tech3371: I thought about that too. I don't see a pattern after 100. I even ask ChatGPT and it didn't find pattern. You could do something like this. Store all non 1-1 value in the dict and then in your code, you can have if statement like Maxine is suggesting.
```
bit8_to_bit12 = {
101: 102
..
..
}
```

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+from bitarray import bitarray
+
+import imap_processing.lo.l0.compression_tables as ct
+from imap_processing.ccsds.ccsds_data import CcsdsData
+from imap_processing.lo.l0.lol0 import LoL0
+
+
+@dataclass
+class ScienceDirectEvents(LoL0):
+    """L1A Science Drect Events data.
```
tech3371:
```suggestion
    """L1A Science Direct Events data.
```

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+    TIME: int
+        time tag for the direct event
+    ENERGY: int
+        energy of the direct event ENA.
+    POS: int
+        Stop position for the direct event. There are 4 quadrants
+        on the at the stop position.
+    CKSM: int
+        This is checksum defined relative to the TOFs
+        condition for golden triples. If golden triples are below
```
tech3371, line 47: what is golden triples?

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+        self._parse_binary()
+        self._decode_fields()
+
+    def _find_tof_decoder_for_case(self):
+        """Get the TOF decoder for this DE's case number."""
+        # Case 0 can either be a gold or silver triple. Gold triples do
+        # not send down the TOF1 value and instead recover the TOF1 value
+        # on the ground using the decompressed checksum
+        if self.case_number == 0:
+            gold_or_silver = int(self.DATA.bin[4])
```
tech3371, line 133: I like the athletic medal naming in the code.

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
+"""L1A Science Direct Events data class."""
+from dataclasses import dataclass
+from itertools import compress
+
+import bitstring
+from bitarray import bitarray
+
+import imap_processing.lo.l0.compression_tables as ct
```
tech3371: can you spell out `ct` to a short name?

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+    def decompress_data(self):
+        """Decompress the Lo Science Direct Events data."""
+        self._find_decompression_case()
+        self._find_tof_decoder_for_case()
+        self._read_tof_calculation_table()
+        self._find_remaining_bits()
+        self._parse_binary()
+        self._decode_fields()
+
+    def _find_tof_decoder_for_case(self):
```
tech3371: Can you order these functions in the order they are called? It's not important but it helps to follow it.

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -0,0 +1,202 @@
...
+        """Get the TOF decoder for this DE's case number."""
+        # Case 0 can either be a gold or silver triple. Gold triples do
```
tech3371: Thank you for these comments again

imap_processing/lo/l0/science_direct_events.py
sdhoyt: I believe a Triple Coincidence is when there are measurements for all the TOF detectors and if the checksum value is below a certain threshold then it's considered a Golden Triple. In that case, the TOF3 is not transmitted and is instead calculated on the ground using the other TOFs and the checksum. I'm not 100% what that translates to conceptually. I think it has something to do with the amount of background suppression. I'm going to ask Colin for more info on this. Last time we met I was too wrapped up in the tables and forgot to ask what a Golden Triple really means.

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Yeah, I agree. I'll reorder

imap_processing/lo/l0/compression_tables.py
sdhoyt: Yeah, there are some fields that aren't transmitted depending on the case number. Do you mean I should have the sum total of bits for a case in the dictionary as well?

imap_processing/lo/l0/compression_tables.py
sdhoyt: Ah, I see what you're saying. Yeah, I'll do that

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Yeah, every time I see someone's PR or issue that has L0 I think "why are they working on Lo?" and then I realize it's a "0" and not an "o"

imap_processing/lo/l0/compression_tables.py
sdhoyt: You mean each table could be a class instead?

greglucas:
I think you accidentally got some of the old docs committed. They changed locations in a recent PR, so you'll need to remove those old references. Hopefully that will fix your doc build.

`docs/source/reference/generated/` -> `docs/source/code-documentation/generated/`

I think this is what you'd need.
```bash
rm -rf docs/source/reference
git add .
git commit
```

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -85,116 +85,155 @@ class ScienceDirectEvents(LoL0):
...
+        The TOF decoder table shows how the TOF bits should be
+        parsed in the binary data.
+
+        Case 0 can either be a gold or silver triple. Gold triples do
+        not send down the TOF1 value and instead recover the TOF1 value
+        on the ground using the decompressed checksum.
+        Cases 4, 6, 10, 12, 13 may be Bronze triples. If it's not a bronze triple,
+        the Position is not transmitted, but TOF3 is. If it is bronze, the table
+        should be used as is. If it's not bronze, position was not transmitted,
+        but TOF3 was transmitted.
```
maxinelasp, line 130: Wow, this is complicated! But, I think your explanation and the structure of your code line up nicely

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -85,116 +85,155 @@ class ScienceDirectEvents(LoL0):
...
+    def _decode_fields(self, remaining_bit_coefficients, parsed_bits):
         """Use the parsed data and TOF coefficients to decode the binary."""
-        for field, tof_bits in self.parsed_bits.items():
-            needed_bits = tof_bits[-len(self.remaining_bits[field]) :]
+        for field, tof_bits in parsed_bits.items():
+            needed_bits = tof_bits[-len(remaining_bit_coefficients[field]) :]
             # Use the TOF coefficients and the bits for the current field to
             # calculate the decompressed value. Also round to 2 because the TOF
             # coefficients only have 2 decimals.
+            print(f"NEEDED BITS: { bitarray(needed_bits).tolist()}")
```
maxinelasp: Should these be left in? Or just used for debugging?

maxinelasp (approved):
This looks great! Nice job on some seriously complicated bit manipulations! I like your updates to the data structures and everything.

sdhoyt:
Would everyone be ok with me adding `pass` at the beginning of each unit test since I needed to comment out the class's init for the unit tests to run since I don't have data currently? Once I get a Sci-De packet this should all run normally. Or would it be worth trying to fake space-packet-parser output?

greglucas:
> Would everyone be ok with me adding pass at the beginning of each unit test since I needed to comment out the class's init for the unit tests to run since I don't have data currently? Once I get a Sci-De packet this should all run normally. Or would it be worth trying to fake space-packet-parser output?

You should add a `pytest.xfail()` or `pytest.skip()` (either to individual tests, or to the module) instead. But, yes that is fine with me to get this in.

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Thanks for catching that! Removing those print statements

PR #298, opened 2023-12-21
Updated the pkt def xtce

GFMoraga:
# Change Summary

## Overview

I just updated the xtce packet definitions with the new excel sheet. Only thing that changed is CAPS.


bourque (approved):
Thanks for doing that!

PR #299, opened 2023-12-21
Add instructions for rebasing in git workflow docs

bourque:
# Change Summary

## Overview

Updates the git workflow docs to use rebasing instead of pulling/merging, as I think that is our preferred method? (Might be opening a can of worms here 😅)

## Updated Files

-`docs/source/development-guide/style-guide/git-and-github-workflow.rst`
   - Updated 'keeping your fork updated' section to use rebasing


laspsandoval (approved):
Thanks!

tech3371 (approved):

maxinelasp:
Maybe we can leave the instructions for `git pull` in there and just add this information as another option? I find rebasing to be quite tricky and I wouldn't necessarily want people who are inexperienced with git to use it, since it's easy to wreck stuff.

PR #300, opened 2023-12-21
Initial CoDICE L1a algorithm for lo-sw-species data

bourque:
# Change Summary

## Overview

This PR expands the current CoDICE L1a code to add initial support for packets with science data (currently only supports `LO_SW_SPECIES_COUNTS` apid).

There is still work to be done to get this fully functioning for this APID, but I think it is at a good point for a review+merge before the PR gets too big.

## New Files

- `imap_processing/codice/data/esa_sweep_values.csv`
  - ESA Sweep table values (determines energy levels)
- `imap_processing/codice/data/lo_stepping_values.csv`
  - Lo stepping table values (determines acq times) 

## Updated Files

  - Added initial support for Lo sunward species counts data product
  - Added lookup tables for compression, collapse tables, ESA sweep tables, Lo stepping table 
  - Minor formatting and value updates 
  - Updated filename convention


imap_processing/codice/codice_l1a.py
```diff
@@ -12,17 +11,292 @@
...
+    plan_step : int
+        Plan step that was active when the data was acquired and processed. In
+        conjunction with ``plan_id``, describes which counters are included
+        in the data packet
+    view_id : int
+        Provides information about how data was collapsed and/or compressed
+    use_simulated_data : bool
+        When ``True``, simulated science data is generated and used in the
+        processing pipeline. This is useful for development and testing in the
+        absence of actual CoDICE testing data.
```
greglucas: Are you able to pass this simulated data directly in somehow instead of requiring a flag inside this class?

In the `__init__` method you are passing in only single values of ints, etc... In the real case how would you send in the bytes, or would it be processed before this step?

To give an example I'm wondering if you can do something like:
`Class(binary_data)`, then in the test suite you'd bass in `Class(_generate_simulated_data(length_in_bits))`, and in the real world you'd do something like `Class(parse_packets(packet_list))`

imap_processing/codice/codice_l1a.py
```diff
@@ -36,19 +310,48 @@ def codice_l1a(packets: list[space_packet_parser.parser.Packet]) -> str:
         The path to the CDF file that was created
     """
     # Group data by APID and sort by time
+    print("Grouping the data by APID")
     grouped_data = group_by_apid(packets)
+    grouped_data = {1154: [], 1156: []}  # Temporary during development
```
greglucas: It might be helpful to use the named values from the Enums you created instead of these integers. ` CODICEAPID.COD_NHK.value` or something similar.

imap_processing/codice/codice_l1a.py
```diff
@@ -36,19 +310,48 @@ def codice_l1a(packets: list[space_packet_parser.parser.Packet]) -> str:
...
+            file = imap_data_access.ScienceFilePath.generate_from_inputs(
+                "codice", "l1a", "hk", "20210101", "20210102", "v01-01"
+            )
+
+        elif apid == CODICEAPID.COD_LO_SW_SPECIES_COUNTS:
+            print(f"processing {apid} packet")
+            packets = grouped_data[apid]
+            table_id, plan_id, plan_step, view_id = get_params(packets)
+
+            pipeline = CoDICEL1a(table_id, plan_id, plan_step, view_id)
```
greglucas: I was going to say that I wasn't entirely sure this deserved a class or could just be functions. I think you are re-using a lot of the id's each time, but I'm still confused how the data is getting sent in, if it would be in the pipeline initialization or in one of these supporting functions.

My minor suggestion at this point would be add pipeline to your class name. I like that you set it to that variable here because I think that helps me understand what this is used for now that I see it :)
```suggestion
            pipeline = CoDICEL1aPipeline(table_id, plan_id, plan_step, view_id)
```

imap_processing/codice/utils.py
```diff
@@ -60,17 +59,15 @@ class CoDICECompression(IntEnum):
...
-    LOSSY_B_LOSSLESS = 6
+    NO_COMPRESSION = 0
+    LOSSY_A = 1
+    LOSSY_B = 2
+    LOSSLESS = 3
+    LOSSY_A_LOSSLESS = 4
+    LOSSY_B_LOSSLESS = 5
 
 
-def add_metadata_to_array(
```
greglucas, line 71: Why remove the packet type hint? Is that for the simulated data issue?

greglucas (commented):
I see you have a TODO to change the prints to loggers, that is my only real minor suggestion, other than a few more inline.

imap_processing/codice/codice_l1a.py
bourque: My intention was to completely remove this flag once I get test data to work with. Perhaps I can just add a note that this flag is only temporary?

imap_processing/codice/codice_l1a.py
bourque: Thanks, I think this is a good suggestion -- it is more explicit and thus more readable.

My motivation for using a class over a function here was to avoid having to pass around so many variables to various functions. With a class, I can set the 4 'main' variables as well as whatever values are going to end up in the CDF as attributes to the class, and that makes the method calls much cleaner IMO.

That being said, I am open to ultimately changing the overall design here, but so far this is serving me well while I am in the development phase. And it might make sense to change things up once I get test data -- Then I won't have to rely on simulating the 4 'main' parameters or the science data bytes.

imap_processing/codice/utils.py
bourque: I think I removed it to get around an error, but I just added it back fixed it!

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -41,7 +43,6 @@ def test_codice_l1a(l0_test_data: list[space_packet_parser.parser.Packet]) -> st
         pytest fixture used to provide a temporary directory during testing
     """
 
-    cdf_filename = codice_l1a(l0_test_data)
+    cdf_filename = process_codice_l1a(l0_test_data, tmp_path)
```
greglucas: You shouldn't need the `tmp_path` now after the dev changes.

```suggestion
    cdf_filename = process_codice_l1a(l0_test_data)
```

greglucas (approved):
Looks like there are pre-commit failures that need to be fixed and one minor merge thing that I added an inline note about, otherwise looks good to me.

imap_processing/codice/constants.py
```diff
@@ -1,5 +1,137 @@
...
+    4: 1,
+    5: 2,
+    6: 4,
+    7: 5,
+    8: 6,
+    9: 7,
+}
+
+# ESA Sweep table ID lookup table
+# The combination of plan_id and plan_step determine the ESA sweep Table to use
```
tech3371, line 60: If I remember, these plan_id and plan_step comes from packet data, right?

May be at the top of this file, it may be helpful to identify which values are coming from packet data. This may help identify which values are passed in vs looked up.

imap_processing/codice/decompress.py
```diff
@@ -81,6 +84,7 @@ def _apply_lzma_lossless(compressed_value: int) -> int:
     decompressed_value : int
         The 24- or 32-bit decompressed value
     """
+    print(compressed_value)
```
tech3371: Did you mean to remove this?

imap_processing/codice/codice_l1a.py
```diff
@@ -12,18 +11,299 @@
...
+        -------
+        dataset : xr.Dataset
+            The dataset used in the L1a CDF file
+        """
+        # TODO: Properly implement this
+
+        # metadata_arrays = collections.defaultdict(list)
+        #
+        epoch_time = xr.DataArray(
+            [np.datetime64("2010-01-01T00:01:06.184")],
```
tech3371: now, you can import this launch time.

imap_processing/codice/codice_l1a.py
```diff
@@ -12,18 +11,299 @@
...
+            )
+            self.science_values = [
+                decompress(compressed_value, self.compression_algorithm)
+                for compressed_value in compressed_values
+            ]
+
+        # TODO: Figure out how to properly unpack the data
+        # 128 e/q steps of 12 spin sectors x 5 positions
+        # Chunk the data by energy steps
+        num_bytes = len(self.science_values)
```
tech3371, line 254: is it number of bits? I don't remember now. I think there was mention of byte count.

imap_processing/codice/codice_l1a.py
```diff
@@ -12,18 +11,299 @@
...
+        energy_steps = 128
+        chunk_size = len(self.science_values) // energy_steps
```
tech3371, line 256: Can you add comment on this? Or may be add in the future?

imap_processing/codice/codice_l1a.py
```diff
@@ -37,16 +317,70 @@ def codice_l1a(packets: list[space_packet_parser.parser.Packet]) -> str:
...
+            logger.debug(f"{apid} is currently not supported")
+            continue
+
+        elif apid == CODICEAPID.COD_HI_SECT_SPECIES_COUNTS:
+            logger.debug(f"{apid} is currently not supported")
+            continue
 
     # Write data to CDF
-    cdf_filename = write_cdf(data)
+    # cdf_filename = write_cdf(data)
```
tech3371, line 378: If you are going to write cdf file here in the future, don't forget to keep track of files paths as you process multiple apid(s). Because you will need that to upload data from cli.py.

tech3371 (commented):
few comments but looks good to me.

imap_processing/codice/decompress.py
bourque: Yes, thanks!

imap_processing/codice/codice_l1a.py
bourque: This is most definitely going to change in the next PR, so I will address that then.

PR #302, opened 2024-01-03
Setup docker template access to EFS

### LINKED ISSUE #242: Setup docker template access to EFS
Setup Docker templates to allow access to the EFS

laspsandoval:
# Change Summary

## Overview
Sets up docker template that can be used to push images to the ECR, provides instructions of how to do that, and provides access to EFS containing SPICE kernels.

## New Files
- Dockerfile
   - Dockerfile that installs imap_processing and its dependencies
- docker.rst
   - Provides description of how to build and run docker image. Also contains information of how to use the image in the ECR

## Deleted Files
None

## Updated Files
- run_processing.py
   - I added a really brief example of how spice kernels from the ECR could be accessed.
- index.rst
   - Added docker.rst to toctree.
- poetry.lock
   - Updated

## Testing
- I made some minor changes to sds-data-manager, which I will describe in another PR for that repo. I then populated the EFS using the lambda Tenzin created (worked beautifully by the way). And then I executed a Step Function manually and looked at the Batch Job log. Picture is attached.

Screenshot 2024-01-03 at 11 30 07 AM


greglucas (commented):
Shouldn't this be in sds-data-manager? You have AWS information here, and the processing repository should be agnostic to any AWS work and able to be run locally.

I'm not following why we need all these steps... `pip install imap_processing` should get us all the scripts we need to do any processing steps. So it should be a 2-3 line Dockerfile I'd hope. (You can also remove the install git if you change to install from a zip archive for now)
* The script should be installed and accessed as `imap_processing --flags`.
* If `tools/` are needed during processing, then those should probably be moved up into the main package to get their dependencies automatically?

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# TODO: install imap_processing instead of copying it once a new version is released
+# Install imap_processing
+#RUN pip install imap_processing
+# TODO: delete this section once new version released for imap_processing
+RUN pip install git+https://github.com/IMAP-Science-Operations-Center/imap_processing.git@dev
+
+# Copy over only the necessary scripts
+COPY imap_processing/run_processing.py $IMAP_PROCESS_DIRECTORY/run_processing.py
+COPY tools $IMAP_PROCESS_DIRECTORY/tools
```
tech3371: Do we need to copy `tools` folder? I think it's only used locally to create XTCE.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN pip install --upgrade pip
+
+# TODO: install imap_processing instead of copying it once a new version is released
+# Install imap_processing
+#RUN pip install imap_processing
+# TODO: delete this section once new version released for imap_processing
+RUN pip install git+https://github.com/IMAP-Science-Operations-Center/imap_processing.git@dev
+
+# Copy over only the necessary scripts
+COPY imap_processing/run_processing.py $IMAP_PROCESS_DIRECTORY/run_processing.py
```
tech3371: I think it should come in the pip package. If it didn't, we could find out why.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+COPY tools $IMAP_PROCESS_DIRECTORY/tools
+
+# Create the /mnt/spice directory
+RUN mkdir -p /mnt/spice
+
+# Define the entrypoint of the container
+ENTRYPOINT ["python", "/opt/imap/run_processing.py"]
```
tech3371: I haven't tested it but I wonder if you could change this to the pip path. Then you won't need to copy in above line.
```suggestion
ENTRYPOINT ["python", "<pip path>/imap_processing/run_processing.py"]
```
I believe you can get <pip path> by doing `pip show imap_processing`

docs/source/development-guide/docker.rst
```diff
@@ -0,0 +1,37 @@
...
+
+This page describes how to build and run a Docker Image Locally and in AWS.
+
+Building and Running a Docker Image Locally
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To build the image run the following command from the directory containing the Dockerfile. You might add -t option to tag your image
+and --rm to remove intermediate containers after the build is done.
+
+    `docker build -t <image name> --rm .`
```
tech3371:
```suggestion
    `docker build -t <image name>:<tag name> --rm .`
```

docs/source/development-guide/docker.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Now we can run our image.
+
+    `docker run --rm -it --volume="$(pwd)/imap_processing/efs:/mnt/spice" <image name> --instrument <instrument> --level <level>`
```
tech3371:
```suggestion
    `docker run --rm -it --volume="$(pwd)/imap_processing/efs:/mnt/spice" <image name>:<tag name> --instrument <instrument> --level <level>`
```

docs/source/development-guide/docker.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Build the Docker image.
+
+    `docker build -t <image name> .`
+
+Tag the image and push to the ECR.
+
+    `docker tag <tag> <ECR URI>`
+
+    `docker push <ECR URI>`
```
tech3371: Is this the place where we would set image tag with version?

Dockerfile
laspsandoval: I removed the example I had of how to use the kernels, which is what I was using the tools folder for. But I removed the example.

docs/source/development-guide/docker.rst
laspsandoval: We can assign the tag name during the build :)

imap_processing/cli.py
```diff
@@ -36,11 +37,10 @@ def _parse_args():
         f"The data level to process. Acceptable values are: {processing_levels}"
     )
 
-    parser = argparse.ArgumentParser(description=description)
+    parser = argparse.ArgumentParser(prog="imap_cli", description=description)
```
tech3371, line 39: I was wondering what this was but looks like it was from poetry. cool

tech3371 (approved):
looks good to me. Thank you!

imap_processing/cli.py
```diff
@@ -7,17 +7,18 @@
 
 Use
 ---
-    python run_processing.py <instrument> <data_level>
+    python cli.py <instrument> <data_level>
```
greglucas: `imap_cli --instrument <instrument> --level <data_level>`

imap_processing/cli.py
```diff
@@ -162,11 +162,12 @@ def process(self):
         print(f"Processing IMAP-Ultra {self.level}")
 
 
-if __name__ == "__main__":
```
greglucas, line 165: You still need the `if __name__ == "__main__" call main() block for running it as `python cli.py` So it is best to leave that in there for either option.

Dockerfile
greglucas, line 1: I'd prefer this doesn't live at the root level. Maybe move it into an `examples/` folder?

Also perhaps call it `some_unique_name.Dockerfile` or `Dockerfile.some_unique_name` so it isn't just a base Dockerfile?

greglucas (commented):
OK, since this is mostly documentation at this point I'm fine with keeping it in this repository since it is more public-facing. But, I think the Dockerfile should be moved out of our base repository level.

imap_processing/cli.py
```diff
@@ -7,17 +7,18 @@
...
 import argparse
 import sys
 from abc import ABC, abstractmethod
+from typing import Optional
 
 from imap_processing import instruments, processing_levels
 
 
-def _parse_args():
+def _parse_args(cli_args: list):
```
greglucas: Why do we need the `cli_args` here and in `main()`? I was pretty sure in my testing that there was not extra args passed into the script's main function call...

https://docs.python.org/3/library/argparse.html#parsing-arguments
> In a script, [parse_args()](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_args) will typically be called with no arguments, and the [ArgumentParser](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser) will automatically determine the command-line arguments from [sys.argv](https://docs.python.org/3/library/sys.html#sys.argv).

docs/source/development-guide/docker.rst
```diff
@@ -0,0 +1,37 @@
...
+
+This page describes how to build and run a Docker Image Locally and in AWS.
+
+Building and Running a Docker Image Locally
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To build the image run the following command from the directory containing the Dockerfile. You might add -t option to tag your image
+and --rm to remove intermediate containers after the build is done.
+
+    `docker build -t <image name>:<tag name> --rm .`
```
greglucas: Do you want to give an example of how to build the image you added? You'll need the `-f Dockerfile.name` now because it isn't the default Dockerfile (sorry for the misdirection there)!

```suggestion
    `docker build -f examples/Dockerfile.efs -t <image name>:<tag name> --rm .`
```

greglucas (approved):
Thanks for the clear explanations and descriptions, I think this is nice with the explicit example.

I would still like to see the `cli_args` removed if that isn't actually needed/used.

bourque:
@all-contributors please add @laspsandoval for infrastructure and ideas

PR #303, opened 2024-01-03
Make a default data directory

greglucas:
# Change Summary

## Overview

This adds a global data directory to maintain all files underneath that is user-settable via an environment variable or command line argument. The purpose is to start moving towards the same logical folder structure we are using in AWS, but locally so all our file management is consistent. I ended up touching more than I wanted to here to try and add some tests and bring the current tests into pathlib parity. I'm happy to try and break some of this out if it is too large for a review now.

## New Dependencies
n/a

## New Files
n/a

## Deleted Files
n/a

## Updated Files
Many updates to various path handling functions.

## Testing
I added a few subprocess calls to verify that instantiating via environment variable works as expected.

### DIFF
--- a/imap_processing/__init__.py
+++ b/imap_processing/__init__.py
@@ -13,12 +13,25 @@ __version__ = "0.1.0"
 # packet definitions directory path.
 #
 # This directory is used by the imap_processing package to find the packet definitions.
+import os
 from pathlib import Path
+
+# NOTE: Use a config dictionary, so it is a mutable global object,
+#       otherwise updating previous imports from other modules
+#       wouldn't have been updated globally (for example if referencing a string).
+config = {"DATA_DIR": Path(os.getenv("IMAP_DATA_DIR") or Path.cwd() / "imap-data")}
+"""imap_processing configuration dictionary.
+
+DATA_DIR : This is where the file data is stored and organized by instrument and level.
+    The default location is in the current working directory, but can be
+    set on the command line using the --data-dir option, or through
+    the environment variable IMAP_DATA_DIR.
+"""
 
 # Eg. imap_module_directory = /usr/local/lib/python3.11/site-packages/imap_processing
 imap_module_directory = Path(__file__).parent
 
-instruments = [
+INSTRUMENTS = [
     "codice",
     "glows",
     "hi",
@@ -31,7 +44,7 @@ instruments = [
     "ultra",
 ]
 
-processing_levels = {
+PROCESSING_LEVELS = {
     "codice": ["l0", "l1a", "l1b", "l2"],
     "glows": ["l0", "l1a", "l1b", "l2"],
     "hi": ["l0", "l1a", "l1b", "l1c", "l2"],
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -1,10 +1,14 @@
 """Various utility functions to support creation of CDF files."""
 
-import os
+import logging
+from pathlib import Path
+from typing import Optional
 
 import numpy as np
 import xarray as xr
 from cdflib.xarray import xarray_to_cdf
+
+import imap_processing
 
 
 def calc_start_time(shcoarse_time: int):
@@ -36,7 +40,9 @@ def calc_start_time(shcoarse_time: int):
 
 
 def write_cdf(
-    data: xr.Dataset, description: str = "", mode: str = "", directory: str = ""
+    data: xr.Dataset,
+    description: str = "",
+    directory: Optional[Path] = None,
 ):
     """Write the contents of "data" to a CDF file using cdflib.xarray_to_cdf.
 
@@ -49,16 +55,19 @@ def write_cdf(
 
     Parameters
     ----------
-        data (xarray.Dataset): The dataset object to convert to a CDF
-        description (str): The description to insert into the file name after the
+        data : xarray.Dataset
+            The dataset object to convert to a CDF
+        description : str, optional
+            The description to insert into the file name after the
                             orbit, before the SPICE field.  No underscores allowed.
-        mode (str): Instrument mode
-        directory (str): The directory to write the file to
+        directory : pathlib.Path
+            The directory to write the file to. The default is obtained
+            from the global imap_processing.config["DATA_DIR"].
 
     Returns
     -------
-        str
-            The name of the file created
+        pathlib.Path
+            Path to the file created
     """
     # Determine the start date of the data in the file,
     # based on the time of the first dust impact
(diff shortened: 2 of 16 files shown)


imap_processing/__init__.py
```diff
@@ -4,12 +4,19 @@
...
+# The default location is in the current working directory, but can be
+# set on the command line using the --data-dir option, or through
+# the environment variable IMAP_DATA_DIR.
+DATA_DIR = Path(os.getenv("IMAP_DATA_DIR") or Path.cwd() / "imap-data")
+
 # Eg. imap_module_directory = /usr/local/lib/python3.11/site-packages/imap_processing
 imap_module_directory = Path(__file__).parent
 
-instruments = [
+INSTRUMENTS = [
```
greglucas, line 34: I changed these to a typical GLOBAL caps pattern since they seemed like globals. I can change these back if this is just noise.

imap_processing/cdf/utils.py
```diff
@@ -98,7 +105,21 @@ def write_cdf(
         + date_string
         + f"_v{data.attrs['Data_version']}.cdf"
     )
-    filename_and_path = os.path.join(directory, filename)
+
+    if directory is None:
```
greglucas, line 109: We could get rid of directory altogether and force everyone to go through the `imap_processing.DATA_DIR`? Then in all tests we'd just set that to a temporary location.

pyproject.toml
```diff
@@ -82,5 +83,5 @@ ignore = ["D104", "PLR2004", "S101"]
 [tool.ruff.pydocstyle]
 convention = "numpy"
 
-[project.scripts]
-imap_processing = "imap_processing.run_processing"
+[tool.poetry.scripts]
+imap_cli = "imap_processing.run_processing:main"
```
greglucas: It looks like poetry doesn't respect `project.scripts` so need to change this over to `tool.poetry.scripts`. @laspsandoval you changed this in one of your previous PRs that you then declined. Did you have a different better name for this? I couldn't find it immediately, so happy to hear your thoughts on the naming here.

`imap_processing` is not great because it clashes with our package/directory name so when you are in the current directory it tries to run that folder instead of the installed script.

tech3371 (approved):
It looks good to me. To double check, DATA_DIR is where data is downloaded to when we are using download API, right?

greglucas:
> It looks good to me. To double check, DATA_DIR is where data is downloaded to when we are using download API, right?

Yes, you could set it to wherever you want, and if you wanted to set it to different places that is fine too.

@tech3371 Thoughts on whether this should be the place we **write** files explicitly to? i.e. this comment: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/303#discussion_r1441039324

imap_processing/cdf/utils.py
tech3371: This directory is to save processed data file and then I assume, we upload data from this directory to data bucket using upload API. Will we want to use `imap_processing.DATA_DIR`? If so, then yeah.

greglucas:
I apologize for the churn/expansion here, this is significantly updated since you reviewed @tech3371.

* Use a config dictionary, so it is a mutable global object, otherwise updating previous imports from other modules wouldn't have been updated globally. (changing an item in `imap_processing.config` will show up to someone else referencing the `imap_processing.config` object, but previously if you would have imported `imap_processing.DATA_DIR` pointing to a string, that wouldn't have received an update if somewhere else changed that value)
* Remove <mode> as a filename attribute and option to pass into `write_cdf()`
* Default to the directory structure the project has agreed to
* Create a global test fixture that sets the data directory to a temporary path so no one has to worry about the scope of their test files being written.

imap_processing/__init__.py
```diff
@@ -4,12 +4,22 @@
 # packet definitions directory path.
 #
 # This directory is used by the imap_processing package to find the packet definitions.
+import os
 from pathlib import Path
 
+config = {"DATA_DIR": Path(os.getenv("IMAP_DATA_DIR") or Path.cwd() / "imap-data")}
```
tech3371, line 22: can we summary of this comment in there or is it too much? It might help someone understand why it was done that way.
```
Use a config dictionary, so it is a mutable global object, otherwise updating previous imports from other modules 
wouldn't have been updated globally. (changing an item in imap_processing.config will show up to someone else
 referencing the imap_processing.config object, but previously if you would have imported 
imap_processing.DATA_DIR pointing to a string, that wouldn't have received an update if somewhere else changed
 that value)
```

imap_processing/cdf/utils.py
tech3371: will this go away now?

imap_processing/cdf/utils.py
```diff
@@ -36,7 +40,9 @@ def calc_start_time(shcoarse_time: int):
 
 
 def write_cdf(
-    data: xr.Dataset, description: str = "", mode: str = "", directory: str = ""
+    data: xr.Dataset,
+    description: str = "",
+    directory: Optional[Path] = None,
```
tech3371, line 45: This too will go away?

pyproject.toml
tech3371: Between, Laura's PR and this PR, you may have merge conflict depending on who merge first.

tech3371: I am not good with names. But I agree about choosing a good name for the script file. `imap_processing` or `cli` is not good. one is confusing with package and other is too generic. since we are using `imap_cli`, we could change script file name to be that too unless others have good name suggestion.

tech3371 (approved):
Couple minor questions and comment. It looks good to me still though.

imap_processing/__init__.py
greglucas: Sure, added the first two lines as a comment.

imap_processing/cdf/utils.py
greglucas: I am leaving it for now in case someone wants to explicitly put their file somewhere other than the base data directory. We can remove it later if we want, but I did add an explicit test to make sure this worked.

imap_processing/cdf/utils.py
greglucas: See comment above. Leaving for now. We can remove it later if we want.

pyproject.toml
greglucas: No problem. It will be an easy rebase, so I'm happy to resolve after hers goes in.

Agreed on moving things around further! Maybe do that as a follow-up PR when someone with good naming skills comes in :)

imap_processing/cdf/utils.py
bourque, line 45: Should we also update the `Parameters` section of the docstring to reflect that `directory` is now a `Path` object instead of the `str`? (Looks like that whole section is not formatted correctly too)

imap_processing/tests/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+import imap_processing
+
+
+@pytest.fixture(autouse=True)
+def _set_global_data_dir(tmp_path):
+    """Set the global data directory to a temporary directory."""
+    _original_data_dir = imap_processing.config["DATA_DIR"]
+    imap_processing.config["DATA_DIR"] = tmp_path
+    yield
+    imap_processing.config["DATA_DIR"] = _original_data_dir
```
bourque, line 13: This is a neat way to do this!

imap_processing/tests/cdf/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    # TODO: Update this when launch time is updated
+    launch_time = np.datetime64("2010-01-01T00:01:06.184")
+    assert calc_start_time(0) == launch_time
+    assert calc_start_time(1) == launch_time + np.timedelta64(1, "s")
+
+
+def test_write_cdf(tmp_path):
+    # Set up a fake dataset
+    # lots of requirements on attributes, so depend on SWE for now
+    ds = xr.Dataset(
```
bourque:
```suggestion
    dataset = xr.Dataset(
```

imap_processing/tests/cdf/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        {"Epoch": ("Epoch", [1, 2, 3])},
+        attrs=swe_l1a_global_attrs.output()
+        | {"Logical_source": "imap_test_l1", "Data_version": "01"},
+    )
+    ds["Epoch"].attrs = ConstantCoordinates.EPOCH
+
+    fname = write_cdf(ds, description="test-description")
```
bourque:
```suggestion
    filename = write_cdf(ds, description="test-description")
```

bourque (approved):
Looks great, I like the changes. This will keep us more consistent/organized. I just have a few non-blocking suggestions.

imap_processing/cdf/utils.py
greglucas: Good catch! I updated all of this to be the Numpy docstring style. I think we can use the type-hints to automagically generate these if we wanted to go that route as well and avoid having to manually document these...

imap_processing/tests/cdf/test_utils.py
greglucas: I updated this and `fname` to spell it out. But, I think there is an argument to be made for keeping it `ds` for xarray dataset, `df` for pandas dataframe, `np` for numpy etc... Their documentation uses `ds` pretty universally, so is it better to stick to that or the spelled out name?
https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html

bourque: I am annoyed that they chose to use `ds` instead of `dataset`, but I am totally fine with keeping it `ds`, `fname`, etc. if that is basically a 'standard' for the particular library.

PR #304, opened 2024-01-04
SWAPI Level 1 Implementation

### LINKED ISSUE #187: Get Primary, secondary, coincidence counts per ESA voltage step and time for SWAPI

tech3371:
# Change Summary
Implementation of SWAPI Level 1 algorithm

## Overview
Code for SWAPI Level 1

## New Files
- docs/source/code-documentation/swapi.rst
   - SWAPI document

## Deleted Files

- deleted file 1

## Updated Files
- docs/source/code-documentation/index.rst
   - Add SWAPI to the document
   - Main file where SWAPI level-1 code exists
- imap_processing/swapi/swapi_utils.py
   - Added ENUM class for SWAPI data mode

## Testing
- imap_processing/swapi/tests/test_swapi_l1.py
   - Tests for SWAPI algorithm code


imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+    # SWEEP_TABLE should all be one value.
+    plan_id = full_sweep_sci["PLAN_ID_SCIENCE"].data.reshape(-1, 12)
+    sweep_table = full_sweep_sci["SWEEP_TABLE"].data.reshape(-1, 12)
+
+    mode = full_sweep_sci["MODE"].data.reshape(-1, 12)
+
+    total_sweeps = len(plan_id)
+    bad_data_start_indices = []
+
+    for index in range(total_sweeps):
```
greglucas: If you're looking to filter the data based on this I think you could use array broadcasting for the checks along your new length-12 axis, something like this (not sure about the specific axes though). Then you could filter the rows out and reshape/flatten to get an n_good_sweeps x 12 length 1d array.
```python
bad_data = (sweep_table != sweep_table[:, 0]).all(axis=1)
```

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+    # TODO: add checks for checksum
+
+    # Get bad data sweep start indices and create
+    # sweep indices.
+    # Eg.
+    # From this: [0 1]
+    # To this: [[ 0  1  2  3  4  5  6  7  8  9 10 11]
+    # [12 13 14 15 16 17 18 19 20 21 22 23]]
+    cycle_start_indices = np.unique(bad_data_start_indices)
+    bad_cycle_indices = np.array(
```
greglucas: Use the `None` broadcasting trick you used on line 202 below? Or use the filtering I mentioned above to avoid this altogether.

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+    # This below line gives back row and column index for
+    # compressed flags. Eg.
+    # (array([0, 1, 2, 2]), array([ 4,  1, 20, 51]))
+    compressed_count_indices = np.where(compression_flag == 1)
+
+    for row, col in zip(*compressed_count_indices):
+        if count_data[row, col] == 0xFFFF:  # Overflow
+            new_count[row, col] = -1
+        elif count_data[row, col] != 0xFFFF:
+            new_count[row, col] = count_data[row, col] * 16
```
greglucas:
```suggestion
    compressed_indices = compression_flag == 1
    new_count[compressed_indices] *= 16
    # if the data was compressed and the count was 0xFFFF, mark it as an overflow
    new_count[compressed_indices & (count_data == 0xFFFF)] = -1
```
For the -1 case, you might want to make sure the dtype of the input array and make sure it isn't an unsigned array.

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+        |    PCEM_CNT
+        |    SCEM_CNT
+        |    COIN_CNT
+        |    PCEM_RNG_ST
+        |    SCEM_RNG_ST
+        |    COIN_RNG_ST
+    total_full_sweeps: int
+        Total number of sweeps
+    """
+    all_cem_data = np.zeros((total_full_sweeps, 72))
```
greglucas:
```suggestion
```
This is overwritten below I think?

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+    # [[ 0  1  2  3  4  5  6  7  8  9 10 11]
+    # [ 1  2  3  4  5  6  7  8  9  10  11  12]
+    # [ 2  3  4  5  6  7  8  9  10  11  12  13]]
+
+    # [[ 0  1  2  3  4  5  6  7  8  9 10 11]
+    # [ 1  2  3  4  5  6  7  8  9  10  11  12]
+    # [ 2  3  4  5  6  7  8  9  10  11  12  13]]]
+    # In other word, we grouped each cem's
+    # data by full sweep.
+    current_cem_counts = current_cem_counts.reshape(6, total_full_sweeps, 12)
```
greglucas: I think you can put a -1 in here to allow numpy to do this reshaping for you without needing to pass in the integer of the total_full_sweeps.
```suggestion
    current_cem_counts = current_cem_counts.reshape(6, -1, 12)
```

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+    # [2  2  2  2  2  2]
+    # [3  3  3  3  3  3]
+    # ...
+    # [12  12  12  12  12  12]],
+    # [[2  2  2  2  2  2]
+    # [3  3  3  3  3  3]
+    # ...
+    # [13  13  13  13  13  13]]
+    # ]
+    all_cem_data = np.stack(current_cem_counts, axis=-1)
```
greglucas, line 384: Could you add a comment about shape of this array in addition to the diagram? I'm a bit confused because there are three reshapes in a row happening and it seems like you might be able to do it in one or two instead? In the above reshape are you able to change that to a different order or dimension like (6, -1) to achieve your desired array?

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+
+    Parameters
+    ----------
+    dataset : xarray.Dataset
+        L0 data
+    """
+    # ====================================================
+    # Step 1: Filter full cycle data
+    # ====================================================
+    full_sweep_indices = get_indices_of_full_sweep(sci_dataset)
```
greglucas, line 402: Does "full sweep" only include the good sweeps? I'm wondering if the naming here should be "good_sweep_indices" because I think you could go through a full sweep but have bad data. If that is how the SWAPI team labels them then ignore this comment.

Edit: I see, below you are filtering out the bad data and switching over to good_sweep there.

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+
+    # Filter full sweep data using indices returned from above line
+    full_sweep_sci = sci_dataset.isel({"Epoch": full_sweep_indices})
+
+    # Find indices of bad sweep cycles
+    bad_data_indices = check_for_bad_data(full_sweep_sci)
+    if len(bad_data_indices) > 0:
```
greglucas: See my comment in the routine above where I think you might be able to do some of this filtering in there, similar to what you're doing with the isel and Epoch indices.

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+
+    # ====================================================
+    # Step 2: Process good sweep data
+    # ====================================================
+    total_packets = len(good_sweep_sci["SEQ_NUMBER"].data)
+
+    # It takes 12 sequence data to make one full sweep
+    total_sequence = 12
+    total_full_sweeps = total_packets // total_sequence
+    # These array will be of size (number of good sweep, 72)
```
greglucas, line 420: Thanks for this detailed comment!

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,520 @@
...
+    # Epoch time. Should be same dimension as number of good sweeps
+    epoch_time = good_sweep_sci["Epoch"].data.reshape(total_full_sweeps, 12)[:, 0]
+    epoch_time = xr.DataArray(
+        epoch_time,
+        name="Epoch",
+        dims=["Epoch"],
+    )
+
+    # There are 72 energy steps
+    energy = xr.DataArray(np.arange(72), name="Energy", dims=["Energy"])
```
greglucas, line 445: We should verify what they want this called. I would vote for "step" or "Energy Step" instead of energy because there aren't units attached to the integer values.

imap_processing/swapi/tests/test_swapi_l1.py
```diff
@@ -1 +1,211 @@
...
+    bad_data_indices = check_for_bad_data(ds)
+    assert len(bad_data_indices) == 0
+
+    # Check for bad MODE data.
+    # This test returns this indices because MODE has 0, 1 values
+    # for the first two sweeps.
+    # TODO: update test when we update MODE from HVENG to HVSCI
+    ds["MODE"] = xr.DataArray(np.repeat(np.arange(total_sweeps), 12))
+    bad_data_indices = check_for_bad_data(ds)
+    assert np.all(bad_data_indices == np.arange(0, 24))
```
greglucas: `np.testing.assert_array_equal()` might be nice in all these assertions because it has pretty prints of the bad indices if things fail.

imap_processing/swapi/tests/test_swapi_l1.py
```diff
@@ -1 +1,211 @@
...
+    """Test for decompress count"""
+    # Test compressed and not overflow case
+    raw_value = np.array([[12]])
+    compression_flag = np.array([[1]])
+    expected_value = raw_value[0][0] * 16
+    returned_value = decompress_count(raw_value, compression_flag)
+    assert returned_value[0][0] == expected_value
+
+    # Test compressed and overflow case
+    raw_value = np.array([[0xFFFF]])
```
greglucas: Put these all together into a single array to make sure that is also handled well in the mixed-use version.

```python
# compressed + no-overflow, compressed + overflow, no compression
raw_values = np.array([[12, 0xFFF, 12]])
compression_flag = np.array([[1, 1, 0]])
expected = np.array([[12*16, -1, 12]])
assert_array_equal(

File: imap_processing/swapi/tests/test_swapi_l1.py [hunk 13]
```diff
@@ -1 +1,211 @@
...
+    assert np.all(start_indices == [12])
+
+    ds["SEQ_NUMBER"] = np.arange(3, 29)
+    start_indices = find_sweep_starts(ds)
+    assert np.all(start_indices == [])
+
+
+def test_get_full_indices():
+    """Test for correct full sweep indices"""
+    time = np.arange(14)
```
greglucas: Maybe test that you get two sweeps when expected? One is a bit easier to get right on accident.
```suggestion
    time = np.arange(26)
```

greglucas (commented):
Looking good. I've got some minor numpy suggestions, but overall this is looking like a nice cleanup!

imap_processing/swapi/l1/swapi_l1.py
tech3371: Yeah, first we filter out only full sweep data. Then filter out bad data. Then process good sweep data. that's what I got from their discussion.

imap_processing/swapi/l1/swapi_l1.py
tech3371: I do that below like you are thinking. Is that what you mean? Did you want me to change anything here?

imap_processing/swapi/l1/swapi_l1.py
tech3371: I was trying stay consistent with I used for SWE.

imap_processing/swapi/l1/swapi_l1.py
tech3371: we discussed about this on Slack. Will push changes soon.

imap_processing/swapi/l1/swapi_l1.py
tech3371: I added comment about the shape.

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -1 +1,505 @@
...
+
+    Parameters
+    ----------
+    full_sweep_sci : xarray.Dataset
+        Science data that only contains full sweep data.
+
+    Returns
+    -------
+    List
+        List of sweep indices of bad data
```
greglucas:
```suggestion
    numpy.ndarray
        Sweep indices of bad data
```

imap_processing/swapi/l1/swapi_l1.py
greglucas: I was wondering if you wanted to do the filtering instead of just the indices in the other routine. But I think it is fine here.

A few other thoughts here though that might make it easier to follow.
* Return the good_indices from the helper function and log the bad stuff there, just keep this area for filtering the datasets, not also calculating other indices. Push this logic up.
* Then I think you can do `good_sweep_sci = full_sweep_sci.isel()` no matter what is returned, not needing the if/else.

imap_processing/swapi/tests/test_swapi_l1.py
```diff
@@ -1 +1,344 @@
...
+    assert returned_value[0][0] == expected_value
+
+    # Test not compressed case
+    raw_value = np.array([[12]])
+    compression_flag = np.array([[0]])
+    expected_value = 12
+    returned_value = decompress_count(raw_value, compression_flag)
+    assert returned_value[0][0] == expected_value
+
+    # compressed + no-overflow, compressed + overflow, no compression
```
greglucas: Just an FYI that I think you can remove all of the above 3-blocks since they are really duplicates of this test that combines them all, and this is more robust in a sense.

imap_processing/swapi/tests/test_swapi_l1.py
```diff
@@ -1 +1,344 @@
...
+    ds = xr.Dataset({"SEQ_NUMBER": sequence_number}, coords={"Epoch": time})
+
+    start_indices = find_sweep_starts(ds)
+    np.testing.assert_array_equal(start_indices, [0, 12])
+
+    ds["SEQ_NUMBER"].data[:12] = np.arange(3, 15)
+    start_indices = find_sweep_starts(ds)
+    np.testing.assert_array_equal(start_indices, [12])
+
+    ds["SEQ_NUMBER"] = np.arange(3, 29)
```
greglucas: Can you add a quick note about what this is testing? Just glancing at it I would have expected it to identify sequence starts...

greglucas (approved):
This is a really nice example of the xarray/numpy data manipulations and commenting the progress through this. Thanks for taking the time to investigate how to get rid of those for-loops too!

imap_processing/swapi/l1/swapi_l1.py
tech3371: Yes. I tried that and that worked out well and it was easier to follow.

PR #306, opened 2024-01-05
MNT: Switch from black to ruff

greglucas:
# Change Summary

## Overview

This is one less dependency, and ruff is faster and produces the same diff currently.

reference: https://docs.astral.sh/ruff/formatter/


bourque (approved):
Cool! I didn't know that `ruff` did that. Looks good, and thanks for updating the documentation too!

PR #309, opened 2024-01-06
PERF: Speed up bitstring conversions in idex

greglucas:
# Change Summary

## Overview

Bitstring was allocating many new objects which added significant overhead. We can just use Python's native int conversions and indexing into strings to achieve the same result.

performance script:
`python -m cProfile -o orig.prof -m pytest imap_processing/idex/tests`

**current version**
image

**new version**
image

There are originally over 3M calls to ba2int and over 1M calls to _readlist, versus 228k and 0 in the new version.

### DIFF
--- a/imap_processing/idex/idex_packet_parser.py
+++ b/imap_processing/idex/idex_packet_parser.py
@@ -821,11 +821,15 @@ class RawDustEvent:
 
         The very last 4 numbers are bad usually, so remove those
         """
-        w = bitstring.ConstBitStream(bin=waveform_raw)
         ints = []
-        while w.pos < len(w):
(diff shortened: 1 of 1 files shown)


bourque (approved):
The more you know! Thanks Greg. I'm curious what prompted you to discover this?

greglucas:
> The more you know! Thanks Greg. I'm curious what prompted you to discover this?

I was running a lot of tests and didn't understand how we could be so slow... I think there will likely be others here as well, but I've always had my eye on bitstring and cdflib being pure python and slow for our use-cases, so it sort of reaffirmed this to me when running cProfile.

PR #310, opened 2024-01-09
PERF: Remove buffer size limitation

greglucas:
# Change Summary

## Overview

No need to explicitly overwrite the default size.

Similar to #309 and seeing things got slower, but a different solution here.

`python -m cProfile -o orig.prof -m pytest imap_processing/tests/glows`
Goes from 60s down to 20s on my local machine.

From the docs: https://space-packet-parser.readthedocs.io/en/latest/autoapi/space_packet_parser/parser/index.html

> buffer_read_size_bytes (int, Optional) – Number of bytes to read from e.g. a BufferedReader or socket binary data source on each read attempt. Default is 4096 bytes.

I wonder if this should be set globally for everyone if we do want to update the size of the reads...

### DIFF
--- a/imap_processing/glows/l0/decom_glows.py
+++ b/imap_processing/glows/l0/decom_glows.py
@@ -58,10 +58,7 @@ def decom_packets(
 
     with open(packet_file_path, "rb") as binary_data:
         try:
-            glows_packets = glows_parser.generator(
-                binary_data,
-                buffer_read_size_bytes=5790778,
-            )
+            glows_packets = glows_parser.generator(binary_data)
 
             for packet in glows_packets:
                 apid = packet.header["PKT_APID"].derived_value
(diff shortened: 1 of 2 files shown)


maxinelasp:
I'm still a little concerned about this, because it's not clear to me what the right value should be. But I'm ok with using the default for now until we figure out if we need to overwrite this value. Just want to note that the test file is very small, so it isn't necessarily a perfect test for this value.

maxinelasp (approved):

greglucas:
Great question, I just asked this to the space packet parser Slack channel.

greglucas:
@maxinelasp , I just merged this in per our conversation that it should be closer to packet size than file size, but the suggestion was to just leave it default until we find that limiting.

I think follow-up work could be to set this value to the fixed packet size for instruments if that makes sense or is needed. Maybe there are some utilities on the xtce parser to get that value for us before reading the stream...

PR #311, opened 2024-01-09
TST: Re-use test fixtures within the module if possible

greglucas:
# Change Summary

## Overview

This speeds things up by not having to read the data so often.

Luckily or unluckily GLOWS was the only one that was easy to update and tests still passed. IDEX mutated their read-in data, so had to it has to be re-read each time, and Ultra was using function-level fixtures which can't be increased in scope, but seemingly those could be updated as well.


maxinelasp (approved):

maxinelasp:
Thanks for catching this, I'll keep it in mind for future tests too!

PR #313, opened 2024-01-12
MNT: Rename description to descriptor in filename variable

greglucas:
# Change Summary

## Overview

Change "description" to "descriptor" for the filenames.

Additionally, `descriptor` is not optional anymore, so set a default of `sci` in the field. Or do we want to just force everyone to input this when calling `write_cdf()`?

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -41,7 +41,7 @@ def calc_start_time(shcoarse_time: int):
 
 def write_cdf(
     data: xr.Dataset,
-    description: str = "",
+    descriptor: str,
     directory: Optional[Path] = None,
 ):
     """Write the contents of "data" to a CDF file using cdflib.xarray_to_cdf.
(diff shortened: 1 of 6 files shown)


imap_processing/cdf/utils.py
```diff
@@ -41,7 +41,7 @@ def calc_start_time(shcoarse_time: int):
 
 def write_cdf(
     data: xr.Dataset,
-    description: str = "",
+    descriptor: str = "sci",
```
tech3371: To you question, I think we shouldn't make this optional input now since this field is pretty generic and can be so different based on individual instrument. 

Or if we put it in dataset.attrs, then we can abstract from that too. But could be up for discussion.

tech3371 (approved):

imap_processing/cdf/utils.py
greglucas: > Or if we put it in dataset.attrs, then we can abstract from that too. But could be up for discussion.

I like this idea, but I am also not sure how it would look because the instrument already has a "Descriptor" which is: "SWE>Solar Wind Electron" which isn't what we want unfortunately.

greglucas: OK, I just made it non-optional now. We can look into the attrs idea later.

tech3371 (approved):

imap_processing/cdf/utils.py
tech3371: that sounds good. I can do that in follow-up PR.

PR #314, opened 2024-01-12
MNT: Cleanup ruff codes

greglucas:
# Change Summary

## Overview

Update the `pyproject.toml` ruff flags that we are ignoring.

This looks like a large diff, but is largely a reorganization within the IDEX code. There was a large trigger_dict block that was being set in each object, but it was more of a global constant with information. I pulled that outside the scope of the object and index into the NamedTuples later instead. I also updated it so that the "calc_*" functions don't return things now, but rather set the values to the attributes directly, so I changed the name of those to make it clear we are setting things.

### DIFF
--- a/imap_processing/idex/idex_packet_parser.py
+++ b/imap_processing/idex/idex_packet_parser.py
@@ -26,6 +26,429 @@ SCITYPE_MAPPING_TO_NAMES = {
     64: "Ion_Grid",
 }
 
+"""
+Creates a large dictionary of values from the FPGA header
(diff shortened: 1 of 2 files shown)


bourque (approved):
Thanks Greg! Looks good to me.

greglucas:
I saw that the MAPPING global was actually never used. I turned that into an Enum now in a separate following commit if you wanted to check that at all too.

bourque:
Looks good!

PR #317, opened 2024-01-19
ENH: Add i/o capability to communicate with the data archive

greglucas:
# Change Summary

## Overview
This adds the ability to download, query, and upload files from the data archive. This is currently opinionated and requires all files to be located within the `DATA_DIR` global configuration directory. I purposefully left off a `url=` argument to the functions to try and make us go through a global configuration url. If people are against that we can remove it. But, it seems like it might be nice to start enforcing some of these limits rather than keeping too many options available.

Outside of this scope: Adding it to the cli, which would make sense after this is decided upon.

## New Files

- io.py
   - Contains the download(), query(), upload() functions

## Updated Files

- __init__.py
   - added API_URL to the config dictionary

## Testing

Lots of mock url testing. But, it could be better once we know what our query endpoints will return and things like that. I've left that off for now though since it is still somewhat TBD and we can update/improve these later then.

### DIFF
--- a/imap_processing/__init__.py
+++ b/imap_processing/__init__.py
@@ -19,9 +19,13 @@ from pathlib import Path
 # NOTE: Use a config dictionary, so it is a mutable global object,
 #       otherwise updating previous imports from other modules
 #       wouldn't have been updated globally (for example if referencing a string).
-config = {"DATA_DIR": Path(os.getenv("IMAP_DATA_DIR") or Path.cwd() / "imap-data")}
+config = {
+    "API_URL": "https://api.dev.imap-mission.com",
+    "DATA_DIR": Path(os.getenv("IMAP_DATA_DIR") or Path.cwd() / "imap-data"),
+}
 """imap_processing configuration dictionary.
 
+API_URL : This is the URL of the data archive API.
 DATA_DIR : This is where the file data is stored and organized by instrument and level.
     The default location is in the current working directory, but can be
     set on the command line using the --data-dir option, or through
--- a/imap_processing/__init__.py
+++ b/imap_processing/__init__.py
@@ -20,12 +20,12 @@ from pathlib import Path
 #       otherwise updating previous imports from other modules
 #       wouldn't have been updated globally (for example if referencing a string).
 config = {
-    "API_URL": "https://api.dev.imap-mission.com",
+    "DATA_ACCESS_API_URL": "https://api.dev.imap-mission.com",
     "DATA_DIR": Path(os.getenv("IMAP_DATA_DIR") or Path.cwd() / "imap-data"),
 }
 """imap_processing configuration dictionary.
 
-API_URL : This is the URL of the data archive API.
+DATA_ACCESS_API_URL : This is the URL of the data access API.
 DATA_DIR : This is where the file data is stored and organized by instrument and level.
     The default location is in the current working directory, but can be
     set on the command line using the --data-dir option, or through
--- a/imap_processing/io.py
+++ b/imap_processing/io.py
@@ -48,7 +48,7 @@ def download(filepath: str) -> Path:
         return destination
 
     # encode the query parameters
-    url = f"{imap_processing.config['API_URL']}"
+    url = f"{imap_processing.config['DATA_ACCESS_API_URL']}"
     url += f"/download?{urlencode({'filename': filepath})}"
     logger.info("Downloading file %s from %s to %s", filepath, url, destination)
 
@@ -106,7 +106,8 @@ def query(
     query_params = {key: value for key, value in locals().items() if value is not None}
     if not query_params:
         raise ValueError("At least one query parameter must be provided")
-    url = f"{imap_processing.config['API_URL']}/query?{urlencode(query_params)}"
+    url = f"{imap_processing.config['DATA_ACCESS_API_URL']}"
+    url += f"/query?{urlencode(query_params)}"
 
     logger.debug("Querying data archive for %s with url %s", query_params, url)
     request = urllib.request.Request(url, method="GET")
@@ -141,7 +142,7 @@ def upload(filepath: Path) -> None:
     # Strip off the data directory to get the upload path + name
     upload_name = str(filepath.relative_to(imap_processing.config["DATA_DIR"]))
 
-    url = f"{imap_processing.config['API_URL']}"
+    url = f"{imap_processing.config['DATA_ACCESS_API_URL']}"
     url += f"/upload?{urlencode({'filename': upload_name})}"
     logger.debug("Uploading file %s to %s", filepath, url)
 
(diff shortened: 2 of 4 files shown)


imap_processing/io.py
```diff
@@ -0,0 +1,162 @@
+"""Input/output capabilities for the IMAP data processing pipeline."""
```
tech3371, line 1: We will use this file to make API call from within batch job, right? Looks good to me.

tech3371 (approved):

imap_processing/io.py
greglucas: Yes, that is the vision at least. I think we'll want to update `write_cdf()` or chain this with it somehow as an example to say:
```python
# Download any dependency data that was passed in
for dependency_file in environment_dependency_files:
    imap_processing.io.download(dependency_file)
# Should have all needed files now to kick off the processing job
imap_processing.swe.science_l1a()
created_cdf_file = imap_processing.cdf.utils.write_cdf(swe_l0_dataset)
# upload file to our archive
imap_processing.io.upload(created_cdf_file)
```
Which will hopefully be wrapped up in our CLI scripts or abstracted out in other ways too.

tech3371: yeah, that sounds like a good plan.

imap_processing/__init__.py
```diff
@@ -19,9 +19,13 @@
...
 #       otherwise updating previous imports from other modules
 #       wouldn't have been updated globally (for example if referencing a string).
-config = {"DATA_DIR": Path(os.getenv("IMAP_DATA_DIR") or Path.cwd() / "imap-data")}
+config = {
+    "API_URL": "https://api.dev.imap-mission.com",
+    "DATA_DIR": Path(os.getenv("IMAP_DATA_DIR") or Path.cwd() / "imap-data"),
+}
 """imap_processing configuration dictionary.
 
+API_URL : This is the URL of the data archive API.
```
tech3371: During SDC tag-up, we were discussion about what to call this API and its endpoints. Many liked the name `data access API`. May be we should call that here too instead of `data archive API`?

greglucas: Sure, I can update the docstring to "data access API".

Should we also change this to `DATA_ACCESS_API` as the key? Or is `API`, `API_URL` (current), or something else short better? I'm open to suggestions here.

bourque: My vote would be `DATA_ACCESS_API_URL` for consistency with what it is named ("Data Access API") and what the key is storing (a URL).

imap_processing/io.py
bourque, line 1: I think naming this `io.py` is fine, but just wanted to mention `api_client.py` or `api_wrapper.py` as other possibilities.

imap_processing/io.py
bourque, line 1: Since this script could be used by external users, it might be nice to expand the module docstrings here a bit and include some examples of how to use the script.

imap_processing/io.py
```diff
@@ -0,0 +1,162 @@
...
+        # Construct the directory structure from the filename
+        # This is for science files that contain the directory structure in the filename
+        # Otherwise, we assume the full path to the file was given
+        parts = filepath.split("_")
+        instrument = parts[1]
+        datalevel = parts[2]
+        startdate = parts[4]
+        year = startdate[:4]
+        month = startdate[4:6]
+        destination = destination / instrument / datalevel / year / month
```
bourque, line 41: Probably beyond the scope of this PR, but it might be useful to pull this out into its own function? I can foresee that we might use this in other parts of our code.

imap_processing/io.py
```diff
@@ -0,0 +1,162 @@
...
+    version : str, optional
+        Data version
+
+    Returns
+    -------
+    list
+        List of files matching the query
+    """
+    # locals() gives us the keyword arguments passed to the function
+    # and allows us to filter out the None values
```
bourque, line 108: Nice use of inline comments here!

imap_processing/tests/test_io.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+@pytest.fixture()
+def mock_download_urlopen():
+    """Mock urlopen to return a file-like object."""
+    mock_data = b"Mock file content"
+    with patch("urllib.request.urlopen") as mock_urlopen:
+        mock_response = mock_urlopen.return_value.__enter__.return_value
+        mock_response.read.return_value = mock_data
+        yield mock_urlopen
```
bourque, line 19: This appears to be repetitive with `mock_query_urlopen()` and `mock_upload_urlopen()`. Is there a way you could pull it out and re-use it?

imap_processing/tests/test_io.py
bourque, line 1: Great test coverage!

bourque (commented):
This is looking good! I just have some minor comments/questions.

imap_processing/io.py
sdhoyt, line 1: Is this meant to replace the sds-api.py script?

imap_processing/__init__.py
greglucas: Changed to DATA_ACCESS_API_URL now.

imap_processing/io.py
greglucas: I prefer shorter names, but can change if you or anyone else has a strong preference for either of those.

imap_processing/io.py
greglucas: I was thinking about holding off on adding those descriptions until we decided on using this as a script versus just functions, because right now it can't actually be used as a script. Thoughts on whether we want to add a script or other info here now versus later?

imap_processing/io.py
greglucas: I agree, I think we should pull it out in a later smaller PR or when we start adding this same code in other places.

imap_processing/tests/test_io.py
greglucas: Thanks for suggesting that! Updated to a single mock_url now, and then a helper function to populate the data return value. Let me know if that makes sense now.

imap_processing/io.py
greglucas: Great question. Currently, it is just functions to help out during the processing. We could go one of two ways. Pull this content out and put it over into the `sds-api.py` script, and depend on that here. I sort of avoided that so we didn't have to worry about releasing versions and syncing things like that between the two repositories. The other would be to add this to our cli or make a new one that could be installed alongside the package.

The major con to turning this into a cli is that a user has to bring along all the `imap_processing` dependencies, whereas the `sds-api.py` script/helper repo could be slimmer and not require that.

Thoughts/preferences on this?

sdhoyt: What syncing would we need to do between the two repositories if we moved this to sds-api? Wouldn't that have its own versioning and be a separate module that we import into imap_processing?

greglucas: syncing releases. So we'd need to make a pypi release of that and this at the same time so they could depend on the right versions of each other.

greglucas:
Created a new repository imap-data-access for this

PR #319, opened 2024-01-30
Lo Star Sensor and Science Counts data classes with bit decompression

sdhoyt:
# Change Summary

## Overview
This PR adds data classes for the IMAP-Lo Start Sensor and Science Counts L1A data. This includes parsing the binary data in python and decompressing the bits from 8 to 12, 8 to 16, or 12 to 16.

## New Dependencies
None

## New Files
- science_counts.py
   - Science count data class to parse binary and decompress integers
- star_sensor.py
   - Star sensor data class to parse binary and decompress integers
- bit_decompression.py
   - utility function for decompressing integers
- 8_to_12_bit.csv
   - CSV with 8 to 12 bit decompression conversion
- 8_to_16_bit.csv
   - CSV with 8 to 16 bit decompression conversion
- 12_to_16_bit.csv
   - CSV with 12 to 16 bit decompression conversion

## Deleted Files
None

## Updated Files
- Star Sensor XTCE
   - updated xtce to read in all data fields as one binary chunk.
- re-organized files into directories

## Testing
Add unit tests:
- test_bit_decompression.py
- test_science_counts.py
   - needed to comment out init to run unit tests because no data is available. Also only testing for resulting field shapes until data is available.
- test_star_sensor.py
   - needed to comment out init to run unit tests because no data is available. Also only testing for resulting field shapes until data is available.


docs/source/code-documentation/lo.rst
```diff
@@ -15,5 +15,8 @@ The L0 code to decommutate the CCSDS packet data can be found below.
     :template: autosummary.rst
     :recursive:
 
-    l0.lo_base
-    l0.science_direct_events
+    l0.utils.lo_base
+    l0.data_classes.science_direct_events
+    l0.data_classes.science_counts
+    l0.data_classes.star_sensor
+    l0.utils.bit_decompression
```
greglucas: I think you might be able to let the recursive nature take over if you wanted to leave those items nested and so you don't have to keep adding new entries here too.

```suggestion
    l0.data_classes
    l0.utils
```

docs/source/conf.py
```diff
@@ -99,6 +99,9 @@
     (r"py:.*", r".*CoDICECompression.*"),
     (r"py:.*", r".*LoBase.*"),
     (r"py:.*", r".*ScienceDirectEvents.*"),
+    (r"py:.*", r".*ScienceCounts.*"),
+    (r"py:.*", r".*StarSensor.*"),
+    (r"py:.*", r".*Decompress.*"),
```
greglucas: Could you change this to `r"*.l0.data_classes.*"` to cover all the cases?

imap_processing/lo/l0/data_classes/science_counts.py
```diff
@@ -0,0 +1,285 @@
...
+            bitstream, 3360, Decompress.DECOMPRESS8TO16, (60, 7)
+        )
+        # The OXYGEN data in the binary is 3360 bits long and each field is
+        # compressed to 8 bits. This data uses 8 to 16 bit decompression.
+        # OXYGEN is a 60, 7 matrix containing Azimuth (60) and Energy (7)
+        self.OXYGEN = self._parse_section(
+            bitstream, 3360, Decompress.DECOMPRESS8TO16, (60, 7)
+        )
+
+    def _parse_section(self, bitstream, section_length, decompression, data_shape):
```
greglucas: Should this and `_extract_binary` be class methods or helper functions since they don't do anything with or on self?

imap_processing/lo/l0/data_classes/science_counts.py
```diff
@@ -0,0 +1,285 @@
...
+        """Extract and decompress science count binary data section."""
+        data_list = list()
+        # stop the bitstream once you reach the end of the data section.
+        bit_stop = bitstream.pos + section_length
+        while bitstream.pos < bit_stop:
+            # Extract the 12 bit long field from the binary chunk and get the integer
+            extracted_integer = bitstream.read(bit_length).uint
+            # The Star Sensor packet uses a 12 to 8 bit compression
+            decompressed_integer = decompress_int(extracted_integer, decompression)
+            data_list.append(decompressed_integer)
```
greglucas: I'm guessing that this could potentially be a slow point in the algorithm. The bitstream reading/manipulations were slow on IDEX, so they could possibly be here as well.
xref: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/309

I think this is fine to get in as-is though and improve performance later, but I'm commenting just in case something sticks out to you as being easy to fix.

Another idea is to try and use `np.frombuffer()` and unpack/repack bits.
```python
array_of_bits = np.frombuffer(bitstring.bytes, 'u1') - ord('0')
np.packbits(array_of_bits[16:24])
# Can also turn it into u16 I think too .view(np.uint16)
```

imap_processing/lo/l0/data_classes/science_counts.py
```diff
@@ -0,0 +1,285 @@
...
+@dataclass
+class ScienceCounts(LoBase):
+    """L1A Science Count data class.
+
+    The Science Counts class handles the parsing
+    and decompression of L0 to L1A data.
+
+    Attributes
+    ----------
+    SHCOARSE: int
```
greglucas: All of these should have a space before the colon I think. I wonder if that was causing issues with the doc build.
https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring
```suggestion
    SHCOARSE : int
```

imap_processing/lo/l0/data_classes/science_counts.py
```diff
@@ -0,0 +1,285 @@
...
+        Spacecraft time.
+    SCI_CNT: str
+        science count compressed binary data
+    START_A: numpy.ndarray
+        Single rates for electon, anode A. 2D array Azimuth(6), Energy(7)
```
greglucas: These descriptions and shapes are really nice, thank you!

imap_processing/lo/l0/data_classes/science_counts.py
```diff
@@ -0,0 +1,285 @@
...
+    POS2: np.ndarray
+    POS3: np.ndarray
+    HYDROGEN: np.ndarray
+    OXYGEN: np.ndarray
+
+    def __init__(self, packet, software_version: str, packet_file_name: str):
+        super().__init__(software_version, packet_file_name, CcsdsData(packet.header))
+        self.parse_data(packet)
+        self._parse_binary()
+        pass
```
greglucas:
```suggestion
```

imap_processing/lo/l0/data_classes/science_counts.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def _parse_binary(self):
+        """Parse the science count binary chunk for each section of data."""
+        # make a bit stream containing the binary for the entire
+        # chunk of Science Count packet data
+        bitstream = ConstBitStream(bin=self.SCI_CNT)
```
greglucas: Is `self.SCI_CNT` a Python string, or is it a bitstring initially? You might be able to just re-use the self.SCI_CNT directly here rather than creating a new stream?

imap_processing/lo/l0/data_classes/science_counts.py
```diff
@@ -0,0 +1,285 @@
...
+
+        # The START_A data in the binary is 504 bits long and each field is
+        # compressed to 12 bits. This data uses 12 to 16 bit decompression.
+        # START_A is a 6, 7 matrix containing Azimuth (6) and Energy (7)
+        self.START_A = self._parse_section(
+            bitstream, 504, Decompress.DECOMPRESS12TO16, (6, 7)
```
greglucas: I like the use of the Enums here!

Can you calculate the num_bits in the parse_section method from the other values you pass in to avoid passing in `504` here. i.e. `num_bits = 12 * (6 * 7)`.

imap_processing/lo/l0/utils/bit_decompression.py
```diff
@@ -0,0 +1,51 @@
...
+            + "Decompress.DECOMPRESS8TO12"
+        )
+
+    # load the decompression table for the method specified
+    data = np.loadtxt(
+        Path(__file__).parent
+        / f"../decompression_tables/{decompression.value}_bit.csv",
+        delimiter=",",
+        skiprows=1,
+    )
```
greglucas: This disk i/o will be called a lot here, so it might help to only load the table once and store it for later use somewhere.
```python
decompression_lookup = {DecompressionEnum: np.load_txt(...) for enum in Decompress}
```

You can bring the `..` up a level to call `.parent` again.
```suggestion
    # load the decompression table for the method specified
    data = np.loadtxt(
        Path(__file__).parent.parent
        / f"decompression_tables/{decompression.value}_bit.csv",
        delimiter=",",
        skiprows=1,
    )
```

imap_processing/tests/lo/test_bit_decompression.py
```diff
@@ -0,0 +1,56 @@
...
+
+from imap_processing.lo.l0.utils.bit_decompression import Decompress, decompress_int
+
+
+def test_decompress_int_8_to_16_bit():
+    """Test decompressing integer from 8 to 16 bits."""
+    ## Arrange
+    decompression = Decompress.DECOMPRESS8TO16
+    compressed_value = 209
+    expected_decompressed_int = 9537
```
greglucas: I'd like to see more parameterization in the tests to avoid the repetition if possible. For example, these cases could be combined here:
```suggestion
@pytest.mark.parameterize(("decompression", "compressed_value", "expected_decompressed_int"),
    [(Decompress.DECOMPRESS8TO16, 209, 9537),
      (Decompress.DECOMPRESS8TO12, 205, 1229),
      (Decompress.DECOMPRESS12TO16, 4068, 62908),
       ])
def test_decompress_int(decompression, compressed_value, expected_decompressed_int):
    """Test decompressing integer from 8 to 16 bits."""
```

imap_processing/tests/lo/test_science_counts.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Going to wait until validation data is available to test that
+# the values are correct. Currently only check for the shape
+# of the resulting data.
+@pytest.mark.skip(reason="no data to initialize with")
+def test_science_counts():
+    """Test the science counts parsing, decompression, and shaping."""
+    ## Arrange
+    sc = ScienceCounts("fake_packet", "version", "packet_name")
+    sc.SCI_CNT = "".join(choice("01") for _ in range(26880))  # noqa: S311
```
greglucas: I'd avoid any random we can, especially if you're just going to skip it for now. Perhaps just use 01 for now (or all 0s)? (Same for the other test file)
```suggestion
    sc.SCI_CNT = "01" * 14440
```

imap_processing/lo/l0/data_classes/science_counts.py
sdhoyt: SCI_CNT is a python string. It gets set to the data.derived_value from the space_packet_parser output

imap_processing/lo/l0/data_classes/science_counts.py
sdhoyt: Wow, yeah I completely missed that lol

imap_processing/lo/l0/utils/bit_decompression.py
sdhoyt: Ok sounds good. I'll make a new function to load all the tables and call that at the top of the ScienceCounts and StarSensor files, so it only needs to be read once per packet.

imap_processing/lo/l0/data_classes/science_counts.py
greglucas: Can you add a comment for why things are commented out currently if there is a TODO still?

imap_processing/lo/l0/data_classes/science_counts.py
greglucas: Thanks for clarifying! I was hoping that was the case and I think we might be able to do some other faster manipulations rather than going back to bitstring within these functions then. We can chat more later about this.

imap_processing/lo/l0/utils/bit_decompression.py
```diff
@@ -0,0 +1,59 @@
...
+    """Load all decompression tables into a dictionary."""
+    # load all decompression tables
+    return {
+        enum: np.loadtxt(
+            Path(__file__).parent.parent / f"decompression_tables/{enum.value}_bit.csv",
+            delimiter=",",
+            skiprows=1,
+        )
+        for enum in Decompress
+    }
```
greglucas: I _think_ you can make this an attribute of the module and then not have to re-load them on every packet even, just once on module import.

```suggestion
DECOMPRESSION_TABLES = {
        enum: np.loadtxt(
            Path(__file__).parent.parent / f"decompression_tables/{enum.value}_bit.csv",
            delimiter=",",
            skiprows=1,
        )
        for enum in Decompress
    }
"""Decompression tables stored as a lookup dictionary."""
```

imap_processing/lo/l0/utils/bit_decompression.py
```diff
@@ -0,0 +1,59 @@
...
+        Decompress.DECOMPRESS12TO16,
+    ]
+    if decompression not in valid_decompression:
+        raise ValueError(
+            "Invalid decompression method. Must be one of the following Enums: "
+            + "Decompress.DECOMPRESS8TO12, Decompress.DECOMPRESS8TO12, "
+            + "Decompress.DECOMPRESS8TO12"
+        )
+    data = decompression_lookup[decompression]
+    return int(data[compressed_value][1])
```
greglucas: I think you can get an int through `loadtxt(..., dtype=int)` directly.
```suggestion
    return data[compressed_value][1]
```

imap_processing/tests/lo/test_science_counts.py
```diff
@@ -0,0 +1,41 @@
+import pytest
+
+from imap_processing.lo.l0.data_classes.science_counts import ScienceCounts
+
+
+# Going to wait until validation data is available to test that
+# the values are correct. Currently only check for the shape
+# of the resulting data.
+@pytest.mark.skip(reason="no data to initialize with")
```
greglucas: Is it worth doing these shape checks for now even? I think that might be useful just to test setting up the classes and reading in your fake data. Just adding a TODO: comment indicating it isn't finished yet?

imap_processing/tests/lo/test_star_sensor.py
```diff
@@ -0,0 +1,21 @@
+from random import choice
```
greglucas: Remove the random here as well.

greglucas (approved):
Some additional minor comments, but otherwise looks good to me.

imap_processing/tests/lo/test_science_counts.py
sdhoyt: I'm skipping these because the data class will crash without space_packet_parser packet output which I don't currently have. The unit tests will run if I comment out the init method for the data class though. So my plan was to comment that function out to test locally and skip the tests when I push to avoid GitHub action errors.

greglucas: xfail might be more appropriate here as well, but not a huge deal either way.
https://pytest.org/en/7.4.x/how-to/skipping.html
> An xfail means that you expect a test to fail for some reason. A common example is a test for a feature not yet implemented, or a bug not yet fixed. When a test passes despite being expected to fail (marked with pytest.mark.xfail), it’s an xpass and will be reported in the test summary.

PR #320, opened 2024-01-30
Fixed typo in pypi URL

bourque:
PyPI expects a hyphen instead of an underscore (though somehow an underscore still works!)


greglucas (approved):

bourque:
@all-contributors please add @bourque for maintenance

PR #330, opened 2024-02-01
Imap cli updates

tech3371:
# Change Summary
This is not meant to be merge. I will create issues to address some of the things we need to think about and address post demo. Opening this PR so that you can see what I push to the ECR image that batch job uses.


examples/Dockerfile.imap_api
```diff
@@ -0,0 +1,24 @@
...
+RUN pip install imap-data-access
+
+# Create the /mnt/spice directory
+RUN mkdir -p /mnt/spice
+
+WORKDIR /app/data
+ENV IMAP_DATA_DIR /app/data
+
+# Define the entrypoint of the container
+ENTRYPOINT ["imap_cli"]
```
greglucas: I think we can trim this container RUNs down? If we aren't using `/mnt/spice` we should comment that out/remove it until it is needed.

NOTE: You should add `imap-data-access` as a dependency in `pyproject.toml` then you won't need to install it in the container and it will be brought along as a dependency when you install imap_processing.

```suggestion
# Dockerfile that installs imap_processing and its dependencies
FROM public.ecr.aws/docker/library/python:3.10-slim

RUN pip install imap_processing

# Create the /mnt/spice directory
RUN mkdir -p /mnt/spice

WORKDIR /app/data
ENV IMAP_DATA_DIR /app/data

# Define the entrypoint of the container
ENTRYPOINT ["imap_cli"]
```

imap_processing/cdf/utils.py
```diff
@@ -87,14 +87,16 @@ def write_cdf(
     # Determine the file name based on the attributes in the xarray
     # Set file name based on this convention:
     # imap_<instrument>_<datalevel>_<descriptor>_<startdate>_
-    # <version>.cdf
```
greglucas: This can probably all fit on one line?

imap_processing/cli.py
```diff
@@ -11,11 +11,18 @@
...
 from abc import ABC, abstractmethod
 from pathlib import Path
 
+import imap_data_access
+from cdflib.xarray import cdf_to_xarray
+
 import imap_processing
+from imap_processing.cdf.utils import write_cdf
+from imap_processing.swe.l1a.swe_l1a import swe_l1a
+from imap_processing.swe.l1b.swe_l1b import swe_l1b
```
greglucas, line 34: I think we should look into this naming later because I find the `from name import name` confusing. I may be in the minority here though of wanting to slim down on our nested modules.

imap_processing/cli.py
```diff
@@ -45,11 +52,27 @@ def _parse_args():
...
+        "Example: '[{instrument: swe, level: l0, version: v00-01}]'"
+    )
 
     parser = argparse.ArgumentParser(prog="imap_cli", description=description)
-    parser.add_argument("--data-dir", type=str, required=False, help=data_dir_help)
     parser.add_argument("--instrument", type=str, required=True, help=instrument_help)
     parser.add_argument("--level", type=str, required=True, help=level_help)
+    parser.add_argument(
+        "--file_path",
+        type=str,
```
greglucas, line 83: You can provide the Path type directly if that is easier.
```suggestion
        type=Path,
```

imap_processing/cli.py
```diff
@@ -169,7 +193,54 @@ class Swe(ProcessInstrument):
...
     def process(self):
         """Perform SWE specific processing."""
+        # self.file_path example:
+        # imap/swe/l1a/2023/09/imap_swe_l1a_sci_20230927_20230927_v01-00.cdf
         print(f"Processing SWE {self.level}")
+        if self.level == "l1a":
+            # create download path
+            download_path = self.file_path.replace("l1a", "l0").replace("cdf", "pkts")
+            print(f"download_path: {download_path}")
+            output_path = imap_data_access.download(download_path)
```
greglucas, line 223: Should this be from the dependency string passed in?

Right now you are back-calculating it from the anticipated product to produce right?

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -44,12 +43,11 @@ def swe_l1b(l1a_dataset: xr.Dataset):
     )
     if apid == SWEAPID.SWE_SCIENCE:
         data = swe_l1b_science(eu_data)
+        if data is None:
+            print("No data to write to CDF")
+            return None
```
```suggestion
            return
```

imap_processing/cli.py
```diff
@@ -45,11 +52,27 @@ def _parse_args():
...
+        "Dependency information in str format."
+        "Example: '[{instrument: swe, level: l0, version: v00-01}]'"
+    )
 
     parser = argparse.ArgumentParser(prog="imap_cli", description=description)
-    parser.add_argument("--data-dir", type=str, required=False, help=data_dir_help)
     parser.add_argument("--instrument", type=str, required=True, help=instrument_help)
     parser.add_argument("--level", type=str, required=True, help=level_help)
+    parser.add_argument(
+        "--file_path",
```
greglucas, line 82: If we can also change the event kickoff, we should try and stick to hyphens only in the CLI arguments.
```suggestion
        "--file-path",
```

imap_processing/cli.py
tech3371: Added TODO.

imap_processing/cli.py
tech3371: TODO

tech3371:
I had worked through all the comments with Greg in-person.

greglucas (approved):

PR #333, opened 2024-02-03
FIX: Make sphinx-contrib an optional dependency only needed for doc builds

greglucas:
# Change Summary

## Overview

Noticed sphinx things getting installed when I tried to install the new 0.2.0 version from pypi. For the code releases we don't need sphinx, so make it an optional dependency and install it with the doc extras.

## New Dependencies

Changed to optional


bourque (approved):
Great catch, and thanks for fixing that!

PR #336, opened 2024-02-13
FIX: Make sure configurations are synced between packages

### LINKED ISSUE #15: BUG - Not accessing IMAP_DATA_DIR env var
### ## Description of the issue

When importing as a library, the code does not correctly read the data directory from the IMAP_DATA_DIR environment variable

1. 2. 3.

1. Clone imap_processing dev commit a29978f3a31d5f1ff10f256245e5eb4ab20a8b37
2. Run the CLI with: `python -m cli --instrument swe --level l1a --file_path imap/swe/l1a/2024/02/imap_swe_l1a_sci_20240203_20240203_v01-01.cdf --dependency "[{'instrument':'swe', 'level':'l0', 'version':'v01-01'}]"` 
3. See that it fails
4. update IMAP_DATA_DIR value to the expected directory (mine was /Users/maha7656/Documents/code/imap/imap_processing/imap_processing/imap-data)
5. see that  it still fails

### Code Snippet:

```python
Code
```

Should read the data directory value from IMAP_DATA_DIR when it's set

Defaults to $PWD/data as data directory

### Additional notes

Reproducing this might need tweaking depending on data in the bucket

_No response_

### Suggested fix?

Read config value differently

greglucas:
# Change Summary

## Overview
This moves all the data configuration handling to the `imap_data_access` package.

An alternative would be to share the configuration dictionary between the packages by doing something like:
`from imap_data_access import config` in `imap_processing.__init__.py` but that seems like too much sharing at that point and we should really keep the configuration dictionaries separate and modify the proper ones where appropriate. This means that I removed the configuration dictionary from `imap_processing` now since it only had data storage methods in it.

I'm not sure if we can cross-repo close issues, but worth a shot.
closes https://github.com/IMAP-Science-Operations-Center/imap-data-access/issues/15

## Deleted Files

- test_data_access
    - This is handled in the other repository and that is there already. Not sure there is a great way to test this here?

## Updated Files
- Updates to the global test fixture to set the imap_data_access utilities to a temporary directory.
- Updates to use `imap_data_access.config` instead of `imap_processing.config`


bourque (approved):
LGTM

Perhaps for a different PR, but we should also remove the `INSTRUMETNS` and `PROCESSING_LEVELS` in the `__init__.py` file and import those from `imap_data_access` as well.

tech3371 (approved):

maxinelasp (approved):
Nice, this looks good. Hopefully this fixes the weird conflicts!

laspsandoval (approved):

PR #339, opened 2024-02-14
Lo SciDE data class fix - multiple DEs in a packet 

### LINKED ISSUE #338: BUG - Lo Science Direct Events data class can't handle multiple direct events in a single packet
### ## Description of the issue

 Lo Science Direct Events data class can't handle multiple direct events in a single packet

1. 2. 3.

_No response_

### Code Snippet:

```python
Code
```

_No response_

_No response_

### Additional notes

_No response_

_No response_

### Suggested fix?

_No response_

sdhoyt:
# Change Summary

## Overview
Updated the Science Direct Events data class for Lo to handled multiple direct events in a single packet.

closes #338 

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
- science_direct_events.py

## Testing
Updated test_science_direct_events unit tests and added a new test for multiple direct events in a binary chunk.

### DIFF
--- a/imap_processing/lo/l0/science_direct_events.py
+++ b/imap_processing/lo/l0/science_direct_events.py
@@ -6,7 +6,6 @@ import bitstring
 from bitarray import bitarray
 
 import imap_processing.lo.l0.decompression_tables as decompress_tables
-from imap_processing.ccsds.ccsds_data import CcsdsData
 from imap_processing.lo.l0.lo_base import LoBase
 
 
@@ -76,7 +75,7 @@ class ScienceDirectEvents(LoBase):
 
     SHCOARSE: int
     COUNT: int
-    DATA: bitstring.Bits
+    DATA: str
     CHKSUM: int
     TOF0: float
     TOF1: float
@@ -88,23 +87,25 @@ class ScienceDirectEvents(LoBase):
 
     def __init__(self, packet, software_version: str, packet_file_name: str):
         """Intialization method for Science Direct Events Data class."""
-        super().__init__(software_version, packet_file_name, CcsdsData(packet.header))
-        self.parse_data(packet)
-        self._decompress_data()
+        # super().__init__(software_version, packet_file_name, CcsdsData(packet.header))
+        # self.parse_data(packet)
+        # self._decompress_data()
 
     def _decompress_data(self):
         """Decompress the Lo Science Direct Events data."""
-        case_number = self._find_decompression_case()
-        tof_decoder = self._find_tof_decoder_for_case(case_number)
-        tof_calculation_binary = self._read_tof_calculation_table(case_number)
-        remaining_bit_coefficients = self._find_remaining_bit_coefficients(
-            tof_calculation_binary
-        )
-        parsed_bits = self._parse_binary(case_number, tof_decoder)
-        # decode_fields will set the TOF class variables
-        self._decode_fields(remaining_bit_coefficients, parsed_bits)
+        bitstream = bitstring.ConstBitStream(bin=self.DATA)
+        for _ in self.COUNT:
+            case_number = self._find_decompression_case(bitstream)
+            tof_decoder = self._find_tof_decoder_for_case(case_number, bitstream)
+            tof_calculation_binary = self._read_tof_calculation_table(case_number)
+            remaining_bit_coefficients = self._find_remaining_bit_coefficients(
+                tof_calculation_binary
+            )
+            parsed_bits = self._parse_binary(case_number, tof_decoder, bitstream)
+            # decode_fields will set the TOF class variables
+            self._decode_fields(remaining_bit_coefficients, parsed_bits)
 
-    def _find_decompression_case(self):
+    def _find_decompression_case(self, bitstream):
         """Find the decompression case for this DE.
 
         The first 4 bits of the binary data are used to
@@ -112,9 +113,11 @@ class ScienceDirectEvents(LoBase):
         The case number is used to determine how to
         decompress the TOF values.
         """
-        return int(self.DATA[0:4].bin, 2)
+        # TODO: Do I need to set the new bitpos or does the read
+        # do that automatically?
+        return bitstream.read(4).uint
 
-    def _find_tof_decoder_for_case(self, case_number):
+    def _find_tof_decoder_for_case(self, case_number, bitstream):
         """Get the TOF decoder for this DE's case number.
 
         The case number determines wich TOF decoder to use.
@@ -130,13 +133,13 @@ class ScienceDirectEvents(LoBase):
         but TOF3 was transmitted.
         """
         if case_number == 0:
-            gold_or_silver = int(self.DATA.bin[4])
+            gold_or_silver = int(bitstream.read(1).bin)
             tof_decoder = decompress_tables.tof_decoder_table[case_number][
                 gold_or_silver
             ]
 
         elif case_number in [4, 6, 10, 12, 13]:
-            is_bronze = int(self.DATA.bin[4])
+            is_bronze = int(bitstream.read(1).bin)
             tof_decoder = decompress_tables.tof_decoder_table[case_number]
 
             if not is_bronze:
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l0/science_direct_events.py
```diff
@@ -76,45 +65,54 @@ class ScienceDirectEvents(LoBase):
...
+    TOF3: np.ndarray
+    TIME: np.ndarray
+    ENERGY: np.ndarray
+    POS: np.ndarray
 
     def __init__(self, packet, software_version: str, packet_file_name: str):
         """Intialization method for Science Direct Events Data class."""
         super().__init__(software_version, packet_file_name, CcsdsData(packet.header))
         self.parse_data(packet)
+        self.TOF0 = np.array([])
```
greglucas, line 82: This extends the individual arrays, there is an alternative of making a list of `ScienceDirectEvent` objects to keep the base structure of what you have and just putting multiples of those together instead if that is useful at all. `packet_events = [DirectEvent1, DirectEvent2, ...]`

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -76,45 +65,54 @@ class ScienceDirectEvents(LoBase):
...
-        case_number = self._find_decompression_case()
-        tof_decoder = self._find_tof_decoder_for_case(case_number)
-        tof_calculation_binary = self._read_tof_calculation_table(case_number)
-        remaining_bit_coefficients = self._find_remaining_bit_coefficients(
-            tof_calculation_binary
-        )
-        parsed_bits = self._parse_binary(case_number, tof_decoder)
-        # decode_fields will set the TOF class variables
-        self._decode_fields(remaining_bit_coefficients, parsed_bits)
+        bitstream = bitstring.ConstBitStream(bin=self.DATA)
```
greglucas, line 93: I liked avoiding the bitstream manipulation you had before if possible. I think this is so you can read x bits off the field and use bitstream to do the indexing for you? I wonder if you could pass in the start/range to your function instead and keep what you had. See below for an example of what I mean.

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -76,45 +65,54 @@ class ScienceDirectEvents(LoBase):
...
+    def _find_decompression_case(self, bitstream):
         """Find the decompression case for this DE.
 
         The first 4 bits of the binary data are used to
         determine which case number we are working with.
         The case number is used to determine how to
         decompress the TOF values.
         """
-        return int(self.DATA[0:4].bin, 2)
+        return bitstream.read(4).uint
```
greglucas, line 113: If you pass the current location in, you could keep this as it was before I think.
```suggestion
        return int(self.DATA[loc:loc+4], 2)
```

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -130,13 +128,13 @@ def _find_tof_decoder_for_case(self, case_number):
         but TOF3 was transmitted.
         """
         if case_number == 0:
-            gold_or_silver = int(self.DATA.bin[4])
+            gold_or_silver = int(bitstream.read(1).bin)
```
greglucas, line 131: Do you want these to be booleans?

imap_processing/lo/l0/science_direct_events.py
```diff
@@ -76,45 +65,54 @@ class ScienceDirectEvents(LoBase):
...
+    DATA: str
     CHKSUM: int
-    TOF0: float
-    TOF1: float
-    TOF2: float
-    TOF3: float
-    TIME: float
-    ENERGY: float
-    POS: float
+    TOF0: np.ndarray
```
maxinelasp, line 70: Related to what Greg said - is this because each direct event has multiple TOF0 values, or is it that `ScienceDirectEvents` contains multiple direct events? How are they organized in the packets? Generally I think you could do this a couple different ways, but it might be good to think about how this class will be used in L1A - like, will you need to divide one L0 class into multiple L1A classes with this design? 

It's hard for me to understand what the underlying data is structured like, but I generally think the classes you use to describe it should align with the data structure as closely as possible to make it easier to use the data in the way you're expecting to. So, if this ndarray is aligned with the packets and the pieces of data, then it looks good to me. But, it may end up being easier to make an additional class here or something which for each piece of data has [TOF0, TOF1, etc] (for example, if TOF0[0] is ALWAYS matched with TOF1[0] and TOF2[0]) and then make a list of that class instead.

maxinelasp (approved):

imap_processing/lo/l0/science_direct_events.py
sdhoyt: The L1A data will have multiple classes merged together. For example, say there are 2 packets for a certain day:

packet 1:
2 DEs
TOF0: [1, 2]

packet 2:
3 DEs
TOF0=[3, 4, 5]

L1A CDF
TOF0 = [1, 2, 3, 4, 5]

imap_processing/lo/l0/science_direct_events.py
sdhoyt: I'm not sure I'm following. The idea is to have a single packet worth of DEs stored in this object and then merged them all together when the CDF is created

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Yeah, I can remove bitstring. It will just make things a little more complicated because different types of DEs require you to check certain bits where other types don't, plus looping over multiple DEs in the packet will require more bookkeeping

imap_processing/lo/l0/science_direct_events.py
sdhoyt: Yeah, I should make them booleans

imap_processing/lo/l0/science_direct_events.py
greglucas: I think this is related to my comment above: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/339#discussion_r1489917218 and I agree with Maxine's questions about trying to determine how you want to organize things. Maybe another way of rephrasing this is: Do you want a `PacketDataClass`, or a `DirectEvent` dataclass? If you keep it as the `DirectEvent` dataclass like you had before then you could have something like passing in a full packet and returning a list of one or more direct events right away:
```python
def get_direct_events(packet):
    # Parse your packet to create N direct event objects
    return [direct_event1, direct_event2, ...]
```

Whereas what you have with the `PacketDataClass` style now with the arrays you'll still need to parse out the individual direct events later, so it kind of creates an extra structure that may not be necessary unless you want to store the packet level information for later use somewhere. (You could also add a field to each DirectEvent object indicating what packet it came from too)

maxinelasp: I think with what you described, your current design is closely aligned (although I would recommend renaming it to one of Greg's suggestions). Overall I agree with Greg's comment about deciding what level of structure would be useful. If you combine them in L1A, would it make sense to split based on the direct events, rather than using arrays? Alternatively, would it make more sense to combine the TOF right away, putting multiple packets into one instance of the class? I do something like that for the GLOWS L1A direct events, where I add multiple packets together into one dataclass as they arrive, so feel free to look at that as an example. 

Thinking about the next step in processing can often help me clear up questions like this. If your next step is going to involve some weird array manipulation across multiple instances, it might make more sense to combine them now. If it's actually super easy and straightforward and you just have all of them anyway, then what you have will totally work as well. Overall, I think if you sort out the structure at the lower level, it can inform later, more complicated levels. If it's complex at L1A, it's probably also complex at L1B, and the complexity carries over as you get further in processing. If it's simple at L1A, it's less likely to cause problems at L2. 🤷  

You know the data and your code, so ultimately if you think what you have makes sense and will be simple to use, that's what's most important. Feel free to also ping me if you want to have a conversation about it - I (clearly) absolutely love digging into the best structure for data management and processing :)

PR #340, opened 2024-02-14
Mag cdf

maxinelasp:
# Change Summary
Updates to MAG to add raw output and initial pass at CDF files.

## Overview

Added CDF files plus a few methods to create them

## New Files

- mag_l1a
   - Call methods for decom
- mag_cdf_attrs
   -  shared instrument attributes for CDF files
   
## Updated Files

- global cdf attrs
   - Created system to add an instrument PI


imap_processing/cdf/global_attrs.py
```diff
@@ -30,24 +33,63 @@ class GlobalConstantAttrs:
...
-        "Instrument_type": "Particles (space)",
+        # TODO: CDF docs say this value should be "IMAP"
         "Mission_group": "IMAP>Interstellar Mapping and Acceleration Probe",
     }
 
+    pi_name: tuple[str] = ("Dr. David J. McComas",)
+    pi_affiliation: tuple[str, str] = (
+        "Princeton Plasma Physics Laboratory",
+        "100 Stellarator Road, Princeton, NJ 08540",
+    )
```
greglucas: It looks like PI_NAME should only be one entry according to the CDF docs. Should this be replaced with the instrument PI instead of appended to with both the mission PI and instrument PI?

@bryan-harter would be good for you to review these CDF changes to see if there is a standard way of putting this into the files that SPDF approves of.

imap_processing/cdf/global_attrs.py
```diff
@@ -88,12 +130,21 @@ class GlobalInstrumentAttrs:
         Descriptor of the instrument (Ex: "IDEX>Interstellar Dust Experiment")
     text: str
         Explanation of the instrument, usually as a paragraph.
+    instrument_type: str default="Particles (space)"
```
greglucas: There should be a space before the colon in numpydoc for all the parameters in the docstrings https://numpydoc.readthedocs.io/en/latest/format.html#parameters
```suggestion
    instrument_type : str, default="Particles (space)"
```

imap_processing/mag/l0/decom_mag.py
```diff
@@ -1,15 +1,21 @@
...
 from space_packet_parser import parser, xtcedef
 
 from imap_processing import imap_module_directory
 from imap_processing.ccsds.ccsds_data import CcsdsData
+from imap_processing.cdf.global_attrs import ConstantCoordinates
+from imap_processing.mag import mag_cdf_attrs
 from imap_processing.mag.l0.mag_l0_data import MagL0, Mode
 
+logging.basicConfig(level=logging.ERROR)
 logger = logging.getLogger(__name__)
```
greglucas: Why don't we want warning messages? I'd just leave this alone, but we should only update this logger and not the global level.
```suggestion
logger = logging.getLogger(__name__)
logger.setLevel(logging.ERROR)
```

imap_processing/mag/l0/decom_mag.py
```diff
@@ -40,11 +46,14 @@ def decom_packets(packet_file_path: str) -> list[MagL0]:
 
     with open(packet_file_path, "rb") as binary_data:
         try:
-            mag_packets = mag_parser.generator(binary_data)
+            mag_packets = mag_parser.generator(
+                binary_data,
+                buffer_read_size_bytes=5790778,  # Todo: what size?
+            )
```
greglucas: Speaking with Gavin earlier in that thread on slack we shouldn't be updating this, and in fact it makes it much slower due to the buffer thrashing issues on space_packet_parser. We can safely use the default.
```suggestion
            mag_packets = mag_parser.generator(binary_data)
```

imap_processing/mag/l0/decom_mag.py
```diff
@@ -53,9 +62,89 @@ def decom_packets(packet_file_path: str) -> list[MagL0]:
                     ]
                     data_list.append(MagL0(CcsdsData(packet.header), *values))
         except ReadError as e:
-            logger.error(
+            logger.warning(
                 f"Found error: {e}\n This may mean reaching the end of an "
                 f"incomplete packet."
             )
```
greglucas: Can you remove this try/except altogether? I believe I provided some code that should fix this. IMO we have a bad packet definition if we are running into this. I'm happy to open a separate PR with that code, just let me know if you want me to do that separately.

imap_processing/mag/l0/decom_mag.py
```diff
@@ -53,9 +62,89 @@ def decom_packets(packet_file_path: str) -> list[MagL0]:
...
+            for key, value in dataclasses.asdict(datapoint).items():
+                if key != "ccsds_header":
+                    norm_data[key].append(value)
+        if datapoint.ccsds_header.PKT_APID == Mode.BURST:
+            burst_data["SHCOARSE"].append(datapoint.SHCOARSE)
+            burst_data["raw_vectors"].append(datapoint.VECTORS)
+
+    # Used in L1A vectors
+    direction = xr.DataArray(
+        np.arange(194),
```
greglucas: Is this a placeholder for now to be filled in with real data later? Or is this a coordinate that is "fixed" in a sense? I was expecting to see something related to `norm_data["raw_vectors"]`, so it might help to put a quick comment in one way or the other.

imap_processing/mag/l0/decom_mag.py
```diff
@@ -53,9 +62,89 @@ def decom_packets(packet_file_path: str) -> list[MagL0]:
...
+        name="Direction",
+        dims=["Direction"],
+        attrs=mag_cdf_attrs.direction_attrs.output(),
+    )
+
+    # print(direction)
```
greglucas:
```suggestion
```

imap_processing/mag/l0/decom_mag.py
```diff
@@ -53,9 +62,89 @@ def decom_packets(packet_file_path: str) -> list[MagL0]:
...
+    # TODO add norm to attrs somehow
+    norm_dataset = xr.Dataset(
+        coords={"Epoch": norm_epoch_time, "Direction": direction},
+        attrs=mag_cdf_attrs.mag_l1a_attrs.output(),
+    )
+
+    norm_dataset["RAW_VECTORS"] = norm_raw_vectors
+
+    # TODO: retrieve the doc for the CDF description
+    print(getattr(MagL0, "__doc__", {}))
```
greglucas: Do you want this print there? If so, I'd add some extra context, otherwise if it is debugging I'd remove it and leave the TODO.

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -0,0 +1,30 @@
...
+    packet_filepath:
+        Packet files for processing
+    cdf_directory:
+        Directory for output
+    """
+    mag_l0 = decom_mag.decom_packets(packet_filepath)
+
+    mag_datasets = decom_mag.export_to_xarray(mag_l0)
+
+    write_cdf(mag_datasets, mode="norm", directory=cdf_directory)
```
greglucas: There is no "mode" anymore, so you might need a rebase to get the updates. I think this would be "descriptor" now.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/497d446ebfe7a6eea45dc6d85e397338708741e4/imap_processing/cdf/utils.py#L41-L44

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -0,0 +1,105 @@
...
+from imap_processing.mag import __version__
+
+text = (
+    "The IMAP magnetometer (MAG) consists of a pair of identical magnetometers "
+    "which each measure the magnetic field in three directions in the vicinity of "
+    "the spacecraft. "
+    "MAG will contribute to our understanding of the acceleration and transport "
+    "of charged particles in the heliosphere. "
+    "MAG design and assembly is led by Imperial College, London. See "
+    "https://imap.princeton.edu/instruments/mag for more details."
```
greglucas, line 18: That page is frustrating to read with the jumping header :) Good job having the patience to grab that text!

imap_processing/mag/tests/mag_multiple_packets.pkts
greglucas, line 1: Curious in GitHub's UI these show as new files. Are they newly added here, or just updated?
These files are in the base repo already it looks like: https://github.com/IMAP-Science-Operations-Center/imap_processing/tree/dev/imap_processing/tests/mag

imap_processing/cdf/global_attrs.py
maxinelasp: Yeah, I wasn't sure if it should include the instrument PI or not. The example files I had from MAG did.

imap_processing/mag/l0/decom_mag.py
maxinelasp: Oh yeah, this is left over from something else. Let me take it out.

imap_processing/mag/l0/decom_mag.py
maxinelasp: Yup, just missed updating it here, I will remove!

imap_processing/mag/l0/decom_mag.py
maxinelasp: Ah yeah that was left in for future work, but shouldn't be merged. I'll remove it and add it in the TODO or my notes

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Yeah, this was a merge fail! Or de-merge? I deleted the commit that did this. I'll update it to match the current version of write_cdf in the code

imap_processing/mag/mag_cdf_attrs.py
maxinelasp: Lol I hate the website design of that page, it's so jerky

imap_processing/mag/tests/mag_multiple_packets.pkts
maxinelasp: Ah, I know what the problem is. These are under /mag/tests, when they should be (and are in dev) under tests/mag. I'll make sure the test file under tests/mag has everything and then remove these.

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Also, this method signature is changed again in my other CLI PR, just so you're aware.

imap_processing/mag/tests/mag_multiple_packets.pkts
maxinelasp: Also I apparently failed to write tests for CDF stuff. I'll write a few mag-specific ones as well before merging.

greglucas: Interesting! I think there is something confusing GitHub somewhere because now it shows them being deleted even though they aren't on the main branch :)

I _think_ once you squash down to a single commit on merge this will resolve itself...

imap_processing/mag/l0/decom_mag.py
greglucas: Awesome, thanks!

imap_processing/mag/l1a/mag_l1a.py
greglucas: :+1:

imap_processing/mag/tests/mag_multiple_packets.pkts
greglucas: Ignore me! I was only reviewing the single commit. Would be nice to squash down still though to avoid added/removed files. Thanks for figuring that out!

imap_processing/cdf/global_attrs.py
```diff
@@ -30,24 +33,63 @@ class GlobalConstantAttrs:
...
+            A tuple of affilitations for each PI. This should include univeristy
+            information. Can be any length
+
+        Returns
+        -------
+        dict
+            Global base attributes, plus pi_names and pi_attributes from the default
+            values and any provided keywords
+        """
+        output_name = list(self.pi_name + pi_names if pi_names else self.pi_name)
```
sdhoyt: Is the purpose of appending to PI names/affiliations defined in this file to add the instrument PI?

imap_processing/cdf/global_attrs.py
```diff
@@ -8,12 +8,15 @@
 This check occurs during the xarray_to_cdf step, which has an example in
```
sdhoyt, line 8: @bryan-harter This isn't something we'd want to change in this PR, but are there any attributes we should add in here to conform to SPASE?

imap_processing/mag/l0/decom_mag.py
```diff
@@ -39,23 +42,95 @@ def decom_packets(packet_file_path: str) -> list[MagL0]:
     data_list = []
 
     with open(packet_file_path, "rb") as binary_data:
```
sdhoyt, line 46: Could you use the decom.decom_packets function here?

imap_processing/mag/tests/mag_multiple_packets.pkts
maxinelasp: Yeah definitely will squash, don't worry 😆 this PR is slightly a mess with all the commits

imap_processing/cdf/global_attrs.py
maxinelasp: Yes

imap_processing/mag/l0/decom_mag.py
maxinelasp: I'd have to loop through twice. Once to get the list of decommed packets from decom_packets, then again to go through that list and apply the MagL0 code to it. One alternative might be to write an optional parameter in decom_packets that includes a dataclass to apply ... Since Python's typing system is so loose, you could then return either a list of packets or a list of, for example, MagL0 objects. Might be worth doing now that more people are using the dataclasses...?

imap_processing/mag/l0/decom_mag.py
maxinelasp: That is a good question that I don't fully remember the answer too. I think this is a variable coordinate that corresponds to the length of the vector (for each time stamp, there should be a vector of length 4 - x, y, z, and sensor range) but for these L0 RAW CDF files, there is instead a long string of bytes. So I think this was mostly me testing that I could get that 2d array in the CDF, but this format isn't correct for this output.

imap_processing/mag/l0/decom_mag.py
greglucas: I think you might need to add the new packets you had from one of the other commits, or update the XTCE definition to avoid the `ReadError`s.
https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/7920476930/job/21623597869#step:6:145

imap_processing/cdf/global_attrs.py
greglucas: Thanks for sending the examples to Bryan and following up with him. It seems like it is acceptable to do it this way and there is not necessarily a standard. We can update this at a later time if need be as well.

imap_processing/mag/l0/decom_mag.py
sdhoyt: You wouldn't know what APIDs are in the file to pass in a data class though, so you'd probably still have to loop through twice. For example, if the file contains science and housekeeping packets, you would need to pass in multiple data classes and have a way to match up the APIDs. I might be missing something here though. But I do see what you're saying about the way you're doing it now avoids the need to loop through the packets twice.

maxinelasp: Ah yeah... that's a good point. You could probably do it with some nasty dictionary shenanigans... but I hate nasty dictionary shenanigans! Personally, I think it's fine to just have every instrument do the opening and reading themselves, since there are special cases (dataclasses) and weird scenarios (what you said)

imap_processing/mag/l0/decom_mag.py
maxinelasp: I addressed this issue by updating the test files AND fixed the XTCE definition!

maxinelasp:
Could I get another review @greglucas  or @sdhoyt with these new updates? Thanks for your prior comments!

imap_processing/cdf/global_attrs.py
bryan-harter: Eric Christian pointed out to me that there's apparently no such thing as an "Instrument PI" on IMAP, they are "Instrument Leads".  So PI name would probably just be reserved for Dave.  I'll get confirmation later if we need another global attribute for Instrument Leads

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -1,72 +1,54 @@
...
-    burst_test_file = current_directory / "mag_single_burst.pkts"
-    l0 = decom_packets(str(burst_test_file))[0]
 
-    assert l0.ccsds_header.PKT_APID == 1068
-    assert l0.ccsds_header.SRC_SEQ_CTR == 1
-    assert l0.ccsds_header.PKT_LEN == 1821
+def test_mag_decom():
+    current_directory = Path(__file__).parent
+    burst_test_file = current_directory / "mag_l0_test_data.pkts"
+    l0 = decom_packets(str(burst_test_file))
```
greglucas: I think you could leave this as a path being passed in so you don't have to cast to string in all of these.

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -1,72 +1,54 @@
...
-        PRI_COARSETM=435954625,
-        PRI_FNTM=64610,
-        SEC_COARSETM=435954625,
-        SEC_FNTM=64607,
-    )
-    assert l0 == expected
+    expected_output = pd.read_csv(current_directory / "mag_l0_test_output.csv")
 
+    index = 0
+    for test in l0:
```
greglucas:
```suggestion
    for index, test in enumerate(l0):
```
and remove the index += at the end of the loop

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -1,72 +1,54 @@
...
-    assert len(l0) == 125
-    for val in l0:
-        assert val.ccsds_header.PKT_APID == 1052
-        assert len(val.VECTORS) > 0
+    output_data = export_to_xarray(l0)
+    required_attrs = list(
+        global_attrs.GlobalInstrumentAttrs("", "", "").output().keys()
+    )
+    print(required_attrs)
+    print(output_data.attrs)
```
greglucas:
```suggestion
```

greglucas (approved):
Looks good to me! Some minor nits, but looks good overall. Thanks for updating the tests and removing the try/except, that is satisfying to see :)

imap_processing/cdf/global_attrs.py
bryan-harter: Eventually we'll probably want to add an attribute for the spase_DatasetResourceID, but it's not actually necessary.  We can discuss the SPASE requirement more at the SDWG tomorrow, but the internal format of our CDF files doesn't need to be anything special to comply to SPASE

imap_processing/cdf/global_attrs.py
maxinelasp: Sounds good. I'll remove this change from the PR and leave it as just Dave.

PR #341, opened 2024-02-14
CLI updates

maxinelasp:
# Change Summary
CLI updates to rework how the CLI is used

## Overview

Updated CLI to run MAG files and changed the API

# IMPORTANT NOTE: 
This is built off of the other MAG CDF changes (in another branch) and is dependent on those. Please ignore duplicate files (anything MAG specific, under the mag/ directory) from there, they will go away once that change is merged. I can't open a PR into the base repository but compare to my branch, but I want to open this to comments. 

Only files to review here: `imap_processing/cli.py`, `imap_processing/cdf/utils.py`

This is also dependent on some imap-data-access changes, up for review [here](https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/18). If you want to test this locally, set up the changes as a [git dependency](https://python-poetry.org/docs/dependency-specification/#git-dependencies) in poetry, or checkout the imap-data-access repo and set up a [file dependency](https://python-poetry.org/docs/dependency-specification/#path-dependencies).  

BOTH dependent PRs will need to be merged (and in imap-data-access, released) before this PR can be.


imap_processing/cdf/global_attrs.py
```diff
@@ -8,12 +8,15 @@
 This check occurs during the xarray_to_cdf step, which has an example in
```
maxinelasp: Don't review this file

imap_processing/cdf/utils.py
```diff
@@ -2,9 +2,7 @@
 
 import logging
```
maxinelasp, line 3: YES review this file

imap_processing/cli.py
```diff
@@ -11,10 +11,11 @@
 """
 
 import argparse
```
maxinelasp, line 13: YES review this file

imap_processing/mag/l0/decom_mag.py
```diff
@@ -1,19 +1,24 @@
-"""Decom MAG data packets using MAG packet definition."""
```
maxinelasp: Don't review this file here

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -1,4 +1,4 @@
-"""Data class for MAG Level 0 data."""
```
maxinelasp: Don't review this file here

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -0,0 +1,23 @@
+"""Methods for decomming packets, processing to level 1A, and writing CDFs for MAG."""
```
maxinelasp, line 1: Don't review this file here (though you can reference it)

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -0,0 +1,105 @@
+"""Shared attribute values for MAG CDF files."""
```
maxinelasp: Do not review this file

imap_processing/mag/tests/test_mag_decom.py
```diff
@@ -0,0 +1,72 @@
+import dataclasses
```
maxinelasp: Don't review this file here

imap_processing/tests/cdf/test_attr_classes.py
```diff
@@ -12,11 +12,30 @@
 
 
 def test_global_attrs():
```
maxinelasp: Don't review this file here

imap_processing/cdf/utils.py
```diff
@@ -56,74 +50,31 @@ def write_cdf(
     ----------
         data : xarray.Dataset
             The dataset object to convert to a CDF
+        filepath: Path
+            The output path, including filename, to write the CDF to.
         descriptor : str
             The descriptor to insert into the file name after the
             orbit, before the SPICE field.  No underscores allowed.
```
bourque: I think this could also be removed since it is not used?

imap_processing/cli.py
bourque, line 1: Could you also update the usage example in the module docstrings to reflect these changes?

imap_processing/cli.py
```diff
@@ -130,11 +168,70 @@ class ProcessInstrument(ABC):
...
+    def download_dependencies(self):
+        """Download the dependencies for the instrument.
+
+        Returns
+        -------
+        file_list: list[str]
+            A list of file paths to the downloaded dependencies.
+        """
+        file_list = []
+        for dep in self.dependencies:
```
```suggestion
        for dependency in self.dependencies:
```

imap_processing/cli.py
```diff
@@ -130,11 +168,70 @@ class ProcessInstrument(ABC):
...
+            try:
+                # TODO: Validate dep dict
+                # TODO: determine what dependency information is optional
+                val = imap_data_access.query(
```
```suggestion
                value = imap_data_access.query(
```

imap_processing/cli.py
```diff
@@ -196,6 +293,26 @@ class Mag(ProcessInstrument):
     def process(self):
         """Perform MAG specific processing."""
         print(f"Processing MAG {self.level}")
+        file_paths = self.download_dependencies()
```
bourque, line 307: Larger philosophical question outside the scope of this PR: since we are really now starting to implement these, should we move these classes to a separate module? Either one module per instrument, or one module that stores all of the instruments?

imap_processing/cli.py
```diff
@@ -78,19 +92,49 @@ def _parse_args():
...
     parser.add_argument(
         "--dependency",
         type=str,
         required=True,
-        help=depdency_help,
+        help=dependency_help,
+    )
+
+    parser.add_argument(
+        "--use-remote",
```
bourque: It might be better if we name this with the word "upload" somewhere in it, since that is what it is used for? Maybe `--upload-to-cloud` or `--upload-to-aws` or `--upload-to-sdc`?

imap_processing/cli.py
bourque: I fear that if we leave this the way it is, this module is going to get very large!

imap_processing/cdf/utils.py
maxinelasp: Merge fail 😆

imap_processing/cli.py
maxinelasp: `--api-upload`?

imap_processing/cli.py
maxinelasp: Yeah good question! Yeah, I think it would make sense to move the ProcessInstrument class and all the subclasses into a different file and keep the CLI stuff here

imap_processing/cdf/utils.py
```diff
@@ -56,74 +50,31 @@ def write_cdf(
     ----------
         data : xarray.Dataset
             The dataset object to convert to a CDF
+        filepath: Path
```
tech3371, line 53: Do we assume here that user had already constructed path using `imap-data-access` library which validates filename?

imap_processing/cli.py
tech3371: I like this one `--upload-to-sdc`. It kind of tells what this flag do.

imap_processing/cli.py
```diff
@@ -213,52 +332,52 @@ def process(self):
         """Perform SWE specific processing."""
         # self.file_path example:
         # imap/swe/l1a/2023/09/imap_swe_l1a_sci_20230927_20230927_v01-00.cdf
+        dependencies = self.download_dependencies()
         print(f"Processing SWE {self.level}")
+
+        # TODO: currently assumes just the first path returned is the one to use
```
tech3371, line 347: Yes. I can update this afterwards.

imap_processing/mag/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "01"
```
tech3371: I don't know if you use this version for filename. If you do, this is not in the format it expects and will throw error.

imap_processing/cdf/utils.py
maxinelasp: Yes, perhaps I should add that into the comment. This assumes that the path is correct, or at least that you are sure it is the path you want to write the file to. I think this method should not be responsible for validating that - for example, if you want to write test files locally that don't match our naming convention, this method should allow for that.

imap_processing/cli.py
maxinelasp: Cool, I figured I could leave more of the SWE stuff for you to finalize, but I just didn't want to break it in my PR 😆

imap_processing/mag/__init__.py
maxinelasp: No, this is the version for the instrument code, which I believe only gets written into the CDF files. The CLI gets a version passed in, and does not generate versions at all.

imap_processing/cdf/utils.py
tech3371: ok

imap_processing/cli.py
```diff
@@ -61,9 +75,9 @@ def _parse_args():
         "The data level to process. Acceptable values are: "
         f"{imap_processing.PROCESSING_LEVELS}"
     )
-    depdency_help = (
+    dependency_help = (
         "Dependency information in str format."
-        "Example: '[{instrument: swe, level: l0, version: v00-01}]'"
+        "Example: '[{'instrument': 'swe', 'level': 'l0', 'version': 'v00-01'}]'"
```
tech3371: can we update example to match new values passed in?
```
[
        {
            'instrument': 'mag',
            'data_level': 'l0',
            'descriptor': 'sci',
            'version': 'v00-01',
            'start_date': '20231212',
            'end_date': '20231212'
        }]
   ```

File: imap_processing/cli.py [hunk 19]
```diff
@@ -130,11 +168,70 @@ class ProcessInstrument(ABC):
     ----------
     level : str
         The data level to process (e.g. ``l1a``)
+    file_path : str
+        The full path to the output file in the S3 bucket
+    dependencies : list[dict]
+        A list of dictionaries containing the dependencies for the instrument in the
+        format: "[{'instrument': 'swe', 'level': 'l0', 'version': 'v00-01'}]"
```
tech3371: same here. some doc string update

imap_processing/cli.py
tech3371: I think you did great job on that. I saw SWE processed successfully last week.

imap_processing/mag/__init__.py
tech3371: cool.

tech3371 (commented):
Couple minor doc string updates. Otherwise, it looks good to me!

imap_processing/codice/utils.py
```diff
@@ -112,7 +113,7 @@ def create_dataset(packets: list[space_packet_parser.parser.Packet]) -> xr.Datas
         add_metadata_to_array(packet, metadata_arrays)
 
     epoch_time = xr.DataArray(
-        metadata_arrays["SHCOARSE"],
+        [calc_start_time(shcoarse) for shcoarse in metadata_arrays["SHCOARSE"]],
```
tech3371: I think this is a topic of discussion. CDF wants one thing but I think instrument team may still want raw spacecraft time.

pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 bitstring = ">=4.0.1"
-cdflib = "==1.2.3"
+cdflib = ">=1.2.3"
```
tech3371: Did we decide to update this to latest release Bryan made this week which is `1.2.4` or wait? Sorry, I don't remember what we decided.

imap_processing/cli.py
```diff
@@ -61,36 +86,63 @@ def _parse_args():
...
-    #     type=Path,
-    #     required=True,
-    #     help="Full path to the file in the S3 bucket.",
-    # )
+    parser.add_argument("--data-level", type=str, required=True, help=level_help)
+
+    # TODO: Remove this dependency from batch_starter, then remove it from here.
     parser.add_argument(
-        "--file_path",
+        "--file-path",
```
tech3371: I think we are safe to delete this argument now.

imap_processing/cli.py
```diff
@@ -61,36 +86,63 @@ def _parse_args():
...
+        required=False,
+        help="DEPRECATED: Full path to the output file in the S3 bucket.",
+    )
+
+    parser.add_argument(
+        "--start-date",
         type=str,
         required=True,
-        help="Full path to the file in the S3 bucket.",
+        help="Start time for the output data. Format: YYYY-MM-DD",
```
tech3371:
```suggestion
        help="Start time for the output data. Format: YYYYMMDD",
```

imap_processing/cli.py
```diff
@@ -61,36 +86,63 @@ def _parse_args():
...
+    )
+
+    parser.add_argument(
+        "--end-date",
+        type=str,
+        required=False,
+        help="End time for the output data. If not provided, start_time will be used "
+        "for end_time. Format: YYYY-MM-DD",
```
tech3371:
```suggestion
        "for end_time. Format: YYYYMMDD",
```

imap_processing/cli.py
```diff
@@ -61,36 +86,63 @@ def _parse_args():
...
+    )
+    # TODO: Will need to add some way of including pointing numbers
+
+    parser.add_argument(
+        "--version",
+        type=str,
+        required=True,
+        help="Version of the data",
```
tech3371:
```suggestion
        help="Version of the data. Format: vxx-xx",
```

imap_processing/cli.py
```diff
@@ -104,37 +156,97 @@ def _validate_args(args):
...
 
 class ProcessInstrument(ABC):
     """An abstract base class containing a method to process an instrument.
 
     Attributes
     ----------
-    level : str
+    data_level : str
         The data level to process (e.g. ``l1a``)
+    file_path : str
```
tech3371: can we update doc strings to match new inputs?

imap_processing/cli.py
```diff
@@ -147,63 +259,85 @@ class Codice(ProcessInstrument):
...
+
+        if self.data_level == "l1a":
+            # File path is expected output file path
+            if len(file_paths) > 1:
+                raise ValueError(
+                    f"Unexpected dependencies found for MAG L1A:"
+                    f"{file_paths}. Expected only one dependency."
+                )
+            filename = imap_data_access.ScienceFilePath.generate_from_inputs(
+                "mag", "l1a", "raw", self.start_date, self.end_date, "v01-01"
```
tech3371: was there reason for using hardcoded version? should we use `self.version` now?

imap_processing/cli.py
```diff
@@ -281,10 +415,18 @@ def main():
...
     cls = getattr(sys.modules[__name__], args.instrument.capitalize())
-    instrument = cls(args.level, args.file_path)
+    instrument = cls(
+        args.data_level,
+        args.dependency,
+        args.start_date,
+        args.end_date,
+        args.version,
+        args.upload_to_sdc,
+    )
```
tech3371: a dump question, how does say `--start-date` gets converted to `args.start_date`? I looked but I am not seeing it where this conversion happens.

imap_processing/codice/utils.py
maxinelasp: Well for now, the tests fail if there isn't datetime64, so I think regardless we will want to convert from SHCOARSE into datetime64. Is it ok if I leave this for now?

pyproject.toml
maxinelasp: This poetry.lock file does upgrade it, but it should still work with `1.2.3` so I left it like this. I can force a version bump if that makes sense though

imap_processing/cli.py
maxinelasp: Good catch!

imap_processing/cli.py
maxinelasp: Oh yeah, you're totally right. Good catch!

imap_processing/cli.py
maxinelasp: You know, I left it this way with the intent to go back to it, and then it all worked so I never did. Looks like `args` automatically converts the arguments from a hyphen to an underscore.

imap_processing/codice/utils.py
tech3371: yes. We can address that in upcoming weeks.

tech3371 (approved):
Once you test that this works in AWS, it looks good to me! And we can address other comments in upcoming PR.

bourque (approved):
Lots of work went into this PR, nice job!

imap_processing/codice/utils.py
bryan-harter: You can have multiple time variables in your CDF file, so there can be another variable for spacecraft clock time.  

But the one variable named "epoch" must be of the type "CDF_TT2000".  And all DEPEND_0s need to point to the "epoch" variable.

imap_processing/cli.py
```diff
@@ -11,10 +11,11 @@
 """
 
 import argparse
-import os
 import sys
 from abc import ABC, abstractmethod
+from json import loads
```
greglucas, line 16: I find "loads" to be pretty generic and not knowing where it comes from is confusing. I'd slightly prefer this to be: `import json` and then use it as `json.loads()` where needed.

imap_processing/cli.py
```diff
@@ -45,13 +64,19 @@ def _parse_args():
...
+        '--data-level "l1a"'
+        ' --start-date "20231212"'
+        '--end-date "20231212"'
+        '--version "v00-01"'
+        '--dependency "['
+        '   {"instrument": "mag",'
+        '   "data_level": "l0"',
+        '   "descriptor": "sci"',
+        '   "version": "v00-01"',
+        '   "start_date": "0231212"',
```
```suggestion
        '   "start_date": "20231212"',
```

imap_processing/cli.py
```diff
@@ -61,36 +86,55 @@ def _parse_args():
...
         required=True,
-        help=depdency_help,
+        help=dependency_help,
+    )
+
+    parser.add_argument(
+        "--upload-to-sdc",
+        action="store_true",
+        required=False,
+        help="Upload completed output files to IMAP API.",
```
```suggestion
        help="Upload completed output files to the IMAP SDC.",
```

imap_processing/cli.py
```diff
@@ -104,37 +148,101 @@ def _validate_args(args):
...
+        dependency_str: str,
+        start_date: str,
+        end_date: str,
+        version: str,
+        upload_to_sdc: bool,
+    ) -> None:
+        self.data_level = data_level
+
+        # Convert string into a dictionary
+        self.dependencies = loads(dependency_str.replace("'", '"'))
```
greglucas, line 203: Are we dumping malformed json somewhere upstream? If so we should correct it there. Single quotes aren't allowed in the JSON spec, so I'm not sure we should be doing the replacement beforehand.

imap_processing/cli.py
```diff
@@ -104,37 +148,101 @@ def _validate_args(args):
...
+
+        self.start_date = start_date
+        self.end_date = end_date
+        if not end_date:
+            self.end_date = start_date
+            print(f"Setting end time to start time: {start_date}")
```
greglucas, line 209: warn/log instead of printing? There is no way to avoid this print otherwise.

imap_processing/cli.py
```diff
@@ -104,37 +148,101 @@ def _validate_args(args):
...
+        file_list: list[str]
+            A list of file paths to the downloaded dependencies.
+        """
+        file_list = []
+        for dependency in self.dependencies:
+            try:
+                # TODO: Validate dep dict
+                # TODO: determine what dependency information is optional
+                # TODO: Add in timestamps and descriptor to query
+                return_query = imap_data_access.query(
```
greglucas, line 228: This would mean I can't run local processing without an internet connection/access to the SDC servers? Do we need to add some checks in the query to add a "local" command to query the local files too? This can come in follow-up issues/PRs

imap_processing/cli.py
```diff
@@ -104,37 +148,101 @@ def _validate_args(args):
...
+
+            if not return_query:
+                raise FileNotFoundError(
+                    f"File not found for required dependency "
+                    f"{dependency} while attempting to create file."
+                    f"This should never occur "
+                    f"in normal processing."
+                )
+
+            file_list.append(imap_data_access.download(return_query[0]["file_path"]))
```
greglucas, line 244: Above you have the "Unable to download files" message. Should this be moved up into that try/except block so we also catch the HTTPError there?

How is return_query sorted? I would guess we want the latest file `[-1]` or all of them even by iterating over the list returned?

imap_processing/cli.py
```diff
@@ -147,63 +255,85 @@ class Codice(ProcessInstrument):
...
 
 class Swapi(ProcessInstrument):
     """Process SWAPI."""
 
     def process(self):
         """Perform SWAPI specific processing."""
-        print(f"Processing SWAPI {self.level}")
+        print(f"Processing SWAPI {self.data_level}")
+        # for dep in self.dependency:
+        #     imap_data_access.query()
```
greglucas: self.download_dependencies()? Or just remove the comments for now if not needed.

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -1,23 +1,23 @@
...
     """
     Process MAG L0 data into L1A CDF files at cdf_filepath.
 
     Parameters
     ----------
     packet_filepath:
         Packet files for processing
-    cdf_directory:
+    cdf_filepath:
         Directory for output
```
greglucas, line 17: cdf_filepath is not super descriptive to me of what that is referring to (cdf inputs, outputs, IMAP base directory), if this is where you want the output to go I'd suggest `output_filepath` as a more explicit naming. But, not blocking at all on this.

imap_processing/tests/idex/test_l1_cdfs.py
```diff
@@ -21,22 +21,30 @@ def decom_test_data():
     return PacketParser(test_file)
 
 
+@pytest.mark.skip(reason="Need to fix new ISTP error in IDEX CDF")
```
greglucas:
```suggestion
@pytest.mark.xfail(reason="Need to fix new ISTP error in IDEX CDF")
```

imap_processing/tests/idex/test_l1_cdfs.py
```diff
@@ -21,22 +21,30 @@ def decom_test_data():
...
-    file_name = write_cdf(decom_test_data.data, descriptor="test")
     date_to_test = "20250724"
-    assert file_name.name == (
+
+    input_name = Path(
         f"{decom_test_data.data.attrs['Logical_source']}_test_"
         f"{date_to_test}_{date_to_test}_v{idex.__version__}.cdf"
     )
+
+    file_name = write_cdf(decom_test_data.data, input_name)
```
greglucas: In the new implementation of the `write_cdf()` are we expanding the `input_name` to the imap-data-access data location? (Which in the test suite we set to a temporary directory)

I'm concerned that these are going to be written to the local directories and start adding files we don't want while running the tests. This is skipped currently, but taking off the skip locally a file was produced there.

greglucas (changes requested):
I am confused where data is supposed to live and just want that clarified before merging since there are a lot of binary files in various locations now.

We have a new `data/imap/mag` folder nested under testing, which I am not sure should be in the tests, or is it a test artifact that only needs to be produced and thus should be in temporary directories?

https://github.com/IMAP-Science-Operations-Center/imap_processing/tree/dev/imap_processing/data/imap/mag

We also have this file now too: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/imap_swe_l1_test_20100101_20100101_v01.cdf

I guess this is just a concern on my part that we will accidentally be doing `git commit .` and adding unnecessary/unexpected files.

imap_processing/cli.py
maxinelasp: Right now, we would expect only one file to return for a given dependency query. That's not the case right now, but Tenzin needs to update the dependency management in batch_starter first.

imap_processing/tests/idex/test_l1_cdfs.py
maxinelasp: Right, I'll update the input_name path to write to a temp directory, I didn't catch that. I'm also open to adding back in the `directory` keyword to `write_cdf`

maxinelasp:
I'll go in and check the tests to see if they're producing files outside of a temp directory. The `imap/mag` folder is just a fail from me adding it, I think, since that structure is from `imap-data-access`.

imap_processing/cli.py
maxinelasp: I'm not sure, this piece is left over from the first draft (just moved). If we want to change the JSON that would be in batch_starter.

tech3371: This is what get's passed to batch job(aka here). I took this from most recent logs
```
Submitting job with this command - ['--instrument', 'mag', '--data-level', 'l1b', '--start-date', '20231212', '--end-date', '20231212', '--version', 'v00-01', '--dependency', "[{'instrument': 'mag', 'data_level': 'l1a', 'version': 'v00-01'}]", '--upload-to-sdc']
```

imap_processing/cli.py
tech3371: To expand more on what Maxine said, before Maxine had above query like this
```
                return_query = imap_data_access.query(
                    instrument=dependency["instrument"],
                    data_level=dependency["data_level"],
                    version=dependency["version"],
                    start_date=dependency["start_date"],
                    end_date=dependency["end_date"],
                    descriptor=dependency["descriptor"],
                )
```
I asked her to temporarily changed that to this for now
```
                return_query = imap_data_access.query(
                    instrument=dependency["instrument"],
                    data_level=dependency["data_level"],
                    version=dependency["version"],
                )
```
The reason was because I don’t have all the information in the dependency list yet. That’s not done yet. Right now, below is what's passed to batch job, as I mentioned above. This is one of the TODO in the big action items list post demo ticket. I haven’t got around to it because that involves many changes such as database changes and make sure we didn’t mess up dependency checks. I know I am not gonna be able to make that change in time if I was to do it.
```
['--instrument', 'mag', '--data-level', 'l1b', '--start-date', '20231212', '--end-date', '20231212', '--version', 'v00-01', '--dependency', "[{'instrument': 'mag', 'data_level': 'l1a', 'version': 'v00-01'}]", '--upload-to-sdc']
```

greglucas:
> I'll go in and check the tests to see if they're producing files outside of a temp directory. The imap/mag folder is just a fail from me adding it, I think, since that structure is from imap-data-access.

Thanks! I think this was my primary concern and reason for not wanting to allow individual output locations at first, just so we could keep things a little more isolated somehow. I'm not sure what the best approach is here though because I don't think we'd want both the directory and path allowed as they might conflict if you pass in a full path. Is there a way to make the filepath argument optional and deduce the path structure from the xarray dataset passed in? So default to no path and then write everything to the imap-data-access location (temporary within the tests), but allow users to specify a path if they know what they are doing and want to push it somewhere else.

imap_processing/cli.py
greglucas: Sounds good! Just wanted to make sure this had a reason, and both of you are on it which is great :)

imap_processing/cli.py
greglucas: @tech3371 where does that command get created? I think it just needs a `json.dumps()` surrounding the dependency argument which will invert that string escaping for us.

tech3371: here: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/lambda_code/SDSCode/batch_starter.py#L315

maxinelasp:
> Is there a way to make the filepath argument optional and deduce the path structure from the xarray dataset passed in? 

I think this is a good idea... but do you want to block the PR on this? Ultimately I think the answer might be to pass in a `ScienceFilePath` instance or something, but I'm not sure. Ultimately, that question isn't super related to the CLI changes, and I don't think it's worth trying to fix here, since this PR has such an impact on sds-data-manager. 

I did remove the files you were concerned about (thanks for catching that!) and update the tests that generate CDFs - can you take another look?

greglucas (approved):

greglucas:
> I did remove the files you were concerned about (thanks for catching that!) and update the tests that generate CDFs - can you take another look?

Yep, this was what I was blocking over, so thanks for fixing it! Merge away before you take off :)

PR #342, opened 2024-02-15
HIT housekeeping data class

sdhoyt:
# Change Summary

## Overview
Added a housekeeping data class for decommutation to L1A.

## New Dependencies
None

## New Files
- housekeeping.py
   - new data class for HIT housekeeping L1A data
- hit_base.py
   - base data class for all of the HIT L1A data classes.

## Deleted Files
None

## Updated Files
- hit.rst
   - added new files to docs
- Updated the XTCE files based on new Telem definitions

## Testing
Added test_houskeeping.py to unit test the housekeeping data. This unit test includes validation for CCSDS decommutation.


imap_processing/hit/l0/data_classes/housekeeping.py
```diff
@@ -0,0 +1,219 @@
...
+        # Each Leak field is 10 bits long
+        leak_bit_length = 10
+        # There are 64 leak fields
+        num_leak_fields = 64
+        # The leak fields appear in the packet in ascending order, so to append
+        # the leak fields in the correct order, the binary will be parsed 
+        # from right to left.
+        for leak_idx in range(leak_bit_length * num_leak_fields, 0, -leak_bit_length):
+            leak_list.append(leak_bits[leak_idx - leak_bit_length:leak_idx].uint)
+        self.LEAK_I = np.array(leak_list)
```
greglucas: You can initialize the shape of the array right away and then fill in the values and avoid bitstring as well I think.
```suggestion
        # Each Leak field is 10 bits long
        leak_bit_length = 10
        # There are 64 leak fields
        num_leak_fields = 64
        self.LEAK_I = np.empty(num_leak_fields, dtype=np.uint16)
        # The leak fields appear in the packet in ascending order, so to append
        # the leak fields in the correct order, the binary will be parsed 
        # from right to left.
        for i, leak_idx in enumerate(range(leak_bit_length * num_leak_fields, 0, -leak_bit_length)):
            self.LEAK_I[i] = int(self.LEAK_I_RAW[leak_idx - leak_bit_length:leak_idx], 2)
            # or something like this to iterate through and then just fill backwards:
            # self.LEAK_I[-i] = int(self.LEAK_I_RAW[i * leak_bit_length:(i + 1) * leak_bit_length], 2)
```

Also, are the bits supposed to be read in reverse as well (little-endian vs big-endian matter). Or are they just stored in reverse?

imap_processing/hit/l0/utils/hit_base.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        attributes = [field.name for field in fields(self)]
+
+        # For each item in packet, assign it to the matching attribute in the class.
+        for key, item in packet.data.items():
+            value = (
+                item.derived_value if item.derived_value is not None else item.raw_value
+            )
+            if "SPARE" in key:
+                continue
```
greglucas, line 47: I wonder if there is a way to get XTCE / space packet parser to ignore a field like this?

imap_processing/hit/l0/utils/hit_base.py
```diff
@@ -0,0 +1,53 @@
...
+            if key not in attributes and "SPARE" not in key:
```
greglucas: You already checked for it above I think
```suggestion
            if key not in attributes:
```

imap_processing/tests/hit/test_housekeeping.py
```diff
@@ -0,0 +1,92 @@
...
+from imap_processing.hit.l0.data_classes.housekeeping import Housekeeping
+
+
+def test_houskeeping():
+    # The HK validation data's first row was not in the CCSDS file sent, so that has
+    # been manually removed from the validation file on the SDC end. The last
+    # packet in the CCSDS file also does not correspond to any of the rows in
+    # the validation file, so the last packet in the CCSDS file is removed for
+    # this test. These issues were discovered / confirmed with the HIT Ops
+    # engineer who delivered the data.
```
greglucas, line 14: What joy :) Thanks for your patience in finding these! And especially documenting this.

imap_processing/tests/hit/test_housekeeping.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    xtce_file = (
+        pathlib.Path(__file__).parent.parent.parent
+        / "hit/packet_definitions/P_HIT_HSKP.xml"
+    )
+
+    validation_data = pd.read_csv(validation_file)
+    leak_columns = [col for col in validation_data.columns if col.startswith("LEAK")]
+    packets = decom.decom_packets(test_file.resolve(), xtce_file.resolve())
+    del packets[-1]
```
greglucas:
```suggestion
    packets = decom.decom_packets(test_file.resolve(), xtce_file.resolve())[:-1]
```

imap_processing/tests/hit/test_housekeeping.py
```diff
@@ -0,0 +1,92 @@
...
+    pkt_idx = 0
+
+    for packet in packets:
```
greglucas: I'd remove the pkt_idx tracking yourself and let enumerate capture it.
```suggestion
    for pkt_idx, packet in enumerate(packets):
```

tools/xtce_generation/xtce_generator_hit.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+
+from telemetry_generator import TelemetryGenerator
+
+
+def main():
+    """HIT XTCE generator"""
+    instrument_name = "hit"
+    current_directory = Path(__file__).parent
+    module_path = f"{current_directory}/../../imap_processing"
```
greglucas: import imap_module_path from the main package?

imap_processing/hit/l0/data_classes/housekeeping.py
sdhoyt: They're just stored in reverse. There's 64 leak fields in the packet and they're ordered as leak_64, leak_63, .... so I just need to reverse the field order when I make the array

imap_processing/hit/l0/utils/hit_base.py
sdhoyt: Good question. I'll check the docs

imap_processing/hit/l0/utils/hit_base.py
sdhoyt: forgot to remove that part. thanks!

imap_processing/hit/l0/data_classes/housekeeping.py
sdhoyt: Are we trying to move away from the `bitstring` library entirely?

greglucas: IMO we should. Let bitstring be a detail for space_packet_parser, but not bring it into imap_processing if we can avoid it. I found it to be really slow. If we need to do complicated bit manipulations then I don't see much of an issue, but everything I've seen so far is not _too_ bad to handle ourselves. Note that bitstring has already made this a string for you, so you're sort of going back to bitstring then rather than using the original bits.

sdhoyt:
I'm getting a "Too many statements" ruff error (PLR0915) because of the length of my `test_houskeeping.py` unit test function. Is this something we think can be ignored or should I break this test up? It seems weird to spread testing class attributes across multiple functions.

greglucas:
> I'm getting a "Too many statements" ruff error (PLR0915) because of the length of my test_houskeeping.py unit test function. Is this something we think can be ignored or should I break this test up? It seems weird to spread testing class attributes across multiple functions.

Yes, you can ignore it IMO. But, if you're looking to reduce boilerplate you have to write you could do something like this for many of those assertions:
```python
# Turn this
assert hk.LAST_BAD_CMD == validation_data["LAST_BAD_CMD"][pkt_idx]
# into:
keys = ["LAST_BAD_CMD", ...]
for key in keys:
    assert getattr(hk, key) == validation_data[key][pkt_idx]
```

docs/source/code-documentation/hit.rst
```diff
@@ -15,4 +15,6 @@ The L0 code to decommutate the CCSDS packet data can be found below.
     :template: autosummary.rst
     :recursive:
 
-    l0.hit_l1a_decom
+    l0.data_classes
+    l0.utils
+    l0.hit_l1a_decom.py
```
greglucas: I think the extra `.py` might be causing the doc build failure.
```suggestion
    l0.hit_l1a_decom
```

imap_processing/tests/hit/test_housekeeping.py
```diff
@@ -0,0 +1,96 @@
...
+    for pkt_idx, packet in enumerate(packets):
+        hk = Housekeeping(packet, "0.0", "hskp_sample.ccsds")
+        for field in hk_fields:
+            if field == "LEAK_I":
+                # Need to create an array out of the separate leak columns in the
+                # validation data
+                assert (
+                    hk.LEAK_I
+                    == np.array(validation_data.loc[pkt_idx, leak_columns].tolist())
+                ).all()
```
greglucas: Definitely a nit, so feel free to ignore, but I would pull this out and test it individually before the loop since this is only one odd-ball. Also, numpy has nice array testing utilities, which will also print out which entries are failing if you want to start using those for the array tests too.

```suggestion
                # Need to create an array out of the separate leak columns in the
                # validation data
                np.testing.assert_array_equal(hk.LEAK_I, validation_data.loc[pkt_idx, leak_columns])
```

imap_processing/hit/packet_definitions/P_HIT_IALRT.xml
greglucas, line 1: Do you want to keep the I-ALiRT definition in the repo? @laspsandoval might want to reference it and keep it in, although I think she is creating a new one manually from these as reference so this might not be explicitly needed.

greglucas (approved):
Looks good to me. Just need to fix the doc build.

imap_processing/hit/packet_definitions/P_HIT_IALRT.xml
sdhoyt: Yeah, I can remove the packet definition. I forgot to ignore it when I did one of the xtce generations.

imap_processing/tests/hit/test_housekeeping.py
sdhoyt: That's cool, I didn't realize numpy had the testing module

imap_processing/hit/l0/utils/hit_base.py
```diff
@@ -0,0 +1,53 @@
+"""General HIT L0 data class used for parsing data and setting attributes."""
+from dataclasses import dataclass, fields
+
+from imap_processing.ccsds.ccsds_data import CcsdsData
+
+
+@dataclass
+class HITBase:
```
maxinelasp, line 8: Nice, I like this

imap_processing/hit/packet_definitions/P_HIT_AUT.xml
```diff
@@ -50,11 +47,11 @@
 			<xtce:Parameter name="PKT_LEN" parameterTypeRef="UINT16">
 				<xtce:LongDescription>CCSDS Packet Length (number of bytes after Packet length minus 1)</xtce:LongDescription>
 			</xtce:Parameter>
-			<xtce:Parameter name="SHCOARSE" parameterTypeRef="UINT32">
-				<xtce:ShortDescription>CCSDS Packet Sec Header</xtce:ShortDescription>
+			<xtce:Parameter name="SC_TICK" parameterTypeRef="UINT32">
```
maxinelasp: I thought SHCOARSE or MET was required - is it missing here because it's a housekeeping packet?

imap_processing/hit/packet_definitions/P_HIT_HSKP.xml
```diff
@@ -63,7 +71,7 @@
 				<xtce:LongDescription>CCSDS Packet Length (number of bytes after Packet length minus 1)</xtce:LongDescription>
 			</xtce:Parameter>
 			<xtce:Parameter name="SHCOARSE" parameterTypeRef="UINT32">
```
maxinelasp, line 73: But here, it's still named SHCOARSE? Why?

imap_processing/tests/hit/test_housekeeping.py
```diff
@@ -0,0 +1,91 @@
...
+    # the validation file, so the last packet in the CCSDS file is removed for
+    # this test. These issues were discovered / confirmed with the HIT Ops
+    # engineer who delivered the data.
+    test_file = imap_module_directory / "tests/hit/test_data/hskp_sample.ccsds"
+    validation_file = (
+        imap_module_directory / "tests/hit/validation_data/hskp_sample_raw.csv"
+    )
+    xtce_file = imap_module_directory / "hit/packet_definitions/P_HIT_HSKP.xml"
+
+    validation_data = pd.read_csv(validation_file)
```
maxinelasp, line 21: Nice, I like this method of validation

maxinelasp (approved):
Looks good!

imap_processing/hit/packet_definitions/P_HIT_AUT.xml
sdhoyt: No you're correct. HIT is calling it SC_TICK in their Telem definition and I guess I forgot to change it back to SHCOARSE when I got an updated definition sheet and re-generated the XTCE. I'm going to remove all the HIT XTCE that isn't housekeeping.

imap_processing/hit/packet_definitions/P_HIT_HSKP.xml
sdhoyt: Yeah, I did remember to change the updated sheet to SHCOARSE for housekeeping since this is the one I was working on for this PR. Going to remove all non-hk xtce  for this PR

PR #345, opened 2024-02-19
HIT Message Log Data Class

sdhoyt:
# Change Summary

## Overview
This PR creates a data class for the HIT Message Log packet. This is a simple packet containing only Spacecraft time and a variable length text field. This PR includes validation of the test data decommutation. HIT asked for the message text to remain in binary form, so there is no conversion to strings.

## New Dependencies
None

## New Files
- message_log.py
- P_HIT_MSGLOG.xml

## Deleted Files
None

## Updated Files
None

## Testing
created `test_message_log.py` to test the decommutation using validation data delivered by HIT.


imap_processing/tests/hit/test_message_log.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+    xtce_file = imap_module_directory / "hit/packet_definitions/P_HIT_MSGLOG.xml"
+
+    validation_data = pd.read_csv(validation_file)
+    packets = decom.decom_packets(test_file.resolve(), xtce_file.resolve())
+
+    for pkt_idx, packet in enumerate(packets):
+        msg_log = MessageLog(packet, "0.0", "mslog_sample.ccsds")
+        assert msg_log.SHCOARSE == validation_data["SHCOARSE"][pkt_idx]
+        assert msg_log.TEXT == bitstring.Bits(hex=validation_data["TEXT"][pkt_idx]).bin
```
greglucas, line 24: Some questions here: Do you want the message log as a string of binary 0s and 1s? Or do you want that as a bytes-like object? `b"TEXT"` or something encoded/decoded with ascii ("TEXT".encode("ascii"))

I'm wondering if you actually want to avoid bitstring here and compare against the specific validation_data provided rather than manipulating that?

greglucas (approved):
I do question whether we want to be doing this on the SDC side of things. These messages are theoretically for the POC and operations folks. Are they relevant to science data processing?

imap_processing/tests/hit/test_message_log.py
tech3371: good question. Curious to know what this does
```
bitstring.Bits(hex=validation_data["TEXT"][pkt_idx]).bin
```

sdhoyt: Eric Christian said the messages could be left in binary for the L1A product. The validation data is in hex so I used the Bits class to convert to binary (there might be better way to do it?). I could double check with them on how they want the data stored. This is one of those products that doesn't really make sense for us to be producing an L1A file for, so maybe I should push back on this product as a whole?

sdhoyt:
HIT has indicated that this data product does not need to be produced by the SDC. Waiting for confirmation from them before closing the PR.

sdhoyt:
Still waiting for confirmation from HIT that they don't need an L1A message log product, but going to close this anyway because I'm expecting they won't.

PR #349, opened 2024-02-23
MAG CDF updates

maxinelasp:
# Change Summary
Updating mag to produce burst CDF files as well as norm

## Overview

- Updated mag CDF generation to write burst and norm data into different datasets
- Updated L1A mag processing to handle production of 2 files
- Addressed a bug that came up with a test file provided by mag, which broke with differing length vectors

## Updated Files

- cli.py/mag_l1a.py
   - Updated to create two files

- decom_mag.py
   - moved xarray creation into a new method, updated to split data by norm and burst 
   - 
## Testing

Added some additional testing around xarray and CDF generation


maxinelasp:
I don't need this merged before I leave (although it would be nice...), but I wanted to put it up for review since it is what I deployed and will leave running for the MAG team.

imap_processing/mag/l0/decom_mag.py
```diff
@@ -71,58 +71,118 @@ def export_to_xarray(l0_data: list[MagL0]):
...
+    shcoarse_data = np.zeros(len(l0_data))
+
+    support_data = defaultdict(list)
+
+    for index, datapoint in enumerate(l0_data):
+        vector_len = len(datapoint.VECTORS)
+        if vector_len > vector_data.shape[1]:
+            # If the new vector is longer than the existing shape, first reshape
+            # vector_data and pad the existing vectors with zeros.
+            vector_data = np.pad(
```
maxinelasp, line 129: I have no clue why this formatting is so weird... if anyone knows how to make ruff happy here I'd be happy to hear it

imap_processing/cli.py
```diff
@@ -22,6 +23,8 @@
 
 import imap_processing
 
+__logger__ = logging.getLogger(__name__)
```
greglucas: Our pattern has been to set the local logger variable and then `logger.info()` later.
```suggestion
logger = logging.getLogger(__name__)
```

imap_processing/mag/l0/decom_mag.py
```diff
@@ -71,58 +71,118 @@ def export_to_xarray(l0_data: list[MagL0]):
...
+        xarray dataset for generating burst data CDFs
     """
     # TODO split by mago and magi using primary sensor
     # TODO split by norm and burst
-    norm_data = defaultdict(list)
-    burst_data = norm_data.copy()
+    norm_data = []
+    burst_data = []
 
     for datapoint in l0_data:
```
greglucas: Call this packet?
I consider datapoint a measurement, but I think your packet could contain multiple datapoints.

imap_processing/mag/l0/decom_mag.py
```diff
@@ -71,58 +71,118 @@ def export_to_xarray(l0_data: list[MagL0]):
...
+
+    Returns
+    -------
+    norm_data : xr.Dataset
+        xarray dataset for generating burst data CDFs
+    burst_data : xr.Dataset
+        xarray dataset for generating burst data CDFs
     """
     # TODO split by mago and magi using primary sensor
     # TODO split by norm and burst
```
greglucas: Remove this TODO now?

imap_processing/mag/l0/decom_mag.py
```diff
@@ -71,58 +71,118 @@ def export_to_xarray(l0_data: list[MagL0]):
...
+
+    norm_dataset = None
+    burst_dataset = None
+
+    if len(norm_data) > 0:
+        norm_dataset = generate_dataset(norm_data)
+    if len(burst_data) > 0:
+        burst_dataset = generate_dataset(burst_data)
+
+    return norm_dataset, burst_dataset
```
greglucas, line 100: Rather than returning these as a two-tuple, would you possibly want to return a single dictionary similar to what you had before?
`datasets["norm"] = norm_dataset`
That might make it easier to extend later with `datasets["outboard"]["norm"]`

imap_processing/mag/l0/decom_mag.py
```diff
@@ -71,58 +71,118 @@ def export_to_xarray(l0_data: list[MagL0]):
...
+    ----------
+    l0_data : list[MagL0]
+        List of sorted L0 MAG data.
+
+    Returns
+    -------
+    dataset : xr.Dataset
+        xarray dataset with proper CDF attributes and shape.
+    """
+    vector_data = np.zeros((len(l0_data), len(l0_data[0].VECTORS)))
```
greglucas, line 119: Rather than padding below, you might be able to calculate it upfront here with a max.
```suggestion
    vector_data = np.zeros((len(l0_data), max(lambda x: len(x.VECTORS) for x in l0_data))
```

imap_processing/mag/l0/decom_mag.py
greglucas: That is a common complaint with the auto-formatters from the scientific community and why many people hate using them. I think ruff might have some flags to try and make this better, but I gave another suggestion above to avoid the padding entirely which might be easier.

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -1,23 +1,34 @@
 """Methods for decomming packets, processing to level 1A, and writing CDFs for MAG."""
+import logging
 from pathlib import Path
 
 from imap_processing.cdf.utils import write_cdf
 from imap_processing.mag.l0 import decom_mag
 
+__logger__ = logging.getLogger(__name__)
```
greglucas: nit: `logger =`

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -35,15 +36,49 @@ def test_mag_decom():
...
-    assert all([item in list(output_data.attrs.keys()) for item in required_attrs])
-    assert all([item is not None for _, item in output_data.attrs.items()])
+    assert all([item in list(norm_data.attrs.keys()) for item in required_attrs])
+    assert all([item is not None for _, item in norm_data.attrs.items()])
+
+    assert all([item in list(burst_data.attrs.keys()) for item in required_attrs])
+    assert all([item is not None for _, item in burst_data.attrs.items()])
+
+    expected_norm_len = 17
+    print(norm_data.dims["Epoch"])
```
greglucas:
```suggestion
```

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -35,15 +36,49 @@ def test_mag_decom():
...
+    output = write_cdf(norm_data, test_data_path_norm)
+    assert Path.exists(test_data_path_norm)
+
+    input_xarray = cdf_to_xarray(output)
+    assert input_xarray.attrs.keys() == norm_data.attrs.keys()
+
+    test_data_path_burst = tmp_path / "mag_l1a_raw-burst_20210101_20210102_v01-01.cdf"
+
+    output = write_cdf(burst_data, test_data_path_burst)
+    assert Path.exists(test_data_path_burst)
```
greglucas: Might want to assert it doesn't exist before the functions as well? (Similar above this)
```suggestion
    assert test_data_path_burst.exists()
```

greglucas (commented):
Just some minor nits / suggestions. Otherwise looks good!

imap_processing/mag/l0/decom_mag.py
maxinelasp: That requires looping through all the data, whereas what I have loops through once and takes the performance hit on resizing the np array. I think the array is not likely to get expanded very often, so this is better performance overall (also, because np in general is way faster than normal python stuff)

imap_processing/mag/l0/decom_mag.py
maxinelasp: Yeah, I like that

imap_processing/mag/l0/decom_mag.py
greglucas: :+1: Maybe a question for another time then is why are we getting resizes at all, is this expected? I would have figured the packets to all contain X number of vectors.

maxinelasp: No, the vector count can vary based on the settings of the instrument

greglucas: Interesting, I would have expected it to only be a difference between burst and normal mode, but otherwise, be measured at the same frequency/cadence.

Another follow-on here then is whether `0` is an OK fill value for this data, or do you want `np.nan` to indicate that the entry didn't have any vector measurements past a certain point?

I guess I'm not entirely clear on what the shifting shape means. I think the shape is `(npackets, ndirections)` by looking later at the xarray dataset, but is it OK to have a shifting directions variable, or do we need different `ndirections` variables within the same dataset to distinguish these? A linearly spaced ndirections instead of only at the beginning (padding 0s to the end)?

I might be getting confused with thinking about higher level data products though, and just don't understand enough about the lower level organization to expect here. So just keep these questions in mind, but you don't need to address anything now.

greglucas (approved):

imap_processing/mag/l0/decom_mag.py
maxinelasp: I actually might come back to this when I add the rest of the L1A processing. I don't usually like returning dictionaries like that if I have a known number of outputs, and I think I will come up with something better when I add in the other kinds of processing...

imap_processing/mag/l0/decom_mag.py
maxinelasp: I agree that "direction" is not the correct axis for this particular CDF. It should be something else, with "Direction" being every other data product. But, I am planning to discuss this with mag when I get back. They didn't reply with their thoughts before now, so it will have to stay as my first draft for now 🤷 I asked if 0 was an ok fill value and they said yes. The byte vectors are varying lengths but you can calculate the number of fixed-length vectors from other data so the length per packet is known when processing.

PR #356, opened 2024-03-04
Added l1a cdf generation for all apids

laspsandoval:
# Change Summary

## Overview
Creates xarray and cdf for all apids (previously only auxiliary apid)

## Updated Files
- decom_ultra.py
   - minor formatting
- ultra_l1a.py
   - Creates xarray and cdf for all apids (previously only auxiliary apid)

## Testing
- test_ultra_l1a.py
   - Tests the creation of the xarray and cdf


imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -38,13 +41,70 @@ def xarray_aux(decom_ultra_aux: dict):
...
+
+        dataset = xr.Dataset(
+            coords={"Epoch": epoch_time, "Row": row, "Column": column},
+            attrs=ultra_cdf_attrs.ultra_l1a_attrs.output(),
+        )
+
+    return dataset
+
+
+def xarray(decom_ultra: dict, apid: int):
```
greglucas: This would shadow `xarray` itself so causes some confusion. Adding a verb to the function would help here.
```suggestion
def create_dataset(decom_ultra: dict, apid: int):
```

tech3371 (approved):
I agree with Greg on updating function name. Besides that, it looks good to me!

imap_processing/tests/ultra/unit/test_ultra_l1a.py
```diff
@@ -59,15 +87,152 @@ def test_xarray_aux(decom_ultra_aux, aux_test_path):
...
+    specific_epoch_data = dataset.sel(Epoch="2022-05-30T22:52:00.184000")["START_RF"]
+    startrf_list = specific_epoch_data.values.tolist()
+    startrf_attr = dataset.variables["START_RF"].attrs
+
+    expected_startrf_attr = dataclasses.replace(
+        ultra_cdf_attrs.ultra_metadata_attrs,
+        catdesc="START_RF",
+        fieldname="START_RF",
+        label_axis="START_RF",
+    ).output()
```
bryan-harter, line 107: How much should I be commenting on the internals of the CDFs in these reviews do you all think? For example, I know the SPDF would complain about these catdesc, fieldname, label_axis. But then again, our priority in the coming months should probably focus on the actual algorithms themselves?

greglucas: I think you should flag it immediately when you see it! Much better to fix these things earlier and then others on the team can keep iterating and fixing their data products as we go. I think you should be as detailed as possible in your reviews, it is super helpful!

d-linko (approved):

imap_processing/tests/ultra/unit/test_ultra_l1a.py
laspsandoval: @bryan-harter yes please comment at will! I would rather get it right the first time rather than go back again later. There will likely be enough of that already!

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -73,27 +135,43 @@ def xarray_aux(decom_ultra_aux: dict):
                 catdesc=key,  # TODO: short and long descriptions
                 fieldname=key,
             ).output()
+            dims = ["Epoch"]
+        # TOF packetdata has multiple dimensions
+        elif key == "PACKETDATA":
+            attrs = dataclasses.replace(
+                ultra_cdf_attrs.ultra_metadata_attrs,
```
bryan-harter: For things that aren't strings, you'll probably just want to use "support_data" rather than "metadata".

bryan-harter:
Sorry I missed the first review of the L1a products, so I'll write some stuff here in general about them. 

So all of your files definitely meet the letter of the ISTP law, and that may be good enough, but I was going to comment about other things that Andriy and Bobby have told me in the past.  At today's SDWG meeting, the SPDF said that L1 products wouldn't be displayed on CDAWeb, so some of these I might not even worry about for L1 products.

I'll post my comments throughout your pull request so we can start different chains to discuss if needed

bryan-harter:
All the variables are display_type=no_plot, which I wasn't sure if the SPDF would ask about.  I was going to say if there was an important variable somewhere that your instrument team might want to plot, it might be worth adding the real plot type

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -38,13 +41,70 @@ def xarray_aux(decom_ultra_aux: dict):
...
+        )
+    else:
+        row = xr.DataArray(
+            # Number of pixel rows
+            np.arange(54),
+            name="Row",
+            dims=["Row"],
+            attrs=dataclasses.replace(
+                ultra_cdf_attrs.ultra_support_attrs,
+                catdesc="ROW",  # TODO: short and long descriptions
```
bryan-harter: Yeah as you note, catdesc, fieldnam, and lablaxis would need better descriptions for each variable.  Reminder than catdesc is about 80 characters, fieldnam is about 20-30 characters, and lablaxis is like, 6 characters.  

Additionally, the SPDF said we should include better VAR_NOTES for my primary variables

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -38,13 +41,70 @@ def xarray_aux(decom_ultra_aux: dict):
...
+                fieldname="ROW",
+                var_type="ignore_data",
```
bryan-harter: Should this be support_data instead of ignore? Especially since they're used as coordinates in the xarray below?

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -38,13 +41,70 @@ def xarray_aux(decom_ultra_aux: dict):
...
+        Packet APID.
+
+    Returns
+    -------
+    dataset : xarray.Dataset
+        Data in xarray format.
+    """
+    dataset = initiate_data_arrays(decom_ultra, apid)
+
+    for key, value in decom_ultra.items():
```
bryan-harter, line 187: If you have any variables in here that are more important that the others, primary variables could use an additional attribute of VARIABLE_PURPOSE=PRIMARY_VAR

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -28,7 +31,7 @@ def xarray_aux(decom_ultra_aux: dict):
     """
     # Converted time
     time_converted = []
-    for time in decom_ultra_aux["SHCOARSE"]:
+    for time in decom_ultra["SHCOARSE"]:
```
bryan-harter: We had discussed this in our meeting on monday, but Epoch must always be monotonically increasing.  So you can't have duplicate epoch numbers in your variable, and you'd unfortunately have to make a separate dimension to hold them.

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -73,27 +135,43 @@ def xarray_aux(decom_ultra_aux: dict):
...
                 ultra_cdf_attrs.ultra_metadata_attrs,
                 catdesc=key,  # TODO: short and long descriptions
                 fieldname=key,
                 label_axis=key,
             ).output()
+            dims = ["Epoch"]
 
         dataset[key] = xr.DataArray(
             value,
             name=key,
```
bryan-harter: Technically, in our CDF documentation, we mentioned that variable names would all be lower case with fields separated by underscores.  I'm totally aware that we haven't been doing that so far, including in my IDEX files, but I was copying what was the standard convention for MMS when I wrote that section.  

So for example, "COIN_TYPE" might now be "ultra_coin-type", or more specifically, "ultra_coin-type_aux_l1a"

laspsandoval: I'll bring this up with the ultra instrument teams. But put as a TODO for now.

PR #360, opened 2024-03-13
TST: Rely on temp directory for test file cleanup

greglucas:
When these files failed before the unlink() step, the files got left behind in the local directory. Use pytest's tmppath fixture to clean this up automatically.


tech3371 (approved):
Thank you!

PR #362, opened 2024-03-13
MNT: Use module-level logger

greglucas:
Rather than using the global logger and updating the configuration globally, we should get a logger instance for each module that wants it and submit log messages to that specific logger.


bourque (approved):
Good idea!

PR #363, opened 2024-03-13
MNT: Update file naming convention via imap-data-access upgrade

greglucas:
# Change Summary

## Overview

Bumps imap-data-access to v0.5.0 where there is a new filename requirement on the mission.

This is a major update to the filename convention, so there are lots of areas touched in this PR. Most are just updating the filenaming, but there is one other major update to the `write_cdf()` routine.

`write_cdf()` was changed to get all of its filename information from the attributes of the passed-in dataset directly, and uses ScienceFilePath from `imap-data-access`. There is an _optional_ `directory` argument if you do want to write that dataset somewhere else, but I think this makes it a bit cleaner without requiring every instrument to create their filenames, but instead focus on the metadata, and then you get the filename for free.

## New Dependencies
`imap-data-access>=0.5`


greglucas:
Also note that I have just set all descriptors to "sci" for now. I'm not sure whether we are setting those in the CDF somewhere, or if we want to just set the CDF metadata attributes with that information directly. This is because the `Logical_source` previously did **not** contain the descriptor field, but now we are told it should. So it somewhat changes up the structure of attribute inheritance for all of the instruments and I wasn't sure how we wanted to do that, so figured it was easiest to leave for each instrument to follow-up individually themselves after this goes in to correct it.

greglucas:
Some notes/thoughts:
* We need to standardize "Epoch" vs "epoch" in particular, and other variable names. Our documentation is not clear on what we need to do as it includes both styles: https://imap-processing.readthedocs.io/en/latest/external-tools/cdf/cdf_requirements.html#required-epoch-variable Currently, this causes issues if we want to infer filenames from that variable.
* We should switch from `l1a` based global descriptor to per file based global descriptors. This is in regards to this comment: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/363#issuecomment-1996006461 where I was going to push that off, but maybe I should just update these descriptors to be per file right now in this PR?

laspsandoval (approved):
Looks good to me.

docs/source/code-documentation/tools/xarray-to-cdf.rst
```diff
@@ -126,15 +126,15 @@ The following python code is the **minimum** code you'd need to store this data
...
+      "Data_version": "001",
       "Descriptor": "JIM>Just an Ion Monitor",
       "TEXT": (
          "JIM is a fictitious instrument that counts ions at 3 different energies on "
          "the IMAP mission. This is where a detailed description of the instrument "
          "goes, as well as the type of data in the file.  For example, if a (mode) "
          "or (descriptor) exist they can be described here."
       ),
-      "Logical_file_id": "imap_jim_l1_mode_description_20250101_v01",
+      "Logical_file_id": "imap_jim_l1_mode_description_20250101_v001",
```
tech3371:
```suggestion
      "Logical_file_id": "imap_jim_l1_mode-description_20250101_v001",
```

imap_processing/cli.py
```diff
@@ -154,7 +150,7 @@ def _validate_args(args):
     if args.instrument not in imap_data_access.VALID_INSTRUMENTS:
         raise ValueError(
             f"{args.instrument} is not in the supported instrument list: "
-            f"{imap_processing.INSTRUMENTS}"
+            f"{imap_data_access.VALID_INSTRUMENTS}"
         )
     if args.data_level not in imap_processing.PROCESSING_LEVELS[args.instrument]:
```
tech3371, line 155: should we also use valid data level from `imap-data-access`?

imap_processing/codice/cdf_attrs.py
```diff
@@ -44,14 +44,14 @@
...
-    logical_source="imap_codice_l1a",
+    logical_source="imap_codice_l1a_sci",
     logical_source_desc="IMAP Mission CoDICE Instrument Level-1A Data",
     instrument_base=codice_base,
 )
 
 codice_l1b_global_attrs = GlobalDataLevelAttrs(
     data_type="L1B->Level-1B",
-    logical_source="imap_cpdice_l1b",
+    logical_source="imap_cpdice_l1b_sci",
```
tech3371: is that a typo?
```suggestion
    logical_source="imap_codice_l1b_sci",
```

tech3371 (approved):
couple minor comments. Otherwise, it looks good to me!

docs/source/code-documentation/tools/xarray-to-cdf.rst
greglucas: fixed, thanks!

imap_processing/codice/cdf_attrs.py
greglucas: good catch, I was only updating the latter thing and not looking at the text before the "sci" portion! Fixed now.

imap_processing/cli.py
greglucas: I looked at this, and the VALID_PROCESSING_LEVELS is generic over in that repository and not "per instrument" like we are checking here. So I held off on that for now.

I agree though, we should look at making that update, but I think we should push it off to a follow-up PR.

tech3371: Yeah, that sounds good!

tech3371:
Looks great. Thank you so much for doing this!

greglucas:
OK, I just pushed up one more commit with a bunch of TODOs that will hopefully help everyone with what the previous naming of the file/descriptor was to know what to go back to once we update the global attributes.

bourque (approved):
This looks good! And thanks for making the changes for IDEX and CoDICE!

PR #365, opened 2024-03-18
Remove Warnings

laspsandoval:
# Change Summary

## Overview
There were warnings when using space_packet_parser. These changes remove the warnings.

## Updated Files
- ULTRA_SCI_COMBINED.xml
   - Added abstract="true" so that CCSDS would be seen as the header and not just a packet.
   - Changed equation for calculating apid 896 (Events) so that it uses PKT_LEN instead of COUNT.
- decom_tools.py
   - Put in restraint of 166 bits per event.


greglucas (approved):
I tested this on the tests that were emitting a bunch of WARNING log messages before and this seems to fix those. Nice work tracking this down!

PR #366, opened 2024-03-19
MNT: Rename Epoch to epoch lowercased for variable names

greglucas:
# Change Summary

## Overview

As discussed on the call and with SPDF, we can use any variable names we want for the DEPEND_* variables. For consistency we are moving towards lowercase variable names. This is a heavy handed find/replace of all "Epoch" and replacing it with "epoch". I reverted some of the documentation and label axis names to be capitalized as they were more used for sentences rather than as variable names.

I didn't update other variable names in this go-around, but can if people want the full find/replace now.


docs/source/code-documentation/tools/xarray-to-cdf.rst
```diff
@@ -205,30 +205,30 @@ The following python code is the **minimum** code you'd need to store this data
...
                   data=[1,2,3], # Fabricated energy bins
-                  dims=("Energy_dim"),
+                  dims=("energy_dim"),
                   attrs=energy_bins_attrs,
                )
    counts_xr = xr.DataArray(
                name="imap_jim_counts_(mode)_(descriptor)",
                data=[[1,2,3], [4,5,6], [7,8,9]], # Fabricated data
-               dims=("Epoch", "Energy_dim"),
+               dims=("epoch", "Energy_dim"),
```
laspsandoval: 'energy_dim'

laspsandoval (approved):
Good gracious. Sorry you had to do that! But thank you. One single comment. Looks good.

maxinelasp (approved):
LGTM, might want to mention in slack once you merge that people will need to go update their other CDF coordinates to match the all-lowercase convention. This is a good change though, now it will flag my brain to make everything match 😆

bourque (approved):
Looks good! I agree with others -- I think this is the right way to go.

docs/source/code-documentation/tools/xarray-to-cdf.rst
greglucas: Good catch, thanks! Updated this one.

PR #367, opened 2024-03-19
Docs for calibration files

maxinelasp:
# Change Summary
These are **PROPOSED** docs for calibration files. This has not been discussed by the SDC at all, this is a design I came up with during discussions with the MAG team. Please bring up any and all comments and concerns about calibration files. This is simply a rough draft to start somewhere.

### DIFF
--- a/docs/source/data-access-api/calibration-files.rst
+++ b/docs/source/data-access-api/calibration-files.rst
@@ -1,0 +1,146 @@
+.. _calibration_files:
+
+Calibration Files
+=================
+
+Calibration files are files used in processing that are provided by instrument teams
+to the SDC. These are files which contain values that could change over the course of
+the mission. Although they are called calibration files here, they may not contain calibration data,
+but instead any kind of data which is used in processing, but may infrequently change in time.
+
+Any values which could be changed by the instrument team should be in a calibration file.
+
+These files will be tracked in the SDC and uploaded to the `SPDF archive <https://spdf.gsfc.nasa.gov/>`_.
+Therefore, they should use unique filenames which clearly distinguish them from each other. Beyond that, they
+can follow any format that is convenient for the instrument team and the SDC.
+
+This document lays out the requirements for calibration files in the SDC and the expected file naming
+convention.
+
+File Naming Convention
+----------------------
+
+File overview
+^^^^^^^^^^^^^^
+
+In this document, I describe two main kinds of calibration files:
+
+#. Universal calibration files
+#. Time window calibration files
+
+**Universal calibration files** are files which are expected to change rarely. Generally, there should be one file
+which covers all times.
+
+**Time window calibration files** are files which are expected to change more frequently, and are only valid for a certain
+time range.
+
+The main distinction between these types of files is the way they are applied to reprocessing. Universal calibration files would generally
+use the most recent version for all time ranges and reprocessing. Time window calibration files should be defined only for a specific time range,
+and reprocessing should usually use the same file as the first processing.
+
+These two types of files are mostly used for creating the filename and processing conventions, so if your calibration doesn't neatly fit into
+either category, that's okay. The main thing is that all files should try and follow the conventions in this document.
+
+Specific examples
+^^^^^^^^^^^^^^^^^
+
+As an example, we will be looking at the Mag calibration files. These are supplemental files provided by the instrument team for use in
+L1 and L2 processing.
+
+The first file (L1B rotation matrices) is expected to change rarely,
+with the most recent version being used for all data. This file contains rotation calibration matrices from
+the boom to the instrument. These files would only change if there was a physical change to the boom or if there is
+some more accurate information available, so they are universal calibration files.
+
+The second file contains matrices which are applied to create the L2 and L2pre data. They are updated
+approximately monthly, and contain a start and end time for validity. During reprocessing, the data file with the most
+recent version and a time range that includes the data will be used. Since these are files which specifically cover a time range,
+they are time window calibration files.
+
+These two cases cover the two main types of calibration files. They both will be uploaded to the SDC
+with the API, and the correct file to use will be determined by the filename at processing time.
+
+File name format
+^^^^^^^^^^^^^^^^
+
+The file name for each file should be the following format::
+
+    <instrument>_<calibration_name>_<start_date>_<end_date>_<version>.<extension>
+
+Since these names are only used in instrument processing, they are flexible and can be adapted to
+accommodate different needs. This is more of a rule of thumb than a strict requirement.
+
+* ``<instrument>``: The instrument name associated with the calibration file
+* ``<calibration_name>``: A descriptive name for the calibration file which distinguishes between
+  other calibration files used by the instrument. Rather than "mag-calibration", this should be similar to
+  "mag-rotation-matrices" or "mag-l2pre-calibration." Multiple words should be split by hyphens.
+* ``<start_date>``: The start time of the validity of the calibration file, in the format "YYYYMMDD". This is required for
+  time window calibration files, and is optional for universal calibration files, or files that should be applied in all time
+  ranges where there isn't a file with a higher version available.
+* ``<end_date>``: The end time of the validity of the calibration file, in the format "YYYYMMDD". This is optional for files,
+  with the understanding that if end_date is not provided, the file is valid until a file with a later start_date and no end_date.
+* ``<version>``: The version of the file, with the format `v000`. This version should be incremented with EVERY new file
+  for a type of calibration file, even if the time span does not overlap. This is used to resolve conflicts if multiple files could be used.
+* ``<extension>``: The file extension. This is likely to be CDF, but could be any format that is convenient for the instrument team and SDC.
+
+File naming examples
+^^^^^^^^^^^^^^^^^^^^
+
+An L1B MAG rotation matrix calibration file:
+
+* ``mag_l1b-rotation-calibration-file_v001.cdf``
+
+This file is assumed to be valid for all times.
+
+An L1B MAG rotation matrix calibration file with a start date:
+
+* ``mag_l1b-rotation-calibration-file_20250101_v002.cdf``
+
+This file is only valid after 2025-01-01. If both of these files were uploaded, the second file would be used
+for all times after 2025-01-01, since the version is higher than ``mag_l1b-rotation-calibration-file_v001.cdf``. However,
+dates before 2025-01-01 would use the first file.
+
+Most global calibration files should be applied to all time ranges, but there may be cases where a file should
+only be used after a certain date - for example, if something physical occurred on the spacecraft on a date which changes
+all data after that date, but does not affect data before that date.
+
+An L2 MAG time window calibration file:
+
+* ``mag_l2-calibration_20250101_20250201_v001.cdf``
+
+This is valid between 2025-01-01 and 2025-02-01.
+
+* ``mag_l2-calibration_20250201_v002.cdf``
+
+This is valid starting at 2025-02-01, and is assumed to be valid until a new file without an end date is uploaded.
+
+If there is a file in the time range with a higher version, then that file should be used. Say we have
+an additional file uploaded:
+
+``mag_l2-calibration_20250301_20250302_v003.cdf``
+
+For all data between 2025-03-01 and 2025-03-02, this file would be used. However, for data after 2025-03-02,
+``mag_l2-calibration_20250201_v002.cdf`` would be used, because that is still the highest version file for that time range.
+
+This way, it is possible to upload corrections for smaller time ranges without affecting the long term calibration files.
+
+.. note::
+    The versions increment for all new `mag_l2-calibration` files, not just those with overlapping time spans.
+    This helps make it clear in every case which file is newer and should be used. This requirement should be
+    enforced by the SDC API.
+
+    Calibration files with different names can share versions, so the mag_l2-calibration and mag_l1b-rotation-calibration can have duplicate versions.
+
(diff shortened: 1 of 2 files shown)


docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
+.. _calibration_files:
+
+Calibration Files
+=================
+
+Calibration files are files used in processing that are provided by instrument teams
+to the SDC. These are files which contain values that could change over the course of
+the mission. Although they are called calibration files here, they may not contain calibration data,
+but instead any kind of data which is used in processing, but may infrequently change in time.
```
bourque: If we are defining the files this way, I wonder if it is better to call them "auxiliary files"?

Just putting it out there for discussion (I think I would still actually prefer to call them calibration files myself, mostly because spelling out "auxiliary" is a nightmare).

docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
...
+use the most recent version for all time ranges and reprocessing. Time window calibration files should be defined only for a specific time range,
+and reprocessing should usually use the same file as the first processing.
+
+These two types of files are mostly used for creating the filename and processing conventions, so if your calibration doesn't neatly fit into
+either category, that's okay. The main thing is that all files should try and follow the conventions in this document.
+
+Specific examples
+^^^^^^^^^^^^^^^^^
+
+As an example, we will be looking at the Mag calibration files. These are supplemental files provided by the instrument team for use in
```
bourque:
```suggestion
As an example, we will be looking at the MAG calibration files. These are supplemental files provided by the instrument team for use in
```

docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
...
+
+* ``<instrument>``: The instrument name associated with the calibration file
+* ``<calibration_name>``: A descriptive name for the calibration file which distinguishes between
+  other calibration files used by the instrument. Rather than "mag-calibration", this should be similar to
+  "mag-rotation-matrices" or "mag-l2pre-calibration." Multiple words should be split by hyphens.
+* ``<start_date>``: The start time of the validity of the calibration file, in the format "YYYYMMDD". This is required for
+  time window calibration files, and is optional for universal calibration files, or files that should be applied in all time
+  ranges where there isn't a file with a higher version available.
+* ``<end_date>``: The end time of the validity of the calibration file, in the format "YYYYMMDD". This is optional for files,
+  with the understanding that if end_date is not provided, the file is valid until a file with a later start_date and no end_date.
```
bourque: Would we ever expect multiple calibration files per-day of the same type? In other words, do we need to consider adding a `HHMMSS` timestamp here as well for finer granularity?

docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
...
+
+The file name for each file should be the following format::
+
+    <instrument>_<calibration_name>_<start_date>_<end_date>_<version>.<extension>
+
+Since these names are only used in instrument processing, they are flexible and can be adapted to
+accommodate different needs. This is more of a rule of thumb than a strict requirement.
+
+* ``<instrument>``: The instrument name associated with the calibration file
+* ``<calibration_name>``: A descriptive name for the calibration file which distinguishes between
```
bourque: I wonder if we should _require_ that the word `calibration` (or perhaps `cal` for short) be in the `calibration_name` so that it can be easily distinguished / searched on from our other data?

bourque (approved):
I like the naming convention, it is very consistent with the one we use for science data. I have a few questions for discussion, but I think this is a great draft!

docs/source/data-access-api/calibration-files.rst
maxinelasp: I agree that "calibration" is a little too specific for the files, but I think I've only ever heard them described as calibration files, regardless of if they are used for calibration. So I stuck with that use case.

docs/source/data-access-api/calibration-files.rst
maxinelasp: 🤷  I wasn't sure. I figure if anyone will care it will be MAG and I can ask them once we have finished SDC in team discussions on it and get that figured out quickly.

docs/source/data-access-api/calibration-files.rst
maxinelasp: That's an interesting proposal, although it would be counterintuitive for some types of files that should fall under this type of file naming/management system, such as Laura's lookup tables. I do like the idea of easily distinguishing them from other random files, though...

docs/source/data-access-api/calibration-files.rst
greglucas: "ancillary" is another term I've heard used before. But you've explicitly explained it here so this seems like it might be fine to be left as "calibration".

docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
...
+This document lays out the requirements for calibration files in the SDC and the expected file naming
+convention.
+
+File Naming Convention
+----------------------
+
+File overview
+^^^^^^^^^^^^^^
+
+In this document, I describe two main kinds of calibration files:
```
greglucas:
```suggestion
In this document, there are two main kinds of calibration files:
```

docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
...
+
+These two cases cover the two main types of calibration files. They both will be uploaded to the SDC
+with the API, and the correct file to use will be determined by the filename at processing time.
+
+File name format
+^^^^^^^^^^^^^^^^
+
+The file name for each file should be the following format::
+
+    <instrument>_<calibration_name>_<start_date>_<end_date>_<version>.<extension>
```
greglucas: Do we want to put `imap_` here as well to be closer to the science file format?

docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
...
+* ``<instrument>``: The instrument name associated with the calibration file
+* ``<calibration_name>``: A descriptive name for the calibration file which distinguishes between
+  other calibration files used by the instrument. Rather than "mag-calibration", this should be similar to
+  "mag-rotation-matrices" or "mag-l2pre-calibration." Multiple words should be split by hyphens.
+* ``<start_date>``: The start time of the validity of the calibration file, in the format "YYYYMMDD". This is required for
+  time window calibration files, and is optional for universal calibration files, or files that should be applied in all time
+  ranges where there isn't a file with a higher version available.
+* ``<end_date>``: The end time of the validity of the calibration file, in the format "YYYYMMDD". This is optional for files,
+  with the understanding that if end_date is not provided, the file is valid until a file with a later start_date and no end_date.
+* ``<version>``: The version of the file, with the format `v000`. This version should be incremented with EVERY new file
```
greglucas, line 79: I'm not sure I follow this sentence. I'm reading it as this is an always incrementing number with every single file? If I want to apply a daily calibration file, I would have to use `*_20240101_v001.txt` then the next day would be `_20240102_v002.txt` even though they don't overlap?

docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Most global calibration files should be applied to all time ranges, but there may be cases where a file should
+only be used after a certain date - for example, if something physical occurred on the spacecraft on a date which changes
+all data after that date, but does not affect data before that date.
+
+An L2 MAG time window calibration file:
+
+* ``mag_l2-calibration_20250101_20250201_v001.cdf``
+
+This is valid between 2025-01-01 and 2025-02-01.
```
greglucas: Could you add inclusive / exclusive for these dates here to indicate whether 2025-02-01 is a valid day in this calibration file or not.

greglucas (commented):
In this scheme we will have a lot of calibration files hanging around that the processing may have to think about which versions/date ranges to get. I'm wondering if we want to think about combining these individual calibration files together when a new file lands at the SDC into a single "latest_v123" calibration file that the processing can just use automatically that spans the entire time range without having to piece together the individual cal files. Moving the stitching of cal files from the processing routines to the SDC ingest routines.

docs/source/data-access-api/calibration-files.rst
```diff
@@ -0,0 +1,146 @@
...
+Any values which could be changed by the instrument team should be in a calibration file.
+
+These files will be tracked in the SDC and uploaded to the `SPDF archive <https://spdf.gsfc.nasa.gov/>`_.
+Therefore, they should use unique filenames which clearly distinguish them from each other. Beyond that, they
+can follow any format that is convenient for the instrument team and the SDC.
+
+This document lays out the requirements for calibration files in the SDC and the expected file naming
+convention.
+
+File Naming Convention
```
tech3371: can we remove this and rename below subsection title `File name format` to be `File Naming Convention`? On the  page, for some reason, it doesn't feel right. I thought that it will be about filename convention but it has information about calibration and noticed it has file name convention subsections.

tech3371 (approved):
just minor request. otherwise it looks good.

docs/source/data-access-api/calibration-files.rst
maxinelasp: Yes, for a given calibration file, the version would need to increment. In your example, if someone later uploaded a file that spanned across `20240101`, the version number would make it explicit that file is later, without needing the uploader to keep track of what time spans exist and what their version numbers are. I'm definitely open to making this a suggestion and not a requirement, but at the very least for MAG, where they will be having variable time span files and need a lot of control over reprocessing numbers etc it's the best I could come up with for always resolving weird edge cases.

maxinelasp: "the uploader" referring there to the people uploading the calibration files

docs/source/data-access-api/calibration-files.rst
maxinelasp: Oh yeah good call!

maxinelasp:
>  I'm wondering if we want to think about combining these individual calibration files together when a new file lands at the SDC into a single "latest_v123" calibration file that the processing can just use automatically that spans the entire time range without having to piece together the individual cal files. Moving the stitching of cal files from the processing routines to the SDC ingest routines.

I really like this idea! Definitely the backend will need to use some kind of database/shared file/helper lambda or something to get the files, since that will be a common piece of work that we shouldn't re-implement all over the place. Partly why I'd like most instruments to use a similar design even if it's overkill in some cases

docs/source/data-access-api/calibration-files.rst
greglucas: :+1: I would only worry that 999 might not be enough numbers in that case. I could definitely see people wanting to upload a batch of daily files or something like that, or even if it is monthly files, but then we want to update those monthly files later we now double the number of versions we are using. These are details though that can pretty easily be worked around.

maxinelasp: Yeah, that's a good point, something to keep in mind for sure.

docs/source/data-access-api/calibration-files.rst
vmartinez-cu: I've used "ancillary" at my previous job for supporting data and documentation files. That was for ocean data. Is there a convention in this industry?

docs/source/data-access-api/calibration-files.rst
vmartinez-cu: I like the idea too. If not in the name, is there a database field where file type could be tracked to easily query these files?

vmartinez-cu (approved):
Well written and easy to follow. Nice job on the examples provided for further clarification.

docs/source/data-access-api/calibration-files.rst
maxinelasp: Yeah, it almost certainly would fall under the indexer lambda, which currently adds a line in a database for science files to track them. Presumably we would track these the same way. Also, they will be stored in a different location in S3, so at least from the SDC side it will be easy to distinguish them from everything else. But, if you just have a file in isolation, for example in an archive, it wouldn't be clear... maybe we can just require pre-approval from an SDC member of the names to make sure they're fully descriptive? Thoughts @bourque ?

bourque: I was thinking about this from the perspective of a user who just has a directory with a whole bunch of various files from an instrument in it. I reckon it would be nice for this hypothetical user to be able to do something like `ls *cal*` or something along those lines. I'm not sure how important or relevant this would be in practice though.

maxinelasp: 🤔 Interesting idea. I do think that should be up to the instrument teams, if they put "cal" in their names to distinguish them then that's valid... but I'm very on the fence about it, I'm happy to call me outvoted if you and Veronica think it's useful!

vmartinez-cu: How is data delivered to users? If it's all placed in one directory, I think there is some value to distinguishing files via their names. When delivering ocean acoustic data to the public in my last job, calibration and ancillary data were organized in their own directories since the user community needed this level of granularity. Do we know what the user community needs here? Perhaps the instrument teams can provide some insight?

maxinelasp: That's a good way of looking at it, I think. We'll deliver all files through the API so we have a lot of control over what data delivery looks like to end users. In the backend, they will also be separated in different directories. I think if we followed the existing patterns, users would find ancillary files by making a call like `imap-data-access --instrument mag --ancillary` which would only return ancillary files

PR #369, opened 2024-03-21
IMAP-Hi L1A DE after Discussion with Paul

### LINKED ISSUE #190: Unpack and process IMAP-Hi packets to L1A

tech3371:
# Change Summary

## Overview
- Update DE with few time changes and addition based on discussion with Paul
- Add logic to handle if file starts without metaevent

### DIFF
--- a/imap_processing/hi/l1a/science_direct_event.py
+++ b/imap_processing/hi/l1a/science_direct_event.py
@@ -14,40 +14,24 @@ from imap_processing.hi import hi_cdf_attrs
 # fast the time was ticking. It is in microseconds.
 LOOKED_UP_DURATION_OF_TICK = 3999
 
+SECOND_TO_NS = 1e9
+MILLISECOND_TO_NS = 1e6
+MICROSECOND_TO_NS = 1e3
 
-def get_direct_event_time(met_seconds: int, met_subseconds: int, de_tag: int):
+
+def get_direct_event_time(time_in_ns):
     """Create MET(Mission Elapsed Time) time using input times.
 
     Parameters
     ----------
-    met_seconds : int
-        MET time in seconds
-    met_subseconds : int
-        MET subseconds in milliseconds
-    de_tag : int
-        Direct event time tag in ticks
+    time_in_ns : int
+        Time in nanoseconds
 
     Returns
     -------
     met_datetime : numpy.datetime64
         Human-readable MET time
     """
-    # Combine these direct event times to DE MET time:
-    #   seconds (from metaevent)
-    #   subseconds (milliseconds) (from metaevent)
-    #   de_tag (milliseconds) (from direct event)
-    #   LOOKED_UP_DURATION_OF_TICK (milliseconds)
-
-    seconds_to_ns = 1e9
-    millisecond_to_ns = 1e6
-    microsecond_to_ns = 1e3
-
-    time_in_ns = (
-        met_seconds * seconds_to_ns
-        + met_subseconds * millisecond_to_ns
-        + de_tag * LOOKED_UP_DURATION_OF_TICK * microsecond_to_ns
-    )
-
     met_datetime = launch_time + np.timedelta64(int(time_in_ns), "ns")
 
     return met_datetime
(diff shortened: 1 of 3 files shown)


bourque (approved):
LGTM!

imap_processing/hi/l1a/science_direct_event.py
```diff
@@ -166,52 +161,80 @@ def break_into_bits_size(binary_data: str):
     return [binary_data[i : i + 48] for i in range(0, len(binary_data), 48)]
```
sdhoyt: I think it's helpful to assign numbers to variables so it's more clear what they are. Like here if you could set 48 to something like field_bit_length (?). Same with conversions like above 1e9 could be assigned to sec_to_ns or something like that to make it more clear what's happening.

imap_processing/hi/l1a/science_direct_event.py
```diff
@@ -166,52 +161,80 @@ def break_into_bits_size(binary_data: str):
...
-        # metaevent is a way to store information
-        # about bigger portion of time information. Eg.
-        # metaevent stores information about, let's say
-        # "20240319T09:30:01". Then direct event time
-        # tag stores information of time ticks since
-        # that time. Then we use those two to combine and
-        # get exact time information of each event.
+    ccsds_met = []
+
+    # How to handle if first event is not metaevent? This
```
sdhoyt, line 194: should each direct event always have a metaevent associated with it? Or is it only an issue if the first event in a packet is missing an metaevent?

imap_processing/hi/l1a/science_direct_event.py
```diff
@@ -166,52 +161,80 @@ def break_into_bits_size(binary_data: str):
...
+    if first_metaevent_index is None:
+        return None
+    elif first_metaevent_index != 0:
+        # Discard all direct events until we see next metaevent
+        # TODO: log a warning
+        de_data_list = de_data_list[first_metaevent_index:]
+        packet_met_time = packet_met_time[first_metaevent_index:]
+
+    for index, event in enumerate(de_data_list):
         if event["start_bitmask_data"] == 0:
```
sdhoyt, line 215: Is a start_bit_mask_data = 0 an indication that the current item in the list is a metaevent? I might be misinterpreting this, but it looks like the metaevents are separate dictionaries in the list? If that's the case and there should be a metaevent associated with each event, would it be easier to merge the metaevent and event into one dictionary?

imap_processing/hi/l1a/science_direct_event.py
```diff
@@ -166,52 +161,80 @@ def break_into_bits_size(binary_data: str):
...
+
             # set time and esa step values to
             # be used for direct event followed by
             # this metaevent
             int_subseconds = event["subseconds"]
             int_seconds = event["seconds"]
             current_esa_step = event["esa_step"]
+            # Add half a tick once per algorithm document
+            # and Paul Janzen.
+            haf_tick = LOOKED_UP_DURATION_OF_TICK // 2
```
sdhoyt: "half" typo?

sdhoyt (commented):
Looks good! Sorry about leaving some comments (mostly questions) about some lines of code that I think you updated in the last PR. I didn't have time to finish reviewing the last one.

tech3371:
> Looks good! Sorry about leaving some comments (mostly questions) about some lines of code that I think you updated in the last PR. I didn't have time to finish reviewing the last one.

@sdhoyt Thank you for leaving comments from last PR. I still need to improve my communication with PR.

imap_processing/hi/l1a/science_direct_event.py
tech3371: Yes. That's a great point. Will make those changes.

imap_processing/hi/l1a/science_direct_event.py
tech3371: It's only an issue with first event in the first packet of the data file.

sdhoyt:
@tech3371 Don't worry about it! I need to improve my punctuality on reviewing PRs

imap_processing/hi/l1a/science_direct_event.py
tech3371: Yes. start_bit_mask_data = 0 indicates that it's a metaevent. metaevent and direct event are both 48-bits and but the first two bits indicates which type it is. Based on type, it needs to parse 48-bits differently. I had to keep them in the same list because order is important here. Every direct event will start with a metaevent and those direct events uses time information from the metaevent to build their full time information.

imap_processing/hi/l1a/science_direct_event.py
tech3371: yes. good catch!

tech3371:
I am going to merge this in the afternoon. Please don't hesitate to comment now or after merging.

imap_processing/hi/l1a/science_direct_event.py
```diff
@@ -163,55 +175,87 @@ def break_into_bits_size(binary_data: str):
...
             int_subseconds = event["subseconds"]
             int_seconds = event["seconds"]
             current_esa_step = event["esa_step"]
+            # Add half a tick once per algorithm document
+            # and Paul Janzen.
+            half_tick = LOOKED_UP_DURATION_OF_TICK // 2
+            # convert microseconds to millieseconds to
+            # match subseconds time format
+            half_tick_ms = half_tick // 1e3
+            int_subseconds += half_tick_ms
```
greglucas: Could you add a link to the page in the algorithm document for this? Might be easier to track it down in the future.

Do you know what values you expect this to be, it doesn't seem quite right to be dividing `3999` by // 2 into an integer `1999`, do you want that to be `1999.5` instead? Then the next line produces a float half_tick_ms because of 1e3, so you might want to switch that over to 1000 instead. But also note that it is producing 1 ms, from an initial value of 3999 which is smaller than half a tick.

```
>>> 3999 // 2 // 1000
1
>>> 3999 // 2 // 1e3
1.0
```

tech3371: I will add the section from algorithm document. In the document it said like in the code block and Paul suggested it doing that. I will think about this and double check with him.
```
Take the MET(in code, it's direct event MET time), add half a tick
```
The reason he wants to add half a tick is because event could be in between ticks and he wanted to add this to account for that.

greglucas: Currently, it is closer to 1/4 of a tick if it is returning 1, so it'd be worth checking with him.

Another option is to change to `int_ns` here so you go to nanoseconds instead of milliseconds, or make milliseconds a float value instead of integer. Because in the following routine you turn the time into nanoseconds since MET.

tech3371: After looking through it closely. I made a big mistake with calculation. Looks like I was converting it ns even though my comment said I was converting it millisecond. Thank you for catching it. Will push change soon and if you could double check, I appreciate it.

greglucas (approved):
I think it looks right now. It might be a good idea to add a few explicit tests for these functions to make sure they are doing what you expect. Ignoring anything with the packets and just focusing on the lists you are expecting and manually manipulating them.

```python
meta_event = ...
direct_event = {"seconds": 5, "subseconds": 2, "esa_step": 1, ...}
list_with_meta_first = [meta_event, direct_event, direct_event]
list_with_meta_middle = [direct_event, meta_event, direct_event]
list_with_meta_last = [direct_event, direct_event, meta_event]
...
create_dataset(list_with_*, [time])
```

PR #370, opened 2024-03-25
Updating mag CDF generation to match new filenames

maxinelasp:
# Change Summary
Pretty minor updates to match changes made to cdf generation

## Overview

Updated decom_mag and cli to use new CDF generation updates

## Updated Files

- decom_mag
   - Updating filenames to match filename updates, fixing variable names, fixing types
- cli.py & mag_l1a.py
   - Changing to use new write_cdf method signature 
- mag_attrs
   - updating to put info in attrs needed for generation, updating based on some comments about CDF files 

## Testing

Required some updates in tests

### DIFF
--- a/imap_processing/mag/l1a/mag_l1a.py
+++ b/imap_processing/mag/l1a/mag_l1a.py
@@ -8,7 +8,7 @@ from imap_processing.mag.l0 import decom_mag
 logger = logging.getLogger(__name__)
 
 
-def mag_l1a(packet_filepath, output_filepath_norm, ouptput_filepath_burst):
+def mag_l1a(packet_filepath):
     """
     Process MAG L0 data into L1A CDF files at cdf_filepath.
 
@@ -16,19 +16,22 @@ def mag_l1a(packet_filepath, output_filepath_norm, ouptput_filepath_burst):
     ----------
     packet_filepath :
         Packet files for processing
-    output_filepath_norm :
-        Full directory and filename for raw-norm CDF file
-    ouptput_filepath_burst :
-        Full directory and filename for raw-burst CDF file
     """
     mag_l0 = decom_mag.decom_packets(packet_filepath)
 
     mag_norm, mag_burst = decom_mag.export_to_xarray(mag_l0)
 
     if mag_norm is not None:
-        write_cdf(mag_norm, Path(output_filepath_norm))
-        logger.info(f"Created CDF file at {output_filepath_norm}")
+        file = write_cdf(mag_norm)
+        logger.info(f"Created CDF file at {file}")
 
     if mag_burst is not None:
-        write_cdf(mag_burst, Path(ouptput_filepath_burst))
-        logger.info(f"Created CDF file at {output_filepath_norm}")
+        file = write_cdf(mag_burst)
+        logger.info(f"Created CDF file at {file}")
+
+
+if __name__ == "__main__":
+    filepath_burst = Path("mag_IT_data/MAG_SCI_BURST.bin")
+    filepath_norm = Path("mag_IT_data/MAG_SCI_NORM.bin")
+    mag_l1a(filepath_norm)
+    mag_l1a(filepath_burst)
--- a/imap_processing/mag/l1a/mag_l1a.py
+++ b/imap_processing/mag/l1a/mag_l1a.py
@@ -1,6 +1,5 @@
 """Methods for decomming packets, processing to level 1A, and writing CDFs for MAG."""
 import logging
-from pathlib import Path
 
 from imap_processing.cdf.utils import write_cdf
 from imap_processing.mag.l0 import decom_mag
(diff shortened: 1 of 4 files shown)


imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -8,27 +8,30 @@
...
+        logger.info(f"Created CDF file at {file}")
 
     if mag_burst is not None:
-        write_cdf(mag_burst, Path(ouptput_filepath_burst))
-        logger.info(f"Created CDF file at {output_filepath_norm}")
+        file = write_cdf(mag_burst)
+        logger.info(f"Created CDF file at {file}")
+
+
+if __name__ == "__main__":
```
tech3371: Is this for local testing?

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -25,16 +25,23 @@
     "Magnetic Fields (space)",
 )
 
-mag_l1a_attrs = GlobalDataLevelAttrs(
-    # TODO: data type should include "norm" and "burst" L1A-norm>Level-1A-normal-rate
-    "L1A>Level-1A",
+mag_l1a_norm_raw_attrs = GlobalDataLevelAttrs(
+    "L1A-raw-norm>Level-1A-raw-normal-rate",
```
tech3371, line 29: Here, Andriy said that this data_type should be datalevel_descriptor. I think it's ok to have `-` in descriptor. This is what I think he may want. @bryan-harter, what do you think?
```suggestion
    "L1A-raw-norm>l1a_raw-normal-rate",
```

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -25,16 +25,23 @@
...
     # Should also include data type
-    # TODO: replace "sci" with descriptor "norm" / "burst"
-    logical_source="imap_mag_l1a_sci",
+    logical_source="imap_mag_l1a_norm-raw",
     logical_source_desc="IMAP Mission MAG Instrument Level-1A Data.",
     instrument_base=mag_base,
 )
 
+mag_l1a_burst_raw_attrs = GlobalDataLevelAttrs(
+    "L1A-raw-burst>Level-1A-raw-burst-rate",
```
tech3371, line 37: same as above

tech3371 (commented):
couple comments but looks great!

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: yup, should not be in here. Thanks for catching!

imap_processing/mag/mag_cdf_attrs.py
maxinelasp: I think this was a direct note from an email he sent me about descriptors - but I'd appreciate Bryan's thoughts

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -77,10 +77,11 @@ def test_mag_raw_cdf_generation(tmp_path):
     input_xarray = cdf_to_xarray(output)
     assert input_xarray.attrs.keys() == norm_data.attrs.keys()
 
-    test_data_path_burst = tmp_path / "mag_l1a_raw-burst_20210101_20210102_v01-01.cdf"
+    test_data_path_burst = output_path / "imap_mag_l1a_burst-raw_20231025_v001.cdf"
 
     assert not test_data_path_burst.exists()
     output = write_cdf(burst_data, test_data_path_burst)
+    print(output)
```
bourque:
```suggestion
```

bourque (approved):
LGTM!

imap_processing/mag/mag_cdf_attrs.py
tech3371: Ah..interesting. My notes said differently. Good catch.

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -60,15 +59,16 @@ def test_mag_raw_xarray():
...
 
+    output_path = tmp_path / "imap" / "mag" / "l1a" / "2023" / "10"
+
     norm_data, burst_data = export_to_xarray(l0)
 
-    test_data_path_norm = tmp_path / "mag_l1a_raw-normal_20210101_20210102_v01-01.cdf"
+    test_data_path_norm = output_path / "imap_mag_l1a_norm-raw_20231025_v001.cdf"
 
     assert not test_data_path_norm.exists()
     output = write_cdf(norm_data, test_data_path_norm)
```
greglucas: You should check this, but I was thinking that this might actually not be testing what we think it is here. I think it might be creating a path with that filename as a part of the directory structure `imap_mag_l1a_norm-raw_20231025_v001.cdf/imap/mag/...`

In the other tests I just removed all of the `tmp_path` fixtures and asserted based on t he filename from `output.name == expected_name`, which might simplify this a bit. Then we can rely on the `write_cdf()` tests for the specific directory structure expectations.

maxinelasp: Ok, I think I understand what you mean. I'll update this now.

imap_processing/mag/mag_cdf_attrs.py
bryan-harter: Sorry, better late than never? I do go back and look at reviews eventually lol.  

I'm pretty sure it can just be "L1A-raw-norm>Level-1A Raw Normal Rate".  No need for hyphens or underscores on the other side of the ">".

maxinelasp: Thanks! I'll update in my next mag PR!

PR #371, opened 2024-03-25
Process IMAP-Hi Housekeeping data and create CDF file

### LINKED ISSUE #190: Unpack and process IMAP-Hi packets to L1A

### LINKED ISSUE #191: Write IMAP-Hi L1A data to CDF file

tech3371:
# Change Summary

## Overview
- unpacks housekeeping data
- test that we are able to write data to CDF file

### DIFF
--- a/imap_processing/hi/hi_cdf_attrs.py
+++ b/imap_processing/hi/hi_cdf_attrs.py
@@ -137,8 +137,26 @@ ccsds_met_attrs = ScienceAttrs(
 # descriptor: instrument name - This is in global attributes
 # data_type: <data_level>_<descriptor> - this is here in DataLevelAttrs
 hi_de_l1a_attrs = GlobalDataLevelAttrs(
-    data_type="L1A>l1a_de",
+    data_type="L1A-de>Level-1A Direct Event",
     logical_source="imap_hi_l1a_de",
     logical_source_desc=("IMAP-HI Instrument Level-1A Direct Event Data."),
     instrument_base=hi_base,
+)
+
+hi_hk_l1a_attrs = GlobalDataLevelAttrs(
+    data_type="L1A-hk>Level-1A Housekeeping",
+    logical_source="imap_hi_l1a_hk",
+    logical_source_desc=("IMAP-HI Instrument Level-1A Housekeeping Data."),
+    instrument_base=hi_base,
+)
+
+hi_hk_l1a_metadata_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=GlobalConstants.INT_MAXVAL,
+    display_type="no_plot",
+    depend_0="epoch",
+    format="I12",
+    units="int",
+    var_type="support_data",
+    variable_purpose="PRIMARY",
 )
--- a/imap_processing/hi/hi_cdf_attrs.py
+++ b/imap_processing/hi/hi_cdf_attrs.py
@@ -153,7 +153,6 @@ hi_hk_l1a_attrs = GlobalDataLevelAttrs(
 hi_hk_l1a_metadata_attrs = ScienceAttrs(
     validmin=0,
     validmax=GlobalConstants.INT_MAXVAL,
-    display_type="no_plot",
     depend_0="epoch",
     format="I12",
     units="int",
--- a/imap_processing/hi/l1a/hi_l1a.py
+++ b/imap_processing/hi/l1a/hi_l1a.py
@@ -2,9 +2,12 @@
 import logging
 
 from imap_processing.hi.l0 import decom_hi
+from imap_processing.hi.l1a.housekeeping import create_dataset
 from imap_processing.hi.l1a.science_direct_event import science_direct_event
 from imap_processing.hi.utils import HIAPID
 from imap_processing.utils import group_by_apid
+
+logger = logging.getLogger(__name__)
 
 
 def hi_l1a(packet_file_path: str):
@@ -32,14 +35,17 @@ def hi_l1a(packet_file_path: str):
             # TODO: Add processing for science count data
             continue
         elif apid == HIAPID.H45_SCI_DE:
-            logging.debug(
+            logger.info(
                 "Processing direct event data for [%s] packets", HIAPID.H45_SCI_CNT.name
             )
 
             data = science_direct_event(grouped_data[apid])
             processed_data.append(data)
         elif apid == HIAPID.H45_APP_NHK:
-            # TODO: Add processing for housekeeping data
-            continue
+            logger.info(
+                "Processing housekeeping data for [%s] packets", HIAPID.H45_APP_NHK.name
+            )
+            data = create_dataset(grouped_data[apid])
+            processed_data.append(data)
 
     return processed_data
--- a/imap_processing/hi/l1a/housekeeping.py
+++ b/imap_processing/hi/l1a/housekeeping.py
@@ -1,0 +1,80 @@
+"""Unpack IMAP-Hi housekeeping data."""
+import collections
+import dataclasses
+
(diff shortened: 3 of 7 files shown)


pyproject.toml
```diff
@@ -33,7 +33,7 @@ bitstring = ">=4.0.1"
 cdflib = "==1.2.6"
 imap-data-access = ">=0.5.0"
 python = ">=3.9,<4"
-space_packet_parser = ">=4.1.0"
+space_packet_parser = ">=4.2.0rc3"
```
tech3371, line 36: TODO: update this to new version

imap_processing/hi/l1a/hi_l1a.py
```diff
@@ -32,14 +33,17 @@ def hi_l1a(packet_file_path: str):
             # TODO: Add processing for science count data
             continue
         elif apid == HIAPID.H45_SCI_DE:
-            logging.debug(
+            logging.info(
```
greglucas: We should get a logger in this module to follow the other instruments.
```suggestion
            logger.info(
```

imap_processing/hi/l1a/housekeeping.py
```diff
@@ -0,0 +1,92 @@
...
+    packets : list
+        packet list
+
+    Returns
+    -------
+    xr.dataset
+        dataset with all metadata field data in xr.DataArray
+    """
+    metadata_arrays = collections.defaultdict(list)
+    description_arrays = collections.defaultdict(list)
```
greglucas: I don't think you're using this as a list later but just a standard dictionary mapping.
```suggestion
    description_arrays = {}
```

imap_processing/hi/l1a/housekeeping.py
```diff
@@ -0,0 +1,92 @@
...
+
+    for data_packet in packets:
+        # Add metadata to array
+        for key, value in data_packet.header.items():
```
greglucas: You can avoid the two for loops by combining the two dictionaries together I think.
```suggestion
        for key, value in (data_packet.header | data_packet.data).items():
```

imap_processing/hi/l1a/housekeeping.py
```diff
@@ -0,0 +1,92 @@
...
+            # convert key to lower case to match SPDF requirement
+            data_key = key.lower()
+            metadata_arrays.setdefault(data_key, []).append(value.raw_value)
+            # add it once since description should be same for all packets
+            if data_key not in description_arrays:
+                description_arrays[data_key] = (
+                    value.long_description
+                    if value.long_description
+                    else value.short_description
+                )
```
greglucas: Might as well just overwrite it then for simplicity?
```suggestion
            description_arrays[data_key] = value.long_description or value.short_description
```
You can also use the `value or other_value` to avoid the if/else if you want to simplify even further.

imap_processing/hi/l1a/housekeeping.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    metadata_arrays = collections.defaultdict(list)
+    description_arrays = collections.defaultdict(list)
+
+    for data_packet in packets:
+        # Add metadata to array
+        for key, value in data_packet.header.items():
+            # convert key to lower case to match SPDF requirement
+            data_key = key.lower()
+            metadata_arrays.setdefault(data_key, []).append(value.raw_value)
```
greglucas: You set this as a `defaultdict[list]` so you should be able to append directly to your item and it will take care of the setdefault for you.
```suggestion
            metadata_arrays[data_key].append(value.raw_value)
```

imap_processing/hi/l1a/housekeeping.py
tech3371: good catch!

imap_processing/hi/l1a/housekeeping.py
tech3371: Thank you! Great point.

imap_processing/hi/l1a/housekeeping.py
tech3371: cool. I wasn't sure which way it was prefer. Will this or pick short description if long description was None?

greglucas: > Will this or pick short description if long description was None?

Yes, it will check `value.long_description` and then fall back to `value.short_description` if `long_description` is None.

```bash
>>> None or "abc"
'abc'
```

tech3371: Just tested and it does get one that's not None.

imap_processing/hi/l1a/housekeeping.py
```diff
@@ -0,0 +1,79 @@
...
+            data_key = key.lower()
+            metadata_arrays[data_key].append(value.raw_value)
+            # description should be same for all packets
+            description_dict[data_key] = (
+                value.long_description or value.short_description
+            )
+
+    epoch_converted_time = [
+        calc_start_time(sc_time) for sc_time in metadata_arrays["ccsds_met"]
+    ]
```
greglucas, line 45: Not for this PR, but it might be nice if `calc_start_time` could handle an array and you could instead call:
`epoch_converted_times = calc_start_time(metadata_arrays["ccsds_met"])`

greglucas (approved):
Nice, the for loops are much easier to follow now I think.

imap_processing/hi/l1a/housekeeping.py
tech3371: I agree. Many of us are doing this. Would be nice to call a common util function for this.

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -142,3 +143,21 @@
     logical_source_desc=("IMAP-HI Instrument Level-1A Direct Event Data."),
     instrument_base=hi_base,
 )
+
+hi_hk_l1a_attrs = GlobalDataLevelAttrs(
+    data_type="L1A>l1a_hk",
```
bryan-harter: I believe this should be something like "L1a-hk>Level-1A Housekeeping Data". To the right of the ">" is the full written-out description of the data type

imap_processing/hi/l1a/housekeeping.py
```diff
@@ -0,0 +1,79 @@
...
+
+from imap_processing.cdf.global_attrs import ConstantCoordinates
+from imap_processing.cdf.utils import calc_start_time
+from imap_processing.hi.hi_cdf_attrs import (
+    hi_hk_l1a_attrs,
+    hi_hk_l1a_metadata_attrs,
+)
+
+
+def create_dataset(packets):
```
bryan-harter: Sorry if this has been discussed already, but how do we feel about type hints? I find them useful if for no other reason than when I step away from IMAP for awhile I need to remind myself what the main objects are that we're all using

```suggestion
from typing import List
from space_packet_parser.parser import Packet

def create_dataset(packets: List[Packet]) -> xr.Dataset:
```

imap_processing/hi/l1a/housekeeping.py
```diff
@@ -0,0 +1,79 @@
...
+    """Create dataset for each metadata field.
+
+    Parameters
+    ----------
+    packets : list
```
bryan-harter: Or maybe just add something here: 

```suggestion
    packets : list[Packets]
```

imap_processing/hi/l1a/housekeeping.py
tech3371: I believed it's not been discussed. Cool. I will make that change.

tech3371:
pre-commit.ci autofix

PR #374, opened 2024-03-27
Update IMAP-Hi DE CDF attrs

### LINKED ISSUE #191: Write IMAP-Hi L1A data to CDF file

tech3371:
# Change Summary

## Overview
Minor changes to IMAP-Hi DE attrs to be able to create CDF file.


imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -45,9 +45,12 @@
 )
 
 # Direct event attrs
-esa_step_attrs = AttrBase(
+esa_step_attrs = ScienceAttrs(
     validmin=0,
     validmax=10,
+    format="I2",
+    units="int",
```
```suggestion
    units=" ",
```
I think this should be blank here since it is unitless. Or possibly `"#"` to indicate a number?

https://spdf.gsfc.nasa.gov/istp_guide/vattributes.html

> UNITS --- required if not using UNIT_PTR (optional for time variables)
is a character string (no more than 20 characters, but preferably 6 characters) representing the units of the variable, e.g., nT for magnetic field. If the standard abbreviation used is short then the units value can be added to a data listing heading or plot label. Use a blank character, rather than "None" or "unitless", for variables that have no units (e.g., a ratio or a direction cosine).
For CDF_TIME_TT2000: SI measurement unit: s, ms(milliseconds for EPOCH variables), ns(nanoseconds for CDF_TIME_TT2000), ps(picoseconds for EPOCH16)

tech3371: looks like I can avoid setting that parameter and our default for that is empty string. I will remove those units parameter. I thought it's required. My bad.

bryan-harter (approved):

PR #375, opened 2024-03-27
MNT: Change default display_type to be no_plot

greglucas:
# Change Summary

ISTP compliance requires a string for display_type, so default to `no_plot` and then people can override it later if they want. But this at least gives an ISTP compliant default.

Remove the skips from IDEX now that it is compliant


vmartinez-cu (approved):
Looks good to me

bourque (approved):

PR #377, opened 2024-03-28
TST: Move codice xfail into a separate test

greglucas:
# Change Summary

## Overview

Explicitly test that the function raises what we think rather than xfailing.

### DIFF
--- a/imap_processing/tests/codice/test_decompress.py
+++ b/imap_processing/tests/codice/test_decompress.py
@@ -43,5 +43,5 @@ def test_decompress(compressed_value: int, algorithm: IntEnum, expected_result: 
 def test_decompress_raises():
     """Tests that the ``decompress`` function raises with an unknown algorithm"""
 
-    with pytest.raises(ValueError, match="some_unsupported_algorithm is not"):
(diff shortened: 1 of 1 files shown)


imap_processing/tests/codice/test_decompress.py
```diff
@@ -39,3 +38,10 @@ def test_decompress(compressed_value: int, algorithm: IntEnum, expected_result:
 
     decompressed_value = decompress(compressed_value, algorithm)
     assert decompressed_value == expected_result
+
+
+def test_decompress_raises():
+    """Tests that the ``decompress`` function raises with an unknown algorithm"""
+
+    with pytest.raises(ValueError, match="some_unsupported_algorithm is not"):
```
bourque: Wasn't sure if this was here on purpose

```suggestion
    with pytest.raises(ValueError, match="some_unsupported_algorithm"):
```

bourque (approved):
Thanks for doing this!

imap_processing/tests/codice/test_decompress.py
greglucas: I did actually do it on purpose but it does read funny when it abruptly ends like that, so good suggestion!

PR #379, opened 2024-03-28
MNT: Remove directory option from write_cdf

greglucas:
# Change Summary

## Overview

We are now explicitly requiring a directory structure for the science files to be written to. A user should update the `imap_data_access.config["DATA_DIR"]` location if they want to write to a different location because the `ScienceFilePath` uses that field to construct the path.

Note that previously `directory` was not working how we were expecting because of the `ScienceFilePath.construct_path()` usage, so this doesn't change behavior, just removes code that didn't do what it was supposed to.

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -1,8 +1,6 @@
 """Various utility functions to support creation of CDF files."""
 
 import logging
-from pathlib import Path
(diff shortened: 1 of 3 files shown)


tech3371 (approved):

maxinelasp (approved):
LGTM

PR #382, opened 2024-03-28
MNT: Add codespell and fix spelling mistakes found

greglucas:
# Change Summary

## Overview

Add codespell as a pre-commit hook. Only apply to
python, rest, and markdown files for now. There are some spelling mistakes in the XML files, but that is from external sources, so lets ignore those for now.

Fix the spelling errors it found.


bryan-harter:
Hey neat!

bourque (approved):
Cool stuff. And with this we are one step closer to not having to think about the code we write! 😂

PR #384, opened 2024-04-01
Mag l1a processing

maxinelasp:
# Change Summary
Finishing MAG L1A processing, including creating vectors and timestamping them. This still needs some work on the timestamps (they're still in spacecraft time and should be in J2000) and some CDF generation work. 

## Overview

Data classes are in mag_l1a_data. L1A processing is in mag_l1a_data and mag_l1a. I refactored some earlier work to split up normal and burst mode upon packet read, as these packets are not mixed together again and it makes the code cleaner. And moved from bytearray to numpy array in early L1a processing. 

## New Dependencies

None

## New Files

- mag_l1a_data
   - dataclasses for L1A
   
## Updated Files

- mag_l1a
   - Added L1A processing
- mag_decom
   - Moved around some of the packet sorting steps and simplified processing 
   
## Testing

Added new tests for L1A, primarily using validation data provided by MAG. I plan to go back and add one more byte-processing test.


imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -99,9 +101,17 @@ def __post_init__(self):
         """
         if isinstance(self.VECTORS, str):
             # Convert string output from space_packet_parser to bytearray
```
greglucas:
```suggestion
            # Convert string output from space_packet_parser to an array of individual big endian bytes
```

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -99,9 +101,17 @@ def __post_init__(self):
...
-            self.VECTORS = bytearray(
-                int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big")
+            self.VECTORS = np.frombuffer(
+                int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"),
+                dtype=np.dtype(">b"),
             )
 
+        if isinstance(self.VECTORS, bytearray):
+            self.VECTORS = np.array(self.VECTORS, dtype=np.dtype(">b"))
```
greglucas: Will you be getting both options, string and bytearray input, or is this from the previous version and now we are only taking the string path?

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -16,14 +19,115 @@ def mag_l1a(packet_filepath):
...
+        if mag_l0.COMPRESSION:
+            raise NotImplementedError("Unable to process compressed data")
+
+        primary_start_time = TimeTuple(mag_l0.PRI_COARSETM, mag_l0.PRI_FNTM)
+        secondary_start_time = TimeTuple(mag_l0.SEC_COARSETM, mag_l0.SEC_FNTM)
+
+        # seconds of data in this packet is the SUBTYPE plus 1
+        seconds_per_packet = mag_l0.PUS_SSUBTYPE + 1
+
+        # now we know the number of secs of data in the packet, and the data rates of
```
greglucas: Nice comments, thank you for adding all of these!

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -16,14 +19,115 @@ def mag_l1a(packet_filepath):
...
+        # VECSEC is already decoded in mag_l0
+        total_primary_vectors = seconds_per_packet * mag_l0.PRI_VECSEC
+        total_secondary_vectors = seconds_per_packet * mag_l0.SEC_VECSEC
+
+        primary_vectors, secondary_vectors = MagL1a.process_vector_data(
+            mag_l0.VECTORS, total_primary_vectors, total_secondary_vectors
+        )
+
+        # Primary sensor is MAGO (most common expected case)
+        if mag_l0.PRI_SENS == 0:
```
greglucas, line 85: There is quite a bit of repetition between the two cases in setting up the classes. It isn't too bad as-is, but figured I would bring this up in case you think it might be easy to do something like the following so you only have two MagL1a calls with varying parameters calculated beforehand:
```python
primary = MagL1a(True, bool(mag_l0.MAGO_ACT if mag_l0.PRI_SENS else mag_l0.MAGI_ACT), ...
secondary = MagL1a(False, ...)

if mag_l0.PRI_SENS == 0:
    mago.append(primary)
    magi.append(secondary)
else:
    mago.append(secondary)
    magi.append(primary)
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -0,0 +1,253 @@
...
+    coarse_time: int
+    fine_time: int
+
+    def __add__(self, seconds: float):
+        """
+        Add a number of seconds to the time tuple.
+
+        Parameters
+        ----------
+        seconds : int
```
greglucas:
```suggestion
        seconds : float
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        sample_time_interval = 1 / self.vectors_per_second
+        previous_time = self.start_time
+        for index, vector in enumerate(self.vectors):
+            if index == 0:
+                new_vector = Vector(vector, self.start_time)
+            else:
+                new_vector = Vector(vector, previous_time + sample_time_interval)
+
+            previous_time = new_vector.timestamp
```
greglucas: I think this can be simplified a bit by updating the time at the end of your loop iteration.
```suggestion
        sample_time_interval = 1 / self.vectors_per_second
        current_time = self.start_time
        for index, vector in enumerate(self.vectors):
            new_vector = Vector(vector, current_time)
            # Increment our time counter before the next iteration
            current_time = new_vector.timestamp + sample_time_interval
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -0,0 +1,259 @@
...
+                y = (
+                    ((vector_data[pos + 2] & 0xFF) << 8)
+                    | ((vector_data[pos + 3] & 0xFF) << 0)
+                ) & 0xFFFF
+                z = (
+                    ((vector_data[pos + 4] & 0xFF) << 8)
+                    | ((vector_data[pos + 5] & 0xFF) << 0)
+                ) & 0xFFFF
+                rng = (vector_data[pos + 6] >> 6) & 0x3
+                pos += 6
```
greglucas, line 189: Again minor (and I realize you got this from elsewhere), but I would probably write this with variables being parameterized and then used in the lookups a single time to try and avoid the duplication of those lookups. No need to change here unless you like this form better.
```python
if i % 4 == 0:
    left_mask = 0xFF
    left_shift = 0
elif i % 4 == 1:
    left_mask = 0x3F
    left_shift = 2
...
x = ((vector_data[pos + 0] & left_mask) << (left_shift + 8))
        | (vector_data[pos + 1] & 0xFF) << left_shift))
       & 0xFFFF
```

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -5,13 +5,16 @@
...
-from imap_processing.mag.l0.decom_mag import decom_packets, export_to_xarray
+from imap_processing.mag import mag_cdf_attrs
+from imap_processing.mag.l0.decom_mag import decom_packets, generate_dataset
 
 
 def test_mag_decom():
     current_directory = Path(__file__).parent
     burst_test_file = current_directory / "mag_l0_test_data.pkts"
-    l0 = decom_packets(burst_test_file)
+    packets = decom_packets(str(burst_test_file))
```
greglucas, line 15: Does your decom_packets function require a string now, or could you keep that as Path?

greglucas (approved):
Very minor comments from me. Looks really nice overall!

imap_processing/mag/l0/mag_l0_data.py
maxinelasp: I had both cases happen during testing (I assume due to space_packet_parser) but let me double check this. It could be that I had both cases happen because I had already converted VECTORS to a bytearray or something.

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Yeah I thought about moving it, but I thought it was clearer this way what is actually happening. It's a little subtle where primary/secondary becomes mago/magi so I thought it would be better to be very explicit in the code.

imap_processing/tests/mag/test_mag_decom.py
maxinelasp: Good catch, copy+paste fail!

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Nice, that's way better

maxinelasp:
pre-commit.ci autofix

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -90,19 +94,24 @@ class MagL0:
     PRI_FNTM: int
     SEC_COARSETM: int
     SEC_FNTM: int
-    VECTORS: bytearray
+    VECTORS: np.ndarray | str
```
tech3371, line 98: Can you add in the doc string why this would be either array or string? This seems like unusual case.

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -90,19 +94,24 @@ class MagL0:
...
-            )
+        # Convert string output from space_packet_parser to numpy array of
+        # big-endian bytes
+        self.VECTORS = np.frombuffer(
+            int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"),
+            dtype=np.dtype(">b"),
+        )
+
+        # Remove buffer from end of vectors
+        if len(self.VECTORS) % 2:
```
tech3371, line 115: What does this `% 2` indicate?

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -17,14 +20,115 @@ def mag_l1a(packet_filepath):
...
+    burst_data = packets["burst"]
 
-    if mag_norm is not None:
-        file = write_cdf(mag_norm)
-        logger.info(f"Created CDF file at {file}")
+    if norm_data is not None:
+        mag_norm_raw = decom_mag.generate_dataset(
+            norm_data, mag_cdf_attrs.mag_l1a_norm_raw_attrs.output()
+        )
+        file = write_cdf(mag_norm_raw)
```
tech3371, line 32: If you are writing CDF file here, are you tracking output filepath somewhere that upload API can use later in `cli.py`?

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -17,14 +20,115 @@ def mag_l1a(packet_filepath):
...
+        if mag_l0.COMPRESSION:
+            raise NotImplementedError("Unable to process compressed data")
+
+        primary_start_time = TimeTuple(mag_l0.PRI_COARSETM, mag_l0.PRI_FNTM)
+        secondary_start_time = TimeTuple(mag_l0.SEC_COARSETM, mag_l0.SEC_FNTM)
+
+        # seconds of data in this packet is the SUBTYPE plus 1
+        seconds_per_packet = mag_l0.PUS_SSUBTYPE + 1
+
+        # now we know the number of secs of data in the packet, and the data rates of
```
tech3371: What is `secs`? is it seconds?

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -17,14 +20,115 @@ def mag_l1a(packet_filepath):
...
+        # each sensor, we can calculate how much data is in this packet and where the
+        # byte boundaries are.
+
+        # VECSEC is already decoded in mag_l0
+        total_primary_vectors = seconds_per_packet * mag_l0.PRI_VECSEC
```
tech3371, line 77: love these descriptive variable names! Makes it easy to follow

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -0,0 +1,254 @@
...
+        Vectors are grouped into primary sensor and secondary sensor, and returned as a
+        tuple (primary sensor vectors, secondary sensor vectors)
+
+        Written by MAG instrument team
+
+        Parameters
+        ----------
+        vector_data : np.ndarray
+            Raw vector data, in bytes. Contains both primary and secondary vector data
+            (first primary, then secondary)
```
tech3371, line 144: As usual, great comments!

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -0,0 +1,254 @@
...
+        primary_vectors = []
+        secondary_vectors = []
+
+        # Since the vectors are stored as 50 bit chunks but accessed via hex (4 bit
+        # chunks) there is some shifting required for processing the bytes.
+        # However, from a bit processing perspective, the first 48 bits of each 50 bit
+        # chunk corresponds to 3 16 bit signed integers. The last 2 bits are the sensor
+        # range.
+
+        for i in range(primary_count + secondary_count):  # 0..63 say
```
tech3371, line 172: love the side comments!

tech3371 (approved):
minor question but looks great!

imap_processing/mag/l0/mag_l0_data.py
maxinelasp: Sure!

imap_processing/mag/l0/mag_l0_data.py
maxinelasp: The length of the vectors has to be an even number. If theres any number left over from %2, meaning self.Vectors is not an even number, then there is a buffer at the end that needs to be removed. I'm not totally sure _why_ there is a buffer, but in some test data the vector length would be 201 with a zero padding at the end. I think it has to do with CCSDS weirdness, it's present in their packet spreadsheet.

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: You're right that this needs to happen, I'll figure it out and put it in the CDF PR (incoming soon)

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Yeah seconds, I'll update the comment

imap_processing/mag/l0/mag_l0_data.py
tech3371: Now that makes sense. If you can add this comment in there, I think it will help down the road looking back.

imap_processing/mag/l1a/mag_l1a.py
tech3371: sounds good!

PR #386, opened 2024-04-02
Added SPICE examples for Annotated Events

laspsandoval:
# Change Summary

## Overview
Provides some spice tools and an example of how to use them. Also contains code that puts particle velocities in different reference frames.

## New Files
- spice_examples.py
   - Far from perfect, but a good start, I think. This shows how some of the spice_utils can be used, gets time ranges for the attitude kernels, and puts the particle velocity into different reference frames.
- New kernels
   - Added kernels from https://lasp.colorado.edu/galaxy/display/IMAP/Data
   
## Deleted Files
- Deleted some kernels that are no longer needed.

## Updated Files
- spice_utils.py
   - removed list_attitude_coverage since querying spice kernels provides the same (cleaner) info

## Testing
- test_spice_utils.py
   - removed test for list_attitude_coverage
   - updated tests to account for new kernels
- test_spice_examples.py
   - tests spice_examples.py

Note for testing: de430.py is too large to commit and push to Github. Therefore to get the tests to run properly it will need to manually be added to the tests/test_data/spice directory.

## Important: Hacks
- imap_wkcp.tf
   - I added a section for 'IMAP_DPS'. DPS = despun reference frame which will need to be created by the SDC for each repointing.
- spice_examples.py
   -  I added the following line, which can be used to dynamically change the IMAP_DPS frame epoch.
spice.pdpool("FRAME_-43906_FREEZE_EPOCH", [802094471.0])

This will need to be readdressed. Nick Dutton is putting together a memo of how to handle it. This is just a hack to get the current code to work properly.


tools/spice/spice_examples.py
```diff
@@ -0,0 +1,132 @@
...
+    This of course will change.
+    """
+    time = direct_events["tdb"]
+    vultra = direct_events["vultra"]  # particle velocity in the ultra frame
+
+    # Particle velocity from instrument to spacecraft frame
+    vscbody = spice.mxv(spice.pxform("IMAP_ULTRA_45", "IMAP_SPACECRAFT", time), vultra)
+
+    # Particle velocity from spacecraft to dps frame
+    vdpssc = spice.mxv(spice.pxform("IMAP_SPACECRAFT", "IMAP_DPS", time), vscbody)
```
maxinelasp: Could we rename some of these variables to write out exactly what they are? using some common acronyms is fine (sc = spacecraft) but it's hard to understand this variable. 

for example:
vdpssc = velocity_sc_to_dps (if that is what this is?)
imapdpsvel = imap_dps_velocity
etc

tools/spice/spice_examples.py
```diff
@@ -0,0 +1,132 @@
...
+        ID of the instrument or spacecraft body.
+
+    Returns
+    -------
+    start: float
+        Start time in ET.
+    end: float
+        End time in ET.
+    """
+    # Get the first CK path
```
maxinelasp, line 33: Can you also define the acronyms used in the comments/docs? I think it's ok to leave "CK" as a common acronym in the variable names here, but it should be made clear in the docs what the "CK Kernel" is used for and contains.

tools/spice/spice_examples.py
```diff
@@ -0,0 +1,132 @@
...
+    end_utc = spice.et2utc(end, "C", 0)
+
+    logger.info(f"Coverage Interval for ID {id}:")
+    logger.info(f"Start Time (UTC): {start_utc}")
+    logger.info(f"End Time (UTC): {end_utc}")
+
+    return start, end
+
+
+def get_particle_velocity(direct_events):
```
maxinelasp: Also might want to mention in the docs here that this method assumes the kernel is loaded. Alternatively, if the intent is always for the end user to access through `build_annotated_events`, this method should say that and be preceded by an underscore to indicate it is a "private" method (`_get_particle_velocity`)

tools/tests/unit/test_spice_examples.py
```diff
@@ -0,0 +1,100 @@
...
+
+    spice.kclear()
+
+    spice.furnsh(kernels)
+    ck_kernel = list_files_with_extensions(kernel_directory, [".ck"])
+    spice.furnsh(ck_kernel)
+
+    get_particle_velocity(direct_events)
+
+    spice.kclear()
```
maxinelasp: Not for now, and not for here, but we should come up with some design to abstract away this `spice.kclear()` `spice.furnsh` `spice.kclear()` sequence in spice_utils. (perhaps using `with` keyword? https://www.geeksforgeeks.org/with-statement-in-python/)

maxinelasp (changes requested):
Exciting to see some spice stuff getting started! I'd like the documentation beefed up and the variable names made more clear before merge, but this looks good otherwise!

tools/spice/spice_examples.py
```diff
@@ -0,0 +1,132 @@
...
+
+import logging
+
+import spiceypy as spice
+
+from tools.spice.spice_utils import list_all_constants
+
+# Logger setup
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
```
tech3371:
```suggestion
logger = logging.getLogger(__name__)
```

tools/spice/spice_examples.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # Get the IDs from the CK file
+    ck_id = spice.ckobj(ck_kernel[0])
+
+    # Check if the specified ID is in the CK file
+    if int(id) not in ck_id:
+        raise ValueError(f"ID {id} not found in CK file")
+
+    # Get the coverage window
+    cover = spice.ckcov(ck_path, int(id), True, "SEGMENT", 0, "TDB")
```
tech3371, line 38: Can you add links like this above these functions for reference or for anyone who doesn't know about this yet?
https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.ckcov

I know it may be duplicate effort but knowing what those parameters are very helpful.

tools/spice/spice_examples.py
```diff
@@ -0,0 +1,132 @@
...
+    Returns
+    -------
+    vultra_heliosphere_frame: np.ndarray
+        Particle velocity in the heliosphere frame.
+
+    #Note: Using a single event time/velocity for now.
+    This of course will change.
+    """
+    time = direct_events["tdb"]
+    vultra = direct_events["vultra"]  # particle velocity in the ultra frame
```
tech3371: Since this is an example file, it's ok as it is. We can discuss how to do this in a generic way in the future. Time input makes sense. `vultra`, looks like velocity of particle is already calculated here and is it something Hi and Lo need to calculate?

tools/spice/spice_examples.py
```diff
@@ -0,0 +1,132 @@
...
+    end: float
+        End time in ET.
+    """
+    # Get the first CK path
+    ck_path = ck_kernel[0]
+
+    # Get the IDs from the CK file
+    ck_id = spice.ckobj(ck_kernel[0])
+
+    # Check if the specified ID is in the CK file
```
greglucas: Isn't this done by SPICE already? I thought SPICE raises an error for us already, so no need to check this ourselves first.

tools/spice/spice_examples.py
greglucas: Can SPICE not go directly between the two frames rather than needing to do it to an intermediate frame first?
```suggestion
    # Particle velocity from instrument to DPS frame
    vscbody = spice.mxv(spice.pxform("IMAP_ULTRA_45", "IMAP_DPS", time), vultra)
```

tools/spice/spice_examples.py
```diff
@@ -0,0 +1,132 @@
...
+        # We need a CK that defines the DPS frame. Nick Dutton putting
+        # together a memo on how this will work.
+        spice.pdpool("FRAME_-43906_FREEZE_EPOCH", [802094471.0])
+
+        time_event = spice.et2utc(direct_events["tdb"], "C", 0)
+
+        logger.info(f"Time (UTC): {time_event}")
+
+        # Just for demo purposes
+        list_all_constants()
```
greglucas: I don't think this does anything without logging added, so just remove?
```suggestion
```

tools/tests/test_data/spice/L1_de431.bsp
greglucas, line 1: 👁️ I did not realize these 4MB files were included before. I personally really don't want these large files in the repo because it just keeps growing then. We can add an automated download routine for these files from NAIF if we need them in CI jobs.

tools/tests/unit/test_spice_examples.py
```diff
@@ -0,0 +1,100 @@
...
+
+    return direct_events
+
+
+def test_get_attitude_timerange(kernels, kernel_directory):
+    """Tests get_attitude_timerange function."""
+
+    spice.kclear()
+
+    spice.furnsh(kernels)
```
greglucas: If there is an error in this routine, or we run tests in multiple processes in the future this will cause issues. We need to put this in an `with spice.Kernels():` call to properly clean up if issues arise during these tests.

tools/tests/unit/test_spice_examples.py
greglucas: I disagree, I think it is definitely for here and should be updated. This will mess things up if there is an error during the test or we run tests in different orders and things aren't cleaned up properly.

greglucas (changes requested):
We should discuss whether these kernels should be a part of the repository or not. In general I think these are external and shouldn't be included, they are really large and they aren't final yet.

tools/tests/unit/test_spice_examples.py
tmplummer: For EMUS, we abstracted away having to think too hard about furnishing kernels by writing a decorator that was added to any function making spiceypy calls. Here is the signature and documentation from that decorator:
```
def ensure_spice(f_py: callable = None,
                 time_kernels_only: bool = False,
                 start_time: int = None, end_time: int = None, time_units: str = 'sclk',
                 kernel_type: str = None):
    """
    Before trying to understand this piece of code, read this:
    https://stackoverflow.com/questions/5929107/decorators-with-parameters/60832711#60832711

    Decorator/wrapper that tries to ensure that a metakernel is furnished in as complete a way as possible.

    ======================
    Control flow overview:
    ======================
    1. Try simply calling the wrapped function naively.
    --> SUCCESS? Great! We're done.
    --> SpiceyError? Go to step 2.

    2. Furnish metakernel at EMUS_METAKERNEL_PATH
    --> SUCCESS? Great, return the original function again (so it can be re-run).
    --> KeyError? Seems like EMUS_METAKERNEL_PATH isn't set, no problem. Go to step 3.

    3. Did we get start_time and end_time parameters passed to the wrapper?
    --> YES? Cool, that means the user thinks we need a full metakernel to run this function. Go fetch one and furnish
        and return the original function (so it can be re-run).
    --> NO? There's still another possibility. Go te step 4.

    4. Did we get the parameter time_kernels_only=True?
    --> YES? We only need LSK and SCLK kernels to run this function. Go fetch those and furnish and return the original
        function (so it can be re-run).
    --> NO? Dang. This is sort of the end of the line. Re-raise the error generated from the failed spiceypy function
     call but add a better message to it.

    ======
    Usage:
    ======
    Three ways to use this object:
    1) A decorator with no arguments (will never query SDC API)
    ```
    @ensure_spice
    def my_spicey_func(a, b):
        pass
    ```
    2) A decorator with hardcoded, permanent parameters with which to fetch a metakernel from the API. This is useful
    if we only need the latest SCLK and LSK kernels for the function involved.
    ```
    @ensure_spice(time_kernels_only=True)
    def my_spicey_api_func(a, b):
        pass
    ```
    3) An explicit wrapper function, providing dynamic parameters to the SDC API call.
    ```
    wrapped = ensure_spice(spicey_func, start_et=362098457, end_time=37485342345, kernel_type='final')
    result = wrapped(a, b)
    ```

    :param f_py: callable, required, The function requiring SPICE that we are going to wrap if being used explicitly,
        Otherwise None, in which case ensure_spice is being used, not as a function wrapper (see l2a_processing.py) but
        as a true decorator without an explicit function argument.
    :param time_kernels_only: bool, optional, Specify that we only need to furnish time kernels (if
        EMUS_METAKERNEL_PATH is set, we still just furnish that metakernel.
    :param start_time: int, optional, Start ephemeris time to pass to the SDC SPICE API
    :param end_time: int, optional, End ephemeris time to pass to the SDC SPICE API
    :param time_units: str, optional, Defaults to SCLK seconds but may also be set to 'J2000' (ephemeris time)
    :param kernel_type: str, optional, Force specific kernel type (recon, final) in API call.
    """
```

I would recommend doing something similar for IMAP. I'm not sure where adding this type of feature fits with the current spice objectives. There are certainly some decisions that need to be made regarding the handling of spice kernels and metakernels that need to be make prior to implementation.

maxinelasp: A decorator seems like a great way to do this, I like it. Plus using shared code/ideas from other projects would be great!

bourque:
@all-contributors please add @subagonsouth for review!

tools/tests/test_data/spice/L1_de431.bsp
laspsandoval: I'll remove all kernels that can be downloaded from NAIF. Gavin Medley has some code that finds the most recent NAIF kernels and downloads them. That might be out of the scope for this PR since it would be quite a bit of additional code. He may have taken that code from EMM, which Tim would be familiar with if he is up for adding it in. But anyways, for now I'll remove the kernels and put that as a TODO.

tools/spice/spice_examples.py
laspsandoval: Yes. It should be an input. I'm not quite there yet, but I think it should be calculated for Ultra L1b.

tools/tests/unit/test_spice_examples.py
```diff
@@ -0,0 +1,82 @@
...
+        "vultra": np.array(
+            [-5.7718470e08, -8.8437594e08, -9.5545958e08], dtype=np.float32
+        ),
+        "tdb": 802100000.0,
+    }
+
+    return direct_events
+
+
+@pytest.mark.xfail(reason="Download NAIF kernels")
```
greglucas, line 55: Nice, thanks for adding these messages and xfails to indicate what we need to do to fix them up later!

greglucas (approved):
Thanks for removing the large kernels from this. I'm happy to push that discussion off to a follow-up PR.

One final question is whether we could remove all of the updates to the kernels here and just leave them at what they were previously (no removals, no additions) because you've xfailed all of those tests. Then we go back in a follow-up PR to add/remove what is necessary when we can look at adding download / other helper functions.

It has my approval as the additions are small-ish currently, but my preference would be to remove all binary file additions here if it is feasible and add them later.

maxinelasp (approved):
Addressed my issues, thanks! LGTM

PR #387, opened 2024-04-02
Update space_packet_parser version

tech3371:
# Change Summary
Update space_packet_parser version and fixed one I-ALiRT test that was failing.


imap_processing/tests/ialirt/unit/test_decom_ialirt.py
```diff
@@ -64,10 +64,8 @@ def test_generate_xarray_empty_file(tmp_path, xtce_ialirt_path):
     binary_file_path = tmp_path / "empty.ccsds"
     binary_file_path.touch()
 
-    expected_error_msg = (
-        "Reading off the end of the data. "
-        "Tried to read .* bits when only 0 available."
-    )
+    # TODO: Look into why it's giving different error message
+    expected_error_msg = "negative shift count"
```
greglucas: This test doesn't make sense to me, can we just remove it?

> Test that an error is raised if an empty file is passed to generate_xarray.

We are making sure space_packet_parser raises an error with bad data, but we can rely on them to test that.

cc @laspsandoval

bourque (approved):

imap_processing/tests/ialirt/unit/test_decom_ialirt.py
tech3371: I am happy to remove it if Laura is ok.

PR #393, opened 2024-04-05
CDF attrs bug fix

tech3371:
# Change Summary
added `File_naming_convention` to global attribute and fixed data_type in all instruments.

### DIFF
--- a/imap_processing/mag/mag_cdf_attrs.py
+++ b/imap_processing/mag/mag_cdf_attrs.py
@@ -52,7 +52,7 @@ mag_l1b_attrs = GlobalDataLevelAttrs(
 )
 
 mag_l1c_attrs = GlobalDataLevelAttrs(
-    "L1A_SCI>Level-1C Science Data",
+    "L1C_SCI>Level-1C Science Data",
     # TODO: replace "sci" with descriptor "norm" / "burst"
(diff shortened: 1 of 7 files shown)


maxinelasp (approved):
LGTM, comments are clear

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -44,15 +44,15 @@
...
+    "L1B_SCI>Level-1B Science Data",
     # TODO: replace "sci" with descriptor "norm" / "burst"
     logical_source="imap_mag_l1b_sci",
     logical_source_desc="IMAP Mission MAG Instrument Level-1B Data.",
     instrument_base=mag_base,
 )
 
 mag_l1c_attrs = GlobalDataLevelAttrs(
-    "L1A>Level-1C",
+    "L1A_SCI>Level-1C Science Data",
```
vmartinez-cu: Should "L1A_SCI" be "L1C_SCI" for the l1c attrs? Just looking at l1b attrs for comparison, but I could be missing something important.

tech3371: good catch. Yes. it should be.

sdhoyt (approved):
Thanks for fixing this!

PR #394, opened 2024-04-08
Added SWAPI science data CDF attribute

### LINKED ISSUE #188: Write SWAPI L1A data to CDF file

tech3371:
# Change Summary

## Overview
- minor fix and couple TODOs after SWAPI reviewed my code.
- added CDF attribute for SWAPI
- change upper case to lower case in dataset


imap_processing/swapi/swapi_cdf_attrs.py
bourque, line 1: I like the way this is organized, I am going to steal some ideas from this for CoDICE :)

bourque (approved):
Looks good!

imap_processing/swapi/swapi_cdf_attrs.py
tech3371: for sure!

imap_processing/swapi/swapi_cdf_attrs.py
```diff
@@ -0,0 +1,106 @@
...
+)
+
+# dimensions attributes
+energy_dim_attrs = AttrBase(
+    validmin=0,
+    validmax=72,
+    format="I2",
+    var_type="support_data",
+    fieldname="Energy Step",
+    catdesc="Energy step number.",
```
bryan-harter: Probably should be a bit more detailed, like "Energy steps at 72 energies from {x}-{y}eV", whatever the min and max energy are in the bins

imap_processing/swapi/swapi_cdf_attrs.py
```diff
@@ -0,0 +1,106 @@
...
+    label_axis="Energy Step",
+)
+
+# science data attributes
+counts_attrs = ScienceAttrs(
+    validmin=GlobalConstants.INT_FILLVAL,
```
bryan-harter: validmin should be different from the fillval, and it should probably be a realistic number (like 0?)

imap_processing/swapi/swapi_cdf_attrs.py
```diff
@@ -0,0 +1,106 @@
...
+    depend_1="energy",
+)
+
+# Uncertainty data is in Float and requires a different fill value,
+# format, min value.
+uncertainty_attrs = ScienceAttrs(
+    validmin=GlobalConstants.DOUBLE_FILLVAL,
+    validmax=GlobalConstants.INT_MAXVAL,
+    format="E13.5",
+    label_axis="Counts Uncertainty",
```
bryan-harter: label axis is generally just a 10 character label, I think just "Uncertainty" would be better here. Or "Cnts. Unc." which I think Andriy recommended to me on IDEX

imap_processing/swapi/swapi_cdf_attrs.py
```diff
@@ -0,0 +1,106 @@
...
+    var_type="data",
+    depend_0="epoch",
+    depend_1="energy",
+)
+
+compression_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=1,
+    format="I1",
+    label_axis="Compression Flag",
```
bryan-harter: Probably just "Flag" for the label_axis here.

imap_processing/swapi/swapi_cdf_attrs.py
tech3371: Good point. I don't know the exact energy level they use yet. These are id that will map to energy value in their lookup table. What you do you think if I change descriptions like this
```suggestion
    catdesc="Energy step id for lookup table",
```

imap_processing/swapi/swapi_cdf_attrs.py
tech3371: yes. min should be zero. I will change that.

imap_processing/swapi/swapi_cdf_attrs.py
tech3371: I like "Uncertainty"

imap_processing/swapi/swapi_cdf_attrs.py
tech3371: Is "Flag" too general?

bryan-harter: I believe the idea is that, if you were plotting it via CDAWeb, the FIELDNAM would appear above/below the plot as the title. In this case, the plot title would be "Compression Flag".  

And then LABLAXIS just labels the y axis on that plot.  So because the FIELDNAM is on the plot, the viewer would know what "Flag" is referring to here.

tech3371: I see. That makes sense.

bryan-harter:
Looks good!

PR #395, opened 2024-04-08
MAG L1A CDF

maxinelasp:
# Change Summary
Add the rest of L1A MAG files, including first draft of CDF and a refactor of the code I just put up for review (should have thought a little harder about that code)

## Overview

Besides the CDF changes, I reworked the way the MAG L1A data looks, by making sure there is a 1:1 relationship between MAG L1A objects and CDF files (each containing 24 hours of data). Also removed the Vector class because nparrays are way more efficient here. 

## Testing

Most of the tests are the same, with some minor changes due to format updates.


imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -46,84 +48,95 @@ def __add__(self, seconds: float):
...
+        """
+        return self.coarse_time + self.fine_time / MAX_FINE_TIME
 
 
 @dataclass
 class MagL1a:
     """
     Data class for MAG Level 1A data.
 
+    TODO: This object should end up having 1 days worth of data.
```
maxinelasp, line 68: This is mostly true right now, but I'm leaving some of the work for when SPICE kernels get involved.

imap_processing/cdf/global_attrs.py
```diff
@@ -86,11 +86,15 @@ class ConstantCoordinates:
...
+        # By default the time is assumed to correspond to instantaneous time or
+        # center of accumulation/measurement period. But if, for some good reason,
+        # time corresponds to the beginning or end or any other part of
+        # accumulation/measurement period, that has to be stated in CATDESC.
+        "CATDESC": "Time, number of nanoseconds since J2000 with leap seconds included",
         "FIELDNAM": "epoch",
         "FILLVAL": GlobalConstants.INT_FILLVAL,
-        "FORMAT": "a2",
         "LABLAXIS": "epoch",
+        "FORMAT": "",  # Supposedly not required, fails in xarray_to_cdf
```
tech3371, line 97: I would think format is required since every data we write has some type of data type such as int, string, float or date. I don't know what the format would be for `EPOCH` though since it's a datetime object. @bryan-harter may know.

imap_processing/cdf/utils.py
```diff
@@ -12,15 +12,15 @@
...
+def calc_start_time(shcoarse_time: float) -> np.datetime64:
     """Calculate the datetime64 from the CCSDS secondary header information.
 
     Since all instrument has SHCOARSE or MET seconds, we need convert it to
     UTC. Took this from IDEX code.
 
     Parameters
     ----------
-    shcoarse_time: int
+    shcoarse_time: float
```
tech3371, line 23: All of instrument's unpacked SHCOARSE should be 32-bits int like you had in `MagLo` class. Does it get converted to float in the MAG processing code? Since this is used by everyone, I want to ask if this effect others if you change the date type here?

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -110,8 +110,7 @@ def __post_init__(self):
             dtype=np.dtype(">b"),
         )
 
-        # Remove buffer from end of vectors. Vector data needs to be in 50 bit chunks,
-        # and may have an extra byte at the end from CCSDS padding.
+        # Remove buffer from end of vectors
```
tech3371: was those comments reverting to old one from rebase? I like the old comments :)

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -81,54 +148,120 @@ def process_packets(mag_l0_list: list[MagL0]) -> dict[str, list[MagL1a]]:
...
+    dataset_attrs: dict
+        Global attributes for the dataset, as created by mag_attrs
+
+    Returns
+    -------
+    dataset : xr.Dataset
+        One xarray dataset with proper CDF attributes and shape containing MAG L1A data.
+    """
+    # TODO: Just leave time in datetime64 type with vector as dtype object to avoid this
+    time_data = mag_l1a.vectors[:, 4].astype(np.dtype("datetime64[ns]"), copy=False)
```
tech3371: I see the same thing below. What does 4 in `[:, 4]` represent?

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -42,6 +42,38 @@
     instrument_base=mag_base,
 )
 
+mag_l1a_norm_mago_attrs = GlobalDataLevelAttrs(
+    "L1A-raw-norm-mago>Level 1A MAGo normal rate",
```
tech3371: I recently found out that we need to have in this format, `data_type=<datalevel>_<descriptor>`. Can you change same thing in the below attrs as well?
```suggestion
    "L1A_raw-norm-mago>Level 1A MAGo normal rate",
```

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -65,15 +97,30 @@
...
-    catdesc="Magnetic field vectors",
+    catdesc="Raw, unprocessed magnetic field vector data in bytes",
     depend_0="epoch",
     depend_1="direction",
     display_type="time_series",
     fieldname="Magnetic Field Vector",
-    label_axis="Magnetic field vector",
+    label_axis="Raw binary magnetic field vector data",
+    fill_val=GlobalConstants.INT_MAXVAL,
+    format="I3",
```
tech3371, line 104: I think you need to increase format to something like `I12` or higher because you value range is int min and max which integer of 3-digit(`I3`) won't be enough. Same for others. I have recently noticed those inconsistent things in my work as well. I am starting to learn more about CDF.

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -118,3 +176,80 @@
     display_type="time_series",
     label_axis="Magnetic field vector",
 )
+
+# Catdesc (<80 chars), Fieldnam (<30 chars)
+
+catdesc_fieldname_l0 = {
+    # TODO: Don't include PUS_SPARE1?
+    "PUS_SPARE1": ["Spare header from ESA Standard", "Spare header"],
```
tech3371, line 178: could you change these list to dictionary as well? I was wondering in above code what those 0 or 1 index mean. Having those key referenced may be much easier to follow without looking up here.
```suggestion
    "PUS_SPARE1": { "catdesc": "Spare header from ESA Standard", "fieldnam": "Spare header"},
```

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -25,22 +29,76 @@ def mag_l1a(packet_filepath):
...
+
+    Returns
+    -------
+    generated_files: list[str]
+        A list of generated filenames
+    """
+    if not packet_data:
+        return []
+
+    mag_raw = decom_mag.generate_dataset(packet_data, raw_attrs)
```
tech3371, line 81: can you call attrs here instead of passed in? Just curious
```suggestion
    mag_raw = decom_mag.generate_dataset(packet_data, mag_cdf_attrs.mag_l1a_burst_raw_attrs.output())
```

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -25,22 +29,76 @@ def mag_l1a(packet_filepath):
...
+
+    file = write_cdf(mag_raw)
+    logger.info(f"Created RAW CDF file at {file}")
+
+    generated_files = [file]
+
+    l1a = process_packets(packet_data)
+
+    for _, mago in l1a["mago"].items():
+        norm_mago_output = generate_dataset(mago, mago_attrs)
```
tech3371, line 91: same here and `magi`
```suggestion
        norm_mago_output = generate_dataset(mago, mag_cdf_attrs.mag_l1a_burst_mago_attrs.output())
```

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -81,54 +148,120 @@ def process_packets(mag_l0_list: list[MagL0]) -> dict[str, list[MagL1a]]:
...
+        )
+        secondary_timestamped_vectors = MagL1a.calculate_vector_time(
+            secondary_vectors, mag_l0.SEC_VECSEC, secondary_start_time
+        )
+
+        # Sort primary and secondary into MAGo and MAGi by 24 hour chunks
+        # TODO: Individual vectors should be sorted by day, not the whole packet
+        mago_is_primary = mag_l0.PRI_SENS == 0
+
+        mago_day = primary_day if mago_is_primary else secondary_day
```
tech3371, line 168: So, `mago_day` and `magi_day` is the date of earliest datetime in the data packets based on which primary sensor value? Looks like you envisioning to add handlers for multiple days of data in your TODO.

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -81,54 +148,120 @@ def process_packets(mag_l0_list: list[MagL0]) -> dict[str, list[MagL1a]]:
...
-                mag_l0.SHCOARSE,
-                secondary_vectors,
-            )
+    return {"mago": mago, "magi": magi}
 
-        mago.append(mago_l1a)
-        magi.append(magi_l1a)
 
-    return {"mago": mago, "magi": magi}
+def generate_dataset(mag_l1a: MagL1a, dataset_attrs: dict):
```
tech3371: For a second, I thought there is a circular loop happening here because of variable and function named same thing here. I thought `mag_l1a` was calling function `mag_l1a` which would result in circular loop but you were not. can you rename this variable name to something else to avoid this confusion?

tech3371: Can you add a test function where it calls `mag_l1a()`? It may help in understanding which order these functions are run.

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -81,35 +76,41 @@ def test_process_vector_data():
...
-    assert test_data.vectors[3] == Vector((3, 3, 3, 4), TimeTuple(10001, 32769))
+    test_vecsec = 2
+    start_time = TimeTuple(10000, 0)
+
+    test_data = MagL1a.calculate_vector_time(test_vectors, test_vecsec, start_time)
+
+    converted_start_time_ns = calc_start_time(start_time.to_seconds())
+
+    skips_ns = np.timedelta64(int(1 / test_vecsec * 1e9), "ns")
+    print(skips_ns)
```
tech3371: remove?

imap_processing/__init__.py
```diff
@@ -48,4 +48,4 @@
 }
 
 # Reference start time (launch time or epoch)
-launch_time = np.datetime64("2010-01-01T00:01:06.184")
+launch_time = np.datetime64("2010-01-01T00:01:06.184", "ns")
```
tech3371, line 51: Thank you for catching this!

imap_processing/cdf/defaults.py
```diff
@@ -28,4 +28,5 @@ class GlobalConstants:
     INT_MAXVAL = np.iinfo(np.int64).max
     DOUBLE_FILLVAL = np.float64(-1.0e31)
     MIN_EPOCH = -315575942816000000
+    # TODO: This is set to 2029, should be further out. Determine the correct value.
```
tech3371: good catch!

imap_processing/cdf/global_attrs.py
maxinelasp: This was from an email from Andriy, apparently it isn't needed for Epoch but 🤷

imap_processing/cdf/utils.py
maxinelasp: This is used not for SHCOARSE time, but for a more precise time from the MAG instrument which includes fractional seconds. But, this is a good point, that changing the type here might affect other instruments. I'm happy to make a different method if you'd prefer to keep the SHCOARSE separate from other spacecraft time conversions. To answer your question, it shouldn't affect others.

imap_processing/mag/l0/mag_l0_data.py
maxinelasp: Damn, merge fail. I'll put it back the way it was.

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: This is getting the timestamp off the `L1A.vectors` numpy array. Each vector has [x, y, z, range, timestamp] in the array. I'll add a comment making that clear here.

imap_processing/mag/mag_cdf_attrs.py
maxinelasp: Yes, good catch, I should have updated this after your attr PR!

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Yes, exactly. Eventually I need to split the individual vectors down by day as well, which should involve a minor change to this code. I was planning on leaving it, because there is some weirdness with time adjustment that I need to implement as well, so I didn't want to write a bunch of code and end up not using it.

imap_processing/cdf/defaults.py
maxinelasp: All Andriy! I wasn't actually sure how to calculate a different Epoch, which is why I didn't change it (or what a good number would be)

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Yes, I'll also rename and add comments to the functions. I agree that it's confusing, I tried to be clear but there are a lot of `mag_l1a`s running around still!

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Same issue, different attrs for norm and burst

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Norm and burst use different attrs, so unfortunately not.

imap_processing/cdf/utils.py
maxinelasp: Alternatively, I could make it clear in this function that it isn't just for SHCOARSE but for other types of time conversions? I assumed that we will eventually replace this with SPICE time conversions, so I could borrow it in a temporary way for MAG processing too.

maxinelasp:
I did get an email from MAG with some pretty specific and in some cases somewhat extensive changes, but I'll put those in a different review to keep it small (er) here.

imap_processing/cdf/global_attrs.py
tech3371: interesting! It makes sense little bit since the language it uses may not have datetime type? cool. Then let's leave EPOCH's format as you have it.

imap_processing/cdf/utils.py
tech3371: Since all tests passed, I think it's ok to leave it as it is. We are going to need to refactor this function or implement another function do below code since many of us are doing this similar thing. We can handle this in future PR.
```
epoch_converted_time = [calc_start_time(time) for time in epoch_time]
```

imap_processing/mag/l1a/mag_l1a.py
tech3371: I appreciate that!

tech3371 (approved):
It looks good to me but may be good to have look by another person in case I missed something.

imap_processing/cdf/global_attrs.py
```diff
@@ -86,11 +86,15 @@ class ConstantCoordinates:
     """
 
     EPOCH: ClassVar[dict] = {
-        "CATDESC": "Default time",
+        # By default the time is assumed to correspond to instantaneous time or
+        # center of accumulation/measurement period. But if, for some good reason,
+        # time corresponds to the beginning or end or any other part of
+        # accumulation/measurement period, that has to be stated in CATDESC.
```
bourque, line 92: Thanks for adding this clarification!

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -25,22 +29,76 @@ def mag_l1a(packet_filepath):
...
+    -------
+    generated_files: list[str]
+        A list of generated filenames
+    """
+    if not packet_data:
+        return []
+
+    mag_raw = decom_mag.generate_dataset(packet_data, raw_attrs)
+
+    file = write_cdf(mag_raw)
```
bourque:
```suggestion
    filepath = write_cdf(mag_raw)
```

Not very important, but I find this to be a bit more readable (`filepath` indicates to me that this is a `str`/`Path` object and not some sort of object with file contents)

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -81,54 +148,120 @@ def process_packets(mag_l0_list: list[MagL0]) -> dict[str, list[MagL1a]]:
...
-    return {"mago": mago, "magi": magi}
+def generate_dataset(mag_l1a: MagL1a, dataset_attrs: dict):
+    """
+    Generate a Xarray dataset for L1A data to output to CDF files.
+
+    Global_attrs should contain all info about mago/magi and burst/norm distinction, as
+     well as any general info in the global attributes.
+
+     Assumes each MagL1a object is a single day of data, so one MagL1a object has one
+     CDF file output.
```
bourque, line 224:
```suggestion
    well as any general info in the global attributes.

    Assumes each MagL1a object is a single day of data, so one MagL1a object has one
    CDF file output.
```

imap_processing/mag/l1a/mag_l1a.py
bourque:
```suggestion
def generate_dataset(mag_l1a: MagL1a, dataset_attrs: dict) -> xr.Dataset:
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -46,84 +48,95 @@ def __add__(self, seconds: float):
 
         return TimeTuple(coarse, fine)
 
+    def to_seconds(self):
```
bourque:
```suggestion
    def to_seconds(self) -> float:
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -46,84 +48,95 @@ def __add__(self, seconds: float):
...
-    vectors_per_second: int
-    expected_vector_count: int
-    seconds_of_data: int
     SHCOARSE: int
-    vectors: list
+    vectors: np.array
 
-    def __post_init__(self):
+    @staticmethod
+    def calculate_vector_time(vectors, vectors_per_second, start_time):
```
bourque:
```suggestion
    def calculate_vector_time(vectors, vectors_per_second, start_time) -> np.array:
```

imap_processing/mag/mag_cdf_attrs.py
```diff
@@ -42,6 +42,38 @@
     instrument_base=mag_base,
 )
 
+mag_l1a_norm_mago_attrs = GlobalDataLevelAttrs(
+    "L1A-raw-norm-mago>Level 1A MAGo normal rate",
+    # Should also include data type
+    logical_source="imap_mag_l1a_norm-mago",
+    logical_source_desc="IMAP Mission MAG Instrument Level-1A Data.",
```
bourque: If these are all going to be the same, you could make a `mag_l1a_desc` variable for this string to avoid some redundancy? If you think it improves readability. i.e.

```python
mag_l1a_desc = "IMAP Mission MAG Instrument Level-1A Data."
logical_source_desc=mag_l1a_desc
```

imap_processing/mag/mag_cdf_attrs.py
bourque: I like how you've organized these!

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -81,35 +76,41 @@ def test_process_vector_data():
 
 
 def test_time_tuple():
-    test_time_tuple = TimeTuple(439067318, 64618)
+    ex_time_tuple = TimeTuple(439067318, 64618)
```
bourque:
```suggestion
    example_time_tuple = TimeTuple(439067318, 64618)
```

bourque (approved):
Nice work! I just have some nitpicky suggestions if you so choose.

imap_processing/cdf/defaults.py
bryan-harter: Might be my bad, I believe I took an example number on the ISTP website and assumed it was the "default" VALIDMAX attribute.  MMS uses "2100-01-01T00:00:00" as their VALIDMAX so I vote we use that too.  We are using MMS's VALIDMIN, which is "1900-01-01T00:00:00".  

That would make the value: 

3155630469184000000

imap_processing/cdf/global_attrs.py
bryan-harter: Whelp that's going to make the code slightly messier.  I'll try to quickly get a change to cdflib to allow blank FORMAT when the variable is named "epoch".

maxinelasp: No need to rush it, I think, but thanks!

imap_processing/cdf/defaults.py
maxinelasp: Seems like a good number, I'll update it!

imap_processing/mag/mag_cdf_attrs.py
maxinelasp: I actually suspect that this should specify norm/burst/mago/magi in it... 🤔

bryan-harter: I think they should probably all be a little different, like this one should be IMAP Mission MAG O (whatever "O" stands for) Normal Rate Level-1A Data

bryan-harter (approved):

imap_processing/mag/mag_cdf_attrs.py
bryan-harter: Oh we made this comment just a few minutes apart lol, I hadn't seen that you wrote when I made my comment.

maxinelasp: O stands for outboard, I stands for inboard. Basically specifying which instrument and where.

PR #402, opened 2024-04-17
CoDICE lo-sw-species-counts test data

### LINKED ISSUE #252: Produce CoDICE L1A CDFs
Remaining work:

- [x] #767 
- [x] #768 
- [x] #769 
- [x] #770 
- [x] #771 
- [x] #772 
- [x] #773 
- [x] #774 
- [x] #775 
- [x] #776 
- [x] #785 
- [x] #809 
- [x] #992 
- [x] #1154 
- [x] #1155

bourque:
# Change Summary

## Overview

This PR contains changes to implement newly acquired CoDICE `lo-sw-species-counts` test data and produce a corresponding L1a CDF

## New Files

- `imap_processing/tests/codice/data/lo_fsw_view_5_ccsds.bin`
   - New `lo-sw-species-counts` test data

## Deleted Files

- `imap_processing/tests/codice/data/sample_science_data.bin`
   - Old test data that is not needed anymore

## Updated Files

   - Some update CDF attributes for `lo-sw-species-counts` data
   - Added (back) parameter for specifying XTCE file to use for decom, so that different types of packets can be decommed instead of just housekeeping
   - Removed code for producing simulated data, as that is no longer needed with the newly acquired test data
   - Changes to produce CDF file for `lo-sw-species-counts`
   - Added lists for variable names for species, used for `lo-sw-species-counts` CDFs
- `imap_processing/tests/codice/test_codice_l0.py` and `imap_processing/tests/codice/test_codice_l1a.py`
   - Updated tests to reflect new `xtce_document` parameter


bourque:
Tests are failing because the test data size is different from what the packet definition expects -- I will need to modify the packet definition to use variable-length data.

imap_processing/codice/cdf_attrs.py
```diff
@@ -86,29 +48,41 @@
     variable_purpose="PRIMARY",
 )
 
-l1a_science_attrs = ScienceAttrs(
+l1a_hskp_attrs = GlobalDataLevelAttrs(
+    data_type="L1A_SCI->Level-1A Science Data",
+    logical_source="imap_codice_l1a_hskp",
+    logical_source_desc="IMAP Mission CoDICE Instrument Level-1A Data",
```
greglucas:
```suggestion
    logical_source_desc="IMAP Mission CoDICE Instrument Level-1A Housekeeping Data",
```

imap_processing/codice/cdf_attrs.py
```diff
@@ -86,29 +48,41 @@
...
+    instrument_base=codice_base,
+)
+
+l1a_lo_sw_species_attrs = GlobalDataLevelAttrs(
+    data_type="L1A_SCI->Level-1A Science Data",
+    logical_source="imap_codice_l1a_lo-sw-species-counts",
+    logical_source_desc="IMAP Mission CoDICE Instrument Level-1A Data",
```
greglucas: Expand SW too?
```suggestion
    logical_source_desc="IMAP Mission CoDICE Instrument Level-1A SW Species Counts",
```

imap_processing/codice/codice_l0.py
```diff
@@ -15,21 +15,22 @@
     packet_list = decom_packets(packet_file, xtce_document)
 """
 
-from imap_processing import decom, imap_module_directory
+from imap_processing import decom
 
 
-def decom_packets(packet_file: str) -> list:
+def decom_packets(packet_file: str, xtce_document) -> list:
```
greglucas: Rather than passing in the XTCE document everywhere you want to use this would it be easier to keep a mapping in this and then pass in a string of the type you want to decommutate? That way only this one file needs to know the path locations and not the tests too.

```suggestion
_PACKET_TYPES = {"housekeeping": path_to_hk_definition, ...}
def decom_packets(packet_file: str, packet_type="housekeeping") -> list:
    if packet_type not in _PACKET_TYPES:
        raise ...
    xtce_document = _PACKET_TYPES[packet_type]
```

imap_processing/codice/codice_l1a.py
```diff
@@ -80,101 +73,59 @@ class CoDICEL1aPipeline:
...
             attrs=ConstantCoordinates.EPOCH,
         )
 
-        species_array = xr.DataArray(
-            name="species",
-            data=np.zeros(36864),
-            dims=("species"),
-            attrs=species_attrs,
+        energy = xr.DataArray(
+            np.arange(128),
```
greglucas, line 107: Do we need to clarify that this is an index / steps in the name at all rather than just "energy"

imap_processing/codice/codice_l1a.py
```diff
@@ -80,101 +73,59 @@ class CoDICEL1aPipeline:
...
-            data = xr.DataArray(
-                name=species_name,
-                data=list(species_data),
-                dims=("species"),
-                attrs=species_attrs,
+        for species_data, varname, fieldname in zip(
+            self.data, LO_SW_SPECIES_VARNAMES, LO_SW_SPECIES_FIELDNAMES
+        ):
+            species_data_arr = [int(item) for item in species_data]
+            species_data_arr = np.array(species_data_arr).reshape(-1, 128)
```
greglucas, line 122:
```suggestion
            species_data_arr = np.array(species_data).astype(int).reshape(-1, 128)
```

imap_processing/codice/constants.py
```diff
@@ -6,8 +6,28 @@
 
 from imap_processing.codice.utils import CoDICECompression
 
+# CDF-friendly FIELDNAMES for lo-sw-species-counts data product
+LO_SW_SPECIES_FIELDNAMES = [
+    "SW - H+",
```
greglucas: What is SW? Will there ever be another identifier, or should that be dropped because we are already in the `sw-species` file?

imap_processing/codice/constants.py
```diff
@@ -6,8 +6,28 @@
...
+    "SW - Si",
+    "SW - Fe lowQ",
+    "SW - Fe highQ",
+    "SW - He+ (PUI)",
+    "SW - CNO+ (PUI)",
+]
+
 # CDF-friendly variable names for lo-sw-species-counts data product
-LO_SW_SPECIES_NAMES = [
+LO_SW_SPECIES_VARNAMES = [
```
greglucas: Would it be easier to see the mapping between the two lists if it is more explicit as a list of tuples?

```python
LO_SW_SPECIES_FIELDNAMES_VARNAMES = [
    ("SW - H+", "hplus"),
    ...
]

File: imap_processing/tests/codice/data/sample_science_data.bin
Comment (greglucas) [inline imap_processing/tests/codice/data/sample_science_data.bin:1]: Is this "bad" data now?

File: imap_processing/codice/codice_l0.py (hunk 3 above)
Comment (bourque) [reply imap_processing/codice/codice_l0.py]: Thanks, great suggestion. I didn't feel great about adding this parameter in, I like this solution better!

File: imap_processing/codice/cdf_attrs.py [hunk 8]
```diff
@@ -86,29 +48,41 @@
...
+
+energy_attrs = AttrBase(
     validmin=0,
-    validmax=GlobalConstants.INT_MAXVAL,
-    display_type="TBD",
-    depend_0="epoch",
-    depend_1="Energy",
-    depend_2="Counts",
+    validmax=127,
     format="I12",
```
tech3371: nit picking
```suggestion
    format="I3",
```

imap_processing/codice/codice_l1a.py
tech3371: I wonder that about my instruments too. I curious to know too!

imap_processing/codice/codice_l1a.py
```diff
@@ -366,17 +290,14 @@ def process_codice_l1a(packets) -> str:
...
-            # sorted_packets = sort_by_time(packets, "SHCOARSE")
+            packets = sort_by_time(grouped_data[apid], "SHCOARSE")
 
             # Get the four "main" parameters for processing
-            table_id, plan_id, plan_step, view_id = get_params(apid)
+            table_id, plan_id, plan_step, view_id = get_params(packets[0])
 
             # Run the pipeline to create a dataset for the product
             pipeline = CoDICEL1aPipeline(table_id, plan_id, plan_step, view_id)
+            pipeline.packets = packets
```
tech3371: If user forget to initialized `pipeline.packets` here, then looks like `pipeline.unpack_science_data()` and `pipline.create_science_dataset()` will fail since it's not in the init function. Can you add a check and may be raise value missing error?

imap_processing/codice/codice_l1a.py
bourque: I will check with Joey on if the name matters within the CDF file, but for this python variable name, I can change it to `energy_steps`.

imap_processing/codice/constants.py
bourque: These names are what CoDICE wants in the CDF file, so we will need to stick with "SW" here. But I will add some clarification on what these acronyms mean ("Sunward" in this case).

imap_processing/codice/constants.py
bourque: Great suggestion!

imap_processing/tests/codice/data/sample_science_data.bin
bourque: Yes, it was some simulated data that Joey created that I was using, but it was always meant to be temporary.

imap_processing/codice/codice_l1a.py
```diff
@@ -80,101 +73,59 @@ class CoDICEL1aPipeline:
...
         # Create a data variable for each species
-        for species_data, species_name in zip(self.data, LO_SW_SPECIES_NAMES):
-            data = xr.DataArray(
-                name=species_name,
-                data=list(species_data),
-                dims=("species"),
-                attrs=species_attrs,
+        for species_data, varname, fieldname in zip(
+            self.data, LO_SW_SPECIES_VARNAMES, LO_SW_SPECIES_FIELDNAMES
+        ):
```
tech3371: I wonder if you can set this relation between `self.data` and those two list differently. I don't have a good suggestion at the moment. But something to ponder for later.

tech3371 (approved):
couple minor comments. Looks like test data helped a lot. Nice!

imap_processing/codice/codice_l1a.py
bourque: Good catch!

bourque: On second thought, I think it makes sense to pass in the `packets` as a parameter to those two methods, so I will do that.

Issue #77, opened 2023-08-25
SPIKE - CDF file generation 


bryan-harter:
## Description

I'm trying to gather my thoughts here, because I can think of 100 different ways to structure the code and I'm not sure what would be useful for you all and the L3 developers.  These go in order from least hand-holding/overhead to most hand-holding/overhead.  

## Requirements
- Allow instrument data to be written to a CDF file easily, without needing to know much about the internal structure of the CDF file.  

### Nice to have or Goal Requirements
- Checks if ISTP compliant
- Checks if compliant with mission naming standards

## Related tickets
None

## Follow up tickets
This spike is not considered complete until at least one follow up issue is created. 
- [ ] Placeholder ticket

# Solution 1
Simply read instrument data directly into an xarray.DataSet() object.  When you're done manipulating it, call cdflib.xarray.xarray_to_cdf to generate the file.  Keep a dictionary of your attributes for the dataset somewhere in the code.  

## Pros:
- Working directly with xarray is nice, because xarray has some neat data analysis tools that could make processing easier
- More freedom to tweak things in the xarray as you're processing

## Cons:
- You don't know if your xarray/CDF is ISTP compliant until you call the "xarray_to_cdf" function
- You don't know if your xarray meets mission criteria (for example, all variable names must be lower case, etc).  That is only caught outside of the code (code reviews, SPDF reviews, science board meetings, etc)
- No easy way to make a skeleton CDF.  
- Forces us all to use xarray DataSet objects and know how they work
- xarray works best once you have the complete dataset.  In otherwords, you really can't create an empty object and then start appending data to it as you read packets.  It will error because it always wants the size/shape to match between all the variables in the dataset.  So you need to store the data in some intermediate variable, and then read them in all at once to xarray.     

## Additional notes:
- This means there are very few CDF utility functions, other than the one build into cdflib  

# Solution 2
Same as above, but with an extra function living in a utils directory: "verify_imap_cdf()" which loops through the variables in the xarray,DataSet object and ensures they meet certain criteria.  For example:
1) all the names meet mission standards 
2) all the attributes are ISTP compliant 
3) Dimensions of the data match what the attributes say they are
4) etc

## Pros:
- Same as solution 1
- Removes some cons from solution 1

## Cons:
- Forces you to call an extra function before generating the cdfs, what if you forget?!
- Still, xarray works best when you have the complete dataset.  See the Con in Solution 1.  
[...]

PR #403, opened 2024-04-17
Lo L1A Science Counts CDF Creation

sdhoyt:
# Change Summary

## Overview
This PR adds attributes and dataset creation for Lo L1A Science Counts data product

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
- imap_processing/lo/l1a/lo_cdf_attrs.py
   - added attributes for L1A science counts
- imap_processing/lo/l1a/lo_l1a_write_cdfs.py
   - added dataset creation for science counts

## Testing
Added units tests for data set data fields and cdf write file naming: imap_processing/tests/lo/test_lo_l1a_write_cdfs.py

Resulting CDF from unit test can be found here: https://lasp.colorado.edu/galaxy/pages/viewpage.action?pageId=197632025


imap_processing/lo/l1a/lo_cdf_attrs.py
```diff
@@ -41,10 +41,16 @@
     instrument_base=lo_base,
 )
 
+lo_sci_cnt_l1a_attrs = GlobalDataLevelAttrs(
+    data_type="L1A_SCICNT>Level-1A Science Counts",
+    logical_source="imap_lo_l1a_scicnt",
+    logical_source_desc="IMAP Mission IMAP-Lo Instrument Level-1A Data",
```
greglucas, line 47: There are two IMAPs here. Do we want to just put "IMAP Mission Lo instrument" or "IMAP-Lo Instrument ..."

imap_processing/lo/l1a/lo_cdf_attrs.py
```diff
@@ -41,10 +41,16 @@
...
+    instrument_base=lo_base,
+)
+
 # TODO: Add rest of data products for L1A, L1B, L1C
 
-# TODO: Figure out what attributes I need for the
-# energy, time, mode, checksum, and position fields
+## Science Direct Events Attributes ##
```
greglucas, line 53: Is there a reason for 2 pound signs before/after here and 3 for the next one? Is the next one a subsection of this?

imap_processing/lo/l1a/lo_cdf_attrs.py
```diff
@@ -58,3 +64,109 @@
     units="seconds",
     label_axis="ToF",
 )
+
+### Science Counts Attributes ###
+lo_anode_rate_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=GlobalConstants.INT_FILLVAL,
```
greglucas, line 71: This reads odd being the valid maximum, but then putting a Global Constant Fill Value which is usually outside of the maximum I think. Should we add a `GlobalConstants.INT_MAX` or something to distinguish them?

imap_processing/lo/l1a/lo_cdf_attrs.py
```diff
@@ -58,3 +64,109 @@
...
+    fieldname="Anode Single Rate",
+    format="I12",
+    var_type="data",
+    units="seconds",
+    label_axis="anode single rate",
+)
+lo_ion_rate_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=GlobalConstants.INT_FILLVAL,
+    catdesc="Ion Single Rate",
```
greglucas, line 85: Should this be plural?
```suggestion
    catdesc="Ion Singles Rate",
```

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
```diff
@@ -121,3 +126,158 @@ def create_lo_scide_dataset(sci_de: list):
...
+    scide: list
+        List of ScienceCounts data class objects.
+
+    Returns
+    -------
+    xarray.Dataset
+        Lo L1A Science Counts Dataset.
+    """
+    # TODO: How should the times be computed? Using START_A + SHCOARSE for now
+    sci_cnt_times = np.concatenate(
```
greglucas, line 146: I don't think you are concatenating here but just creating an array?
```suggestion
    sci_cnt_times = np.array(
```

Below you are doing this already and then adding dtype too, so I think you could add dtype here as well or `.astype(...)`

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
```diff
@@ -121,3 +126,158 @@ def create_lo_scide_dataset(sci_de: list):
...
+
+    # create the dataset without populating data
+    sci_cnt_dataset = xr.Dataset(
+        data_vars={},
+        attrs=lo_cdf_attrs.lo_sci_cnt_l1a_attrs.output(),
+        # TODO: figure out how to convert time data to epoch
+        coords={"epoch": sci_cnt_epoch},
+    )
+
+    # populate the dataset fields
```
greglucas, line 166: There is a lot of repetition below, can you abstract any of it out into a helper function or similar? You are also iterating through `sci_cnt` N times which would be nice to turn into just 1 iteration and extract all of your variables at once.

```python
# Array of shape (N, 9 [START_A, START_C, ...])
packet_array = np.array([(sci_cnt_data.START_A, sci_cnt_data.START_C, ...) for sci_cnt_data in sci_cnt])
# Loop and create datasets
for i, (name, attrs) in enumerate(zip(("start_a", "start_c", ...), (lo_cdf_attrs.lo_anode_rate_attrs.output(), ...))):
    ds[name] = xr.DataArray(packet_array[:, i], dims="epoch", attrs=attrs)
```

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
```diff
@@ -121,3 +126,158 @@ def create_lo_scide_dataset(sci_de: list):
...
+    sci_cnt_dataset["start_a"] = xr.DataArray(
+        np.concatenate([sci_cnt_data.START_A for sci_cnt_data in sci_cnt]),
+        dims="epoch",
+        attrs=lo_cdf_attrs.lo_anode_rate_attrs.output(),
+    )
+    sci_cnt_dataset["start_c"] = xr.DataArray(
+        np.concatenate([sci_cnt_data.START_C for sci_cnt_data in sci_cnt]),
+        dims="epoch",
+        attrs=lo_cdf_attrs.lo_anode_rate_attrs.output(),
```
greglucas, line 175: This is using the same attributes as above, is that OK or does it need a different name for the variable or anything like that to distinguish it?

imap_processing/tests/lo/test_lo_l1a_write_cdfs.py
```diff
@@ -80,11 +154,52 @@ def test_create_lo_scide_dataset(direct_events):
...
+        "pos3",
+        "hydrogen",
+        "oxygen",
+    ]
+
+    expected_epoch = np.array(
+        np.array(
+            [1712787078, 1712787090, 1712787100, 1712787150, 1712787170, 1712787200]
+        )
+        * 1e9,
```
greglucas, line 188: Can you bring this outside and only create one array and then multiply by `1e9` later? Or set the `datetime64` to be seconds

tmplummer (commented):
This is not a comment that is specific to this PR, but this is the first time I am taking a look at how data is flowing from parsed packets to CDF files. This is certainly not meant to pick on you @sdhoyt .

Much of the code seems so formulaic and repeated that I wonder if it wouldn't be better to come up with a configuration file that captures all of the necessary details and a utility that uses that configuration to automatically define the CDF attributes and populate the `xr.DataArray` objects.

sdhoyt:
> This is not a comment that is specific to this PR, but this is the first time I am taking a look at how data is flowing from parsed packets to CDF files. This is certainly not meant to pick on you @sdhoyt .
> 
> Much of the code seems so formulaic and repeated that I wonder if it wouldn't be better to come up with a configuration file that captures all of the necessary details and a utility that uses that configuration to automatically define the CDF attributes and populate the `xr.DataArray` objects.

@subagonsouth That's a good idea. I've been thinking about how overwhelming this file is going to get as I need to add more attributes and have been trying to think of an alternative. Are you thinking something like a json file with the attributes that are unpacked to set `ScienceAttrs()`? For example:

```python
sci_cnt_dataset["oxygen"] = xr.DataArray(
        np.concatenate([sci_cnt_data.OXYGEN for sci_cnt_data in sci_cnt]),
        dims="epoch",
        attrs=ScienceAttrs(**config["lo_oxygen"]).output(),
    )
```

tmplummer:
I guess that I need a github tutorial. I want to be able to reply to this comment:


> > This is not a comment that is specific to this PR, but this is the first time I am taking a look at how data is flowing from parsed packets to CDF files. This is certainly not meant to pick on you @sdhoyt .
> > Much of the code seems so formulaic and repeated that I wonder if it wouldn't be better to come up with a configuration file that captures all of the necessary details and a utility that uses that configuration to automatically define the CDF attributes and populate the `xr.DataArray` objects.
> 
> @subagonsouth That's a good idea. I've been thinking about how overwhelming this file is going to get as I need to add more attributes and have been trying to think of an alternative. Are you thinking something like a json file with the attributes that are unpacked to set `ScienceAttrs()`? For example:
> 
> ```python
> sci_cnt_dataset["oxygen"] = xr.DataArray(
>         np.concatenate([sci_cnt_data.OXYGEN for sci_cnt_data in sci_cnt]),
>         dims="epoch",
>         attrs=ScienceAttrs(**config["lo_oxygen"]).output(),
>     )
> ```


I was thinking even a level higher than that (if I was even thinking of something concrete, not just an abstract idea)
Something like perhaps a CSV file that defines all of the attributes in a CDF and links them to the dataclass in some way (ideally it would reference all the way back to the packet, but I see that things like decompression get in the way of that). For example:
```
name,validmin,validmax,catdesc,depend_0,display_type,fill_val,fieldname,format,var_type,units,label_axis,dims,data_class_field
oxygen,0,INT_FILLVAL,Oxygen Rate,epoch,time_series,INT_FILLVAL,Oxygen Rate,I12,data,seconds,oxygen rate,OXYGEN
```
Then
```python
# Some pseudo code here
with read_csv('config.csv') as csv:
    for row in csv:
        attrs = ScienceAttrs(**get_science_attrs_from_row(row))
        sci_cnt_dataset[row.name] = xr.DataArray(
             np.concatenate([sci_cnt_data[row.data_class_field] for sci_cnt_data in sci_cnt]),
            dims=row.dims,
            attrs=attrs.output(),
        ) 
```

greglucas:
@subagonsouth, I remember seeing these yaml configurations at one point for an alternative way to do this. Is this similar to what you're suggesting?

https://github.com/HERMES-SOC/hermes_core/blob/640c23f9b636527a899970b665c1bdb6835bfd12/hermes_core/data/hermes_default_global_cdf_attrs_schema.yaml

https://github.com/HERMES-SOC/hermes_core/blob/640c23f9b636527a899970b665c1bdb6835bfd12/hermes_core/util/schema.py

Here is a related issue with some discussion as well: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/77

imap_processing/lo/l1a/lo_cdf_attrs.py
sdhoyt: I did that to differentiate that comment as a section. I'm trying to figure out a better way to keep the attributes file from getting out of hand. The sections were just something to indicate it in the meantime. I can remove the doubles though, it's not that helpful.

imap_processing/lo/l1a/lo_cdf_attrs.py
sdhoyt: The fill val is negative, so do you mean the validmin should change? I thought the fill vals didn't count as a valid number to account for in the range since it's meant to indicate missing data.

imap_processing/lo/l1a/lo_cdf_attrs.py
sdhoyt: Yeah it should be

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
sdhoyt:
```python
np.concatenate([sci_cnt_data.START_A for sci_cnt_data in sci_cnt])
```

`sci_cnt` is a list of ScienceCount objects and the attributes that I'm pulling from each object are numpy arrays, so without the concatenation this would result in a list of numpy arrays, but I'm trying to concatenate those arrays together. I do need to check with Colin to make sure this is correct for these fields though.

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
sdhoyt: Yeah, I know the repetition isn't sustainable. I was trying to stay in line with how I saw others setting these since it's easier to read / debug, but a lot of these products have too many fields to do that. I'll change it to loop through the field names and possibly move to Tim's config file suggestion in the future.

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
sdhoyt: Yeah, they're both anode singles rates fields, just different anodes so I _think_ it's ok to have the same attributes. That's something I need Colin to review though.

imap_processing/tests/lo/test_lo_l1a_write_cdfs.py
sdhoyt: Yeah, that was a result of trying to figure out why the times weren't lining up and I forgot to clean that code up.

imap_processing/lo/l1a/lo_cdf_attrs.py
greglucas: I think the doubles is fine, I was just wondering why you used doubles here and triples below.
```python
## Comment here ##
...
### Comment in next section ###
```

Another option would be to do similar to our docs and put underlines or similar after this heading
```python
# Header
# -------
```

But, this is all fine as-is, I was just curious

imap_processing/lo/l1a/lo_cdf_attrs.py
greglucas: Then this would read as `valid maximum is -199999` or whatever it is. I think this validmax should be the maximum value this rate could take which would be more like `99999` or a positive number. The max shouldn't be less than the min and I'm guessing SPDF would complain about.

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
greglucas: Oh interesting, this wasn't clear to me that START_A is an array, that sounds like an individual quantity to me. Maybe it woiuld be helpful to change the variables to START_As plural, or START_A_arr to indicate it is an array/multiple possible values?

imap_processing/lo/l1a/lo_cdf_attrs.py
sdhoyt: Oh sorry, I misunderstood your comment and didn't notice that in the code. Yeah, I put the wrong value there for validmax. I'll update that to INT_MAXVAL

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
sdhoyt: Yeah, I have information about the shape of those variables in the docstring for the data class: 
```python
"""
START_A : numpy.ndarray
        Single rates for electron, anode A. 2D array Azimuth(6), Energy(7)
"""
```

I'm calling it START_A because that's what is in the packet definition which is also what Lo wants in the L1A CDF as the names for the arrays, so I want to keep the names consistent throughout. Are you ok with that being a little unclear to keep the naming consistent? If there's a question about the data shape, there should be enough information in the class docs.

imap_processing/lo/l1a/lo_cdf_attrs.py
sdhoyt: Gotcha, yeah, I didn't meant to use a different number of `#`. I'll make that less confusing.

imap_processing/lo/l1a/lo_cdf_attrs.py
```diff
@@ -58,3 +64,109 @@
     units="seconds",
     label_axis="ToF",
 )
+
+### Science Counts Attributes ###
+lo_anode_rate_attrs = ScienceAttrs(
```
tech3371, line 69: Is this data going to be int or float? So far, rate has been float for me. If it's float, some of these attrs default value may need to change.

imap_processing/lo/l1a/lo_cdf_attrs.py
tech3371: looks like others were using INT_FILLVAL instead of INT_MAXVAL as well.

imap_processing/lo/l1a/lo_cdf_attrs.py
```diff
@@ -58,3 +64,109 @@
...
+    validmin=0,
+    validmax=GlobalConstants.INT_FILLVAL,
+    catdesc="Anode Single Rate",
+    depend_0="epoch",
+    display_type="time_series",
+    fill_val=GlobalConstants.INT_FILLVAL,
```
tech3371, line 75: If this data is float type, you can use float fillval option from this class.

imap_processing/lo/l1a/lo_cdf_attrs.py
```diff
@@ -58,3 +64,109 @@
...
+    fieldname="Anode Single Rate",
+    format="I12",
```
tech3371, line 77: If this data is float type, you can use `E12.2`.

imap_processing/lo/l1a/lo_cdf_attrs.py
```diff
@@ -58,3 +64,109 @@
...
+    var_type="data",
+    units="seconds",
+    label_axis="triple coincidence rate",
+)
+lo_disc_rate_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=GlobalConstants.INT_FILLVAL,
+    catdesc="Discarded Rate",
+    depend_0="epoch",  # TODO: Not sure if this is right
```
tech3371, line 125:
```suggestion
    depend_0="epoch",
```

imap_processing/tests/lo/test_lo_l1a_write_cdfs.py
```diff
@@ -63,6 +65,78 @@ def direct_events():
...
+        "POS2",
+        "POS3",
+        "HYDROGEN",
+        "OXYGEN",
+    ]
+
+    scicnt1.SHCOARSE = np.array([1712787078, 1712787090, 1712787100])
+    scicnt2.SHCOARSE = np.array([1712787150, 1712787170, 1712787200])
+    for field in fields:
+        setattr(scicnt1, field, np.array([0, 0, 0]))
```
sdhoyt, line 124: These test inputs array shapes don't make sense for this data. Need to make them 6x7 and 60x7 arrays

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
```diff
@@ -121,3 +126,158 @@ def create_lo_scide_dataset(sci_de: list):
...
+        List of ScienceCounts data class objects.
+
+    Returns
+    -------
+    xarray.Dataset
+        Lo L1A Science Counts Dataset.
+    """
+    # TODO: How should the times be computed? Using START_A + SHCOARSE for now
+    sci_cnt_times = np.concatenate(
+        [sci_cnt_data.START_A + sci_cnt_data.SHCOARSE for sci_cnt_data in sci_cnt]
```
sdhoyt, line 147: START_A is 6x7 matrix, this doesn't make sense. I'll use SHCOARSE only for now. Still need to get more info from Lo on how to store the epoch.

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
```diff
@@ -121,3 +126,158 @@ def create_lo_scide_dataset(sci_de: list):
...
+        name="epoch",
+        attrs=ConstantCoordinates.EPOCH,
+    )
+
+    # create the dataset without populating data
+    sci_cnt_dataset = xr.Dataset(
+        data_vars={},
+        attrs=lo_cdf_attrs.lo_sci_cnt_l1a_attrs.output(),
+        # TODO: figure out how to convert time data to epoch
+        coords={"epoch": sci_cnt_epoch},
```
sdhoyt, line 163: the coords should have azimuth and energy instead of epoch?

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
```diff
@@ -121,3 +126,158 @@ def create_lo_scide_dataset(sci_de: list):
...
+
+    # populate the dataset fields
+    sci_cnt_dataset["start_a"] = xr.DataArray(
+        np.concatenate([sci_cnt_data.START_A for sci_cnt_data in sci_cnt]),
+        dims="epoch",
+        attrs=lo_cdf_attrs.lo_anode_rate_attrs.output(),
+    )
+    sci_cnt_dataset["start_c"] = xr.DataArray(
+        np.concatenate([sci_cnt_data.START_C for sci_cnt_data in sci_cnt]),
+        dims="epoch",
```
sdhoyt, line 174: All of these should have azimuth and energy dimensions instead of epoch

sdhoyt: or maybe they should have azimuth, energy, and epoch dimensions

imap_processing/lo/l1a/lo_cdf_attrs.py
sdhoyt: I'm not doing any conversions to engineering units in L1A so these should all be ints

tmplummer:
> @subagonsouth, I remember seeing these yaml configurations at one point for an alternative way to do this. Is this similar to what you're suggesting?
> 
> https://github.com/HERMES-SOC/hermes_core/blob/640c23f9b636527a899970b665c1bdb6835bfd12/hermes_core/data/hermes_default_global_cdf_attrs_schema.yaml
> 
> https://github.com/HERMES-SOC/hermes_core/blob/640c23f9b636527a899970b665c1bdb6835bfd12/hermes_core/util/schema.py
> 
> Here is a related issue with some discussion as well: #77

@greglucas Good question. I certainly don't have something specific in mind. I find it advantageous to have product definitions all in one place, usually in a configuration file rather than in code. Based on the discussion in #77, I can see that you all have put lots of thought into this already. I am new to CDF and the tools being used to generate them, so I am probably the least qualified to make any decisions regarding implementation of this concept.

The HERMES-SOC approach looks useful for defining the products. It seems to me that it would be useful to add a piece that maps the dataclass(es) to the `xr.DataSet`. This would allow for generic code, common to all of the instruments, to do the work of populating the `xr.DataSet`.

All of this is outside the scope of this issue IMO.

imap_processing/lo/l1a/lo_l1a_write_cdfs.py
vmartinez-cu: Tenzin updated a generic create_dataset function in utils.py that might be useful for looping through fields and creating xarrays. See PR #532

tech3371:
Should we close this PR?

PR #404, opened 2024-04-18
Updating CLI to fix MAG 

maxinelasp:
# Change Summary
Missed this section during my refactor. Just a minor change to remove old filename management.

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -67,8 +67,13 @@ def write_cdf(dataset: xr.Dataset):
     start_time = np.datetime_as_string(dataset["epoch"].values[0], unit="D").replace(
         "-", ""
     )
+
+    # Will now accept vXXX or XXX formats, as batch starter sends versions as vXXX.
     r = re.compile(r"v\d{3}")
-    if r.match(dataset.attrs["Data_version"]) is None:
+    if (
+        not isinstance(dataset.attrs["Data_version"], str)
+        or r.match(dataset.attrs["Data_version"]) is None
+    ):
         version = f"v{int(dataset.attrs['Data_version']):03d}"  # vXXX
     else:
         version = dataset.attrs["Data_version"]
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -1,6 +1,7 @@
 """Various utility functions to support creation of CDF files."""
 
 import logging
+import re
 
 import imap_data_access
 import numpy as np
@@ -66,7 +67,11 @@ def write_cdf(dataset: xr.Dataset):
     start_time = np.datetime_as_string(dataset["epoch"].values[0], unit="D").replace(
         "-", ""
     )
-    version = f"v{int(dataset.attrs['Data_version']):03d}"  # vXXX
+    r = re.compile(r"v\d{3}")
+    if r.match(dataset.attrs["Data_version"]) is None:
+        version = f"v{int(dataset.attrs['Data_version']):03d}"  # vXXX
+    else:
+        version = dataset.attrs["Data_version"]
     repointing = dataset.attrs.get("Repointing", None)
     science_file = imap_data_access.ScienceFilePath.generate_from_inputs(
         instrument=instrument,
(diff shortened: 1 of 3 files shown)


imap_processing/cdf/utils.py
```diff
@@ -66,7 +67,11 @@ def write_cdf(dataset: xr.Dataset):
     start_time = np.datetime_as_string(dataset["epoch"].values[0], unit="D").replace(
         "-", ""
     )
-    version = f"v{int(dataset.attrs['Data_version']):03d}"  # vXXX
+    r = re.compile(r"v\d{3}")
+    if r.match(dataset.attrs["Data_version"]) is None:
+        version = f"v{int(dataset.attrs['Data_version']):03d}"  # vXXX
```
maxinelasp, line 77: The original version was throwing errors because batch starter provides the version as "vXXX"

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -78,6 +84,11 @@ def process_and_write_data(
     if not packet_data:
         return []
 
+    # TODO: Rework attrs to be better
+    raw_attrs["Data_version"] = data_version
```
maxinelasp, line 88: This is a temporary fix, I wanted to run L1A in the cloud today for MAG but I have rewritten the way these attributes work in my WIP branch.

imap_processing/cdf/utils.py
tech3371: I see. can you add this comment in the code?

tech3371 (approved):
looks great!

imap_processing/cdf/utils.py
maxinelasp: Sure!

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -40,19 +40,25 @@ def mag_l1a(packet_filepath) -> list[Path]:
...
     return generated_files
 
 
 def process_and_write_data(
-    packet_data: list[MagL0], raw_attrs: dict, mago_attrs: dict, magi_attrs: dict
+    packet_data: list[MagL0],
+    raw_attrs: dict,
+    mago_attrs: dict,
+    magi_attrs: dict,
+    data_version: str,
```
tmplummer, line 61: It is unclear to me which part of these changes is a temporary. If the addition of `data_version` to this function signature is not temporary, please add it to the docstring.

imap_processing/cli.py
```diff
@@ -311,23 +311,13 @@ def process(self):
...
-                # TODO: figure out data_dir, because now this fails.
-                #  Should switch to using IMAP_DATA_DIR env var.
-                if filename_norm.exists():
-                    logger.info(f"Uploading file: {filename_norm}")
-                    imap_data_access.upload(filename_norm)
-                if filename_burst.exists():
-                    logger.info(f"Uploading file: {filename_burst}")
-                    imap_data_access.upload(filename_burst)
+                if len(output_files) == 0:
+                    print("No files to upload.")
```
tmplummer, line 317: Should these `print`s be logging calls?

imap_processing/cdf/utils.py
```diff
@@ -66,7 +67,11 @@ def write_cdf(dataset: xr.Dataset):
     start_time = np.datetime_as_string(dataset["epoch"].values[0], unit="D").replace(
         "-", ""
     )
-    version = f"v{int(dataset.attrs['Data_version']):03d}"  # vXXX
+    r = re.compile(r"v\d{3}")
+    if r.match(dataset.attrs["Data_version"]) is None:
```
tmplummer: `r.match` will raise a TypeError if `dataset.attrs["Data_version"]` is not a string. Is this a possibility?

tmplummer (commented):
Looks good overall.

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: All of it is temporary, including data_version in this change. I need to add a bunch of global attributes so I'm reworking the way this works.

imap_processing/cli.py
maxinelasp: When they were logging calls, the AWS batch runner wouldn't print them into the cloudwatch logs

imap_processing/cdf/utils.py
maxinelasp: It is a possibility but I think it would be an incorrect usage of dataset attributes. But, I can add a catch here if it's an int.

PR #531, opened 2024-04-23
Chilean partner example

laspsandoval:
# Change Summary

## Overview
This is not intended to be merged into imap_processing. This is solely for the purposes of reviewing the code. I did not commit SPICE kernels. The README lists the SPICE kernels that may be downloaded from https://lasp.colorado.edu/galaxy/display/IMAP/Data.

## New Files
- README.md
   - Instructions for developers
 - spice_examples_partners.py
   - Calculates the azimuth, elevation, and range from observer to target. The observer is at latitude, longitude, and altitude of Longovilo-Hacienda, Chile. The target is the IMAP spacecraft.
 - spice_output_data.txt
   - Test data created by spice_examples_partners.py.

## Testing
 - test_spice_examples_partners.py
   - I calculated the azimuth and elevation of the Sun wrt the observer in Longovilo-Hacienda, Chile and then compared that with an external library (pvlib.solarposition import get_solarposition) for a single day. The results match. On the same plot please find the results of the IMAP satellite as well.

Screenshot 2024-04-23 at 1 29 50 PM

## Context
The group in Chile we are working with are from Entel, the Chilean telecommunications company, and specifically, the group who oversees the ground stations in Longovilo Chile. The rationale for the request is that they are trying to understand if the motion control system of their 30 and/or 32-m will be sufficient for IMAP. Right now they have a pretty manual process that requires them to manually program in the antenna pointing at each step position and they are proposing an upgrade to their control system but want to make sure that they can adequately track IMAP in the sky with the upgrades they are planning.
 
Hopefully that context helps a bit. Furthermore, they are not familiar with deep space satellite assets and initially struggled a bit to understand tracking a satellite that wasn’t in earth orbit thus the rationale for helping them understand the differences with tracking and receiving from something like IMAP at L1.


tools/spice/spice_examples_partners.py
```diff
@@ -0,0 +1,289 @@
...
+    observer_position_ecef = latitude_longitude_to_ecef(longitude, latitude, altitude)
+
+    method = "Ellipsoid"  # Only method supported
+    abcorr = "None"  # Aberration correction
+    azccw = False  # Azimuth measured clockwise from the positive y-axis
+    elplsz = True  # Elevation increases from the XY plane toward +Z
+    obsctr = "EARTH"  # Name of the center of motion
+    obsref = "IAU_EARTH"  # Body-fixed, body-centered reference frame wrt obsctr
+
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.azlcpo
```
tmplummer, line 162: Reading this documentation prompted the following thought...
I don't know what the precision requirements are for pointing the antenna, but to be exact, you actually want to point at were the spacecraft was when it sent the data which is the current time minus the one-way-light time. So you would have to shift your azimuth elevation pointing table by that amount.
Again, I don't think that matters here but I found it interesting to think about.

tmplummer (approved):
This looks great. I like that you validated by comparing to an external source for the sun location.

tools/spice/spice_examples_partners.py
```diff
@@ -0,0 +1,289 @@
...
+            azimuth, elevation, range_to_target = (
+                f"{azimuth:.4f}",
+                f"{elevation:.4f}",
+                f"{range_to_target:.4f}",
+            )
+
+            # Format data to be written to text file
+            data_line = (
+                f"{time}, {time_utc}, {azimuth}, " f"{elevation}, {range_to_target}\n"
+            )
```
greglucas: Minor nits:
* Put the format strings directly where you plan to use them here rather than the two-part format strings.
* Remove the space after the comma for a csv file.
* Do you want to include any other values like you had before with the J2000 x/y/z values?
```suggestion
            data_line = (
                f"{time},{time_utc},{azimuth:.4f},{elevation:.4f},{range_to_target:.4f}\n"
            )
```

tools/spice/spice_examples_partners.py
```diff
@@ -0,0 +1,289 @@
...
+            str(kernel_directory / "IMAP_launch20250429_1D.bsp"),
+            start_et_input,
+            stop_et_input,
+        )
+
+        # Get azimuth (degrees) and elevation (degrees) and
+        # range (km) to the target for each time step
+        for time in range(int(start_et_input), int(stop_et_input), time_interval):
+            # Convert to UTC time for readability
+            time_utc = spice.et2utc(time, "C", 0)
```
greglucas: I find the ISOC to be easier and more portable to read.

```suggestion
            time_utc = spice.et2utc(time, "ISOC", 0)
```

Options from: https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/et2utc_c.html
```bash
                  "C"      "1986 APR 12 16:31:09.814"
                  "D"      "1986-102 // 16:31:12.814"
                  "J"      "JD 2446533.18834276"
                  "ISOC"   "1987-04-12T16:31:12.814"
                  "ISOD"   "1987-102T16:31:12.814"
```

tools/spice/spice_examples_partners.py
```diff
@@ -0,0 +1,289 @@
...
+
+    # Define the path to the kernel directory relative to this script
+    kernel_directory = Path(__file__).parents[0]
+    kernels = get_kernels(kernel_directory)
+
+    # Load spice kernels and perform calculations.
+    with (
+        spice.KernelPool(kernels),
+        open(
+            "../tests/test_data/spice_partner_example/spice_output_data.txt", "w"
```
greglucas: I'd make it a csv extension just to be explicit.
```suggestion
            "../tests/test_data/spice_partner_example/spice_output_data.csv", "w"
```

greglucas (commented):
I think this looks great. I don't know much about the SPICE stuff, but in general it looks good and is what I'd expect for following the Sun around.

One other plot to think about would be an az/el (x/y) plot with the dots color coded with time to illustrate the spacecraft and sun going across the sky from the current position. The antenna will only be able to pick up the signal when the elevation is above 0 (probably higher than that), but give a good indication when day/night is and observer times of interest.

tools/spice/spice_examples_partners.py
laspsandoval: Good point. I looked into spice.azlcpo and there is the parameter abcorr. If set to "LT+S" it accounts for this. So that I think that should solve your concern.

"LT+S"     Correct for one-way light time and
                             stellar aberration using a Newtonian
                             formulation. This option modifies the
                             state obtained with the "LT" option to
                             account for the observer's velocity
                             relative to the solar system
                             barycenter. The result is the apparent
                             state of the target---the position and
                             velocity of the target as seen by the
                             observer.

laspsandoval:
Responded to comments and did not merge. Sent to ALiRT project management 4/24/24.

PR #532, opened 2024-04-24
updates to utils create_dateset function

tech3371:
# Change Summary

## Overview
Updated `imap_processing/utils.py`'s create_dataset function to be more generic and updated IMAP-Hi Housekeeping processing to use it.

## Updated Files
Two main files that was changed
- imap_processing/utils.py
- imap_processing/hi/l1a/housekeeping.py

### DIFF
--- a/imap_processing/hi/l1a/housekeeping.py
+++ b/imap_processing/hi/l1a/housekeeping.py
@@ -1,8 +1,5 @@
 """Unpack IMAP-Hi housekeeping data."""
 
-import collections
-import dataclasses
-
 import xarray as xr
 from space_packet_parser.parser import Packet
 
@@ -10,11 +7,11 @@ from imap_processing.cdf.global_attrs import ConstantCoordinates
 from imap_processing.cdf.utils import calc_start_time
 from imap_processing.hi.hi_cdf_attrs import (
     hi_hk_l1a_attrs,
-    hi_hk_l1a_metadata_attrs,
 )
+from imap_processing.utils import create_dataset
 
 
-def create_dataset(packets: list[Packet]) -> xr.Dataset:
+def process_housekeeping(packets: list[Packet]) -> xr.Dataset:
     """Create dataset for each metadata field.
 
     Parameters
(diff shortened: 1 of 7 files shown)


imap_processing/utils.py
```diff
@@ -101,17 +101,22 @@ def convert_raw_to_eu(dataset: xr.Dataset, conversion_table_path, packet_name):
     return dataset
 
 
-def create_dataset(packets: list[Packet], met_name="shcoarse") -> xr.Dataset:
+def create_dataset(
```
tmplummer, line 105: Ooooo. I didn't know about this function. Love that it is being made more generic. If I am not mistaken it can be used for any product that is generated directly from the packet storing everything in one-dimensional, 12-bit CDF variables? Some details about usage restrictions should be added to the docstring.

I wonder if functionality could be added to allow for custom unpacking of certain keys in the input packets as well as CDF variables that are not type "I12"?

imap_processing/utils.py
```diff
@@ -101,17 +101,22 @@ def convert_raw_to_eu(dataset: xr.Dataset, conversion_table_path, packet_name):
...
         packet list
-    met_name : str
-        metadata name to use as epoch time
+    met_name : str, Optional
+        Default is "shcoarse" because many instrument uses it.
+        This key is used to get spacecraft time for epoch dimension.
+    include_header: bool, Optional
+        Whether to include CCSDS header data in the dataset
+    skip_keys: list, Optional
+        Keys to skip in the metadata
```
tmplummer, line 123: What is `metadata` in this context? I think it means that matching fields in the input packets are not included in the output dataset?

imap_processing/utils.py
```diff
@@ -122,8 +127,21 @@ def create_dataset(packets: list[Packet], met_name="shcoarse") -> xr.Dataset:
...
     for data_packet in packets:
+        current_data = (
+            (data_packet.header | data_packet.data)
+            if include_header
+            else data_packet.data
+        )
+
+        # Drop keys using skip_keys
+        if skip_keys is not None:
+            current_data = {
```
tmplummer: This is very minor considering a relatively small size of the `current_data` dict, but this would modify the dictionary in place rather than replicating a new one.
```suggestion
            for key in skip_keys:
                current_data.pop(key, None)
```

imap_processing/utils.py
```diff
@@ -101,17 +101,22 @@ def convert_raw_to_eu(dataset: xr.Dataset, conversion_table_path, packet_name):
...
 
-    # TODO: change IMAP-Hi to use this and others as needed.
-
     Parameters
     ----------
     packets : list[Packet]
         packet list
-    met_name : str
-        metadata name to use as epoch time
+    met_name : str, Optional
```
tmplummer: Do some packets contain a value that is valid to directly use as the epoch coordinate? The few examples that I have seen require a function that converts what is the packet to what is desired to be stored in the epoch coordinate. For example, a SPICE call to convert CCSDS_MET to a datetime object?

tmplummer (commented):
Great so see some work to make a function more generic and reusable.

imap_processing/utils.py
tmplummer: I see now that I was misinterpreting the `format="I12"` and now understand that this is to display an integer with a field width of 12. Ignore that part of my comment.

imap_processing/hi/l1a/housekeeping.py
```diff
@@ -1,20 +1,17 @@
...
 from imap_processing.cdf.utils import calc_start_time
 from imap_processing.hi.hi_cdf_attrs import (
     hi_hk_l1a_attrs,
-    hi_hk_l1a_metadata_attrs,
 )
+from imap_processing.utils import create_dataset
 
 
-def create_dataset(packets: list[Packet]) -> xr.Dataset:
+def process_housekeeping(packets: list[Packet]) -> xr.Dataset:
```
bourque, line 12: I made a very similar change on CoDICE, so I feel validated 😃

imap_processing/utils.py
```diff
@@ -101,17 +101,22 @@ def convert_raw_to_eu(dataset: xr.Dataset, conversion_table_path, packet_name):
     return dataset
 
 
-def create_dataset(packets: list[Packet], met_name="shcoarse") -> xr.Dataset:
+def create_dataset(
+    packets: list[Packet], met_name="shcoarse", include_header=True, skip_keys=None
```
bourque: I wonder if something like `spacecraft_time_key` would be better / more explicit than `met_name`. When I look at `met_name` it is hard to tell what it is used for.

imap_processing/utils.py
bourque: Nice improvements to this function!

imap_processing/utils.py
```diff
@@ -122,8 +127,21 @@ def create_dataset(packets: list[Packet], met_name="shcoarse") -> xr.Dataset:
     description_dict = {}
 
     for data_packet in packets:
+        current_data = (
```
bourque:
```suggestion
        data_to_include = (
```

I was a bit confused at first what `current` meant in this context. I think it is fine to leave as is but just wanted to offer this suggestion.

bourque (approved):
I like how `create_dataset` has evolved over the last several months 😄

imap_processing/hi/l1a/housekeeping.py
tech3371: yay! same mind think alike

imap_processing/utils.py
tech3371: @subagonsouth let me know if you like clarification on this.

yes. I was trying to make this function more generic as I go along.

imap_processing/utils.py
tech3371: yes. I appreciate better name suggestions because I was not happy with what I had.

imap_processing/utils.py
tech3371: In this phase, I am not converting because some uses this exact key to sort data by this time. May be we should change time here in the format the CDF desire(datetime64 object). I could do that in upcoming PR.

imap_processing/utils.py
tech3371: Here, I am using metadata as keys from packet definition after CCSDS header keys. I appreciate other term suggestion.

imap_processing/utils.py
tech3371: Thank you!

tech3371:
I am going to merge this so that others can use it sooner. We can update this more in future PR as we go along.

PR #533, opened 2024-04-24
Updates based on SPDF feedback

tech3371:
# Change Summary

## Overview
Updates to IMAP-Hi and SWAPI CDF attributes based on SPDF feedback

### DIFF
--- a/imap_processing/cdf/defaults.py
+++ b/imap_processing/cdf/defaults.py
@@ -26,8 +26,12 @@ class GlobalConstants:
 
     INT_FILLVAL = np.iinfo(np.int64).min
     INT_MAXVAL = np.iinfo(np.int64).max
+    FLOAT_MAXVAL = np.finfo(np.float64).max
     DOUBLE_FILLVAL = np.float64(-1.0e31)
     # 1900-01-01T00:00:00
     MIN_EPOCH = -315575942816000000
     # 2100-01-01T00:00:00
     MAX_EPOCH = 3155630469184000000
+    # 32-bits max int value
+    UBIT32_MAXVAL = 2**32 - 1
+    UBIT16_MAXVAL = 2**16 - 1
--- a/imap_processing/cdf/defaults.py
+++ b/imap_processing/cdf/defaults.py
@@ -33,5 +33,5 @@ class GlobalConstants:
     # 2100-01-01T00:00:00
     MAX_EPOCH = 3155630469184000000
     # 32-bits max int value
-    UBIT32_MAXVAL = 2**32 - 1
-    UBIT16_MAXVAL = 2**16 - 1
+    UINT32_MAXVAL = np.iinfo(np.uint32).max
+    UINT16_MAXVAL = np.iinfo(np.uint16).max
--- a/imap_processing/cdf/global_attrs.py
+++ b/imap_processing/cdf/global_attrs.py
@@ -213,17 +213,26 @@ class AttrBase:
     display_type : str default="no_plot"
         The display type of the plot (ex "no_plot"), required
     catdesc : str, default=None
-        The category description, "CATDESC" attribute, required
+        The category description, "CATDESC" attribute, required.
+        Max 80 characters. If need to write more, use "var_notes".
     fieldname : str, default=None
-        The fieldname, "FIELDNAM" attribute
+        The fieldname, "FIELDNAM" attribute. Max 30 characters.
+        fieldname and label_axis parameter are closely related.
+        fieldname and label_axis are used in plot. fieldname
+        is used as title of the plot, label_axis is used as axis label.
+        For example, fieldname='Time of flight', label_axis='TOF'
     var_type : str, default="support_data"
-        The type of data
+        The type of data. Valid options are:
+            1. "data" - plotted and listed on CDAWeb and accessed by API,
+            2. "support_data" - not plotted
+            3. "metadata" - reserved for character type variables.
     fill_val : np.int64, default=Constants.INT_FILLVAL
         The values for filling data
     scale_type : str, default="linear"
         The scale of the axis, "SCALETYP" attribute
     label_axis : str, default=None
         Axis label, "LABLAXIS" attribute. Required. Should be close to 6 letters.
+        See fieldname parameter for more information.
     format : str, default=None
         The format of the data, in Fortran format
     units : str, default=None
--- a/imap_processing/hi/hi_cdf_attrs.py
+++ b/imap_processing/hi/hi_cdf_attrs.py
@@ -8,6 +8,7 @@ from imap_processing.cdf.global_attrs import (
 )
 from imap_processing.hi import __version__
 
+# TODO: add more information about dataset
 text = (
     "IMAP-Hi consists of two identical, single-pixel"
     "high-energy energetic neutral atom (ENA) imagers"
@@ -15,27 +16,9 @@ text = (
     "relative to the spacecraft spin axis. These "
     "imagers measure neutral atoms entering our solar "
     "system from the outer edge of the heliosphere as "
-    "they move towards the Sun. The ENA imagers collect"
-    " the neutral atoms, sort them by type, and then map "
-    "their incident direction from the outer heliosphere. "
-    "IMAP-Hi uses a time-of-flight (TOF) section to "
-    "identify hydrogen (H) and helium (He) and heavier "
-    "atoms such as carbon (C), nitrogen (N), oxygen (O), "
-    "and neon (Ne). With each spin of the spacecraft, the "
-    "imagers sample swaths in the sky that include the "
-    "ecliptic poles and four additional locations in the "
-    "ecliptic plane. Some low latitude regions, that "
-    "contain ENA emissions from the nose and tail of the "
-    "heliosphere, as well as most of the IBEX Ribbon and "
-    "Belt, are sampled twice within as little as 1.5 "
-    "months which allows it to explore short ENA "
-    "variability."
-    "IMAP-Hi's design and assembly is led by Los Alamos "
-    "National Laboratory (LANL) in collaboration with "
-    "Southwest Research Institute (SwRI), University of "
-    "New Hampshire (UNH), and University of Bern (UBe)."
-    "See https://imap.princeton.edu/instruments/hi for "
-    "more details."
+    "they move towards the Sun. "
+    "See https://imap.princeton.edu/instruments/imap-hi for "
+    "more details. "
 )
 
 hi_base = GlobalInstrumentAttrs(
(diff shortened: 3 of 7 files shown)


imap_processing/cdf/defaults.py
```diff
@@ -26,8 +26,12 @@ class GlobalConstants:
...
     INT_MAXVAL = np.iinfo(np.int64).max
+    FLOAT_MAXVAL = np.finfo(np.float64).max
     DOUBLE_FILLVAL = np.float64(-1.0e31)
     # 1900-01-01T00:00:00
     MIN_EPOCH = -315575942816000000
     # 2100-01-01T00:00:00
     MAX_EPOCH = 3155630469184000000
+    # 32-bits max int value
+    UBIT32_MAXVAL = 2**32 - 1
+    UBIT16_MAXVAL = 2**16 - 1
```
greglucas: I'd suggest either U32, or UINT32 instead of BIT. Also move up to the section with the other ints/floats.
```suggestion
    UINT32_MAXVAL = np.iinfo(np.uint32).max
    UINT16_MAXVAL = np.iinfo(np.uint16).max
```

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -50,87 +33,102 @@
...
     depend_0="epoch",
 )
 
 de_tag_attrs = ScienceAttrs(
-    validmin=GlobalConstants.INT_FILLVAL,
-    validmax=GlobalConstants.INT_MAXVAL,
-    format="I6",
-    label_axis="Direct event time tag",
+    validmin=0,
+    validmax=GlobalConstants.UBIT16_MAXVAL,
```
greglucas: More generically, can we just derive these from the datatype of the variable itself? That way we don't have to specify things here, but can just rely on the code to properly get the min/max values based upon the datatype. `np.iinfo(ds["variable"].dtype).max`

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -50,87 +33,102 @@
...
-    catdesc=(
-        "CCSDS MET. "
+    catdesc="CCSDS mission elapsed time (MET)",
+    var_notes=(
+        "CCSDS mission elapsed time (MET). "
         "It's a 32-bits integer value that represents the "
         "CCSDS Mission Elapsed Time (MET) in seconds."
     ),
-    fieldname="CCSDS MET",
+    fieldname="Mission Elapse Time(MET)",
```
greglucas: spelling: Elapse -> Elapsed

imap_processing/hi/l1a/science_direct_event.py
```diff
@@ -283,13 +285,31 @@ def create_dataset(de_data_list: list, packet_met_time: list) -> xr.Dataset:
         trigger_id, dims="epoch", attrs=hi_cdf_attrs.trigger_id_attrs.output()
     )
     dataset["tof_1"] = xr.DataArray(
-        tof_1, dims="epoch", attrs=hi_cdf_attrs.tof_attrs.output()
+        tof_1,
+        dims="epoch",
+        attrs=dataclasses.replace(
+            hi_cdf_attrs.tof_attrs,
+            fieldname="Time of Flight(TOF) 1",
```
greglucas: Do we want a space between Flight / parentheses?
```suggestion
            fieldname="Time of Flight (TOF) 1",
```

imap_processing/swapi/swapi_cdf_attrs.py
```diff
@@ -46,20 +38,20 @@
...
 counts_attrs = ScienceAttrs(
     validmin=0,
-    validmax=GlobalConstants.INT_MAXVAL,
-    format="I12",
-    units="Counts",
-    label_axis="Counts",
+    validmax=GlobalConstants.UBIT16_MAXVAL,
+    format="I5",
+    units="counts",
+    label_axis="<replaced-in-code>",
```
greglucas, line 46: Can you leave these out of the initializer call, or are they required?

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -50,87 +33,102 @@
...
+    label_axis="Energy Step",
     display_type="time_series",
     catdesc=(
-        "ESA step. "
+        "ESA step (0-10), nominally 9, but possibly 8 or 10,"
+        " 0 is likely a background test"
+    ),
+    var_notes=(
+        "ESA (electrostatic analyzer) step. "
         "It's a 4-bits integer value that represents the "
```
```suggestion
        "It's a 4-bit integer value that represents the "
```

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -8,34 +8,17 @@
 )
 from imap_processing.hi import __version__
 
+# TODO: add more information about dataset
```
greglucas, line 11: I think we allow 88 characters for line lengths. It might be easier to read some of this if we try to take up as much of that linewidth as possible in these descriptions rather than so much line-wrapping and taking up more vertical space.

I've noticed this elsewhere too in our codebase though, so maybe that is just a personal preference on my part.

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -50,87 +33,102 @@
...
 trigger_id_attrs = ScienceAttrs(
-    validmin=GlobalConstants.INT_FILLVAL,
+    validmin=1,
     validmax=3,
     format="I1",
-    label_axis="Trigger ID",
+    label_axis="ID",
     display_type="time_series",
     catdesc=(
+        "Trigger ID of which detector was hit first. 1,2,3 for detector A,B,C resp"
```
greglucas: Just to verify they count 1-3 and not 0-2?
```suggestion
        "Trigger ID of which detector was hit first. 1, 2, or 3 for detector A, B, or C respectively"
```

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -50,87 +33,102 @@
...
+    ),
+    var_notes=(
         "Trigger ID is a 2-bits. It represents the trigger "
```
```suggestion
        "The trigger ID is 2-bits. It represents "
```

imap_processing/cdf/defaults.py
tech3371: oh cool. Didn't know that was option. Thanks!

imap_processing/hi/hi_cdf_attrs.py
tech3371: I don't have strong preference here. I will change these two to be 88 characters.

imap_processing/hi/hi_cdf_attrs.py
tech3371: curious to know why you removed `s`. If so, should I remove for others as well?

imap_processing/hi/hi_cdf_attrs.py
tech3371: Just to be sure, let's say today's data's max value of de_tag value is 10k, then the max of this file data's will be 10k. If next day's data's max value of de_tag is 20k, then the max of that file's data will 20k. Is that right?

imap_processing/hi/hi_cdf_attrs.py
tech3371: I had to remove those spaces and shorten respectively to fit the 80 char limit feedback from Andriy but also have meaningful comment.

imap_processing/swapi/swapi_cdf_attrs.py
tech3371: They are required for data type. It also helps to know which variable I am replacing if I was duplicating the code or on first glance.

imap_processing/cdf/global_attrs.py
```diff
@@ -213,17 +213,26 @@ class AttrBase:
...
+        fieldname and label_axis parameter are closely related.
+        fieldname and label_axis are used in plot. fieldname
+        is used as title of the plot, label_axis is used as axis label.
+        For example, fieldname='Time of flight', label_axis='TOF'
     var_type : str, default="support_data"
-        The type of data
+        The type of data. Valid options are:
+            1. "data" - plotted and listed on CDAWeb and accessed by API,
+            2. "support_data" - not plotted
+            3. "metadata" - reserved for character type variables.
```
bourque, line 228: This is helpful, thanks!

bourque, line 228: What is meant by "character type variables" here?

bourque (approved):
Looks good!

imap_processing/hi/hi_cdf_attrs.py
greglucas: I _think_ this validmax means the global max and shouldn't be changing across files, but I really don't know.

My suggestion was basically to infer the datatype rather than manually input it here if that is possible. It likely isn't worth doing here, was more just a thought on possibilities.

imap_processing/hi/hi_cdf_attrs.py
greglucas: It is a minor gramatical update, it depends on how you are using it. Here I think it is `a 4 bit data value`, but there are `4 bits of information`. So somewhat confusing when to use plural vs singular and I don't think every place should be updated.

imap_processing/hi/hi_cdf_attrs.py
greglucas: Ahh, interesting! Sorry I didn't realize this field was character limited. Would it be OK to just put "Trigger ID of the detector that was hit first" and leave the 1/2/3 text to the following longer text section?

imap_processing/cdf/global_attrs.py
tech3371: I took this from Andriy's feedback. I took that as character type meaning data array with string but we should ask him for clarify.

imap_processing/hi/hi_cdf_attrs.py
tech3371: Oh ok. That's what I was doing with `UINT32_MAXVAL` to set global max since I know that de_tag won't have value great than 16-bits int value. We can discuss that as team and figure out how we want to set global max.

tech3371: I tried to incorporate `dtype` in the science data. Can you let me know what you think or is it what you are thinking?

greglucas: Yes, that is nice :) and it gets me thinking that we should actually check other instruments and datafiles to make sure we are using the correct dtypes, because the default is `int64` and if you only need `uint8` then we are wasting quite a bit of space on disk.

```python
xr.DataArray([0, 1, 2]).dtype  # int64
xr.DataArray(np.asarray([0, 1, 2], np.uint8)).dtype  # uint8
```

greglucas (approved):

imap_processing/hi/hi_cdf_attrs.py
tech3371: Totally agree. We should do that in upcoming PR. I can help with that.

Issue #251, opened 2023-10-10
Validate CoDICE L1A CDFs


bourque:
- [x] #1192 

CDFs to validate:

- [x] `hskp`
- [x] `lo-counters-aggregated`
- [x] `lo-counters-singles`
- [x] `hi-counters-aggregated`
- [x] `hi-counters-singles`
- [x] `hi-priorities`
- [x] `lo-sw-priority`
- [x] `lo-nsw-priority`
- [x] `lo-sw-angular`
- [x] `lo-nsw-angular`
- [x] `lo-sw-species`
- [x] `lo-nsw-species`
- [x] `lo-pha`
- [x] `hi-pha`
- [x] `hi-omni`
- [x] `hi-sectored`
- [x] `lo-ialirt`
- [ ] `hi-ialirt`

PR #530, opened 2024-04-23
CoDICE L1a lo-nsw-species-counts

bourque:
# Change Summary

## Overview

This PR adds support for `lo-nsw-species-counts` data and creates corresponding CDF. It also includes some minor updates on APID names.

## New Files

- `imap_processing/tests/codice/data/lo_fsw_view_6_ccsds.bin`
    - Test data for `lo-nsw-species-counts` APID 

## Updated Files

- `imap_processing/codice/cdf_attrs.py`
    - Added CDF attrs for `lo-nsw-species-counts` data
- `imap_processing/codice/codice_l0.py` 
    - Added mapping to `lo-nsw-species-counts` packet definition
- `imap_processing/codice/codice_l1a.py`
    - Added support for `lo-nsw-species-counts` APID, and some updates to make some functions/methods a bit more generalized 
- `imap_processing/codice/constants.py` 
    - Added CDF-friendly names for `lo-sw-species-counts`
- `imap_processing/codice/packet_definitions/P_COD_LO_NSW_SPECIES_COUNTS.xml`
    - Packet definition now uses variable length data   
- `imap_processing/codice/utils.py`
    - Added/updated some APIDs and their names
- `imap_processing/tests/codice/test_codice_l1a.py`   
    - Added test for `lo-nsw-species-counts` CDF

PR #534, opened 2024-04-24
Initial Codice L1A validation tests

bourque:
# Change Summary

## Overview

This PR adds a new test for validation of CoDICE L1a CDFs. We currently don't have validation data from CoDICE, but this sets up the test to be able to compare data values between the CDFs generated by the pipeline and CoDICE validation files. The validation files I am currently using in these tests are just the CDFs that I generated myself. So not a great test, but the goal here is to just set up the framework so that once CoDICE validation files are acquired, they can just be swapped in. 

Relates to #251 

## New Files

- `imap_processing/tests/codice/data/temp_validation_<descriptor>.cdf`
   - (Temporary) validation files for the currently supported data products: `hskp`, `lo-sw-species-counts`, and `lo-nsw-species-counts`

## Updated Files

    - `process_codice_l1a` function now returns the `xarray` `dataset` in addition to the CDF filename so that the `dataset` can easily be tested
    - Added test that compares data values for validation CDFs and CDFs generated by the pipeline


imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -40,5 +57,32 @@ def test_codice_l1a(test_file: Path, expected_filename: str):
...
+
+    TODO: This tests currently uses validation files that the pipeline itself
+    generates, rather than a CDF file generated by the CoDICE instrument team.
+    Once proper validation files are acquired, this test function should point
+    to those. This function currently just serves as a framework for validating
+    files, but does not actually validate them.
+    """
+
+    # Retrieve the validation file contents
+    validation_data = cdflib.xarray.cdf_to_xarray(validation_file)
```
greglucas: We have a `write_cdf()` helper, this got me to thinking that maybe we should also have a `load_cdf()` helper that basically just calls this function from `imap_processing` instead of needing to remember the nested imports of cdflib? Allowing us to leave `cdflib` as an implementation detail to our end users.

greglucas (changes requested):
I'm happy to be overridden and anyone can dismiss this review, but I personally don't like hosting temporary datasets in git repositories because they grow out of control then.

imap_processing/tests/codice/test_codice_l1a.py
bourque: I like the idea! I can take note of this and add that in a separate PR for the team to review/discuss.

bourque:
@greglucas Do you have any suggestions on how to deal with the temporary validation files? Should I just read them from a local directory?

greglucas:
> @greglucas Do you have any suggestions on how to deal with the temporary validation files? Should I just read them from a local directory?

Yes that is definitely an option and then add a `pytest.skip()` if that isn't found possibly. I guess I'm not sure at the ultimate end goal. Are you anticipating getting CDF files as validation data? I think most of our other instruments have been providing csv or text files, so it isn't as easy as just direct comparisons and we've had to only compare specific variables that were provided in the csv file I believe.

imap_processing/tests/codice/test_codice_l1a.py
greglucas: Yes, definitely a separate PR, thanks for volunteering! ;) Just wanted to write that down while I noticed it here :)

bourque:
> > @greglucas Do you have any suggestions on how to deal with the temporary validation files? Should I just read them from a local directory?
> 
> Yes that is definitely an option and then add a `pytest.skip()` if that isn't found possibly. I guess I'm not sure at the ultimate end goal. Are you anticipating getting CDF files as validation data? I think most of our other instruments have been providing csv or text files, so it isn't as easy as just direct comparisons and we've had to only compare specific variables that were provided in the csv file I believe.

I planned on removing these `temp_validation_*.cdf` files once I acquired some sort of CoDICE validation data (whether it were CDF, CSV, txt, etc.) and updating these tests to compare the arrays to whatever the validation data may be. The goal was really to just set up some sort of test that gathered the data arrays and put them in a place ready to be compared to something. (see 

Perhaps I can remove these temp files, and just mark the tests as `xfail` for now?

This also reminds me of your comment on my last PR (https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/530#discussion_r1576669263) -- I still need to add a test for the shapes and lengths of these data arrays

greglucas:
> Perhaps I can remove these temp files, and just mark the tests as xfail for now?

That seems reasonable to me and addresses my issue of growing the repository size. I'm not sure if `skip` or `xfail` would be better? `@pytest.mark.skip(reason="no way of currently testing this")`

pyproject.toml
```diff
@@ -76,7 +76,8 @@ lint.ignore = ["D104", "PLR2004", "RUF200", "S311"]
 
 [tool.ruff.lint.per-file-ignores]
 # S603 unchecked input in subprocess call is fine in our tests
-"*/tests/*" = ["D", "S101", "S603"]
+# PT006 Wrong type passed to first argument of @pytest.mark.parametrize
+"*/tests/*" = ["D", "S101", "S603", "PT006"]
```
bourque, line 80: I had to add this because I am using a fixture within a parametrized test (I think) ... I couldn't find a way around this error so I decided to just ignore it. Happy to dig further into it if anyone thinks it's an issue.

imap_processing/codice/codice_l1a.py
```diff
@@ -37,6 +35,7 @@
 from imap_processing.utils import group_by_apid, sort_by_time
 
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
```
tech3371, line 38: @greglucas cleaned up logger at one point. Could it be because logging uses `INFO` by default?

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -2,43 +2,143 @@
...
+    packets = decom_packets(request.param)
+    dataset, cdf_filename = process_codice_l1a(packets)
+    return dataset, cdf_filename
+
+
+@pytest.mark.parametrize(
+    "test_l1a_data, expected_filename",
+    list(zip(TEST_PACKETS, EXPECTED_FILENAMES)),
+    indirect=["test_l1a_data"],
+)
```
tech3371, line 63: nice refactoring and I learned something new about `indirect`.

pyproject.toml
tech3371: @greglucas may know about this

tech3371 (approved):
looks good to me!

imap_processing/codice/codice_l1a.py
bourque: I think this can be removed. I put it in when I was messing around with trying to get log messages printed to the screen during a local `pytest` run (I still haven't found a good solution for that 🤷🏻)

tech3371: When I am running test locally, I write print statement and then run the pytest with `-s` parameter. It will print all the print statements. Eg.
```
pytest imap_processing/tests/swapi -s
```

bourque: Do you know of a way to do the same thing only with `logger.info` statements instead of `print` statements? It seems that `logger.info` statements don't get shown when the tests pass, only when they fail.

I found that `pytest -rP` does what I am looking for, though it also includes some information I don't want to see.

imap_processing/codice/codice_l1a.py
```diff
@@ -8,11 +8,9 @@
     from imap_processing.codice.codice_l0 import decom_packets
     from imap_processing.codice.codice_l1a import codice_l1a
     packets = decom_packets(packet_file)
-    cdf_filename = codice_l1a(packets)
+    dataset, cdf_filename = process_codice_l1a(packets)
```
greglucas: No need to change in this PR, I'm just throwing this idea out, but what about adding the filename to the dataset as an attribute? `dataset.cdf_filename = cdf_filename` or `dataset.attrs["filename"] = cdf_filename`
so that later you could just grab it from the dataset without needing to pass tuples around.

imap_processing/codice/codice_l1a.py
```diff
@@ -293,14 +292,16 @@ def process_codice_l1a(packets) -> str:
...
+        ``xarray`` dataset containing the science data and supporting metadata
     cdf_filename : str
         The path to the CDF file that was created
     """
     # Group data by APID and sort by time
     grouped_data = group_by_apid(packets)
 
     for apid in grouped_data.keys():
-        print(f"\nProcessing {CODICEAPID(apid).name} packet")
+        logger.info(f"\nProcessing {CODICEAPID(apid).name} packet")
```
greglucas, line 302: I like this change :)

imap_processing/codice/codice_l1a.py
tech3371: Adding this in our `test/conftest.py` did the trick for me. But you still have to add `-s` to see the log statement. This is from in conversation with ChatGPT.
```
def pytest_configure(config):
    # Set up logging to output to the console at DEBUG level
    logging.basicConfig(level=logging.DEBUG, format='%(asctime)s - %(levelname)s - %(message)s')

    # If you know the specific logger used by your application, set it directly:
    logger.setLevel(logging.INFO)
```

imap_processing/tests/codice/test_codice_l1a.py
greglucas: I agree this is a nice way of doing it! Out of curiosity, if you were to transpose the order of the variables would you need the indirect still? Going from `EXPECTED_ARRAY_SHAPES = ..., EXPECTED_ARRAY_SIZES = ...` to `HSK_EXPECTATIONS = (shape, size, ...)` and parameterize over `HSK, SCI, ...`

pyproject.toml
greglucas: I _think_ this might be because it wants it to be spelled out as a list/tuple of options instead of a string.
https://docs.astral.sh/ruff/rules/pytest-parametrize-names-wrong-type/
But I think the string is readable enough so this works IMO and I have also been caught by this error, so it might be better to just ignore it in our tests.

"test_l1a_data, expected_shape" -> ("test_l1a_data", "expected_shape")

greglucas (approved):
Approving, but make sure you squash the commits so the extra files don't get caught in the history.

imap_processing/codice/codice_l1a.py
bourque: Thanks for sharing, I will try it out!

imap_processing/codice/codice_l1a.py
bourque: I like it!

imap_processing/tests/codice/test_codice_l1a.py
bourque: Interesting idea. I can explore this in a future PR, especially when I go to add more data products.

PR #535, opened 2024-04-24
IMAP Hi Histogram L1A processing

tmplummer:
# Change Summary
Add L1A processing of Hi Hist packets.

## Overview
I am opening this PR before finalizing test coverage in order to get feedback before fully fleshing out test coverage.

## New Dependencies
None

## New Files
   - House the L1A processing specific to the hist packet (SCI_CNT)

## Deleted Files
None

## Updated Files
- imap_processing/hi/hi_cdf_attrs.py
   - add histogram specific cdf attributes
- imap_processing/hi/l1a/hi_l1a.py
   - add code into branch for processing histogram packets
   - add branch for unexpected APIDs with logging
- imap_processing/hi/packet_definitions/hi_packet_definition.xml
   - Remove comment about Acquisition Start Time in description of ESA_STEP. This was an artifact from IBEX.
- imap_processing/test/hi/test_l1a.py
   - preliminary test coverage for L1A processing of histogram packets

## Testing
One unit test added - I plan to refine this test and possibly add more after getting initial feedback

### DIFF
--- a/imap_processing/hi/hi_cdf_attrs.py
+++ b/imap_processing/hi/hi_cdf_attrs.py
@@ -2,6 +2,7 @@
 
 from imap_processing.cdf.defaults import GlobalConstants
 from imap_processing.cdf.global_attrs import (
+    AttrBase,
     GlobalDataLevelAttrs,
     GlobalInstrumentAttrs,
     ScienceAttrs,
@@ -138,13 +139,64 @@ ccsds_met_attrs = ScienceAttrs(
 hi_de_l1a_attrs = GlobalDataLevelAttrs(
     data_type="L1A_DE>Level-1A Direct Event",
     logical_source="imap_hi_l1a_de",
-    logical_source_desc=("IMAP-HI Instrument Level-1A Direct Event Data."),
+    logical_source_desc=("IMAP-Hi Instrument Level-1A Direct Event Data."),
     instrument_base=hi_base,
 )
 
 hi_hk_l1a_attrs = GlobalDataLevelAttrs(
     data_type="L1A_HK>Level-1A Housekeeping",
     logical_source="imap_hi_l1a_hk",
-    logical_source_desc=("IMAP-HI Instrument Level-1A Housekeeping Data."),
+    logical_source_desc=("IMAP-Hi Instrument Level-1A Housekeeping Data."),
     instrument_base=hi_base,
+)
+
+hi_hk_l1a_metadata_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=GlobalConstants.INT_MAXVAL,
+    depend_0="epoch",
+    format="I12",
+    units="int",
+    var_type="support_data",
+    variable_purpose="PRIMARY",
+)
+
+# Histogram attributes
+hi_hist_l1a_global_attrs = GlobalDataLevelAttrs(
+    data_type="L1A_CNT>Level-1A Histogram",
+    logical_source="imap_hi_l1a_hist",
+    logical_source_desc=("IMAP-Hi Instrument Level-1A Histogram Data."),
+    instrument_base=hi_base,
+)
+
+hi_hist_l1a_angle_attrs = AttrBase(
+    validmin=0,
+    validmax=360,
+    format="I3",
+    units="deg",
+    label_axis="ANGLE",
+    display_type="time_series",
+    catdesc="Angle bin centers for Histogram data.",
+    fieldname="ANGLE",
+    fill_val=GlobalConstants.INT_FILLVAL,
+    var_type="support_data",
+)
+
+hi_hist_l1a_esa_step_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=2**4 - 1,
+    depend_0="epoch",
+    format="I2",
+    catdesc="4-bit ESA Step Number",
+    var_type="support_data",
+    variable_purpose="PRIMARY",
+)
+
+hi_hist_l1a_counter_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=2**12 - 1,
+    depend_0="epoch",
+    depend_1="angle",
+    format="I4",
+    var_type="support_data",
+    variable_purpose="PRIMARY",
 )
--- a/imap_processing/hi/l1a/hi_l1a.py
+++ b/imap_processing/hi/l1a/hi_l1a.py
@@ -3,6 +3,7 @@
 import logging
 
 from imap_processing.hi.l0 import decom_hi
+from imap_processing.hi.l1a.histogram import create_dataset as hist_create_dataset
 from imap_processing.hi.l1a.housekeeping import process_housekeeping
 from imap_processing.hi.l1a.science_direct_event import science_direct_event
 from imap_processing.hi.utils import HIAPID
@@ -26,27 +27,29 @@ def hi_l1a(packet_file_path: str):
     """
     unpacked_data = decom_hi.decom_packets(packet_file_path)
 
-    # group data by appId
+    # group data by apid
     grouped_data = group_by_apid(unpacked_data)
 
     # Process science to l1a.
     processed_data = []
     for apid in grouped_data.keys():
         if apid == HIAPID.H45_SCI_CNT:
-            # TODO: Add processing for science count data
-            continue
+            logger.info(
+                "Processing histogram data for [%s] packets", HIAPID.H45_SCI_CNT.name
+            )
+            data = hist_create_dataset(grouped_data[apid])
         elif apid == HIAPID.H45_SCI_DE:
             logger.info(
-                "Processing direct event data for [%s] packets", HIAPID.H45_SCI_CNT.name
+                "Processing direct event data for [%s] packets", HIAPID.H45_SCI_DE.name
             )
 
             data = science_direct_event(grouped_data[apid])
-            processed_data.append(data)
         elif apid == HIAPID.H45_APP_NHK:
             logger.info(
                 "Processing housekeeping data for [%s] packets", HIAPID.H45_APP_NHK.name
             )
             data = process_housekeeping(grouped_data[apid])
-            processed_data.append(data)
-
+        else:
+            raise RuntimeError(f"Encountered unexpected APID [{apid}]")
+        processed_data.append(data)
     return processed_data
--- a/imap_processing/hi/l1a/histogram.py
+++ b/imap_processing/hi/l1a/histogram.py
@@ -1,0 +1,145 @@
+"""Unpack IMAP-Hi histogram data."""
+
+import dataclasses
+
+import numpy as np
+import xarray as xr
+from space_packet_parser.parser import Packet
+
+from imap_processing.cdf.global_attrs import ConstantCoordinates
+from imap_processing.cdf.utils import calc_start_time
+from imap_processing.hi import hi_cdf_attrs
+
+# TODO: Verify that these names are OK for counter variables in the CDF
+# define the names of the 24 counter arrays
+# contained in the histogram packet
+QUALIFIED_COUNTERS = (
+    "qual_ab",
+    "qual_c1c2",
+    "qual_ac1",
+    "qual_bc1",
+    "qual_abc1",
+    "qual_ac1c2",
+    "qual_bc1c2",
+    "qual_abc1c2",
+)
+LONG_COUNTERS = (
+    "long_a",
+    "long_b",
+    "long_c",
+    "long_ab",
+    "long_c1c2",
+    "long_ac1",
+    "long_bc1",
+    "long_abc1",
+    "long_ac1c2",
+    "long_bc1c2",
+    "long_abc1c2",
+)
+TOTAL_COUNTERS = ("total_a", "total_b", "total_c", "fee_de_sent", "fee_de_recd")
+
+
+def create_dataset(packets: list[Packet]) -> xr.Dataset:
+    """Create dataset for a number of Hi Histogram packets.
+
+    Parameters
+    ----------
+    packets : list[Packet]
+        packet list
+
+    Returns
+    -------
+    xr.dataset
+        dataset with all metadata field data in xr.DataArray
+    """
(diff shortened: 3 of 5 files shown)


imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -158,3 +159,54 @@
     var_type="support_data",
     variable_purpose="PRIMARY",
 )
+
+# Histogram attributes
+hi_hist_l1a_global_attrs = GlobalDataLevelAttrs(
+    data_type="L1A_CNT>Level-1A Histogram",
+    logical_source="imap_hi_l1a_hist",
+    logical_source_desc=("IMAP-HI Instrument Level-1A Histogram Data."),
```
greglucas: I think we've been using Hi and Lo rather than all caps for instrument names, but someone else can probably confirm.
```suggestion
    logical_source_desc=("IMAP-Hi Instrument Level-1A Histogram Data."),
```

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -158,3 +159,54 @@
...
+    instrument_base=hi_base,
+)
+
+hi_hist_l1a_angle_attrs = AttrBase(
+    validmin=GlobalConstants.INT_FILLVAL,
+    validmax=GlobalConstants.INT_MAXVAL,
```
greglucas: Are these supposed to be angles in degrees?
```suggestion
    validmin=0,
    validmax=360,
```

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -158,3 +159,54 @@
...
+    format="I9",
+    label_axis="ANGLE",
+    display_type="time_series",
+    catdesc=("Angle bin" "Bin centers for Histogram data."),
```
```suggestion
    catdesc=("Angle of bin centers for Histogram data."),
```

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -158,3 +159,54 @@
...
+    fill_val=GlobalConstants.INT_FILLVAL,
+    var_type="support_data",
+)
+
+hi_hist_l1a_esa_step_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=2**4 - 1,
+    depend_0="epoch",
+    format="I2",
+    units="int",
```
greglucas: Are these unitless quantities? I think that is supposed to be left empty in the units category, not the datatype.
```suggestion
    units="",
```

imap_processing/hi/l1a/hi_l1a.py
```diff
@@ -26,27 +27,31 @@ def hi_l1a(packet_file_path: str):
...
         elif apid == HIAPID.H45_APP_NHK:
             logger.info(
                 "Processing housekeeping data for [%s] packets", HIAPID.H45_APP_NHK.name
             )
-            data = create_dataset(grouped_data[apid])
-            processed_data.append(data)
+            data = hk_create_dataset(grouped_data[apid])
+        else:
+            logger.warning(f"Encountered unexpected APID [{apid}]")
+            continue
```
greglucas: Is this OK to have happen, or should we be raising RunTimeError when this occurs to attempt to abort?

imap_processing/hi/l1a/histogram.py
```diff
@@ -0,0 +1,122 @@
...
+    coords["angle"] = xr.DataArray(
+        np.arange(2, 360, 4),
+        name="angle",
+        dims=["angle"],
+        attrs=hi_cdf_attrs.hi_hist_l1a_angle_attrs.output(),
+    )
+
+    data_vars = dict()
+    data_vars["esa_step"] = xr.DataArray(
+        np.empty(len(packets), np.dtype("uint8")),
```
greglucas: minor nit: I have a slight preference for being explicit with adding a keyword argument to the empty method call and using the builtin `np.uint8` for consistency with other places we've been using dtypes in the codebase.
```suggestion
        np.empty(len(packets), dtype=np.uint8),
```

imap_processing/hi/l1a/histogram.py
```diff
@@ -0,0 +1,122 @@
...
+            fieldname="ACQ_START",
+            label_axis="ACQ_START",
+        ).output(),
+    )
+
+    # This dictionary defines the names and order of the 24, 90 element counter arrays
+    # TODO: This implementation will result in a CDF with variables consisting of the
+    #    counter_def_dict key + tuple string. Need to verify if this is OK with
+    #    instrument team
+    counter_def_dict = {
```
greglucas: I found it hard to parse the nested for loops at first glance, so I am just giving a quick suggestion in case this appeals to you or sparks another idea. But, the way it currently works is fine as well if you like that style better.
```python
qual_list = ["QUAL_" + x for x in ("AB", "C1C2", "AC1", "BC1", "ABC1", "AC1C2", "BC1C2", "ABC1C2")]
long_list = ["LONG_" + ...]
misc_list = ["MISC_" + ...]
for key_counter in qual_list + long_list + misc_list:
    data_vars[key_counter] = ...
```

imap_processing/hi/l1a/histogram.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # unpack the packets data into the Dataset
+    for i_epoch, packet in enumerate(packets):
+        dataset.epoch.data[i_epoch] = calc_start_time(
+            packet.data["CCSDS_MET"].raw_value
+        )
+        dataset.esa_step.data[i_epoch] = packet.data["ESA_STEP"].raw_value
+        dataset.acq_start.data[i_epoch] = packet.data["ACQ_START"].raw_value
+
+        counters = ConstBitStream(bin=packet.data["COUNTERS"].raw_value)
```
greglucas: I've been trying to limit our usage of bitstring when possible for performance reasons. Are we able to convert these values ourselves with native Python functionality? Here is a quick previous example: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/309

imap_processing/hi/packet_definitions/hi_packet_definition.xml
```diff
@@ -123,12 +123,13 @@
 			</xtce:Parameter>
             <!-- SCI_DE end -->
             <!-- SCI_CNT start -->
-            <xtce:Parameter name="ESA_STEP" parameterTypeRef="UINT16" shortDescription="Lower 4-bits represent the ESA Step Number: 4-bit ESA Step Enumeration 12-bit Start of Acquisition Time">
-				<xtce:LongDescription>
-					Lower 4-bits represent the ESA Step Number:
-					4-bit ESA Step Enumeration
-					12-bit Start of Acquisition Time
-				</xtce:LongDescription>
+            <xtce:Parameter name="ACQ_START" parameterTypeRef="UINT12">
```
greglucas: Nice :)

imap_processing/hi/hi_cdf_attrs.py
```diff
@@ -158,3 +159,54 @@
...
+    data_type="L1A_CNT>Level-1A Histogram",
+    logical_source="imap_hi_l1a_hist",
+    logical_source_desc=("IMAP-HI Instrument Level-1A Histogram Data."),
+    instrument_base=hi_base,
+)
+
+hi_hist_l1a_angle_attrs = AttrBase(
+    validmin=GlobalConstants.INT_FILLVAL,
+    validmax=GlobalConstants.INT_MAXVAL,
+    format="I9",
```
tech3371: If we make @greglucas 's suggestion on min and max, then we can update this format too to match that.
```suggestion
    format="I3",
```

imap_processing/hi/l1a/hi_l1a.py
tech3371: Good question. I think we should log it because if we have other good apid that we could process, it would be good to process those. But good point about how to raise this error without aborting all.

imap_processing/hi/hi_cdf_attrs.py
tmplummer: I confirmed that Lo cdf global attributes are not all caps. I will change the other occurrences of "HI" in this file.

imap_processing/hi/hi_cdf_attrs.py
tmplummer: Right. I don't know what the units are for this field. I would think that it is some fraction of a second, but am unsure.
@tech3371 With Vivek leaving/off the project who should I ask for clarification about what this value is?

imap_processing/hi/l1a/hi_l1a.py
tmplummer: Valid point.

imap_processing/hi/hi_cdf_attrs.py
tech3371: Yaxi, yaxi.liu@swri.org

imap_processing/hi/l1a/histogram.py
```diff
@@ -0,0 +1,122 @@
...
+        dims=["epoch"],
+        attrs=dataclasses.replace(
+            hi_cdf_attrs.hi_hist_l1a_acq_start_attrs,
+            catdesc=packets[0].data["ACQ_START"].long_description,
+            fieldname="ACQ_START",
+            label_axis="ACQ_START",
+        ).output(),
+    )
+
+    # This dictionary defines the names and order of the 24, 90 element counter arrays
```
tech3371: This is what I was thinking of doing if it helps. https://github.com/tech3371/imap_processing/blob/hi_l1a/imap_processing/hi/l1a/science_histogram.py

Then You could have counters dimension with those values, `"AB", "C1C2", "AC1", "BC1",...`, angle dimension like you had.

imap_processing/hi/l1a/histogram.py
tech3371: I remember this discussion. But for this case, it seems like bitstring may be the best choice since we need to chunk data into `uint12`. Other ways was not readable or too much workaround.

greglucas: Would this work to get a list of ints from that data? This doesn't seem too bad to me.
```python
data_string = packet.data["COUNTERS"].raw_value
list_of_ints = [int(data_string[i*12:(i+1)*12], 2) for i in range(90*24)]
```

Profiling it is 10x faster:
```
❯ python -m timeit 'import bitstring; x = "10" * 6 * 90 * 24; counters = bitstring.ConstBitStream(bin=x); data = counters.readlist("2160*uint12")'
50 loops, best of 5: 7.66 msec per loop
❯ python -m timeit 'import bitstring; x = "10" * 6 * 90 * 24; counters = bitstring.ConstBitStream(bin=x); data = [int(x[i*12:(i+1)*12], 2) for i in range(90*24)]'
500 loops, best of 5: 742 usec per loop
```

tmplummer: Wow, 10x faster!
I have essentially the same implementation staged to commit. I am working through a few other changes in this PR as well.

tech3371: I am amazed by solutions you guys come up with. Nice!

imap_processing/hi/hi_cdf_attrs.py
tmplummer: I might have resolution on what this field is and what the units should be later today. Other units have been removed.

imap_processing/hi/l1a/hi_l1a.py
tmplummer: I'm not sure how to resolve this. Getting an unexpected APID would indicate a bug in something upstream that didn't filter APIDs correctly or called the wrong processing function for a given L0 file. I don't think that we should plan for this to happen regularly so it seems like raising a RunTimeError would be appropriate in order to fail loudly.

greglucas: I share your concern about a logger warning not being loud enough and possibly going undetected which is why I've been stressing the raise instead of warn. But, I don't have a major preference on that and it is likely not a big deal. You can choose what you feel is appropriate here and we can change things later if we need to, I don't think anyone is blocking over this question.

imap_processing/hi/l1a/histogram.py
```diff
@@ -0,0 +1,155 @@
...
+        # unpack 24 arrays of 90 12-bit unsigned integers
+        counters_binary_data = packet.data["COUNTERS"].raw_value
+        counter_ints = [
+            int(counters_binary_data[i * 12 : (i + 1) * 12], 2) for i in range(90 * 24)
+        ]
+        # populate the dataset with the unpacked integers
+        for i_counter, counter in enumerate(
+            (*QUALIFIED_COUNTERS, *LONG_COUNTERS, *TOTAL_COUNTERS)
+        ):
+            dataset[counter][i_epoch] = counter_ints[
```
greglucas, line 74: Above you used `.data[i_epoch]`, but here you use just `[i_epoch]` is there a difference between these setters?

greglucas (approved):
Nice work!

greglucas:
You do still need to appease pre-commit it looks like before merging.

imap_processing/hi/l1a/histogram.py
tmplummer: Good catch. I have no experience with xarray.DataArray. Both methods of setting values appear to do the same thing but I believe the correct/preferred way is without the `data` attribute. I'll make that change.

tmplummer: Interestingly enough, if one tries to modify a coordinate variable without using the `data` attribute, xarray raises `TypeError: IndexVariable values cannot be modified`. Interestingly, using the `data` attribute has the desired effect with no error.

greglucas: Ahh, so the issue is with updating the coordinates only then? And I suppose it is because you are looping through this only updating one at a time instead of wanting to use a full `assign()` or `assign_coords()` that would reset the entire array. With the `.data` you are accessing the underlying numpy array and then mutating that without touching any of the metadata held within the xarray class.

Good to know, thanks for digging in!

bourque:
@all-contributors please add @subagonsouth for code!

PR #536, opened 2024-04-25
MNT: Remove direct bitstring dependency

greglucas:
# Change Summary

## Overview

We do still bring it in via space packet parser, but we don't directly require it anymore.

~~Opening in draft for now in case @subagonsouth does still need this for Hi processing.~~

## New Dependencies

One less dependency

## Updated Files

- pyproject.toml
   - removed dependency
- docs/conf.py
  - removed intersphinx mapping


greglucas:
This is ready for review now as I think we are able to avoid its use for now still.

bourque (approved):

PR #540, opened 2024-04-29
CoDICE L1a lo-sw-priority-counts

bourque:
# Change Summary

## Overview

This PR adds support for CoDICE `lo-sw-priority-counts` data products in the L1a pipeline.

## New Files

- `imap_processing/tests/codice/data/lo_fsw_view_3_ccsds.bin`
  - Test data containing `lo-sw-priority-counts` CCSDS packet 

## Updated Files

  - Added CDF attrs for `lo-sw-priority-counts` CDFs
  - Added mapping from `lo-sw-priority-counts` CCSDS packet file to XTCE definition
  - Added support for `lo-sw-priority-counts` processing
  - Added CDF-friendly names for `lo-sw-priority-counts` dataset variables
- `imap_processing/codice/packet_definitions/P_COD_LO_PRIORITY_COUNTS.xml`
  - Now using variable length data
  - Added tests for `lo-sw-priority-counts` data


imap_processing/codice/codice_l1a.py
```diff
@@ -116,15 +117,15 @@ def create_science_dataset(self, packets: list) -> xr.Dataset:
         )
 
         # Create a data variable for each species
-        for species_data, name in zip(self.data, self.species_names):
+        for variable_data, name in zip(self.data, self.variable_names):
```
bourque, line 121: I changed the name here as to be more generalized and allow to iterate over priority data in addition to species data.

imap_processing/codice/cdf_attrs.py
```diff
@@ -55,20 +55,29 @@
     instrument_base=codice_base,
 )
 
-l1a_lo_sw_species_attrs = GlobalDataLevelAttrs(
+l1a_lo_sw_species_counts_attrs = GlobalDataLevelAttrs(
```
bourque, line 58: Added "counts" to the name here to be consistent with data product name.

imap_processing/codice/codice_l1a.py
```diff
@@ -116,15 +117,15 @@ def create_science_dataset(self, packets: list) -> xr.Dataset:
...
             varname, fieldname = name
-            species_data_arr = [int(item) for item in species_data]
-            species_data_arr = np.array(species_data_arr).reshape(
+            variable_data_arr = [int(item) for item in variable_data]
+            variable_data_arr = np.array(variable_data_arr).reshape(
```
greglucas, line 124: I forget if I mentioned this as an option before, but I think you might be able to do this casting? Or put the dtype directly in the call `np.array(variable_data, dtype=np.int64)`.
```suggestion
            variable_data_arr = np.array(variable_data).astype(int).reshape(
```

greglucas (approved):

imap_processing/codice/codice_l1a.py
bourque: I have been trying to get this to work but have been running into `OverflowError: Python int too large to convert to C long` errors. Something I need to look into further. I will add a TODO for this and try to tackle it in a future PR.

greglucas: Interesting, currently what do you get from `variable_data_arr.dtype`, my guess is that it might be of `float` type then and not integer if we are getting overflows otherwise. That is definitely curious.

```python
>>> np.array([1e45], dtype=int)
  File "<stdin>", line 1, in <module>
OverflowError: Python int too large to convert to C long
>>> np.array([1e10], dtype=int)
array([10000000000])
>>> np.array([1e45]).dtype
dtype('float64')
```

bourque: I think the issue is that the original form of `variable_data` is a very long string of ones and zeros, and so I need to split that string up into individual integer bits before constructing the array?

```python
In [1]: import numpy as np

In [2]: variable_data = '10100101011001101001000000000000000101011101100000001111000000000001000000000000000011110110000000110000011101000101101100011100010001100111111
   ...: 11011000000100000000000000000000000000000000000000100010110010101101'

In [3]: np.array(variable_data)
Out[3]: 
array('1010010101100110100100000000000000010101110110000000111100000000000100000000000000001111011000000011000001110100010110110001110001000110011111111011000000100000000000000000000000000000000000000100010110010101101',
      dtype='<U211')

In [4]: np.array(variable_data).astype(int)
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
Cell In[4], line 1
----> 1 np.array(variable_data).astype(int)

OverflowError: Python int too large to convert to C long

In [5]: variable_data_arr = [int(item) for item in variable_data]

In [6]: variable_data_arr
Out[6]: 
[1,
 0,
 1,
 0,
 0,
 1,
 0,
 1,
 0,
 1,
 1,
 0,
 0,
 1,
 1,
 0,
 1,
 0,
 0,
 1,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 1,
 0,
 1,
 0,
 1,
 1,
 1,
 0,
 1,
 1,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 1,
 1,
 1,
 1,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 1,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 1,
 1,
 1,
 1,
 0,
 1,
 1,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 1,
 1,
 0,
 0,
 0,
 0,
 0,
 1,
 1,
 1,
 0,
 1,
 0,
 0,
 0,
 1,
 0,
 1,
 1,
 0,
 1,
 1,
 0,
 0,
 0,
 1,
 1,
 1,
 0,
 0,
 0,
 1,
 0,
 0,
 0,
 1,
 1,
 0,
 0,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 0,
 1,
 1,
 0,
 0,
 0,
 0,
 0,
 0,
 1,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 0,
 1,
 0,
 0,
 0,
 1,
 0,
 1,
 1,
 0,
 0,
 1,
 0,
 1,
 0,
 1,
 1,
 0,
 1]

```

greglucas: Ahh, I didn't realize it was still coming in as a string of bits.

Here is the quick way to get that as an array of integers: `np.array(list(data), dtype=int)` (or bool?). Do you want 1s and 0s, or do you want to say that each energy step is a separate integer composed of a few bits ["101", "111", ...]?
`np.array([int(data[i*length:(i+1)*length], 2) for i in range(self.energy_steps)])`

PR #546, opened 2024-05-03
Glows l1a cdf first draft

maxinelasp:
# Change Summary
Added CDF generation for GLOWS L1A. This involved some minor rearranging of attributes within the GLOWS L1A dataclass to facilitate easier generation. ie aligning more closely with the final output, rather than the structures used in documentation. 

## Overview

Added glows_l1a.py for L1A generation. Added GLOWS CDF attributes. Updated some dataclasses. Added testing.

## New Files

- glows_l1a.py
   - method for creating histogram and direct event CDF files

## Updated Files

- glows_l1a_data.py
   - Moved around attributes
- glows_cdf_attrs.py
   - Created CDF attributes for L1a
   - 
## Testing

I added validation tests using generated data to fully test L1a.


imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,394 @@
...
+            # If the direct event is part of a sequence and is not the first,
+            # append it to the last direct event in the list
+            de_by_day[de_day][-1].append(de)
+        else:
+            de_by_day[de_day].append(DirectEventL1A(de))
+
+    return de_by_day
+
+
+def generate_de_dataset(
```
maxinelasp, line 100: I expect this structure to change after feedback from GLOWS, I just wanted a first draft for discussion.

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,190 @@
...
+    "L1A_de>Level-1A direct event",
+    logical_source="imap_glows_l1a_de",
+    logical_source_desc="IMAP Mission GLOWS Direct Event Level-1A Data.",
+    instrument_base=glows_base,
+)
+
+bins_attrs = AttrBase(
+    validmin=0,
+    validmax=70,  # doc says maximum count per bin is 66.7
+    catdesc="Counts of direct events for photon impacts.",
```
tech3371: Andriy said we don’t need `.` for this and fieldname
```suggestion
    catdesc="Counts of direct events for photon impacts",
```

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,190 @@
...
+    fieldname="Counts of direct events.",
```
tech3371:
```suggestion
    fieldname="Counts of direct events",
```

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,190 @@
...
+    format="F7.4",  # Float with 4 digits
```
tech3371: @bryan-harter Is this correct if float data is only 2 digits followed by 4 digits after the decimal?
```suggestion
    format="F2.4",  # Float with 4 digits
```

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,190 @@
...
+    label_axis="Counts",
+)
+
+# TODO Update this
+per_second_attrs = AttrBase(
+    validmin=0,
+    validmax=300,
+    catdesc="Direct events recorded approximately per second",
+    fieldname="List of direct events ",
+    format="F7.4",  # Float with 4 digits
```
tech3371: This too and in other places too if above suggestion is correct.
```suggestion
    format="F3.4",  # Float with 4 digits
```

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,190 @@
...
+
+# TODO Update this
+# TODO rename this
+direct_event_attrs = AttrBase(
+    validmin=0,
+    validmax=300,  # TODO what is a reasonable max
+    catdesc="Direct events recorded approximately per second",
+    fieldname="List of direct events ",
+    format="F7.4",  # Int with 10 digits
+    var_type="data",
```
tech3371, line 91: Do you add `epoch` dimension in the code? Per Slack’s conversation, sounds like we need `epoch` dimension for `var_type=data`.

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,190 @@
...
+    display_type="time_series",
+    label_axis="Counts",
+)
+
+histogram_attrs = ScienceAttrs(
+    validmin=0,
+    validmax=70,  # doc says maximum count per bin is 66.7, value has a 71
```
tech3371, line 98: Does the comment means `validmax` should be 71?

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,190 @@
...
+    display_type="time_series",
+    label_axis="Counts",
+    fill_val=GlobalConstants.INT_FILLVAL,
+    units="counts",
+    var_type="data",
+)
+
+metadata_attrs = ScienceAttrs(
+    depend_0="epoch",
+    validmin=GlobalConstants.INT_FILLVAL,
```
tech3371: Andriy’s seems to not like fillval as min value. He wants us to use some min value instead. May be zero?

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,190 @@
...
+    validmax=GlobalConstants.INT_MAXVAL,
+    display_type="time_series",
+    fill_val=GlobalConstants.INT_FILLVAL,
+    var_type="metadata",
+    format="I6",
```
tech3371: If we use `I6`, we will need to use `validmax` that is also 6-digits.

tech3371 (commented):
I only looked through CDF stuff. I made comments based on feedback Andriy gave me. I struggle with these tedious things too. I will look at GLOWS l1a refactor code later but don’t let me block you if others review and approve. I will look at it offline.

imap_processing/glows/glows_cdf_attrs.py
maxinelasp: This is something I have asked the GLOWS team about

imap_processing/glows/glows_cdf_attrs.py
maxinelasp: Yeah, I do. The generated files look good in SKTEditor.

imap_processing/glows/glows_cdf_attrs.py
maxinelasp: Good call, I will investigate a good value for this. Unfortunately it's used for a bunch of attributes but they're decently similar

imap_processing/glows/glows_cdf_attrs.py
```diff
@@ -0,0 +1,234 @@
...
+    "solar cycles. "
+    "GLOWS design and assembly is led by the Space Research Center, Warsaw, Poland "
+    "(CBK PAN). See https://imap.princeton.edu/instruments/glows for more details."
+)
+
+glows_base = GlobalInstrumentAttrs(
+    __version__,
+    "GLOWS>GLObal Solar Wind Structure",
+    text,
+    "Photons (space)",  # TODO: is this accurate?
```
greglucas: Probably "Imagers (space)" would be my guess? Here are the valid values: https://spdf.gsfc.nasa.gov/istp_guide/gattributes.html#Instrument_type

imap_processing/glows/l0/decom_glows.py
```diff
@@ -27,7 +27,7 @@ class GlowsParams(Enum):
 
 
 def decom_packets(
-    packet_file_path: str,
+    packet_file_path: Path,
```
greglucas, line 30: Nice, I like this :)

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+    -------
+    generated_files: list[Path]
+        List of the paths of the generated CDF files
+    """
+    # TODO: Data version inside file as well?
+    # Create glows L0
+    hist_l0, de_l0 = decom_packets(packet_filepath)
+
+    de_by_day = process_de_l0(de_l0)
+    hists_by_day = dict()
```
greglucas: We've been using `collections.defaultdict(list)` in other places which would save you having to check if it is present below.

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        # Split by IMAP start time
+        # TODO: Should this be MET?
+        hist_day = calc_start_time(hist.SEC).astype("datetime64[D]")
+        if hist_day not in hists_by_day:
+            hists_by_day[hist_day] = []
+        hists_by_day[hist_day].append(hist_l1a)
+
+    # Generate CDF files for each day
+    generated_files = []
+    for _, hist_l1a_list in hists_by_day.items():
```
greglucas:
```suggestion
    for hist_l1a_list in hists_by_day.values():
```

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        dataset = generate_histogram_dataset(hist_l1a_list, data_version)
+        generated_files.append(write_cdf(dataset))
+
+    for _, de_l1a_list in de_by_day.items():
```
greglucas:
```suggestion
    for de_l1a_list in de_by_day.values():
```

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        Dictionary with keys of days and values of lists of DirectEventL1A objects.
+        Each day has one CDF file associated with it.
+    """
+    de_by_day = dict()
+
+    for de in de_l0:
+        de_day = calc_start_time(de.MET).astype("datetime64[D]")
+        if de_day not in de_by_day:
+            de_by_day[de_day] = [DirectEventL1A(de)]
+        elif de.SEQ != 0:
```
greglucas, line 90: Does this need to be sorted at some point in the future to make sure we are getting a sorted sequence? I think we'll be getting everything sorted by time, but something worth thinking about.

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        "housekeeping_valid": [],
+        "is_pps_autogenerated": [],
+        "hv_test_in_progress": [],
+        "pulse_test_in_progress": [],
+        "memory_error_detected": [],
+    }
+
+    for index, de in enumerate(de_l1a_list):
+        # Set the timestamp to the first timestamp of the direct event list
+        epoch_time = calc_start_time(de.l0.MET).astype("datetime64[ns]")
```
greglucas, line 164: I think others are doing this casting to nanosecond datetime64 too. I wonder if we should move this into the return of `calc_start_time` directly 🤔  No need to change here, but something to consider later.

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        Dataset containing the GLOWS L1A direct event CDF output
+    """
+    # TODO: Block header per second, or global attribute?
+
+    time_data = np.zeros(len(de_l1a_list), dtype="datetime64[ns]")
+    # TODO: Should each timestamp point to a list of direct events, each with a
+    #  timestamp? Or should the list be split out to make the timestamps?
+
+    # Each DirectEventL1A class covers 1 second of direct events data
+    direct_events = np.zeros((len(de_l1a_list), len(de_l1a_list[0].direct_events), 4))
```
greglucas, line 125: Could you add a note about what shape you are expecting this to be? `(npackets, max(nevents/packet), 4)` What is 4 as well?

Is the first dimension npackets, and if so, could we flatten that dimension and make this shape `(nevents, 4)`? I find the ragged possibility of the arrays a bit hard to follow. How do we know that there was fill values in earlier packets after you've expanded it later? I might just not be following everything though.

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+
+        direct_events[index, : len(de.direct_events), :] = new_de
+        time_data[index] = epoch_time
+
+        # support_data["flight_software_version"].append(
+        # str(de.l0.ccsds_header.VERSION))
+        support_data["seq_count_in_pkts_file"].append(
+            int(de.l0.ccsds_header.SRC_SEQ_CTR)
+        )
+        support_data["number_of_de_packets"].append(int(de.l0.LEN))
```
greglucas, line 198: Doesn't space_packet_parser already return this as a uint, so no need to cast? https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/2f3ba40f68de2d745e87ec51509cfed39a86bbba/imap_processing/glows/packet_definitions/GLX_COMBINED.xml#L180

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+    epoch_time = xr.DataArray(
+        time_data,
+        name="epoch",
+        dims=["epoch"],
+        attrs=ConstantCoordinates.EPOCH,
+    )
+
+    direct_event = xr.DataArray(
+        np.arange(4),
+        name="direct_event",
```
greglucas, line 214: Is this like an enum, a possibility of 0, 1, 2, 3 labeled direct event in the packet?

This is kind of opposite of the ENA instruments because they are saying "An event happened at this exact time", whereas here they're saying "this many events happened over this time range" I think?

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        dims=["direct_event"],
+        attrs=glows_cdf_attrs.direct_event_attrs.output(),
+    )
+
+    # TODO come up with a better name
+    per_second = xr.DataArray(
+        np.arange(direct_events.shape[1]),
```
greglucas, line 221: Is this supposed to be the bin centers of a histogram? So you want to say this many events were in this bin?

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        hist_data[index] = hist.histograms
+
+        support_data["flags_set_onboard"].append(hist.flags["flags_set_onboard"])
+        support_data["is_generated_on_ground"].append(
+            int(hist.flags["is_generated_on_ground"])
+        )
+
+        for key in support_data.keys():
+            if key not in ["flags_set_onboard", "is_generated_on_ground"]:
+                support_data[key].append(hist.__getattribute__(key))
```
greglucas, line 345: I'd suggest avoiding calling dunder methods directly if we don't need to. Are you able to use `getattr(hist, key)` instead?

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        name="bins",
+        dims=["bins"],
+        attrs=glows_cdf_attrs.bins_attrs.output(),
+    )
+
+    hist = xr.DataArray(
+        hist_data,
+        name="histograms",
+        dims=["epoch", "bins"],
+        coords={"epoch": epoch_time, "bins": bins},
```
greglucas, line 370: I don't think you need to pass in the `coords` directly to the DataArrays if you specify the coords at the Dataset level. As long as the dims are pointing to a coordinate array in the dataset (which you have done below) it should pick up that mapping.

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -200,14 +242,27 @@ def __init__(self, level0: HistogramL0):
...
+        # In L1a, these are left as unit encoded values.
+        self.number_of_spins_per_block = l0.SPINS
+        self.number_of_bins_per_histogram = l0.NBINS
+        self.number_of_events = l0.EVENTS
+        self.filter_temperature_average = l0.TEMPAVG
+        self.filter_temperature_variance = l0.TEMPVAR
+        self.hv_voltage_average = l0.HVAVG
+        self.hv_voltage_variance = l0.HVVAR
+        self.spin_period_average = l0.SPAVG
+        self.spin_period_variance = l0.SPAVG
```
greglucas: Duplicate entry, is there a VAR here too?
```suggestion
        self.spin_period_average = l0.SPAVG
        self.spin_period_variance = l0.SPVAR
```

imap_processing/glows/utils/constants.py
```diff
@@ -72,3 +76,12 @@ class DirectEvent:
     timestamp: TimeTuple
     impulse_length: int
     multi_event: bool = False
+
+    def to_array(self):
```
greglucas: nit, but I'd call this `to_list()` to be explicit that you aren't a numpy array just yet.

imap_processing/tests/glows/test_glows_l1a_data.py
```diff
@@ -22,6 +25,22 @@ def decom_test_data():
...
+    hist_l1a = []
+
+    for hist in decom_test_data[0]:
+        hist_l1a.append(HistogramL1A(hist))
+
+    de_l1a_dict = process_de_l0(decom_test_data[1])
+
+    de_l1a = []
+    for _, value in de_l1a_dict.items():
+        de_l1a += value
```
greglucas:
```suggestion
    de_l1a = list(de_l1a_dict.values())
```

imap_processing/tests/glows/test_glows_l1a_data.py
```diff
@@ -361,3 +391,120 @@ def test_generate_status_data():
...
+def test_expected_results(l1a_test_data):
+    _, de_data = l1a_test_data
+
+    # Validation data is generated from the code sent over by GLOWS team. Contains the
+    # first 20 packets
+    validation_data = pd.read_csv(
+        Path(__file__).parent / "direct_events_validation_data_l1a.csv"
+    )
+
+    for index in validation_data.index:
```
greglucas, line 407: It might be good to add an assertion that `len(validation_data) == 20` before this loop to make sure we are actually iterating through something. Otherwise we could just pass over this loop without any assertions and still pass the test if validation_data was empty.

imap_processing/tests/glows/test_glows_l1a_data.py
```diff
@@ -361,3 +391,120 @@ def test_generate_status_data():
...
+            == validation_data["pulse_test_in_progress"][index]
+        )
+        assert (
+            de.status_data.memory_error_detected
+            == validation_data["memory_error_detected"][index]
+        )
+
+        assert de.l0.LEN == validation_data["number_of_de_packets"][index]
+
+        validation_de = ast.literal_eval(validation_data["de_data"][index])
```
greglucas: Nice use of `literal_eval`!
Could you bring it up into the original read_csv as a converter? `read_csv(*, converters={"de_data": ast.literal_eval})`

imap_processing/tests/glows/test_glows_l1a_data.py
```diff
@@ -361,3 +391,120 @@ def test_generate_status_data():
...
+            == validation_de[0]
+        )
+        assert (
+            de.direct_events[
+                validation_data["de_data_counter"][index]
+            ].timestamp.subseconds
+            == validation_de[1]
+        )
+        assert (
+            de.direct_events[validation_data["de_data_counter"][index]].impulse_length
```
greglucas, line 504: Is this where the 4 is coming from earlier that I commented on? Rather than your Dataset coordinates being `np.arange(4)`, could you make it `["timestamp seconds", "timestamp subseconds", "impulse length", "multi event"]`? Storing the strings as the coordinate variable would make the enum more explicit and easier to follow I think.

imap_processing/glows/glows_cdf_attrs.py
maxinelasp: The description text says specifically it's non-imaging... but from that list, I don't know what else would fit better.

imap_processing/tests/glows/test_glows_l1a_data.py
maxinelasp: I did not realize this was something read_csv could do, nice!

maxinelasp: Also, a lazy use of literal_eval instead of adding new columns 😆

imap_processing/tests/glows/test_glows_l1a_data.py
maxinelasp: I don't think I know your prior comment, but yes, I can update the dataset coordinates. I didn't realize they could be strings!

maxinelasp: That actually appears to crash cdflib with an error that I have been discussing with Bryan. So I'm going to leave it as is and make a comment, and maybe go back if that error is a bug

greglucas: :( sigh, I really dislike CDF, netCDF would make this much easier.

maxinelasp: On the bright side, using CDF files builds character!

tech3371: Coordinates or data could be strings, but we need to use make sure data type of CDF attrs are setup properly. You could use `StringAttrs` class from global_attrs.py

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -249,8 +311,7 @@ class DirectEventL1A:
     """
 
     l0: DirectEventL0
-    block_header: dict
```
tech3371, line 252: since you removed this dictionary, do you need to update some of the comments/doc string or are they still accurate?

imap_processing/glows/l1/glows_l1a_data.py
```diff
@@ -157,40 +164,75 @@ class HistogramL1A:
...
+    filter_temperature_variance: int
+        Variance of filter temperature in the block, from L0.TEMPVAR. Uint encoded.
+    hv_voltage_average: int
+        Average HV voltage in the block, from L0.HVAVG. Uint encoded.
+    hv_voltage_variance: int
+        Variance of HV voltage in the block, from L0.HVVAR. Uint encoded.
+    spin_period_average: int
+        Average spin period in the block, from L0.SPAVG. Uint encoded.
+    spin_period_variance: int
+        Variance of spin period in the block, from L0.SPAVG. Uint encoded.
```
tech3371: similar to Greg's comment
```suggestion
        Variance of spin period in the block, from L0.SPVAR. Uint encoded.
```

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+
+from imap_processing.cdf.global_attrs import ConstantCoordinates
+from imap_processing.cdf.utils import calc_start_time, write_cdf
+from imap_processing.glows import __version__, glows_cdf_attrs
+from imap_processing.glows.l0.decom_glows import decom_packets
+from imap_processing.glows.l0.glows_l0_data import DirectEventL0
+from imap_processing.glows.l1.glows_l1a_data import DirectEventL1A, HistogramL1A
+
+
+def glows_l1a(packet_filepath: Path, data_version: str) -> list[Path]:
```
tech3371, line 18: can you import data `data_version` using import or is there reason for passing it as argument?

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+        support_data["flags_set_onboard"].append(hist.flags["flags_set_onboard"])
+        support_data["is_generated_on_ground"].append(
+            int(hist.flags["is_generated_on_ground"])
+        )
+
+        for key in support_data.keys():
+            if key not in ["flags_set_onboard", "is_generated_on_ground"]:
+                support_data[key].append(hist.__getattribute__(key))
+        for key in time_metadata.keys():
+            time_metadata[key].append(hist.__getattribute__(key).to_seconds())
```
tech3371: Can you add some comments to help follow this code?

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+
+    epoch_time = xr.DataArray(
+        time_data,
+        name="epoch",
+        dims=["epoch"],
+        attrs=ConstantCoordinates.EPOCH,
+    )
+
+    bins = xr.DataArray(
+        np.arange(3600),  # TODO: Is it always 3600 bins?
```
tech3371: I was wondering what 3600 was in above line. cool. can you refactor this number into variable name? Also, is this angle bins?

imap_processing/glows/l1/glows_l1a.py
tech3371: Agree. If you move line 371 up above these, then you won't need these coordinates.

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+
+    time_data = np.zeros(len(de_l1a_list), dtype="datetime64[ns]")
+    # TODO: Should each timestamp point to a list of direct events, each with a
+    #  timestamp? Or should the list be split out to make the timestamps?
+
+    # Each DirectEventL1A class covers 1 second of direct events data
+    direct_events = np.zeros((len(de_l1a_list), len(de_l1a_list[0].direct_events), 4))
+    global_attributes = glows_cdf_attrs.glows_l1a_de_attrs.output()
+
+    global_attributes["ground_software_version"] = __version__
```
tech3371, line 128: My first question was related to this. I was wondering if you could do similar to this line and avoid having to pass data_version parameter.

imap_processing/glows/l1/glows_l1a.py
tech3371: I thought this line in `calc_start_time` already does it but if not, I agree. we should that conversion in there as well.
```
    time_delta = np.timedelta64(int(shcoarse_time * 1e9), "ns")
```

imap_processing/glows/l1/glows_l1a.py
```diff
@@ -0,0 +1,405 @@
...
+                    ),
+                    (0, de_len - direct_events.shape[1]),
+                    (0, 0),
+                ),
+                "constant",
+                constant_values=(0,),
+            )
+        new_de = np.array([event.to_list() for event in de.direct_events])
+
+        direct_events[index, : len(de.direct_events), :] = new_de
```
tech3371: Can you explain/add comment about what this is doing?

tech3371 (commented):
I looked at the l1a processing code. I think I understand most of it on very high level. Some logics are more complicated than I can understand. I hope GLOWS team can give you better feedback. Let me know if you like me to look at anything.

imap_processing/glows/l1/glows_l1a.py
maxinelasp: This is a different version number to data_version. Data_version needs to be passed in because it is only known at run time - it's the version number of the file generated.

maxinelasp:
Mentioned this in person, but just for posterity: this PR is my first draft good faith attempt at GLOWS CDF files, and has a lot of open questions that need to be answered by the GLOWS team before it is complete. The main goal is just to get CDF files as outputs so development can continue.

imap_processing/tests/glows/test_glows_l1a_cdf.py
```diff
@@ -0,0 +1,82 @@
...
+    packet_path = current_directory / "glows_test_packet_20110921_v01.pkts"
+    histograms_l0, de_l0 = decom_glows.decom_packets(packet_path)
+
+    histograms_l1a = [HistogramL1A(hist) for hist in histograms_l0]
+
+    de_l1a_dict = process_de_l0(de_l0)
+
+    de_l1a = []
+    for _, value in de_l1a_dict.items():
+        de_l1a += value
```
greglucas:
```suggestion
    de_l1a = [value for value in de_l1a_dict.values()]
```

imap_processing/tests/glows/test_glows_l1a_cdf.py
```diff
@@ -0,0 +1,82 @@
...
+        elif key == "flags":
+            assert dataset["flags_set_onboard"].data[0] == item["flags_set_onboard"]
+            assert (
+                dataset["is_generated_on_ground"].data[0]
+                == item["is_generated_on_ground"]
+            )
+        elif key not in ["histograms", "ground_software_version", "pkts_file_name"]:
+            assert dataset[key].data[0] == item
+
+    assert (dataset["histograms"].data[-1] == histograms_l1a[-1].histograms).all()
```
greglucas: You are testing the first one right away in this and the last one here, do you want to test them all?
```suggestion
    for i in range(len(dataset["histograms"].data):
        assert (dataset["histograms"].data[i] == histograms_l1a[i].histograms).all()
```

imap_processing/tests/glows/test_glows_l1a_cdf.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_generate_de_dataset(l1a_data):
+    _, de_l1a = l1a_data
+    dataset = generate_de_dataset(de_l1a, "v001")
+    assert len(dataset["epoch"].values) == len(de_l1a)
+
+    # TODO fix this test
```
greglucas: Remove this if it is currently working, or indicate what you mean by "fix" it for when we come back to it later and forget.

imap_processing/tests/glows/test_glows_l1a_data.py
```diff
@@ -22,6 +25,22 @@ def decom_test_data():
...
+    hist_l1a = []
+
+    for hist in decom_test_data[0]:
+        hist_l1a.append(HistogramL1A(hist))
+
+    de_l1a_dict = process_de_l0(decom_test_data[1])
+
+    de_l1a = []
+    for value in de_l1a_dict.values():
+        de_l1a += value
```
greglucas: Minor nit to make this similar to your other test code block
```suggestion
    hist_l1a = [HistogramL1A(hist) for hist in decom_test_data[0]]

    de_l1a_dict = process_de_l0(decom_test_data[1])

    de_l1a = [value for value in de_l1a_dict.values()]
```

greglucas (approved):
Very minor nits. Looks good to me though and I like all the comments you added in to clarify later and make sure we follow up on things!

PR #364, opened 2024-03-18
hi direct event l1a processing

tech3371:
# Change Summary
TODO: will update this description tomorrow.

## New Files

- new file 1
   - description of new file 1's purpose

## Deleted Files

- deleted file 1
   - explanation for why file was deleted

## Updated Files

- updated file 1
   - description of change 1 in file 1
   - description of change 2 in file 2
- updated file 2
   - descipriton of change 1 in file 2

PR #549, opened 2024-05-07
WIP - Define L0 data class attributes for HIT science packet

### LINKED ISSUE #512: Create Science Packet data classes
### Summary

The Science packet must be decommutated and written to one, or multiple data classes. There's a couple options I'm thinking (I'm sure there are other options as well):

Option 1:
- decom and decompress the science packet data and write to a single L0 data class
- split out the decompressed data in the L0 data class to two separate L1A data classes that will correspond to the two L1A science products (Count rates and Pulse Height Event)

Option 2:
- Decom and decompress the science packet data, then immediately split out and write the data to the two L1A data classes.

I think L1A also includes converting the data to engineering units, so it might be better to write decom, decompress, and write the raw values to an L0 data class, then convert to engineering units when writing the data to L1A data classes.

vmartinez-cu:
# Change Summary
Defined L0 data class attributes for HIT science packets

## Overview
Created an L0 data class for the HIT science packet and defined required attributes. Some attributes are `InitVar` attributes that will hold binary data that need to be decommutated using methods that haven't been written yet. 

## New Dependencies
xlrd was needed to generate the xtce file so I added this to the project dependencies. This addition is not the
result of new code from this PR.

## New Files
- science_packet.py
   - L0 data class for the HIT science packet. Contains some InitVars attributes that will only be used for initialization
- P_HIT_SCIENCE.xml
   - xtce file for the HIT science packet
- TLM_HIT_modified.xls
   - Telemetry packet definitions. This file was modified to consolidate repeated fields as bytes for easier 
      parsing and processing using code

## Deleted Files
None

## Updated Files
The following files needed to be updated to add the xlrd dependency required for running xtce_generator_hit.py
- poetry.lock
- pyproject.toml

## Testing
None


pyproject.toml
```diff
@@ -47,6 +47,7 @@ pytest-cov = {version="^4.0.0", optional=true}
 ruff = {version="==0.2.1", optional=true}
 sphinx = {version="*", optional=true}
 sphinxcontrib-openapi = {version="^0.8.3", optional=true}
+xlrd = "^2.0.1"
```
greglucas: In the tools section below we have `openpyxl` added already. Would you be able to use that package instead by installing the package with the tools extras? If you do need this, it should be added as an optional dependency and then added to the tools extras because it isn't required during actual processing.
```suggestion
```

tools/xtce_generation/TLM_HIT_modified.xls
greglucas, line 1: I see that we have added some `.xlsx` files, but left others out, so I'm not sure what our typical workflow is for this, hopefully someone else can chime in on what our standard is. My preference would be to remove this file and only add the generated `.xml` file to not grow the repo size too much.

greglucas (commented):
Are you using the HIT XTCE generator code to produce the new packet file? I don't see the new `xlrd` dependency used anywhere, so I'm curious if you need to add an extra Python file/script in the tools still.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/tools/xtce_generation/xtce_generator_hit.py

Do you have any sample data to add a test for this new code?

imap_processing/hit/packet_definitions/P_HIT_SCIENCE.xml
```diff
@@ -0,0 +1,371 @@
...
+			<xtce:Parameter name="NFORMAT" parameterTypeRef="UINT16" shortDescription="nothing currently goes in this slot">
+				<xtce:LongDescription>events handled b the telemetry event formatter</xtce:LongDescription>
+			</xtce:Parameter>
+			<xtce:Parameter name="NASIDE" parameterTypeRef="UINT16" shortDescription="A-side events">
+				<xtce:LongDescription>events the software assumed were A-side events</xtce:LongDescription>
+			</xtce:Parameter>
+			<xtce:Parameter name="NBSIDE" parameterTypeRef="UINT16" shortDescription="B-side events">
+				<xtce:LongDescription>events the software assumed were B-side events</xtce:LongDescription>
+			</xtce:Parameter>
+			<xtce:Parameter name="NERROR" parameterTypeRef="UINT16" shortDescription="events that caused a processing error - should never happen">
```
maxinelasp, line 256: I don't know if this is a generator problem or not, but it really bothers me that this file mixes parameter definitions for short and long descriptions. I'd prefer to see: 

```
<xtce:ShortDescription>events the software assumed were B-side events</xtce:ShortDescription>
<xtce:LongDescription>events the software assumed were B-side events</xtce:LongDescription>
```

tools/xtce_generation/TLM_HIT_modified.xls
maxinelasp: I prefer to leave only `.xml` and have those be the source of truth

pyproject.toml
maxinelasp: +1

maxinelasp (approved):
Overall looks good, I like the dataclass, but a few comments. Also, they're ok, but I would prefer to see some more descriptive documentation for the attributes where it makes sense. That depends heavily on what is important for later processing though, and most of them are good now.

pyproject.toml
vmartinez-cu: It appears that the xlrd dependency is needed if the TLM packet definition file is in .xls format. I changed the format to .xlsx and this dependency is no longer necessary. I will remove it from pyproject.toml.

We do have sample data available for this so that will be next to work on for this new code. Probably in a new PR.

tools/xtce_generation/TLM_HIT_modified.xls
vmartinez-cu: Agreed, I'll remove the .xlsx file

imap_processing/hit/packet_definitions/P_HIT_SCIENCE.xml
vmartinez-cu: Good catch. The [HIT housekeeping xml](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/hit/packet_definitions/P_HIT_HSKP.xml) uses the structure you prefer. @sdhoyt, do you recall running into any formatting issues with the short and long descriptions when generating the housekeeping xml?

sdhoyt: I don't think I ran into any issues with housekeeping. What's weird is the text in the long description tag is the text in the short description column in the spreadsheet. The long description in the spreadsheet is EVPRATES_14 (which is weirdly shorter). Seems like something weird might be going on with the generator. Maybe try regenerating and add some print statements in the generator code where it grabs the short and long descriptions to see what's going on.


Update: I was looking at an old spreadsheet with the same name. the long description in the xtce does match the description in the spreadsheet

sdhoyt (approved):

imap_processing/hit/packet_definitions/P_HIT_SCIENCE.xml
vmartinez-cu: It looks like the telemetry_generator.py code is setup to write the short description as a parameter attribute and the long description as a sub element.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/440ed182c7fd269e026275f8c6b2fd5a66e988fd/tools/xtce_generation/telemetry_generator.py#L271-L276

I could update this part of the code as follows to get both descriptions as sub elements. Thoughts?

```python
            # Add descriptions if they exist
            if pd.notna(row.get("shortDescription")):
                description = Et.SubElement(parameter, "xtce:ShortDescription")
                description.text = row.get("shortDescription")
            if pd.notna(row.get("longDescription")):
                description.text = row.get("longDescription")
```

maxinelasp: If you wouldn't mind, I would really prefer them to match. Seems like it's a really easy change

vmartinez-cu: Oh, it looks like the generator was setup the way I suggested above and was changed in March https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/364

@tech3371 Do you recall the motivation for this update? I don't want to change it back if there was need for this structure in the xml files.

tools/xtce_generation/TLM_HIT_modified.xls
tech3371: No standard yet. I have added couples lately but only kept definitions tab that we use in science processing for reference.

imap_processing/hit/packet_definitions/P_HIT_SCIENCE.xml
tech3371: This update was made because the way we were doing before was wrong according to XTCE document. Short description should be in the parameter tab. This was updated when Gavin was implementing feature of getting short and long description from unpacked data. He spotted this bug.

tech3371: We all will need to update our XTCE definition if we plan to use short and long description from unpacked data. It shouldn't break unpacking data if we leave it as it is. We won't see short description I believe.

tech3371 (approved):
Looks good to me! I will send a Slack message about short description update since that everyone is impacted by that change. My bad on bad communication on this update.

imap_processing/hit/packet_definitions/P_HIT_SCIENCE.xml
vmartinez-cu: This is helpful background, thanks for sharing! I will leave telemetry_generator.py as is

PR #550, opened 2024-05-07
Refactor imap_processing.cli.ProcessInstrument

tmplummer:
# Change Summary
Refactor `ProcessInstrument` base class
Add Hi L1a processing to cli Hi class
Add test coverage for cli.py

### Refactor `ProcessInstrument`
I modified the ProcessInstrument.process method to be a `final` method which cannot be overridden in inheriting classes. The method calls three methods that can be overridden: `pre_processing`, `do_processing`, and `post_processing`. 
- `pre_processing` is defined so that default behavior is to download dependencies
- `do_processing` is an abstractmethod that must be defined in any inheriting class
- `post_processing` is defined so that default behavior is to upload any results from `do_processing` to the SDC if the `--upload-to-sdc` argument is set to True.
### Add Hi L1a processing
After the refactor I added Hi L1a processing using the new methods in the base class.

## New Files
- tests/test_cli.py
   - Added some test coverage for the cli.py module

## Updated Files
- imap_processing/cli.py
   - refactor ProcessInstrument class
   - add L1a processing to Hi class
   - modify other instrument classes to use refactored ProcessInstrument class

## Testing
Added some test coverage for cli.py. I am interested in getting feedback on whether my `@mock.patch` and `@pytest.mark.parameterize` use is up to standards/best practices that you all use.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -249,7 +249,7 @@ class ProcessInstrument(ABC):
 
         Attributes
         ----------
-        products: list[str]
+        products : list[str]
         A list of file paths to upload to the SDC.
         """
         if self.upload_to_sdc:
@@ -278,9 +278,9 @@ class ProcessInstrument(ABC):
         """
         Complete pre-processing.
 
-        For this baseclass, pre-processing consists of downloading
-        dependencies for processing. Child classes can override this
-        method to customize the pre-processing actions.
+        For this baseclass, pre-processing consists of downloading dependencies
+        for processing. Child classes can override this method to customize the
+        pre-processing actions.
 
         Returns
         -------
@@ -299,12 +299,13 @@ class ProcessInstrument(ABC):
 
         Attributes
         ----------
-        dependencies: list
-        List of dependencies to process
+        dependencies : list
+            List of dependencies to process
 
         Returns
         -------
-        List of products produced
+        list
+            List of products produced
         """
         raise NotImplementedError
 
@@ -318,8 +319,8 @@ class ProcessInstrument(ABC):
 
         Attributes
         ----------
-        products: list[str]
-        A list of file paths (products) produced by do_processing method.
+        products : list[str]
+            A list of file paths (products) produced by do_processing method.
         """
         self.upload_products(products)
 
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -16,6 +16,7 @@ import sys
 from abc import ABC, abstractmethod
 from json import loads
 from pathlib import Path
+from typing import final
 from urllib.error import HTTPError
 
 import imap_data_access
@@ -31,6 +32,7 @@ import imap_processing
 # In code:
 #   call cdf.utils.write_cdf
 from imap_processing.cdf.utils import load_cdf, write_cdf
+from imap_processing.hi.l1a import hi_l1a
 from imap_processing.mag.l1a.mag_l1a import mag_l1a
 from imap_processing.swe.l1a.swe_l1a import swe_l1a
 from imap_processing.swe.l1b.swe_l1b import swe_l1b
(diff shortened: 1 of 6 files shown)


imap_processing/cli.py
```diff
@@ -241,67 +243,159 @@ def download_dependencies(self):
...
+        A list of file paths to upload to the SDC.
+        """
+        if self.upload_to_sdc:
+            if len(products) == 0:
+                logger.info("No files to upload.")
+            for filename in products:
+                logger.info(f"Uploading file: {filename}")
+                imap_data_access.upload(filename)
+
+    @final
```
bourque, line 262: Cool, this is the first time i've seen this used and I learned something!

imap_processing/cli.py
```diff
@@ -241,67 +243,159 @@ def download_dependencies(self):
...
         return file_list
 
-    @abstractmethod
+    def upload_products(self, products: list[str]):
+        """
+        Upload data products to the IMAP SDC.
+
+        Attributes
+        ----------
+        products: list[str]
```
```suggestion
        products : list[str]
```

Super nit-picky but I think `numpydoc` conventions expect a space on either side of the colon.

imap_processing/cli.py
```diff
@@ -241,67 +243,159 @@ def download_dependencies(self):
...
+        A list of file paths to upload to the SDC.
+        """
+        if self.upload_to_sdc:
+            if len(products) == 0:
+                logger.info("No files to upload.")
+            for filename in products:
+                logger.info(f"Uploading file: {filename}")
+                imap_data_access.upload(filename)
```
bourque, line 260: Should we throw an error or warning here if the `upload_to_sdc` attribute is `False?`

imap_processing/cli.py
```diff
@@ -241,67 +243,159 @@ def download_dependencies(self):
...
+        Complete pre-processing.
+
+        For this baseclass, pre-processing consists of downloading
+        dependencies for processing. Child classes can override this
+        method to customize the pre-processing actions.
+
+        Returns
+        -------
+        List of dependencies downloaded from the IMAP SDC.
+        """
```
bourque: Also super nit-picky but I think we should strive for consistency in using up the 80-character limit in our docstrings

```suggestion
        """
        Complete pre-processing.

        For this baseclass, pre-processing consists of downloading dependencies
        for processing. Child classes can override this method to customize the
        pre-processing actions.

        Returns
        -------
        List of dependencies downloaded from the IMAP SDC.
        """
```

imap_processing/cli.py
```diff
@@ -241,67 +243,159 @@ def download_dependencies(self):
...
+        Abstract method that processes the IMAP processing steps.
+
+        All child classes must implement this method. Input and
+        outputs are typically lists of file paths but are free to
+        any list.
+
+        Attributes
+        ----------
+        dependencies: list
+        List of dependencies to process
```
```suggestion
        dependencies : list
            List of dependencies to process
```

imap_processing/cli.py
```diff
@@ -241,67 +243,159 @@ def download_dependencies(self):
...
+
+        Returns
+        -------
+        List of products produced
```
```suggestion
        Returns
        -------
        list
            List of products produced
```

imap_processing/cli.py
```diff
@@ -241,67 +243,159 @@ def download_dependencies(self):
...
+        Complete post-processing.
+
+        Default post-processing consists of uploading newly generated
+        products to the IMAP SDC. Child classes can override this method
+        to customize the post-processing actions.
+
+        Attributes
+        ----------
+        products: list[str]
+        A list of file paths (products) produced by do_processing method.
```
```suggestion
        products : list[str]
            A list of file paths (products) produced by do_processing method.
```

bourque (commented):
This is looking good to me! I just have some minor suggestions

imap_processing/cli.py
tmplummer: Thanks for the help with my numpydoc formatting. This is the first time I have used it. I just changed my pycharm setting to numpy docstrings so hopefully that will help.

imap_processing/cli.py
tmplummer: I think that we could log a warning but I believe that the intent of having the `--upload-to-sdc` option in the CLI is so that users can run processing without uploading products to the SDC by default.

imap_processing/cli.py
tmplummer: Not super nit-picky. Thanks for suggesting this.

maxinelasp (approved):
I like this, nice way to standardize

PR #552, opened 2024-05-09
Add launch_time parameter to cdf.utils.calc_start_time

bourque:
# Change Summary

## Overview

This PR adds an optional parameter to `cdf.utils.calc_start_time` for a user-specified launch time. This change is motivated by the desire from CoDICE to use their own launch time that is slightly different from the one we define in `imap_processing.__init__.py`.

## Updated Files

- `imap_processing/cdf/utils.py`
   - Added new `launch_time` parameter, which defaults to using the `launch_time` specified at the package level
- `imap_processing/tests/cdf/test_utils.py`
   - Expanded test to test the new parameter

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -3,6 +3,7 @@
 import logging
 import re
 from pathlib import Path
+from typing import Optional
 
 import imap_data_access
 import numpy as np
@@ -14,7 +15,9 @@ from imap_processing import launch_time
 logger = logging.getLogger(__name__)
 
 
-def calc_start_time(shcoarse_time: float) -> np.datetime64:
+def calc_start_time(
+    shcoarse_time: float, launch_time: Optional[np.datetime64] = launch_time
+) -> np.datetime64:
     """Calculate the datetime64 from the CCSDS secondary header information.
 
     Since all instrument has SHCOARSE or MET seconds, we need convert it to
@@ -24,6 +27,8 @@ def calc_start_time(shcoarse_time: float) -> np.datetime64:
     ----------
     shcoarse_time: float
         Number of seconds since epoch (nominally the launch time)
+    launch_time : np.datetime64
+        The time of launch to use as the baseline
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


imap_processing/cdf/utils.py
```diff
@@ -14,7 +15,9 @@
 logger = logging.getLogger(__name__)
 
 
-def calc_start_time(shcoarse_time: float) -> np.datetime64:
+def calc_start_time(
+    shcoarse_time: float, launch_time: Optional[np.datetime64] = launch_time
```
greglucas, line 19: No need to change this here, but I find it odd we have `launch_time` in the root level of `imap_processing`. There are conflicting uses of it as well with IDEX also using a different value, and `hi` the only instrument using this value outside of this helper function.
https://github.com/search?q=repo%3AIMAP-Science-Operations-Center%2Fimap_processing%20launch_time&type=code

greglucas (approved):

vmartinez-cu (approved):

imap_processing/cdf/utils.py
tech3371: I thought I updated all places it uses it. Looks like I missed IDEX. I started seeing that we were storing launch_time in different places and thought we should keep it in one place and update it when we get real one.

Is there reason for making this argument optional or adding it? Just curious

bourque: My motivation for adding the parameter is because CoDICE prefers to use a `launch_time` of `2010-01-01T00:01:06` instead of `2010-01-01T00:01:06.184` 🤷🏻 

Alternatively, I could just take care of this within the CoDICE code by essentially making a CoDICE-specific `calc_start_time` function.

bourque: Oh I think I misunderstood you question. Are you asking whether it is best to make the parameter optional or required?

tech3371: Your first response was what I was wondering. Cool

greglucas: Maybe this shouldn't be a cdf_utils helper, but a more general helper for any time offset from launch epoch?

Right now hi is creating a `get_direct_event_time()` using the global launch time, but that is really just the same function as here. `calc_time_from_launch(offset, units="ns", launch_time=some_default)` could then be used by hi with their current default, idex with theirs, and codice with theirs.

I still think this is good to go and can be cleaned up later.

tech3371: Ah ok. That makes sense!

tech3371 (approved):

PR #553, opened 2024-05-09
Add Codecov token

bourque:
Hoping this will fix the Codecov reports which are currently stalling because (i think) we have reached our upload limit.


bourque:
I think that worked!

greglucas (changes requested):
I don't think we should be putting the token in the public repository. I'd yank that token if you can.
I'll open an alternative following these instructions to add a secret to the repo and use that to see if it works.
https://docs.codecov.com/docs/adding-the-codecov-token

PR #557, opened 2024-05-20
CoDICE L1a various fixes

bourque:
# Change Summary

## Overview
This PR contains a few updates to the CoDICE L1a algorithm, namely:
- Updated CDF data variables, attributes, and attribute value names
- Documented what still needs to be addressed with TODOs
- Updated packet definitions

## Updated Files

  - Various updates to CDF attributes/names
  - Added lo sweeping values and acquisition times to xarray dataset
   - Various updates to CDF names and values 
- `imap_processing/codice/data/lo_stepping_values.csv`
  - Updated values 
- `imap_processing/codice/packet_definitions/P_COD_LO_*.xml`
  - Some updates to packet definitions
  - Updated expected values 
- `tools/xtce_generation/TLM_COD.xlsx`
  - Updated telemetry defenitions 
- `tools/xtce_generation/xtce_generator_codice.py`
  - Updated APIDs


imap_processing/codice/codice_l1a.py
```diff
@@ -163,16 +182,18 @@ def get_acquisition_times(self):
...
-
-        # TODO: Expand acquisition times list so that each energy step has an
-        # associated time
+        self.acquisition_times = []
+        energy_steps = lo_stepping_values[
+            ["e1", "e2", "e3", "e4", "e5", "e6", "e7", "e8"]
+        ].astype(str)  # convert to string to avoid confusion with table index value
+        for step_number in range(128):
+            index = energy_steps.isin([str(step_number)]).any(axis=1).idxmax()
+            self.acquisition_times.append(lo_stepping_values.acq_time[index])
```
tech3371: can you add some comments for clarification? Appreciate it

imap_processing/codice/cdf_attrs.py
```diff
@@ -83,35 +75,83 @@
...
+    validmin=0,
+    validmax=127,
+    var_type="support_data",
+    scale_type="linear",
+)
+
+esa_sweep_attrs = AttrBase(
+    catdesc="ElectroStatic Analyzer Energy Values",
+    display_type="no_plot",
+    fieldname="ESA V",
```
tech3371: can you spell `V` here?

imap_processing/codice/cdf_attrs.py
```diff
@@ -83,35 +75,83 @@
...
+    fill_val=GlobalConstants.INT_FILLVAL,
+    format="I12",
```
tech3371:
```suggestion
    format="I19",
```

tech3371 (approved):
couple minor comments

imap_processing/codice/codice_l1a.py
```diff
@@ -94,13 +98,14 @@ def create_science_dataset(self, packets: list) -> xr.Dataset:
...
-        # TODO: Resolve "Python into too large to convert to C long" error
         """
         epoch = xr.DataArray(
-            [calc_start_time(packets[0].data["SHCOARSE"].raw_value)],
+            [
+                calc_start_time(
+                    packets[0].data["ACQ_START_SECONDS"].raw_value,
+                    launch_time=np.datetime64("2010-01-01T00:01:06.184", "ns"),
+                )
+            ],
```
greglucas: Do you want a start time per packet? Otherwise, I think this is just a single epoch value as your coordinate.
```suggestion
            [
                calc_start_time(
                    packet.data["ACQ_START_SECONDS"].raw_value,
                    launch_time=np.datetime64("2010-01-01T00:01:06.184", "ns"),
                ) for packet in packets
            ],
```

imap_processing/codice/codice_l1a.py
greglucas: I agree, I'm not sure I follow this either, so it would be good to document why you're doing isin, any, and idxmax on the previous line as well.

It looks like you're doing == but using isin, are you not able to do a comparison on the arrays? `energy_steps == str(step_number)`

imap_processing/codice/codice_l1a.py
```diff
@@ -243,20 +291,18 @@ def unpack_science_data(self, packets: list):
...
         science_values = packets[0].data["DATA"].raw_value
 
         # Divide up the data by the number of priorities or species
-        num_bytes = len(science_values)
+        num_bits = len(science_values)
         chunk_size = len(science_values) // self.num_counters
 
         self.data = [
-            science_values[i : i + chunk_size] for i in range(0, num_bytes, chunk_size)
+            science_values[i : i + chunk_size] for i in range(0, num_bits, chunk_size)
```
greglucas, line 312: Do you know if these will be integers and could convert them right away?

```suggestion
            int(science_values[i : i + chunk_size], 2) for i in range(0, num_bits, chunk_size)
```

imap_processing/codice/packet_definitions/P_COD_LO_NSW_SPECIES_COUNTS.xml
```diff
@@ -33,6 +33,9 @@
 			<xtce:IntegerParameterType name="UINT20" signed="false">
 				<xtce:IntegerDataEncoding sizeInBits="20" encoding="unsigned" />
 			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="UINT23" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="23" encoding="unsigned" />
+			</xtce:IntegerParameterType>
 			<xtce:IntegerParameterType name="UINT24" signed="false">
```
greglucas: I haven't looked closely, but is `UINT24` unused now and should be removed?

tools/xtce_generation/TLM_COD.xlsx
greglucas, line 1: I think we decided not to commit the `xlsx` files and only the `xml`.

https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/549#discussion_r1593085199

imap_processing/codice/codice_l1a.py
bourque: I will certainly add more clarification in the code but also want to clarify here. In full disclosure, this code is a chat-gpt inspired solution 😅 

The table I am using looks like this:

Screenshot 2024-05-22 at 09 02 10

The goal of line 195 is to try to figure out a corresponding row value for each number between 0 and 127.

imap_processing/codice/codice_l1a.py
bourque: Thanks, nice catch. I think this is a relic from some debugging-past.

imap_processing/codice/codice_l1a.py
bourque: Not quite, these are intended to be a list of 128 bits

imap_processing/codice/packet_definitions/P_COD_LO_NSW_SPECIES_COUNTS.xml
bourque: Yes, thanks!

imap_processing/codice/codice_l1a.py
bourque: This can be simplified to:

```python
index = (energy_steps == str(step_number)).any(axis=1).idxmax()
```

The `axis=1` is finding in which _row_ the step number is found, and `idxmax()` is converting that to an index number.

greglucas: As an alternative suggestion. If they are all unique, I think you might be able to use `np.nonzero(energy_steps == str(step_number))[0]` with `nonzero` giving you the indices of the entries that matched your expression.
https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html

I think the `idxmax()` is throwing me off because it seems like it is being used for more than one return value. `argwhere` or `argmax` might be suitable as well.

bourque: Thanks. It looks like this also works:

```python
index = np.argmax(energy_steps == str(step_number), axis=1).argmax()
```

I don't really find any of these solutions to be very readable/straightforward to be honest, but perhaps this one is the easiest to follow(?)

greglucas: I still think you just want to grab the single value though and not necessarily the "max"?

https://stackoverflow.com/questions/69329249/how-do-i-find-the-index-of-an-specific-value-in-numpy-array

```python
# will return (5, 6) row/col of the first match
np.argwhere(a == 0)[0]
# I think will return 5, the row of the first match
np.argwhere(a == 0)[0][0]
```

bourque: Right, so the `argwhere` solution would be:

```
index = np.argwhere(energy_steps == str(step_number)[0][0]
```

So of all these solutions, which is the best?

```python
index = (energy_steps == str(step_number)).any(axis=1).idxmax()
```

```python
index = np.argmax(energy_steps == str(step_number), axis=1).argmax()
```

```python
index = np.argwhere(energy_steps == str(step_number)[0][0]
```

PR #558, opened 2024-05-21
Removed all non-science data for SWE

tech3371:
# Change Summary
Checked with Ruth from SWE and she ok of removing non-science data that she can access from POC interface.

## Overview
Deleted all code that was processing non-science data
Deleted tests and test data of non-science data


imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -41,17 +35,8 @@ def swe_l1b(l1a_dataset: xr.Dataset):
...
-    else:
-        data = eu_data
-        # Update global attributes to l1b global attributes
-        data.attrs.update(swe_cdf_attrs.swe_l1b_global_attrs.output())
-        # TODO: replace "sci" with proper descriptor, or add to global attributes
-        data.attrs["descriptor"] = "sci"
+    data = swe_l1b_science(eu_data)
+    if data is None:
+        print("No data to write to CDF")
+        return
```
greglucas: Do we want this as an info/warning instead? Also, an empty `return` still returns `None`, so is the same here as `return data`, which is `return None` in this case. Alternatively, would you want to put this logging/info print into the swe_l1b science routine itself to say we didn't get any data there?

```suggestion
        logger.info("No data to write to CDF")
```

imap_processing/tests/swe/test_swe_l1b.py
```diff
@@ -78,63 +77,24 @@ def test_swe_l1b(decom_test_data):
...
-        "HVPS_TEMP",
-        "LVPS_BACK_BOARD_TEMP",
-    ]
-
-    # Check EU values from housekeeping data
-    for field in eu_field_list:
-        assert round(hk_l1b[field].data[1], 5) == round(validation_data[field], 5)
-
 
+@pytest.mark.xfail(reason="Will update test data with multiple packets")
```
greglucas, line 81: Does L1B need multiple packets? It looks like you are just validating filenames below, so I would have expected this to still work.

greglucas (approved):

imap_processing/tests/swe/test_swe_l1b.py
tech3371: I can't create l1b CDF file if I don't have any good processed data. It takes 4 packets to create one good cycle data. In upcoming PR, Yaxi gave me bin file with multiple good packets data. I plan to update this in upcoming PR. small PR with intention

PR #562, opened 2024-05-23
ENH/tools: Add metadata helper script to convert excel to yaml

greglucas:
This adds the ability to go from an Excel metadata definition spreadsheet to yaml that the processing code can easily read in without relying on any Excel reading within the processing suite.

Currently, it is hard-coded to work with SWAPI's metadata Excel spreadsheet as a proof of concept. Happy to hear feedback/ideas for improvements or things that might go wrong.


tools/metadata_generation/swapi_metadata_generator.py
```diff
@@ -0,0 +1,97 @@
...
+        A dictionary of the global attributes, where the keys are the attribute names
+        and the values are the attribute descriptions.
+    """
+    sheet = sheet.set_index("NAME")
+    # subset by global_att field type, then only keep the CATDESC column
+    sheet = sheet[sheet["FIELD_TYPE"] == "global_att"][["CATDESC"]]
+    return sheet.to_dict()["CATDESC"]
+
+
+def get_file_attributes(sheet, global_attrs=None):
```
tech3371: Is the purpose of having optional parameter for `global_attrs` is so that user can define their own and pass it in?

tools/metadata_generation/swapi_metadata_generator.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    sheets = pd.read_excel(excel_path, sheet_name=None)
+
+    # Dictionary to store all of the metadata
+    metadata_dict = {}
+    metadata_dict["global_attrs"] = get_global_attributes(sheets["global_attrs"])
+
+    for key in sheets:
+        if not key.startswith("imap_"):
+            # Only worry about the imap_* sheets
```
tech3371:
```suggestion
            # Only worry about the imap_* tabs
```

tools/metadata_generation/swapi_metadata_generator.py
```diff
@@ -0,0 +1,97 @@
...
+    output_json_path : str | pathlib.Path
+        The path to the output JSON file.
+    """
+    sheets = pd.read_excel(excel_path, sheet_name=None)
+
+    # Dictionary to store all of the metadata
+    metadata_dict = {}
+    metadata_dict["global_attrs"] = get_global_attributes(sheets["global_attrs"])
+
+    for key in sheets:
```
tech3371:
```suggestion
    for tab in sheets:
```

tools/metadata_generation/swapi_metadata_generator.py
```diff
@@ -0,0 +1,97 @@
...
+            # Only worry about the imap_* sheets
+            # ignore support data, and codes
+            continue
+        metadata_dict[key] = get_file_attributes(
+            sheets[key], metadata_dict["global_attrs"]
+        )
+
+        # Save it out to a file
+        with open(output_json_path, "w") as f:
+            f.write(json.dumps(metadata_dict, indent=4))
```
tech3371: Thank you for adding this tab. It was helpful to review!

tools/metadata_generation/swapi_metadata_generator.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    # Add the global attributes for this specific sheet
+    file_attrs = {"global_attrs": get_global_attributes(sheet)}
+    # Add the global attributes from the global_attrs sheet (if any)
+    file_attrs["global_attrs"].update(global_attrs or {})
+
+    # Move onto variable and coordinate metadata
+    # Get everything that isn't a global attribute and drop that column
+    sheet = sheet[sheet["FIELD_TYPE"] != "global_att"].drop(columns=["FIELD_TYPE"])
+    sheet = sheet.set_index("NAME")
```
tech3371: I think I understand what this does. But to double check, does this `drop` call filters out all non-global attrs and gets those columns and assign it to `sheet`? It may be helpful to have different name than `sheet`. I suggested something but you are better with names. Can you pick another name?
```suggestion
    data_or_coord_data = sheet[sheet["FIELD_TYPE"] != "global_att"].drop(columns=["FIELD_TYPE"])
    data_or_coord_data = data_or_coord_data.set_index("NAME")
```

tech3371 (commented):
This makes sense to me. I had some comments. Thank you for getting to this really quickly!

tools/metadata_generation/swapi_metadata_generator.py
greglucas: I was thinking there might be multiple different global_attrs that need to be put in at different times, but that turned out not to be the case after all, we can make this mandatory if that'd be easier. But, if someone wants to put all global_attrs in every tab/sheet, then they could do that and not have a separate global_attrs tab/sheet.

tools/metadata_generation/swapi_metadata_generator.py
greglucas: They are technically separate "sheets" I believe, it just so happens to be displayed as a tab in the program.
I have a preference for sticking with "sheet" rather than "tab", but not a major preference and could be convinced to change if others also want this update.

tools/metadata_generation/swapi_metadata_generator.py
greglucas: This is modifying the structure/format of "sheet". I would call it `df` or `variable_data` if anything else I guess since it is a pandas dataframe. It is pretty standard in pandas to overwrite your local `df` variable as you modify it, so I'm not sure how helpful renaming would be though.

There is a comment above this saying we are getting all rows that aren't equal to `global_att` as our subset. Then we are getting rid of that column because we don't want to propagate it in the output metadata. Or do we need it in the CDF metadata as well?

greglucas: Updated to `variables`

tools/metadata_generation/swapi_metadata_generator.py
greglucas: I agree, it was overwhelming before as a single line!

tools/metadata_generation/swapi_metadata_generator.py
greglucas: I changed this to `sheet_name` to keep the "sheet" qualifier rather than tab. TBD on whether we move to tabs still pending further comments from others.

tools/metadata_generation/swapi_metadata_generator.py
tech3371: Ah ok. Yeah, that makes sense.

tools/metadata_generation/swapi_metadata_generator.py
tech3371: I was thinking `variable_or_coord_data` when I made suggestions but wrote something else. `variables` sounds good.

No, we don't need that column after that filter.

tools/metadata_generation/swapi_metadata_generator.py
tech3371: I am curious what others think as well.

tech3371:
To me, this looks great as a temporary tool for SIT-3! And then extend it or replace as needed post SIT-3.

maxinelasp:
How hard would it be to add more complicated spreadsheets? For example, in MAG, I have 4 outputs that are all very similar, but have a few differences. I'd prefer not to duplicate everything 4 times - do you think it would be feasible to use this method with that?

greglucas:
> How hard would it be to add more complicated spreadsheets? For example, in MAG, I have 4 outputs that are all very similar, but have a few differences. I'd prefer not to duplicate everything 4 times - do you think it would be feasible to use this method with that?

@maxinelasp That is a really good question. I _think_ it would be feasible, but, I think it will depend on where we want to make those simplifications (in the spreadsheets through cell references, in this tool during the conversion process, or in the code that uses this output JSON dictionaries). Do you have a sense of where you'd want to remove the duplication?

tools/metadata_generation/swapi_metadata_generator.py
tmplummer: I know what both mean... but I do think that excel refers to them as sheets.

vmartinez-cu: "sheets" is consistent with what I've seen in code that processes excel files. Maybe you can clarify that 'sheets' and 'tabs' are synonymous terms in the doc string for clarification

imap_processing/swapi/swapi_metadata.json
```diff
@@ -0,0 +1,246 @@
...
+            "VALIDMAX": 3.15563e+18,
+            "FILLVAL": -9.22337e+18,
+            "VAR_TYPE": "support_data",
+            "SCALE_TYP": "linear",
+            "TIME_BASE": "J2000",
+            "TIME_SCALE": "Terrestrail Time",
+            "MONOTON": "INCREASE",
+            "REFERENCE_POSITION": "Rotating Earth Geoid"
+        },
+        "<to-be-replaced-in-code>": {
```
tmplummer: It is probably worth coming up with a standard syntax for items that will be replaced in code.

tools/metadata_generation/swapi_metadata_generator.py
tmplummer: Should this be called `get_sheet_attributes`?

tools/metadata_generation/swapi_metadata_generator.py
```diff
@@ -0,0 +1,97 @@
...
+    # Move onto variable and coordinate metadata
+    # Get everything that isn't a global attribute and drop that column
+    variables = sheet[sheet["FIELD_TYPE"] != "global_att"].drop(columns=["FIELD_TYPE"])
+    variables = variables.set_index("NAME")
+
+    # Iterate through the rows of the sheet, each row being a variable or coordinate
+    # Then drop any NaN values and convert the row to a dictionary of key/values like:
+    # {"CATDESC": "description", "VAR_TYPE": "support_data", ...}
+    for row, data in variables.iterrows():
+        file_attrs[row] = data.dropna().to_dict()
```
tmplummer: How does dropping all NaN values affect ISTP compliance for the required attributes?

imap_processing/swapi/swapi_metadata.json
```diff
@@ -0,0 +1,246 @@
...
+            "Mission_group": "IMAP>Interstellar Mapping and Acceleration Probe",
+            "PI_name": "Dr. David J. McComas",
+            "PI_affiliation": "Princeton Plasma Physics Laboratory",
+            "File_naming_convention": "source_descriptor_datatype_yyyyMMdd_vNNN",
+            "Data_version": 1,
+            "Descriptor": "swapi>The Solar Wind and Pickup Ion",
+            "TEXT": "The Solar Wind and Pickup Ion (SWAPI) instrument measures several different elements of the solar wind, including hydrogen (H) and helium (He) ions, and, on occasion, heavy ions produced by large events from the Sun. See https://imap.princeton.edu/instruments/swapi for more details. SWAPI level-1 data contains primary, secondary, coincidence counts per ESA voltage step and time. Level-2 data contains the same data as level-1 but counts are converted to rates by dividing counts by time.",
+            "Instrument_type": "Particles (space)"
+        },
+        "epoch": {
```
tmplummer: How can we make sure that `epoch` has consistent attributes across all data products?

imap_processing/swapi/swapi_metadata.json
```diff
@@ -0,0 +1,246 @@
+{
+    "global_attrs": {
```
tmplummer: I think it is worth still having a "global" global_attrs definition for attributes that are common and should be identical across all data products.

tmplummer (commented):
This is great, Greg. Everything looks good to me... but I did make some comments/questions of things mostly dealing with keeping consistency across instruments for common attributes.

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -477,64 +475,34 @@ def process_swapi_science(sci_dataset):
...
-        attrs=dataclasses.replace(
-            counts_attrs,
-            fieldname="Primary CEM counts",
-            label_axis="PCEM cnts",
-            catdesc="Primary Channel Electron Multiplier (CEM) counts",
-        ).output(),
+        attrs=metadata["swp_pcem_counts"],
     )
     dataset["swp_scem_counts"] = xr.DataArray(
         np.array(swp_scem_counts, dtype=np.uint16),
```
tmplummer: I added a DATA_TYPE column to my spreadsheet thinking it could be used to specify the `dtype` when creating the `DataArray`s.

tools/metadata_generation/swapi_metadata_generator.py
laspsandoval: I know both.

laspsandoval:
Great temporary tool. I'm ready to use it!

laspsandoval (approved):
It looks good. Great temporary tool. I'm ready to use it.

imap_processing/swapi/l1/swapi_l1.py
greglucas: I believe it could be accessed like `dtype=np.dtype(metadata["swp_scem_counts"]["DATA_TYPE"])` if you wanted to set that here in that way.

One thing that leads to is that with the "dump everything into metadata" approach, there would be a lot of extra metadata attributes then if we care about that showing up in the output data files or not. Something worth thinking about whether we'd want to pop it off the dict at some point, or just live with it also being an attribute field.

imap_processing/swapi/swapi_metadata.json
greglucas: This is from that sheet.

I think it might be more descriptive to relabel the nested one with the same name below this in "imap_swapi_l1_sci" to something like "dataset_attrs", which is composed from this **and** the extra fields in that spreadsheet.

imap_processing/swapi/swapi_metadata.json
greglucas: I think this is a great question, and I'm not sure you can with this approach. You'd either have to check in the spreadsheets (maybe link/reference more cells), or check further downstream with I hope what Maxine's approach will provide.

tools/metadata_generation/swapi_metadata_generator.py
greglucas: Good idea to update this. I opted for `get_dataset_attributes` as I think each "sheet" we are interested in here is a dataset. But the `global_attrs` is also a sheet, so maybe that is a better distinguisher? Let me know if you have a strong preference for "sheet" here and we can use that too as it is much better than file.

tools/metadata_generation/swapi_metadata_generator.py
greglucas: Another great question! This is something we should talk about next week. Where do these checks occur. Do we want it to occur in this tool, upstream in the spreadsheet itself, or downstream in some other validator (possibly just relying on cdflib?). I was definitely going for the "just convert" approach here and was assuming validation is done elsewhere.

imap_processing/swapi/swapi_metadata.json
tmplummer: Maxine's approach does provide this! And also addresses my concern about global attribute consistency across instruments and products.

tools/metadata_generation/swapi_metadata_generator.py
tmplummer: I agree... `get_dataset_attributes` is better yet.

tmplummer (approved):
This is great. Connecting this code to Maxine's code and working out some details will provide a nice way to easily modify/update product definitions and maintain consistency across instruments and products.

tech3371 (approved):

bourque (approved):
I agree with everyone else's comments here -- I think this is a nice improvement and will allow for better consistency!

greglucas:
Just a note that we will be discussing this and #564 on Thursday, so I won't be merging this until after this is discussed then and decided on the direction we want to go more generally. Folks should also review that PR to get ideas and thoughts on benefits/drawbacks to a combination of approaches or one or the other, or something else entirely.

greglucas:
I have just pushed up some changes that gets this a little closer to the format expected by #564. There is still quite a bit of repetition and lack of inheritance, but this should at least be an OK start for people to get the basic text and formatting working. Look at the recent `*.yaml` files generated. It would be more work to try and figure out how to add tags/anchors, if anyone else wants to take a stab at that feel free to pick this up and extend it.

maxinelasp (approved):
Yaml files LGTM, although as you say they are a bit repetitive. It's a great start for people, I think

greglucas:
@tech3371 I have removed all SWAPI related files, so you'll have to update that with the yaml portion from Maxine's upcoming PR. This is now purely an "extra tool" PR that people can use to make a first cut at yaml metadata files and has no impact on the actual processing code at this point.

greglucas:
Merging this in so people can use it if need be for their initial metadata work. It does not touch any used code and doesn't update anything other than the tools area.

PR #563, opened 2024-05-23
[WIP] New file defining HIT CDF attributes

vmartinez-cu:
# Change Summary
Created a new file to define HIT CDF attributes (WIP)

## Overview
This file defines HIT CDF attributes. It currently only has attributes for the housekeeping packets. These packets have over 50 data variables. I defined each of these in dictionaries for L1A and L1B products. I'm not sure that this is the best approach. I'm considering creating data classes for each data variable and their attributes instead since these attributes won't need to change (as far as I know). Open to ideas on different approaches!

## New Dependencies
None

## New Files
- hit_cdf_attrs.py
   - Defines CDF attributes for the HIT data products. This will be used by the data processing pipeline

## Deleted Files
None

## Updated Files
- init.py
   - Added version. This gets imported into hit_cdf_attrs.py

## Testing
None yet


imap_processing/hit/l1a/hit_cdf_attrs.py
```diff
@@ -0,0 +1,762 @@
...
+
+
+# Dictionary of housekeeping attributes that are common between L1A and L1B
+l1a_l1b_hk_attrs = {
+    "fsw_version_a": replace(
+        hit_hk_base_attrs,
+        validmax=3,
+        catdesc="Flight Software Version Number (A.B.C bits)",
+        fieldname="Flight Software Version Number A",
+        label_axis="Flight software version number a",
```
tmplummer: The SPDF required attribute is spelled "LABLAXIS" and says that this field should only be 6-10 characters. Something like "FSW A" for this variable?
I do find the SPDF examples page useful to look at for trying to figure out exactly what they want for each attribute: https://spdf.gsfc.nasa.gov/istp_guide/variables.html#data_eg3

imap_processing/hit/l1a/hit_cdf_attrs.py
```diff
@@ -0,0 +1,762 @@
...
+    ),
+    "ebox_p2d0vd": replace(
+        hit_hk_base_attrs,
+        validmin=1,  # doc says min 1.79
+        validmax=3,  # doc says max 2.5
+        catdesc="+2.0VD Ebox (Digital)",
+        fieldname="+2.0VD Ebox (Digital)",
+        label_axis="+2.0VD Ebox (Digital)",
+        units="V",
+        format="F1.2",
```
tmplummer: If I understand fortran formats correctly for floats, it is Fw.d where w=total number of digits in the output, d=number of digits to the right of the decimal point. So, for this value I think you want `"F3.2"`.

imap_processing/hit/l1a/hit_cdf_attrs.py
```diff
@@ -0,0 +1,762 @@
...
+        catdesc="Board Temp Mounted in the Middle of Board on Opposite Side of "
+        "Hottest Component",
+        fieldname="Board Temp",
+        label_axis="LVPS temp",
+        units="C",
+        format="I2",
+    ),
+    "ebox_3d4vd": replace(
+        hit_hk_base_attrs,
+        validmin=3,  # doc says min 3.1
```
tmplummer, line 704: I'm curious about what documentation these limits are coming from?
If it is the limits tab in the TLM spreadsheet, we should verify the meaning. Typically, there are yellow and red limits designated in those spreadsheets that the OPS folks use to trigger alarms which send out notifications that there is something on the s/c or instrument that needs to be investigated. I would interpret the limits in the CDF files differently.

tmplummer (commented):
Wow, this is a lot of work that shows great attention to detail. I had a few comments about things I noticed in the variable attributes.

imap_processing/hit/l1a/hit_cdf_attrs.py
```diff
@@ -0,0 +1,762 @@
...
+
+descriptor = "HIT>IMAP High-energy Ion Telescope"
+
+hit_description_text = (
+    "The High-energy Ion Telescope (HIT) measures the elemental composition, "
+    "energy spectra, angle distributions, and arrival times of high-energy ions. "
+    "HIT delivers full-sky coverage from a wide instrument field-of-view (FOV) to "
+    "enable a high resolution of ion measurements, such as observing shock-accelerated "
+    "ions, determining the origin of the solar energetic particles (SEPs) spectra, and "
+    "resolving particle transport in the heliosphere."
```
tech3371, line 21:
```suggestion
    "resolving particle transport in the heliosphere."
    "See https://imap.princeton.edu/instruments/hit for more details. "
```

imap_processing/hit/l1a/hit_cdf_attrs.py
```diff
@@ -0,0 +1,762 @@
...
+        fieldname="Flight Software Version Number A",
+        label_axis="Flight software version number a",
+        format="I1",
+    ),
+    "fsw_version_b": replace(
+        hit_hk_base_attrs,
+        validmax=15,
+        catdesc="Flight Software Version Number (A.B.C bits)",
+        fieldname="Flight Software Version Number B",
+        label_axis="Flight software version number b",
```
tech3371: To add to Tim's point, `label_axis` could be like one word or two words. Adding this suggestion as an example.
```suggestion
        label_axis="Version",
```

tech3371 (approved):
Thank you for all the detailed work. Nice work! I hope we can look for tools to do this more automatically later. Also, in the future, we can discuss to what detail we want to write for data product such as housekeeping which doesn't get used passed level-1. And SPDF doesn't have strick requirement for level-1 data.

imap_processing/hit/l1a/hit_cdf_attrs.py
tmplummer: Never mind about the attribute name. I tested that `label_axis` correctly gets converted to `LABLAXIS` and it does.

vmartinez-cu: Thanks for confirming the attribute name! Also, thanks for the character limit info for label axis. I will double check this for the other fields as well. Hopefully users of this data will know what "FSW A" means! Sean and I have a meeting with the instrument team tomorrow which will help refine some of these details

imap_processing/hit/l1a/hit_cdf_attrs.py
vmartinez-cu: This is very helpful! I haven't worked with fortran formats before. I thought the first number was the number of digits left of the decimal and the second number was the number of digits right of the decimal. I'll update these values

imap_processing/hit/l1a/hit_cdf_attrs.py
vmartinez-cu: That's correct. I pulled the red low and red high values form the limits tab of the TLM spreadsheet. Good to know that CDF limits may not need to match values in the limits tab. I'll try to get some clarification on these values when I meet with the instrument team.

imap_processing/hit/l1a/hit_cdf_attrs.py
tech3371: The reason we are able to keep label axis value short is because plot title(`fieldname` of cdf attrs) will contain more descriptive information and label axis will be shorter version of it. I hope that helps clear some confusion.

PR #564, opened 2024-05-23
SwxSOC cdf investigation

maxinelasp:
# Change Summary
This is a first draft pass of using the SWXSOC CDF style of attribute generation. This actually doesn't depend on their code, and more rewrites part of it to pull out the common pieces. Ideally, I would want to put this in its own repo, perhaps in the LASP repository, so both us and the Hermes team could make improvements to it. But this is just a proof of concept for now. 

## Overview

This mostly works with the MAG L1A RAW CDF files, with some minor fixes to the way the variable bases work still needed. 
There is also a mismatch between the global schema provided by HERMES and the added global attrs we have, but I think that will be easy to fix. 

## New Dependencies

Added the HERMES-SOC as a dependency, but it's not really needed

This still requires some coordination with the HERMES team - I think ideally I would pull the shared pieces out into a different repo, give them credit, and then if they wanted, they could replace the duplicate pieces in their repo with the library. Still an open question where that repo should live, who should maintain, etc - LMK thoughts!


imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+mag_metadata_attrs: 
+    VAR_TYPE: metadata
+    base: default_attrs
+    
+mag_flag_attrs:
+    VALIDMIN: 0
+    VALIDMAX: 1
+    DISPLAY_TYPE: time_series
+    FILLVAL: 255
```
laspsandoval, line 43: Is this supposed to be the max value? Fill value 255 seems odd.

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,150 @@
...
+direction_attrs: 
+    CATDESC: Magnetic field vector
+    FIELDNAM: \[xyz\] magnetic field vector
+    FORMAT: I3
+    VAR_TYPE: support_data
+    DISPLAY_TYPE: time_series
+    LABLAXIS: Magnetic field vector
+    base: default_attrs
+    
+PUS_SPARE1:
```
laspsandoval: Should these be lowercase?

imap_processing/mag/l0/decom_mag.py
```diff
@@ -66,7 +67,7 @@ def decom_packets(packet_file_path: str | Path) -> dict[str, list[MagL0]]:
     return {"norm": norm_data, "burst": burst_data}
 
 
-def generate_dataset(l0_data: list[MagL0], dataset_attrs: dict) -> xr.Dataset:
+def generate_dataset(l0_data: list[MagL0], cdf_attrs: ImapCdfData, logical_source_id: str) -> xr.Dataset:
```
laspsandoval: Nice!

greglucas:
Does this give us metadata validation as well as the schema?

I'm wondering if there is an easy way to go from spreadsheet -> this yaml format -> code. I'm guessing there is and we'd just need to update the PR I opened.

One thing that worries me here is the complexity this adds versus a straight dictionary lookup. Could we have the classes inherit from MutableMapping to make them dict-like? Or am I missing another reason for why we want to use the extra functions to access the metadata.

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,150 @@
...
+  
+mag_raw_vector_attrs:
+  CATDESC: Raw unprocessed magnetic field vector data in bytes
+  DEPEND_1: direction
+  FIELDNAM: Magnetic Field Vector
+  LABLAXIS: Raw binary magnetic field vector data
+  FORMAT: I3
+  UNITS: counts
+  VAR_TYPE: data
+  base: default_attrs
```
tmplummer: Oh, interesting that this defines a base which will be default values if unspecified. I think that I like this... although the defaults won't work for many datatypes.

pyproject.toml
```diff
@@ -35,6 +35,7 @@ python = ">=3.9,<4"
 space_packet_parser = ">=4.2.0"
 spiceypy = ">=6.0.0"
 xarray = '>=2023.0.0'
+hermes_core = {git = "https://github.com/HERMES-SOC/hermes_core.git", branch="main"}
```
Alrobbertz: The HERMES version is more stable for now, but I'm moving a lot of functionality to the https://github.com/swxsoc/swxsoc repo. The `SWXDataSchema` is the "generalized" version of the Schema class that does not have the hermes-specific attributes and adds in the layering niceties. This is the class/version that I would like to split out into its own repo, maybe something like `swxschema`. This repo might be volatile in the upcoming weeks/months as I'm extending the schema for FITS/SOLARNET metadata attributes so we should discuss whether we want to do the division before or after I want to make a lot of volatile changes.

imap_processing/cdf/config/imap_default_global_cdf_attrs.yaml
tmplummer, line 1: I really like that this keeps the single source for global attributes common to all data products.

tmplummer (commented):
This is great. It solves some of the questions I raised when reviewing @greglucas PR that converts a spreadsheet to JSON. It seems like a combination of the two would provide the complete solution.

imap_processing/cdf/config/imap_mag_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,46 @@
...
+    Logical_source_description: IMAP Mission MAGi Normal Rate Instrument Level-1A Data.
+
+
+imap_mag_l1a_burst-mago:
+    Data_type: L1A_raw-burst-mago>Level 1A MAGo burst rate
+    Logical_source: imap_mag_l1a_burst-mago
+    Logical_source_description: IMAP Mission MAGo Burst Rate Instrument Level-1A Data.
+
+
+imap_mag_l1a_burst-magi:
```
tech3371, line 43: good that you are keeping data level global attrs separate. It was confusing at times when we were keeping all together.

tech3371 (commented):
I agree with Tim. Combination of both would be good at the end. I think we are going to need both components because excel sheet is going to be used for initial CDF data attrs definition and used as discussion method since most instrument teams are not heavily interested in looking at code or file produced from the code. Excel has been very helpful to get feedback from them much faster. Then afterward, we will need a tool to read the excel and get in the format that this tool is going to need. Now, with this tool, we could work with SWxSOC and others to write generic validation code that is similar across any CDF data being produced. That's what I could think so far. I will some more for the discussion.

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
maxinelasp: You can create a different base if needed, or overwrite the defaults if they don't work

maxinelasp:
Reply to @greglucas 

> Does this give us metadata validation as well as the schema?

It validates against the schema but the current version of the code does not do validation beyond that. 

> I'm wondering if there is an easy way to go from spreadsheet -> this yaml format -> code. I'm guessing there is and we'd just need to update the PR I opened.

Yes, I think so, if we decided to go that way. I have further thoughts that I'll bring up in the meeting

> One thing that worries me here is the complexity this adds versus a straight dictionary lookup. Could we have the classes inherit from MutableMapping to make them dict-like? Or am I missing another reason for why we want to use the extra functions to access the metadata.

This was just personal style from me, I think having a dictionary lookup would make a lot of sense too now that I have actually finished. It was just a level of abstraction in case I wanted to add something in between class creation and retrieval.

imap_processing/cdf/config/imap_mag_global_cdf_attrs.yaml
maxinelasp: Yeah I agree, this is clearer for me.

maxinelasp:
Re: Excel sheet, my strong preference would be for us to use the YAML files as the source of truth and use them to generate both the code and the spreadsheets for discussion. So we would edit the YAML files, but we could still use spreadsheets for discussions. I feel like if we invite edits from external people, there's a good chance they would (inadvertently) make a breaking change to the format and we would need to edit/verify everything carefully anyway. But, this is a discussion for our meeting about this.

imap_processing/cdf/config/imap_mag_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,46 @@
...
+TEXT: >
+  The IMAP magnetometer (MAG consists of a pair of identical magnetometers
+  which each measure the magnetic field in three directions in the vicinity of
+  the spacecraft.
+  MAG will contribute to our understanding of the acceleration and transport
+  of charged particles in the heliosphere.
+  MAG design and assembly is led by Imperial College, London. See
+  https://imap.princeton.edu/instruments/mag for more details.
+
+imap_mag_l1a_norm-raw:
```
tmplummer, line 12: Now that I am working on replicating this yaml for Hi, I have a comment:
It is not apparent in the global attrs yaml how the dependency works like it is in the variable attrs yaml. I really like being able to see that dependency directly in the yaml without having to snoop in the code to figure out how things like `Descriptor` and `TEXT` get propagated into product specific global attrs.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,192 @@
...
+        """
+        # TODO: Move this validation into the load step
+        template = OrderedDict()
+        for attr_name, attr_schema in self.global_attribute_schema.items():
+            if (
+                attr_schema["required"]
+                and not attr_schema["derived"]
+                and attr_name not in self.global_attributes
+            ):
+                # TODO throw an error?
```
greglucas: Seems like if it is required, yes we should error. Maybe add a warning for now?
```suggestion
                # TODO throw an error?
                warnings.warn(f"Missing {attr_name} from the template")
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,192 @@
...
+                template[attr_name] = None
+            elif attr_name in self.global_attributes:
+                template[attr_name] = self.global_attributes[attr_name]
```
greglucas: Could move this up to the first if block and warning down and then remove the `if attr_name in self.global_attributes` because you know that is False if you've hit further down the line.

imap_processing/cdf/config/imap_mag_global_cdf_attrs.yaml
maxinelasp: Maybe we could rework the global attrs to work the same way, with the default and then the specifics? Then when retrieving the attribute, you would always specify the logical_source_id.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,192 @@
...
+            path to the yaml file to load.
+
+        Returns
+        -------
+            Loaded yaml.
+        """
+        with open(file_path, "r") as file:
+            return yaml.safe_load(file)
+
+    def global_attribute_template(self, logical_source_id=None) -> OrderedDict:
```
greglucas: Is there a need for `OrderedDict`? Python dicts are now guaranteed ordered, so I find it easier to just use them if there isn't a specific reason for the `OrderedDict` implementation.
```suggestion
    def global_attribute_template(self, logical_source_id=None) -> dict:
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,192 @@
...
+        }
+        self.timetypes = [
+            const.CDF_EPOCH.value,
+            const.CDF_EPOCH16.value,
+            const.CDF_TIME_TT2000.value,
+        ]
+
+        # Load Default IMAP Global Attributes
+        self.global_attributes = self._load_yaml_data(
+            str(self.source_dir / DEFAULT_GLOBAL_CDF_ATTRS_FILE)
```
greglucas: You are casting to a lot of strings. If you already have a Path it looks like you should be able to keep it as a Path in your loader function as well already I think. I'd suggest removing all str() casts throughout.
```suggestion
            self.source_dir / DEFAULT_GLOBAL_CDF_ATTRS_FILE
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,192 @@
...
+        # We should add some way of extending the schema and remove all the HERMES specific stuff
+        # Data Validation, Complaiance, Derived Attributes
+        self.global_attribute_schema = self._load_default_global_attr_schema()
+
+        # Data Validation and Compliance for Variable Data
+        self.variable_attribute_schema = (
+            self._load_default_variable_attr_schema()
+        )
+
+        self.cdftypenames = {
```
greglucas: This is interesting, because it is using ctypes, which is specific to `spacepy`'s cdf writer and not cdflib which we are using... So I'm curious if we need this or what the value is for the IMAP project rather than numpy types, or are we going to run into any other issues?

imap_processing/mag/l0/decom_mag.py
```diff
@@ -120,27 +121,28 @@ def generate_dataset(l0_data: list[MagL0], dataset_attrs: dict) -> xr.Dataset:
         np.arange(vector_data.shape[1]),
         name="direction",
         dims=["direction"],
-        attrs=mag_cdf_attrs.raw_direction_attrs.output(),
+        attrs=cdf_attrs.variable_attribute_template("raw_direction_attrs"),
```
greglucas: What does "template" mean here? I think of a template more as the schema and now you've filled it out so we are doing a lookup/get.

```suggestion
        attrs=cdf_attrs.get_variable_attribute("raw_direction_attrs"),
        # or
        attrs=cdf_attrs.variable_attributes["raw_direction_attrs"],
        # or could you just have one global lookup that accounted for global and variable lookups?
        attrs=cdf_attrs["raw_direction_attrs"]
```

Do you need "attrs" on the direction, or could you make it a one-to-one mapping of `direction` in the name here?

pyproject.toml
greglucas: Remove now that it isn't a dependency.
```suggestion
```

greglucas (commented):
Some minor suggestions.

One other thing that I'm not completely following is the organization/nesting of global/variable attributes. What does "global" mean, I think it is actually "dataset" in this context, or is there different levels of "global".

Would it make sense to do more nesting within the yaml files themselves rather than within the Python code with the load_global and load_variable? Something like:
```yaml
# Dataset attributes
global_dataset: &global_dataset
  PI_NAME: ...

datasets:
  imap_mag_l1a-raw:
    <<: *global_dataset

variables:
  epoch: ...
  var1: ...
```

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: Well, right now we're using a schema that might not match, so it will always throw a warning

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: Are they really? I didn't realize that! Yeah, I'll remove ordereddict

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: the move to accept Path types was a late addition, but agree with this so I will update it.

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: This is from the HERMES soc so I'm not sure! I can remove this part and leave the numpy types...? I'm not really sure what IMAP will end up needing vs what we would want in the generic version of the code.

imap_processing/mag/l0/decom_mag.py
maxinelasp: I agree with you on template and will rename that method. 

But, the step where the attributes are retrieved is where the validation against the schema occurs. So, I actually think it should stay a method rather than directly accessing the method. There isn't really a good place to validate otherwise, since we can add partial definitions of variable files which might not include all the required variables in the schema to start. 

For now, it does just return a dictionary because there isn't a variable schema that we can use yet (it's contained inside hermes_default_variable_cdf_attrs_schema.yaml, but includes a bunch of hermes specific stuff. I'd like to move it into its own file.) but once it does, variable_attribute_template() (or whatever it is renamed to) will validate against the schema before returning.

maxinelasp:
> One other thing that I'm not completely following is the organization/nesting of global/variable attributes. What does "global" mean, I think it is actually "dataset" in this context, or is there different levels of "global".

Here, "global" refers to global file attributes. In xarray, this would be Dataset attributes. "variable" refers to everything else - in xarray, DataArray level attributes. There are multiple "global" values because there is a distinction between project specific and instrument or level specific attributes - but all global values are eventually put in the same level at the dataset. 

> Would it make sense to do more nesting within the yaml files themselves rather than within the Python code with the load_global and load_variable? 

Unfortunately you can't alias/anchor across different files. We either need to have multiple files and do some legwork in the code to combine them, duplicate mission level attributes, or have one massive file for everything. I'd prefer to have the code combine the files. Tim brought up moving some of the instrument attributes into the level attributes using aliases though, which is possible because it's in the same file.

imap_processing/mag/l0/decom_mag.py
greglucas: OK, I think that makes sense, some of this depends on the organization as well and where we want to access things from. I'd still personally use the dict lookup and injecting the validation in `__getitem__`, but the method access is fine too 👍 

```python
class CDFAttrs(dict):
    def __getitem__(self, key):
        val = dict.__getitem__(self, key)
        # Do your validation checks on the key or value
        self._validate(val)
        return val
```

maxinelasp: Could also use properties, which would probably be my preference.

imap_processing/cdf/config/imap_mag_global_cdf_attrs.yaml
maxinelasp: I put a change in to hopefully mitigate this? @subagonsouth I'd appreciate your thoughts

imap_processing/cdf/cdf_attribute_manager.py
greglucas: I didn't follow what this was actually used for. My assumption is some kind of validation, which makes me think that it might help HERMES validation because they use `spacepy` and these data types, but not us because we use a different set of datatypes. So, I think this is up to you whether this is needed or not, it just stuck out to me as something that might be more project-specific.

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,157 @@
+default_attrs: &default
```
laspsandoval, line 1: Where is the attribute for epoch?

imap_processing/cdf/cdf_attribute_manager.py
Alrobbertz: We use this for metadata derivations leveraging some of the `spacepy` code! There is some code that will help us identify the CDF variable types based on the numpy types. We then use this for setting default values for VALIDMIN, VALIDMAX, and FORMAT attributes. You shouldn't need this or the `const.py` if you're not going to use this for any attribute derivations using `spacepy`.

imap_processing/cdf/config/shared/hermes_default_global_cdf_attrs_schema.yaml
```diff
@@ -0,0 +1,278 @@
+CDF_Lib_version:
+  description: > 
+    Version of the CDF Binaries library used to generate the CDF File
+  default: null
+  derived: true
+  required: true # NOT Required in ISTP Guide (Derived)
+  validate: false
+  overwrite: false
```
Alrobbertz: This is unique for HERMES since we're using `spacepy` which relies on the CDF binaries. You should be able to remove this for IMAP since you're using CDFLIB. You might want to re-name or change the description so its more clear this is not the CDF Binaries version.

imap_processing/cdf/config/shared/hermes_default_global_cdf_attrs_schema.yaml
```diff
@@ -0,0 +1,278 @@
...
+DOI:
+  description: > 
+    DOI is a persistent Unique Digital Identifier with the form
+    https://doi.org/<PREFIX>/<SUFFIX> with the <PREFIX> identifying the DOI
+    registration authority and the <SUFFIX> identifying the dataset. The DOI should point to
+    a landing page for additional information about the dataset. DOIs are typically created by
+    the SPASE naming authority or archive.
+  default: null
+  derived: false
+  required: true # NOT Required in ISTP Guide (Recommended)
```
Alrobbertz: There are a lot of attributes here which are NOT required for ISTP, but are required for HERMES. I think I did a decent job adding comments on these cases. 

There are also som attributes that are marked required for HERMES, like Data_Level, which is not ISTP required, but it is used in our derivation functions for generating Logical_file_id.

imap_processing/cdf/config/shared/hermes_default_global_cdf_attrs_schema.yaml
```diff
@@ -0,0 +1,278 @@
...
+  validate: true
+  overwrite: true
+HERMES_version:
+  description: >
+    Version of `hermes_core` originally used to generate the given CDF File
+  default: null
+  derived: true
+  required: true # NOT Required in ISTP Guide (Derived)
+  validate: false
+  overwrite: false
```
Alrobbertz: It might be better to have an `imap_processing_version` attribute here? This is so we can tell which version of our software was used to generate different data products.

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: Thanks @Alrobbertz! I'll take them out of this PR and we can think about adding it in the generic version.

imap_processing/cdf/config/shared/hermes_default_global_cdf_attrs_schema.yaml
```diff
@@ -0,0 +1,278 @@
+CDF_Lib_version:
+  description: > 
+    Version of the CDF Binaries library used to generate the CDF File
+  default: null
+  derived: true
```
Alrobbertz: You may also want to remove the derived keys here if you're not going to auto-generate any of the metadata attributes. 

I was thinking of _not_ including these in the "generalized" version of the schema, and only including them in sub-classed versions of the schema.

imap_processing/cdf/config/shared/hermes_default_global_cdf_attrs_schema.yaml
maxinelasp: Ok, thank you! I think I'll leave the variables that are not required for ISTP in, since we probably will use some of them as well. But I will remove the obvious HERMES-specific ones and rename the file to be more generic.

imap_processing/cdf/config/shared/hermes_default_global_cdf_attrs_schema.yaml
maxinelasp: That's something that should go into a separate IMAP global CDF schema (not yet created)

imap_processing/cdf/config/shared/hermes_default_global_cdf_attrs_schema.yaml
greglucas: Is there a value in having multiple global_cdf_attrs schemas that can be mixed/matched as missions need?

For example, have an `ISTP_metadata_schema.yaml`, `SOLARNET_metadata_schema.yaml`, `IMAP_metadata_schema.yaml`, `SWxSOC_metadata_schema.yaml`, ... Where you would have ISTP / SOLARNET / external requirements able to be used by any mission and then each mission could pick and choose which individual metadata schemas they want to use and also add their own in their own repos so we don't blow up the main one with extra metadata for mission-specific things?

maxinelasp: Yeah, I do like that actually! I definitely think we should have mission specific schema as well as the global ones, but I like the idea of having multiple global ones too.

Alrobbertz: Yes, this is something we've been working on is "layering" multiple YAML files together. https://swxsoc.readthedocs.io/en/latest/generated/api/swxsoc.util.schema.SWXSchema.html#swxsoc.util.schema.SWXSchema

This is how I envisioned it to be integrated, although I haven't had to time to fully do integration for HERMES yet. https://github.com/HERMES-SOC/hermes_core/blob/04013bd3748309f0cf88ff924f7ae12b30ee57b2/hermes_core/util/schema.py

Alrobbertz: I'm also currently working on adding SOLARNET yaml files for FITS standards. Draft here: https://github.com/swxsoc/swxsoc/pull/5

maxinelasp:
pre-commit.ci autofix

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+import yaml
+
+DEFAULT_GLOBAL_CDF_ATTRS_FILE = "imap_default_global_cdf_attrs.yaml"
+DEFAULT_GLOBAL_CDF_ATTRS_SCHEMA_FILE = "default_global_cdf_attrs_schema.yaml"
+DEFAULT_VARIABLE_CDF_ATTRS_SCHEMA_FILE = "default_variable_cdf_attrs_schema.yaml"
+
+
+class CdfAttributeManager:
+    """
+    Class for creating and managing CDF attributes based out of yaml files.
```
tech3371, line 19: very helpful comments!

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+    This class is based on the HERMES SWxSOC project for managing CDF attributes, but
+    is intended to be a flexible and very lightweight way of managing CDF attribute
+    creation and validation.
+
+    To use, you can load one or many global and variable attribute files:
+
+    ```
+    cdf_attr_manager = CdfAttributeManager(data_dir)
+    cdf_attr_manager.load_global_attributes("global_attrs.yaml")
+    cdf_attr_manager.load_global_attrs("instrument_global_attrs.yaml")
```
greglucas:
```suggestion
    cdf_attr_manager.load_global_attributes("instrument_global_attrs.yaml")
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+
+class CdfAttributeManager:
+    """
+    Class for creating and managing CDF attributes based out of yaml files.
+
+    This class is based on the HERMES SWxSOC project for managing CDF attributes, but
+    is intended to be a flexible and very lightweight way of managing CDF attribute
+    creation and validation.
+
+    To use, you can load one or many global and variable attribute files:
```
greglucas, line 25: You might want to add a sentence on which way the update works for loading. Feel free to modify my suggestion or move it elsewhere.
```suggestion
    To use, you can load one or many global and variable attribute files
    (the latter files will overwrite duplicate entries in the previous ones):
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+    cdf_attr_manager.load_global_attributes("global_attrs.yaml")
+    cdf_attr_manager.load_global_attrs("instrument_global_attrs.yaml")
+    cdf_attr_manager.load_variable_attrs("variable_attrs.yaml")
+    ```
+
+    You can then get the global and variable attributes:
+
+    ```
+    # Instrument ID is optional for refining the attributes used from the file
+    global_attrs = cdf_attr_manager.get_global_attributes(instrument_id)
```
greglucas, line 44: Is `global_attrs` indexed by dataset after the `instrument_id`? Or is the dataset supposed to be a part of `instrument_id`?

It might be good to put a quick example here of what `global_attrs` looks like after the call.
```python
# global_attrs == {"raw": {...}}
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+    variable_attrs = cdf_attr_manager.get_variable_attributes(variable_name)
+    ```
+
+    The variable and global attributes are validated against the schemas upon calling
+    `get_global_attributes` and `get_variable_attributes`.
+
+    Attributes
+    ----------
+    source_dir: Path
```
greglucas:
```suggestion
    source_dir : Path
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+        The directory containing the schema and variable files - nominally config/
+    """
+
+    def __init__(self, data_dir: Path):
+        """
+        Initialize the CdfAttributeManager and read schemas from data_dir.
+
+        Parameters
+        ----------
+        data_dir: Path
```
greglucas: numpydoc should have a space before the colon as well.
```suggestion
        data_dir : Path
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+        with open(file_path) as file:
+            return yaml.safe_load(file)
+
+    def get_global_attributes(self, instrument_id=None) -> dict:
+        """
+        Generate a dictionary global attributes based off the loaded schema and attrs.
+
+        Validates against the global schema to ensure all required variables are
+        present. It can also include instrument specific global attributes if
+        instrumet_id is set.
```
greglucas, line 153: typo + We should highlight that instrument_id is a quantity. I don't remember if we are bolding these are double-backticking them.
```suggestion
        **instrument_id** is set.
```

laspsandoval (approved):

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+
+        If a logical_source_id is provided, the level and instrument specific
```
greglucas: Where would `logical_source_id` be provided from? Do you mean its in the variable files, or is this supposted to be `instrument_id`?

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
maxinelasp: The corresponding variables are uppercase in my code so I left it uppercase here. This won't end up in the CDF files, it's purely internal. In the CDF files the variables are lowercased!

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+                and attr_name in self.global_attributes[instrument_id]
+            ):
+                output[attr_name] = self.global_attributes[instrument_id][attr_name]
+            elif attr_schema["required"] and attr_name not in self.global_attributes:
+                # TODO throw an error
+                output[attr_name] = None
+
+        return output
+
+    def load_variable_attrs(self, file_name: str) -> None:
```
greglucas: I think we should spell it out and be consistent with `load_global_attributes()`
```suggestion
    def load_variable_attributes(self, file_name: str) -> None:
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        Add variable attributes for a given filename.
+
+        Parameters
+        ----------
+        file_name: str
```
greglucas:
```suggestion
        file_name : str
```

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
maxinelasp: I believe this was given to me either by the MAG team or by SPDF, because I wouldn't have picked that value!

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
maxinelasp: That is in the global attribute file, since it should be the same across the board for everyone. It's accessed via the same instance of a class as these variables, so they end up together in code for easy access.

imap_processing/cdf/imap_cdf_manager.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self):
+        """Set the path to the config directory."""
+        super().__init__(Path(__file__).parent / "config")
+
+    def add_instrument_global_attrs(self, instrument: str):
+        """Add instrument specific global attributes."""
+        # Looks for file named "imap_{instrument}_global_cdf_attrs.yaml"
+        self.load_global_attributes(f"imap_{instrument}_global_cdf_attrs.yaml")
+
+    def add_instrument_variable_attrs(self, instrument, level):
```
greglucas:
```suggestion
    def add_instrument_variable_attrs(self, instrument: str, level: str):
```

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -0,0 +1,58 @@
...
+    cdf_manager = CdfAttributeManager(Path(__file__).parent.parent / "config")
+
+    cdf_manager.load_global_attributes("imap_default_global_cdf_attrs.yaml")
+
+    # Expected failure: file_naming_convention is not in schema.
+    # for attr in cdf_manager.global_attributes.keys():
+    #     assert attr in cdf_manager.global_attribute_schema.keys()
+
+    assert (
+        cdf_manager.global_attributes["Mission_group"]
```
greglucas, line 17: I like this lookup style, thanks for adding it!

imap_processing/cdf/config/shared/default_global_cdf_attrs_schema.yaml
```diff
@@ -0,0 +1,238 @@
...
+  overwrite: false
+Discipline:
+  description: >
+    This attribute describes both the science discipline and sub discipline. For HERMES,
+    this value should always be "Space Physics>Magnetospheric Science."
+  default: Space Physics>Magnetospheric Science
+  required: true
+  validate: true
+  overwrite: false
+Generation_date:
```
tech3371, line 72: Do we need to add this here? Just curious because we use different filename convention than its default. We need to use `source_descriptor_datatype_yyyyMMdd_vNNN`
```suggestion
File_naming_convention:
     description: Filename convention
     default: source_datatype_descriptor_yyyyMMdd
     required: false
     validate: false
     overwrite: false
 Generation_date:
```

greglucas (approved):

imap_processing/cdf/config/shared/default_global_cdf_attrs_schema.yaml
```diff
@@ -0,0 +1,238 @@
...
+  description: >
+    Date stamps the creation of the file using the syntax yyyymmdd, e.g., "
+  default: null
+  required: false # NOT Required in ISTP Guide (Recommended)
+  validate: true
+  overwrite: true
```
tech3371, line 78: Can you add comment on what these values mean?

imap_processing/cdf/config/shared/default_variable_cdf_attrs_schema.yaml
```diff
@@ -0,0 +1,466 @@
...
+  - DISPLAY_TYPE
+  - FIELDNAM
+  - FILLVAL
+  - FORMAT
+  - LABLAXIS
+  - SI_CONVERSION
+  - UNITS
+  - VALIDMIN
+  - VALIDMAX
+  - VAR_TYPE
```
tech3371, line 449: Does this mean that if data is of `data` type, it needs these fields? Just curious

imap_processing/cdf/config/imap_mag_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,49 @@
+Descriptor: &desc
```
tech3371, line 1: Learned this new thing about way of reusing variables in yaml. cool

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
tech3371, line 1: to my previous question about if `data` type, which fields are required, I wonder if we can have some defaults if `data`, `support_data` or `metadata` and so on. Not for this PR but in the future.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Returns
+        -------
+        the dict representing the variable schema
+        """
+        default_schema_path = (
+            self.source_dir / "shared" / DEFAULT_VARIABLE_CDF_ATTRS_SCHEMA_FILE
+        )
+        # Load the Schema
+        return CdfAttributeManager._load_yaml_data(default_schema_path)
```
tech3371, line 110: curious why this was using `CdfAttributeManager` to load the schema and previous one uses `self._load_yaml_data`?

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def load_global_attributes(self, file_path: str):
```
tech3371, line 112: Is this function way to upload additional global attribute? If so, can we add a comment about this?

tech3371 (approved):
It looks great! I had few clarification questions but otherwise, it looks great.

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: Should be instrument_id

imap_processing/cdf/config/shared/default_global_cdf_attrs_schema.yaml
maxinelasp: Yes, in the IMAP specific global CDF attributes we do need to add file_naming_convention. That isn't an ISTP standard though, so it should go in a different file (eg `imap_global_cdf_attrs_schema.yaml`)

imap_processing/cdf/config/shared/default_variable_cdf_attrs_schema.yaml
maxinelasp: I believe yes

imap_processing/cdf/config/imap_mag_l1a_variable_attrs.yaml
maxinelasp: Yeah definitely that should be part of the validation step!

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: It is a static method, so either is valid BUT they should be the same. I'll change the other ones to use CdfAttributeManager instead of self.

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: Yes, will update the comment

PR #566, opened 2024-05-28
L1b outline for Ultra

laspsandoval:
# Change Summary

## Overview
This is an outline of what I will do for Ultra L1b. I will use the code developed by Greg to create "ultra_metadata.json". For now I am using "ultra_metadata_example.json". 

## New Files
- ultra_l1b.py
   - Added outline with TODOs of what I will do for Ultra L1b SIT-3
- ultra_metadata_example.json
   - Example of how I will use this json. It will be replaced by a final file that is created using the excel spreadsheet and will not be merged into the PR. **Please do not review.**

## Updated Files
- cli.py
   - Added description to the query. Without this change I get the name of the same file multiple times.
- test_ultra_l1a.py and ultra_utils.py
   - Nomenclature change
- __init__.py
   - Version formatting adjustment
 - ultra_l1a.py
   - Formatting change

## Testing
- test_ultra_l1b.py
   - Tests ultra_l1b.py
- Confirmed expected files appeared in AWS.


imap_processing/ultra/ultra_metadata_example.json
laspsandoval, line 1: Just a reminder this is not my final ultra_metadata.json. It is only used for the sake of creating a first draft of L1b processing. I will not merge this in.

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def ultra_l1b(data_dict: dict):
+    """
+    Process ULTRA L1A data into L1B CDF files at output_filepath.
+
+    Parameters
+    ----------
+    data_dict: : dict
+        Dependencies dictionary.
```
tech3371: Is this dictionary including the data itself and its dependent data(s)? If so, can we update this comment. Also, can you include information about what the key value is and value value is?

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Returns
+    -------
+    output_datasets : list of xarray.Dataset
+        List of xarray.Dataset
+    """
+    output_datasets = []
+    instrument_id = int(re.search(r"45|90", next(iter(data_dict.keys()))).group())
```
tech3371: This regex is assuming that the key(eg. filename prefix) has 45 and 90 in the name. I wonder if it's safe to do this check this way. This could be a question for future.

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -0,0 +1,85 @@
...
+    # Load metadata from the metadata file
+    with open(Path(__file__).parent.parent / "ultra_metadata_example.json") as f:
+        metadata = json.loads(f.read())[name]
+
+    epoch = dataset.coords["epoch"]
+
+    annotated_de_attrs = metadata["dataset_attrs"]
+
+    dataset = xr.Dataset(
+        coords={"epoch": epoch.astype("datetime64[ns]")},
```
tech3371: If I understand correctly, l1a `epoch` should have already been in `datetime46[ns]`. I don't think you need to do this again.

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -0,0 +1,85 @@
...
+        attrs=annotated_de_attrs,
+    )
+
+    dataset["x_front"] = xr.DataArray(
+        np.zeros(len(epoch), dtype=np.float32),
+        dims=["epoch"],
+        attrs=metadata["x_front"],
```
tech3371, line 44: Filling zeros for data is great. Is `x_front` an actual data variable of annotated DE?

tech3371 (commented):
It looks like a good start. Some of it may change based on tomorrow discussion.

imap_processing/ultra/l1b/ultra_l1b.py
laspsandoval: Yes. It is the event position on the front foil. L1b-L1c will have zeros. I am almost done with extended events so most of these will be filled soon, but I don't want to risk running out of time.

imap_processing/cli.py
```diff
@@ -229,6 +230,7 @@ def download_dependencies(self):
                     instrument=dependency["instrument"],
                     data_level=dependency["data_level"],
                     version=dependency["version"],
+                    descriptor=dependency["descriptor"],
```
greglucas, line 232: Do all instruments include a descriptor in their dependency list now?

imap_processing/cli.py
```diff
@@ -478,6 +480,14 @@ def do_processing(self, dependencies: list):
             datasets = ultra_l1a.ultra_l1a(dependencies[0])
             products = [write_cdf(dataset) for dataset in datasets]
             return products
+        elif self.data_level == "l1b":
+            data_dict = defaultdict(list)
```
greglucas: Below you are setting the field for data_dict to a dataset, you aren't appending to it, so I think this should just be a normal dictionary?
```suggestion
            data_dict = {}
```

imap_processing/tests/ultra/unit/test_ultra_l1b.py
```diff
@@ -0,0 +1,31 @@
...
+def test_create_dataset(mock_data_dict):
+    """Tests that dataset is created as expected."""
+    dataset = create_dataset(mock_data_dict, "imap_ultra_l1b_45annotated-de")
+
+    assert "epoch" in dataset.coords
+    assert dataset.coords["epoch"].dtype == "datetime64[ns]"
+    assert dataset.attrs["Logical_source"] == "imap_ultra_l1b_45annotated-events"
+    assert "x_front" in dataset.data_vars
+    assert dataset["x_front"].attrs["UNITS"] == "mm"
+    assert (dataset["x_front"].values == 0).all()
```
greglucas: This will test the shape as well.
```suggestion
    np.testing.assert_array_equal(dataset["x_front"], np.zeros(3))
```

imap_processing/tests/ultra/unit/test_ultra_l1b.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_create_dataset(mock_data_dict):
+    """Tests that dataset is created as expected."""
+    dataset = create_dataset(mock_data_dict, "imap_ultra_l1b_45annotated-de")
+
+    assert "epoch" in dataset.coords
+    assert dataset.coords["epoch"].dtype == "datetime64[ns]"
+    assert dataset.attrs["Logical_source"] == "imap_ultra_l1b_45annotated-events"
+    assert "x_front" in dataset.data_vars
```
greglucas: I think this is implied below because you're accessing `dataset["x_front"]`, so it would fail below if it wasn't present.

imap_processing/tests/ultra/unit/test_ultra_l1b.py
```diff
@@ -0,0 +1,31 @@
...
+    assert dataset["x_front"].attrs["UNITS"] == "mm"
+    assert (dataset["x_front"].values == 0).all()
+
+
+def test_ultra_l1b(mock_data_dict):
+    """Tests that L1b data is created."""
+    output_datasets = ultra_l1b(mock_data_dict)
+
+    assert len(output_datasets) == 1
```
greglucas, line 33: Do you want to say anything about the write_cdf / naming involved, or is that tested enough in the above create_dataset test?

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -0,0 +1,86 @@
...
+        Dictionary including the data itself and
+        its dependent data.
+
+    Returns
+    -------
+    output_datasets : list of xarray.Dataset
+        List of xarray.Dataset
+    """
+    output_datasets = []
+    instrument_id = int(re.search(r"45|90", next(iter(data_dict.keys()))).group())
```
greglucas: I'm not sure I follow exactly what this is doing, I think it is seeing whether 45 or 90 is in the first dictionary key? Could you add a note, or alternatively, could it be written `instrument_id = 45 if "45" in list(data_dict.keys())[0] else 90`?

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -0,0 +1,86 @@
...
+
+import json
+import logging
+import re
+from pathlib import Path
+
+import numpy as np
+import xarray as xr
+
+logger = logging.getLogger(__name__)
```
greglucas: Unused logger?
```suggestion
```

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -0,0 +1,86 @@
...
+
+def ultra_l1b(data_dict: dict):
+    """
+    Process ULTRA L1A data into L1B CDF files at output_filepath.
+
+    Parameters
+    ----------
+    data_dict: : dict
+        Dictionary including the data itself and
+        its dependent data.
```
greglucas: There is an extra colon here and the dictionary is already implied by the type annotation
```suggestion
    data_dict : dict
        The data itself and its dependent data.
```

imap_processing/cli.py
laspsandoval: Yes. They always have. It just wasn't being queried before.

imap_processing/tests/ultra/unit/test_ultra_l1b.py
```diff
@@ -0,0 +1,41 @@
...
+from imap_processing.ultra.l1b.ultra_l1b import create_dataset, ultra_l1b
+
+
+@pytest.fixture()
+def mock_data_dict():
+    epoch = np.array(
+        [760591786368000000, 760591787368000000, 760591788368000000],
+        dtype="datetime64[ns]",
+    )
+    data_dict = {"imap_ultra_l1a_45de": xr.Dataset(coords={"epoch": epoch})}
```
tech3371:
```suggestion
    data_dict = {"imap_ultra_l1a_45sensor-de": xr.Dataset(coords={"epoch": epoch})}
```

imap_processing/tests/ultra/unit/test_ultra_l1b.py
```diff
@@ -0,0 +1,41 @@
...
+    return data_dict
+
+
+def test_create_dataset(mock_data_dict):
+    """Tests that dataset is created as expected."""
+    dataset = create_dataset(mock_data_dict, "imap_ultra_l1b_45annotated-de")
```
tech3371:
```suggestion
    dataset = create_dataset(mock_data_dict, "imap_ultra_l1b_45sensor-de")
```

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -0,0 +1,81 @@
...
+    -------
+    dataset : xarray.Dataset
+        xarray.Dataset
+    """
+    # TODO: this will change to actual l1b data dictionary
+    # For now we are using it for retrieving the epoch.
+    dataset = data_dict["imap_ultra_l1a_45de"]
+
+    # Load metadata from the metadata file
+    with open(Path(__file__).parent.parent / "ultra_metadata_example.json") as f:
```
tech3371, line 29: I am assuming this file will change now based on yesterday discussion outcome. But if you want to do this update in future PR to keep this thing moving, that sounds good to me.

tech3371 (approved):

PR #567, opened 2024-05-29
Codice L1A updates for SIT-3

bourque:
# Change Summary

## Overview

This PR makes some more changes to the CoDICE L1a pipeline to add better support for products for SIT-3

## New Files

- `imap_processing/codice/packet_definitions/P_COD_LO_SW_PRIORITY_COUNTS.xml`
- `imap_processing/codice/packet_definitions/P_COD_LO_NSW_PRIORITY_COUNTS.xml`
   - Priority counts XTCE definition is now split into two, one for each Sunward and Non-sunward

## Deleted Files

- `imap_processing/codice/packet_definitions/P_COD_LO_PRIORITY_COUNTS.xml`
   - Priority counts XTCE definition is now split into two, one for each Sunward and Non-sunward

## Updated Files

- `docs/source/development-guide/style-guide/naming-conventions.rst`
   - Updated valid descriptors
   - Moved attribute definitions to be in alphabetical order
   - Added code to determine the XML to raw data file mapping based on descriptor instead of hard-coded
   - Various cleanup, updated TODOs, and defining the data arrays as arrays of `1`s to get things flowing for SIT-3
   - Added `CATDESC`s for a bunch of CDF attributes
- `imap_processing/codice/packet_definitions/*.xml`
   - Renamed files to be consistent with the valid descriptors expected for CoDICE
   - Updated XTCE files after being run through updated `space_packet_parser` version
- `imap_processing/tests/codice/data/imap_codice_l0_*.pkts`
   - Renamed simulated data files to reflect proper naming convention for L0 data
- `imap_processing/tests/codice/test_codice_l0.py`
   - Renamed L0 simulated data file to reflect proper naming convention for L0 data
   - Updated simulated data filenames and re-ordered to alphabetical order
- `tools/xtce_generation/xtce_generator_codice.py`
   - Reodered list of APIDs to be in increasing order


imap_processing/codice/codice_l0.py
```diff
@@ -42,7 +32,14 @@ def decom_packets(packet_file: Path) -> list:
     list : list
         all the unpacked data.
     """
+    descriptor = (
```
greglucas, line 38: There are a lot of split/replace going on here. Could you give a quick example of input -> output of what these transforms are doing.

imap_processing/codice/codice_l1a.py
```diff
@@ -296,21 +275,22 @@ def unpack_science_data(self, packets: list):
...
-        self.data = [
-            science_values[i : i + chunk_size] for i in range(0, num_bits, chunk_size)
-        ]
+        # science_values = packets[0].data["DATA"].raw_value
+        # num_bits = len(science_values)
+        # chunk_size = len(science_values) // self.num_counters
+        # self.data = [
+        #     science_values[i : i + chunk_size] for i in range(0, num_bits, chunk_size)
+        # ]
+        self.data = ["1" * 128 for _ in range(self.num_counters)]
```
greglucas: Are you making this all ones, 128 * num_counters length? Maybe: `["1"] * 128 * self.num_counters` here.

greglucas (commented):
Would it make sense to combine the packet definitions into a single "COMBINED" file? I think that is how we are doing it on some of the instruments. That would enable you to load a single XTCE file and parse all APIDs I think, but it would likely make the single XTCE much larger and harder to find things.

imap_processing/tests/codice/test_codice_l1a.py
greglucas, line 1: It might be worth appending some of the binary test files together to make sure you have two APIDs coming in the same file and can handle that case of multiple APIDs?

No need to do this here, but something to keep in mind.

bourque: Yes, good suggestion! I plan to change the CoDICE pipeline to be able to handle multiple-apids-per-file sometime post-SIT3 (for now I have been assuming one-packet-per-file since that is how I have been receiving simulated data). I will make note to add that test then.

bourque:
> Would it make sense to combine the packet definitions into a single "COMBINED" file? I think that is how we are doing it on some of the instruments. That would enable you to load a single XTCE file and parse all APIDs I think, but it would likely make the single XTCE much larger and harder to find things.

This might be worth trying, but regardless I would like to wait until after SIT-3 to tackle this. Given the copious amount of different CoDICE APIDs (13), the resulting XTCE file would indeed be quite massive. Something I will think about!

imap_processing/codice/codice_l1a.py
```diff
@@ -87,29 +89,23 @@ def __init__(self, table_id: int, plan_id: int, plan_step: int, view_id: int):
...
         """
         epoch = xr.DataArray(
-            [
-                calc_start_time(
-                    packet.data["ACQ_START_SECONDS"].raw_value,
-                    launch_time=np.datetime64("2010-01-01T00:01:06.184", "ns"),
-                )
-                for packet in packets
-            ],
+            [start_time],
```
tech3371, line 108: Was there reason for taking out rest of packet times and keeping only first packet time? Just curious. Is this for SIT-3 testing purpose?

tech3371 (approved):
I just had one question. Otherwise, it looks good.

imap_processing/codice/codice_l1a.py
bourque: Yeah, I think this will likely change post SIT-3 when I work on getting the pipeline to read in multiple APIDs. But right now the pipeline handles one-APID-per-file, so there is only one start time.

PR #570, opened 2024-05-29
CoDICE L1A cli

bourque:
# Change Summary

## Overview

This PR adds CoDICE L1A processing to the cli. I also made a few changes to the CoDICE pipeline to be more consistent with how other instruments are running their processing (namely passing in a file path to the L0 file instead of passing in already-decommed packets)

## Testing

This appears to be working for housekeeping data when I test this locally, but fails on other data products, I think because the `descriptor` field is not yet fully implemented as a parameter in `cli.py`.

### DIFF
--- a/imap_processing/codice/codice_l1a.py
+++ b/imap_processing/codice/codice_l1a.py
@@ -22,6 +22,7 @@ from imap_processing import imap_module_directory
 from imap_processing.cdf.global_attrs import ConstantCoordinates
 from imap_processing.cdf.utils import calc_start_time, write_cdf
 from imap_processing.codice import cdf_attrs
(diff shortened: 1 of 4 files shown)


imap_processing/codice/codice_l1a.py
```diff
@@ -349,13 +350,13 @@ def get_params(packet) -> tuple[int, int, int, int]:
     return table_id, plan_id, plan_step, view_id
 
 
-def process_codice_l1a(packets) -> xr.Dataset:
+def process_codice_l1a(file_path: str) -> xr.Dataset:
```
```suggestion
def process_codice_l1a(file_path: Path | str) -> xr.Dataset:
```

greglucas (approved):

PR #571, opened 2024-05-31
Dynamic version

greglucas:
# Change Summary

## Overview

This adds dynamic version resolution from the git tag/git commit.

* Add log messages during processing to indicate the version
* Add the version to `write_cdf()` so it is output in our saved datasets as a dataset attribute.

## New Dependencies

It adds a new build system dependency on `poetry-dynamic-versioning` that injects the metadata resolution during build-time but is not a package dependency.

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -10,13 +10,14 @@ import numpy as np
 import xarray as xr
 from cdflib.xarray import cdf_to_xarray, xarray_to_cdf
 
-from imap_processing import launch_time
+import imap_processing
 
 logger = logging.getLogger(__name__)
 
 
 def calc_start_time(
-    shcoarse_time: float, launch_time: Optional[np.datetime64] = launch_time
+    shcoarse_time: float,
+    launch_time: Optional[np.datetime64] = imap_processing.launch_time,
 ) -> np.datetime64:
     """Calculate the datetime64 from the CCSDS secondary header information.
 
@@ -129,6 +130,8 @@ def write_cdf(dataset: xr.Dataset):
     # Insert the final attribute:
     # The Logical_file_id is always the name of the file without the extension
     dataset.attrs["Logical_file_id"] = file_path.stem
+    # Add the processing version to the dataset attributes
+    dataset.attrs["ground_software_version"] = imap_processing.__version__
 
     # Convert the xarray object to a CDF
     xarray_to_cdf(
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -52,6 +52,7 @@ def _parse_args():
     --instrument "mag"
     --data-level "l1a"
     --start-date "20231212"
+    --end-date "20231212"
     --version "v001"
     --dependency "[
         {
@@ -101,6 +102,29 @@ def _parse_args():
     )
 
     parser = argparse.ArgumentParser(prog="imap_cli", description=description)
+    # TODO: Add version here and change our current "version" to "data-version"?
+    # parser.add_argument(
+    #     "--version",
+    #     action="version",
+    #     version=f"%(prog)s {imap_processing.__version__}",
+    # )
+    # Logging level
+    parser.add_argument(
+        "--debug",
+        help="Print lots of debugging statements",
+        action="store_const",
+        dest="loglevel",
+        const=logging.DEBUG,
+        default=logging.WARNING,
+    )
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        help="Add verbose output",
+        action="store_const",
+        dest="loglevel",
+        const=logging.INFO,
+    )
     parser.add_argument("--instrument", type=str, required=True, help=instrument_help)
     parser.add_argument("--data-level", type=str, required=True, help=level_help)
 
@@ -275,9 +299,15 @@ class ProcessInstrument(ABC):
         of new products (files).
         3. Post-processing actions such as uploading files to the IMAP SDC.
         """
+        logger.info(f"IMAP Processing Version: {imap_processing.__version__}")
+        logger.info(f"Processing {self.__class__.__name__} level {self.data_level}")
+        logger.info("Beginning preprocessing (download dependencies)")
         dependencies = self.pre_processing()
+        logger.info("Beginning actual processing")
         products = self.do_processing(dependencies)
+        logger.info("Beginning postprocessing (uploading data products)")
         self.post_processing(products)
+        logger.info("Processing complete")
 
     def pre_processing(self):
         """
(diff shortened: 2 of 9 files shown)


imap_processing/cdf/utils.py
```diff
@@ -128,6 +129,8 @@ def write_cdf(dataset: xr.Dataset):
     # Insert the final attribute:
     # The Logical_file_id is always the name of the file without the extension
     dataset.attrs["Logical_file_id"] = file_path.stem
+    # Add the processing version to the dataset attributes
+    dataset.attrs["imap_processing_version"] = imap_processing.__version__
```
greglucas: Open to other naming suggestions here for that dataset level attribute.

imap_processing/cli.py
```diff
@@ -101,6 +101,29 @@ def _parse_args():
     )
 
     parser = argparse.ArgumentParser(prog="imap_cli", description=description)
+    # TODO: Add version here and change our current "version" to "data-version"?
```
greglucas, line 104: I'm not sure how hard this would be to change, but most CLIs allow `cli --version` to get the cli version resolution. That would be nice here if possible, but figured I wouldn't change this right this moment.

imap_processing/tests/ultra/unit/test_ultra_l1a.py
```diff
@@ -219,7 +219,8 @@ def test_cdf_aux(
     dataset_aux = create_dataset({ULTRA_AUX.apid[0]: decom_ultra_aux})
     input_xarray_aux = load_cdf(test_data_path)
 
-    assert input_xarray_aux.attrs.keys() == dataset_aux.attrs.keys()
+    # write_cdf() injects some attributes that are not in the xarray
```
greglucas, line 222: We should really be testing explicit expectations I think, so it shouldn't matter if we are injecting extra attributes in write_cdf, we really want to know that the ones that were in the xarray dataset made it through to the output.

pyproject.toml
```diff
@@ -1,13 +1,14 @@
...
+build-backend = "poetry_dynamic_versioning.backend"
 
 [tool.poetry]
 name = "imap-processing"
-version = "0.2.0"
+version = "0.0.0"
 description = "IMAP Science Operations Center Processing"
 authors = ["IMAP SDC Developers <imap-sdc@lists.lasp.colorado.edu>"]
 readme = "README.md"
+include = ["imap_processing/_version.py"]
```
greglucas, line 12: Note that we will be producing a `_version.py` file now, but it is gitignored so that will be dynamically updating everytime we install but not committed to the repo.

pyproject.toml
```diff
@@ -1,13 +1,14 @@
 [build-system]
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning>=1.0.0,<2.0.0"]
+build-backend = "poetry_dynamic_versioning.backend"
 
 [tool.poetry]
 name = "imap-processing"
-version = "0.2.0"
+version = "0.0.0"
```
greglucas, line 8: This gets changed during build/release, but not committed to the repository.

imap_processing/cdf/utils.py
maxinelasp: GLOWS requested "ground_software_version", so maybe along those lines?

imap_processing/cli.py
maxinelasp: I like that

pyproject.toml
maxinelasp: Add a comment that this is updated during release?

maxinelasp (approved):

imap_processing/cdf/utils.py
greglucas: :+1: That would be fine with me too. The only issue I'd see is if we want to put the versions of multiple software packages being used in, like do we want to start including `imap-data-access-version`, or something else informative too?

pyproject.toml
greglucas: Added!

imap_processing/cli.py
tech3371: To your question about how hard it is to make this, we will need to change here and update batch starter lambda and its test to use `data-version` instead of `version`. I think it will be ok.

pyproject.toml
tech3371: Just curious. How often do we need to install to update this? Is this done per release?

tech3371 (approved):
It makes sense on high-level. I may have questions later on.

pyproject.toml
greglucas: Every time you install the project it will update this from the latest tag/commit for you. You never have to touch this, it is all automated. If we tag on release, it should pick up that tag and set that as the version number.

imap_processing/cli.py
greglucas: OK, do you want me to do that now, or hold off until after SIT-3 then?

imap_processing/cdf/utils.py
maxinelasp: Probably I would prefer "ground_software_version", since presumably updating the `imap-data-access` dependency would result in a new `ground_software_version`. I'd prefer it to be super obvious to people what the main version they should care about is, and we can add more specific info later/if needed.

imap_processing/cli.py
maxinelasp: I thought we already were using `data-version`? If we are using `version` then yeah leave it for later, removing/changing functionality from the API is annoying

imap_processing/cdf/utils.py
greglucas: I would actually prefer to call it `imap_processing_version` then because that is super explicit that is this package's version we are referring to. IMO `ground_software_version` is more encompassing and could mean container hash, imap_processing version hash, something else we manually set/stamp ourselves as this is a version.

@maxinelasp I see that you put `ground_software_version` in GLOWS. Would this mess things up if we also injected an `imap_processing_version` (or overwrite the `ground_software_version` metadata you're already declaring) here?

maxinelasp: It wouldn't break anything, but I would have to ask and see if they're ok with the name change or if I would need to duplicate that 🤷

maxinelasp: But it's fine with me

bourque (approved):
Looks good to me!

imap_processing/cdf/utils.py
greglucas: `ground_software_version` was voted the winner.

@maxinelasp, you put in GLOWS metadata that it is of the format `vXXX`, with the dynamic versioning it will actually be whatever the current tag is: `v0.1.2` or if you're in development it will show you commit info as well: `0.2.0.post173.dev0+cde8921` Is that OK for your situation, or do we need to update the GLOWS documentation/attributes to account for this?

maxinelasp: Nope, that should work, and I probably just got confused about which version I was using there. I would expect it to be the semantic version of the repo.

PR #572, opened 2024-06-03
MNT: Minimize warnings and prepare for the future

greglucas:
# Change Summary

## Overview

There were over 4000+ warnings thrown before, so this is to remove those and make sure the useful warnings do appear and don't get lost in the noise.

- This ignores warnings that cdflib is throwing and have already been fixed upstream.
- Hi: change from dims -> sizes as xarray suggests
- Lo: cast to nanosecond precision when creating xarray datetime64


bourque (approved):
LGTM

PR #581, opened 2024-06-04
Sit 3 Ultra L1b and L1c

laspsandoval:
# Change Summary

## Overview
Added code to process L1b and L1c for Ultra

## New Files
**New code for creating datasets and tests:**
ultra_l1b.py
ultra_l1c.py
test_ultra_l1b.py
test_ultra_l1c.py

**template algorithm code for l1b:**
badtimes.py
cullingmask.py
de.py
extendedspin.py

**template algorithm code for l1c:**
histogram.py
pset.py

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
- Adds ultra global attributes

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
- Add ultra variables

imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
- Add ultra variables

## Deleted Files
- ultra_metadata_example.json
   - Switched to yaml

## Updated Files
- imap_processing/cli.py
   - Add processing code for l1c, l1b
- imap_processing/test_cli.py
   - Add test code for  l1c, l1b

## Testing
I will test that batch job functions properly after I push the image to the ECR.


imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,151 @@
...
+  Ultra>IMAP Ultra-Energy (IMAP-Ultra) Energetic Neutral Atom Imager
+TEXT: &txt >
+  Ultra captures images of very energetic neutral atoms, particularly hydrogen (H) atoms,
+  produced in the solar system at the heliosheath, the region where the solar wind slows,
+  compresses, and becomes much hotter as it meets the interstellar medium (ISM).
+  The instrument also measures the distribution of solar wind electrons and protons, and
+  the magnetic field. See https://imap.princeton.edu/instruments/ultra for more details.
+
+imap_ultra_l1b_45sensor-de:
+    Descriptor: 45sensor-de
```
tech3371: I think this is suppose to be this in this line and in others as well. @maxinelasp is that right?
```suggestion
    Descriptor: *desc
```

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,151 @@
...
+    TEXT: *txt
+    Data_level: 1B
```
tech3371, line 14: Interesting because `Data_type` is suppose to be `<datalevel>_<descriptor>` as well. Looks like they have two places to store data level.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,310 @@
...
+  # TODO: come back to format
+  UNITS: ns
+
+eventtype:
+  <<: *default
+  CATDESC: Eventtype (1-2 top and bottom stop types; 8-15 ssd stop types)
+  FIELDNAM: eventtype
+  LABLAXIS: eventtype
+  # TODO: come back to format
+  UNITS: ' '
```
greglucas: Can this be an empty string, or does it have to have a space in it?
```suggestion
  UNITS: ''
```

imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
```diff
@@ -0,0 +1,134 @@
...
+  # TODO: come back to format
+  UNITS: degrees
+
+esa_step:
+  <<: *default
+  CATDESC: ESA step (0-10), nominally 9, but possibly 8 or 10, 0 is likely a background test. Energy bin.
+  FIELDNAM: esa_step
+  LABLAXIS: esa step
+  # TODO: come back to format
+  UNITS: " "
```
greglucas, line 38: Here you use double quotes, in the other file you use single quotes. I don't think it matters which one here, but it'd be nice to be consistent.

imap_processing/tests/ultra/unit/test_ultra_l1b.py
```diff
@@ -6,18 +6,41 @@
 
 
 @pytest.fixture()
-def mock_data_dict():
+def mock_data_l1a_dict():
+    # Create sample data for the xarray Dataset
+    epoch = np.arange(
+        "2024-02-07T15:28:37", "2024-02-07T16:24:50", dtype="datetime64[s]"
+    )[:5]
```
greglucas: Can you make the array length-5 instead of creating extras and then truncating?
```suggestion
    epoch = np.arange(
        "2024-02-07T15:28:37", "2024-02-07T15:28:42", dtype="datetime64[s]"
    )
```

imap_processing/tests/ultra/unit/test_ultra_l1b.py
```diff
@@ -26,13 +49,33 @@ def test_create_dataset(mock_data_dict):
...
     assert (
         output_datasets[0].attrs["Logical_source_description"]
-        == "IMAP-Ultra Instrument Level-1B Annotated Event Data."
+        == "IMAP-Ultra Instrument Level-1B Extended Spin Data."
     )
+
+
+def test_ultra_l1b_error(mock_data_l1a_dict):
+    """Tests that L1a data throws an error."""
+    data_dict = mock_data_l1a_dict.copy()
```
greglucas: I don't think you need this because your fixture should be creating a new mock dataset for every test. If this is needed then we should look at refactoring the tests somehow because there will be action-at-a-distance happening depending on what order the tests are run and very confusing debugging/test assertions.
```suggestion
```

imap_processing/ultra/l1b/badtimes.py
```diff
@@ -0,0 +1,31 @@
...
+    badtimes_dict["spin_number"] = extended_spin_dict["spin_number"]
+    badtimes_dict["spin_start_time"] = extended_spin_dict["spin_start_time"]
+    badtimes_dict["avg_spin_period"] = extended_spin_dict["avg_spin_period"]
+    badtimes_dict["rate_start_pulses"] = extended_spin_dict["rate_start_pulses"]
+    badtimes_dict["rate_stop_pulses"] = extended_spin_dict["rate_stop_pulses"]
+    badtimes_dict["rate_coin_pulses"] = extended_spin_dict["rate_coin_pulses"]
+    badtimes_dict["rate_processed_events"] = extended_spin_dict["rate_processed_events"]
+    badtimes_dict["rate_rejected_events"] = extended_spin_dict["rate_rejected_events"]
+
+    return badtimes_dict
```
greglucas: I'm not following why you are adding all of the copy logic manually. This is a lot of extra code IMO. I don't think we need placeholders at this point. I'm also not following why it is a dict and not a dataset or an array?

```suggestion
    # TODO: filter out bad times from the extended_spin_dict entries
    return extended_spin_dict.copy()
```

imap_processing/ultra/l1b/extendedspin.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def calculate_extendedspin(data_dict):
+    """
+    Create dictionary with defined datatypes for Extended Spin Data.
+
+    Parameters
+    ----------
+    data_dict: : dict
+        L1a data dictionary.
```
greglucas: I'd suggest passing in the dataset of interest directly. You are using only one key `"imap_ultra_l1a_45sensor-rates"` below, so passing that dataset in would make it a bit clearer IMO.

imap_processing/ultra/l1b/extendedspin.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Returns
+    -------
+    extendedspin_dict : dict
+        Dictionary containing the data.
```
greglucas: Can you make this a dataset instead of a dictionary? Each of your keys below would be separate data arrays.

Could also loop over the float and uint values.
```python
for name in ["spin_number", "spin_start_time", ...]:
    da = xr.DataArray((name, epoch, np.zeros(len(epoch), dtype=np.uint64)))
```

imap_processing/ultra/l1b/extendedspin.py
```diff
@@ -0,0 +1,39 @@
...
+    extendedspin_dict["epoch"] = epoch
+    extendedspin_dict["spin_number"] = np.zeros(len(epoch), dtype=np.uint64)
+    extendedspin_dict["spin_start_time"] = np.zeros(len(epoch), dtype=np.float64)
+    extendedspin_dict["avg_spin_period"] = np.zeros(len(epoch), dtype=np.float64)
+    extendedspin_dict["rate_start_pulses"] = np.zeros(len(epoch), dtype=np.float64)
+    extendedspin_dict["rate_stop_pulses"] = np.zeros(len(epoch), dtype=np.float64)
+    extendedspin_dict["rate_coin_pulses"] = np.zeros(len(epoch), dtype=np.float64)
+    extendedspin_dict["rate_processed_events"] = np.zeros(len(epoch), dtype=np.float64)
+    extendedspin_dict["rate_rejected_events"] = np.zeros(len(epoch), dtype=np.float64)
+    extendedspin_dict["quality_hk"] = np.zeros(len(epoch), dtype=np.uint16)
```
greglucas: Are you getting more than 255 quality flags down? I'd suggest `uint8` if possible.

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -15,67 +20,92 @@ def create_dataset(data_dict, name):
...
-
-    for name in l1b_products:
-        # TODO: perform l1b calculations here.
-        # For now we are using the L1A data (data_dict) as a placeholder.
-        dataset = create_dataset(data_dict, name)
+    if f"imap_ultra_l1a_{instrument_id}sensor-rates" in data_dict:
+        extendedspin_dict = calculate_extendedspin(data_dict)
+        extendedspin_dataset = create_dataset(
+            extendedspin_dict, f"imap_ultra_l1b_{instrument_id}sensor-extendedspin"
+        )
```
greglucas: I see now, you're calling your create_dataset function here from those dictionaries. Can you pull this `create_dataset` call into the `calculate_extended_spin` function so you are going dataset -> dataset in that routine?

imap_processing/ultra/l1c/pset.py
```diff
@@ -0,0 +1,131 @@
...
+import numpy as np
+
+
+def calculate_pset_exposure(l1b_de_dict, l1b_extendedspin_dict, l1b_cullingmask_dict):
+    """
+    Create dictionary with defined datatype for Pointing Set Grid Data.
+
+    Parameters
+    ----------
+    l1b_de_dict: : dict
```
greglucas: Extra colons on all of these. Also suggest dataset
```suggestion
    l1b_de_dict : dict
```

imap_processing/ultra/l1c/ultra_l1c.py
```diff
@@ -0,0 +1,133 @@
...
+    output_datasets : list of xarray.Dataset
+        List of xarray.Dataset
+    """
+    data_dict = {}
+    for dependency in dependencies:
+        dataset = load_cdf(dependency, to_datetime=True)
+        data_dict[dataset.attrs["Logical_source"]] = dataset
+
+    output_datasets = []
+    instrument_id = 45 if "45" in next(iter(data_dict.keys())) else 90
```
greglucas:
```suggestion
    instrument_id = 45 if "45" in data_dict.keys()[0] else 90
```

imap_processing/ultra/l1c/ultra_l1c.py
```diff
@@ -0,0 +1,133 @@
...
+    ):
+        histogram_dict = calculate_histogram(
+            data_dict[f"imap_ultra_l1a_{instrument_id}sensor-histogram"],
+            data_dict[f"imap_ultra_l1b_{instrument_id}sensor-cullingmask"],
+        )
+        histogram_dataset = create_dataset(
+            histogram_dict, f"imap_ultra_l1c_{instrument_id}sensor-histogram"
+        )
+        output_datasets.append(histogram_dataset)
+    elif (
```
greglucas, line 32: What happens if you get all 5 of these datasets passed to you at once? The first if will always hit, but you also want to process these files too right? I'm wondering if this should just be a separate `if` block

greglucas (commented):
My overarching comment would be to try and go to datasets sooner if possible and pass those around directly to your functions rather than dictionaries.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
laspsandoval: Bryan Harter says it needs a space.

greglucas: Bummer, another CDF annoyance for me to live with ;) Thanks for clarifying!

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
maxinelasp: I think this descriptor should be something like `ULTRA45> IMAP Ultra-Energy Energetic Neutral Atom Imager` or something like that?[ ISTP standard](https://spdf.gsfc.nasa.gov/istp_guide/attributes.html#Descriptor)

imap_processing/ultra/l1c/ultra_l1c.py
laspsandoval: This was a ruff fix. I'll try it your way, but I might have to do it ruff's way.

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,151 @@
...
+    Data_level: 1C
+    Data_type: L1C_Sensitivity>Level-1C Counts
+    Logical_source: imap_ultra_l1c_45sensor-pset-counts
+    Logical_source_description: IMAP-Ultra Instrument Level-1C Pointing Set Grid Counts Data.
+
+imap_ultra_l1c_90sensor-pset-counts:
+    Descriptor: 90sensor-pset-counts
+    TEXT: *txt
+    Data_level: 1C
+    Data_type: L1C_Sensitivity>Level-1C Counts
```
maxinelasp: I also think this data_type might need to include a 90/45 distinction. I also think they will need to be distinct? I don't remember what SPDF told me about it, but I think they need to be very specific.

imap_processing/ultra/l1c/ultra_l1c.py
laspsandoval: There shouldn't be a time in which the batch job command has all 5 dependencies. It would be in two separate jobs.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,310 @@
...
+
+quality_instruments:
+  <<: *default
+  CATDESC: Spin filter derived from instruments other than Ultra. Bitwise flagging used to filter the spin.
+  FIELDNAM: quality_instruments
+  LABLAXIS: quality instruments
+  # TODO: come back to format
+  UNITS: ' '
```
maxinelasp: Nice, I think this whole file is very clean which I love to see!

tech3371 (commented):
I looked at everything at very high level. The logics flows and I see what you are doing. I really think we should ask Ultra to think about combining data products. As you are aware, there a lot of duplicates data variable in their data products and it would be very beneficial to combine for both us and users.

imap_processing/ultra/l1b/cullingmask.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Returns
+    -------
+    cullingmask_dict : dict
+        Dictionary containing the data.
+    """
+    cullingmask_dict = {}
+
+    # Placeholder for bitwise filtering based on quality flags
+    cullingmask_dict["epoch"] = extended_spin_dict["epoch"]
```
maxinelasp: Would it be reasonable to make these dictionaries dataclasses instead? You could subclass to include different or extended data, or to use them for different cases, or even just to distinguish them. Basically make all these keys attributes in the dataclass?

 I know this is just a placeholder so it's possible that suggestion is overkill, but I find dataclasses provide some abstraction so I can easily go back in and add logic. For example, I could just start by initializing the dataclass and moving that around, but then later, add in logic to cull bad datapoints, without needing to update everywhere that the dataclass is used.

imap_processing/ultra/l1b/de.py
```diff
@@ -0,0 +1,53 @@
+"""Calculates Annotated Direct Events."""
+
+import numpy as np
+
+
+def calculate_de(data_dict):
+    """
+    Create dictionary with defined datatypes for Direct Event Data.
```
maxinelasp: Similar point - if you're creating a dictionary with the same keys every time, a dataclass will provide some nice convenience factors, while also being more stable and safe than a dictionary (by providing tracing back to what the keys should be, providing additional documentation for each key, and by disallowing someone to update the dictionary later.) Feel free to message me on slack for a zoom call if you want to discuss.

maxinelasp: This would be instead of the dataset suggestion by Greg - a dataset here would also function in a very similar way so it would also solve a lot of the same issues as a dataclass will. So, it comes down to you being the expert to figure out which way works the best.

maxinelasp (commented):
I had a few suggestions. I also have a halfway-finished alternative way of converting one dataset to another that I am hoping to share with the team tomorrow, which could be a pretty natural fit for here and would remove some of the dataset -> dictionary -> dataset complexities. I'm happy to explain that to you if that sounds interesting.

imap_processing/ultra/l1b/de.py
laspsandoval: @maxinelasp I saw your post after I had changed it based on Greg's comment of a dataset. What I think I will do is leave it as a dataset for now, but if you use a dataclass for MAG I'll take a look and maybe make modifications.

imap_processing/ultra/l1b/cullingmask.py
laspsandoval: I love this idea. Let's discuss. Do you plan on using it for MAG? I always like seeing examples to get ideas.

maxinelasp: Yup, definitely happy to talk about it. I am actually doing something very interesting for GLOWS that would line up quite well with what you have here - if we plan to meet end of day today or tomorrow, it should be complete enough for you to get a full idea of how it works!

imap_processing/ultra/l1b/de.py
maxinelasp: A dataset functions very similarly in this case, and in some ways is a better match than a dataclass. I can walk you though my style of using a dataclass, and then you can decide which you like better. 

A dataset is more closely aligned with the inputs and outputs, which is nice. A dataclass you have a bit more control over the structure and on expanding the variables (or, including variables that are not in the output, things like that) so it does come down to a style choice at that point. Both are better than dicts!

laspsandoval:
Verified that it works in batch.

imap_processing/ultra/l1c/histogram.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Returns
+    -------
+    dataset : xarray.Dataset
+        Dataset containing the data.
+    """
+    histogram_dict = {}
+
+    # Placeholder for calculations
+    dataset = data_dict["imap_ultra_l1a_45sensor-histogram"]
```
greglucas: Here you are hardcoding "45sensor", but you use `instrument_id` to pass in, so I think you need to handle both 45 and 90.

More explicitly though I think you could pass in the `histogram_dataset` here and rely on the "get" logic out in the other calling function.

imap_processing/ultra/l1c/ultra_l1c.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    output_datasets = []
+    instrument_id = 45 if any("45" in key for key in data_dict.keys()) else 90
+
+    if (
+        f"imap_ultra_l1a_{instrument_id}sensor-histogram" in data_dict
+        and f"imap_ultra_l1b_{instrument_id}sensor-cullingmask" in data_dict
+    ):
+        histogram_dataset = calculate_histogram(
+            data_dict,
```
greglucas:
```suggestion
            data_dict[f"imap_ultra_l1a_{instrument_id}sensor-histogram"],
```

imap_processing/ultra/l1c/ultra_l1c.py
```diff
@@ -0,0 +1,50 @@
...
+    ):
+        pset_dataset = calculate_pset(
+            data_dict, f"imap_ultra_l1c_{instrument_id}sensor-pset"
+        )
+
+        output_datasets.extend(
+            [
+                pset_dataset,
+            ]
+        )
```
greglucas: You are doing append/extend on single iteration items. You can just set the object directly to the list you want and I think it will be more clear. You also use `append` above this, so if you don't want to set explicitly then I'd suggest using `append` here insted of `extend`
```suggestion
        output_datasets = [pset_dataset]
```

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,95 @@
+# Define descriptors
```
tech3371: remove?

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,310 @@
...
+  # Assumed values for all variable attrs unless overwritten
+  DEPEND_0: epoch
+  DISPLAY_TYPE: time_series
+  FILLVAL: -9223372036854775808
+  FORMAT: I12
+  VALIDMIN: -9223372036854775808
+  VALIDMAX: 9223372036854775807
+  VAR_TYPE: data
+
+epoch:
```
tech3371, line 11: cool. Others were wondering how to do this.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,310 @@
+default_attrs: &default
+  # Assumed values for all variable attrs unless overwritten
+  DEPEND_0: epoch
+  DISPLAY_TYPE: time_series
+  FILLVAL: -9223372036854775808
+  FORMAT: I12
```
tech3371: Based on below min and max value, it should be `I19`

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,310 @@
...
+  VALIDMIN: -9223372036854775808
```
tech3371: Should min be `0`?

tech3371 (approved):
Thank you for testing in Batch! That's a huge step. You can address those minor comments in future PR.

laspsandoval:
Will add more data structure in the next PR.

PR #582, opened 2024-06-04
Move CoDICE to YAML-based CDF attribute definition

bourque:
# Change Summary

## Overview

This PR replaces the dataclasses-based CDF attribute definitions with the new YAML-based approach introduced in #564 

## New Files

- `imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml`
  - YAML files to define CoDICE CDF attributes 

## Deleted Files

  - Removed file holding the dataclasses-based definitions

## Updated Files

  - Updates to utilize the new YAML-based approach


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,65 @@
...
+    <<: *default
+    CATDESC: ElectroStatic Analyzer Energy Values
+    FIELDNAM: ESA Voltage
+    FORMAT: I19
+    LABLAXIS: ESA V
+    UNITS: V
+    VALIDMIN: 0
+    VAR_TYPE: support_data
+
+# TODO: Define CATDESCs and FIELDNAMs for metadata fields, e.g.:
```
bourque: I have a whole bunch of these to add so I will do that sometime later

imap_processing/codice/codice_l1a.py
```diff
@@ -115,12 +121,15 @@ def create_science_dataset(self, start_time: np.datetime64) -> xr.Dataset:
             np.arange(self.num_energy_steps),
             name="energy",
             dims=["energy"],
-            attrs=cdf_attrs.energy_attrs.output(),
+            attrs=cdf_attrs.variable_attributes["energy_attrs"],
```
maxinelasp: Similar to my comment to Sean, the schema validation step is run when the attributes are retrieved - so they should be accessed using `get_variable_attributes` instead. Alternatively, a property would also solve this, and eliminate a common source of confusion - so feel free to update `get_variable_attributes` to be a python property instead.

imap_processing/codice/codice_l1a.py
```diff
@@ -104,6 +105,11 @@ def create_science_dataset(self, start_time: np.datetime64) -> xr.Dataset:
         xr.Dataset
             ``xarray`` dataset containing the science data and supporting metadata
         """
+        cdf_attrs = ImapCdfAttributes()
+        cdf_attrs.add_instrument_global_attrs("codice")
+        cdf_attrs.add_instrument_variable_attrs("codice", "l1a")
+
+        # TODO: Is there a way to get the attrs from the YAML-based method?
```
bourque, line 112: It wasn't quite clear to me how I might get the same information as is defined in `ConstantCoordinates.EPOCH` from the YAML files

imap_processing/codice/utils.py
```diff
@@ -140,7 +144,10 @@ def create_hskp_dataset(packets) -> xr.Dataset:
 
     dataset = xr.Dataset(
         coords={"epoch": epoch},
-        attrs=cdf_attrs.l1a_hskp_attrs.output(),
+        attrs={
+            **cdf_attrs.get_global_attributes(),
+            **cdf_attrs.global_attributes["imap_codice_l1a_hskp"],
```
bourque: Is there a better way to do this? I want to be able to get one dictionary that contains the default global attributes (i.e. `Project`, `Source_name`, etc.) with the instrument- and data product-specific attributes (i.e. `Data_type`, `Logical_source`, etc.)

imap_processing/codice/codice_l1a.py
```diff
@@ -115,12 +121,15 @@ def create_science_dataset(self, start_time: np.datetime64) -> xr.Dataset:
...
+            attrs=cdf_attrs.variable_attributes["energy_attrs"],
         )
 
         dataset = xr.Dataset(
             coords={"epoch": epoch, "energy": energy_steps},
-            attrs=self.cdf_attrs.output(),
+            attrs={
+                **cdf_attrs.get_global_attributes(),
+                **cdf_attrs.global_attributes[self.dataset_name],
+            },
```
maxinelasp: If you just pass `self.dataset_name` into `get_global_attributes` it will add those attributes into the returned dictionary so you can remove this line. 
```suggestion
            attrs=cdf_attrs.get_global_attributes(self.dataset_name)
```

imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,61 @@
...
+    Data_type: L1A_hskp->Level-1A Housekeeping Data
+    Logical_source: imap_codice_l1a_hskp
+    Logical_source_description: IMAP Mission CoDICE Instrument Level-1A Housekeeping Data.
+
+imap_codice_l1a_lo_sw_angular_counts:
+    Descriptor: *desc
+    TEXT: *txt
+    Data_type: L1A_lo-sw-angular-counts->Level-1A Lo Sunward Angular Counts Data
+    Logical_source: imap_codice_l1a_lo-sw-angular-counts
+    Logical_source_description: IMAP Mission CoDICE Instrument Level-1A Lo Sunward Angular Counts Data.
```
tech3371: It flows better for me but what you had was good too.
```suggestion
    Logical_source_description: IMAP Mission CoDICE Lo Level-1A Sunward Angular Counts Data.
```

imap_processing/codice/codice_l1a.py
```diff
@@ -132,28 +141,27 @@ def create_science_dataset(self, start_time: np.datetime64) -> xr.Dataset:
                 -1, self.num_energy_steps
             )
 
-            counter_attrs = cdf_attrs.counters_attrs
-            counter_attrs["FIELDNAM"] = fieldname
-            counter_attrs["CATDESC"] = catdesc
+            cdf_attrs.variable_attributes["counters_attrs"]["FIELDNAM"] = fieldname
```
maxinelasp: You shouldn't be updating the dictionary like this - any logic for setting attributes should stem from the YAML files. Don't make me regret using python dicts! If you absolutely need to do it like this, create a copy of the variable attributes file and update that one instead of going back into the cdf_attrs instance.

maxinelasp (changes requested):
Overall looks quite good, with the caveat that I only checked the format of the YAML files and not the contents. 

Mostly you need to correct a few places where you're accessing attributes directly and not with the `get` functions that validate before outputting. This is showing me that the API for that class needs to be changed and Greg was correct, so it should be a property instead of a get function.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,65 @@
...
+    VALIDMIN: 0
+    VALIDMAX: 8388607  # max value for a signed 24-bit integer
+
+energy_attrs:
+    <<: *default
+    CATDESC: Energy per charge (E/q) sweeping step
+    FIELDNAM: Energy step
+    FORMAT: I3
+    LABLAXIS: energy
+    UNITS: dN
```
```suggestion
    UNITS: ' '
```

imap_processing/codice/codice_l1a.py
```diff
@@ -104,6 +105,11 @@ def create_science_dataset(self, start_time: np.datetime64) -> xr.Dataset:
         xr.Dataset
             ``xarray`` dataset containing the science data and supporting metadata
         """
+        cdf_attrs = ImapCdfAttributes()
+        cdf_attrs.add_instrument_global_attrs("codice")
+        cdf_attrs.add_instrument_variable_attrs("codice", "l1a")
```
tech3371, line 110: It's great to see how easy of transition it is. I was concerned about that! Nice job @maxinelasp and all!

imap_processing/codice/codice_l1a.py
maxinelasp: Yeah looking at it now I might have... removed it? Or something? I remember moving that over but it's not in the files... :( 
Anyway, it should probably be in either its own file or in some kind of default IMAP variable attrs yaml file.

imap_processing/codice/utils.py
maxinelasp: Yes there is, made a comment

imap_processing/codice/codice_l1a.py
tech3371: This is interesting. I looked in HERMES repo and having hard time finding it.

tech3371 (approved):
looks great!

imap_processing/codice/codice_l1a.py
tech3371: Could this be what we need to use for `EPOCH`? https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/cdf/config/shared/default_variable_cdf_attrs_schema.yaml#L5

maxinelasp: Yeah that's the schema for it, but we still need to define the attribute somewhere. I'm not sure where it should live, maybe even in its own file? It's such a special case

imap_processing/codice/codice_l1a.py
bourque: Agreed! I thought it was really straightforward!

imap_processing/codice/codice_l1a.py
bourque: Ah, perfect!

imap_processing/codice/codice_l1a.py
bourque: So instead I could define a new variable attribute for each "counter" I have here (there are 16 total in this case) that inherits the `counters_attrs` and just switches out the `CATDESC` and `FIELDNAM`. Something like this:

```yaml
counters_attrs: &counters
    <<: *default
    CATDESC: Fill in at creation
    DEPEND_0: epoch
    DEPEND_1: energy
    DISPLAY_TYPE: time_series
    FIELDNAM: Fill in at creation
    LABL_PTR_1: energy
    UNITS: counts
    VALIDMIN: 0
    VALIDMAX: 8388607  # max value for a signed 24-bit integer
    
counter1_attrs:
    <<: *counters
    CATDESC: catdesc for counter 1 of 16
    FIELDNAM: fieldname for counter 1 of 16
    
counter2_attrs:
    <<: *counters
    CATDESC: catdesc for counter 2 of 16
    FIELDNAM: fieldname for counter 2 of 16

...etc
```

maxinelasp: Yup, exactly. All the data should be in the YAML files, nothing added later.

imap_processing/codice/codice_l1a.py
```diff
@@ -121,47 +121,42 @@ def create_science_dataset(self, start_time: np.datetime64) -> xr.Dataset:
...
 
         # Create a data variable for each species
         for variable_data, variable_name in zip(self.data, self.variable_names):
-            fieldname = self.variable_names[variable_name]["fieldname"]
-            catdesc = self.variable_names[variable_name]["catdesc"]
-
             variable_data_arr = np.array(list(variable_data), dtype=int).reshape(
                 -1, self.num_energy_steps
             )
+            cdf_attrs_key = (
```
maxinelasp, line 137: Nice, I like this

maxinelasp (approved):
Much better, thanks for making those changes!

PR #583, opened 2024-06-05
Create HIT L1A CDF for Housekeeping

vmartinez-cu:
Added code to create HIT L1A CDFs

# Change Summary
This PR creates a CDF file for the HIT housekeeping packet. We're working with housekeeping for the SIT-3 deadline, but most of the code is applicable to other packet apids and will be expanded on after SIT-3.

## Overview

- Renamed hit_l1a_decom.py to hit_l1a.py and moved file to l1a directory
- Added code to create hit l1a xarray datasets and write CDF files
- Updated test_hit_decom.py to reflect changes in l1a processing code
- Updated hit_cdf_attrs.py to add adc_channels variable and attributes to use as a second dims for the leak_i data variable. Also, updated some data variable names to match fields in hk data classes to easily find matching attributes
- Added test for hit_l1a code

## New Dependencies
- test_hit_decom.py
   - Moved tests to new test file  

## New Files
- test_hit_l1a.py
   - unit tests for l1a data processing

## Deleted Files
None

## Updated Files
- hit_l1a.py
   - Renamed hit_l1a_decom.py to hit_l1a.py
   - Moved file to l1a directory
   - Added code to populate housekeeping data classes, create datasets, and write CDF files
- hit_cdf_attrs.py
   - Added "adc_channels" metadata variable and defined attributes for it. This is needed as a second dim for the leak_i data variable which is a 2d array
   - Updated some data variable names to match fields in the housekeeping data class to more easily find cdf attributes

## Testing
test_hit_l1a.py was added to test the l1a data processing code. Also added tests from test_hit_decom.py to this file and deleted test_hit_decom.py


imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+
+    HIT_AUT = 1250  # Autonomy
+    HIT_HSKP = 1251  # Housekeeping
+    HIT_SCIENCE = 1252  # Science
+    HIT_IALRT = 1253  # I-ALiRT
+    HIT_MEMDUMP = 1255  # Memory dump
+
+
+def hit_l1a_data(packet_file: str):
```
greglucas:
```suggestion
def hit_l1a_data(packet_file: Path | str):
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+        Path to the CCSDS data packet file.
+
+    Returns
+    -------
+    dict
+        List of file paths to CDF data product files
+    """
+    # Decom, sort, and group packets by apid
+    packets = decom_packets(packet_file)
+    sorted_packets = sorted(packets, key=lambda x: x.data["SHCOARSE"].derived_value)
```
greglucas: Use the global utility function here too?
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/1efb973eb1830c4e3407c84e1199d8b83c1a7899/imap_processing/utils.py#L19
```suggestion
    sorted_packets = sort_by_time(packets, "SHCOARSE")
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+        "shcoarse",
+        "ground_sw_version",
+        "packet_file_name",
+        "ccsds_header",
+        "leak_i_raw",
+    ]
+    datasets = create_datasets(grouped_data, skip_keys)
+
+    # Create CDF files
+    logger.info("Creating CDF files for HIT L1A data")
```
greglucas, line 78: I think that most of these should be "debug" level otherwise they'll add a lot of extra text to our logs that we don't necessarily need to see all the time.

We should discuss our logging policy more broadly as a team because I don't think we're currently very uniform in what messages get placed at what level.

Not a blocker here, just a note/thought for later.

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+    Returns
+    -------
+    dict
+        grouped data by apid
+    """
+    logger.info("Grouping packet values for each apid")
+    grouped_data = utils.group_by_apid(unpacked_data)
+
+    # Create data classes for each packet
+    for apid in grouped_data.keys():
```
greglucas:
```suggestion
    for apid in grouped_data:
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+        if apid == HitAPID.HIT_HSKP:
+            logger.info(f"Grouping housekeeping packets - APID: {apid}")
+            for i, packet in enumerate(grouped_data[apid]):
+                # convert data to data classes
+                grouped_data[apid][i] = Housekeeping(packet, "0.0", "hskp_sample.ccsds")
```
greglucas: Do you want to replace your current `grouped_data[apid]` entries, or do you want to make a new entry / container?
You can also shorten this to a list comprehension.
```suggestion
            grouped_data[apid] = [Housekeeping(packet, "0.0", "hskp_sample.ccsds") for packet in grouped_data[apid]]
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+
+def create_datasets(data: dict, skip_keys=None):
+    """
+    Create a dataset for each APID in the data.
+
+    Parameters
+    ----------
+    data : dict
+        A single dictionary containing data for all instances of an APID.
+    skip_keys: list, Optional
```
greglucas, line 148: add a space for numpydoc
```suggestion
    skip_keys : list, Optional
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+    logger.info("Creating datasets for HIT L1A data")
+    processed_data = {}
+    for apid, data_packets in data.items():
+        metadata_arrays = defaultdict(list)
+        for packet in data_packets:
+            # Add metadata to an array
+            for field in fields(packet):
+                field_name = field.name
+                field_value = getattr(packet, field_name)
+                print(f"{field_name}: {field_value}")
```
greglucas:
```suggestion
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+                # convert key to lower case to match SPDF requirement
+                data_key = field_name.lower()
+                metadata_arrays[data_key].append(field_value)
+
+        # Create xarray data arrays for dependencies
+        epoch_time = xr.DataArray(
+            np.array(metadata_arrays["shcoarse"], dtype="datetime64[s]"),
```
greglucas: Just an FYI that SHCOARSE is an integer, so I think you will need to use the `calc_start_time()` function to get this value into a `datetime64[s]`, otherwise this is actually going to be showing 1970's unix epoch time values.

Edit:
here is the warning that shows up in the test logs: https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/9388783053/job/25854766240#step:6:67

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+                # convert data to data classes
+                grouped_data[apid][i] = Housekeeping(packet, "0.0", "hskp_sample.ccsds")
+        else:
+            raise RuntimeError(f"Encountered unexpected APID [{apid}]")
+
+    logger.info("Finished grouping packet data")
+    return grouped_data
+
+
+def create_datasets(data: dict, skip_keys=None):
```
greglucas, line 140: Can you use the global utility function instead of writing your own?
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/1efb973eb1830c4e3407c84e1199d8b83c1a7899/imap_processing/utils.py#L131

imap_processing/tests/hit/test_hit_l1a.py
```diff
@@ -0,0 +1,106 @@
...
+    skip_keys = [
+        "shcoarse",
+        "ground_sw_version",
+        "packet_file_name",
+        "ccsds_header",
+        "leak_i_raw",
+    ]
+    datasets = hit_l1a.create_datasets(grouped_data, skip_keys=skip_keys)
+    assert len(datasets.keys()) == 1
+    assert isinstance(datasets[1251], xr.Dataset)
```
greglucas: I was confused here at first, because I thought you were making 1251 datasets 😂  I'd suggest something like this renaming instead (also use the APID Enum to avoid the number)
```suggestion
    datasets_by_apid = hit_l1a.create_datasets(grouped_data, skip_keys=skip_keys)
    assert len(datasets.keys()) == 1
    assert isinstance(datasets_by_apid[hit_l1a.HITApid.HSKP], xr.Dataset)
```

imap_processing/tests/hit/test_hit_l1a.py
```diff
@@ -0,0 +1,106 @@
...
+        "ground_sw_version",
+        "packet_file_name",
+        "ccsds_header",
+        "leak_i_raw",
+    ]
+    datasets = hit_l1a.create_datasets(grouped_data, skip_keys=skip_keys)
+    cdf_filepaths = []
+    for dataset in datasets.values():
+        cdf_file = write_cdf(dataset)
+        cdf_filepaths.append(cdf_file)
```
greglucas: You're rewriting the `hit_l1a_data` code here I think, do you mean to call that function so you're testing it?
```suggestion
    cdf_filepaths = hit_l1a.hit_l1a_data(packet_filepath)
```

greglucas: This will also help with the `codecov` failures I think: https://app.codecov.io/gh/IMAP-Science-Operations-Center/imap_processing/pull/583/blob/imap_processing/hit/l1a/hit_l1a.py

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+        Housekeeping
+    HIT_SCIENCE : int
+        Science
+    HIT_IALRT : int
+        I-ALiRT
+    HIT_MEMDUMP : int
+        Memory dump
+    """
+
+    HIT_AUT = 1250  # Autonomy
```
sdhoyt: HIT_AUT and HIT_MEMDUMP can be deleted from here

imap_processing/hit/hit_cdf_attrs.py
```diff
@@ -61,10 +62,22 @@
 # TODO: update label_axis with values provided by instrument team.
 #  Waiting for info
 
+# Define
```
tech3371: remove?

imap_processing/tests/hit/test_hit_decom.py
```diff
@@ -1,15 +1,14 @@
 import pytest
```
sdhoyt: I think this test should be moved into `test_hit_l1a.py`

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: I borrowed this code from the global utility function. The global function takes in a list of packets produced by space packet parser. The HIT data is stored as data classes to further decommutate some of the data, so I wasn't able to use the global function as is. If other instruments are also using data classes then maybe this global function could be modified to handle both packet formats

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Returns
+    -------
+    dict
+        List of file paths to CDF data product files
+    """
+    # Decom, sort, and group packets by apid
+    packets = decom_packets(packet_file)
+    sorted_packets = sorted(packets, key=lambda x: x.data["SHCOARSE"].derived_value)
+    grouped_data = group_data(sorted_packets)
```
tech3371, line 60: Similar to Greg's comment, there is this `group_by_apid()` that you can use to group your data per apid.https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/e1234fe0f69ab76a3d3484114abb8f200b2e1989/imap_processing/utils.py#L39

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+
+    # Create datasets
+    # TODO define keys to skip for each apid. Currently just have
+    #  a list for housekeeping.
+    skip_keys = [
+        "shcoarse",
+        "ground_sw_version",
+        "packet_file_name",
+        "ccsds_header",
+        "leak_i_raw",
```
sdhoyt, line 73: Don't need to make the change in this PR, but for the future if we continue to create the housekeeping CDF, we can use the builtin `dataclasses` class `InitVar` so `leak_i_raw` doesn't show up when you extract the object's field names. 

Also after SIT-3, we might want to include the CCSDS header, packet file name, and sw version in the CDF (we'll need to talk to Eric C and Grant about that), so you might only need to skip shcoarse here in the future.

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,208 @@
...
+    ]
+    datasets = create_datasets(grouped_data, skip_keys)
```
tech3371, line 75: Also, This function may be helpful here. https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/e1234fe0f69ab76a3d3484114abb8f200b2e1989/imap_processing/utils.py#L131

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Nice! I missed this global function. Thanks for the suggestion

sdhoyt (approved):
Great work! Like I said before, I didn't see being able to get the sample data to populate the L1A CDF in time for SIT-3, so this is really great.

imap_processing/hit/l1a/hit_l1a.py
tech3371: SWAPI has example of how it's using above functions, https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/e1234fe0f69ab76a3d3484114abb8f200b2e1989/imap_processing/swapi/l1/swapi_l1.py#L587

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Good point. A team discussion on logging would be great

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: We're using data classes for the packets rather than the space packet parser formatted packets to create the datasets. Do you see any advantages with creating a new container instead of updating `grouped_data[apid]`? That would be an easy update to make

vmartinez-cu: I prefer the list comprehension. Thanks!

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Good catch. I thought I had cleaned these print statements up

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: InitVar is perfect for this variable. I added a TODO so we won't forget about this and updating the list of fields to skip

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: I use that global function in my custom group_data function to group the packets from space packet parser. However, the data need to be stored as data classes for further decommutation steps and so my function replaces the space packet parser packets with those data classes in the grouped_data dictionary it outputs. I could rename this function so that it isn't confused with the global group_by_apid function. Thoughts?

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: I looked at those examples when I was working on this code and they were super helpful! I had to create a custom create_dataset function because our packets are data classes and the global function expects a list of space packet parser packets. Maybe there's a better way to handle this difference in data format?

imap_processing/hit/l1a/hit_l1a.py
maxinelasp: GLOWS and MAG do something similar and don't use that global utility function.

imap_processing/hit/l1a/hit_l1a.py
maxinelasp: Like previously mentioned, I do some dataclass oriented sorting (especially in MAG) if you want to take a look. Maybe there is shared work in that design that we can pull out.

imap_processing/tests/hit/test_hit_l1a.py
vmartinez-cu: My last job did not have good testing practices, so I'm a novice at unit tests. Your feedback is much appreciated!

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,212 @@
...
+    ----------
+    HIT_HSKP: int
+        Housekeeping
+    HIT_SCIENCE : int
+        Science
+    HIT_IALRT : int
+        I-ALiRT
+    """
+
+    HIT_HSKP = 1251  # Housekeeping
```
maxinelasp: These comments are duplicated from the docstring, remove

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -0,0 +1,212 @@
...
+    ]
+    datasets = create_datasets(grouped_data, skip_keys)
+
+    # Create CDF files
+    logger.info("Creating CDF files for HIT L1A data")
+    cdf_filepaths = []
+    for dataset in datasets.values():
+        cdf_file = write_cdf(dataset)
+        cdf_filepaths.append(cdf_file)
+    logger.info("L1A CDF files created")
```
maxinelasp: This function is great. My only note would be to add "cdf_filepaths" to the output logging here.

maxinelasp (approved):
Looks good to me!

imap_processing/hit/l1a/hit_l1a.py
tech3371: Oh ok. I should have look at function for more details. Yeah, I thought this function was doing same thing as utils function. Sounds good.

imap_processing/hit/l1a/hit_l1a.py
tech3371: I see. Yeah, we can revisit that in future PR.

imap_processing/hit/l1a/hit_l1a.py
tech3371: I struggle with this too! :)

tech3371 (approved):

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Updated!

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Yes, these comments are redundant. Good catch 👍

PR #584, opened 2024-06-06
Created SWE CDF Attrs YAML

tech3371:
# Change Summary

## Overview
- created SWE YAML for l1a and l1b
- updated code to use yaml
- fixes to utils function
- removed old cdf attrs


imap_processing/cdf/config/imap_swe_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,25 @@
...
+    in the heliosphere. SWE will contribute to our understanding of the acceleration
+    and transport of charged particles in the heliosphere.
+    SWE design and assembly is led by the Princeton Plasma Physics Laboratory. See
+    https://imap.princeton.edu/instruments/swe for more details.
+  Data_version: "001"
+  Instrument_type: "Particles (space)"
+
+imap_swe_l1a_sci:
+  <<: *instrument_base
+  # NOTE: Right now, this Data_level is required to produce valid CDF
```
bourque, line 14: Was that introduced in a new version of `cdflib`? I think I missed that.

imap_processing/cdf/config/imap_swe_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,206 @@
+# <=== Coordinates ===>
```
bourque, line 1: I like these organizational comments, I am going to steal the idea!

imap_processing/cdf/config/imap_swe_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,206 @@
...
+  VALIDMAX: 9223372036854775807
+  UNITS: ns
+  VAR_TYPE: support_data
+  SCALETYP: linear
+  MONOTON: INCREASE
+  TIME_BASE: J2000
+  TIME_SCALE: Terrestrial Time
+  REFERENCE_POSITION: Rotating Earth Geoid
+
+# <=== Data Variables ===>
```
bourque, line 47: It is nit-picky/personal preference, but I would find it to be a bit more readable if there was a blank line between each data variable

imap_processing/cdf/config/imap_swe_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,206 @@
+# <=== Coordinates ===>
+spin_angle:
+  CATDESC: 180 measurements per quarter cycle. See VAR_NOTES for details
+  FIELDNAM: Spin Angle
+  FILLVAL: -9223370000000000000
```
bourque, line 6: I am wondering if there is a way to (and if it is a good idea to) define some of these constant numbers in a separate YAML file (e.g. `global_constants.yaml`) and reference them in our instrument-specific yaml files, instead of defining them multiple times within a file.

bourque (approved):
Looks good to me! I just had some minor comments/questions.

imap_processing/cdf/config/imap_swe_global_cdf_attrs.yaml
tech3371: This is introduced by both our schema usage and `cdflib` version we are using. New `cdflib` catches this error. What's happening behind the scene was that new schema sorts global attrs in alphebetic order and it set those key that we don't set value for to `None` value. Then `cdflib` version we have doesn't catch this `None` value. If a global attrs is set to `None`, `cdflib`skips adding remaining global attrs.

imap_processing/cdf/config/imap_swe_l1a_variable_attrs.yaml
tech3371: I agree. I don't how easy it is to inherit other values from other yaml files. But would be nice to define some constant value in yaml file.

imap_processing/swe/utils/swe_utils.py
```diff
@@ -43,83 +24,10 @@ def add_metadata_to_array(data_packet, metadata_arrays):
...
-        elif key == "APP_MODE":
-            # We need to get derived value for this because it's used in
-            # filename.
-            metadata_arrays.setdefault(key, []).append(
-                value.raw_value
-                if value.derived_value is None
-                else str.lower(value.derived_value)
-            )
         else:
             metadata_arrays.setdefault(key, []).append(value.raw_value)
```
greglucas:
```suggestion
        metadata_arrays.setdefault(key, []).append(value.raw_value)
```

imap_processing/tests/cdf/test_utils.py
```diff
@@ -32,11 +35,11 @@ def test_dataset():
...
         },
-        attrs=swe_l1a_global_attrs.output()
+        attrs=swe_attrs.get_global_attributes("imap_swe_l1a_sci")
         | {
-            "Logical_source": "imap_swe_l1_sci",
+            "Logical_source": "imap_swe_l1a_sci",
             "Data_version": "001",
-            "Logical_file_id": "imap_swe_l1_sci_20100101_v001",
+            "Logical_file_id": "imap_swe_l1a_sci_20100101_v001",
         },
```
greglucas: Aren't some (all?) of these already set in the `get_global_attributes()` yaml definitions now? So you may not need to update this anymore.

greglucas (approved):

imap_processing/tests/cdf/test_utils.py
tech3371: `Logical_file_id` is not still. But good point about cleaning up this `or` option.

imap_processing/cdf/config/imap_swe_l1a_variable_attrs.yaml
maxinelasp: I agree with both of you, but it's not really a thing in YAML unfortunately. My idea was to use a placeholder word (like "INTFILLVAL") that is then replaced by the code when the attributes are read in, but I kind of... don't like that so I never did it. There may be some yaml feature I don't know about that would solve this problem but at least at first look I couldn't turn anything up :(

imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,267 @@
...
+  VALIDMAX: 9223372036854775807
+  UNITS: ns
+  VAR_TYPE: support_data
+  SCALETYP: linear
+  MONOTON: INCREASE
+  TIME_BASE: J2000
+  TIME_SCALE: Terrestrial Time
+  REFERENCE_POSITION: Rotating Earth Geoid
+
+# <=== Data Variables ===>
```
maxinelasp, line 68: I like this comment

imap_processing/swe/l1a/swe_science.py
```diff
@@ -139,99 +137,70 @@ def swe_science(decom_data):
...
+        dataset[metadata_field] = xr.DataArray(
             value,
             dims=["epoch"],
-            attrs=dataclasses.replace(
-                swe_cdf_attrs.swe_metadata_attrs,
-                catdesc=key,
-                fieldname=key,
-                label_axis=key,
-                depend_0="epoch",
-            ).output(),
```
maxinelasp, line 234: I love seeing how much code is removed here 😆 so satisfying!

maxinelasp (approved):
Very nice! I didn't closely check the values of the CDF attributes, but I figure you are probably better equipped for that than me 😅

imap_processing/swe/l1a/swe_science.py
tech3371: same here!

PR #597, opened 2024-06-07
BUGFIX - load_cdf results in added xarray specific attributes

tmplummer:
Remove xarray variable and coordinate attributes that get injected by… cdf_to_xarray

# Change Summary
* Added optional keyword `remove_xarray_attrs` that defaults to True to `load_cdf`. When set True, attributes that the current version of cdflib `cdf_to_xarray` injects into the coordinate and variable attributes of the xarray.Dataset will be removed.

## Overview
Using SKTEditor to verify ISTP compliance, I discovered a 'units' attribute that was being added to the CDF file generated by loading a CDF to xarray.Dataset and then writing the xarray.Dataset back out to CDF. Digging further, I found that three attributes are added: ["units", "standard_name", "long_name"]. These will now be stripped off in the `load_cdf` function.

## Updated Files
   - Added optional keyword `remove_xarray_attrs` that defaults to True to `load_cdf`
   - Added code to strip off unwanted variable and coordinate attributes if `remove_xarray_attrs` is truthy
   - Added test coverage for removal of unwanted attributes

## Testing
Test coverage added

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -82,9 +82,8 @@ def load_cdf(
     # TODO: This can be removed if/when feature is added to cdf_to_xarray to
     #      make adding these attributes optional
     if remove_xarray_attrs:
-        xarray_attrs = [v for k, v in ISTP_TO_XARRAY_ATTRS.items()]
         for key in dataset.variables.keys():
-            for xarray_key in xarray_attrs:
+            for xarray_key in  ISTP_TO_XARRAY_ATTRS.values():
                 dataset[key].attrs.pop(xarray_key, None)
 
     return dataset
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -2,7 +2,6 @@
 
 import logging
 import re
-from collections import ChainMap
 from pathlib import Path
 from typing import Optional
 
@@ -84,9 +83,9 @@ def load_cdf(
     #      make adding these attributes optional
     if remove_xarray_attrs:
         xarray_attrs = [v for k, v in ISTP_TO_XARRAY_ATTRS.items()]
-        for _, data_array in ChainMap(dataset.data_vars, dataset.coords).items():
-            for key in xarray_attrs:
-                data_array.attrs.pop(key, None)
+        for key in dataset.variables.keys():
+            for xarray_key in xarray_attrs:
+                dataset[key].attrs.pop(xarray_key, None)
 
     return dataset
 
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -83,7 +83,7 @@ def load_cdf(
     #      make adding these attributes optional
     if remove_xarray_attrs:
         for key in dataset.variables.keys():
-            for xarray_key in  ISTP_TO_XARRAY_ATTRS.values():
+            for xarray_key in ISTP_TO_XARRAY_ATTRS.values():
                 dataset[key].attrs.pop(xarray_key, None)
 
     return dataset
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -2,6 +2,7 @@
 
 import logging
 import re
+from collections import ChainMap
 from pathlib import Path
 from typing import Optional
 
@@ -9,6 +10,7 @@ import imap_data_access
 import numpy as np
 import xarray as xr
 from cdflib.xarray import cdf_to_xarray, xarray_to_cdf
+from cdflib.xarray.cdf_to_xarray import ISTP_TO_XARRAY_ATTRS
 
 import imap_processing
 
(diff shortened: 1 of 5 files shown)


imap_processing/cdf/utils.py
```diff
@@ -72,6 +79,15 @@ def load_cdf(file_path: Path, **kwargs: dict) -> xr.Dataset:
...
             dataset.attrs[attribute] = value[0]
 
+    # Remove attributes specific to xarray plotting from vars and coords
+    # TODO: This can be removed if/when feature is added to cdf_to_xarray to
+    #      make adding these attributes optional
+    if remove_xarray_attrs:
+        xarray_attrs = [v for k, v in ISTP_TO_XARRAY_ATTRS.items()]
+        for _, data_array in ChainMap(dataset.data_vars, dataset.coords).items():
+            for key in xarray_attrs:
+                data_array.attrs.pop(key, None)
```
greglucas: I think this might be able to be simplified and not need to use ChainMap.
```suggestion
        for key in (dataset.data_vars | dataset.coords).keys():
            for xarray_key in ISTP_TO_XARRAY_ATTRS.values():
                dataset[key].attrs.pop(xarray_key, None)
```

imap_processing/tests/cdf/test_utils.py
```diff
@@ -66,6 +68,13 @@ def test_load_cdf(test_dataset):
     dataset = load_cdf(file_path)
     assert isinstance(dataset, xr.core.dataset.Dataset)
 
+    # Test removal of attributes that are added on by cdf_to_xarray and
+    # are specific to xarray plotting
+    xarray_attrs = ["units", "standard_name", "long_name"]
+    for _, data_array in ChainMap(dataset.data_vars, dataset.coords).items():
+        for attr in xarray_attrs:
+            assert attr not in data_array.attrs.keys()
```
greglucas:
```suggestion
            assert attr not in data_array.attrs
```

greglucas (approved):
Minor suggestions to try and simplify, but I like this otherwise 👍

imap_processing/cdf/utils.py
tmplummer: Thanks for the suggestion. It almost worked but lead me to finding that the `xarray.Dataset.variables` attribute is a lower level access to a dictionary of all coordinates and variables. `Dataset.data_vars` and `Dataset.coords` return different typed objects and so the pipe operator does not work.

imap_processing/cdf/utils.py
```diff
@@ -72,6 +78,15 @@ def load_cdf(file_path: Path, **kwargs: dict) -> xr.Dataset:
         if isinstance(value, list) and len(value) == 1:
             dataset.attrs[attribute] = value[0]
 
+    # Remove attributes specific to xarray plotting from vars and coords
+    # TODO: This can be removed if/when feature is added to cdf_to_xarray to
+    #      make adding these attributes optional
+    if remove_xarray_attrs:
+        xarray_attrs = [v for k, v in ISTP_TO_XARRAY_ATTRS.items()]
+        for key in dataset.variables.keys():
+            for xarray_key in xarray_attrs:
```
greglucas: I think you're just keeping the values? (Similar in the other testing file where you iterate over items and then ignore the key, I'd just access `.values()` directly.
```suggestion
        for key in dataset.variables.keys():
            for xarray_key in  ISTP_TO_XARRAY_ATTRS.values():
```

imap_processing/cdf/utils.py
greglucas: Awesome, even better! Good to know there is that other attribute, thanks for digging in :)

imap_processing/cdf/utils.py
```diff
@@ -48,13 +49,18 @@ def calc_start_time(
     return launch_time + time_delta
 
 
-def load_cdf(file_path: Path, **kwargs: dict) -> xr.Dataset:
+def load_cdf(
+    file_path: Path, remove_xarray_attrs: bool = True, **kwargs: dict
```
tech3371, line 53: Are you going to add `to_datetime=True` in another PR? Just curious

tech3371 (approved):
LGTM

greglucas:
pre-commit.ci autofix

imap_processing/cdf/utils.py
tmplummer: Oh... thanks for reminding me. I'll add that here.

tmplummer: I have added this change to this PR. @tech3371 and @greglucas, can you take another look at this PR for those changes. I wasn't sure if my implementation of setting that default is too opaque and should be done through an explicit keyword argument.

imap_processing/cdf/utils.py
```diff
@@ -63,6 +69,8 @@ def load_cdf(file_path: Path, **kwargs: dict) -> xr.Dataset:
     dataset : xr.Dataset
         The ``xarray`` dataset for the CDF file
     """
+    if "to_datetime" not in kwargs:
+        kwargs["to_datetime"] = True
```
tech3371, line 76: Can we add a comment in the function doc string that we are setting this by default? Otherwise, it looks great to me

tech3371 (approved):

imap_processing/cdf/utils.py
greglucas: :+1: I think that this looks fine. Something to consider though is that I think the datetime stuff is going to be automatically handled for us by cdflib when we upgrade (I'm not positive on this though), so do we want to add this now or can we just wait until we hit the cdflib upgrade?

We also have a bunch of uses of instruments already injecting this, so maybe remove it from the other areas of the code as well if we are going to add it?

tech3371: I don't think this is in the `cdflib` upgrade. @bryan-harter Can you confirm? I think `cdflib` still keeps this as optional parameter. But Bryan mentioned this in our Slack which is little different from this. Is that what you are thinking?
```
Going along with the above, "datetime64_to_tt2000" is no longer needed in the xarray_to_cdf function. I decided to just make that the default behavior because why else would you have a datetime object if you didn't want a CDF time variable
```

imap_processing/cdf/utils.py
vmartinez-cu: Nice improvement to this function!

vmartinez-cu (approved):
LGTM

PR #600, opened 2024-06-07
MNT/TST: Prepare for cdflib upgrade

greglucas:
# Change Summary

## Overview

When writing a cdf file out, the data needs to be monotonically increasing. Until we get good validation data we can remove the duplicates to appease the newer cdflib versions.


tech3371 (approved):

tmplummer (approved):
Thanks for doing this, Greg.

Issue #594, opened 2024-06-07
CDF attribute manager: Create instrument-specific schemas


maxinelasp:
### Summary

We should expand the existing ISTP/Global schemas in cdf_attribute_manager to include IMAP specific global attributes or expected variables, such as file_naming_convention. This should involve creating a new schema file that extends but does not replace the existing global schema file, and writing code so that you can add multiple schemas to the class. 

Any IMAP specific changes should go in imap_attribute_manager, but try to put most of the code in the generic cdf_attribute_manager. 

### Requirements

The more specific the better.
  - [ ] Additional global schemas can be added to the attribute manager
  - [ ] Additional variable schemas can be added to the attribute manager
  - [ ] Test that is currently marked "xfail" should pass
  - [ ] Code should be mostly in cdf_attribute_manager, and generic across multiple missions.

tech3371:
I believe this is completed because of SAMMI.

PR #603, opened 2024-06-07
Adding LABL_PRT_i to 2D or higher Dimension Data Type - SWE

tech3371:
# Change Summary

## Overview
- changes to add LABL_PTR_i
- fixes for filename convention error on SKEditor

### DIFF
--- a/imap_processing/cdf/config/shared/default_global_cdf_attrs_schema.yaml
+++ b/imap_processing/cdf/config/shared/default_global_cdf_attrs_schema.yaml
@@ -69,6 +69,14 @@ Discipline:
   required: true
   validate: true
   overwrite: false
+File_naming_convention:
+  description: >
+    If File_naming_convention was not set, it uses default setting:
+      source_datatype_descriptor_yyyyMMdd
+  default: source_datatype_descriptor_yyyyMMdd
+  required: false
+  validate: false
+  overwrite: true
 Generation_date:
   description: >
     Date stamps the creation of the file using the syntax yyyymmdd, e.g., "
--- a/imap_processing/swe/l1a/swe_science.py
+++ b/imap_processing/swe/l1a/swe_science.py
@@ -160,6 +160,7 @@ def swe_science(decom_data):
     )
 
     # NOTE: LABL_PTR_1 should be CDF_CHAR.
+    # TODO: Make this label and spin_angle data match. Do same for polar_angle
     spin_angle_label = xr.DataArray(
         [str(i) for i in np.arange(180)],
         name="spin_angle_label",
--- a/imap_processing/swe/l1a/swe_science.py
+++ b/imap_processing/swe/l1a/swe_science.py
@@ -159,11 +159,27 @@ def swe_science(decom_data):
         attrs=cdf_attrs.get_variable_attributes("spin_angle"),
     )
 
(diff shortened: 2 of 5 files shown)


greglucas (approved):
I can't say I like adding extra variables for labels to a dataset, but everyone's already heard how much I dislike the CDF/ISTP stuff anyways 😂 

If this is what we have to do, then I guess we've got to do it.

sdhoyt (approved):
Tried what you did in my code and it worked. Thanks for this!

imap_processing/cdf/config/shared/default_global_cdf_attrs_schema.yaml
```diff
@@ -69,6 +69,14 @@ Discipline:
   required: true
   validate: true
   overwrite: false
+File_naming_convention:
```
tmplummer, line 72: I was also going to add this to the schema but saw that @maxinelasp created a ticket related to this. Issue #594

imap_processing/swe/l1a/swe_science.py
```diff
@@ -159,13 +159,29 @@ def swe_science(decom_data):
         attrs=cdf_attrs.get_variable_attributes("spin_angle"),
     )
 
+    # NOTE: LABL_PTR_1 should be CDF_CHAR.
+    spin_angle_label = xr.DataArray(
+        [str(i) for i in np.arange(180)],
```
tmplummer: I would recommend generating this based on the values in `spin_angle` so that if those change, the label also changes. (This comment goes for all of the label variables)
I would think that CDAWeb should do this based on the DEPEND_i attributes. I will try to remember to ask Andriy about this next time.

tmplummer (commented):
Thanks for forging the path forward for multi-dimensional CDF variables.

imap_processing/swe/l1a/swe_science.py
tech3371: I think it's a good idea. I will update that in future PR if you don't mind.

greglucas: I think we should just do it here for explicitness and don't think it will be too hard to incorporate in this PR.

```suggestion
        spin_angle.astype(str),
```
(might need a `.values` or similar to get the numpy data)

Another question that this brings up is you have 180 bins here, but the angle resolution in the YAML file says 12 degrees, so I'm a bit confused about that mismatch.

tech3371: That `astype(str)` is a great idea. If that, then it may be easier change that what I was thinking initially.

180 is 180 measurement per quarter cycle. This is just for l1a. At l1a, it has not information about angle. What you are thinking is l1b data product description. It's binned in 30 bins and it's 12 degree nominally.

greglucas: This is the yaml descriptor you're referencing here. So I think you'd need to update this YAML description to be different for l1a and l1b then.
```
spin_angle_label:
  CATDESC: Spin angle step. Angle resolution is 12 degree nominally
  FIELDNAM: 12 Degree Resolution
  FORMAT: A3
  VAR_TYPE: metadata
```

tech3371: Ah ok. I will double check this. Thank you for catching it!

PR #605, opened 2024-06-10
Glows l1b processing

maxinelasp:
# Change Summary
This completes the first draft of GLOWS L1B processing. The missing pieces and TODOs are primarily because I haven't heard back from GLOWS about the L1A CDF file, and should be reasonably simple to complete once the L1A is completely final. 

## Overview

Created glows_l1b.py and glows_l1b_data.py to apply glows L1B processing. This does not include the CDF generation, which will be coming in a second PR (hopefully today), however, the end result are tuples of DataArrays with the correct dimensions, so it should be simple to add attributes and assemble them into Datasets 🤞 

## New Files

- glows_l1b.py
   - Run processing code for GLOWS L1B
- glows_l1b_data.py 
   - Dataclasses for organizing and running specific algorithms on variables to convert them all from L1A to L1B.

## Testing

There is room for improvement with the tests, particularly for specific algorithms, but the actual conversion from L1A -> L1B is very strictly enforced and will throw errors if dimensions don't align, there are missing inputs or outputs, etc. So, the fact that test_process_histograms and test_process_de pass without failing is key.


imap_processing/glows/l1b/glows_l1b.py
```diff
@@ -0,0 +1,88 @@
...
+    return l1b_fields
+
+
+def process_histogram(l1a: xr.Dataset) -> xr.Dataset:
+    """Process the histogram data from the L1A dataset and return the L1B dataset."""
+    dataarrays = [l1a[i] for i in l1a.keys()]
+
+    dims = [[] for i in l1a.keys()]
+    # 32 is the number of attributes in the HistogramL1B class.
+    new_dims = [[] for i in range(34)]
```
sdhoyt, line 108: I think I'm confused about what this is doing. You're creating a list of empty lists for each of the attributes in the Histogram L1B class, plus ones for bins and flag? Where are the index and field mappings coming from (the 0, 22, 23, then the last 4 indexes in the list). Are these all dimensions of the data or attributes?

maxinelasp: They are the dimensions of the data. If the dimension of the data is 1D on epoch, it just needs to be an empty array. If it is more dimensions than that, it needs to include the other dimensions. So, the indices are for the variables which are 2D (or higher.) I need to describe the output dimensions of those variables so the dataarrays are properly aligned. 

It's kind of complicated, I can try and rewrite if it would help make things more clear... I couldn't really figure out a good way of selecting the indices of the new dimensions out of the dataclass, but I can use a dictionary to make it more clear if that would help. Either way I'll add more comments, but would you want to see that change? Did that answer your question?

The dimensions are the hardest part of this style of code. Luckily, most of the dimensions here are just 1D so they're easier to deal with.

maxinelasp: Also, this step does not include ANY attributes yet, so it's only the data and DataArray.

sdhoyt: Ah ok, I think I understand. I think that was part of what confused me is coords is also a parameter for the xr.DataSet. But that's a dimension of your data as well?

Instead of hardcoding could you pull those dimension strings from the YAML? Something like this?
`dims = [
                value
                for key, value in attr_mgr.get_variable_attributes(field).items()
                if "DEPEND" in key
            ]
`

Do you think you could run into an issue if you always assume epoch is a dimension? For example, for Lo I have support data that doesn't vary by time, so epoch is not a dimension. That might not be an issue for GLOWS though.

Sorry for all the questions about this!

maxinelasp: Yeah, that's an interesting idea! I'll try and do that, I think that's better and then it's coming from somewhere concrete as well. 

Data variables that don't vary by time are a special case and would not work with this design, unfortunately. You could just leave them out of the arguments (here, `*dataarrays`) and handle them separately. I'd probably write a static method to apply any calculations to them and do them after the ufunc step. In this case, they all vary by time, but they do need that epoch dimension to align everything. 

That being said, the apply_ufunc method is really powerful and complicated, so it's possible that you could still pass in non-time varying variables if you needed (perhaps through kwargs?)

maxinelasp: Is it ok if I leave this for now, and apply your idea in the next CDF PR? I don't want to add the CDF stuff to this PR and make it bigger.

maxinelasp: And you're right that "coords" is a bad choice of dimension name, lol. I'll change that now

sdhoyt: Yeah sounds good! The vectorized function is a really cool idea.

sdhoyt (approved):

imap_processing/glows/l1b/glows_l1b.py
maxinelasp: @sdhoyt I updated the docs to cover what I explained here. Thanks for the review!

imap_processing/glows/ancillary/l1b_conversion_table_v001.json
greglucas, line 1: Is this the file that you were wondering about whether to put in AWS or not? Do we anticipate this changing frequently or infrequently? It actually doesn't seem so bad to keep this next to the code rather than having to request to download it every time we want to process.

imap_processing/glows/ancillary/l1b_conversion_table_v001.json
```diff
@@ -0,0 +1,42 @@
+{
+    "description": "Table for conversion/decoding ancillary parameters collected onboard by IMAP/GLOWS",
+    "version": "0.1",
```
greglucas: Do you want to track the version in the filename?
```suggestion
    "version": "001",
```

imap_processing/glows/l1b/glows_l1b.py
```diff
@@ -0,0 +1,142 @@
...
+    # Flags is a constant length. It has a dimension of "flags"
+    new_dims[-3] = ["flags"]
+
+    # glows_times and pulse_lengths should be dimension of "per_second", the same as
+    # the input.
+    new_dims[-2] = ["per_second"]
+    new_dims[-1] = ["per_second"]
+
+    l1b_fields = xr.apply_ufunc(
+        lambda *args: tuple(dataclasses.asdict(DirectEventL1B(*args)).values()),
```
greglucas, line 73: I'm not sold on this style personally, but again I'm happy to be the odd one out. I'd find this much easier to follow to use typical numpy-style broadcasting on things and I think it would be much faster because you'd iterate in C rather than iterate over each time via dataclasses. (I also think this might be thrashing disk I/O opening the params file for each iteration)

Just to be explicit, something like this is what I would personally find easier follow with processing functions being applied rather than class-oriented methods.
```python
def decode_ds(ds, param_key, encoded_key):
        params = ds[param_key]
        # compute parameters a and b:
        param_a = (2 ** params["n_bits"] - 1) / (params["max"] - params["min"])
        param_b = -params["min"] * param_a

        return (ds["encoded_key"] - param_b) / param_a).astype(np.single)

# Then call it on the dataarrays of interest
output_data_array = decode_ds(l1a_ds, "hv_voltage", "hv_voltage_encoded")
...

# Or apply this "decode" function to the dataset and let the numpy vectorization take place there without needing to go to dataclasses at each epoch timestep.
output_data_array = xr.apply_ufunc(decode_ds, l1a_ds, ...)
```

imap_processing/glows/utils/constants.py
```diff
@@ -33,9 +35,11 @@ def __post_init__(self):
         object.__setattr__(self, "seconds", final_seconds)
         object.__setattr__(self, "subseconds", final_subseconds)
 
-    def to_seconds(self) -> float:
+    def to_seconds(self) -> np.single:
```
greglucas, line 38: Does GLOWS want float32 values? I'm wondering if it would be easier to keep all the math at 64bit (or whatever is passed in by default) and then just cast to single precision when saving?

imap_processing/tests/glows/test_glows_l1b.py
```diff
@@ -0,0 +1,247 @@
...
+        encoded_val,
+        encoded_val,
+        time_val,
+        time_val,
+        time_val,
+        time_val,
+    )
+
+    output = process_histogram(hist_dataset)
+    print(output)
```
greglucas:
```suggestion
```

imap_processing/tests/glows/test_glows_l1b_data.py
```diff
@@ -0,0 +1,70 @@
+import numpy as np
+import pytest
+
+from imap_processing.glows.l1b.glows_l1b_data import AncillaryParameters, DirectEventL1B
+
+
+def test_glows_l1b_ancillary_file():
+    fake_good_input = {
```
greglucas, line 8: Is this the same as your ancillary data fixture?

imap_processing/tests/glows/test_glows_l1b_data.py
```diff
@@ -0,0 +1,70 @@
...
+            "n_bits": 8,
+            "p01": 0.0,
+            "p02": 0.0,
+            "p03": 0.0,
+            "p04": 0.0,
+        },
+    }
+
+    ancillary = AncillaryParameters(fake_good_input)
+    for key in fake_good_input.keys():
```
greglucas: You don't need to call `keys()` directly on dicts (same with `len(dict.keys()) == len(dict)` above)
```suggestion
    for key in fake_good_input:
```

imap_processing/glows/ancillary/l1b_conversion_table_v001.json
maxinelasp: Yes, this is it. This one is not expected to change frequently, although I like the idea of the GLOWS team being able to update it without input from us...

imap_processing/glows/ancillary/l1b_conversion_table_v001.json
maxinelasp: Yes, good call. I renamed the file but it should match the other file versioning format, I think

imap_processing/glows/l1b/glows_l1b.py
maxinelasp: You're right about the I/O, I need to move that outside of the class anyway. Personally, I prefer the way it's organized in the dataclass rather than having loads of functions, but I can see the advantages of both. That being said, I don't actually think either method is much faster or slower (excluding the IO thing) because of the way `apply_ufunc` [is written](https://docs.xarray.dev/en/stable/generated/xarray.apply_ufunc.html). 

If I wanted it to be faster, I would need to write my own pre-vectorized functions so I could get that advantage from C, but as it is now, I am pretty sure it's doing similar amounts of work either way. In any case, I think this is still faster than what most of the other instruments are using or my previous designs, because it is guaranteed to only loop through the data one time. It doesn't need to loop through again to set up the datasets or to rearrange the dimensions. Plus, most of the real work inside the methods does take advantage of those numpy vectorizations that give speed. 

I prefer having it organized in a class so it's super explicit what the output is expected to be, especially in for example the direct events class where attributes are dropped or combined. I think it would be very confusing in a function call and difficult to tell at a glance what the actual output is.

imap_processing/glows/utils/constants.py
maxinelasp: I kind of just picked `np.single` at random, but actually thinking about it, it probably makes more sense to have doubles everywhere. I'll update it.

imap_processing/tests/glows/test_glows_l1b_data.py
maxinelasp: This is the same format as the ancillary data file with different input values.

imap_processing/glows/l1b/glows_l1b.py
maxinelasp: I did a quick profiler test with a simple example of both, and here are the results for the method: 
` 6438 function calls (6294 primitive calls) in 0.004 seconds`

vs the dataclass:
`6648 function calls (6495 primitive calls) in 0.004 seconds`

So, the dataclass might be a hair slower, but it's certainly comparable as long as the methods are well-written.

greglucas: Thanks for checking! Can you paste the quick example script?

> I prefer having it organized in a class so it's super explicit what the output is expected to be, especially in for example the direct events class where attributes are dropped or combined. I think it would be very confusing in a function call and difficult to tell at a glance what the actual output is.

👍 Just a difference of opinion, so that sounds good to me. I find functions easier to interpret at first glance 🤷

maxinelasp:
```
import dataclasses
from dataclasses import dataclass

import xarray as xr
import numpy as np

import cProfile

def foo(test_dataset):
    @dataclass
    class TestClass:
        direct_events: np.ndarray
        support_data_1: float
        support_data_2: float

        def __post_init__(self):
            self.support_data_1 = self.support_data_1 + 1
            self.support_data_2 = self.support_data_2 * 3

    def mapping(direct_events, support_data_1, support_data_2):
        result1 = direct_events
        result2 = support_data_1 + 1
        result3 = support_data_2 * 3
        return result1, result2, result3

    # Input and output core dims should include all dims EXCEPT Epoch
    # de_out, data_1_out, data_2_out = xr.apply_ufunc(
    #     mapping,
    #     test_dataset["direct_events"],
    #     test_dataset["support_data_1"],
    #     test_dataset["support_data_2"],
    #     input_core_dims=[["per_seconds"], [], []],
    #     output_core_dims=[["per_seconds"], [], []],
    #     vectorize=True,
    # )

    de_out2, data_1_out2, data_2_out2 = xr.apply_ufunc(
        lambda *args: tuple(dataclasses.asdict(TestClass(*args)).values()),
        test_dataset["direct_events"],
        test_dataset["support_data_1"],
        test_dataset["support_data_2"],
        input_core_dims=[["per_seconds"], [], []],
        output_core_dims=[["per_seconds"], [], []],
        vectorize=True,
    )

direct_events = np.array([[[1, 2, 3, 4], [5, 6, 7, 8]],
                              [[9, 10, 11, 12], [13, 14, 15, 16]],
                              [[17, 18, 19, 20], [21, 22, 23, 24]]])

direct_events_2 = np.array([[1, 2, 3, 4],
                            [9, 10, 11, 12],
                            [17, 18, 19, 20]])

support_data_1 = np.array([-2, -3, -4])
support_data_2 = np.array([1, 2, 3])
print(direct_events.shape)
time_data = np.array([100, 101, 102])

epoch = xr.DataArray(
    time_data,
    name="epoch",
    dims=["epoch"]
)

direct_event = xr.DataArray(
    np.arange(4),
    name="direct_event",
    dims=["direct_event"])

per_seconds = xr.DataArray(
    np.arange(4),
    name="per_seconds",
    dims=["per_seconds"])

de = xr.DataArray(
    direct_events_2,
    name="direct_events",
    dims=["epoch", "per_seconds"],
    coords={"epoch": epoch, "per_seconds": per_seconds})

test_dataset = xr.Dataset(
    coords={"epoch": time_data}
)

test_dataset["direct_events"] = de
test_dataset["support_data_1"] = xr.DataArray(
    support_data_1,
    name="support_data_1",
    dims=["epoch"],
    coords={"epoch": epoch}
)

test_dataset["support_data_2"] = xr.DataArray(
    support_data_2,
    name="support_data_2",
    dims=["epoch"],
    coords={"epoch": epoch}
)


cProfile.run('foo(test_dataset)')
```

greglucas: That's awesome @maxinelasp, thanks for providing it. I'm always a little leary of small array profiles. Here is a quick modification with the takeaway being that I'm still not sold on this approach, but I'm not going to block either.

```
# cProfile.run('foo_dataclass(test_dataset)')  # 7601328 function calls (7301312 primitive calls) in 1.383 seconds
# cProfile.run('foo_xarray_ufunc(test_dataset)')  # 901328 function calls (901312 primitive calls) in 0.172 seconds
# cProfile.run('foo_direct_call(test_dataset)')  # 278 function calls (276 primitive calls) in 0.000 seconds
```

```python
import dataclasses
from dataclasses import dataclass

import xarray as xr
import numpy as np

import cProfile

@dataclass
class TestClass:
    direct_events: np.ndarray
    support_data_1: float
    support_data_2: float

    def __post_init__(self):
        self.support_data_1 = self.support_data_1 + 1
        self.support_data_2 = self.support_data_2 * 3

def mapping(direct_events, support_data_1, support_data_2):
    result1 = direct_events
    result2 = support_data_1 + 1
    result3 = support_data_2 * 3
    return result1, result2, result3

def foo_dataclass(test_dataset):
    return xr.apply_ufunc(
        lambda *args: tuple(dataclasses.asdict(TestClass(*args)).values()),
        test_dataset["direct_events"],
        test_dataset["support_data_1"],
        test_dataset["support_data_2"],
        input_core_dims=[["per_seconds"], [], []],
        output_core_dims=[["per_seconds"], [], []],
        vectorize=True,
    )

def foo_xarray_ufunc(test_dataset):
    # Input and output core dims should include all dims EXCEPT Epoch
    return xr.apply_ufunc(
        mapping,
        test_dataset["direct_events"],
        test_dataset["support_data_1"],
        test_dataset["support_data_2"],
        input_core_dims=[["per_seconds"], [], []],
        output_core_dims=[["per_seconds"], [], []],
        vectorize=True,
    )


def foo_direct_call(test_dataset):
    return mapping(
        test_dataset["direct_events"],
        test_dataset["support_data_1"],
        test_dataset["support_data_2"]
    )

ntimes = 100000
nevents = 400
per_seconds = 30

direct_events = np.arange(ntimes*nevents*per_seconds).reshape(ntimes, nevents, per_seconds)
direct_events_2 = np.arange(ntimes*per_seconds).reshape(ntimes, per_seconds)

support_data_1 = np.arange(ntimes)
support_data_2 = np.arange(ntimes)
print(direct_events.shape)
time_data = np.arange(ntimes)

epoch = xr.DataArray(
    time_data,
    name="epoch",
    dims=["epoch"]
)

direct_event = xr.DataArray(
    np.arange(nevents),
    name="direct_event",
    dims=["direct_event"])

per_seconds = xr.DataArray(
    np.arange(per_seconds),
    name="per_seconds",
    dims=["per_seconds"])

de = xr.DataArray(
    direct_events_2,
    name="direct_events",
    dims=["epoch", "per_seconds"],
    coords={"epoch": epoch, "per_seconds": per_seconds})

test_dataset = xr.Dataset(
    coords={"epoch": time_data}
)

test_dataset["direct_events"] = de
test_dataset["support_data_1"] = xr.DataArray(
    support_data_1,
    name="support_data_1",
    dims=["epoch"],
    coords={"epoch": epoch}
)

test_dataset["support_data_2"] = xr.DataArray(
    support_data_2,
    name="support_data_2",
    dims=["epoch"],
    coords={"epoch": epoch}
)


a1, b1, c1 = foo_dataclass(test_dataset)
a2, b2, c2 = foo_xarray_ufunc(test_dataset)
a3, b3, c3 = foo_direct_call(test_dataset)

# Verify all functions provide same results
np.testing.assert_array_equal(a1, a2)
np.testing.assert_array_equal(a1, a3)
np.testing.assert_array_equal(b1, b2)
np.testing.assert_array_equal(b2, b3)
np.testing.assert_array_equal(c1, c2)
np.testing.assert_array_equal(c2, c3)

# cProfile.run('foo_dataclass(test_dataset)')  # 7601328 function calls (7301312 primitive calls) in 1.383 seconds
# cProfile.run('foo_xarray_ufunc(test_dataset)')  # 901328 function calls (901312 primitive calls) in 0.172 seconds
# cProfile.run('foo_direct_call(test_dataset)')  # 278 function calls (276 primitive calls) in 0.000 seconds
```

maxinelasp: I investigated your code a bit more, and turns out a lot of the slowdown from the dataclass is from the `dataclasses.asdict` method, which makes a deep copy. If I just manually write a function to return the tuple without a deepcopy: 

```
def output(self):
  return tuple(getattr(self, field.name) for field in dataclasses.fields(self))
```

Then I get something a lot closer to the direct function call: 

`2601395 function calls (2601379 primitive calls) in 0.448 seconds`

This is also a fixed slow down, so if the function calls are slower, the difference between the two methods stays at ~0.3s (which I checked by just adding a load of factorial calls in both methods)

My instincts are that they should be a lot closer together than that, so I think the slowdown is coming from inside the dataclass implementation. But, I still think it's worth the trade off for a very very minor performance hit to have a class which perfectly describes the output file and has docs right there for every attribute.

Also, thank you for the nerd snipe, I was supposed to be doing things today 😂 instead I was figuring out where that difference came from.

greglucas: :) Glad you're enjoying doing the deep dive!

> But, I still think it's worth the trade off for a very very minor performance hit to have a class which perfectly describes the output file and has docs right there for every attribute.

👍 That is fine, I just have the opposite opinion and find it more confusing to track that documentation than through the function calls, but no worries this is your instrument!

Note that the direct function call is actually the one I'd prefer, avoiding `xr.apply_ufunc` altogether. That is still more than 100x faster (maybe more because I'm just getting 0.000s).

PR #611, opened 2024-06-11
Fix for data_version in file output, using SWE as an example

maxinelasp:
# Change Summary
Right now, most instruments are not setting the `Data_version` global attribute correctly. This input value should come from the CLI and correspond to the version of the output file (in the filename, the "V002", for example) and not come from the code version (the "ground_software_version" attribute).

This change proposes a fix and a change to the CDF attribute manager which would allow for some dynamic values to be inserted into the CDF attributes. 

## Overview

* Updated the class to include a dyanamic way to add global variables
* Updated SWE as an example for what this would look like

## Alternatives
An alternative idea would be to make Data_version a required input into `ImapCdfManager` rather than allowing for dynamic input of global attributes. The add_global_attribute method is more flexible, but also might encourage misuse.

### DIFF
--- a/imap_processing/swe/l1a/swe_l1a.py
+++ b/imap_processing/swe/l1a/swe_l1a.py
@@ -12,7 +12,7 @@ from imap_processing.utils import group_by_apid, sort_by_time
 logger = logging.getLogger(__name__)
 
 
-def swe_l1a(file_path):
+def swe_l1a(file_path, data_version=None):
     """Process SWE l0 data into l1a data.
 
     Receive all L0 data file. Based on appId, it
@@ -23,6 +23,9 @@ def swe_l1a(file_path):
     ----------
     file_path: pathlib.Path
         Path where data is downloaded
+    data_version: str
+        Data version to write to CDF files and the Data_version CDF attribute.
+        Should be in the format Vxxx
 
     Returns
     -------
@@ -40,4 +43,4 @@ def swe_l1a(file_path):
     sorted_packets = sort_by_time(grouped_data[SWEAPID.SWE_SCIENCE], "ACQ_START_COARSE")
     logger.debug("Processing science data for [%s] packets", len(sorted_packets))
 
-    return swe_science(decom_data=sorted_packets)
+    return swe_science(decom_data=sorted_packets, data_version=data_version)
--- a/imap_processing/tests/swe/test_swe_l1b.py
+++ b/imap_processing/tests/swe/test_swe_l1b.py
@@ -69,11 +69,11 @@ def test_swe_l1b(decom_test_data):
 def test_cdf_creation():
     """Test that CDF file is created and has the correct name."""
     test_data_path = "tests/swe/l0_data/2024051010_SWE_SCIENCE_packet.bin"
-    l1a_datasets = swe_l1a(imap_module_directory / test_data_path)
+    l1a_datasets = swe_l1a(imap_module_directory / test_data_path, "v002")
 
     sci_l1a_filepath = write_cdf(l1a_datasets)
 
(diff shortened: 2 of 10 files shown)


tmplummer (commented):
Overall, these changes look good to me.

It does make me wonder how the call to the CLI knows, a priori, what version the output file should be. I would think that version would need to be determined after data has been processed and the full filename is known except for the version string. But this is based on assumptions I have about how file versions are handled.

tmplummer (commented):
Should the Data_version definition be removed from imap_processing/cdf/config/imap_default_global_cdf_attrs.yaml ? Or at least changed to "999" or something so that we know that it has not been set dynamically?

imap_processing/tests/swe/test_swe_l1b.py
```diff
@@ -69,11 +69,11 @@ def test_swe_l1b(decom_test_data):
 def test_cdf_creation():
     """Test that CDF file is created and has the correct name."""
     test_data_path = "tests/swe/l0_data/2024051010_SWE_SCIENCE_packet.bin"
-    l1a_datasets = swe_l1a(imap_module_directory / test_data_path)
+    l1a_datasets = swe_l1a(imap_module_directory / test_data_path, "v002")
 
     sci_l1a_filepath = write_cdf(l1a_datasets)
 
-    assert sci_l1a_filepath.name == "imap_swe_l1a_sci_20240510_v001.cdf"
+    assert sci_l1a_filepath.name == "imap_swe_l1a_sci_20240510_v002.cdf"
```
greglucas, line 78: This is `v002`, but the l1b below this is still `v001`, it looks like the version is set to `None` by default so I'm wondering where the `v001` is getting picked up from. I think this is related to @subagonsouth's comment about putting in a `v999` version attribute or something else to make it obvious we aren't producing the default data version.

maxinelasp: I did not update swe_l1b to retrieve the data version from the method call yet, so it's still using the old version.

maxinelasp:
@subagonsouth I do think it should be removed from there, thanks for catching it. If it is not included anywhere, attempting to generate the attributes should fail, so we can know right away.

maxinelasp:
As for the question about how the CLI knows, that's something that we need to do in sds-data-manager. As of now, I believe it just takes the version number of the input file, but @tech3371  has some drafts for version management that will solve it.

tmplummer (approved):
My comments have been addressed and I am comfortable with this implementation for the time being. Thanks for putting this together.

imap_processing/swe/l1a/swe_l1a.py
```diff
@@ -12,7 +12,7 @@
 logger = logging.getLogger(__name__)
 
 
-def swe_l1a(file_path):
+def swe_l1a(file_path, data_version=None):
```
tech3371, line 15: should we make all data_version to be a required input? At this point, the cli input should give us correct version to use (given that we implement logic in the sds-data-manager to find the correct version through database).

tech3371 (approved):
Just had one comment. Besides, it looks great! and Thank you for getting this in really quick.

imap_processing/swe/l1a/swe_l1a.py
maxinelasp: I think it probably should be required, but I made it optional for this change because I didn't want to have to update every test that uses this method 😁 I left of the swe_l1b change for the same reason - just trying to get it up quickly as an example.

tech3371: Ah ok. That makes total sense. I can update that in future PR. Thank you!

tech3371:
Can you merge this when you get chance? I have updated other instruments but will need to use new method for some instruments.

PR #614, opened 2024-06-12
Lo L1C CDF Creation for SIT-3

sdhoyt:
# Change Summary

## Overview
This PR adds adds the following: 

- Lo L1C CDF YAML attribute definitions for Pointing Sets (minimum goal for SIT-3)
- CDF Creation for Pointing Sets (minimum goal for SIT-3)
   - I have no sample data so there is temporary code in here to get things working for SIT-3
- Updates to cli.py to kick off Lo L1C processing

## New Dependencies
None

## New Files
- imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
   - handles Lo L1C processing

## Deleted Files
None

## Updated Files
- imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
   - Added L1C global attrs
- imap_processing/cli.py
   - added trigger for Lo L1C processing

## Testing
- added the unit test test_lo_l1b to test L1B processing
- ran cli.py and checked the correct file was created in the database


imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -0,0 +1,171 @@
...
+            pointing_bins.values.astype(str),
+            name="pointing_bins_label",
+            dims=["pointing_bins_label"],
+            attrs=attr_mgr.get_variable_attributes("pointing_bins_label"),
+        )
+        dataset = xr.Dataset(
+            coords={
+                "epoch": epoch_time,
+                "pointing_bins": pointing_bins,
+                "pointing_bins_label": pointing_bins_label,
```
tmplummer, line 121: I see that you are perhaps following what @tech3371 did. I didn't catch this in her PR, but I wonder if the label variables should really be coordinates or are they just regular variables?

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+
+    # Loop through the data fields that were pulled from the
+    # data class. These should match the field names given
+    # to each field in the YAML attribute file
+    for data_field in data_fields:
+        field = data_field.name.lower()
+        # Create a list of all the dimensions using the DEPEND_I keys in the
+        # YAML attributes
+        dims = [
```
tmplummer, line 135: Nice. I like that you are extracting the dims form the attributes. I need to implement this in my pointing set code.

imap_processing/tests/lo/test_lo_l1c.py
```diff
@@ -0,0 +1,94 @@
...
+
+    assert len(dataset.pointing_start.shape) == 1
+    assert dataset.pointing_start.shape[0] == 1
+    assert len(dataset.pointing_end.shape) == 1
+    assert dataset.pointing_end.shape[0] == 1
+    assert len(dataset.mode.shape) == 1
+    assert dataset.mode.shape[0] == 1
+    assert len(dataset.pivot_angle.shape) == 1
+    assert dataset.pivot_angle.shape[0] == 1
+    assert len(dataset.triples_counts.shape) == 3
```
tmplummer: I think that the individual asserts for each variable can be done in a single line such as:
```suggestion
    np.testing.assert_array_equal(dataset.triples_counts.shape, (1, 3600, 7))
```

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
```diff
@@ -0,0 +1,195 @@
...
+  UNITS: ' '
+  FORMAT: I1
+  VAR_TYPE: support_data
+  LABLAXIS: ESA
+  LABL_PTR_1: esa_step_label
+
+pointing_bins_label:
+  CATDESC: Pointing bins
+  FIELDNAM: Pointing bins
+  FORMAT: A3
```
tmplummer: I think that this has to be A4 to fit any value > 999.

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
```diff
@@ -0,0 +1,195 @@
...
+  MONOTON: INCREASE
+  TIME_BASE: J2000
+  TIME_SCALE: Terrestrial Time
+  REFERENCE_POSITION: Rotating Earth Geoid
+
+# Non-epoch Coordinates
+esa_step_label:
+  CATDESC: ESA Steps
+  FIELDNAM: ESA Steps
+  FORMAT: A3
```
tmplummer: Can this just be A1 since esa_step will always be one character?

tmplummer (commented):
This is great. I have a few comments. Thanks for the nice idea of how to generate the `dims` keyword argument needed for the `xr.DataArray` constructor.

imap_processing/lo/l1c/lo_l1c.py
greglucas: I agree this is nice. I wonder if we should add this as a general helper to the AttributeManager class (not in this PR). `get_variable_dims()` (naming TBD) that would encapsulate this logic for everyone.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -0,0 +1,171 @@
...
+            dataset[field] = xr.DataArray(
+                data=[1],
+                dims=dims,
+                attrs=attr_mgr.get_variable_attributes(field),
+            )
+        # TODO: This is temporary.
+        #  The data type will be set in the data class when that's created
+        elif field == "exposure_time":
+            dataset[field] = xr.DataArray(
+                data=[7 * [np.float16(1)]],
```
greglucas: I'd suggest using numpy array shapes for all of these temporary data fields which is a bit more explicit IMO.

```suggestion
                data=np.ones((1, 7), dtype=np.float16),
```

imap_processing/tests/lo/test_lo_l1c.py
greglucas: Maybe just compare against an array itself which will also check shape for you: `np.testing.assert_array_equal(dataset.triples_counts, np.ones((1, 3600, 7)))`

imap_processing/tests/lo/test_cdfs/imap_lo_l1a_de_20100101_v001.cdf
greglucas, line 1: It doesn't look like the l1a files added here aren't being read in anywhere and thus are unused I think? Can you remove these from the commits?

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: Yeah, I need to get better at recognizing when I can use numpy

imap_processing/tests/lo/test_cdfs/imap_lo_l1a_de_20100101_v001.cdf
sdhoyt: I added these to a test_l1a_cdfs folder in my last PR, but decided to move them to a general test_cdfs folder and forgot to delete the original ones. I'll delete the extras. In a future PR I will remove all the test cdfs entirely and create a dataset in the unit test

greglucas: Perfect, the git move rename is much better. Thanks!

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: That's a good question. If they can be normal variables, I feel like it still makes more sense to have them as coordinates since they're basically string versions of the coordinates. I haven't looked at it much yet, but I'm hoping there is a good way to just convert the data to a string and use that as the label within write_cdf or cdflib instead of having to define attributes for the label since all the attributes are easily determined (ie. string length) or straight from the coordinate attributes.

greglucas (approved):

tmplummer (approved):
Looks good to me.

PR #616, opened 2024-06-12
HIT L1B CDF Creation for SIT-3

vmartinez-cu:
# Change Summary

## Overview
This PR adds code for L1B processing of housekeeping data for SIT-3. Updates will be needed to handle science data once we have more information from the instrument team post SIT-3.

## New Dependencies
None

## New Files
- hit_l1b.py
   - handles l1b processing. currently just handles housekeeping data
- test_hit_l1b.py
   - tests creating l1b cdf housekeeping product

## Deleted Files
None

## Updated Files
- hit_cdf_attrs.py
   - updated fill_val for variables with float values
- cli.py
   - added l1b processing for HIT

## Testing
test_hit_l1b.py tests creating l1b cdf file for housekeeping


imap_processing/cli.py
```diff
@@ -440,6 +441,17 @@ def do_processing(self, dependencies):
             products = hit_l1a(dependencies[0])
             return products
 
+        elif self.data_level == "l1b":
+            if len(dependencies) > 1:
```
sdhoyt, line 445: I missed this in the last PR and it doesn't matter at the moment, but the science files won't be 1 to 1. For L1A and L1B there could be multiple dependencies, so those two `if len(dependencies) > 1` checks should go away. It it's not super straightforward to remove them now it can wait.

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -0,0 +1,165 @@
...
+    # Create datasets
+    datasets = []
+    if "_hk" in logical_source:
+        dataset = create_hk_dataset()
+        datasets.append(dataset)
+
+    # Create CDF files
+    logger.info("Creating CDF files for HIT L1B data")
+    cdf_filepaths = []
+    for dataset in datasets:
```
sdhoyt, line 49: I'm not sure how things will work with the batch starter and dependency config file when we have one L1A science file creating three L1B files. I'm not sure if batch start will kick off three separate jobs with a different downstream dependency each time or if they will all be created in the same job. Nothing to change for this, that's something I'm not sure about that we might need to check

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -0,0 +1,165 @@
...
+                for key, value in hit_cdf_attrs.l1b_hk_attrs[field].output().items()
+                if "DEPEND" in key
+            ]
+
+            # TODO: This is temporary.
+            #  The data will be set in the data class when that's created
+            if field == "leak_i":
+                # 2D array - needs two dims
+                hk_dataset[field] = xr.DataArray(
+                    [[0, 0, 1], [0, 1, 0], [0, 0, 1]],
```
sdhoyt: I think ADC channel dimension for leak current is supposed to have a length of 64. It might be good to make these three lists 64 indices long. You could do something like `3 * [64 * [1]]` to give you three entries of 64 index long lists of 1.

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Create xarray data arrays for dependencies
+    epoch_time = xr.DataArray(
+        data=epoch_converted_time,
+        name="epoch",
+        dims=["epoch"],
+        attrs=ConstantCoordinates.EPOCH,
+    )
+
+    adc_channels = xr.DataArray(
```
sdhoyt, line 98: Should change this to 64 instead of 3

sdhoyt (commented):
Great work! just some minor comments

imap_processing/hit/l1b/hit_l1b.py
tech3371: I think it's ok at the moment since we have not added dependency passed `l1b` level for hit. It just won't kickoff process after this.

tech3371: but can we indent these line in `elif` statement.
```
if hk:
  process housekeping
  append to dataset list
elif science:
   process science data
         append processed data to dataset list
```

imap_processing/cli.py
sdhoyt: I have no idea what I was thinking when I wrote this. There are no files in L1A or L1B that rely on multiple dependencies, so what you have is correct. You can ignore this comment.

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -0,0 +1,165 @@
...
+        np.array(np.arange(3), dtype=np.uint16),
```
greglucas:
```suggestion
        np.arange(3, dtype=np.uint16),
```

imap_processing/hit/l1b/hit_l1b.py
greglucas: `np.ones((3, 64))`

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -0,0 +1,165 @@
...
+                "ebox_m5d7va",
+                "ref_p5v",
+                "l1ab_bias",
+                "l2ab_bias",
+                "l34a_bias",
+                "l34b_bias",
+                "ebox_p2d0vd",
+            ]:
+                hk_dataset[field] = xr.DataArray(
+                    [np.float16(1), np.float16(1), np.float16(1)],
```
greglucas:
```suggestion
                    np.ones(3, dtype=np.float16),
```

imap_processing/hit/l1b/hit_l1b.py
greglucas: Maybe make it a variable to use in your shape.size creation below as well?
`nchannels = 64` ... `np.ones((nepochs, nchannels))`

imap_processing/tests/hit/test_data/imap_hit_l1a_hk_20100105_v001.cdf
greglucas, line 1: This is quite a large file, especially if we possibly aren't going to be producing hk data in the future. I have a mild preference for removing from this commit. Can you create this l1a dataset in your l1b code for re-use, rather than saving it to the repo?

imap_processing/tests/hit/test_hit_l1b.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@pytest.fixture(scope="module")
+def dependency():
+    """Set path to test data file"""
+
+    cdf_filepath = (
+        imap_module_directory / "tests/hit/test_data/imap_hit_l1a_hk_20100105_v001.cdf"
+    )
+    l1a_data = load_cdf(cdf_filepath)
```
greglucas:
```suggestion
    pkts_filepath = (
        imap_module_directory / "tests/hit/test_data/..."
    )
    l1a_data = load_cdf(l1a_function(pkts_filepath))
```

vmartinez-cu: This is much better, thanks for the suggestion. Code has been updated

imap_processing/tests/hit/test_data/imap_hit_l1a_hk_20100105_v001.cdf
vmartinez-cu: Agreed. File has been removed from the repo

greglucas (approved):

sdhoyt (approved):

PR #617, opened 2024-06-12
Codice l1a CDF fixes

bourque:
# Change Summary

## Overview

This PR fixes some issues with the CoDICE L1a CDFs I was producing; namely adding the missing "Instrument_type" attribute, and implementing a workaround for an issue where `cdflib` doesn't properly write/read a CDF with only one `epoch` value.  I also updated the L0 houskeeping filename to be consistent with the L1a file it produces.


tech3371 (approved):

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -26,7 +26,7 @@ energy_attrs:
     VALIDMAX: 127
     VAR_TYPE: support_data
 
-epoch:
+epoch_attrs:
```
greglucas, line 29: Why did you change this naming? I thought it should be mapped to the variable names directly.

imap_processing/codice/codice_l1a.py
```diff
@@ -97,12 +97,12 @@ def create_science_dataset(self, start_time: np.datetime64) -> xr.Dataset:
         cdf_attrs.add_instrument_global_attrs("codice")
         cdf_attrs.add_instrument_variable_attrs("codice", "l1a")
 
-        # Define DataArrays for the coordinates
+        # Define coordinates
         epoch = xr.DataArray(
-            [start_time],
+            [start_time, start_time],  # TODO: Fix after SIT-3 (see note below)
```
greglucas: Can you make these different values too? The upcoming cdflib will complain if they are duplicates.
`[start_time, start_time + np.timedelta64(1, "s")]`

imap_processing/codice/codice_l1a.py
```diff
@@ -111,16 +111,32 @@ def create_science_dataset(self, start_time: np.datetime64) -> xr.Dataset:
...
+            },
             attrs=cdf_attrs.get_global_attributes(self.dataset_name),
         )
 
-        # Create a data variable for each species
+        # Create a data variable for each counter
         for variable_data, variable_name in zip(self.data, self.variable_names):
-            variable_data_arr = np.array(list(variable_data), dtype=int).reshape(
-                -1, self.num_energy_steps
+            # TODO: Currently, cdflib doesn't properly write/read CDF files that
```
tmplummer, line 137: It would be good to broadcast this issue a bit more widely. There are several/many of us with products that have a single epoch value that should be made aware of the issue. I'd like to know some more details about how the issue manifests.

tmplummer (approved):
Changes look good to me.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: I was messing with this when trying to debug some errors, but decided to keep it because I thought it was more explicit (and thus better). Happy to change it back if you feel strongly about it.

imap_processing/codice/codice_l1a.py
bourque: Ah yes, great catch!

imap_processing/codice/codice_l1a.py
bourque: Good suggestion, I will mention it in slack

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
greglucas: Nope, I just like consistency, but it appears we aren't consistent at the moment when looking through other instruments yamls, so no problem choosing one here :)

greglucas (approved):

PR #618, opened 2024-06-12
Glows l1b cdf

maxinelasp:
# Change Summary
GLOWS L1B CDF first draft

## Overview

Added GLOWS CDF attribute files and added GLOWS L1A and L1B to the CLI. This completes work for GLOWS prior to the SIT.


imap_processing/glows/l1b/glows_l1b.py
```diff
@@ -104,7 +211,7 @@ def process_histogram(l1a: xr.Dataset) -> xr.Dataset:
     dataarrays = [l1a[i] for i in l1a.keys()]
 
     dims = [[] for i in l1a.keys()]
-    # 34 is the number of output attributes in the HistogramL1B class.
+    # 35 is the number of output attributes in the HistogramL1B class.
```
greglucas: Do you need to update the range(34) call below too then? There is mismatch between 34/35 here.

imap_processing/glows/l1b/glows_l1b.py
```diff
@@ -114,28 +221,31 @@ def process_histogram(l1a: xr.Dataset) -> xr.Dataset:
 
     # This preserves the dimensions for the histogram output
     new_dims[0] = ["bins"]
-    new_dims[22] = ["bins"]  # For imap_spin_angle_center - a per-bin value
+    new_dims[21] = ["bins"]  # For imap_spin_angle_center - a per-bin value
```
greglucas: This is pretty cryptic and seems easy to get off-by-one errors (for example, I am unclear why this has changed all of a sudden now). Is there a way we can have `new_dims` be string lookup instead of index-based?

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -277,9 +278,11 @@ def __post_init__(
         )
 
         # TODO: double check that this time is in unix time and is the correct variable
-        self.unique_identifier = np.datetime_as_string(
-            np.datetime64(int(self.imap_time_last_pps), "ns"), "s"
-        )
+        # TODO: This cannot be in the data because it's a string, put it in the
```
greglucas, line 281: We are putting labels in variables now, so it should be OK to have strings defined in CDF variable outputs. Just fine to come back to this later though.

imap_processing/cdf/config/imap_glows_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,336 @@
+int_fillval: &int_fillval -9223372036854775808
+valid_max: &valid_max 3155630469184000000
+
+epoch_attrs:
```
greglucas: I commented on @bourque's PR as well because I thought we were doing a 1-1 mapping of variable names to the yaml attrs definitions, but now I see this might be the convention. I'm also confused though because there are also some variables below without the `_attrs` portion. Is there a convention we should be following?

maxinelasp: Here, the `_attrs` indicate dimensions, but that is ... not a good way of doing that, so I'll rename them. But there aren't really any conventions yet for this documents, so it's a little bit wild west...

imap_processing/glows/l1b/glows_l1b_data.py
maxinelasp: It would cause errors within cdflib when writing. I also am not convinced that this should be a data variable, so it maybe should be a global attribute instead? This is another todo that is waiting on commentary from the GLOWS team.

imap_processing/glows/l1b/glows_l1b.py
maxinelasp: The code is correct, the number is wrong. I might just remove the number from the comment because it's confusing.

imap_processing/glows/l1b/glows_l1b.py
maxinelasp: `new_dims` must be index based, but I can write a lookup method using the dataclass to generate it.

maxinelasp: Rewrote it and I think it's way better, actually!

imap_processing/cdf/config/imap_glows_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,27 @@
...
+  GLOWS>GLObal Solar Wind Structure
+TEXT: &txt >
+  The GLObal Solar Wind Structure (GLOWS) is a non-imaging single-pixel Lyman-alpha
+  photometer that will be used to observe the sky distribution of the helioglow to
+  better understand the evolution of the solar wind structure.
+  The Lyman-alpha photon counts from these observations can be used to build a more
+  comprehensive picture of the solar wind structure and how it changes through the
+  solar cycles.
+  GLOWS design and assembly is led by the Space Research Center, Warsaw, Poland
+  (CBK PAN). See https://imap.princeton.edu/instruments/glows for more details.
```
tech3371: Don't forget to add this here and other areas. This has been causing issues with generating valid CDF files.
```suggestion
  (CBK PAN). See https://imap.princeton.edu/instruments/glows for more details.
Instrument_type: Particles (space)
```

tech3371 (approved):

PR #619, opened 2024-06-12
Data version update

tech3371:
# Change Summary

## Overview
Updated `Data_version` CDF global attrs for these instrument:
- CoDICE
- Hi
- HIT
- IDEX
- Lo
- SWAPI
- Ultra
- SWE

### DIFF
--- a/imap_processing/tests/hi/test_l1a.py
+++ b/imap_processing/tests/hi/test_l1a.py
@@ -11,10 +11,12 @@ def test_sci_de_decom(create_de_data):
     """Test science direct event data"""
 
     # Process using test data
-    processed_data = hi_l1a(packet_file_path=create_de_data(HIAPID.H45_SCI_DE.value))
+    processed_data = hi_l1a(
+        packet_file_path=create_de_data(HIAPID.H45_SCI_DE.value), data_version="v001"
+    )
 
     assert processed_data[0].attrs["Logical_source"] == "imap_hi_l1a_45sensor-de"
-    assert processed_data[0].attrs["Data_version"] == "001"
+    assert processed_data[0].attrs["Data_version"] == "v001"
 
     # unique ESA steps should be [1, 2]
     assert np.array_equal(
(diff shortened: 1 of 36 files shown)


imap_processing/tests/hi/test_l1a.py
```diff
@@ -11,10 +11,12 @@ def test_sci_de_decom(create_de_data):
...
 
     # Process using test data
-    processed_data = hi_l1a(packet_file_path=create_de_data(HIAPID.H45_SCI_DE.value))
+    processed_data = hi_l1a(
+        packet_file_path=create_de_data(HIAPID.H45_SCI_DE.value), data_version="v001"
+    )
 
     assert processed_data[0].attrs["Logical_source"] == "imap_hi_l1a_45sensor-de"
-    assert processed_data[0].attrs["Data_version"] == "001"
+    assert processed_data[0].attrs["Data_version"] == "v001"
```
tmplummer: Are we sure that we want to store "vXXX" in the `Data_version` global attribute? It seems like "v" is already implied by the attribute name.

tmplummer (approved):
Overall, looks good to me.

imap_processing/tests/hi/test_l1a.py
tech3371: This one is tricky because `write_cdf` appends `v` if it doesn't exist already or format int into `vxxx`. But lot of tests are testing `dataset` before it's written to cdf. That's why I wrote it as it is. But for `hi`, looks you had version as `001` which I can convert it back to what you had.

tech3371: On second thought, I will try to convert version back to `001` and let `write_cdf` handle adding v to it to be consistent across all.

tmplummer: I think that consistency is more important. No need to change anything here, but it might be good to have a future discussion about this.

tech3371: I agree. I changed that already :)

maxinelasp (approved):
LGTM! That PR is merged, so you can uncomment the "add_global_attribute" things :)

greglucas (approved):
Looks good. Something to consider in the future is whether we could somehow add this to `write_cdf()` instead of needing to pass it to every calling function.

Just to expand on that thought so I don't forget later: Have all instruments return datasets rather than filepaths, have the cli write out the datasets as appropriate for everyone, which would then mean we wouldn't need to pass the dataset version into processing routines.

bourque (approved):
Thank you for doing this!

PR #622, opened 2024-06-13
Inital Codice L1b processing (for SIT-3)

bourque:
# Change Summary

## Overview

This PR adds initial support for CoDICE L1b data products. It is pretty bare bones for now, just to get data products produced for the SIT-3 test (for example, the data arrays are simply being converted from integers to floats). This will be revisited after SIT-3 to actually apply appropriate processing.

## New Files

- `imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml`
  - Added L1b attributes 
- `imap_processing/codice/codice_l1b.py`
  - Initial implementation of L1b processing  
- `imap_processing/tests/codice/test_codice_l1b.py`
  - Tests for L1b processing  

## Updated Files

- `imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml`
  - Added L1b attributes 
- `imap_processing/cli.py`
  - Added L1b processing


imap_processing/codice/codice_l1b.py
```diff
@@ -0,0 +1,196 @@
...
+    else:
+        l1b_dataset = create_science_dataset(l1a_dataset, cdf_attrs, dataset_name)
+
+    # Write the dataset to CDF
+    logger.info(f"\nFinal data product:\n{l1b_dataset}\n")
+    l1b_dataset.attrs["Data_version"] = data_version
+    l1b_dataset.attrs["cdf_filename"] = write_cdf(l1b_dataset)
+    logger.info(f"\tCreated CDF file: {l1b_dataset.cdf_filename}")
+
+    return l1b_dataset
```
sdhoyt, line 196: Could return just the filename instead of the whole dataset?

imap_processing/codice/codice_l1b.py
```diff
@@ -0,0 +1,196 @@
...
+
+    if "hskp" in dataset_name:
+        l1b_dataset = create_hskp_dataset(l1a_dataset, cdf_attrs)
+
+    else:
+        l1b_dataset = create_science_dataset(l1a_dataset, cdf_attrs, dataset_name)
+
+    # Write the dataset to CDF
+    logger.info(f"\nFinal data product:\n{l1b_dataset}\n")
+    l1b_dataset.attrs["Data_version"] = data_version
```
sdhoyt: Does adding the version directly replace needing to make this call? `attr_mgr.add_global_attribute("Data_version", data_version)`

sdhoyt (approved):
Everything looks good to me. Just a couple minor questions. 

Also, the unit test failure your getting is because I messed up a Lo merge. If you rebase that should go away.

imap_processing/codice/codice_l1b.py
bourque: Yeah, I think I could do that for L1b. For L1a, I am returning the whole dataset to make it easier to write unit tests (in that way, my test code doesn't need to go open the file), but since my L1b tests are just checking the filename, I could just return that instead.

imap_processing/codice/codice_l1b.py
bourque: Good catch. i just forgot to update this to use that method of setting the `Data_version`.

imap_processing/codice/codice_l1b.py
sdhoyt: Gotcha, yeah, whatever works better for you

PR #623, opened 2024-06-13
Mag L1B/L1C cdf

maxinelasp:
# Change Summary
This is the change to create MAG L1B and L1C output files. It does a first pass at MAG L1B processing as well, since it's very simple. Mag L1C processing is not complete.

## Overview

- Added MAG L1B processing first draft
- Added MAG L1B and L1C attributes
- Added MAG L1C basic output
- Added L1B and L1C to cli.py

I have tested this against the CLI locally and it works great. I may need to update the SDS dependencies for L1C.


imap_processing/mag/l1b/mag_l1b.py
```diff
@@ -0,0 +1,112 @@
...
+    attribute_manager = ImapCdfAttributes()
+    attribute_manager.add_instrument_global_attrs("mag")
+    attribute_manager.add_global_attribute("Data_version", version)
+
+    # Variable attributes can remain the same as L1A
+    input_logical_source = input_dataset.attrs["Logical_source"]
+    if isinstance(input_dataset.attrs["Logical_source"], list):
+        input_logical_source = input_dataset.attrs["Logical_source"][0]
+
+    logical_source = input_logical_source.replace("l1a", "l1b")
```
sdhoyt, line 42: Nice, this is a cool way to handle the 1 to 1 products

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -0,0 +1,52 @@
...
+        mago or magi.
+    second_input_dataset : xr.Dataset
+        The second input dataset to process. This should be burst if first_input_dataset
+        was norm, or norm if first_input_dataset was burst. It should match the
+        instrument - both inputs should be mago or magi.
+    version : str
+        The version of the output data
+    """
+    # TODO: L1C processing involves filling gaps with burst data.
+    input_logical_source_1 = first_input_dataset.attrs["Logical_source"]
```
sdhoyt, line 29: I probably didn't look at this code long enough to see why that won't work, but could you pass in the two datasets as one list and get the index of of the one with norm below and use that to index the list? Something like

```python
norm_idx = [index for index, value in enumerate(data_list) if 'norm' in value]
data_list[norm_idx[0]]
```

sdhoyt (approved):
Everything looks good to me

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: Maybe? Probably? I think because the logical source is sometimes a list and sometimes not, this approach would end up being difficult to parse.

imap_processing/mag/l1b/mag_l1b.py
```diff
@@ -0,0 +1,112 @@
...
+       calibration_matrix : xr.DataArray
+           The full set of calibration matrices, for each range. Size is ((3, 3, 4))
+    """
+    updated_vector = input_vector.copy()
+
+    updated_vector[:3] = np.matmul(
+        input_vector[:3], calibration_matrix.values[:, :, int(input_vector[3])]
+    )
+
+    return updated_vector
```
greglucas, line 112: Great comments on what this is doing, it was super helpful to read and follow exactly what is going on.

Do you want the "range" to be carried along, or is that just used for lookup here and can then be discarded as you go to higher data levels? If you can discard it, then you could inline it directly.
```suggestion
    return input_vector[:3] @ calibration_matrix.values[:, :, int(input_vector[3])]
```

imap_processing/mag/l1b/mag_l1b.py
```diff
@@ -0,0 +1,112 @@
...
+    calibration_dataset = load_cdf(
+        Path(__file__).parent / "imap_calibration_mag_20240229_v01.cdf"
+    )
+    # TODO: Check validity of time range for calibration
+    if "mago" in input_dataset.attrs["Logical_source"][0]:
+        calibration_matrix = calibration_dataset["MFOTOURFO"]
+    else:
+        calibration_matrix = calibration_dataset["MFITOURFI"]
+
+    l1b_fields = xr.apply_ufunc(
```
greglucas, line 75: You should be able to vectorize this as well I think rather than going through each vector individually. I still am not a fan of `apply_ufunc` ;) This will be way faster and it is a bit tricky to follow but with comments I think it can make quite a bit of sense for the future.

```python
ranges = input_vectors[:, 3].astype(int)  # N
calibration_matrices = calibration_dataset[:, :, ranges].T  # (N, 3, 3)
# grab the (N, 3) input vectors
# add a new axis to make it (N, 1, 3)
# matmul the outer two dimensions to produce (N, 1, 3) rotated vectors
# Remove the extra inner dimension
output_vectors = (input_vectors[:, :3][:, np.newaxis, :] @ calibration_matrices)[:, 0, :]
```

imap_processing/tests/mag/test_mag_l1b.py
```diff
@@ -0,0 +1,77 @@
...
+
+    return output_dataset
+
+
+def test_mag_processing(mag_l1a_dataset):
+    mag_l1a_dataset.attrs["Logical_source"] = ["imap_mag_l1a_norm-mago"]
+
+    mag_l1b = mag_l1b_processing(mag_l1a_dataset)
+
+    assert np.allclose(
```
greglucas: Use the numpy testing helpers if possible. This will handle shape comparisons as well.
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html
```suggestion
    np.testing.assert_allclose(
```

imap_processing/mag/l1b/mag_l1b.py
```diff
@@ -0,0 +1,112 @@
...
+from pathlib import Path
+
+import numpy as np
+import xarray as xr
+
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
+from imap_processing.cdf.utils import load_cdf
+
+
+def mag_l1b(input_dataset: xr.Dataset, version: str) -> xr.Dataset:
```
greglucas, line 12: I like that you're passing Datasets in/out, seems like a nice clean way to do it!

greglucas (approved):
Just some minor comments, looks good overall.

imap_processing/mag/l1b/mag_l1b.py
maxinelasp: The range needs to be carried along

imap_processing/mag/l1b/mag_l1b.py
maxinelasp: There are future changes that I think would be useful to have apply_ufunc for, but I will keep this in mind for my next update if it ends up remaining this way. I agree it's a little overkill for what it is now.

PR #624, opened 2024-06-13
REL/CI: Add build job dependency to release process

greglucas:
# Change Summary

## Overview

Adds an explicit dependency on building the sdist before deploying to PyPI.


bourque (approved):
Thanks for fixing that!

PR #626, opened 2024-06-14
HIT Data_version fixes

tech3371:
# Change Summary

## Overview
Minor fixes to reflect recent changes in `Data_version`

### DIFF
--- a/imap_processing/hit/l1b/hit_l1b.py
+++ b/imap_processing/hit/l1b/hit_l1b.py
@@ -17,13 +17,15 @@ logger = logging.getLogger(__name__)
 # TODO review logging levels to use (debug vs. info)
 
 
-def hit_l1b(l1a_dataset: xr.Dataset):
+def hit_l1b(l1a_dataset: xr.Dataset, data_version: str):
     """Process HIT data to L1B.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


sdhoyt (approved):

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -47,6 +49,7 @@ def hit_l1b(l1a_dataset: xr.Dataset):
     logger.info("Creating CDF files for HIT L1B data")
     cdf_filepaths = []
     for dataset in datasets:
+        dataset.attrs["Data_version"] = data_version
```
sdhoyt, line 52: Does this also need to make this call? `attr_mgr.add_global_attribute("Data_version", data_version)`

sdhoyt: I asked Maxine about this and her explaining how these should be used reminded me that HIT hasn't changed to using YAML, so that's why 

```python
dataset.attrs["Data_version"] = data_version
```
is being used instead of
```python
```

This comment can be ignored for now.

sdhoyt:
@tech3371 Merged this PR for you. Thanks for making this change!

PR #635, opened 2024-06-17
Update for Sit 3

laspsandoval:
# Change Summary

## New Files

- new file 1

## Deleted Files

- deleted file 1

## Updated Files

- updated file 1
- updated file 2


tech3371 (approved):
nice find!

imap_processing/ultra/l0/decom_tools.py
```diff
@@ -182,9 +182,9 @@ def decompress_image(
     blocks_per_row = cols // pixels_per_block
 
     # Compressed pixel matrix
-    p = np.zeros((rows, cols), dtype=np.uint16)
+    p = np.zeros((rows, cols), dtype=np.int16)
```
greglucas, line 185: Do you actually want to allow negative values for these pixel matrices? I worry about the bit manipulations later on with `int` rather than `uint` if we are carrying around that extra negative bit and where that is.

Would you actually want to increase the size to `uint32` or `uint64`?

laspsandoval: I need to revisit why this is throwing an error. It's working now, but I'm putting a todo.

PR #636, opened 2024-06-17
Mag hotfix

maxinelasp:
# Change Summary
Quick change to have MAG processing work. This seems to be related to the release of Numpy 2.0. So I added an explicit typecast and numpy in the dependencies list to keep it at 1.26 for now.


greglucas (approved):

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -163,8 +163,10 @@ def process_packets(
         total_primary_vectors = seconds_per_packet * mag_l0.PRI_VECSEC
         total_secondary_vectors = seconds_per_packet * mag_l0.SEC_VECSEC
 
+        # The raw vectors are of type int8, but the output vectors should be at least
+        # int16.
         primary_vectors, secondary_vectors = MagL1a.process_vector_data(
-            mag_l0.VECTORS, total_primary_vectors, total_secondary_vectors
+            mag_l0.VECTORS.astype(dtype=np.int32), total_primary_vectors, total_secondary_vectors
```
tech3371: glad it was similar thing as Laura

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -163,8 +163,12 @@ def process_packets(
         total_primary_vectors = seconds_per_packet * mag_l0.PRI_VECSEC
         total_secondary_vectors = seconds_per_packet * mag_l0.SEC_VECSEC
 
+        # The raw vectors are of type int8, but the output vectors should be at least
+        # int16.
         primary_vectors, secondary_vectors = MagL1a.process_vector_data(
-            mag_l0.VECTORS, total_primary_vectors, total_secondary_vectors
+            mag_l0.VECTORS.astype(dtype=np.int32),
```
tech3371, line 169: glad it was this issue instead of something worse :)

tech3371 (approved):

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Yeah turns out we were using numpy 2.0, so that probably was causing some weird new issues! Hopefully nothing more should pop up because the version is pinned in pyproject.toml.

PR #637, opened 2024-06-17
Cdf attribute tests

### LINKED ISSUE #592: Update CDF attribute generator unit tests
### Summary

The CDF attribute generator in cdf/cdf_attribute_manager.py needs additional unit tests. There is a start within the tests/ folder, but all the methods in cdf_attribute_manager need to be tested. 

### Requirements

The more specific the better.
  - [ ] Write unit tests for the methods in cdf_attribute_manager that are missing
  - [ ] Create a file for imap_cdf_manager and add tests specific to imap_cdf_manager.py
  - [ ] If needed, expand existing unit tests
  - [ ] If needed, correct or expand documentation

### Related tickets

_No response_

anamanica:
# Change Summary

## Overview
Creating CDF attribute code tests for the two files:
    1.) cdf_attribute_manager.py
    2.) imap_cdf_manager.py

## New Dependencies
None

## New Files
For CDF Attribute Manager:
- imap_default_global_test_cdf_attribues.yaml 
    - Contains test .yaml for global variables
    - General info (tests load_global_attributes)
    - Also used in testing variable attributes
- imap_test_global.yaml
   - Contains test .yaml for global cdf attributes 
   - INSTRUMENT SPECIFIC (tests get_global_attributes)
- imap_test_variable.yaml 
    - Contains test .yaml for variable cdf attributes
    - INSTRUMENT SPECIFIC (tests get_variable_attributes)
 

For IMAP CDF Manager
- test_imap_cdf_manager.py
   - Contains unit tests for imap_cdf_manager.py
 - imap_instrument_global_cdf_attrs.yaml
    - Tests add_instrument_global_attrs
    - Same information as imap_test_global.yaml, but needed a specific file naming convention.
 - imap_instrument2_global_cdf_attrs.yaml
    - Tests add_instrument_global_attrs
    - Same information as imap_swe_global.yaml, but needed a specific file naming convention.
 - imap_instrument_level_variable_attrs.yaml
    - Tests add_instrument_variable_attrs
    - Same information as  imap_test_variable.yaml, but again, needed a specific file naming convention
- imap_instrument2_level2_variable_attrs.yaml
    - Tests add_instrument_variable_attrs
    - Same information as  imap_ultra_l1b_variable.yaml, but again, needed a specific file naming convention

## Deleted Files
None

## Updated Files
- test_cdf_attribute_manager.py
   - Creating and improving unit tests
- imap_cdf_manager.py
   - Added an optional parameter to `__init__` so that I could reach my test files while also preserving the API. 

## Testing
The purpose of this ticket is to specifically create unit tests, and thus all the work this ticket is unit testing.

closes #592


imap_processing/cdf/config/imap_instrument_level_variable_attrs.yaml
```diff
@@ -0,0 +1,23 @@
+default_attrs: &default
```
maxinelasp: These YAML files look good, but they should go under tests/

imap_processing/cdf/config/imap_test_global.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+imap_test_section_4:
+    Descriptor: *desc
+    TEXT: *txt
+    Data_type: L1A_raw-norm-magi>Level 1A MAGi normal rate
+    Logical_source: imap_mag_l1a_norm-magi
+    Logical_source_description: IMAP Mission MAGi Normal Rate Instrument Level-1A Data.
+
+
+imap_test_section_5:
```
maxinelasp: You don't need more than 2 imap_test_sections IMO. What is adding additional sections adding from a test perspective? One test is good to validate that you can get into the section. Two is good to validate that you can choose between the sections. But adding additional sections doesn't test anything extra.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,47 +1,121 @@
...
+    """
+
+    # Test
+
+    # Initialize CdfAttributeManager object which loads in default schema
+    cdf_manager = CdfAttributeManager(Path(__file__).parent.parent / "config")
+
+    # Default global tests
+    # Check false case
+    assert cdf_manager.global_attribute_schema["DOI"]["required"] is False
```
maxinelasp, line 19: It's a very minor thing, but you can also check falseness by saying `assert not [statement]` or check trueness with `assert [statement]`

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,47 +1,121 @@
...
 def test_global_attribute():
+    """
+    Test function that covers:
+        load_global_attributes
+        get_global_attributes
+    """
+
+    # Initialize CdfAttributeManager object which loads in default global variables and schema
     cdf_manager = CdfAttributeManager(Path(__file__).parent.parent / "config")
+    # Test that default information was loaded in from "imap_default_global_cdf_attrs.yaml"
```
maxinelasp: hm, this might actually be a bug within `CdfAttributeManger`. I don't think it should load in any information to start. This is irrelevant to your PR, just a mental note for me to go back and fix that

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,47 +1,121 @@
...
+    assert cdf_manager.global_attributes["Mission_group"] == "IMAP>Interstellar Mapping and Acceleration Probe"
+    assert cdf_manager.global_attributes["PI_name"] == "Dr. David J. McComas"
+    assert cdf_manager.global_attributes[
+               "PI_affiliation"] == "Princeton Plasma Physics Laboratory, 100 Stellarator Road, Princeton, NJ 08540"
+    assert cdf_manager.global_attributes["File_naming_convention"] == "source_descriptor_datatype_yyyyMMdd_vNNN"
 
-    cdf_manager.load_global_attributes("imap_default_global_cdf_attrs.yaml")
+    # Load in different data, test what was carried over
+    cdf_manager.load_global_attributes("imap_default_global_test_cdf_attrs.yaml")
+    assert cdf_manager.global_attributes["Project"] == "STP>Solar-Terrestrial Physics"
```
maxinelasp, line 89: Nice

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,47 +1,121 @@
...
+    assert cdf_manager.global_attributes["Source_name"] == "IMAP>Interstellar Mapping and Acceleration Probe"
+    assert cdf_manager.global_attributes["Discipline"] == "Solar Physics>Heliospheric Physics"
+    assert cdf_manager.global_attributes["Mission_group"] == "Dysfunctional Cats"
+    assert cdf_manager.global_attributes["PI_name"] == "Ana Manica"
+    assert cdf_manager.global_attributes["PI_affiliation"] == "LASP, CU"
+    assert cdf_manager.global_attributes["File_naming_convention"] == "source_descriptor_datatype_yyyyMMdd_vNNN"
+    assert cdf_manager.global_attributes["Data_version"] == 2
+    assert cdf_manager.global_attributes["Test_attribute"] == "test"
 
     # Expected failure: file_naming_convention is not in schema.
```
maxinelasp: This should no longer fail, so you can uncomment it

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,47 +1,121 @@
...
+    # Creating dictionary for loaded information
+    mag_test_global_attrs = cdf_manager.get_global_attributes("imap_test_T1_test")
+
+    assert mag_test_global_attrs["Mission_group"] == "Dysfunctional Cats"
+
+    # Testing if statement
+    assert mag_test_global_attrs["Descriptor"] == "TEST>Testinstrument"
+    assert mag_test_global_attrs["Logical_source"] == "imap_test_T1_test"
+    # Testing non required data from schema
+    assert mag_test_global_attrs["Data_type"] == "T1_test-one>Test-1 test one"
```
maxinelasp: Nice! I'd rename the variable away from "MAG" since it isn't using that attribute file anymore.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -56,3 +130,32 @@ def test_variable_attribute():
...
+    imap_test_variable = cdf_manager.get_variable_attributes("test_field_1")
+
+    # Make sure all expected attributes are here
+    for variable_attrs_2 in cdf_manager.variable_attribute_schema.keys():
+        required_var_attributes = cdf_manager.variable_attribute_schema[variable_attrs_2]
+        if required_var_attributes is True:
+            assert variable_attrs_2 in imap_test_variable.keys()
+
+    # Calling required attributes
+    # assert imap_test_variable["CATDESC"] == "test time"
```
maxinelasp: Why is this commented out?

imap_processing/cdf/tests/test_imap_cdf_manager.py
```diff
@@ -0,0 +1,49 @@
+# from pathlib import Path
+# from imap_processing.cdf.cdf_attribute_manager import CdfAttributeManager
+# This is a test
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
+
+# test add_instrument_global_attrs(self, instrument: str):
+
+
```
maxinelasp, line 6: Nice, I like this file, but don't add much more than this to it. ImapCdfAttributes is very minimal, and you don't want to duplicate tests. Really, the only test needed is that `self.source_dir` is set correctly, because right now that's the only thing done in ImapCdfAttributes, with everything else needing to be tested in `test_cdf_attribute_manager`

maxinelasp (commented):
Looks like a great first start! Not approving only because it's a draft rn, but it's looking really good!

imap_processing/cdf/tests/test_cdf_attribute_manager.py
anamanica: Ah! Okay, I see. This is the code that shows information is loaded in: 
`# Load Default IMAP Global Attributes
        self.global_attributes = CdfAttributeManager._load_yaml_data(
        )
        self.variable_attributes = dict()`

That is line 78 in cdf_attribute_manager.py, and DEFAULT_GLOBAL_CDF_ATTRS_FILE is defined as

`DEFAULT_GLOBAL_CDF_ATTRS_FILE = "imap_default_global_cdf_attrs.yaml"`

in line 12.

maxinelasp: Hmm ok in that case, leave this test in, but I probably will want to update that lol

maxinelasp: Should not block or really affect you

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -152,6 +156,35 @@ def test_global_attribute():
...
+    assert (
+        test_get_global_attrs_2["Logical_source_description"]
+        == "IMAP Mission TEST two document Level-T2."
+    )
+    assert test_get_global_attrs_2["LINK_TEXT"] == "Test two additional info"
+
+    # Trying to update a default global using get_global_attributes does not work.
+    # For example, thing about DOI event.
+
+    # END END END
```
maxinelasp: You can use things like `# ============` to split up sections. But, if you're splitting sections, you can also consider putting it into a different test. It's useful to have small tests because if `test_adding_second_global_attribute_file` fails you know exactly what happened, whereas `test_global_attribute` isn't as clear

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -164,9 +197,6 @@ def test_global_attribute():
         if required_schema is True:
             assert attr_name in test_get_global_attrs.keys()
 
-    # Testing second elif statement
-    # Should throw error
```
maxinelasp: If this is testing that the code successfully throws an error, you can use `pytest.raises`. If this test is failing because the code it's testing is incomplete or wrong, but it shouldn't, that's when you use `pytest.mark.xfail`. Basically, if someone should come back and fix it, `xfail` is your friend.

imap_processing/cdf/imap_cdf_manager.py
```diff
@@ -12,9 +12,13 @@
 class ImapCdfAttributes(CdfAttributeManager):
     """Contains IMAP specific tools and settings for CDF management."""
 
-    def __init__(self):
+    def __init__(self, source_dir_input=None):
```
bourque: Could you add a `Parameters` section to the docstring for the `source_dir_input`?

imap_processing/cdf/tests/imap_default_global_test_cdf_attrs.yaml
```diff
@@ -0,0 +1,7 @@
+Project: STP>Solar-Terrestrial Physics
+Source_name: IMAP>Interstellar Mapping and Acceleration Probe
+Mission_group: Dysfunctional Cats
+PI_name: Ana Manica
```
bourque, line 4: Nice 😂

imap_processing/cdf/imap_cdf_manager.py
```diff
@@ -12,9 +12,13 @@
...
         """Set the path to the config directory."""
-        super().__init__(Path(__file__).parent / "config")
+        if source_dir_input is None:
+            super().__init__(Path(__file__).parent / "config")
+        else:
+            super().__init__(Path(__file__).parent / "config")
```
maxinelasp: You should pass the source_dir_input into the super() init function because there are steps in `__init__` that depend on the source_dir and it would be confusing to have it set to different values for the init vs for the rest of processing. 

Then you can remove line 21

imap_processing/cdf/tests/imap_instrument2_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,24 @@
...
+    The IMAP Solar Wind Electron (SWE) instrument measures the solar wind electrons
+    in the heliosphere. SWE will contribute to our understanding of the acceleration
+    and transport of charged particles in the heliosphere.
+    SWE design and assembly is led by the Princeton Plasma Physics Laboratory. See
+    https://imap.princeton.edu/instruments/swe for more details.
+  Instrument_type: "Particles (space)"
+
+imap_swe_l1a_sci:
+  <<: *instrument_base
+  # NOTE: Right now, this Data_level is required to produce valid CDF
```
maxinelasp: This will always be required, so you can remove this note. It wasn't in the first draft of MAG because it was a first draft - but it is required by everyone.

imap_processing/cdf/tests/imap_instrument2_level2_variable_attrs.yaml
```diff
@@ -0,0 +1,310 @@
...
+
+vx_dps_sc:
+  <<: *default
+  CATDESC: Normalized component of the velocity vector in x direction in the DPS frame at rest WRT spacecraft (velocity w/ spacecraft frame subtracted)
+  FIELDNAM: vx_dps_sc
+  LABLAXIS: vx dps sc
+  # TODO: come back to format
+  UNITS: " "
+
+vy_dps_sc:
```
maxinelasp: This file should be shorter, with just 2 variables, rather than including all of these. For testing purposes, having 2 variables vs 20 variables doesn't add anything, and it makes the tests more complicated. Unless there is something specific you're testing with these that I'm missing?

imap_processing/cdf/tests/shared/default_global_cdf_attrs_schema_test.yaml
```diff
@@ -0,0 +1,246 @@
+DOI:
```
maxinelasp: Is there any way/is it reasonable to use the real schema? I know this goes back to the source_dir question again, but I think it would be valuable to test against the real schema - because if that ever changes, and the person updating doesn't realize that the file is duplicated in the tests, it could cause issues.

imap_processing/cdf/tests/shared/default_variable_cdf_attrs_schema_test.yaml
```diff
@@ -0,0 +1,466 @@
+# Based on :  https://spdf.gsfc.nasa.gov/istp_guide/vattributes.html
+#             HERMES CDF Format Guide
```
maxinelasp: Ditto for this schema, I would prefer to use the "real" one

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+
+    # Initialize CdfAttributeManager object which loads in default schema
+    cdf_manager = CdfAttributeManager(Path(__file__).parent.parent / "config")
+
+    # Default global tests
+    assert cdf_manager.global_attribute_schema["DOI"]["required"] is False
+    assert cdf_manager.global_attribute_schema["Data_level"]["required"] is True
+
+    # Default variable tests
+    assert (
```
maxinelasp, line 23: IMO it doesn't matter whether or not these variables are required or not - which is what you're testing here. Instead, you should check if the keys exist in the variable attribute schema, which means the schema correctly read from the file. 

So, these tests will check that the schema contains "attribute_key", "ABSOLUTE_ERROR," and "required,", so if that variable ended up being optional, the test won't start failing. You can also remove the assert on ln 29 because it's duplicating the check now.
```
assert "attribute_key" in cdf_manager.variable_attribute_schema
assert "ABSOLUTE_ERROR" in cdf_manager.variable_attribute_schema["attribute_key"]
assert "required" in cdf_manager.variable_attribute_schema["attribute_key"]["ABSOLUTE_ERROR"]
```

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+        ]
+        is False
+    )
+    assert (
+        cdf_manager.variable_attribute_schema["attribute_key"]["RESOLUTION"]["required"]
+        is True
+    )
+
+
 # @pytest.mark.xfail(reason="Missing IMAP specific global schema")
```
maxinelasp: Is this test expected to fail? If it runs now, just delete this line rather than commenting it out

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
 def test_global_attribute():
-    cdf_manager = CdfAttributeManager(Path(__file__).parent.parent / "config")
+    """
+    Test function that covers:
+        load_global_attributes
+    """
 
+    # Initialize CdfAttributeManager object which loads in default info
+    cdf_manager = CdfAttributeManager(Path(__file__).parent.parent / "config")
     cdf_manager.load_global_attributes("imap_default_global_cdf_attrs.yaml")
```
maxinelasp, line 43: This should probably be pointing to your test cdf attrs, since we won't be including any IMAP attrs when we move this to a new repo

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+        == "Solar Physics>Heliospheric Physics"
+    )
+
+    # Testing attributes newly loaded
+    assert cdf_manager.global_attributes["Project"] == "STP>Solar-Terrestrial Physics"
+    assert (
+        cdf_manager.global_attributes["Source_name"]
         == "IMAP>Interstellar Mapping and Acceleration Probe"
     )
+    assert cdf_manager.global_attributes["Mission_group"] == "Dysfunctional Cats"
```
maxinelasp, line 94: Ahh ok I see what you're testing here. I like that you're testing that the attributes are overwritten, I would include two test files (test_1 and test_2) and not only test that the attributes are overwritten (like you are here) and that you can add new attributes (like with DOI) but also that you are not required to overwrite attributes (so, for example, remove "Data_version" from your second attribute file and assert that it's the same as in the first attribute file.)

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+    assert cdf_manager.global_attributes["PI_name"] == "Ana Manica"
+    assert cdf_manager.global_attributes["PI_affiliation"] == "LASP, CU"
+    assert cdf_manager.global_attributes["Data_version"] == 2
+    assert cdf_manager.global_attributes["DOI"] == "test"
 
-    assert cdf_manager.global_attributes["Descriptor"] == "MAG>Magnetometer"
+    # Testing that everything loaded into the global attrs is present in
+    #   the global attrs schema
+    for attr in cdf_manager.global_attributes.keys():
+        assert attr in cdf_manager.global_attribute_schema.keys()
```
maxinelasp, line 103: Here, you're actually just testing your test files, which isn't particularly helpful. If you think about it, what lines of code is this testing? You've already tested that the global attributes are loaded in correctly. So I'd remove this assert.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+def test_get_global_attributes():
+    """
+    Test function that covers:
+        get_global_attributes
+    """
+    # Initialize CdfAttributeManager object which loads in default info
+    cdf_manager = CdfAttributeManager(Path(__file__).parent.parent / "config")
+
+    # Change filepath to load test global attributes
+    cdf_manager.source_dir = Path(__file__).parent.parent / "tests"
```
maxinelasp, line 116: Is this so you can get the schema from the "config" file? I like it, that means you're testing against the "real" schema.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+    cdf_manager.load_global_attributes("imap_default_global_test_cdf_attrs.yaml")
+    cdf_manager.load_global_attributes("imap_test_global.yaml")
```
maxinelasp, line 118: This is a nitpick, but I would prefer if  you could rename these files to make it immediately clear what the difference between these files is. Also, remove "IMAP" from the name - it's ok to have IMAP style names inside the files, but it should be clear that these are generic tests.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+    # Testing instrument specific global attributes (first elif statement)
+    assert test_get_global_attrs["Descriptor"] == "TEST>Testinstrument"
+    assert test_get_global_attrs["Data_type"] == "T1_test-one>Test-1 test one"
+    assert test_get_global_attrs["Logical_source"] == "imap_test_T1_test"
+    assert (
+        test_get_global_attrs["Logical_source_description"]
+        == "IMAP Mission TEST one document Level-T1."
+    )
+    # Not given, and not required information
+    assert test_get_global_attrs["Data_level"] is None
```
maxinelasp, line 139: Is data level not required in the schema?

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+    with pytest.raises(KeyError):
+        assert test_get_global_attrs["bad_name"] == "false info"
```
maxinelasp, line 141: You should also test that a global attribute in the file but not in the schema isn't passed along. So for example, add a new attribute `not_passed_along` inside imap_test_global.yaml and then assert that it isn't present in `test_get_global_attrs`

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+
+    # Testing second elif statement
+    test_error_elif = cdf_manager.get_global_attributes("imap_test_T3_test")
+    with pytest.raises(KeyError):
+        assert test_error_elif["Data_type"] == "Does Not Exist"
```
maxinelasp, line 146: You can also just write `assert test_error_elif["Data_type"]`

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+    test_get_global_attrs_2 = cdf_manager.get_global_attributes("imap_test_T2_test")
+    # Testing information previously loaded into global attributes (first if statement)
+    assert test_get_global_attrs_2["Project"] == "STP>Solar-Terrestrial Physics"
+    # Testing first elif statement
+    assert test_get_global_attrs_2["Descriptor"] == "TEST>Testinstrument"
+    # "Data_type" not required according to default schema
+    assert test_get_global_attrs_2["Data_type"] == "T2_test-two>Test-2 test two"
+
+    # Testing how instrument_id operates
+    assert test_get_global_attrs["Project"] == cdf_manager.global_attributes["Project"]
```
maxinelasp, line 177: are these meant to be `test_get_global_attrs_2`? If not, nitpick, but move them into the section with `test_get_global_attrs`

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+    assert (
+        test_get_global_attrs["Data_type"]
+        == cdf_manager.global_attributes["imap_test_T1_test"]["Data_type"]
+    )
+    assert (
+        cdf_manager.global_attributes["imap_test_T1_test"]["Logical_source"]
+        == "imap_test_T1_test"
+    )
+    with pytest.raises(KeyError):
+        assert cdf_manager.global_attributes["imap_test_T1_test"]["Project"]
```
maxinelasp, line 191: These tests are not testing `get_global_attributes` and should be removed (if they're duplicates, which they might be) or put into a different test.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+
+    # Trying to update a default global using get_global_attributes does not work.
+    # For example, thing about DOI event.
+
+    # Testing that required schema keys are in get_global_attributes
+    for attr_name in cdf_manager.global_attribute_schema.keys():
+        required_schema = cdf_manager.global_attribute_schema[attr_name]["required"]
+        if required_schema is True:
```
maxinelasp: Nice

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+        == cdf_manager.global_attributes["imap_test_T1_test"]["Data_type"]
+    )
+    assert (
+        cdf_manager.global_attributes["imap_test_T1_test"]["Logical_source"]
+        == "imap_test_T1_test"
+    )
+    with pytest.raises(KeyError):
+        assert cdf_manager.global_attributes["imap_test_T1_test"]["Project"]
+
+    # Trying to update a default global using get_global_attributes does not work.
```
maxinelasp: I don't really understand this comment... if its a note to self, it should be removed, or if it's a genuine note, could you expand on what you mean?

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
+    # Change filepath to load test global attributes
+    cdf_manager.source_dir = Path(__file__).parent.parent / "tests"
+    cdf_manager.load_global_attributes("imap_default_global_test_cdf_attrs.yaml")
+    cdf_manager.load_global_attributes("imap_test_global.yaml")
+
+    # Changing a dynamic global variable
+    cdf_manager.add_global_attribute("Project", "Test Project")
+    assert cdf_manager.global_attributes["Project"] == "Test Project"
+    test_get_global_attrs = cdf_manager.get_global_attributes("imap_test_T1_test")
+    assert test_get_global_attrs["Project"] == "Test Project"
```
maxinelasp, line 206: Nice test, I like it. Can you also add a test for adding a new attribute that didn't exist in the file? For example, delete one of the required attributes (you can delete it out of cdf_manager.global_attributes) and then assert that `get_global_attributes` fails for that attribute, then run `add_global_attribute` and assert that it now exists.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,322 @@
...
-        for attr in expected_attributes:
-            assert attr in variable_attrs.keys()
+    # Assuring all required attributes are loaded in
+    for attr_name in cdf_manager.variable_attribute_schema["attribute_key"]:
+        attribute = cdf_manager.variable_attribute_schema["attribute_key"][attr_name]
+        if attribute["required"] is True:
+            for exp_attr in expected_attributes:
+                assert (
+                    exp_attr
+                    in cdf_manager.variable_attribute_schema["attribute_key"].keys()
```
maxinelasp:
```suggestion
                    in cdf_manager.variable_attribute_schema["attribute_key"]
```

You don't need `.keys()`. I know this comes from my test but it's still true 😆 Greg always corrects me on this and it might have finally stuck now!

maxinelasp (commented):
Not done with my review yet - but nice work!

imap_processing/cdf/imap_cdf_manager.py
anamanica: This has actually been causing me a lot of weird struggles with testing this class. In CdfAttributeManager , the default schema and IMAPS default variables are in the "config" file path and not guaranteed in the  source_dir_input file path. I can't get the tests to run correctly without loading in that default info first, and then switching the file path... Thoughts?

imap_processing/cdf/tests/shared/default_global_cdf_attrs_schema_test.yaml
anamanica: Yes! Sorry, I actually am using the real schema, I should have deleted these files. This is why we have to initially set the CdfAttributeManager.source_dir to end in the "config" folder to get those real schema...

imap_processing/cdf/tests/test_cdf_attribute_manager.py
anamanica: Yeah! You will see the updated test in some of my new pushes, but I did have a few questions about this:
- The code in get_global_attributes does not throw an error if a required attribute is missing, it just sets that attribute to `None`. Shall this be changed in the method definition?
- I can't add attributes that are instrument specific because all instrument specific data is accessed through dictionary objects. and not CdfAttributeManager objects (such as `cdf_manager.global_attributes["instrument"]` or `instrument = cdf_manager.get_global_attributes("instrument)`). Is that okay?

imap_processing/cdf/imap_cdf_manager.py
maxinelasp: In those tests, you can pass in the "config" path (or pass in nothing and use the default) and then override it. This way it will still get the default schemas and global variables from "config" but you can also access your test files in the other file path.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
maxinelasp: 1. No, don't change the code, but instead validate that it's None.
2. You can add instrument specific attributes through `add_global_attribute` because the global attributes are flat and contain the global attributes for the instruments and for the shared values.

I'm not totally sure I understand your second question, did I answer it?

imap_processing/cdf/tests/test_imap_cdf_manager.py
```diff
@@ -0,0 +1,59 @@
...
+    # Testing reloading data
+    imap_cdf_manager.add_instrument_global_attrs("instrument2")
+
+    # Testing data carried over, and overwritten
+    instrument2_instrument = imap_cdf_manager.get_global_attributes("imap_swe_l1a_sci")
+    assert instrument2_instrument["Data_type"] == "L1A_SCI>Level-1A Science data"
+    assert instrument2_instrument["Project"] == "STP>Solar-Terrestrial Physics"
+
+
+def testing_source_dir():
```
maxinelasp, line 27: This test is sufficient for testing your "else" statement, so you can ignore codecov.

imap_processing/cdf/tests/test_imap_cdf_manager.py
```diff
@@ -0,0 +1,59 @@
...
+    # Create an ImapCdfAttributes object
+    imap_cdf_manager = ImapCdfAttributes("tests")
+    imap_cdf_manager.add_instrument_global_attrs("instrument")
+
+    # Testing data loaded in
+    imap_instrument = imap_cdf_manager.get_global_attributes("imap_test_T1_test")
+    assert imap_instrument["Data_type"] == "T1_test-one>Test-1 test one"
+    assert imap_instrument["Project"] == "STP>Solar-Terrestrial Physics"
+
+    # Testing reloading data
```
maxinelasp, line 18: This is already tested in the test_cdf_attribute_manager code - most of these tests are somewhat duplicating it. It's ok to have some duplicate tests, but generally, you should really try to focus in and make sure you only have one test per thing. If we changed the way we add global attrs, we would have to go and update 2 tests that test the same thing, so it's a little annoying. 

This isn't a blocker for me since this test file is pretty minimal - so I will leave it up to your judgement which tests are duplicating and which ones are adding value. Just wanted to point that out as something to think about.

maxinelasp (commented):
Finished review, ping me when you're ready for another one! Also, let me know if you want to meet to discuss any of this since there are some larger questions here that might be easier to have a conversation about.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
anamanica: I was being a silly goose, that's all.

imap_processing/cdf/imap_cdf_manager.py
anamanica: Hahaha, happy to say I just got it. Thank you so much! I see now.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
anamanica: Ah. I think I got it now, thanks!

imap_processing/cdf/tests/test_imap_cdf_manager.py
anamanica: The particular reason I included both of these tests is because they are loaded in from different files. That being said though, I definitely hear what you are saying. I'll rake through my tests, and get rid of the silly ones.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -203,19 +200,31 @@ def test_add_global_attribute():
...
 
     # Testing adding required global attribute
     cdf_manager.global_attributes.__delitem__("Source_name")
+    # Reloading get_global_attributes to pick up deleted Source_name
+    test_get_global_attrs = cdf_manager.get_global_attributes("imap_test_T1_test")
     with pytest.raises(KeyError):
-        assert cdf_manager.global_attributes["Source_name"] == "Wrong"
-    assert cdf_manager.global_attributes["Source_name"] is None
-
+        assert cdf_manager.global_attributes["Source_name"] == "Does_not_exist"
```
maxinelasp:
```suggestion
        assert cdf_manager.global_attributes["Source_name"] 
```

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -203,19 +200,31 @@ def test_add_global_attribute():
...
+    assert test_get_global_attrs["Source_name"] is None
+    # Adding deleted global attribute
     cdf_manager.add_global_attribute("Source_name", "anas_source")
     assert cdf_manager.global_attributes["Source_name"] == "anas_source"
```
maxinelasp, line 218: Nice, I'd also add:
```suggestion
    assert cdf_manager.global_attributes["Source_name"] == "anas_source"
    test_get_global_attrs = cdf_manager.get_global_attributes("imap_test_T1_test")
    assert test_get_global_attrs["Source_name"] == "anas_source"
```

but that is definitely a nitpick!

imap_processing/cdf/tests/test_imap_cdf_manager.py
```diff
@@ -1,11 +1,13 @@
...
 # from imap_processing.cdf.cdf_attribute_manager import CdfAttributeManager
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 
 
 def test_add_instrument_global_attrs():
-    # Create an ImapCdfAttributes object
-    imap_cdf_manager = ImapCdfAttributes("tests")
+    # Create an ImapCdfAttributes object, set to correct file path
+    imap_cdf_manager = ImapCdfAttributes()
+    imap_cdf_manager.source_dir = Path(__file__).parent.parent / "tests"
```
maxinelasp, line 10: Nice

imap_processing/cdf/tests/test_imap_cdf_manager.py
```diff
@@ -24,7 +26,8 @@ def test_add_instrument_global_attrs():
 
 def testing_source_dir():
     # Create an ImapCdfAttributes object
-    imap_cdf_manager = ImapCdfAttributes("tests")
+    imap_cdf_manager = ImapCdfAttributes()
+    imap_cdf_manager.source_dir = Path(__file__).parent.parent / "tests"
```
maxinelasp: You actually want to leave this step like you had it before - because you're testing that passing "tests" to the init results in source_dir getting set properly.

imap_processing/cdf/imap_cdf_manager.py
```diff
@@ -17,8 +17,7 @@ def __init__(self, source_dir=None):
         if source_dir is None:
             super().__init__(Path(__file__).parent / "config")
         else:
-            super().__init__(Path(__file__).parent / "config")
-            self.source_dir = Path(__file__).parent / source_dir
+            super().__init__(Path(__file__).parent / source_dir)
```
maxinelasp: Actually, could you change this to accept the whole path, rather than assuming the path is a parent to this file?

eg
```
def __init__(self, source_dir: Path=None):
  ...
  else:
    super().__init__(source_dir)
```

maxinelasp (approved):
Nice work! Really really close now, if you address my newest comments then you can go ahead and merge! Congrats for making it through the first review thunderdome 😁

imap_processing/cdf/imap_cdf_manager.py
```diff
@@ -11,11 +11,28 @@
...
+    """
+    Contains IMAP specific tools and settings for CDF management.
 
-    def __init__(self):
-        """Set the path to the config directory."""
-        super().__init__(Path(__file__).parent / "config")
+    Parameters
+    ----------
+    source_dir : Path
+            Source directory.
```
bourque:
```suggestion
    source_dir : Path
            Source directory.
```
```suggestion
    source_dir : Path
        Source directory.
```

Looks like this got double-indented

imap_processing/cdf/imap_cdf_manager.py
```diff
@@ -11,11 +11,28 @@
...
+    """
+
+    def __init__(self, source_dir=None):
+        """
+        Set the path to the config directory.
+
+        Parameters
+        ----------
+        source_dir
```
bourque:
```suggestion
        source_dir
```
```suggestion
        source_dir : Path or None
```

imap_processing/cdf/tests/imap_instrument_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,14 @@
+Descriptor: &desc
+  TEST>Testinstrument
+TEXT: &txt >
+  This file is nothing more than a test yamal. This description? This is a test description.
+  This Testinstrument will contribute to our understanding of the cdf_attribute_manager.py file.
+  This test file is led by me, Ana Manica. Lasp undergraduate employee.
+  I don't have a website. Sorry.
```
bourque: Beautifully written :joy:

imap_processing/cdf/tests/imap_instrument_global_cdf_attrs.yaml
bourque, line 1: Pretty nit-picky here, but I would prefer this file (and the other corresponding yaml files) to be named with `instrument1` instead of just `instrument` -- just to be consistent with the `insturment2` files in the way they are named.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,351 @@
...
     # All variables required to have:
     expected_attributes = [
+        "CATDESC",
+        "DEPEND_0",
         "DISPLAY_TYPE",
+        "FIELDNAM",
         "FILLVAL",
         "FORMAT",
+        "LABLAXIS",
+        "UNITS",
```
bourque, line 255: Yay for alphabetical order!

imap_processing/cdf/tests/test_imap_cdf_manager.py
```diff
@@ -0,0 +1,62 @@
+from pathlib import Path
+
+# from imap_processing.cdf.cdf_attribute_manager import CdfAttributeManager
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
+
+# Test
```
bourque:
```suggestion
```

bourque (approved):
Awesome work! I just have a few non-blocking nit-picky suggestions, if you so chose to accept them.

imap_processing/cdf/imap_cdf_manager.py
bourque: You can ignore this comment, it is outdated!

PR #638, opened 2024-06-17
Improvement to correct numpy update issue for Ultra

laspsandoval:
# Change Summary

## Overview
Improvement over initial correction to fix the datatype

## Updated Files
- pyproject.toml
- poetry.lock
- decom_tools.py : changed back to uint16 and instead put parenthesis so that there are no negative values for the pixel matrices


laspsandoval:
**Not needed to be merged before SIT-3.**

imap_processing/ultra/l0/decom_tools.py
```diff
@@ -182,9 +182,9 @@ def decompress_image(
     blocks_per_row = cols // pixels_per_block
 
     # Compressed pixel matrix
-    p = np.zeros((rows, cols), dtype=np.int16)
+    p = np.zeros((rows, cols), dtype=np.uint16)
```
greglucas, line 191: Reading the comments below I think that `p` should be `np.uint8` because you want to make this value into the range 0-255.
```suggestion
    p = np.zeros((rows, cols), dtype=np.uint8)
```

imap_processing/ultra/l0/decom_tools.py
```diff
@@ -214,8 +214,10 @@ def decompress_image(
                 # Keeps only the last 8 bits of the result of pixel0 - delta_f
                 # This operation ensures that the result is within the range
                 # of an 8-bit byte (0-255)
-                # TODO: Check to make certain no negative values occur here.
-                p[i][column_index] = (pixel0 - delta_f) & 0xFF
+                if delta_f < 0:
+                    p[i][column_index] = (pixel0 + (-delta_f)) & 0xFF
+                else:
+                    p[i][column_index] = (pixel0 - delta_f) & 0xFF
```
greglucas: I'm not sure what we are trying to do here so I am a bit concerned with the overflow checks that are happening and whether this is what we want or not. These should be identical results unless we are going outside of the range of the 0-255 values. Do we always want to go towards 128, or do we want to allow wraparound?

Lets say pixel0 is 0 and delta_f is -1. Do you want that to produce 255 or 1?

```bash
python -c "import numpy; print(numpy.uint8(0) - 1)"
RuntimeWarning: overflow encountered in scalar subtract
255
```

laspsandoval: This is what I want:

delta_f
-1
(pixel0 + (-delta_f)) & 0xFF
np.uint16(1)

delta_f
-2
(pixel0 + (-delta_f)) & 0xFF
np.uint16(2)

laspsandoval: Is we use the same equation this happens:

pixel0
np.uint16(1)
delta_f
1
(pixel0 - delta_f) & 0xFF
np.uint16(0)
(pixel0 + (-delta_f)) & 0xFF
...
OverflowError: Python integer -1 out of bounds for uint16

greglucas: If you use triple backticks you can get inline code formatting, which might make it easier to see the commands you want to show.

So you are OK with OverflowWarnings still and want that to happen? In my example above, we have `pixel0=0, delta_f=1` and the result is `255` with an Overflow warning. You are just trying to avoid the overflow **error** here because you can't do `np.uint8(-1)` as a cast.

I'm still not totally following all of this and what data types each of these values can be. Reading the above comment it says "we are masking it with `0xFF` to ensure it is only a single byte. But does that mean you want to expand to 16-bit space just for the math and then truncate the value to 8-bit space. Or do you really want to do all the math with overflows in 8-bit space?

If you want to go to larger space and then back down again, you might be able to do an explicit cast of the pixel0 value, work in int16 space and then an implicit cast back to uint8 space because you are setting the array value (which does 0xFF for you) `p[i][column_index] = np.int16(pixel0) + delta_f`, avoiding the if/else clause.


There are a lot of files here now. Could you bring it back to just the numpy update?

laspsandoval: Yes sorry about the extra files.

laspsandoval: Ok I made an update that uses int16 for the math and then brings it back to uint8.

imap_processing/ultra/l0/decom_tools.py
```diff
@@ -220,8 +220,9 @@ def decompress_image(
                 # Keeps only the last 8 bits of the result of pixel0 - delta_f
                 # This operation ensures that the result is within the range
                 # of an 8-bit byte (0-255)
-                # TODO: Check to make certain no negative values occur here.
-                p[i][column_index] = (pixel0 - delta_f) & 0xFF
+                # Use np.uint16 for the arithmetic operation to avoid overflow
+                result = np.int16(pixel0) + (-delta_f)
+                p[i][column_index] = np.uint8(result)
```
greglucas: If we want `p` to be uint8, I still think you should set it to uint8 dtype above these loops. Then here:

```suggestion
                # Use np.uint16 for the arithmetic operation to avoid overflow
                # Then implicitly cast back to the p's uint8 dtype for storage
                p[i][column_index] = np.int16(pixel0) - delta_f
```

laspsandoval: The max value is > 255. I see one that is 368 in the test data. I'm therefore setting p to uint16.

greglucas: The comment above should be updated then?
```
                # Keeps only the last 8 bits of the result of pixel0 - delta_f
                # This operation ensures that the result is within the range
                # of an 8-bit byte (0-255)
```

But, if I look at the algorithm document (which I think is this image):
image

That to me indicates that `p` should be in the range 0-255, if you are masking with 0xff there is no way to get a larger value. After log decompression, then I agree that things will get larger and `pp` or `p_decom` here might need to be a different range. But I'm still not seeing it for `p` itself.

greglucas (approved):
Hesitantly approve. I'm still not convinced of the datatypes here, but it could just be me being pedantic or not understanding and there is no issue with using the larger types.

laspsandoval:
Will confirm datatype with Ultra team.

PR #653, opened 2024-06-20
ENH/MNT: Allow array-like time conversion

greglucas:
# Change Summary

We can convert entire arrays at once to nanoseconds since J2000 from integer MET rather than having to loop through in each individual calling location.

This is faster and easier to read IMO.

I changed the name to `met_to_j2000ns`.

## Files changed

I updated many files to use the array indexing and removed the unnecessary for-loops.

### DIFF
--- a/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
@@ -70,7 +70,7 @@ default_epoch: &default_epoch
   TIME_BASE: J2000
   TIME_SCALE: Terrestrial Time
   REFERENCE_POSITION: Rotating Earth Geoid
-  dtype: datetime64[ns]
+  dtype: int64
 
 # ------- L1A DE Section -------
 hi_de_ccsds_met:
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -42,7 +42,9 @@ def convert_met_to_datetime64(
     """
     # Convert MET to int, then to nanoseconds as datetime64 requires int input and
     # some instruments are adding additional float precision to the MET
-    time_array = (np.asarray(met) * 1e9).astype(int).astype("timedelta64[ns]")
+    # NOTE: We need int64 here when running on 32bit systems as plain int will default
+    #       to 32bit and overflow due to the nanosecond multiplication
+    time_array = (np.asarray(met) * 1e9).astype(np.int64).astype("timedelta64[ns]")
     return launch_time + time_array
 
 
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -16,36 +16,49 @@ import imap_processing
 logger = logging.getLogger(__name__)
 
 
-def convert_met_to_datetime64(
+# Reference start time (launch time or epoch)
+# DEFAULT_EPOCH = np.datetime64("2010-01-01T00:01:06.184", "ns")
+IMAP_EPOCH = np.datetime64("2010-01-01T00:00:00", "ns")
+J2000_EPOCH = np.datetime64("2000-01-01T11:58:55.816", "ns")
+
+
+def met_to_j2000ns(
     met: np.typing.ArrayLike,
-    launch_time: Optional[np.datetime64] = imap_processing.launch_time,
-) -> np.datetime64:
-    """Convert mission elapsed time (MET) to numpy datetime64.
+    reference_epoch: Optional[np.datetime64] = IMAP_EPOCH,
+) -> np.typing.ArrayLike:
+    """
+    Convert mission elapsed time (MET) to nanoseconds from J2000.
 
     Parameters
     ----------
-    met: array_like
-        Number of seconds since epoch (nominally the launch time)
-    launch_time : np.datetime64
-        The time of launch to use as the baseline.
+    met : array_like
+        Number of seconds since epoch according to the spacecraft clock.
+    reference_epoch : np.datetime64
+        The time of reference for the mission elapsed time. The standard
+        reference time for IMAP is January 1, 2010 00:00:00 UTC. Per APL's
+        IMAP Timekeeping System Design document.
 
     Returns
     -------
-    np.datetime64
-        Numpy datetime64 representation of the mission elapsed time
+    array_like or scalar, int64
+        The mission elapsed time converted to nanoseconds since the J2000 epoch.
 
+    Notes
     -----
-    This conversion is temporary for now, and will need SPICE in the future.
-    Nick Dutton mentioned that s/c clock start epoch is
-        jan-1-2010-00:01:06.184 ET
-    We will use this for now.
+    This conversion is temporary for now, and will need SPICE in the future to
+    account for spacecraft clock drift.
     """
-    # Convert MET to int, then to nanoseconds as datetime64 requires int input and
-    # some instruments are adding additional float precision to the MET
+    # Mission elapsed time is in seconds, convert to nanoseconds
+    # NOTE: We need to multiply the incoming met by 1e9 first because we could have
+    #       float input and we want to keep ns precision in those floats
     # NOTE: We need int64 here when running on 32bit systems as plain int will default
     #       to 32bit and overflow due to the nanosecond multiplication
-    time_array = (np.asarray(met) * 1e9).astype(np.int64).astype("timedelta64[ns]")
-    return launch_time + time_array
+    time_array = (np.asarray(met) * 1e9).astype(np.int64)
+    # Calculate the time difference between our reference system and J2000
+    j2000_offset = (
+        (reference_epoch - J2000_EPOCH).astype("timedelta64[ns]").astype(np.int64)
+    )
+    return j2000_offset + time_array
 
 
 def load_cdf(
@@ -61,19 +74,14 @@ def load_cdf(
     remove_xarray_attrs : bool
         Whether to remove the xarray attributes that get injected by the
         cdf_to_xarray function from the output xarray.Dataset. Default is True.
-    **kwargs : {dict} optional
-        Keyword arguments for ``cdf_to_xarray``. This function overrides the
-        ``cdf_to_xarray`` default keyword value `to_datetime=False` with
-        ``to_datetime=True`.
+    **kwargs : dict, optional
+        Keyword arguments for ``cdf_to_xarray``.
 
     Returns
     -------
     dataset : xarray.Dataset
         The ``xarray`` dataset for the CDF file.
     """
-    # TODO: remove this when cdflib is updated to version >1.3.0
-    if "to_datetime" not in kwargs:
-        kwargs["to_datetime"] = True
     dataset = cdf_to_xarray(file_path, kwargs)
 
     # cdf_to_xarray converts single-value attributes to lists
@@ -118,9 +126,9 @@ def write_cdf(dataset: xr.Dataset):
     # Create the filename from the global attributes
     # Logical_source looks like "imap_swe_l2_counts-1min"
     instrument, data_level, descriptor = dataset.attrs["Logical_source"].split("_")[1:]
-    start_time = np.datetime_as_string(dataset["epoch"].values[0], unit="D").replace(
-        "-", ""
-    )
+    # Convert J2000 epoch referenced data to datetime64
+    dt64 = J2000_EPOCH + dataset["epoch"].values[0].astype("timedelta64[ns]")
+    start_time = np.datetime_as_string(dt64, unit="D").replace("-", "")
 
     # Will now accept vXXX or XXX formats, as batch starter sends versions as vXXX.
     r = re.compile(r"v\d{3}")
@@ -156,7 +164,6 @@ def write_cdf(dataset: xr.Dataset):
     xarray_to_cdf(
         dataset,
         str(file_path),
-        datetime64_to_cdftt2000=True,
         terminate_on_warning=True,
     )  # Terminate if not ISTP compliant
 
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -53,7 +53,7 @@ def met_to_j2000ns(
     #       float input and we want to keep ns precision in those floats
     # NOTE: We need int64 here when running on 32bit systems as plain int will default
     #       to 32bit and overflow due to the nanosecond multiplication
-    time_array = (np.asarray(met) * 1e9).astype(np.int64)
+    time_array = (np.asarray(met, dtype=float) * 1e9).astype(np.int64)
     # Calculate the time difference between our reference system and J2000
     j2000_offset = (
         (reference_epoch - J2000_EPOCH).astype("timedelta64[ns]").astype(np.int64)
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -16,39 +16,34 @@ import imap_processing
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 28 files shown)


greglucas:
This isn't reassuring that changing the time handling in numpy would cause windows tests to fail! (All linux and mac runners successful) I thought windows only hated me when I was dealing with file paths 😂 

If anyone has thoughts on what could be causing this I'm all ears since I don't have a windows machine to test with.

Edit: Apparently Windows runners are 32bit machines and therefore `int` gets interpreted as a numpy `int32` datatype, so we need to explicitly set it to `int64`.

imap_processing/cdf/utils.py
```diff
@@ -16,37 +16,36 @@
...
-    """Calculate the datetime64 from the CCSDS secondary header information.
-
-    Since all instrument has SHCOARSE or MET seconds, we need convert it to
-    UTC. Took this from IDEX code.
+    """Convert mission elapsed time (MET) to numpy datetime64.
 
     Parameters
     ----------
-    shcoarse_time: float
+    met: array_like
```
tech3371: I am not familiar with `array_like` type. Does this mean that input can be array or one input?

imap_processing/cdf/utils.py
```diff
@@ -16,37 +16,36 @@
...
     -----
     This conversion is temporary for now, and will need SPICE in the future.
     Nick Dutton mentioned that s/c clock start epoch is
         jan-1-2010-00:01:06.184 ET
     We will use this for now.
     """
-    # Get the datetime of Jan 1 2010 as the start date
-    time_delta = np.timedelta64(int(shcoarse_time * 1e9), "ns")
-    return launch_time + time_delta
+    # Convert MET to int, then to nanoseconds as datetime64 requires int input and
```
tech3371: Thank you for refactoring this code!

tech3371 (approved):
looks good to me!

imap_processing/cdf/utils.py
greglucas: Yes, needs to be usable by `np.asarray`. Maybe @daralynnrhode knows if there is a better type annotation to put here for numpydoc style?

imap_processing/cdf/utils.py
```diff
@@ -16,37 +16,36 @@
 logger = logging.getLogger(__name__)
 
 
-def calc_start_time(
-    shcoarse_time: float,
+def convert_met_to_datetime64(
```
bourque: I don't have a strong opinion on this, but wondering if we should instead call this `get_met()` or `convert_to_mission_met()` or something similar? Because we are not simply converting times here, we are returning times that are converted _and_ are now relative to the `launch_time`.

bourque (approved):
I wouldn't mind seeing the function name change, but I don't really have any great suggestions for that, nor feel that strong about it. Otherwise this LGTM!

imap_processing/cdf/utils.py
greglucas: I think that means we need a new name then if you're confused by the current naming! Just to note though we are not getting MET or converting to MET. That is the input here, and then we are converting to a datetime64 based on whenever launch was as you say. I'm trying to get across we are going from MET to earth UTC times.

tmplummer: To be somewhat consistent with SPICE, this could be named `met_to_et_ns` or `met_to_j2000ns`. Ephemeris Time in spice is J2000 seconds, not j2000 nanoseconds. But in either case, this function should not be confused with a conversion to UTC.

greglucas: But we aren't doing either of those, we are converting to UTC numpy datetime64s... Do we need to wait until SPICE gets involved and just remove this function then?

tmplummer: Aha, I hadn't appreciated that `datetime64` is referenced to the epoch `1970-01-01T00:00:00`.

Ultimately, yes SPICE will need to be used to convert from SHCOARSE to J2000 seconds (ET). ET is the time that SPICE requires throughout as an input to functions used to compute geometry.

ET will get converted to J2000 nanoseconds to store in CDF epoch coordinates.

I have not looked at what cdflib does to convert `datetime64` to ISTP compliant epoch in the CDF, but that is another time format translation that we need to be cognizant of. It may be that we want to skip ever translating to `datetime64` and just go from ET directly to J2000 nanoseconds for the CDF files.

tmplummer (approved):
To me, this PR looks good. The `convert_met_to_datetime64` function may become unnecessary after spice integration, but I don't see a reason not to make these changes since the work has already been done.

greglucas:
> To me, this PR looks good. The convert_met_to_datetime64 function may become unnecessary after spice integration, but I don't see a reason not to make these changes since the work has already been done.

@subagonsouth, do you have an idea of what we will be using once we get SPICE integration? I'm not in a rush to get this in if it isn't helpful.

I think some ENA teams have wanted to stay in MET for as long as possible. So, are we actually doing ourselves a disservice by going to datetimes too soon. I'm curious if we should actually be sticking with EPOCH being defined as integer/float seconds since launch in all of our calculations/processing. Then only do a datetime/CDF_EPOCH conversion in the `write_cdf()` routine (could reverse that in `load_cdf()` if we want).

tmplummer:
@greglucas 
> do you have an idea of what we will be using once we get SPICE integration? I'm not in a rush to get this in if it isn't helpful.

Here is what I know:

- SPICE will be needed to convert from SHCOARSE to seconds since J2000. SPICE calls this Ephemeris Time (ET) aka Barycentric Dynamical Time (TDB).  https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/req/time.html#The%20J2000%20Epoch
- It is convenient that SPICE and SPDF need J2000 epoch times in the `epoch` CDF coordinate.
- Paul Janzen has conveyed the need to have MET in Hi files and thus I have designed the products to carry both J2000 nanoseconds and double precision MET seconds into L1B DE products.

What I am unsure of:
- I'm a little unsure of the strict definition of Mission Elapsed Time. In my mind, MET is different than what comes down in the SHCOARSE telemetry. The reason that I say this is that spacecraft clocks drift and can be reset. This is where the sclk kernel comes in. It tells SPICE about clock drift or resets allowing SPICE to correct for this when converting the spacecraft clock to ET.
- I think, unless there is a compelling reason, that we will not use `datetime64` in the processing pipeline once we have SPICE time utilities integrated.

greglucas:
I think we are on the same page @subagonsouth, and I am also confused by the strict terminology.

> I'm a little unsure of the strict definition of Mission Elapsed Time. In my mind, MET is different than what comes down in the SHCOARSE telemetry. The reason that I say this is that spacecraft clocks drift and can be reset. This is where the sclk kernel comes in. It tells SPICE about clock drift or resets allowing SPICE to correct for this when converting the spacecraft clock to ET.

I think this is where I was just assuming a different convention than you here, so it is really good to be discussing this! I was assuming MET was the spacecraft clock signal (SHCOARSE) and it was allowed to drift and then we would correct that later via SPICE as you mentioned. It sounds like maybe we shouldn't use "MET" as it is apparently deprecated

https://cwe.ccsds.org/sea/docs/SEA-TIME/Draft%20Materials/Green%20Book%20draft/References%20in%20Green%20Book/08863870.pdf

> In the past, this SCLK time has sometimes been referred to as Mission Elapsed Time (MET).  It was assumed that the spacecraft clock time zero would be the moment of launch or close to it. This is often no longer the case, so the term “MET” is now deprecated for time correlation

Some other older reading with more clock details: https://ntrs.nasa.gov/api/citations/20030025281/downloads/20030025281.pdf

> I think, unless there is a compelling reason, that we will not use datetime64 in the processing pipeline once we have SPICE time utilities integrated.

Perhaps we should change this value to nanoseconds since J2000 now? New function called `convert_sclk_to_j2000()` which would change the implementation to something like this:
```python
nanosecond_sclk = (np.asarray(sclk) * 1e9).astype(np.int64)
nanosecond_offset = (launch_time - np.datetime64("2000-01-01T12:00").astype("timedelta64[ns]").astype(np.int64)
return nanosecond_sclk + nanosecond_offset
```

tmplummer:
Thanks for linking those documents @greglucas . I looked more closely at the Hi algorithm document and found the use of MET in the document to be consistent with MET being equivalent to spacecraft clock time. I do think that MET is a misleading name but suspect that it may be difficult to get the science team to want to change that name.

> Perhaps we should change this value to nanoseconds since J2000 now? New function called `convert_sclk_to_j2000()` which would change the implementation to something like this:
> 
> ```python
> nanosecond_sclk = (np.asarray(sclk) * 1e9).astype(np.int64)
> nanosecond_offset = (launch_time - np.datetime64("2000-01-01T12:00").astype("timedelta64[ns]").astype(np.int64)
> return nanosecond_sclk + nanosecond_offset
> ```
This seems like a good idea to me. Please note that the J2000 epoch is actually referenced to `"2000-01-01T11:58:55.816"`.

greglucas:
I've now updated this after our group discussion to provide a conversion utility called `met_to_j2000ns` taking Tim's suggestion. I've updated everywhere in the codebase to use this utility function for the `epoch` variable. Please take a look at the latest commit, there was a lot more conflicts and updates needed than I was hoping for, so it'd be good for others to double check that I got the intention right for all the places in the code.

imap_processing/cdf/utils.py
```diff
@@ -16,39 +16,49 @@
...
     """
-    # Get the datetime of Jan 1 2010 as the start date
-    time_delta = np.timedelta64(int(shcoarse_time * 1e9), "ns")
-    return launch_time + time_delta
+    # Mission elapsed time is in seconds, convert to nanoseconds
+    # NOTE: We need to multiply the incoming met by 1e9 first because we could have
+    #       float input and we want to keep ns precision in those floats
+    # NOTE: We need int64 here when running on 32bit systems as plain int will default
+    #       to 32bit and overflow due to the nanosecond multiplication
+    time_array = (np.asarray(met) * 1e9).astype(np.int64)
```
tmplummer: It appears that the multiplication will overflow if `met` is passed in as a 32 bit integer (signed or unsigned). I see that SHCOARSE is a `uint32` in packets. Doing the math on seconds since the IMAP_EPOCH:
```
# 15 years * 365.25 days/year * 24 hours/day * 60 minutes/hour * 60 seconds/minute * 1e9 ns/second
np.log2(15 * 365.25 * 24 * 60 * 60 * 1e9) = 58.7
```

imap_processing/cdf/config/imap_hi_variable_attrs.yaml
```diff
@@ -70,7 +70,7 @@ default_epoch: &default_epoch
   TIME_BASE: J2000
   TIME_SCALE: Terrestrial Time
   REFERENCE_POSITION: Rotating Earth Geoid
-  dtype: datetime64[ns]
+  dtype: int64
```
tmplummer, line 73: I confirmed that the CDF_TIME_TT2000 data type is an `int64`, not a `uint64`.

imap_processing/cdf/utils.py
```diff
@@ -159,7 +164,6 @@ def write_cdf(dataset: xr.Dataset):
     xarray_to_cdf(
         dataset,
         str(file_path),
-        datetime64_to_cdftt2000=True,
```
tmplummer, line 162: Does the `int64` epoch get written correctly in the CDF as datatype: CDF_TIME_TT2000?

tmplummer (commented):
Overall, this looks great. Two questions/concerns that I would like double checked.
Thanks!

imap_processing/cdf/utils.py
greglucas: Very good catch! I was getting lucky before as numpy did the cast to float automatically for us.
```
np.asarray(np.int32(2**30)) -> np.int32
np.asarray(np.int32(2**30)) * 1e9 -> np.float
```
I believe that is changing in numpy2.0 so I've updated it to be explicit by adding in `np.asarray(met, dtype=float)`.

imap_processing/cdf/utils.py
greglucas: Apparently so... I just dumped our test dataset out to disk and ran `cdfdump` on it. One concerning thing is that our MET times in that test file are 1, 2, 3. But we are getting
```
  Record # 1: 2009-12-31T23:59:59.000000000
  Record # 2: 2010-01-01T00:00:00.000000000
  Record # 3: 2010-01-01T00:00:01.000000000
```
But our IMAP_EPOCH is set to 2010-01-01T00:00, so those seem off to me. Are we missing a leapseconds conversion somewhere / am I missing something obvious?

```
Variable Information (0 rVariable, 1 zVariable)
===========================================================
epoch                 CDF_TT2000/1      0:[]    T/  


Variable (1 variable)
=========================================

epoch (No: 0) (Recs: 3)
-----
Data Type:           CDF_TT2000
Dimensionality:      0:[]       (T/)  
Format:              
Pad value:           0000-01-01T00:00:00.000000000
Written Records:     3/3(max)
Allocated Records:   3/3(max)
Blocking Factor:     1 (records)
Attribute Entries:
     CATDESC         (CDF_CHAR/66): "Time, number of nanoseconds since J2000 with leap seconds included"
     FIELDNAM        (CDF_CHAR/5): "epoch"
     FILLVAL         (CDF_TT2000/1): 9999-12-31T23:59:59.999999999
     LABLAXIS        (CDF_CHAR/5): "epoch"
     FORMAT          (CDF_CHAR/1): ""
     UNITS           (CDF_CHAR/2): "ns"
     VALIDMIN        (CDF_TT2000/1): 1990-01-01T00:00:00.000000000
     VALIDMAX        (CDF_TT2000/1): 2099-12-31T00:00:00.000000000
     VAR_TYPE        (CDF_CHAR/12): "support_data"
     SCALETYP        (CDF_CHAR/6): "linear"
     MONOTON         (CDF_CHAR/8): "INCREASE"
     TIME_BASE       (CDF_CHAR/5): "J2000"
     TIME_SCALE      (CDF_CHAR/16): "Terrestrial Time"
     REFERENCE_POSITION (CDF_CHAR/20): "Rotating Earth Geoid"
     DEPEND_0        (CDF_CHAR/5): "epoch"
Variable Data:
  Record # 1: 2009-12-31T23:59:59.000000000
  Record # 2: 2010-01-01T00:00:00.000000000
  Record # 3: 2010-01-01T00:00:01.000000000
```

tmplummer: I believe that you are on the right track asking about leapseconds. My reading of the numpy documentation is that conversion to/from a string is for TAI time, not UTC. See: https://numpy.org/doc/stable/reference/arrays.datetime.html#datetime64-conventions-and-assumptions

Comparing to SPICE, this is what I get:
```
from emmemuspy.utils.time import utc_2_et
IMAP_EPOCH = np.datetime64("2010-01-01T00:00:00", "ns")
J2000_EPOCH = np.datetime64("2000-01-01T11:58:55.816", "ns")
>>> (IMAP_EPOCH - J2000_EPOCH).astype("timedelta64[ns]").astype(np.int64) / 1e9 - utc_2_et("2010-01-01T00:00:00")
-1.9999244809150696
```
This is consistent with the two leapseconds that occured between 2000 and 2010 (one in 2005 and one in 2008).

tmplummer (approved):
This turned into much more work than intended for a temporary function. Nice job!

greglucas:
> This turned into much more work than intended for a temporary function.

That is the way it goes sometimes. @subagonsouth, thanks for the thorough review! It got everyone to think about how we are handling time and come to an agreement for a common path forward I think. Hopefully this helps with the transition over to SPICE for everyone as well.

PR #654, opened 2024-06-20
Xtce combined

greglucas:
# Change Summary

## Overview

I am opening this up to get discussion and see what people think of any of these items I list below. There is a lot going on here, but I tried to break it up into manageable commits and we could easily pick/choose which pieces we do/don't want. My goal here was to see if it'd be "easy" to automate the housekeeping data products, potentially through XTCE calibrators.

## Updates

* Rework and try to simplify some logic in the XTCEgenerator to go through all of the sheets of the Excel files we receive. This creates the combined XTCE file for us without having to give packet apids etc...
* Adds a polynomial calibrator to the parameter types based on the conversion tab of the spreadsheet. We are currently bringing that in manually as a csv I think and using it here: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/a62122a1fa723b20871c45c0b1023454a32cb036/imap_processing/utils.py#L59
  *  There could be downsides to this as I'm guessing it is faster to just do the conversion in numpy as a vectorized process rather than evaluating the polynomial every time with `space_packet_parser`, so maybe it is actually better to leave these conversions out?
  * It is unclear to me whether XTCE can support SEGMENTEDPOLY calibrations and how to handle some of the low/high value stuff with XTCE. That may also be easier to just do in straight numpy math rather than XML-like math.
* Added logic for variable length BYTES fields. It might be nice to calculate these automatically, but I don't know if the spreadsheets give us enough logic to know whether the BYTE field is fixed-length or variable-length. They all seem to be the maximum fixed length possible.
* Added a function to automatically process housekeeping and write a cdf file out. This is very similar to the `create_dataset` https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/a62122a1fa723b20871c45c0b1023454a32cb036/imap_processing/utils.py#L136 function, so this logic should be merged with that probably, but it was easier for me to test in a standalone script.


tools/xtce_generation/decode_hk.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    # Set up containers to store our data
+    packet_dict = defaultdict(list)
+    metadata_dict = defaultdict(dict)
+
+    # Set up the parser from the input packet definition
+    packet_definition = xtcedef.XtcePacketDefinition(xtce_packet_definition)
+    packet_parser = parser.PacketParser(packet_definition)
+
+    with open(packet_file, "rb") as binary_data:
```
tech3371: with how this is setup, I feel like it won't work if we get packet file with multiple apid. I feel like this assume this packet contains only housekeeping data. Did I understand it right?

tools/xtce_generation/decode_hk.py
```diff
@@ -0,0 +1,74 @@
...
+        packet_generator = packet_parser.generator(binary_data)
+        for i, packet in enumerate(packet_generator):
+            if i == 0:
+                # Time is always the first data key on IMAP
+                time_key = next(iter(packet.data.keys()))
+
+            for key, value in (packet.header | packet.data).items():
+                # TODO: Do we want derived_value or raw_value?
+                # Maybe we want these to be l1a and l1b files?
```
tech3371: There was discussion in instrument nomenclature about storing conversion formula in global attribute or some variable. I wonder if we can also do that in the future for this question and store conversion formula that user can use to convert raw to EU.

tools/xtce_generation/decode_hk.py
```diff
@@ -0,0 +1,74 @@
...
+                    metadata_dict[key]["short_description"] = value.short_description
+                    metadata_dict[key]["long_description"] = value.long_description
+                    # Units too?
+
+    # Convert the data to an xarray dataset
+    time_coord = packet_dict.pop(time_key)
+    ds = xr.Dataset(
+        {key: (time_key, val) for key, val in packet_dict.items()},
+        coords={time_key: time_coord},
+    )
```
tech3371: Is this line creating data array for all `packet_dict` data? Will this be updated to use `metadata_dict` as well to set CDF attrs or are we leaving it out since housekeeping will only be process up to level 1 and SPDF said it's ok to be bare minimum?

tools/xtce_generation/telemetry_generator.py
```diff
@@ -26,298 +64,280 @@ class TelemetryGenerator:
...
-        root.attrib["xmlns:xtce"] = f"{self.source_link}"
-        root.attrib["name"] = self.packet_name
-        # xmlns:xtce="http://www.omg.org/space/xtce"
+        self._root = root
+        root.attrib["xmlns:xtce"] = source_link
+        # Subsystem sheet name is used as the base name for this XTCE definition
+        subsystem = self.sheets["Subsystem"]
+        root.attrib["name"] = subsystem.loc[
+            subsystem["infoField"] == "subsystem", "infoValue"
+        ].values[0]
```
tech3371: cool

tools/xtce_generation/telemetry_generator.py
```diff
@@ -26,298 +64,280 @@ class TelemetryGenerator:
...
-        header.attrib["date"] = current_date
-        header.attrib["version"] = "1.0"
+        header.attrib["date"] = str(
+            subsystem.loc[
+                subsystem["infoField"] == "sheetReleaseDate", "infoValue"
+            ].values[0]
+        )
+        header.attrib["version"] = subsystem.loc[
+            subsystem["infoField"] == "sheetReleaseRev", "infoValue"
+        ].values[0]
```
tech3371: nice. with this, we get true date of when packet definition was updated last according to instrument team.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -26,298 +64,280 @@ class TelemetryGenerator:
...
+
+        # Create the XML containers that will be populated later
+        self._setup_xml_containers()
+        # Add the CCSDS Header information to the containers
+        self._setup_ccsds_header()
+        # Create the sequence containers (also adding parameters within)
+        self._create_container_sets()
+
+    def _setup_xml_containers(self):
+        """Create an XML representation of telemetry data."""
```
tech3371: can you add example of what we are creating here in doc string as preview of what's been done here? I think I follow it right now by looking at existing xtce file but would be nice to see it here as well. Eg. seems like we are creating main structure of xtce here.

```
<xtce:SpaceSystem xmlns:xtce="http://www.omg.org/space/xtce" name="H45_SCI_DE">
	<xtce:Header date="2023-12" version="1.0" author="IMAP SDC" />
	<xtce:TelemetryMetaData>
		<xtce:ParameterTypeSet>
		</xtce:ParameterTypeSet>
		<xtce:ParameterSet>
		</xtce:ParameterSet>
		<xtce:ContainerSet>
		</xtce:ContainerSet>
	</xtce:TelemetryMetaData>
</xtce:SpaceSystem>
```

tools/xtce_generation/telemetry_generator.py
```diff
@@ -26,298 +64,280 @@ class TelemetryGenerator:
...
-        Returns:
-        - parameter_set: The updated ParameterSet element.
         """
-        # Process rows from SHCOARSE until the last available row in the DataFrame
-        for index, row in self.pkt.iterrows():
-            if index < 7:
-                # Skip rows until SHCOARSE(also known as MET) which are
-                # not part of CCSDS header
-                continue
+        name = f"{row["packetName"]}_{row["mnemonic"]}"
```
tech3371: With this, we are going to have many `xtce:IntegerParameterType` in`xtce:ParameterTypeSet` if we are using this naming convention. is that what you intended? It seems like we are going to have one for each `mnemonic`.

tools/xtce_generation/telemetry_generator.py
```diff
@@ -26,298 +64,280 @@ class TelemetryGenerator:
...
+                param_ref = Et.SubElement(dynamic_value, "xtce:ParameterInstanceRef")
+                param_ref.attrib["parameterRef"] = "PKT_LEN"
+                linear_adjustment = Et.SubElement(
+                    dynamic_value, "xtce:LinearAdjustment"
+                )
+                linear_adjustment.attrib["slope"] = str(8)
+                # Intercept is the start bit of this parameter - (header bytes - 1) * 8
+                linear_adjustment.attrib["intercept"] = str(-(row["startBit"] - 40))
+
+        if row["convertAs"] == "ANALOG":
```
tech3371: nice!

tools/xtce_generation/telemetry_generator.py
```diff
@@ -26,298 +64,280 @@ class TelemetryGenerator:
...
+            default_calibrator, "xtce:PolynomialCalibrator"
+        )
+        # FIXME: Use lowValue / highValue from the conversion sheet
+        # FIXME: Handle segmented polynomials (only using first segment now)
+        for i in range(8):
+            col = f"c{i}"
+            if conversion[col] != 0:
+                term = Et.SubElement(polynomial_calibrator, "xtce:Term")
+                term.attrib["coefficient"] = str(conversion[col])
+                term.attrib["exponent"] = str(i)
```
tech3371: I don't enough about how xml does calculation but it seems much nicer if we can do it here like you have

tools/xtce_generation/telemetry_generator.py
```diff
@@ -26,298 +64,280 @@ class TelemetryGenerator:
...
+            parameter_type.attrib["name"] = name
+            parameter_type.attrib["signed"] = "false"
 
-        return telemetry_metadata
+            encoding = Et.SubElement(parameter_type, "xtce:IntegerDataEncoding")
+            encoding.attrib["sizeInBits"] = str(parameter_data["lengthInBits"])
+            encoding.attrib["encoding"] = "unsigned"
 
-    def create_remaining_parameters(self, parameter_set):
+    def _add_parameter_type(self, row):
```
tech3371: This is not important but slight personal preference. can you move this down with `_add_parameter()` function?

tech3371 (commented):
Really nice refactor of code. I had few questions.

tools/xtce_generation/telemetry_generator.py
greglucas: Yes, I think we need this because they have to be unique if the same mnemonic has different size or conversion across different packets. (SPARE for instance where it may be 2 bits in one packet and 1 bit in another, ...)

Open to suggestions if there is a better way to handle this.

tools/xtce_generation/telemetry_generator.py
greglucas: Yep, I moved it down into the other section now.

greglucas:
I think I addressed all of @tech3371's comments. I also removed the housekeeping decoder since I moved that work into a separate PR.

@anamanica can you take a look at this and see what you think from a CLI perspective? I moved it into the main package so it will get installed with our main package and usable from anywhere. This doesn't bring in any new dependencies and just exits and tells the user to install the extra required package instead.

```bash
# Will just change the output extension to produce a new file path/to/excel_file.xml
imap_xtce path/to/excel_file.xlsx

# Explicit output location
imap_xtce path/to/excel_file.xlsx --output imap_processing/swapi/packet_definitions/swapi_packet_definition.xml
```

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -0,0 +1,426 @@
...
+    {
+        "name": "PKT_LEN",
+        "lengthInBits": 16,
+        "description": "CCSDS Packet Length "
+        "(number of bytes after Packet length minus 1)",
+    },
+]
+
+
+class XTCEGenerator:
```
maxinelasp, line 59: Is this a complete re-write of telemetry_generator? Or is this a rename/move?

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -0,0 +1,426 @@
...
+                if pd.isna(row.get("packetName")):
+                    # This is a poorly formatted row, skip it
+                    continue
+                name = f"{row['packetName']}_{row['mnemonic']}"
+                parameter_ref_entry = Et.SubElement(
+                    packet_entry_list, "xtce:ParameterRefEntry"
+                )
+                parameter_ref_entry.attrib["parameterRef"] = name
+                # Add this parameter to the ParameterSet too
+                self._add_parameter(row, final_two_rows=i >= final_row - 1)
```
maxinelasp: So this assumes that the last two rows are always dynamic? Is that true for all instruments and HK?

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -0,0 +1,426 @@
...
+                # dynamic value. (there is sometimes an extra checksum in the final row)
+                dynamic_value = Et.SubElement(size_in_bits, "xtce:DynamicValue")
+                param_ref = Et.SubElement(dynamic_value, "xtce:ParameterInstanceRef")
+                param_ref.attrib["parameterRef"] = "PKT_LEN"
+                linear_adjustment = Et.SubElement(
+                    dynamic_value, "xtce:LinearAdjustment"
+                )
+                linear_adjustment.attrib["slope"] = str(8)
+                # Intercept is the start bit of this parameter - (header bytes - 1) * 8
+                linear_adjustment.attrib["intercept"] = str(-(row["startBit"] - 40))
```
maxinelasp: I like having this be automatic, but the assumption about the last 2 rows being dynamic seems like it's not right? I guess if we assume the last row is always dynamic and it isn't then it works the same way, so that part seems fine. But why the last two, if we can only have one dynamic length variable?

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -0,0 +1,426 @@
...
+        ".xml extension."
+    )
+    file_path_help = "Provide the full path to the input excel file."
+
+    parser = argparse.ArgumentParser(prog="imap_xtce", description=description)
+    parser.add_argument("excel_file", type=Path, help=file_path_help)
+    parser.add_argument("--output", type=Path, required=False, help=output_help)
+
+    if not find_spec("openpyxl"):
+        parser.error(
```
maxinelasp, line 401: Does it make sense to just include openpyxl in the imap_processing installer? I understand wanting to keep it small, but I don't love this. However, I can live with it if openpyxl is a huge dependency or something.

imap_processing/tests/ccsds/test_excel_to_xtce.py
```diff
@@ -0,0 +1,60 @@
+"""Testing for xtce generator template."""
```
maxinelasp, line 1: Is this file the same as test_xtce_generator_template?

maxinelasp (commented):
I recently got a request from MAG that would be totally solved by this PR, so it definitely seems good to merge. Sorry for the slow review! There were some updates to the XTCE generator recently, will that cause issues? 

This review assumes that most of excel_to_xtce.py is moved from telemetry_generator.py. I skimmed it but did not review in depth, except the variable length pieces. If there's something else changed, or if it is significantly different, let me know and I'll review more closely. 

Did you generate some example files using this code and compare them to existing ones?

imap_processing/ccsds/excel_to_xtce.py
greglucas: It is a pretty significant refactor, but then I did do a `git mv ...` and it still says it is an added/deleted file for some reason rather than keeping the "diff" from before. I think this might be because it was such a large refactor and so many lines changed that it was easier for git to call it new/deleted.

imap_processing/ccsds/excel_to_xtce.py
greglucas: Great question. I don't love this myself, but didn't know a better way. Maybe just assume any BYTE value is dynamic is better rather than special casing the two final rows? Let me know your thought on setting everything to be dynamic instead.

We do need the last row as that is most instruments. But Hi has a final fixed-length checksum that comes after their BYTE field, so it is why I added the final two to this.

imap_processing/ccsds/excel_to_xtce.py
greglucas: > But why the last two, if we can only have one dynamic length variable?

Is this true? If so, I think we could just make all BYTE entries be dynamic and it will be easier than restricting it to the final two entries.

Testing with Hi and SWAPI, this did appear to work with both the final row and the second-to-last row being correct.

imap_processing/ccsds/excel_to_xtce.py
greglucas: This is somewhat a personal preference of mine where if it isn't going to be used a majority of the time, then we probably don't need it as a direct dependency. It is relatively small, but this should really only be used once to generate the XTCE file I think. Once we have the XTCE file, then the dependency isn't ever used again.

I didn't really want to bring this into the main package and just leave it in the "tools" section, but then the CLI that @anamanica added wouldn't be installed with the package because the tools/ directory isn't installed.

Not a great answer for you, but I'd prefer to keep it optional and not a required dependency.

imap_processing/tests/ccsds/test_excel_to_xtce.py
greglucas: Yes, with a few updates to satisfy the CLI changes.

imap_processing/ccsds/excel_to_xtce.py
greglucas: I pushed up a commit to remove the final two and just make everything dynamic for now.

imap_processing/ccsds/excel_to_xtce.py
maxinelasp: It was my understanding that there could only be one variable-length field in the XTCE. All my knowledge of variable length fields basically comes from [this](https://lasp.slack.com/archives/C05CRQL2LJF/p1695745620068189) slack thread though. [Here](https://lasp.slack.com/archives/C05CRQL2LJF/p1694535292813049) is also some info.

imap_processing/tests/ccsds/test_excel_to_xtce.py
maxinelasp: Great, that's what I assumed so I didn't give it a close review.

imap_processing/ccsds/excel_to_xtce.py
maxinelasp: Ok, just wanted to know what was being updated.

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -288,22 +284,19 @@ def _add_parameter(self, row: pd.Series, final_two_rows: bool = False) -> None:
...
-                fixed_value.text = str(row["lengthInBits"])
+
+            # If it is a byte field consider it a dynamic value.
+            dynamic_value = Et.SubElement(size_in_bits, "xtce:DynamicValue")
+            param_ref = Et.SubElement(dynamic_value, "xtce:ParameterInstanceRef")
+            param_ref.attrib["parameterRef"] = "PKT_LEN"
+            linear_adjustment = Et.SubElement(dynamic_value, "xtce:LinearAdjustment")
+            linear_adjustment.attrib["slope"] = str(8)
+            # Intercept is the start bit of this parameter - (header bytes - 1) * 8
+            linear_adjustment.attrib["intercept"] = str(-(row["startBit"] - 40))
```
maxinelasp: I'm pretty sure you need a "+8" here

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -0,0 +1,419 @@
...
+    path_to_excel_file : Path
+        Path to the excel file.
+    """
+
+    def __init__(self, path_to_excel_file: Path):
+        # Read in all sheets from the excel file
+        self.sheets = pd.read_excel(path_to_excel_file, sheet_name=None)
+        # Set up the packet mapping from packetName to Apid
+        packet_sheet = self.sheets["Packets"]
+        self._packet_mapping = packet_sheet.set_index("packetName")["apId"].to_dict()
```
tech3371, line 88: I tried this with SWE packet definition and SWE seems to use `apIdHex` instead of `apId`. They are same information but saved in different ways.

tech3371: Another thing I was wondering is if we can add option to filter which packets we want to write to xml? Eg. write xtce definition for only science packets and leave out non-science ones. Some way to take in user input for that.

imap_processing/ccsds/excel_to_xtce.py
greglucas: Thanks for talking this through with me! I _think_ I've got it now, and using the "startBit" was actually wrong. We need the sum of all other bits to account for those potential extra fill values. Check out my latest commit to see if you agree with that update or I'm still missing something.

imap_processing/ccsds/excel_to_xtce.py
greglucas: > I tried this with SWE packet definition and SWE seems to use apIdHex instead of apId. They are same information but saved in different ways.

Can you send me their Excel file? Do you mean they don't have an apid column? Or that they have the hex values in the apid column? We could probably have a fallback here to calculate the apid from the hex if needed.

> Another thing I was wondering is if we can add option to filter which packets we want to write to xml? Eg. write xtce definition for only science packets and leave out non-science ones. Some way to take in user input for that.

I debated this and wasn't sure whether that would be useful or not. My thought was that it might be easier to just delete entries from the Excel file first if you don't want to include those rather than having to learn a new input/output flag to use? For example, Hi has a lot of packets listed, but there are only a few tabs listed so then the program outputs:
```
Packet definition for H45_BOOT_HK not found in the excel file.
Packet definition for H45_BOOT_MEMDMP not found in the excel file.
```
So if you don't want certain entries in the XTCE then you could just delete those tabs from the spreadsheet.

tech3371: Deleting tabs/packets we don't want include works for me!

tech3371 (approved):
You have addressed my comments and concerned. This looks great and going to help us tons!

PR #655, opened 2024-06-21
Py-property Update

anamanica:
# Change Summary

## Overview
Making global_attributes and variable_attributes Python Properties to eliminate use of the methods get_global_attributes and get_variable_attributes.

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
- cdf_attribute_manager.py
    - Changing methods to properties with decorators like `@property` and `@global_attributes.setter` 
- test_cdf_attribute_manager.py
    - Expand tests to utilize Python Properties while not breaking old tests.
 
## Testing
Two new test functions added to test_cdf_attribute_manager.py:
    - test_property_variable_attrs
    - test_property_global_attrs
Both tests validate that the python properties work as they should.


imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -76,10 +76,42 @@ def __init__(self, data_dir: Path):
...
+        # Validating required global attributes are present
+        for attr_name in self.global_attribute_schema.keys():
+            required_attr = self.global_attribute_schema[attr_name]["required"]
+            if required_attr is True and attr_name not in self._global_attributes:
+                self._global_attributes[attr_name] = None
+
+        return self._global_attributes
+
+    @global_attributes.setter
+    def global_attributes(self, instrument_id: str):
```
tmplummer: This setter has me confused. I think of setters being used to modify some data in the class, but I don't see that anything is being modified here. The only possible side-effect of calling `self.get_global_attributes(instrument_id)` in this setter is that it validates the global attributes for the input instrument by raising an exception if a required global variable is not defined.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -76,10 +76,42 @@ def __init__(self, data_dir: Path):
...
+            attribute = self.variable_attribute_schema["attribute_key"][attr_name][
+                "required"
+            ]
+            if attribute is True and attr_name not in self._variable_attributes.keys():
+                self._variable_attributes["attribute_key"] = None
+
+        return self._variable_attributes
+
+    @variable_attributes.setter
+    def variable_attributes(self, variable_name: str):
```
tmplummer: Same comment about the setter here.

imap_processing/cdf/imap_cdf_manager.py
```diff
@@ -12,9 +12,13 @@
...
 
-    def __init__(self):
+    def __init__(self, source_dir_input=None):
         """Set the path to the config directory."""
-        super().__init__(Path(__file__).parent / "config")
+        if source_dir_input is None:
+            super().__init__(Path(__file__).parent / "config")
+        else:
+            super().__init__(Path(__file__).parent / "config")
+            self.source_dir = Path(__file__).parent / source_dir_input
```
tmplummer, line 22: If the `super().__init__()` method is intended to be called in both cases, this could be simplified as follows
```suggestion
        super().__init__(Path(__file__).parent / "config")
        if source_dir_input is not None:
            self.source_dir = Path(__file__).parent / source_dir_input
```
But, it isn't clear to me that the default schema files are intended to be loaded if a source_dir is specified.

tmplummer (commented):
Wow, that is a tremendous amount of test coverage! Thanks for being so thorough.

imap_processing/cdf/cdf_attribute_manager.py
anamanica: Thank you! This is my first time working with properties, so I will appreciate all the feedback I can get. I updated the property itself and eliminated the setter, so it makes much more sense now.

PR #657, opened 2024-06-21
MNT: Pass around dataset objects rather than file objects

greglucas:
# Change Summary

## Overview

This is a major update to use datasets as our primary objects that we pass around rather than paths to the datasets on disk. Rather than saving files to disk and writing cdf files in each individual instrument, we can save the CDF files in one common location (CLI) and then pass the datasets around to all of the processing functions.

Many updated files due to the thrashing nature of this change. The tests are now testing against "Logical_source" because we actually don't have the filename in the dataset until the write_cdf() routine. But that is tested already in write_cdf, so not necessary in each individual instrument.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -21,6 +21,7 @@ from typing import final
 from urllib.error import HTTPError
 
 import imap_data_access
+import xarray as xr
 
 import imap_processing
 from imap_processing.cdf.utils import load_cdf, write_cdf
@@ -372,19 +373,22 @@ class ProcessInstrument(ABC):
         """
         raise NotImplementedError
 
-    def post_processing(self, products: list[str]):
+    def post_processing(self, datasets: list[xr.Dataset]):
         """
         Complete post-processing.
 
-        Default post-processing consists of uploading newly generated products
-        to the IMAP SDC. Child classes can override this method to customize the
+        Default post-processing consists of writing the datasets to local storage
+        and then uploading those newly generated products to the IMAP SDC.
+        Child classes can override this method to customize the
         post-processing actions.
 
         Parameters
         ----------
-        products : list[str]
-            A list of file paths (products) produced by do_processing method.
+        datasets : list[xarray.Dataset]
+            A list of datasets (products) produced by do_processing method.
         """
+        logger.info("Writing products to local storage")
+        products = [write_cdf(dataset) for dataset in datasets]
         self.upload_products(products)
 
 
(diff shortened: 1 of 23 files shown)


maxinelasp (commented):
I like this standardization and I think it's a good change. It's nice to decouple the reading and writing of datasets from the processing code for lots of reasons. I didn't look super carefully through the other instruments, but the mag change looks good to me and there aren't other spots in mag or glows that should be updated.

maxinelasp (approved):

imap_processing/cli.py
```diff
@@ -372,19 +373,22 @@ def do_processing(self, dependencies: list):
...
 
         Parameters
         ----------
-        products : list[str]
-            A list of file paths (products) produced by do_processing method.
+        datasets : list[xarray.Dataset]
+            A list of datasets (products) produced by do_processing method.
         """
+        logger.info("Writing products to local storage")
+        products = [write_cdf(dataset) for dataset in datasets]
```
tech3371, line 391: moving this here is a great idea!

Not for this PR, but in the future, I wonder if we can refactor to avoid passing in version as well in `do_processing` function and pass to `write_cdf` function or set `Data_version` here in this function? Because most of us are passing it in only for the purpose adding it in dataset global attrs

imap_processing/cli.py
```diff
@@ -498,18 +498,17 @@ def do_processing(self, dependencies: list):
                     f"{dependencies}. Expected only one dependency."
                 )
             datasets = hi_l1a.hi_l1a(dependencies[0], self.version)
-            products = [write_cdf(dataset) for dataset in datasets]
         elif self.data_level == "l1b":
-            dataset = hi_l1b.hi_l1b(dependencies[0], self.version)
-            products = [write_cdf(dataset)]
+            dependencies = [load_cdf(dependency) for dependency in dependencies]
```
tmplummer, line 502: Thank you for cleaning up these variable names.

imap_processing/cli.py
```diff
@@ -580,9 +579,7 @@ def do_processing(self, dependencies):
                 )
             # read CDF file
             processed_data = PacketParser(dependencies[0], self.version).data
-            cdf_file_path = write_cdf(processed_data)
-            print(f"processed file path: {cdf_file_path}")
-            return [cdf_file_path]
+            return [processed_data]
```
```suggestion
            return [dataset]
```

There are only a few instruments that are using a different variable name other than `dataset` here. Should we change the them so all instruments are consistent?

bourque (approved):
Agree with Maxine -- +1 for standardization and decoupling!

tmplummer (approved):
LGTM

imap_processing/cli.py
greglucas: Yes, great suggestion. I just updated all of the returns to be either `[dataset]` or `datasets` depending on whether one or multiple were being returned. Maybe we should update that later as well to force us to always return a list from the calling procedure... But that will lead to more discussion so lets save it for a follow-up PR.

tech3371 (approved):
Thank you for doing this!

PR #660, opened 2024-06-25
CDF Updates

### LINKED ISSUE #593: CDF Attribute manager: Move methods to properties
### Summary

Instead of using the methods `get_global_attributes` and `get_variable_attributes` to access and validate global and variable attributes within cdf_attribute_manager.py, self.variable_attributes and self.global_attributes should be changed to Python Properties. 

When the properties are accessed, the attributes should be validated against the correct schema (similar to how the getters work now). 

### Requirements

The more specific the better.
  - [ ] Move variable_attributes and global_attributes to properties
  - [ ] Have them be validated during access 
  - [ ] Expand or update unit tests to check new functionality

### Related tickets

_No response_

anamanica:
# Change Summary

## Overview
Cleaning up the CdfAttributeManager class code in order to ensure schema validation when a user is accessing global and variable attributes. This is done by changing `.global_attributes` and `.variable_attributes` to be private properties, establishing the methods `.get_global_attributes` and `.get_variable_attributes` as the way to retrieve a validated, specific dictionary. 

This particular change has been slightly debated. Originally, the methods `.get_global_attriubtes` and `.get_variable_attributes` were going to be changed into Python properties, but that proved to be unintuitive with the functionality of getters and setters. Then, we were going to create a new class to place validation into the `__getitem__` functionality of dictionaries, but this would require creating a class with multiple inheritance and doing some funky things. Overall, it was decided by Maxine and I that the best current solution is the one described above. 

This branch was created from cdf-attribute-tests in order to carry the testing work from that branch onto this one.

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
-cdf_attribute_manger
    - Changing `.global_attributes` and `.variable_attributes` to private methods.
    - Finishing the code of the method `.get_variable_attributes`.
- ultra_l1_utils.py
    - Changing lines 37 and 51 to use `.get_variable_attributes` instead of `.variable_attributes`
- codice_l1b.py
    - Changing line 66 to use `.get_variable_attributes` instead of `.variable_attributes` 
- codice/utils.py
    - Changing line 158 to use `.get_variable_attributes` instead of `.variable_attributes`

## Testing
All covered in test_cdf_attribute_manager.py.

closes #593


imap_processing/cdf/tests/__init__.py
tmplummer, line 1: Commenting on the `__init__.py` file but this is relevant to all files in this directory...

All `.py` files in this directory should be moved to `imap_processing/tests/cdf/`
All `.yaml` files should be moved under the parent directory `imap_processing/tests/cdf/test_data`

Basically, we want the directory structure under `imap_processing/tests` to mirror the directory structure of under `imap_processing` with the exception of test data.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -236,10 +238,22 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
+            Information containing specific variable attributes
+            associated with "variable_name".
         """
-        # TODO: Create a variable attribute schema file, validate here
-        if variable_name in self.variable_attributes:
-            return self.variable_attributes[variable_name]
-        # TODO: throw an error?
-        return {}
+        output = dict()
+        for attr_name in self.variable_attribute_schema.keys():
```
tmplummer: I don't think that this attribute schema validation is working correctly. The reason that I say this is on Hi, I define a variable attribute `dtype` in all of my variable attribute dictionaries that I use when instantiating the xarray.DataArray objects that store the Hi data. Since `dtype` is not defined in the variable attribute schema, I would expect this change to break several Hi processing tests.

I would suggest adding tests that verify the output dictionary looks as expected when the condition `variable_name in self._variable_attributes[variable_name]` is True.

imap_processing/cdf/tests/shared/default_variable_cdf_attrs_schema.yaml
tmplummer, line 1: I would suggest reducing this test schema file to have the minimum number of entries needed for complete test coverage. I think that you can get away with as few as two entries under `attribute_key`. They can be completely made up entries with values specified such that you can test everything that you need to.

imap_processing/cdf/tests/shared/default_global_cdf_attrs_schema.yaml
tmplummer, line 1: This fill can also be reduced to a minimum needed set of global attributes needed for complete test coverage... See my comment on the `default_variable_cdf_attrs_schema.yaml` for more context.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,351 @@
...
+    assert (
+        cdf_manager.variable_attribute_schema["attribute_key"]["RESOLUTION"]["required"]
+        is True
+    )
+
 
-    # Expected failure: file_naming_convention is not in schema.
-    # for attr in cdf_manager.global_attributes.keys():
-    #     assert attr in cdf_manager.global_attribute_schema.keys()
+def test_global_attribute():
```
tmplummer:
```suggestion
def test_global_attribute():
```
```suggestion
def test_load_global_attribute():
```

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -14,6 +14,8 @@
 DEFAULT_GLOBAL_CDF_ATTRS_SCHEMA_FILE = "default_global_cdf_attrs_schema.yaml"
 DEFAULT_VARIABLE_CDF_ATTRS_SCHEMA_FILE = "default_variable_cdf_attrs_schema.yaml"
 
+# Pre-commit fix
```
maxinelasp: Remove this comment

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -236,10 +238,22 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
-        if variable_name in self.variable_attributes:
-            return self.variable_attributes[variable_name]
-        # TODO: throw an error?
-        return {}
+        output = dict()
+        for attr_name in self.variable_attribute_schema.keys():
+            if variable_name in self._variable_attributes:
+                output = self._variable_attributes[variable_name]
+            elif (
+                variable_name is not None
```
maxinelasp: I don't think `variable_name` should ever be None, so this check is not needed. Actually, this whole `elif` block should be removed, because the person calling get_variable_attributes should always provide variable_name. 

This was necessary in get_global_attributes because the instrument_id is an optional parameter, so it can be none.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -1,58 +1,351 @@
...
-    mag_l1a_global_attrs = cdf_manager.get_global_attributes("imap_mag_l1a_norm-raw")
 
+def test_get_global_attributes():
+    """
+    Test function that covers:
+        get_global_attributes
+    """
+
+    # Initialize CdfAttributeManager object which loads in default info
+    cdf_manager = CdfAttributeManager(Path(__file__).parent.parent / "config")
```
tmplummer: Is there a reason to first load the default (not for testing) schema and default global attributes? It makes this test a bit confusing to have that cross pollination that isn't very obvious. Can the test schema be used instead so that this is totally isolated to using test specific schema and global attribute defaults?

This comment applies to all of the test functions in this file.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -236,10 +238,22 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
+                and attr_name in self._variable_attributes["variable_name"]
+            ):
+                output = self._variable_attributes["variable_name"][attr_name]
+            elif (
+                variable_name not in self.variable_attribute_schema
```
maxinelasp: Rather than checking `variable_name` against the schema, we should check the attributes against the schema. So you would rewrite this block as something like:

```
for attribute in self.variable_attribute_schema.keys():
  if attribute in self._variable_attributes[variable_name]:
    add to output
  if attribute not in self._variable_attributes[variable_name] and variable is required:
    add to output as None
```

imap_processing/cdf/imap_cdf_manager.py
```diff
@@ -11,11 +11,28 @@
 
 
 class ImapCdfAttributes(CdfAttributeManager):
-    """Contains IMAP specific tools and settings for CDF management."""
+    """
```
maxinelasp: Looks like this PR includes some stuff from your test PR - that will go away once you merge the test PR.

maxinelasp (commented):
A few changes, but looking good so far!

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: It looks like this is why your tests are failing too!

imap_processing/cdf/tests/test_imap_cdf_manager.py
tmplummer, line 1: Nice job implementing succinct tests in this file that cover only the functionality added by the `ImapCdfAttributes` class!

tmplummer (commented):
This is a ton of work and things are looking pretty good. I agree with the decision to remove the property accessors.

Let me know if you need any clarification on the comments that I left.

imap_processing/cdf/cdf_attribute_manager.py
anamanica: Hahah, yes, thank you so much! I caught this momentarily after I submitted the PR request. Felt very silly. I should have probably started this one as a draft.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -217,13 +216,36 @@ def load_variable_attributes(self, file_name: str) -> None:
...
         # Add variable attributes from file_name. Each variable name should have the
-        # Required subfields as defined in the variable schema.
+        # required sub fields as defined in the variable schema.
         raw_var_attrs = CdfAttributeManager._load_yaml_data(self.source_dir / file_name)
         var_attrs = raw_var_attrs.copy()
 
         self._variable_attributes.update(var_attrs)
 
-    def get_variable_attributes(self, variable_name: str) -> dict:
+    # def get_variable_attributes(self, variable_name: str) -> dict:
```
maxinelasp: Is there a reason why this is left as a comment? We can go back to prior commits and retrieve code if we want it, so there's no need to leave it commented, just delete it. If this is just for reference while you're working that's fine, just make sure it's removed before merging

imap_processing/cdf/tests/__init__.py
anamanica: Sweet! I had to leave one .yaml file under `/tests/cdf` due to the way the code is written in `__init__`, but I will be sure to ask Maxine about that.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -235,25 +257,54 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
-                output = self._variable_attributes[variable_name]
-            elif (
-                variable_name is not None
-                and attr_name in self._variable_attributes["variable_name"]
-            ):
-                output = self._variable_attributes["variable_name"][attr_name]
+        for attr_name in self.variable_attribute_schema["attribute_key"]:
+            # Current case to handle hi instruments and avoid "dtype" key error
+            # TODO: Do we want to keep the check_schema flag?
+            if check_schema is False:
```
maxinelasp: Rather than putting this if check inside the for-loop, it should just be the first line in the method. Right now, if someone adds a variable attribute that isn't in the schema, this code will never get reached because it loops through the attributes in the schema. If `check_schema` is false, then it should entirely skip this

imap_processing/cdf/tests/shared/default_variable_cdf_attrs_schema.yaml
anamanica: Ah yes! Initially I had made a default_test_schema file, but then was asked to use the actual schema file since that is a bit more concrete that the individual instrument files.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -235,25 +257,54 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
         output = dict()
-        for attr_name in self.variable_attribute_schema.keys():
-            if variable_name in self._variable_attributes:
-                output = self._variable_attributes[variable_name]
-            elif (
-                variable_name is not None
-                and attr_name in self._variable_attributes["variable_name"]
-            ):
-                output = self._variable_attributes["variable_name"][attr_name]
+        for attr_name in self.variable_attribute_schema["attribute_key"]:
```
maxinelasp, line 254: I'm not sure why you're looping through the "attribute_key" dictionary from the schema? I think this should be something like:

```
variable_schema = self.variable_attribute_schema.get(variable_name, {})
for attr_name in variable_schema: 
...
```
this code assumes that you have moved the `check_schema` if statement up above it. If we know we don't need to check the schema, we can short-circuit this and skip any schema validation.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -235,25 +257,54 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
+            # TODO: Do we want to keep the check_schema flag?
+            if check_schema is False:
+                if variable_name in self._variable_attributes:
+                    return self._variable_attributes[variable_name]
+                # TODO: throw an error?
+                return {}
+            # Standard, expected cases
+            elif attr_name in self._variable_attributes[variable_name]:
+                output[attr_name] = self._variable_attributes[variable_name][attr_name]
+            elif attr_name in self._variable_attributes:
```
maxinelasp: Why do we have this case?

imap_processing/cdf/tests/test_cdf_attribute_manager.py
anamanica: I believe the reason we are using the default schema for these tests is because they are actually loaded into the CdfAttributeManager class from the `__init__` function. I would have to change the code here if I wanted to use test schema, which I have been told is not ideal.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -235,25 +257,54 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
+                output[attr_name] = self._variable_attributes[attr_name]
+            # Case to handle DEPEND_i schema issues
+            elif attr_name == "DEPEND_i":
+                # range(3) because the highest DEPEND_i value is 3.
+                for i in range(3):
```
maxinelasp: You can't assume that there are only 3 DEPEND_i cases. To address your todo comment below, I also believe depend_0 does need to exist and it needs to be `epoch`.

Suggested rewrite: 
```
elif attr_name == "DEPEND_i":
  # Find all the attributes of variable_name that contain "DEPEND"
  variable_depend_attrs = [key for key in self._variable_attributes[variable_name].keys() if "DEPEND" in key] 
  # Confirm that each DEPEND_i attribute is unique
  if len(set(variable_depend_attrs)) != len(variable_depend_attrs):
    logger.warn(f"Found duplicate DEPEND_i attribute in variable {variable_name}: {variable_depend_attrs}")
  for variable_depend_attr in variable_depend_attrs: 
    output[variable_depend_attr] = self._variable_attributes[variable_name][variable_depend_attr]
```
Also as a TODO, please add a comment to add some additional validation for the depend attributes.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -235,25 +257,54 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
+                        output[attr_name_depend] = self._variable_attributes[
+                            variable_name
+                        ][attr_name_depend]
+            # Includes attr_name != DEPEND_0 to handle lo_l1a. Because
+            # epoch is not a dimension for some fields, DEPEND_i starts at >=1,
+            # and thus should not be included.
             elif (
-                variable_name not in self.variable_attribute_schema
-                and self.variable_attribute_schema["required"]
+                self.variable_attribute_schema["attribute_key"][attr_name]["required"]
```
maxinelasp, line 306: This should be checking `self.variable_attribute_schema[variable_name]` not `self.variable_attribute_schema["attribute_key"]`

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -235,25 +257,54 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
+                and attr_name not in self._variable_attributes[variable_name]
+                and attr_name != "DEPEND_0"
             ):
-                output[variable_name] = None
+                if attr_name == "DEPEND_0":
```
maxinelasp: This doesn't need to be a different if statement

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -235,25 +257,54 @@ def get_variable_attributes(self, variable_name: str) -> dict:
...
+                        ][attr_name_depend]
+            # Includes attr_name != DEPEND_0 to handle lo_l1a. Because
+            # epoch is not a dimension for some fields, DEPEND_i starts at >=1,
+            # and thus should not be included.
             elif (
-                variable_name not in self.variable_attribute_schema
-                and self.variable_attribute_schema["required"]
+                self.variable_attribute_schema["attribute_key"][attr_name]["required"]
+                and attr_name not in self._variable_attributes[variable_name]
+                and attr_name != "DEPEND_0"
```
maxinelasp: Since depend_0 is always required, we don't need this line

imap_processing/cdf/tests/imap_test_variable.yaml
```diff
@@ -15,12 +15,13 @@ test_field_1:
...
   UNITS: test_units
   VAR_TYPE: test_variable_type
-  SCALETYP: test_scaletyp
-  MONOTON: test_monoton
   TIME_BASE: 10
   TIME_SCALE: test_time_scale
   REFERENCE_POSITION: test_reference_position
-  NOT_REQUIRED: test_not_required
+  LEAP_SECONDS_INCLUDED: test_not_required
+  DEPEND_1: test_depend_1
```
maxinelasp, line 22: Should include `DEPEND_0` in your test data. I think it's debatable whether we want to require `DEPEND_0` in the attribute definition since it should always be present, but I think most of us have been explicitly defining it, so your tests should reflect that.

imap_processing/cdf/tests/test_cdf_attribute_manager.py
```diff
@@ -294,58 +294,51 @@ def test_get_variable_attributes():
     imap_test_variable = cdf_manager.get_variable_attributes("test_field_1")
```
maxinelasp, line 294: Can you add a test case for `check_schema=False`?

imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -102,15 +102,15 @@ def allocate_pset_dataset(n_esa_steps: int, sensor_str: str):
     # preallocate coordinates xr.DataArrays
     coords = dict()
     # epoch coordinate has only 1 entry for pointing set
-    attrs = cdf_manager.get_variable_attributes("hi_pset_epoch").copy()
+    attrs = cdf_manager.get_variable_attributes("hi_pset_epoch", False).copy()
```
maxinelasp: nitpick - but - when calling keywords like this, I prefer to have it like `check_schema=False` so someone glancing through this code can see exactly what is going on.
```suggestion
    attrs = cdf_manager.get_variable_attributes("hi_pset_epoch", check_schema=False).copy()
```

maxinelasp (commented):
Some additional comments but it's definitely getting close!!

imap_processing/cdf/tests/test_cdf_attribute_manager.py
maxinelasp: I prefer to use the real schema against the test data so we can avoid the real schema and test schema from getting out of sync, so I asked Ana to do it this way. 

But, since you got a comment on it, can you add a short comment in the code explaining why you're loading in the default schema?

imap_processing/cdf/tests/__init__.py
maxinelasp: The reason we left the tests where they were, is because this code is going to be moved to another repo very soon and hopefully removed from here. So, I wanted it to be isolated from everything else. 

but, in general, all tests should go under `imap_processing/tests`. So I don't feel strongly about this, but I'd slightly prefer to keep it all together to make it easier to move and feel confident I got all of it.

imap_processing/cdf/tests/shared/default_variable_cdf_attrs_schema.yaml
maxinelasp: I think I see where some of the confusion is coming from. Either you should use the test_schema file (if there is some testing reason to keep it) and point to it in the tests (so change the code in Tim's other comment) OR continue using the real schema, leave the tests the same, and remove these unused schemas. 

If I'm missing something here, please let me know!

imap_processing/cdf/cdf_attribute_manager.py
anamanica: Ah, yes. So the reason I was doing this is because the default_variable_attribute_schema has a few different sections. There is the "attribute_keys" section where all the attribute names and their information is stored, and near the bottom of the file, there are the sections "data", "support_data", "metadata" that hold different collections of attribute names.

imap_processing/cdf/cdf_attribute_manager.py
anamanica: Great question! Hahah, I guess my brain was stuck in the format of global_attributes. I'll remove that.

imap_processing/cdf/cdf_attribute_manager.py
anamanica: Ah! This works way better. Much more sophisticated. Thanks! 

Regarding DEFAULT_0, I actually talked with Sean about his tests for l0_l1a, and that is were he told me that DEPEND_0 does not need to be present in some cases he is working with when epoch is not one of the dimensions. I am doing a poor job of explaining this, so I can send you the message he sent me if you would like.

imap_processing/cdf/tests/imap_test_variable.yaml
anamanica: Yes! For `test_field_1`, DEFAULT_0 is defined in default_attrs. However `test_field_3` does not have `DEPEND_0`, but I sent you a few slack questions about this case specifically.

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: You're right, so update the line to be: 
`self.variable_attribute_schema["attribute_keys"].get(variable_name, {})`

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: Ok, for posterity: according to Bryan Harter, Depend_0 must be an epoch, but it isn't actually required for every variable. So we will be updating the schema to make it not required and removing the special cases from the code.

imap_processing/cdf/tests/imap_test_variable.yaml
maxinelasp: Ok, in light of the discussion we had, disregard this comment

imap_processing/cdf/tests/test_cdf_attribute_manager.py
tmplummer: @maxinelasp, Does this mean that you are considering these tests a validation of the real schema? I think of unit tests as only covering the code implementation, not validating the schema.

maxinelasp: Yeah, good question. I can see either argument, but in this case, I think it makes sense to have the real schema be part of the unit tests because it's closely tied to the code. But, I can definitely see the philosophical argument that it's not really part of the "unit" so it shouldn't be tested.

imap_processing/cdf/cdf_attribute_manager.py
anamanica: I think the code has been changed enough that this isn't needed anymore, but please let me know if I am just missing it!

imap_processing/hi/l1a/science_direct_event.py
```diff
@@ -302,7 +302,7 @@ def create_dataset(de_data_list: list, packet_met_time: list) -> xr.Dataset:
     )
 
     for var_name, data in data_dict.items():
-        attrs = cdf_manager.get_variable_attributes(f"hi_de_{var_name}").copy()
+        attrs = cdf_manager.get_variable_attributes(f"hi_de_{var_name}", False).copy()
```
tmplummer:
```suggestion
        attrs = cdf_manager.get_variable_attributes(f"hi_de_{var_name}", check_schema=False).copy()
```

imap_processing/tests/hi/test_hi_l1b.py
```diff
@@ -21,6 +23,7 @@ def test_hi_l1b_hk():
     assert l1b_dataset.attrs["Logical_source"] == "imap_hi_l1b_45sensor-hk"
 
 
+@pytest.mark.xfail()
```
tmplummer, line 24: I will write myself a ticket to fix this test.

tmplummer (approved):
Thanks for persevering through this challenging PR. Sorry that my off-nominal use of the CdfAttributeManager caused problems. I'm happy with this end result.

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: Yes! Looks good now, nice job!

maxinelasp (approved):
Nice job!

PR #661, opened 2024-06-25
Add mypy Pre-Commit Hook for files Pt.2

daralynnrhode:
# Adding the mypy Pre-commit hook for non-instrument files and folders

## Overview
Changes for this PR will occur in the `imap_processing` files  and `imap_processing/cdf` Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

The places where inline ignore comments are will be changed/fixed  before the PR is merged. 

## Updated Files
-imap_processing/utils.py
-imap_processing/decom.py
-imap_processing/cli.py
-imap_processing/cdf/utils.py
-imap_processing/cdf/imap_cdf_manager.py
-imap_processing/cdf/cdf_attribute_manager.py


imap_processing/cdf/utils.py
```diff
@@ -163,4 +167,4 @@ def write_cdf(dataset: xr.Dataset):
         terminate_on_warning=True,
     )  # Terminate if not ISTP compliant
 
-    return file_path
+    return [file_path]
```
maxinelasp: This change is why your tests are failing. Why did you change it to a list? It should return a single path.

daralynnrhode: Without the [] the mypy checks would flag it as an 'Any' type being returns instead of a list. Should the return type for the function then be changed? In order to reflect the single path.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -148,7 +154,7 @@ def add_global_attribute(self, attribute_name: str, attribute_value: str) -> Non
         self.global_attributes[attribute_name] = attribute_value
 
     @staticmethod
-    def _load_yaml_data(file_path: str | Path) -> dict:
+    def _load_yaml_data(file_path: str | Path) -> yaml:
```
maxinelasp, line 154: It does return a dict that is structured like yaml. (`dict` where the keys are strings and the values are dicts)

imap_processing/cdf/utils.py
```diff
@@ -115,7 +119,7 @@ def write_cdf(dataset: xr.Dataset):
 
     Returns
     -------
-    file_path : pathlib.Path
+    file_path : list[Path]
```
maxinelasp: This change should be undone

imap_processing/cli.py
```diff
@@ -353,7 +354,7 @@ def pre_processing(self):
         return self.download_dependencies()
 
     @abstractmethod
-    def do_processing(self, dependencies: list):
+    def do_processing(self, dependencies: list):  # type: ignore[no-untyped-def]
```
maxinelasp, line 361: What does "ignore[no-untyped-def]" mean?

imap_processing/cli.py
```diff
@@ -618,24 +626,31 @@ def do_processing(self, dependencies):
         elif self.data_level == "l1b":
             data_dict = {}
             for dependency in dependencies:
-                dataset = load_cdf(dependency, to_datetime=True)
+                dataset = load_cdf(dependency, to_datetime=True)  # type: ignore[arg-type]
+                # Todo Change
+                # Argument "to_datetime" to "load_cdf" has incompatible
```
maxinelasp: Related to this: https://github.com/python/mypy/issues/1969

imap_processing/cli.py
```diff
@@ -650,6 +665,7 @@ def do_processing(self, dependencies) -> list[Path]:
             List of output files.
         """
         print(f"Processing MAG {self.data_level}")
+        place_holder: list[Path] = []
```
maxinelasp:
```suggestion
        output_files: list[Path] = []
```

imap_processing/cli.py
```diff
@@ -686,12 +702,13 @@ def do_processing(self, dependencies) -> list[Path]:
             output_dataset = mag_l1c(input_data[0], input_data[1], self.version)
             output_files = write_cdf(output_dataset)
```
maxinelasp:
```suggestion
            output_files.extend(write_cdf(output_dataset))
```

imap_processing/cli.py
```diff
@@ -659,7 +675,7 @@ def do_processing(self, dependencies) -> list[Path]:
                     f"{dependencies}. Expected only one dependency."
                 )
             output_files = mag_l1a(dependencies[0], data_version=self.version)
```
maxinelasp:
```suggestion
            output_files.extend(mag_l1a(dependencies[0], data_version=self.version))
```

imap_processing/cli.py
```diff
@@ -659,7 +675,7 @@ def do_processing(self, dependencies) -> list[Path]:
...
-            return output_files
+            return [output_files]
```
maxinelasp:
```suggestion
           
```

imap_processing/cli.py
```diff
@@ -686,12 +702,13 @@ def do_processing(self, dependencies) -> list[Path]:
             output_dataset = mag_l1c(input_data[0], input_data[1], self.version)
             output_files = write_cdf(output_dataset)
             return [output_files]
```
maxinelasp:
```suggestion
            
```

imap_processing/cli.py
```diff
@@ -686,12 +702,13 @@ def do_processing(self, dependencies) -> list[Path]:
...
+        return place_holder
```
maxinelasp:
```suggestion
        return output_files
```

If you accept one of these suggested changes, then suggest them all! It's a good practice to only return at the end of the function (which I failed to do here)

imap_processing/cli.py
```diff
@@ -718,12 +736,13 @@ def do_processing(self, dependencies):
             # Write all processed data to CDF files
             products = [write_cdf(dataset) for dataset in processed_data]
             return products
```
maxinelasp:
```suggestion
            
```

imap_processing/cli.py
```diff
@@ -766,12 +786,13 @@ def do_processing(self, dependencies):
             return [cdf_file_path]
         else:
             print("Did not recognize data level. No processing done.")
+        return cdf_file_path
```
maxinelasp: Same change here, remove all the returns before this one

imap_processing/cli.py
```diff
@@ -814,9 +836,10 @@ def do_processing(self, dependencies: list):
             datasets = ultra_l1c.ultra_l1c(data_dict, self.version)
             products = [write_cdf(dataset) for dataset in datasets]
             return products
+        return products
```
maxinelasp: Same change here about returns

imap_processing/utils.py
```diff
@@ -51,6 +52,7 @@ def group_by_apid(packets: list):
     grouped_packets : dict
         Grouped data by apid.
     """
+    grouped_packets: dict[list]
     grouped_packets = collections.defaultdict(list)
```
maxinelasp:
```suggestion
    grouped_packets: dict[list] = collections.defaultdict(list)
```
Does this work?

maxinelasp (commented):
Nice! I had some comments but great work so far!

imap_processing/cdf/utils.py
maxinelasp: It should be a single path returned. Can you enforce the type on the creation of the variable? eg `file_path: Path = ...`

imap_processing/cli.py
daralynnrhode: It is a mypy inline ignore error command that skips over the error raised. In this case, it is needed because the function itself is not implemented. Once the function is implemented the command can be removed and a return type would need to be specified. I will add a comment noting that the command will need to be removed later.

imap_processing/cli.py
daralynnrhode: I am a little confused as to how to go about solving this issue.

imap_processing/utils.py
daralynnrhode: Yes, it does work.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -78,7 +78,9 @@ def __init__(self, data_dir: Path):
         self.global_attributes = CdfAttributeManager._load_yaml_data(
             self.source_dir / DEFAULT_GLOBAL_CDF_ATTRS_FILE
         )
-        self.variable_attributes = dict()
+        self.variable_attributes = dict()  # type: ignore[var-annotated]
+        # Todo Change, Need type annotation for "variable_attributes"
+        # (hint: "variable_attributes: dict[<type>, <type>] = ...")
```
bourque:
```suggestion
        self.variable_attributes: dict[str, dict] = dict()
```

I played around with this a bit. I think this is the solution.

imap_processing/cdf/cdf_attribute_manager.py
```diff
@@ -93,7 +95,9 @@ def _load_default_global_attr_schema(self) -> dict:
             self.source_dir / "shared" / DEFAULT_GLOBAL_CDF_ATTRS_SCHEMA_FILE
         )
         # Load the Schema
-        return CdfAttributeManager._load_yaml_data(default_schema_path)
+        return CdfAttributeManager._load_yaml_data(default_schema_path)  # type: ignore[no-any-return]
```
bourque: I am not quite sure what to do about this. My understanding is that these yaml files get converted to dictionaries, and the keys of those dictionaries could be anything, in which case saying the type is "Any" is perfectly valid.

Because of this, I am inclined to add `no-any-return` to the list of disabled error codes.

Curious to know what @maxinelasp thinks about this.

imap_processing/cdf/utils.py
```diff
@@ -76,7 +77,10 @@ def load_cdf(
     """
     # TODO: remove this when cdflib is updated to version >1.3.0
     if "to_datetime" not in kwargs:
-        kwargs["to_datetime"] = True
+        kwargs["to_datetime"] = True  # type: ignore[assignment]
+        # ToDO Change
+        # Incompatible types in assignment
+        # (expression has type "bool", target has type "dict[Any, Any]")
```
bourque:
```suggestion
        kwargs["to_datetime"] = True  # type: ignore[assignment]
```

This one confuses me. Yes, the target (`kwargs`) is of type dict, but here we are setting a value to one of the dictionary's _keys_ 🤔 

So I think it is fine to just keep this inline `ignore` comment.

imap_processing/cdf/utils.py
```diff
@@ -163,4 +167,5 @@ def write_cdf(dataset: xr.Dataset):
         terminate_on_warning=True,
     )  # Terminate if not ISTP compliant
 
-    return file_path
+    return file_path  # type: ignore[no-any-return]
+    # TODO Change,  Returning Any from function declared to return "list[Path]"
```
bourque:
```suggestion
    return file_path
```

When I tried this on my end, it didn't throw an error, so I think this `ignore` can just be removed?

imap_processing/cli.py
```diff
@@ -340,7 +341,7 @@ def process(self):
         self.post_processing(products)
         logger.info("Processing complete")
 
-    def pre_processing(self):
+    def pre_processing(self):  # type: ignore[no-untyped-def]
```
```suggestion
    def pre_processing(self) -> list:
```

imap_processing/cli.py
```diff
@@ -356,7 +357,8 @@ def pre_processing(self):
         return self.download_dependencies()
 
     @abstractmethod
-    def do_processing(self, dependencies: list):
+    def do_processing(self, dependencies: list):  # type: ignore[no-untyped-def]
+        # Todo Remove type: ignore once function is implemented.
```
```suggestion
    def do_processing(self, dependencies: list):  # type: ignore[no-untyped-def]
```

Since this is an abstract method that is meant to be implemented in child classes, I don't think it will ever actually be 'implemented' here.

So I think it is fine to keep the inline `ignore` statement here.

imap_processing/cli.py
```diff
@@ -621,24 +630,31 @@ def do_processing(self, dependencies):
         elif self.data_level == "l1b":
             data_dict = {}
             for dependency in dependencies:
-                dataset = load_cdf(dependency, to_datetime=True)
+                dataset = load_cdf(dependency, to_datetime=True)  # type: ignore[arg-type]
+                # Todo Change
+                # Argument "to_datetime" to "load_cdf" has incompatible
+                # type "bool"; expected "dict[Any, Any]"
```
```suggestion
                dataset = load_cdf(dependency, to_datetime=True)  # type: ignore[arg-type]
```

I'm happy with just keeping the inline `ignore` statement here

imap_processing/cli.py
```diff
@@ -621,24 +630,31 @@ def do_processing(self, dependencies):
...
             return [output_file]
 
         elif self.data_level == "l1c":
             data_dict = {}
             for dependency in dependencies:
-                dataset = load_cdf(dependency, to_datetime=True)
+                dataset = load_cdf(dependency, to_datetime=True)  # type: ignore[arg-type]
+                # Todo Change
+                # Argument "to_datetime" to "load_cdf" has incompatible
+                # type "bool"; expected "dict[Any, Any]"
```
```suggestion
                dataset = load_cdf(dependency, to_datetime=True)  # type: ignore[arg-type]
```

Same here

bourque (changes requested):
Nice work here, especially with these weird corner cases. I have a few suggestions to fix things, otherwise this looks good!

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: The keys of the dictionaries are expected to be `string` types always, with the values either being `string` or `dict`. If we can add that type then it should fix the issue. If not, then `any` type is ok here.

imap_processing/cdf/utils.py
maxinelasp: Actually I think this code is wrong. We shouldn't directly update kwargs like this. I think we actually want: 
` to_datetime_arg = kwargs.get("to_datetime", True)` which will get the "to_datetime" key out of kwargs if it exists and otherwise return true, and then, in the line where we use kwargs, we instead want 

`dataset = cdf_to_xarray(file_path, to_datetime=to_datetime_arg, **kwargs)`

Simple code to prove this works: 
```
In [10]: def test_func_1(**kwargs):
    ...: 
    ...:     print(kwargs)
    ...: 

In [11]: def test_func_2(**kwargs):
    ...:     print(kwargs)
    ...:     test_func_1(test="test1", **kwargs)
    ...: 

In [12]: test_func_2(test2="test2")
{'test2': 'test2'}
{'test': 'test1', 'test2': 'test2'}
```

Info on kwargs: https://realpython.com/python-kwargs-and-args/

imap_processing/cli.py
maxinelasp: This is an abstract method, meaning it's defining a method but not expecting it to be implemented. So I don't think we're ever expecting it to be defined... I would add an expected return type of `list` and leave the ignore in place.

imap_processing/cli.py
maxinelasp: Oh yeah this is what I meant, agree with Matt here

imap_processing/cli.py
maxinelasp: We might need to change the type of `**kwargs` in the load_cdf function definition to `bool` instead of `dict`.

maxinelasp: I think this is a real error, but it's also kind of annoying to fix, since the kwargs are ultimately passed into xarray_to_cdf which does not use MyPy and therefore doesn't have a defined type. I think you could avoid the issue by giving `**kwargs` in `load_cdf` a type of `object` like in this issue: `https://github.com/python/mypy/issues/6092`

maxinelasp: If neither suggestion works then just leave the ignore errors statements in

maxinelasp: Also, read about kwargs before attempting to understand what I'm saying here: https://realpython.com/python-kwargs-and-args/

imap_processing/cdf/cdf_attribute_manager.py
daralynnrhode: I can `from typing import Any` which can allow me to do `self.variable_attributes: dict[Any]`. Doing this way allows for the `no-any-return` error to still be allowed. 
However, `self.variable_attributes: dict[str, dict]` does work, so long as this is the right format for the dict.

imap_processing/cdf/utils.py
daralynnrhode: It still throws an error for me, maybe because you disallowed the `no-any-return` error from the comment about the `dict[Any]` declaration.

imap_processing/cdf/cdf_attribute_manager.py
daralynnrhode: Worked on my end!

imap_processing/cdf/cdf_attribute_manager.py
daralynnrhode: @maxinelasp and @bourque 
I can import `Any` from typing and then declare the function to return Any which will not trigger any errors.  I can also tell the function to return `yaml` since that is technically what the `_load_yaml_data()` returns that is being called.

imap_processing/cdf/utils.py
daralynnrhode: Thanks! That got it to work.

imap_processing/cdf/utils.py
daralynnrhode: I still get an error on my end...

bourque (approved):
Thanks for making those changes. We are ignoring some errors here but I think they are for valid reasons. So looks good to me!

imap_processing/cdf/utils.py
```diff
@@ -76,8 +77,8 @@ def load_cdf(
     """
     # TODO: remove this when cdflib is updated to version >1.3.0
     if "to_datetime" not in kwargs:
-        kwargs["to_datetime"] = True
-    dataset = cdf_to_xarray(file_path, kwargs)
+        to_datetime_arg = kwargs.get("to_datetime", True)
```
tmplummer: Comment removed. @maxinelasp proved this works fine.

imap_processing/cdf/utils.py
tmplummer: @maxinelasp, just for reference, if the concern was about mutating the input keyword args, the spec says that wont happen.
>  If the form “**identifier” is present, it is initialized to a new ordered mapping receiving any excess keyword arguments, defaulting to a new empty mapping of the same type.

PR #663, opened 2024-06-26
Add mypy Pre-Commit For Ultra Pt.3

daralynnrhode:
# Adding the mypy Pre-commit hook for ultra.
## Overview
Changes for this PR will occur in the `imap_processing/ultra` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

## Updated Files
-imap_processing/ultra/l1b/ultra_l1b.py
-imap_processing/ultra/l1c/ultra_l1c.py
-imap_processing/ultra/l1b/badtimes.py
-imap_processing/ultra/l1b/cullingmask.py
-imap_processing/ultra/l1b/de.py
-imap_processing/ultra/l1b/extendedspin.py
-imap_processing/ultra/l1c/pset.py
-imap_processing/ultra/l0/ultra_utils.py
-imap_processing/ultra/l0/decom_ultra.py
-imap_processing/ultra/utils/ultra_l1_utils.py


imap_processing/ultra/l0/ultra_utils.py
```diff
@@ -12,13 +12,13 @@ class PacketProperties(NamedTuple):
     apid: list  # List of APIDs
     logical_source: list  # List of logical sources
     addition_to_logical_desc: str  # Description of the logical source
-    width: int  # Width of binary data
-    block: int  # Number of values in each block.
+    width: Union[int, None]  # Width of binary data
```
bourque:
```suggestion
    width: Union[int, None]  # Width of binary data (may be None)
```

For these ones with possible `None` types, might be good just to make a note of that. An unsuspecting reader of this might be confused what this `Union[int, None]` means.

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -77,7 +77,8 @@ def append_tof_params(
             data_dict[key].clear()
 
 
-def append_params(decom_data: dict, packet):
+def append_params(decom_data: dict, packet) -> None:  # type: ignore[no-untyped-def]
+    # Todo Update what packet type is.
```
bourque, line 83: I think these `packet` parameter types are `space_packet_parser.parser.Packet`? (You may need to `import space_packet_parser` to get them working too)

bourque: Alternatively you could `from space_packet_parser.parser import Packet` and just call these `Packet` types.

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -122,12 +123,15 @@ def process_ultra_apids(data: list, apid: int):
     sorted_packets = sort_by_time(data, "SHCOARSE")
 
     process_function = strategy_dict.get(apid)
-    decom_data = process_function(sorted_packets, defaultdict(list))
+    decom_data = process_function(sorted_packets, defaultdict(list))  # type: ignore[misc]
+    # ToDo Change, "None" not callable
```
bourque:
```suggestion
    decom_data = process_function(sorted_packets, defaultdict(list))  # type: ignore[misc]
```

I think we can keep this `ignore` in here. This error is confusing, but luckily someone else has encountered this and documented why its happening: https://cdreimer.medium.com/fixing-none-not-callable-error-message-in-mypy-3538c954549b

So it looks like this could be fixed by changing the above line to:

```python
process_function = strategy_dict.get(apid, lambda *args: False)
```

But then this results in an error in the return type 🤷🏻 . I will leave it to you if you want to try to solve that, but this seems to me like a random one-off case that can be ignored.

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -257,7 +261,9 @@ def process_ultra_rates(sorted_packets: list, decom_data: dict):
             ULTRA_RATES.mantissa_bit_length,
         )
 
-        for index in range(ULTRA_RATES.len_array):
+        for index in range(ULTRA_RATES.len_array):  # type: ignore[arg-type]
+            # TODO Change, Argument 1 to "range" has incompatible type
+            # "Optional[int]"; expected "SupportsIndex"
```
bourque:
```suggestion
        for index in range(ULTRA_RATES.len_array):  # type: ignore[arg-type]
```

Another weird corner case that I think can be ignored

imap_processing/ultra/l0/ultra_utils.py
```diff
@@ -316,7 +317,11 @@ def append_ccsds_fields(decom_data: dict, ccsds_data_object: object):
     ccsds_data_object : object
         CCSDS data object.
     """
-    for field in fields(ccsds_data_object.__class__):
+    for field in fields(ccsds_data_object.__class__):  # type: ignore[arg-type]
```
bourque, line 322: Here is a relevant stackoverflow I found about this: https://stackoverflow.com/questions/77071473/where-can-i-import-dataclassinstance-for-mypy-check

I couldn't seem to get this to work. Feel free to try yourself, but if its too much of a hassle I think we can keep the `ignore`

bourque (changes requested):
Great work. There are some tricky corner cases in this one. I have some suggestions for a few fixes, otherwise it looks good!

imap_processing/ultra/l0/ultra_utils.py
daralynnrhode: I think the return error is because of the changed to how process_function was implemented. I can simply change what the function returns to reflect what the error returns.

imap_processing/ultra/l0/decom_ultra.py
daralynnrhode: I think the return error is because of the changed to how process_function was implemented. I can simply change what the function returns to reflect what the error returns.

imap_processing/ultra/l0/ultra_utils.py
daralynnrhode: I was able to successfully import DataclassInstance and resolve the mypy errors occurring. However, I am unsure if the type changes I made will negatively affect the code and would like them double checked.

PR #666, opened 2024-06-26
Add mypy Pre-Commit for swe Pt.4

daralynnrhode:
# Adding the mypy Pre-commit hook for swe.
## Overview
Changes for this PR will occur in the `imap_processing/swe` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

## Updated Files
-imap_processing/swe/l1a/swe_l1a.py
-imap_processing/swe/l1b/swe_l1b.py
-imap_processing/swe/l1a/swe_science.py
-imap_processing/swe/l1b/swe_l1b_science.py
-imap_processing/swe/l0/decom_swe.py
-imap_processing/swe/utils/swe_utils.py


imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -40,7 +40,7 @@
 }
 
 
-def read_lookup_table(table_index_value: int):
+def read_lookup_table(table_index_value: int) -> list:
```
bourque:
```suggestion
def read_lookup_table(table_index_value: int) -> Any:
```

I think this will solve the errors in this function?

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -254,7 +257,8 @@ def populate_full_cycle_data(
             for step in range(180):
                 # Get esa voltage value from esa lookup table and
                 # use that to get row index in full data array
-                esa_voltage_value = esa_lookup_table.loc[esa_step_number]["esa_v"]
+                esa_voltage_value = esa_lookup_table.loc[esa_step_number]["esa_v"]  # type: ignore[attr-defined]
+                # TODO Change,  "list[Any]" has no attribute "loc"
```
bourque:
```suggestion
                esa_voltage_value = esa_lookup_table.loc[esa_step_number]["esa_v"]
```

I think this error will go away with the above fix (switching function to return type `Any`)?

bourque (approved):
Looks great! I think the few errors you are seeing might be fixed with what I suggested, but also happy to keep the `ignores` if they don't.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -117,7 +118,7 @@ def deadtime_correction(counts: np.ndarray, acq_duration: int):
     return corrected_count
 
 
-def convert_counts_to_rate(data: np.ndarray, acq_duration: int):
+def convert_counts_to_rate(data: np.ndarray, acq_duration: float) -> np.ndarray:
```
tech3371: I wanted to leave this comment. We still want that as `int` because input should be integer.
```suggestion
def convert_counts_to_rate(data: np.ndarray, acq_duration: int) -> np.ndarray:
```

tech3371 (commented):
nice work!

imap_processing/swe/l1b/swe_l1b_science.py
daralynnrhode: @tech3371 
Let me know if the change I made will work. 
Mypy is flagging the type declared because the expression could technically return a float so it wants the variable to also match.

daralynnrhode: Or I suppose another variable could be created that is a float and then set equal to the expression. That way `act_duration` is not being set equal to its manipulated self.

tech3371: I think it's ok. Can you add `TODO: check with SWE team` so that I don't forget to check this in upcoming discussion? Otherwise, it looks good to me!

PR #667, opened 2024-06-26
Add mypy Pre-commit for swapi Pt.5

daralynnrhode:
# Adding the mypy Pre-commit hook for swapi.
## Overview
Changes for this PR will occur in the `imap_processing/swapi` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

This PR was pretty straight forward and short-and-sweet :)

## Updated Files
-imap_processing/swapi/l1/swapi_l1.py


bourque (approved):
Nice and easy. Love to see it.

PR #668, opened 2024-06-26
Add Pre-Commit For mag Pt.6

daralynnrhode:
# Adding the mypy Pre-commit hook for mag.
## Overview
Changes for this PR will occur in the `imap_processing/mag` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

## Updated Files
-imap_processing/mag/l1a/mag_l1a.py
-imap_processing/mag/l0/mag_l0_data.py
-imap_processing/mag/l1a/mag_l1a_data.py
-imap_processing/mag/l1c/mag_l1c.py
-imap_processing/mag/l1b/mag_l1b.py


imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -167,7 +167,10 @@ def process_packets(
         # The raw vectors are of type int8, but the output vectors should be at least
         # int16.
         primary_vectors, secondary_vectors = MagL1a.process_vector_data(
-            mag_l0.VECTORS.astype(dtype=np.int32),
+            mag_l0.VECTORS.astype(dtype=np.int32),  # type: ignore[union-attr]
```
maxinelasp, line 167: I'm confused why this line doesn't work... the method is expecting np.ndarray and that's what this is? Any idea why?

maxinelasp (approved):
Thanks for doing this, LGTM! I did have one question but you can merge over it.

imap_processing/mag/l1a/mag_l1a.py
daralynnrhode: @maxinelasp This is occurring because mypy is picky about the attributes of types listed as Unions. As far as I can tell `mag_l0_data.py` has `VECTORS: np.ndarray | str` the `str` portion of the type does not have an `astype` attribute to it. That is why mypy is flagging it.

daralynnrhode: When declaring with Unions mypy expects both types to have the same attributes

maxinelasp: Ohhhhh I understand. Ok, yeah, then this ignore is good, and I will try to make a mental note to fix that next time I'm updating this code

PR #669, opened 2024-06-27
Add mypy Pre-Commit For Lo Pt.7

daralynnrhode:
# Adding the mypy Pre-commit hook for lo.
## Overview
Changes for this PR will occur in the `imap_processing/lo` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

Most of the ignore comments occur around the `data_fields` object and its type. 

## Updated Files
-imap_processing/lo/l0/data_classes/science_counts.py
-imap_processing/lo/l0/data_classes/science_direct_events.py
-imap_processing/lo/l0/data_classes/star_sensor.py
-imap_processing/lo/l0/utils/lo_base.py
-imap_processing/lo/l0/utils/binary_string.py
-imap_processing/lo/l0/utils/bit_decompression.py
-imap_processing/lo/l1a/lo_data_container.py
-imap_processing/lo/l1a/lo_l1a.py
-imap_processing/lo/l1b/lo_l1b.py
-imap_processing/lo/l1c/lo_l1c.py


sdhoyt (approved):
Everything looks good to me. Interesting to see how bad I've been doing with defining types. Thanks for doing all of this!

PR #670, opened 2024-06-27
Add mypy Pre-Commit For idex Pt.8

daralynnrhode:
# Adding the mypy Pre-commit hook for idex.
## Overview
Changes for this PR will occur in the `imap_processing/idex` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

## Updated Files
-imap_processing/idex/idex_packet_parser.py


bourque (approved):
Looks great!

Issue #664, opened 2024-06-26
Add variable attribute inheritance across yaml files to CdfAttributeManager


tmplummer:
### Algorithm Description:
For Hi, I have leveraged YAML inheritance to define default base attribute definitions for each datatype (uint8, uint16, float32, etc). This significantly reduces repeated definitions for VALIDMIN, VALIDMAX, FILLVAL, and FORMAT and allows for a single source definition of variables that are common to multiple data products. Since there is not currently any way to allow dictionary inheritance across YAML files, this means that I need to define all of my variable attributes in a single YAML file and am unable to use the ImapCdfManager class because it expects files to be broken up by data level. I think that it would be useful to implement the ability for inheritance to work across YAML files.

For an example of use case, see:
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
and the general function for instantiating empty xr.DataArray objects:
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/hi/l1c/hi_l1c.py#L182

tech3371:
ENA is already doing this. If loaded in order, you get inheritance.

PR #671, opened 2024-06-27
662 migrate remaining hi products to use cdfattributemanager

tmplummer:
# Change Summary
Migration of Hi L1A hist and de processing to use CdfAttributeManager.
- hist needed total migration including variable attributes and global attributes
- de only needed to migrate the epoch attributes

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
   - add global attributes for hi_l1a_hist products
   - Add hi_l1a_hist variable attributes
   - Reorganize common variable attribute definition for `ccsds_met` and`esa_stepping_num`
   - Change from using hi_cdf_attrs to CdfAttributeManager
   - Modify how counter name is injected into counter variable attributes
   - Use CdfAttributeManager for epoch attributes
   - Update variable name from `esa_step` to `esa_stepping_num`. 

## Testing
The only testing change was to match the renaming of `esa_step` to `esa_stepping_num`

closes 662

### DIFF
--- a/imap_processing/hi/l1a/histogram.py
+++ b/imap_processing/hi/l1a/histogram.py
@@ -4,9 +4,9 @@ import numpy as np
 import xarray as xr
 from space_packet_parser.parser import Packet
 
-from imap_processing.cdf.utils import met_to_j2000ns
 from imap_processing import imap_module_directory
 from imap_processing.cdf.cdf_attribute_manager import CdfAttributeManager
+from imap_processing.cdf.utils import met_to_j2000ns
 
 # TODO: Verify that these names are OK for counter variables in the CDF
 # define the names of the 24 counter arrays
--- a/imap_processing/hi/l1a/histogram.py
+++ b/imap_processing/hi/l1a/histogram.py
@@ -1,14 +1,12 @@
 """Unpack IMAP-Hi histogram data."""
 
-import dataclasses
-
 import numpy as np
 import xarray as xr
 from space_packet_parser.parser import Packet
 
-from imap_processing.cdf.global_attrs import ConstantCoordinates
 from imap_processing.cdf.utils import met_to_j2000ns
-from imap_processing.hi import hi_cdf_attrs
+from imap_processing import imap_module_directory
+from imap_processing.cdf.cdf_attribute_manager import CdfAttributeManager
 
 # TODO: Verify that these names are OK for counter variables in the CDF
 # define the names of the 24 counter arrays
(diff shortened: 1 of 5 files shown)


imap_processing/hi/l1a/histogram.py
```diff
@@ -92,13 +90,16 @@ def allocate_histogram_dataset(num_packets: int) -> xr.Dataset:
     dataset : xarray.Dataset
         Empty xarray.Dataset ready to be filled with packet data.
     """
+    cdf_manager = CdfAttributeManager(imap_module_directory / "cdf" / "config")
+    cdf_manager.load_global_attributes("imap_hi_global_cdf_attrs.yaml")
+    cdf_manager.load_variable_attributes("imap_hi_variable_attrs.yaml")
```
tech3371: @maxinelasp What is main difference between doing this way and other way I have seen others do? Is there benefit of using one over other? Other way I have seen others do is this way:
```
    # create the attribute manager for this data level
    attr_mgr.add_instrument_global_attrs(instrument="lo")
    attr_mgr.add_instrument_variable_attrs(instrument="lo", level="l1a")
```

tech3371 (commented):
Just had one question to Maxine but otherwise, it looks good!

imap_processing/hi/l1a/histogram.py
tmplummer: I can answer that...
Using `ImapCdfAttributes` adds the automatic setting of the config directory as part of the constructor. It also adds convenience functions for adding global and variable attributes by instrument name. For some reason, I thought that would preclude me from using the variable attribute YAML that I have which lumps all l1 variable attributes into a single file so that I can leverage the YAML inheritance and avoid duplicate entries across separate `l1a`, `l1b`, `l1c`. Now that I am thinking a bit more clearly, I should probably do the following:
```suggestion
    attr_mgr = ImapCdfAttributes()
    attr_mgr.add_instrument_global_attrs(instrument="lo")
    attr_mgr.load_variable_attributes("imap_hi_variable_attrs.yaml")
```

I wrote up a feature request to add inheritance across yaml files: #664
This would allow me to follow what other instruments are doing. I just find it very useful to be able to define defaults for each dtype variable and have consistent, not repeated definitions for variables like `epoch`, `ccsds_met`, and any other variables that have the same definition across products.

imap_processing/hi/l1a/histogram.py
```diff
@@ -107,34 +108,35 @@ def allocate_histogram_dataset(num_packets: int) -> xr.Dataset:
...
+    data_vars["esa_stepping_num"] = xr.DataArray(
         np.empty(num_packets, dtype=np.uint8),
         dims=["epoch"],
-        attrs=hi_cdf_attrs.esa_step_attrs.output(),
+        attrs=cdf_manager.get_variable_attributes("hi_hist_esa_stepping_num"),
     )
 
-    for counter in (*QUALIFIED_COUNTERS, *LONG_COUNTERS, *TOTAL_COUNTERS):
-        data_vars[counter] = xr.DataArray(
+    default_counter_attrs = cdf_manager.get_variable_attributes("hi_hist_counters")
```
laspsandoval: Leave a docstring for what are qualified counters, long counters, etc. unless it is somewhere else in the code. I like how you used only one attribute  instead of duplicating these.

laspsandoval (commented):
Nice work. Looks good to me.

tech3371 (approved):

PR #673, opened 2024-06-28
Add mypy Pre-Commit For ialirt Pt.9

daralynnrhode:
# Adding the mypy Pre-commit hook for ialirt.
## Overview
Changes for this PR will occur in the `imap_processing/ialirt` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

***Very few changes were made in only one file. It is very short. 

## Updated Files
-imap_processing/ialirt/l0/decom_ialirt.py


sdhoyt (approved):
Everything looks good to me

PR #675, opened 2024-06-28
[WIP] Refactor pytest fixtures into plugins directory

tmplummer:
# Note:
*The easiest way to look through this PR is to start with the changes to `imap_processing/tests/conftest.py` then look at new files in the `imap_processing/tests/plugins` directory. Other changes are changes to Hi tests to use the new fixtures.*

# Change Summary
Move common and hi fixtures into plugins directory
Add pytest_plugins definition to root conftest.py
Refactor hi test_data structure
Add fixtures for hi test data

## Overview
This is a proposal for a refactor of how we organize pytest fixtures. The proposal is that we move away from distributed conftest.py files which I find to be challenging to understand as more and more fixtures are added. This centralizes all fixtures that are not specific to a single test file and organizes them by instrument. Pleas let me know what you think about these changes.

For this WIP PR, I moved common and hi fixtures only to demonstrate what the refactor would look like.

## New Dependencies
Non

## New Files
- imap_processing/tests/plugins/__init__.py
- imap_processing/tests/plugins/common_fixtures.py
- imap_processing/tests/plugins/hi_fixtures.py

## Deleted Files
None, some files were moved

## Updated Files
   - Added definition of pytest_plugins pointing to common tests/plugins files
- imap_processing/tests/hi/test_decom.py
   - Use new fixtures for test data location
   - Use new fixtures for test data location
   - Use new fixtures for test data location

## Testing
Many changes to the structure of fixtures


greglucas (commented):
:+1: to breaking these fixtures up into more specific subfolders.

Is there a reason to use `plugins` specifically? I thought plugins were generally for sharing across packages so other external projects could make use of the fixtures.

I've generally seen this written more as nested `conftest.py`s, and I think that would also prevent the `hi_test_data_path` from being available to all tests, and only within `hi` tests instead. Right now, the plugins are imported in the root `conftest.py` meaning they will be available to all tests and then the import order likely matters if multiple people define `ccsds_header_data` differently and want it available to their module tests.

```
tests/
  conftest.py  # global fixtures that everyone uses
  hi/
    conftest.py # hi-specific fixtures (only available in this module)
  swe/
    conftest.py  # swe-specific fixtures
```

tmplummer:
> Is there a reason to use `plugins` specifically? I thought plugins were generally for sharing across packages so other external projects could make use of the fixtures.
> 
> I've generally seen this written more as nested `conftest.py`s, and I think that would also prevent the `hi_test_data_path` from being available to all tests, and only within `hi` tests instead. Right now, the plugins are imported in the root `conftest.py` meaning they will be available to all tests and then the import order likely matters if multiple people define `ccsds_header_data` differently and want it available to their module tests.

I personally dislike the organization of fixtures into 10 or more different `conftest.py` files. I find it hard to find fixtures and determine what fixtures are affecting a test, especially for `autouse=True` fixtures. I also think that it discourages reuse of fixtures because everyone writes their own fixture in their own conftest.py. I do see your point about fixture scope and possible collision of names.

On EMUS, we had some troubles with having different levels of conflicting fixtures which was solved by centralizing all shared fixtures into one main location. Perhaps this is less of a concern for `imap_processing` because we don't need fixtures to deal with DB connections or AWS resources.

greglucas:
This could just be my personal preference and experience working with nested `conftest.py` so I prefer that style. It would be good to get others' preferences/opinions too though as I may be skewed.

Here is a good thread explaining some of the nested conftest uses: https://stackoverflow.com/questions/34466027/what-is-conftest-py-for-in-pytest

> I personally dislike the organization of fixtures into 10 or more different conftest.py files. I find it hard to find fixtures and determine what fixtures are affecting a test

Using plugins would lead to the same thing though I think, it's just the files are called `hi_fixtures` in a single root folder rather than `conftest.py` located in the `hi/` directory. I still wouldn't know which plugin file my fixture came from, I could take a guess, but I don't know for sure. This is sort of an unfortunate limitation of using fixtures, that I can't click into them with my IDE to see where the implementation is coming from (or at least I haven't been able to and this seems like it would be tricky to implement).

> especially for autouse=True fixtures

I actually think that we should rarely be using `autouse=True` because I agree with you that it is hard to see what was happening with the action-at-a-distance stuff when this is the case. The exception is with our current autouse fixtures to muck with temporary directories and environment variables where we want those to isolate everyone without needing to think about it.

> On EMUS, we had some troubles with having different levels of conflicting fixtures which was solved by centralizing all shared fixtures into one main location. Perhaps this is less of a concern for imap_processing because we don't need fixtures to deal with DB connections or AWS resources.

Do you mean you just made a large `conftest.py` with all of the fixtures?

In our AWS CDK repo, we do split the DB tests into a separate folder from the AWS resources tests, so they have different `conftest.py` files with different fixtures associated with those.

tmplummer:
> Do you mean you just made a large `conftest.py` with all of the fixtures?

No, we did something similar to what I did in this PR organizing fixtures by aws, data path, database, and spice fixtures. I grok the organization and scope of conftest.py files but dislike the strict naming requirement and inability to organize fixtures in the case where you have lots of fixtures that need to be at the same level.

> Using plugins would lead to the same thing though I think, it's just the files are called hi_fixtures in a single root folder rather than conftest.py located in the hi/ directory. I still wouldn't know which plugin file my fixture came from, I could take a guess, but I don't know for sure.

Being able to determine a unique name for the file that the fixture is in would allow us to follow a convention such as: Each fixture name should start with the name of the file it is located in.

> I actually think that we should rarely be using autouse=True because I agree with you that it is hard to see what was happening with the action-at-a-distance stuff when this is the case. 

Totally agree with this. There may eventually be a need for one additional `autouse=True` fixture that handles setting up SPICE depending on all the decisions that we make on how to use SPICE.

greglucas:
Did anyone else review this or give you feedback/preferences? I'd leave this open if you're still looking for other comments.

tmplummer:
@greglucas 
> Did anyone else review this or give you feedback/preferences? I'd leave this open if you're still looking for other comments.

No other feedback was given. I'm going to cherry pick some of the Hi specific changes but keep with the standard use of `conftest.py` files. If we find a need for clearer fixture organization as our test suite grows, it is an easy change to implement.

PR #676, opened 2024-07-01
Add mypy Pre-Commit for glows Pt.11

daralynnrhode:
# Adding the mypy Pre-commit hook for glows.
## Overview
Changes for this PR will occur in the `imap_processing/glows` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

The places where inline ignore comments are will be changed/fixed before the PR is merged unless for special instances.

## Updated Files
-imap_processing/glows/l1a/glows_l1a_data.py
-imap_processing/glows/l0/glows_l0_data.py
-imap_processing/glows/l1b/glows_l1b_data.py
-imap_processing/glows/utils/constants.py


maxinelasp (approved):
Looks great, thanks for doing that and for the patience around my weird dataclass field types!

PR #678, opened 2024-07-01
Add mypy Pre-Commit for codice Pt.12 Last

daralynnrhode:
# Adding the mypy Pre-commit hook for codice.
## Overview
Changes for this PR will occur in the `imap_processing/codice` file . Applying the mypy pre-commit and making any necessary changes to then pass the checks. 

The places where inline ignore comments are will be changed/fixed before the PR is merged unless for special instances.
There is only one inline ignore comment and that is within `imap_processing/codice/codice_l1a.py`.

## Updated Files
-imap_processing/codice/utils.py
-imap_processing/codice/codice_l0.py
-imap_processing/codice/decompress.py
-imap_processing/codice/codice_l1b.py
-imap_processing/codice/codice_l1a.py


imap_processing/codice/codice_l1a.py
```diff
@@ -324,7 +327,7 @@ def get_params(packet) -> tuple[int, int, int, int]:
     return table_id, plan_id, plan_step, view_id
 
 
-def process_codice_l1a(file_path: Path | str, data_version: str) -> xr.Dataset:
+def process_codice_l1a(file_path: Path, data_version: str) -> xr.Dataset:
```
sdhoyt, line 330: Is this intentionally removing `str` as a possible type?

imap_processing/codice/decompress.py
```diff
@@ -74,13 +74,13 @@ def _apply_lossy_b(compressed_value: int) -> int:
     return LOSSY_B_TABLE[compressed_value]
 
 
-def _apply_lzma_lossless(compressed_value: int) -> int:
+def _apply_lzma_lossless(compressed_value: bytes) -> int:
```
sdhoyt: Looking at the unit tests, it looks like this should allow both `bytes` and `int`

sdhoyt (commented):
Great work! Just a couple comments

imap_processing/codice/codice_l1a.py
daralynnrhode: @sdhoyt 
Yes, because the `decom_packets(file_path)` function that is called is only specified to take a `Path` type. If the decom function is changed to take both `Path` and `str` then you run into the problem of having a union attribute error which would  need to be ignored for all instances (about 8). Essentially mypy would be expecting both types of the `Union`, Path and str in this case, to have the same attributes but they do not. So either only a Path is passed in or code would need to be added to insure that all strings passed in are then converted to Paths.

imap_processing/codice/decompress.py
daralynnrhode: @sdhoyt 
I had to switch around types for this one because the `lama.decompress()` only takes objects that are `bytes`. If the test cases take `int`'s then to satisfy mypy the int would need to be converted to a byte then passed to the decompress function. 
Should I add a type-checker to change `int` to `byte` if it is passed in? That way the code honors the tests

sdhoyt: I might be missing something, but it looks like decompress expects type `int` for `compressed_value`:

```python
decompress(compressed_value: int, algorithm: IntEnum) -> int:
```

daralynnrhode: @sdhoyt 
Oh, I Understand where the confusion is. `decompress()` is different from `lzma.decompress()`( the previous message got autocorrected). From what I can tell everything expects an `int` to be passed in and passed out. However in `_apply_lzma_lossless()` the `lzma.decompress()` is a function from the lzma module and needs a `bytes` type. 
`_apply_lzma_lossless()` is called within the created `decompress()` function.

Since all the other functions that `compressed_value` will be passed into before or has the option of being passed into only take `int` types, then the type for this case should be int. That means a conversion will need to occur from `int` to `bytes`. 

From what I can tell Matthew understood that the `lzma.decompress()` would return a `byte` so he had it converted to an `int` in order to return the correct type. He just didn't realize that the passed-in `int` needed to be converted to a `byte` before utilizing the function.

daralynnrhode: I went ahead and pushed what I feel it all should look like. Let me know what you think!

sdhoyt: Ah gotcha. Sounds good

sdhoyt (approved):
Looks good to me. Thanks for looking into those questions

PR #679, opened 2024-07-01
Update Management of HIT CDF Attributes to use YAMLs

vmartinez-cu:
# Change Summary

## Overview
Update HIT to use yaml files and cdf attribute manager. Added yaml files and updated l1a and l1b
processing code to use the new process for handling cdf attributes

## New Dependencies
None

## New Files

   - defines global attributes for hit
- imap_hit_l1a_variable_attrs.yaml
  - defines l1a variable attributes for hit
- imap_hit_l1b_variable_attrs.yaml
  - defines l1b variable attributes for hit

## Deleted Files
None

## Updated Files
- hit_l1a.py
   - Added cdf attribute manager
   - Added loop to create a list of dims for variables based on their DEPEND_i attribute fields. L1b processing code already had this, so now both are consistent
- hit_l1b.py
   - Added cdf attribute manager

## Testing
Updated hit l1a and l1b unit tests to reflect changes described above.

### DIFF
--- a/imap_processing/cdf/config/imap_hit_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hit_l1a_variable_attrs.yaml
@@ -1,0 +1,494 @@
+# <=== Defaults ===>
+# Assumed values for all variable attrs unless overwritten
+default_attrs: &default
+  DEPEND_0: epoch
+  DISPLAY_TYPE: time_series
+  FILLVAL: -9223372036854775808
+  FORMAT: I12
+  VALIDMIN: 0
+  VALIDMAX: 9223372036854775807
+  VAR_TYPE: data
+  UNITS: ' '
+  SCALETYP: linear
+
+hk_support_attrs: &support_default
+  <<: *default
+  DISPLAY_TYPE: no_plot
+  VAR_TYPE: ignore_data
+
+# <=== Coordinates ===>
+epoch:
+  CATDESC: Time, number of nanoseconds since J2000 with leap seconds included
+  FIELDNAM: epoch
+  LABLAXIS: epoch
+  FILLVAL: -9223372036854775808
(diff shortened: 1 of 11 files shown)


vmartinez-cu:
pre-commit.ci autofix

maxinelasp (approved):
Nice, I like that you're passing around the attribute manager instance to your functions

imap_processing/cdf/config/imap_hit_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,500 @@
...
+  LABLAXIS: Spin period long
+  FORMAT: I5
+
+leak_i:
+  <<: *support_default
+  DEPEND_1: adc_channels
+  CATDESC: Leakage Current [I]
+  FIELDNAM: Leakage Current [I]
+  LABLAXIS: Current I
+  labl_ptr: adc_channels
```
sdhoyt: Maybe I missed an update since I did this, but for the label ptr attribute in leak_i you should have:

```yaml
LABEL_PTR_1: adc_channels_label
```

where adc_channels_label is:
```yaml
adc_channels_label:
  CATDESC: Analog Digital Converter Channels
  FIELDNAM: ADC Channels
  FORMAT: A2
  VAR_TYPE: metadata
```

sdhoyt (commented):
Looks great! Just had one comment

imap_processing/cdf/config/imap_hit_l1a_variable_attrs.yaml
sdhoyt: Also, the label attribute will need to be added as a data array in the dataset

sdhoyt (commented):
The changes to the L1A YAML and the dataset look good! Looks like you just need to make the same updates in L1B

greglucas (approved):

PR #680, opened 2024-07-02
XTCE generator scripts

### LINKED ISSUE #576: Improvements to XTCE generator script(s)
We have scripts to generate XTCE-formatted XML files for each instrument (example [here](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/tools/xtce_generation/xtce_generator_codice.py)). These are used to generate the ["packet definition files"](https://github.com/IMAP-Science-Operations-Center/imap_processing/tree/dev/imap_processing/codice/packet_definitions) which are used to decommutate packet data from the spacecraft.

There are some improvements we could make to these scripts:
- Make a single, generalized script that can process any instrument, instead of having a script for each instrument. This could be done by passing in command line argument(s) for the instrument to process, the path to the user's telemetry spreadsheet, the path to which to write the packet definition file, etc.
- Improve the information that gets printed to the terminal for the user. Currently it is pretty bare bones.
- Add appropriate unit tests

anamanica:
# Change Summary

## Overview
Updated the xtce_generatpr_template.py to use CLI for any instrument.

## New Dependencies
None

## New Files
- test_xtce_generator_template
    - Wrote a few quick unit tests to ensure CLI is working.

## Deleted Files
None

## Updated Files
- xtce_generator_template
    - Finished implementation to take CLI and generate an xtce file.
- pyproject.toml
    - Added imap_xtce as a tool

## Testing
Three unit tests present in test_xtce_generator_template. Fairly general, but effective.

closes #576


tools/xtce_generation/xtce_generator_template.py
```diff
@@ -1,39 +1,91 @@
...
+    )
+    filename_help = "Provide filename to write packets to."
+    packets_help = (
+        "Provide packet dictionary using packet_name, and app_id."
+        '{"<packet_name>": <app_id>}'
+    )
+
+    parser = argparse.ArgumentParser(prog="imap_xtce", description=description)
+    parser.add_argument("--instrument", type=str, required=True, help=instrument_help)
+    parser.add_argument("--filename", type=str, required=True, help=filename_help)
```
maxinelasp: Could you rename this to "file_name"?

tools/xtce_generation/xtce_generator_template.py
```diff
@@ -1,39 +1,91 @@
...
-    #     "P_COD_HI_PHA": 1169,
-    #     "P_COD_LO_PHA": 1153,
-    #     "P_COD_LO_NSW_SPECIES_COUNTS": 1157,
-    #     "P_COD_HI_OMNI_SPECIES_COUNTS": 1172,
-    # }
-
-    packets = {
-        "<packet_name>": "<app_id_number>",
-    }
+    path_to_excel_file = f"{current_directory}/{args.filename}"
```
maxinelasp: Use `pathlib` here since windows uses `\`

tools/xtce_generation/xtce_generator_template.py
```diff
@@ -1,39 +1,91 @@
...
+    ----------
+    args : argparse.Namespace
+        An object containing the parsed arguments and their values.
+    """
+    if args.instrument not in imap_data_access.VALID_INSTRUMENTS:
+        raise ValueError(
+            f"{args.instrument} is not in the supported instrument list: "
+            f"{imap_data_access.VALID_INSTRUMENTS}"
+        )
+
```
maxinelasp, line 73: Could you also validate that the file exists?

maxinelasp (commented):
Nice, looking great. The tests look awesome!

tools/xtce_generation/xtce_generator_template.py
```diff
@@ -1,47 +1,105 @@
...
+    # Parse arguments, and validate instrument
+    args = _parse_args()
+    _validate_args(args)
+
+    instrument_name = args.instrument
+    current_directory = Path(__file__).parent.parent.parent
+    module_path = current_directory / "imap_processing"
+    packet_definition_path = module_path / instrument_name / "packet_definitions"
+    path_to_excel_file = (
```
sdhoyt: Interested to hear if @maxinelasp agrees, but I'd prefer the command argument be the full path to the excel file rather than needing to put the excel file in the xtce_generation folder. I don't think we want to store the excel sheets in the repo so it's an extra step to copy one in there to run this command.

sdhoyt (commented):
Great work on this!

tools/xtce_generation/xtce_generator_template.py
maxinelasp: Yep, I agree, good call

anamanica: Good stuff! I'll get that implemented, and then request another review.

imap_processing/tests/hi/test_hi_l1b.py
greglucas, line 1: I think this was fixed in https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/681 so shouldn't be added back in here.

tools/tests/test_xtce_generator_template.py
```diff
@@ -0,0 +1,79 @@
...
+# Testing invalid instrument
+@mock.patch("tools.xtce_generation.xtce_generator_template.main")
+def test_main_inval_instr(mock_input):
+    """Testing with invalid instrument."""
+    test_args = [
+        "test_script",
+        "--instrument",
+        "ana",
+        "--file_path",
+        "/Users/anma6676/Desktop/Repositories/imap_processing/tools/xtce_generation/TLM_SWP_20231006-121021.xlsx",
```
greglucas: I'd suggest removing your specific folder structure from tests if possible. That way the bots won't learn who you are :)
```suggestion
        "tools/xtce_generation/TLM_SWP_20231006-121021.xlsx",
```

tools/xtce_generation/xtce_generator_template.py
```diff
@@ -1,47 +1,104 @@
...
-"""
+"""XTCE generator script that uses CLI for generalized format."""
 
+import argparse
+import json
 from pathlib import Path
 
+import imap_data_access
+
 from tools.xtce_generation.telemetry_generator import TelemetryGenerator
```
greglucas, line 9: I am not sure that this will work as an external script, were you able to verify that this works from other directories? We aren't installing the tools folder, so this currently has to be referenced directly from the folder structure I think.

tools/xtce_generation/xtce_generator_template.py
```diff
@@ -1,47 +1,104 @@
...
+    )
+    file_path_help = "Provide full file path to write packets to."
+    packets_help = (
+        "Provide packet dictionary using packet_name, and app_id."
+        '{"<packet_name>": <app_id>}'
+    )
+
+    parser = argparse.ArgumentParser(prog="imap_xtce", description=description)
+    parser.add_argument("--instrument", type=str, required=True, help=instrument_help)
+    parser.add_argument("--file_path", type=str, required=True, help=file_path_help)
```
greglucas: It should be hyphens for the CLI input to be consistent with our other CLIs. That then gets transformed to underscores within the Python code for you.
```suggestion
    parser.add_argument("--file-path", type=str, required=True, help=file_path_help)
```

tools/xtce_generation/xtce_generator_template.py
```diff
@@ -1,47 +1,104 @@
...
+    """Generate xtce file from CLI information given."""
+
+    # Parse arguments, and validate instrument
+    args = _parse_args()
+    _validate_args(args)
+
+    instrument_name = args.instrument
+    current_directory = Path(__file__).parent.parent.parent
+    module_path = current_directory / "imap_processing"
+    packet_definition_path = module_path / instrument_name / "packet_definitions"
```
greglucas, line 93: I was confused at first where the output was going, could you add a note in the main docstring about writing the file to the specific instrument subfolder?

greglucas (commented):
Just a note that I have a proposal to rework some of the underlying XTCE code: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/654 if you wanted to review that as well. I think this PR will slot in nicely with it to enable a command line interface for that, but it might remove some of the arguments here (packets for instance).

maxinelasp (commented):
Looks great to me, once you address Greg's comments which I agree with :)

sdhoyt (commented):
Looks good to me! Thanks for making that update. Ready to approve once Greg's comments are addressed.

imap_processing/tests/hi/test_hi_l1b.py
anamanica: Haha, yes! sorry about that.

tools/xtce_generation/xtce_generator_template.py
anamanica: @greglucas That makes sense! In the external libraries, `imap-data-access` is loaded in, and this is a line of code I totally stole from `cli.py`. Would that make this okay to keep as is?

I tried to download the imap-xtce in a different directory to test it, but was unable to. Whether or not this is user error or actual error I am unsure, so I reached out to the students slack-chat.

greglucas: Thanks for looking into this!

I think this is only an issue here because it is in the "tools" area, which is an extra portion and not something that gets added to the distribution when we type `pip install mypackage`. The way around this would be to inline `TelemetryGenerator`, or put this CLI work into the same file as that so there are no relative imports required.

I wouldn't bother with this. We can know about the limitation and this is an improvement 👍 . That can be added/updated as an extra feature later.

greglucas (approved):

PR #684, opened 2024-07-02
682 imap hi use imapcdfattributes instead of cdfattributemanager

### LINKED ISSUE #682: IMAP Hi - use ImapCdfAttributes instead of CdfAttributeManager
### Algorithm Description:
It was pointed out in PR 671 that I could use the IMAP specific class for CDF attribute management. This ticket is to migrate the remaining Hi uses of the general CdfAttributeManager class to ImapCdfAttributes.

tmplummer:
# Change Summary
This is a small change that replaces instances where I was using CdfAttributManager rather than ImapCdfAttributes.

## Updated Files
   - Replace CdfAttributeManager with ImapCdfAttributes
   - Replace CdfAttributeManager with ImapCdfAttributes
   - Replace CdfAttributeManager with ImapCdfAttributes

Closes #682

### DIFF
--- a/imap_processing/hi/l1b/hi_l1b.py
+++ b/imap_processing/hi/l1b/hi_l1b.py
@@ -6,14 +6,14 @@ import numpy as np
 import xarray as xr
 
 from imap_processing import imap_module_directory
-from imap_processing.cdf.cdf_attribute_manager import CdfAttributeManager
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.hi.utils import HIAPID
 from imap_processing.utils import convert_raw_to_eu
 
 logger = logging.getLogger(__name__)
-CDF_MANAGER = CdfAttributeManager(imap_module_directory / "cdf" / "config")
-CDF_MANAGER.load_global_attributes("imap_hi_global_cdf_attrs.yaml")
-CDF_MANAGER.load_variable_attributes("imap_hi_variable_attrs.yaml")
+ATTR_MGR = ImapCdfAttributes()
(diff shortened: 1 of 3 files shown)


bourque (approved):
LGTM!

imap_processing/hi/l1b/hi_l1b.py
```diff
@@ -6,14 +6,14 @@
...
-from imap_processing.cdf.cdf_attribute_manager import CdfAttributeManager
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.hi.utils import HIAPID
 from imap_processing.utils import convert_raw_to_eu
 
 logger = logging.getLogger(__name__)
-CDF_MANAGER = CdfAttributeManager(imap_module_directory / "cdf" / "config")
-CDF_MANAGER.load_global_attributes("imap_hi_global_cdf_attrs.yaml")
-CDF_MANAGER.load_variable_attributes("imap_hi_variable_attrs.yaml")
+ATTR_MGR = ImapCdfAttributes()
```
tech3371, line 14: above was lower case. Can we lower case this also to be consistent?

tech3371 (approved):

imap_processing/hi/l1b/hi_l1b.py
tmplummer: I implemented this attribute manager as a global attribute in this file and thus followed the convention of using an all uppercase variable name. The reason I did this differently in this file was to avoid instantiating a separate instance in the two functions that use the same attribute manager.

PR #686, opened 2024-07-03
Ultra l1b extended events

laspsandoval:
# Change Summary

## Overview
All of l1b de except any SPICE utilization. Note that I still don't have recent lookup tables so I will update this code once I have them.

## New Files
lookup_utils.py
- Contains tools for lookup tables for l1b.

ultra_l1b_extended.py
- Calculates extended raw events for Ultra l1b.

/lookup_tables (5 files)

imap_processing/tests/ultra/test_data/l0/FM45_40P_Phi28p5_BeamCal_LinearScan_phi28.50_theta-0.00_Ultra_Image_Raw_Event_20240207T102746_withFSWcalcs.csv 
- test data

## Updated Files
imap_ultra_l1b_variable_attrs.yaml
- Added and improved variable descriptions.

conftest.py
- Added some test data.

de.py
- Added direct event data.

## Testing

test_lookup_utils.py
- Tests lookup_utils.py

test_ultra_l1b_extended.py
- Tests ultra_l1b_extended.py

test_ultra_l1b.py
- Tests ultra_l1b.py

## Environment
poetry.lock
pyproject.toml


imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -26,113 +26,157 @@ x_front:
   CATDESC: x front position
   FIELDNAM: Event x-position on front foil
   LABLAXIS: x front position
-  # TODO: come back to format
-  UNITS: mm
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: hundredths of a millimeter
```
tmplummer: SPDF says the following about the UNITS attribute:

> is a character string (no more than 20 characters, but preferably 6 characters) representing the units of the variable

Is there any way to shorten these units? e.g. "m * 10^-5" or "mm / 100"

imap_processing/ultra/l1b/de.py
```diff
@@ -1,7 +1,22 @@
-"""Calculate Annotated Direct Events."""
+"""Calculate Direct Events."""
 
 import numpy as np
 
+from imap_processing.cdf.defaults import GlobalConstants
```
tmplummer: Is `imap_processing.cdf.defaults` going to go away once all CDF generation is migrated to use `ImapCdfAttributes`?

imap_processing/ultra/l1b/de.py
```diff
@@ -23,35 +38,113 @@ def calculate_de(de_dataset, name):
...
+    ssd_r = get_path_length((xf[ssd_indices], ssd_yf), (ssd_xb, ssd_yb), ssd_d)
+    ssd_ctof, ssd_bin = determine_species_ssd(ssd_energy, ssd_tof, ssd_r)
+
+    # Combine ph_yb and ssd_yb along with their indices
+    combined_indices = np.concatenate((ph_indices, ssd_indices))
+
+    de_dict["x_front"] = xf
+
+    yb = np.concatenate((ph_yb, ssd_yb))
+    de_dict["y_back"] = yb[np.argsort(combined_indices)]
```
tmplummer: It looks like you could store the results from `np.argsort(combined_indices)` and use that instead of having numpy perform the sorting algorithm so many times.

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -0,0 +1,158 @@
...
+    ----------
+    dy_lut : np.ndarray
+        Change in y direction used for the lookup table (mm).
+
+    Returns
+    -------
+    yadj : np.ndarray
+        Y adjustment (mm).
+    """
+    yadjust_path = f"{base_path}/yadjust.csv"
```
tmplummer: Specifying the path this way is not platform agnostic. I suggest the following equivalent change for all of these functions that read csv files.
```suggestion
    yadjust_path = base_path / "yadjust.csv"
```

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -0,0 +1,158 @@
+"""Contains tools for lookup tables for l1b."""
+
+import numpy as np
+import pandas as pd
+
+from imap_processing import imap_module_directory
+
+base_path = f"{imap_module_directory}/ultra/lookup_tables"
```
tmplummer: Keeping paths as `Path` objects is platform agnostic.
```suggestion
base_path = imap_module_directory / "ultra" / "lookup_tables"
```

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -0,0 +1,158 @@
...
+
+def get_norm(dn: np.ndarray, key: str, file_label: str):
+    """
+    Correct mismatches between the stop Time to Digital Converters (TDCs).
+
+    There are mismatches between the stop TDCs, i.e., SpN, SpS, SpE, and SpW.
+    Before these can be used, they must be corrected, or normalized,
+    using lookup tables.
+
+    Further description is available on pages 31-32 of the IMAP-Ultra Flight Software
```
tmplummer, line 45: Nice documentation of where to find more info.

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -0,0 +1,243 @@
...
+    get_path_length,
+    get_ph_tof_and_back_positions,
+    get_ssd_offset_and_positions,
+    get_ssd_tof,
+)
+from imap_processing.utils import group_by_apid
+
+
+@pytest.fixture()
+def de_dataset(ccsds_path_theta_0, xtce_path):
```
tmplummer: Is this fixture the same as in test_ultra_l1b.py? If so, it can be moved into a conftest.py and scope set to "sesson" so that the output is cached and not regenerated each test it is used in.

tmplummer (commented):
Great to see some algorithm code being implemented.

This is a very large PR. I was unable to to a thorough review but provided what feedback I could.

imap_processing/ultra/l1b/de.py
laspsandoval: Nice catch. I'll probably model the attributes after what Maxine is doing in her PR. I put it as a TODO for now.

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -0,0 +1,192 @@
...
+        de_dataset, df_filt.Xf.values.astype("float"), "ultra45"
+    )
+
+    index = np.where((df_filt["CoinType"] == 1) | (df_filt["CoinType"] == 2))[0]
+
+    # TODO: This is as close as I can get.
+    #  I suspect that the lookup table is different.
+    test_xc = df_filt["Xc"].iloc[index].astype("float")
+    etof, xc = get_coincidence_positions(de_dataset, tof, "ultra45")
+    assert xc == pytest.approx(test_xc.values, rel=1)
```
greglucas, line 104: Here and elsewhere in the tests it would be helpful to use the numpy testing utilities for array comparisons.
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html

```python
np.testing.assert_allclose(xc, test_xc.values, rtol=...)
```

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -0,0 +1,192 @@
...
+
+    tof, ssd = get_ssd_tof(de_dataset, xf)
+
+    ssd_indices = np.where(de_dataset["STOP_TYPE"] >= 8)[0]
+
+    energy = get_energy_ssd(de_dataset, ssd)
+    test_energy = df_filt["Energy"].iloc[ssd_indices].astype("float")
+
+    # TODO: the first value doesn't match. Look into this.
+    assert np.array_equal(test_energy[1::], energy[1::].astype(float))
```
greglucas, line 181: This seems like it should have an xfail/skip associated with it. A TODO in the tests doesn't seem ideal as I'm not sure we'll get back to it, so it could get lost in time.

```suggestion
    # TODO: the first value doesn't match. Look into this.
    np.testing.assert_array_equal(test_energy[1::], energy[1::].astype(float))
```

imap_processing/ultra/l1b/de.py
```diff
@@ -24,35 +41,113 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
-    de_dict["vz_dps_helio"] = np.zeros(len(epoch), dtype=np.float64)
-    de_dict["eventtimes"] = np.zeros(len(epoch), dtype=np.float64)
+    xf = get_front_x_position(
+        de_dataset["START_TYPE"].data,
+        de_dataset["START_POS_TDC"].data,
+    )
+
+    # Pulse height
+    ph_indices = np.where(
+        (de_dataset["STOP_TYPE"] == 1) | (de_dataset["STOP_TYPE"] == 2)
```
greglucas, line 55: Should STOP_TYPE be an Enum? What do 1 and 2 mean here?

imap_processing/ultra/l1b/de.py
```diff
@@ -24,35 +41,113 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    )[0]
```
greglucas, line 56: You only want the first occurrence, or do you want all sets of indices matching this condition?
Numpy also suggests using `nonzero()` instead to get the indices. https://numpy.org/doc/stable/reference/generated/numpy.where.html

```python
ph_indices = np.nonzero(de_dataset["STOP_TYPE"] == 1) | (de_dataset["STOP_TYPE"] == 2)
```

imap_processing/ultra/l1b/de.py
```diff
@@ -24,35 +41,113 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    ph_d, ph_yf = get_front_y_position(de_dataset["START_TYPE"].data[ph_indices], ph_yb)
+    ph_energy = get_energy_pulse_height(
+        de_dataset["STOP_TYPE"].data[ph_indices], ph_xb, ph_yb
+    )
+    ph_r = get_path_length((xf[ph_indices], ph_yf), (ph_xb, ph_yb), ph_d)
+    ph_ctof, ph_bin = determine_species_pulse_height(ph_energy, ph_tof, ph_r)
+    ph_etof, ph_xc = get_coincidence_positions(de_dataset, ph_tof, "ultra45")
+
+    # SSD
+    ssd_indices = np.where(de_dataset["STOP_TYPE"] >= 8)[0]
```
greglucas, line 69: Suggest changing all "where" to nonzero and making sure you only want the first occurrence or if you want all of the indices. (indices to me means there could be multiples whereas you're only grabbing the first ssd_index currently)
```suggestion
    ssd_indices = np.nonzero(de_dataset["STOP_TYPE"] >= 8)
```

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -0,0 +1,158 @@
...
+    dy_lut : np.ndarray
+        Change in y direction used for the lookup table (mm).
+
+    Returns
+    -------
+    yadj : np.ndarray
+        Y adjustment (mm).
+    """
+    yadjust_path = base_path / "yadjust.csv"
+    yadjust_df = pd.read_csv(yadjust_path).set_index("dYLUT")
```
greglucas, line 30: Do you want to store these as global constants so you don't have to do `pd.read_csv()` multiple times if you call this function several times.

```python
_YADJUST_DF = pd.read_csv(yadjust_path).set_index("dYLUT")
...
```

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -0,0 +1,158 @@
...
+
+    yadj = yadjust_df["dYAdj"].iloc[dy_lut]
+
+    return yadj.values
```
greglucas, line 34: I think you can possibly index directly into the numpy array.
```suggestion
     return yadjust_df["dYAdj"].values[dy_lut]
```

imap_processing/ultra/lookup_tables/FM45_Startup1_ULTRA_IMGPARAMS_20240719.csv
```diff
@@ -0,0 +1,7 @@
...
+,0.172998047,49.3,48.25,0.5,-528,-525,0.001831055,0.067929688,41.75,0.067929688,-39.79492188,0.1,-44.5,-44.5,,,,,,,,,,,,,29.3,37.3,7.1,15.1,-15.1,-7.1,-37.3,-29.3,0.196484375,-6,-7.3,-3.8,-4.2,-3.8,-3.7,-6.3,-5,-5,-6.3,-3.7,-3.8,-4,-4.2,-7.3,-6,5.9,,
+445027647,0.172998047,47.5,48.5,0.5,-528,-525,0.001831055,0.068486328,42.25,0.068486328,-39.75,0.1,-44.5,-44.5,22.6,26.6,22.6,26.6,-40,9,0.666699219,10,0.75,-5,540,540,29.3,37.300000000000004,7.1000000000000005,15.1,-15.1,-7.1,-37.3,-29.3,0.196484375,-6,-7,-3.5,-4,-3.8,-3.5,-6.8,-5.5,-5.5,-6.8,-3.5,-3.8,-4,-3.5,-7,-6,5.300000000000001,50,65
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
```
greglucas, line 7: What are these extra lines for?

pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "poetry_dynamic_versioning.backend"
 [tool.poetry]
 name = "imap-processing"
 # Gets updated dynamically by the poetry-dynamic-versioning plugin
-version = "0.0.0"
+version = "0.3.0.post24.dev0+517786f"
```
greglucas, line 8: This should be left as "0.0.0" and not updated, see the comment above. There is something going wrong with your install if this is getting committed I think.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -0,0 +1,716 @@
...
+    tof : np.array
+        Time of flight (tenths of a nanosecond).
+    t2 : np.array
+        Particle time of flight from start to stop (tenths of a nanosecond).
+    xb : np.array
+        Back positions in x direction (hundredths of a millimeter).
+    yb : np.array
+        Back positions in y direction (hundredths of a millimeter).
+    """
+    indices = np.where(np.isin(de_dataset["STOP_TYPE"], [1, 2]))[0]
```
greglucas, line 57: I would be consistent between the files. Here you are using `isin`, but in the other files you are doing `(STOP_TYPE == 1) | (STOP_TYPE == 2)`

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -0,0 +1,716 @@
...
+    t1 = tofx + tofy  # /2 incorporated into scale
+
+    xb = np.zeros(len(de_dataset["STOP_TYPE"]))
+    yb = np.zeros(len(de_dataset["STOP_TYPE"]))
+
+    # particle_tof (t2) used later to compute etof
+    t2 = np.zeros(len(de_dataset["STOP_TYPE"]))
+    tof = np.zeros(len(de_dataset["STOP_TYPE"]))
+
+    # Stop Type: 1=Top, 2=Bottom
```
greglucas, line 88: Would suggest an Enum for this. `STOP_TYPE.Top, STOP_TYPE.Bottom` or similar

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -0,0 +1,716 @@
...
+    """
+    index_top = np.where(de_dataset["COIN_TYPE"] == 1)[0]
+    index_bottom = np.where(de_dataset["COIN_TYPE"] == 2)[0]
+    index = np.concatenate((index_top, index_bottom))
+
+    # Normalized TDCs
+    # For the stop anode, there are mismatches between the coincidence TDCs,
+    # i.e., CoinN and CoinS. They must be normalized via lookup tables.
+
+    # TpCoinNNorm Top
```
greglucas, line 273: I think there is a lot of duplicated code for the two cases here, so would it make sense to try and factor some of this out? I think you could also subset your dataset instead of keeping track of the indices and using those all the itme.
https://docs.xarray.dev/en/v2024.06.0/generated/xarray.Dataset.where.html

```python
top_ds = de_dataset.where(de_dataset["COIN_TYPE"] == 1)
```

greglucas (commented):
This is too much for me to review in one sitting, but I left some general comments now.

I'd really like to see this broken up into roughly each individual function as a PR, which means this one PR turns into ~15-20. I think this would go much faster and easier for getting feedback on as well.

laspsandoval:
> This is too much for me to review in one sitting, but I left some general comments now.
> 
> I'd really like to see this broken up into roughly each individual function as a PR, which means this one PR turns into ~15-20. I think this would go much faster and easier for getting feedback on as well.

Ok. I'll create separate branches to break it up.

laspsandoval:
Addressed in other PRs.

PR #687, opened 2024-07-08
ENH/MNT: Add a common initial dataset creation routine

greglucas:
# Change Summary

## Overview

This adds a common interface for returning xarray datasets from a packet file. There is one dataset per apid, sorted by time, with derived values automatically converted.

Each dataset is essentially a L0.5, but my hope is that this will be easier for everyone to work with (especially for housekeeping dataset creation). Rather than everyone implementing something fairly similar but slightly different inside their packet iteration routine:
```python
for packet in packets:
    # do a bunch of checks like sorting, apid checks, extraction
```

It would be:

```python
datasets_by_apid = packet_file_to_datasets(packet_file, instrument_definition)
ds = datasets_by_apid[my_instrument.SCIENCE_APID]

... Then later do some work on the dataset
# The dataset is already referenced to our required epoch and is the length of the number of packets in the file with this apid
# This will allow easier iteration as the entire dataset is referenced to the proper dimension of the packet epoch.
ds["my_derirved_variable"] = do_something(ds["my_packet_variable"])
```

I updated SWAPI to use this new function, which simplified things IMO.

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -103,7 +103,7 @@ def load_cdf(
     return dataset
 
 
-def write_cdf(dataset: xr.Dataset) -> Path:
+def write_cdf(dataset: xr.Dataset, **extra_cdf_kwargs: dict) -> Path:
     """
     Write the contents of "data" to a CDF file using cdflib.xarray_to_cdf.
 
@@ -118,6 +118,8 @@ def write_cdf(dataset: xr.Dataset) -> Path:
     ----------
     dataset : xarray.Dataset
         The dataset object to convert to a CDF.
+    **extra_cdf_kwargs : dict
+        Additional keyword arguments to pass to the ``xarray_to_cdf`` function.
 
     Returns
     -------
@@ -166,6 +168,7 @@ def write_cdf(dataset: xr.Dataset) -> Path:
         dataset,
         str(file_path),
         terminate_on_warning=True,
+        **extra_cdf_kwargs,
     )  # Terminate if not ISTP compliant
 
     return file_path
--- a/imap_processing/hi/l1a/histogram.py
+++ b/imap_processing/hi/l1a/histogram.py
@@ -2,10 +2,8 @@
 
 import numpy as np
 import xarray as xr
-from space_packet_parser.parser import Packet
 
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
-from imap_processing.cdf.utils import met_to_j2000ns
 
 # define the names of the 24 counter arrays
 # contained in the histogram packet
@@ -35,30 +33,34 @@ LONG_COUNTERS = (
 TOTAL_COUNTERS = ("total_a", "total_b", "total_c", "fee_de_sent", "fee_de_recd")
 
 
-def create_dataset(packets: list[Packet]) -> xr.Dataset:
+def create_dataset(input_ds: xr.Dataset) -> xr.Dataset:
     """
     Create dataset for a number of Hi Histogram packets.
 
     Parameters
     ----------
-    packets : list[space_packet_parser.ParsedPacket]
-        Packet list.
+    input_ds : xarray.Dataset
+        Dataset of packets.
 
     Returns
     -------
     dataset : xarray.Dataset
         Dataset with all metadata field data in xr.DataArray.
     """
-    dataset = allocate_histogram_dataset(len(packets))
+    dataset = allocate_histogram_dataset(len(input_ds.epoch))
+
+    # TODO: Move into the allocate dataset function?
+    dataset["epoch"].data[:] = input_ds["epoch"].data
+    dataset["ccsds_met"].data = input_ds["ccsds_met"].data
+    dataset["esa_stepping_num"].data = input_ds["esa_step"].data
 
     # unpack the packets data into the Dataset
-    for i_epoch, packet in enumerate(packets):
-        dataset.epoch.data[i_epoch] = met_to_j2000ns(packet.data["CCSDS_MET"].raw_value)
-        dataset.ccsds_met[i_epoch] = packet.data["CCSDS_MET"].raw_value
-        dataset.esa_stepping_num[i_epoch] = packet.data["ESA_STEP"].raw_value
-
+    # (npackets, 24 * 90 * 12)
+    # TODO: Look into avoiding the for-loops below
+    #       It seems like we could try to reshape the arrays and do some numpy
+    #       broadcasting rather than for-loops directly here
+    for i_epoch, counters_binary_data in enumerate(input_ds["counters"].data):
         # unpack 24 arrays of 90 12-bit unsigned integers
-        counters_binary_data = packet.data["COUNTERS"].raw_value
         counter_ints = [
             int(counters_binary_data[i * 12 : (i + 1) * 12], 2) for i in range(90 * 24)
         ]
--- a/imap_processing/swapi/l1/swapi_l1.py
+++ b/imap_processing/swapi/l1/swapi_l1.py
@@ -467,11 +467,10 @@ def process_swapi_science(sci_dataset: xr.Dataset, data_version: str) -> xr.Data
     # ===================================================================
 
     # epoch time. Should be same dimension as number of good sweeps
-    epoch_time = good_sweep_sci["epoch"].data.reshape(total_full_sweeps, 12)[:, 0]
-    # epoch_converted_time = met_to_j2000ns(epoch_time)
-    epoch_converted_time = epoch_time
+    epoch_values = good_sweep_sci["epoch"].data.reshape(total_full_sweeps, 12)[:, 0]
+
     epoch_time = xr.DataArray(
-        epoch_converted_time,
+        epoch_values,
         name="epoch",
         dims=["epoch"],
         attrs=ConstantCoordinates.EPOCH,
--- a/imap_processing/swapi/l1/swapi_l1.py
+++ b/imap_processing/swapi/l1/swapi_l1.py
@@ -53,7 +53,7 @@ def filter_good_data(full_sweep_sci: xr.Dataset) -> np.ndarray:
     # TODO: change comparison to SWAPIMODE.HVSCI once we have
     # some HVSCI data
     # MODE should be HVSCI
-    mode_indices = (mode == SWAPIMODE.HVENG.value).all(axis=1)
+    mode_indices = (mode == SWAPIMODE.HVENG).all(axis=1)
     bad_data_indices = sweep_indices & plan_id_indices & mode_indices
 
     # TODO: add checks for checksum
@@ -527,7 +527,7 @@ def process_swapi_science(sci_dataset: xr.Dataset, data_version: str) -> xr.Data
     # L1 quality flags
     # TODO: Should these be kept in raw format rather than derived into strings?
     dataset["swp_pcem_flags"] = xr.DataArray(
-        np.array(pcem_compression_flags == "RAW", dtype=np.uint8),
+        np.array(pcem_compression_flags, dtype=np.uint8),
         dims=["epoch", "energy"],
         attrs=dataclasses.replace(
             compression_attrs,
@@ -537,7 +537,7 @@ def process_swapi_science(sci_dataset: xr.Dataset, data_version: str) -> xr.Data
         ).output(),
     )
     dataset["swp_scem_flags"] = xr.DataArray(
-        np.array(scem_compression_flags == "RAW", dtype=np.uint8),
+        np.array(scem_compression_flags, dtype=np.uint8),
         dims=["epoch", "energy"],
         attrs=dataclasses.replace(
             compression_attrs,
@@ -547,7 +547,7 @@ def process_swapi_science(sci_dataset: xr.Dataset, data_version: str) -> xr.Data
         ).output(),
     )
     dataset["swp_coin_flags"] = xr.DataArray(
-        np.array(coin_compression_flags == "RAW", dtype=np.uint8),
+        np.array(coin_compression_flags, dtype=np.uint8),
         dims=["epoch", "energy"],
         attrs=dataclasses.replace(
             compression_attrs,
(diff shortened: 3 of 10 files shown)


imap_processing/utils.py
```diff
@@ -254,3 +255,86 @@ def update_epoch_to_datetime(dataset: xr.Dataset) -> xr.Dataset:
...
+                for key, value in packet.data.items():
+                    attribute_dict[apid][key]["short_description"] = (
+                        value.short_description
+                    )
+                    attribute_dict[apid][key]["long_description"] = (
+                        value.long_description
+                    )
+                    # TODO: Additional info from XTCE like units
+
+            # TODO: Include the header too? (packet.header | packet.data)
```
tmplummer: Generally, l1a is close to the unprocessed packet, so I would argue that similar to raw vs. derived value, the header should be included for l1a an not included for l1b.

imap_processing/utils.py
```diff
@@ -254,3 +255,86 @@ def update_epoch_to_datetime(dataset: xr.Dataset) -> xr.Dataset:
...
+        # The time key is always the first key in the data dictionary on IMAP
+        time_key = next(iter(data.keys()))
+        # Convert to J2000 time and use that as our primary dimension
+        time_data = met_to_j2000ns(data[time_key])
+        ds = xr.Dataset(
+            {key.lower(): ("epoch", val) for key, val in data.items()},
+            coords={"epoch": time_data},
+        )
+        for key, value in attribute_dict[apid].items():
+            ds[key.lower()].attrs.update(value)
```
tmplummer: Is the assumption here that all of the required CDF attributes will be defined in the xtce? or would there be another round of updating the variable attributes in the dataset?

imap_processing/utils.py
```diff
@@ -254,3 +255,86 @@ def update_epoch_to_datetime(dataset: xr.Dataset) -> xr.Dataset:
...
+    # {apid1: dataset1, apid2: dataset2, ...}
+    data_dict: dict[int, dict] = dict()
+    attribute_dict: dict[int, dict] = dict()
+
+    # Set up the parser from the input packet definition
+    packet_definition = xtcedef.XtcePacketDefinition(xtce_packet_definition)
+    packet_parser = parser.PacketParser(packet_definition)
+
+    with open(packet_file, "rb") as binary_data:
+        packet_generator = packet_parser.generator(binary_data)
```
tmplummer, line 272: I like that this directly uses the generator rather than converting to a list.

imap_processing/utils.py
```diff
@@ -254,3 +255,86 @@ def update_epoch_to_datetime(dataset: xr.Dataset) -> xr.Dataset:
     dataset = dataset.assign_coords(epoch=epoch)
 
     return dataset
+
+
+def packet_file_to_datasets(
```
tmplummer, line 231: I see quite a bit of overlap with the `create_dataset` function. Some differences are:

- This function creates a dataset for each apid in the input file
- `create_dataset` has an option for including the packet header or not
- `create_dataset` brings in variable attributes from a source external to the xtce. Right now that source has not been updated to use the ImapCdfAttributes class.
- `create_dataset` allows for the option to skip certain fields in the packet.

tmplummer, line 231: It may be useful to optionally skip an apid. The motivation for this is packets with variable length binary data fields.

tmplummer (commented):
Overall, I like this implementation but I think there may need to be some features added in order to be able to replace all usages of the `create_dataset` function. I am also curious to see what it looks like to update variable attributes using the ImapCdfAttributes class.

imap_processing/utils.py
greglucas: OK, I'm happy to do that as well, then we could strip it later.

For the raw vs `derived_value`. Do we not want to use the XTCE functionality of `derived_value` and rather just do that ourselves in the code later? i.e. use the raw_to_eu conversion function instead.

imap_processing/utils.py
greglucas: No, I view this as very much a "just give me an xarray dataset". Then I'll start appending metadata, adding more stuff to it etc... We could leave this out if it'd be easier to understand and use later, but I figured it might be useful for housekeeping datasets.

imap_processing/utils.py
greglucas: I agree, this is heavily based on that function. Would you want all of these capabilities? My thought is that sometimes too many options is not great either and sometimes standardizing on a fixed layout is useful.

I'd actually rather not "skip" the specific fields and complicate this function, but rather just have a user do `dataset.drop("variable")` to remove whatever they want afterwards. So I kind of view this as wanting to give the user everything and then the user can start winnowing down to what they need or transforming from there.

If it would be useful to add some ignore keywords I am open to it though.

greglucas: This handles variable length binary data!

I can see the argument here as well, but again I think that it might be easier for a user to just not access that apid in the dictionary lookup?

imap_processing/utils.py
tmplummer: > For the raw vs `derived_value`. Do we not want to use the XTCE functionality of `derived_value` and rather just do that ourselves in the code later? i.e. use the raw_to_eu conversion function instead.

Good question. I recall a conversation about using the command and telemetry spreadsheet analog conversions to populate the XTCE so that we could essentially skip l1a for housekeeping. Maybe that decision has not been made and just sticking to the raw value is the appropriate thing to do for now?

imap_processing/utils.py
tmplummer: I was really just making an observation about the overlap. I am in agreement that keeping this function simple makes sense.

imap_processing/utils.py
tmplummer: Ok, great. I wonder if it is even useful to include the attributes defined in XTCE or if we should just centralize on using the YAML for all attributes.

greglucas: I ended up removing the attributes and short/long descriptions. It simplified things, so I think makes sense for now.

imap_processing/utils.py
greglucas: I added an optional argument to it that defaults to `derived_value`, but allows one to get the raw value if we want too. I can remove it or change the default if you think this would be better as `raw` by default. I actually think we should use derived if we can, but I'm not wedded to that.

greglucas:
@subagonsouth thanks for the comments. I updated `IMAP-Hi` now with the new function to remove the convert_time function as well. I hope that gives you an idea of how this will look for others, but I'm hoping others can make use of this and then hopefully we can even broadcast numpy array results in your Hi functions ;)

imap_processing/cdf/utils.py
```diff
@@ -149,7 +151,7 @@ def write_cdf(dataset: xr.Dataset) -> xr.Dataset:
         version=version,
         repointing=repointing,
     )
-    file_path = science_file.construct_path()
+    file_path: Path = science_file.construct_path()
```
tech3371: Was this done for a reason? Just curious

imap_processing/utils.py
tech3371: I agree. I think it's gonna save lot of repeated steps that we all are doing separately. This function seems much better. It's decomming, creating dataset and then returning it per apid. Then every processing that happens after that can decide what to keep and what to drop in terms of the actual data. And it's already in the format we need to write to cdf except we need to do minor things to add attrs stuff.

imap_processing/utils.py
```diff
@@ -228,29 +228,77 @@ def create_dataset(
...
+    with open(packet_file, "rb") as binary_data:
+        packet_generator = packet_parser.generator(binary_data)
+        for packet in packet_generator:
+            apid = packet.header["PKT_APID"].raw_value
+            if apid not in data_dict:
+                # This is the first packet for this APID
+                data_dict[apid] = collections.defaultdict(list)
+
+            # TODO: Do we want to give an option to remove the header content?
+            packet_content = packet.data | packet.header
```
tech3371, line 280: I do think it will be helpful to add that option in this function. Removing packet header content will be same across all packet and makes sense to do that here. As for removing data following header, it makes sense to do that in the individual processing code as you mentioned.

imap_processing/utils.py
```diff
@@ -228,29 +228,77 @@ def create_dataset(
...
+                if use_derived_value:
+                    # Use the derived value if it exists, otherwise use the raw value
+                    val = value.derived_value or val
+                data_dict[apid][key].append(val)
+
+    dataset_by_apid = {}
+    # Convert each apid's data to an xarray dataset
+    for apid, data in data_dict.items():
+        # The time key is always the first key in the data dictionary on IMAP
+        time_key = next(iter(data.keys()))
```
tech3371, line 293: I am little confused about the comment. Do you mean that the time key is always the first key if we don't include the packet header and only include the data itself? That is true if that's the case. But if we include packet header than it won't the first key, right? Just want to make sure I understood your comment

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -475,7 +468,8 @@ def process_swapi_science(sci_dataset: xr.Dataset, data_version: str) -> xr.Data
 
     # epoch time. Should be same dimension as number of good sweeps
     epoch_time = good_sweep_sci["epoch"].data.reshape(total_full_sweeps, 12)[:, 0]
-    epoch_converted_time = met_to_j2000ns(epoch_time)
+    # epoch_converted_time = met_to_j2000ns(epoch_time)
```
tech3371: remove?

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -532,8 +526,9 @@ def process_swapi_science(sci_dataset: xr.Dataset, data_version: str) -> xr.Data
     )
 
     # L1 quality flags
+    # TODO: Should these be kept in raw format rather than derived into strings?
```
tech3371, line 528: keep it raw for now. I will ask SWAPI team if we want in derived value.

imap_processing/tests/swapi/test_swapi_l1.py
```diff
@@ -208,10 +203,7 @@ def test_swapi_algorithm(decom_test_data):
 
 def test_process_swapi_science(decom_test_data):
     """Test process swapi science"""
-    grouped_data = group_by_apid(decom_test_data)
-    science_data = grouped_data[SWAPIAPID.SWP_SCI]
-    sorted_packets = sort_by_time(science_data, "SHCOARSE")
-    ds_data = create_dataset(sorted_packets, include_header=False)
+    ds_data = decom_test_data[SWAPIAPID.SWP_SCI]
```
tech3371, line 213: nice!

imap_processing/tests/swapi/test_swapi_l1.py
```diff
@@ -333,5 +325,6 @@ def test_swapi_l1_cdf():
 
     # hk cdf file
     cdf_filename = "imap_swapi_l1_hk_20100101_v001.cdf"
-    cdf_path = write_cdf(processed_data[1])
+    # Ignore ISTP checks for HK data
+    cdf_path = write_cdf(processed_data[1], istp=False)
```
tech3371, line 336: nice!

imap_processing/cdf/utils.py
greglucas: Because mypy was complaining later when it was used. I think our `imap-data-access` doesn't have type hints yet so this adds it for this one use.

imap_processing/utils.py
greglucas: I agree it is different depending on the header, so I avoided that situation here by making it the first key right above this by appending the packet header after the data. So regardless if we want to use the header or not, the SHCOARSE will be the first entry in this content dict.
```python
packet_content = packet.data | packet.header
```

tech3371: I see. Nice! I totally missed the order in which you added those. That makes sense now. Thank you!

imap_processing/swapi/l1/swapi_l1.py
greglucas: Done.

imap_processing/swapi/l1/swapi_l1.py
greglucas: OK, I reverted this change and set `use_derived_value=False` for SWAPI. See the latest commit for the update.

imap_processing/utils.py
greglucas: I am starting to be more of the mindset that we don't want to be adding options just to have them, is there a specific instrument/use-case you know of where we want both cases. So for now, I think keeping all packet content seems reasonable and then downstream people can remove the header if they want. But it seems like more people wanted the header than not?

tech3371: I have not heard strong preference of wanting to remove packet header. I think it came up in past discussion and brought it up just in case. That makes sense and works for me.

imap_processing/swapi/l1/swapi_l1.py
tech3371: thank you!

tech3371 (approved):
Thank you for doing this! Looks great!

greglucas:
@subagonsouth, as an FYI, I did update some of Hi's code in this as well and left some TODO comments here. Let me know if you want me to remove/change any of the Hi updates.

imap_processing/hi/l1a/histogram.py
```diff
@@ -35,30 +33,34 @@
...
-        dataset.ccsds_met[i_epoch] = packet.data["CCSDS_MET"].raw_value
-        dataset.esa_stepping_num[i_epoch] = packet.data["ESA_STEP"].raw_value
+    # TODO: Move into the allocate dataset function?
+    dataset["epoch"].data[:] = input_ds["epoch"].data
+    dataset["ccsds_met"].data = input_ds["ccsds_met"].data
+    dataset["esa_stepping_num"].data = input_ds["esa_step"].data
 
+    # unpack the packets data into the Dataset
+    # (npackets, 24 * 90 * 12)
+    # TODO: Look into avoiding the for-loops below
```
tmplummer, line 59: Yes... Good recommendation!

imap_processing/hi/l1a/histogram.py
```diff
@@ -35,30 +33,34 @@
...
     """
-    dataset = allocate_histogram_dataset(len(packets))
+    dataset = allocate_histogram_dataset(len(input_ds.epoch))
 
-    # unpack the packets data into the Dataset
-    for i_epoch, packet in enumerate(packets):
-        dataset.epoch.data[i_epoch] = met_to_j2000ns(packet.data["CCSDS_MET"].raw_value)
-        dataset.ccsds_met[i_epoch] = packet.data["CCSDS_MET"].raw_value
-        dataset.esa_stepping_num[i_epoch] = packet.data["ESA_STEP"].raw_value
+    # TODO: Move into the allocate dataset function?
```
tmplummer, line 52: These changes look fine for now. The `allocate_histogram_dataset` function was really just my way of avoiding having an intermediate data storage. I wanted to go directly from the packet into the xr.DataSet arrays. With your new function, this is no longer achieving that goal. I will write a ticket to look at this Hi code and address the TODOs.

tmplummer (approved):

PR #688, opened 2024-07-08
Glows l1a cdf

### LINKED ISSUE #683: Move GLOWS L1A processing code to CDF attribute manager
### Summary

Move the GLOWS L1A processing code to use ImapAttributeManager rather than global_attrs and glows_cdf_attrs. 

This will entail moving the L1A attributes out of glows_cdf_attrs.py into YAML file(s), and then updating glows_l1a.py to use ImapAttributeManager rather than glows_cdf_attrs. 

### Requirements

The more specific the better.
  - [ ] delete glows_cdf_attrs.py
  - [ ] Produce valid CDF files for L1A that match the existing CDF files 
  - [ ] Transition attribute management for GLOWS to YAML files

### Related tickets

_No response_

anamanica:
# Change Summary

## Overview
Changed glows l1a files over to using ImapCdfAttributes.

## New Dependencies
None 

## New Files
- imap_glows_l1a_variable_attrs.yaml
    - Changing glows_cdf_attrs.py to .yaml file

## Deleted Files
- glows_cdf_attrs.py
   - This file is no longer needed due to the process of using .yaml files for cdf attributes.

## Updated Files
- imap_glows_global_cdf_attrs.yaml
   - Added global attributes from glows_cdf_attrs.
- glows_l1a.py
   - This file is where the chang from the old attribute management to ImapCdfAttributes occurred.

## Testing
None

closes #683


imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
+# Info found in cdf.defaults.py
+# Info defined from glows_cdf_attrs
+int_fillval: &int_fillval -9223372036854775808
+min_epoch: &min_epoch -315575942816000000
+max_epoch: &max_epoch 3155630469184000000
+
+#Information taken from CDF warnings of missing attributes
+default_attrs: &default_attrs
+  # Resolution is optional?
```
maxinelasp, line 9: Is that error from SKTEditor, cdflib, or from cdf_attribute_manager?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  RESOLUTION: ' '
+  DISPLAY_TYPE: no_plot
+  TIME_BASE: J2000
+  # Guess from other l1a sections, but optional?
+  TIME_SCALE: Terrestrial Time
```
maxinelasp, line 14: This is correct but it's odd that it gave you an error

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  # Guess from other l1a sections
+  REFERENCE_POSITION: Rotating Earth Geoid
```
maxinelasp, line 16: Good guess, probably not correct, but that's a SPICE question too. The reference for this data is the GLOWS instrument, but I don't know what the value of this variable should be. It's fine to leave it. ([More info here](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/Tutorials/pdf/individual_docs/04_concepts.pdf))

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  # Guess from other l1a sections
+  UNITS: ns
```
maxinelasp: I'd say that this should be unitless by default. 
```suggestion
  UNITS: ' '
```

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  FILLVAL: *int_fillval
+  VALIDMIN: *min_epoch
+  CATDESC: ' '
```
maxinelasp: Rather than leaving these required attributes blank, I'd prefer if they weren't included in the default. That way, if someone makes a variable dependent on this, it will throw errors on missing attributes instead of silently writing nothing.

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  DISPLAY_TYPE: time_series
+  LABLAXIS: Counts
+
+event_attrs:
+  <<:  *default_attrs
+  VALIDMMIN: 0
+  VALIDMAX: 300
+  CATDESC: Direct events recorded approximately per second
+  FIELDNAM: List of direct events
+  # FORMAT is different from glows_cdf_attrs.py
```
maxinelasp, line 75: It's probably wrong in glows_cdf_attrs 😅  possibly due to the more strict typing of mypy

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  <<:  *default_attrs
+  DEPEND_0: epoch
+  VALIDMIN: 0
+  VALIDMAX: 1000000000
+  DISPLAY_TYPE: time_series
+  FILL_VAL: *int_fillval
+  VAR_TYPE: metadata
+  FORMAT: I10
+  LABLAXIS: Metadata
+  RESOLUTION: ' '
```
maxinelasp: Should be getting this from default_attrs

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  VALIDMIN: 0
+  VALIDMAX: 70
+  CATDESC: Histogram of photon counts
+  DEPEND_0: epoch
+  DEPEND_1: bins
+  FIELDNAM: Histogram of photon counts
+  # FORMAT is different from glows_cdf_attrs.py
+  # Changed this in order to avoid warnings
+  # Float with 4 digits
+  FORMAT: F0.5
```
maxinelasp: This is meant to be a [fortran format code](https://docs.oracle.com/cd/E19957-01/805-4939/z40007437a2e/index.html) (don't ask me why.) If we wanted a float with four digits, with a maximum of 100, we would need the code `F7.4` (for 7 total digit places, with 4 after the decimal.) In particular, that first number can't be zero and it can't be less than the second number. 

This is a count so it should actually probably be an int... but we can leave it as a float. It has a max of 70, so let's say 3 digits to be safe. `F7.4` is a good code for this attribute.

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  VALIDMAX: 1000000000
+  DISPLAY_TYPE: time_series
+  FILL_VAL: *int_fillval
+  VAR_TYPE: metadata
+  FORMAT: I10
+  LABLAXIS: Metadata
+  RESOLUTION: ' '
+
+# not-in--dicts in generate_de_dataset
+# Should I be defining the metadata_attrs here?
```
maxinelasp: Ah yeah this isn't in the dicts because in between when I first wrote the code and now, Greg added ground_software_version as a global variable for everyone. So actually, all references to ground_software_version should be removed from the GLOWS specific code.

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+ground_software_version:
+  <<: *metadata_attrs
+  CATDESC: Ground software version in the format vXXX
+  FIELDNAM: Ground software version
+
+pkts_file_name:
```
maxinelasp, line 130: I haven't figured out how the string variables work, so this may not be used in the code right now. It should still be included but just FYI

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  FIELDNAM: Last spin ID
+
+imap_start_time:
+  <<: *metadata_attrs
+  CATDESC: Start time of the block, in seconds, in IMAP spacecraft clock time
+  FIELDNAM: Start time in seconds in IMAP time
+  # FORMAT is different from glows_cdf_attrs.py
+  # Changed this in order to avoid warnings
+  # Can range from F0 to F0.9
+  FORMAT: F0
```
maxinelasp: From the other notes - this should be something like `F16.6` (10 digits for seconds makes sense because that's several years, and 6 digits after just feels pretty good to me. So all the time variables should have that format.

maxinelasp: Could you please add "UNITS: seconds" for the timestamps too?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  CATDESC: End time of the block, in seconds, in GLOWS clock time
+  FIELDNAM: End time in seconds in GLOWS time
+  # FORMAT is different from glows_cdf_attrs.py
+  # Changed this in order to avoid warnings
+  # Can range from F0 to F0.9
+  FORMAT: F0.5
+
+flags_set_onboard:
+  <<: *metadata_attrs
+  CATDESC: Flags for bad data [missing PPS, missing time, missing spin phase, missing spin periods, overexposure, nonmonotonic event, data collected at night, HV test, pulse test, memory error]
```
maxinelasp: CATDESC should have a max of 80 characters (I know this was copied from my code)

```suggestion
  CATDESC: Flags for bad data 
```

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  FIELDNAM: Total number of histogram events
+
+filter_temperature_average:
+  <<: *metadata_attrs
+  CATDESC: Filter temperature, averaged per block, and unit " "encoded
+  FIELDNAM: Average filter temperature, unit encoded
+
+filter_temperature_variance:
+  <<: *metadata_attrs
+  CATDESC: Filter temperature variance, per block, and unit " "encoded.
```
maxinelasp:
```suggestion
  CATDESC: Filter temperature variance, per block, and unit encoded.
```

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+  FIELDNAM: Number of histogram bins
+
+number_of_events:
+  <<: *metadata_attrs
+  CATDESC: Total number of events or counts in the histogram
+  FIELDNAM: Total number of histogram events
+
+filter_temperature_average:
+  <<: *metadata_attrs
+  CATDESC: Filter temperature, averaged per block, and unit " "encoded
```
maxinelasp:
```suggestion
  CATDESC: Filter temperature, averaged per block, and unit encoded
```

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -0,0 +1,381 @@
...
+
+memory_error_detected:
+  <<: *metadata_attrs
+  CATDESC: Flag
+  FIELDNAM: Flag
+# End of data_every_second
+
+# Commented out in support_data
+# Do I include *metadata_attrs?
+missing_packet_sequences:
```
maxinelasp, line 367: This is also an open question, I suspect this will end up being a global attribute but it's fine the way it is now (which i believe is commented out)

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -1,20 +1,36 @@
...
-from imap_processing.cdf.global_attrs import ConstantCoordinates
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.cdf.utils import J2000_EPOCH, met_to_j2000ns
-from imap_processing.glows import __version__, glows_cdf_attrs
 from imap_processing.glows.l0.decom_glows import decom_packets
 from imap_processing.glows.l0.glows_l0_data import DirectEventL0
 from imap_processing.glows.l1a.glows_l1a_data import DirectEventL1A, HistogramL1A
 
 
+def create_glows_attr_obj() -> ImapCdfAttributes:
```
maxinelasp: Since this will be used only in one place, it doesn't need to be a method - but I like where your head is! I'm generally very very pro methods (as you can see in this code) but creating the attribute object is very simple so it's overkill.

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -34,14 +50,22 @@ def glows_l1a(packet_filepath: Path, data_version: str) -> list[xr.Dataset]:
...
     hist_l0, de_l0 = decom_packets(packet_filepath)
 
+    # Process the direct event data into a dictionary grouped by the day.
     de_by_day = process_de_l0(de_l0)
+    # Dictionary is used to store histogram data grouped by day
     hists_by_day = defaultdict(list)
 
-    # Create histogram L1A and filter into days based on start time
+    # TODO: Is there a reason process_de_l0 is its own function and we have the
+    #   code for histrogram data here?
```
maxinelasp: Answer to your todo: the histogram data is already split by day, so it doesn't really need more than this, and it's simple enough to leave here instead of in a method. 

On the other hand, the direct events are more complicated because they have this sequencing where one DirectEventL1A object is composed of multiple packets, so I put it in another method. This is also partly future-proofing - I expect that process_de_l0 method to be a lot more complicated when I address all the weird corner cases that come from combining packets, whereas the histogram code will never be more complicated than this.

On the other hand, it's kind of weird that the histogram doesn't have its own method. So, you do have a point that maybe it should.

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -50,18 +74,23 @@ def glows_l1a(packet_filepath: Path, data_version: str) -> list[xr.Dataset]:
         hists_by_day[hist_day].append(hist_l1a)
 
     # Generate CDF files for each day
+    # Loop through the grouped histogram and direct event data,
+    # generate datasets for each day,
+    # Add to output list.
```
maxinelasp: Could you tidy up a few of these comments? If it's obvious what is happening from the code, then adding comments just gives more places where the code and the comment might fall out of sync. (For example, if I were to change `generate_histogram_dataset` to return a path to a generated file rather than a dataset, I wouldn't need to change any code here.... except this comment, which I might forget to do.)

This is partly a style thing, but I prefer to have minimal comments in the code and make the variables and functions very descriptive so you can understand just by reading the code. Obviously it's fine if you add comments to track for yourself (I do that a lot as well) but any unnecessary ones should be cleaned up in the PR.

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -50,18 +74,23 @@ def glows_l1a(packet_filepath: Path, data_version: str) -> list[xr.Dataset]:
...
     output_datasets = []
     for hist_l1a_list in hists_by_day.values():
-        dataset = generate_histogram_dataset(hist_l1a_list, data_version)
+        dataset = generate_histogram_dataset(hist_l1a_list, data_version, glows_attrs)
```
maxinelasp: Nice, I like the way the attrs are handled here

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -50,18 +74,23 @@ def glows_l1a(packet_filepath: Path, data_version: str) -> list[xr.Dataset]:
...
     for de_l1a_list in de_by_day.values():
-        dataset = generate_de_dataset(de_l1a_list, data_version)
+        dataset = generate_de_dataset(de_l1a_list, data_version, glows_attrs)
         output_datasets.append(dataset)
 
+    # Return generated Datasets
     return output_datasets
 
 
+# Combining packets with direct event sequences that span multiple packets.
```
maxinelasp: There shouldn't be a comment like this above the function call, all the associated documentation for the function should be in the docstring. If the documentation for the function is unclear, always feel free to add to it there. My notes about comments falling out of sync don't apply to official docstrings, because people updating the code should know to update official docstrings if they're affected. 

Comment leaving is kind of complicated 😅 Personally, I try to make all my code self-documenting, which means it doesn't _require_ comments. (Except official docstrings for methods and classes and stuff) Feel free to ask me more about this in the next student meeting if I'm being confusing.

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -82,24 +111,32 @@ def process_de_l0(
...
+    # Return the processed Direct Events
     return de_by_day
 
 
+# Generate xarray dataset(array of dimensions) from a list of DirectEventL1a objects
 def generate_de_dataset(
-    de_l1a_list: list[DirectEventL1A], data_version: str
+    de_l1a_list: list[DirectEventL1A],
+    data_version: str,
+    imap_object: ImapCdfAttributes,
```
maxinelasp: Could you rename this to include "cdf" and "attributes"? In this case, "imap" is not necessary. I'd name it something like "glows_cdf_manager" or "glows_cdf_attributes" to make it clear it's a glows-specific thing, or "cdf_attributes" or "cdf_manager". 

This comes back to the self-documenting thing: "imap_object" is super vague and could mean almost anything. "glows_cdf_manager" is obviously related to glows and cdf, and combined with "get_variable_attribute" it's super clear what is happening there.

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -118,19 +157,19 @@ def generate_de_dataset(
...
+    # Create a 3D array to store the direct events data
     # Each DirectEventL1A class covers 1 second of direct events data
     direct_events = np.zeros((len(de_l1a_list), len(de_l1a_list[0].direct_events), 4))
-    global_attributes = glows_cdf_attrs.glows_l1a_de_attrs.output()
 
-    global_attributes["ground_software_version"] = __version__
-    global_attributes["Data_version"] = data_version
-    # In header: block header, missing seqs
-    # Time varying - statusdata
+    imap_object.add_global_attribute("Data_version", data_version)
```
maxinelasp: Since this is the same for both de and histograms, can you pull it out and move it next to where you're creating the `ImapCdfAttributes` manager? Then you don't need to pass `data_version` into either generate_dataset method.

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -118,19 +157,19 @@ def generate_de_dataset(
...
+    # TODO: What is ground_software_version??
```
maxinelasp: Oh yeah, this also came up in another comment. This would mean the version of `imap_processing`, which now is automatically added by some changes Greg made. So it's no longer necessary here.

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -203,27 +245,30 @@ def generate_de_dataset(
         for key in data_every_second.keys():
             data_every_second[key].append(de.status_data.__getattribute__(key))
 
+    # Convert arrays and dictionaries into xarray 'DataArray' objects
     epoch_time = xr.DataArray(
         time_data,
         name="epoch",
         dims=["epoch"],
-        attrs=ConstantCoordinates.EPOCH,
+        # attrs=ConstantCoordinates.EPOCH,
```
maxinelasp: Remove these commented out lines

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -289,19 +329,24 @@ def generate_histogram_dataset(
...
     """
+    # Store timestamps for each HistogramL1A object.
     time_data = np.zeros(len(hist_l1a_list), dtype="datetime64[ns]")
     # TODO Add daily average of histogram counts
     # TODO compute average temperature etc
     # Data in lists, for each of the 25 time varying datapoints in HistogramL1A
 
     hist_data = np.zeros((len(hist_l1a_list), 3600))
 
+    imap_object.add_global_attribute("Data_version", data_version)
```
maxinelasp: Ditto move this and remove data_version from this method

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -353,31 +398,31 @@ def generate_histogram_dataset(
...
         dims=["epoch", "bins"],
         coords={"epoch": epoch_time, "bins": bins},
-        attrs=glows_cdf_attrs.histogram_attrs.output(),
+        attrs=imap_object.get_variable_attributes("histogram_attrs"),
     )
 
-    glows_attrs = glows_cdf_attrs.glows_l1a_hist_attrs.output()
-    glows_attrs["Data_version"] = data_version
+    glows_attrs_global = imap_object.get_global_attributes("imap_glows_l1a_hist")
+    glows_attrs_global["Data_version"] = data_version
```
maxinelasp: Hey, this line is a duplicate! I never noticed. Can you remove it, we're already setting the global attribute for Data_version elsewhere

maxinelasp: Actually you can get rid of `glows_attrs_global` entirely and just directly set `attrs = imap_object.get_global_attributes("imap_glows_l1a_hist")`

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -267,18 +310,15 @@ def generate_de_dataset(
...
+            attrs=imap_object.get_variable_attributes(key),
         )
 
+    # Return this 'Dataset'
     return output
 
 
 def generate_histogram_dataset(
-    hist_l1a_list: list[HistogramL1A], data_version: str
+    hist_l1a_list: list[HistogramL1A], data_version: str, imap_object: ImapCdfAttributes
```
maxinelasp: Ditto renaming imap_object

imap_processing/tests/glows/test_glows_l1a_cdf.py
```diff
@@ -32,9 +33,21 @@ def l1a_data():
     return (histograms_l1a, de_l1a)
 
 
+def test_create_glows_attr_obj():
```
maxinelasp: I appreciate that you thought to add a test for this! (or codecov made you do it, lol) but everything in this function is just calling out to _other_ functions, and those functions are already tested, so this is a duplicate test. 

This is related to the updating comments thing from above - if some functionality of get_variable_attributes changes internally, we don't want to have this test in a different part of the code start failing.

imap_processing/tests/glows/test_glows_l1a_cdf.py
```diff
@@ -32,9 +33,21 @@ def l1a_data():
...
+            "attribute_key"
+        ][variable_attrs]["required"]
+        if required_var_attributes is True:
+            assert variable_attrs in glows_his_attrs.keys()
+
+
 def test_generate_histogram_dataset(l1a_data):
     histograms_l1a, _ = l1a_data
-    dataset = generate_histogram_dataset(histograms_l1a, "v001")
+    glows_attrs = create_glows_attr_obj()
```
maxinelasp: Ah, I see now why you made that a method. In this case, I'll leave it up to you whether you want to leave it as a method or just duplicate in the tests. I know I said "Don't Repeat Yourself" but it's usually ok to do in unit tests because it makes it easier to change small pieces (like, if I wanted to add a new variable attribute file or something in the unit tests.)

Sorry, I feel like so much of my advice is contradictory.... IMO coding is an art, not a science, so some of it is style preferences, and some of it is to move towards my ultimate goal to have clean, beautiful, logical code. Like poetry... you want to take away all the words that are unnecessary and only leave the things that are necessary.

maxinelasp (commented):
Long review again! Really nice work overall, good job following the existing style of the code and YAML files while still putting your own touch on things. I don't think it will take long to get this to a good state.

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: I am pretty sure these were from cdf_attribute_manager or cdflib... These were the warnings we came across together in the terminal on pycharm.

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: Hmmm... Yeah, it shows in the schema that this variable is required.

maxinelasp: Ah, schema error. Ok, leave it, and I will make a mental todo to fix that in SAMMI

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
maxinelasp: I think most of those were from cdf_attribute_manager, probably schema issues

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: Right! So the reason  I wrote `CATDESC` and `FIELDNAM` into the `default_attrs` is because the file will throw errors if they are not present even in these general sections (for lack of better phrasing). Is there a way I can bypass this?

maxinelasp: huh that's... wrong. Are `default_attrs` accidentally getting used when writing the file somewhere or something? Can you send the errors you see?

imap_processing/tests/glows/test_glows_l1a_cdf.py
anamanica: This is wonderful advice! Yes, coding is definitely an art. I was thinking about this, and while normally I would vote on the side of taking it out as a method, I actually think it works well in this case. We use it in both `generate_de_dataset`, and `generate_histogram_dataset`, we are able to load in global and variable attributes in one place, and we are able to load in the `data_version` directly to this method. Thoughts?

maxinelasp: Up to you, your logic makes sense to me

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -12,14 +12,14 @@ default_attrs: &default_attrs
...
+  # Guess from other l1a sections, prob wrong
   REFERENCE_POSITION: Rotating Earth Geoid
   # Guess from other l1a sections
-  UNITS: ns
+  UNITS: ' '
   FILLVAL: *int_fillval
   VALIDMIN: *min_epoch
-  CATDESC: ' '
-  FIELDNAM: ' '
+  # CATDESC: Default attributes used for multiple variables
```
maxinelasp: Just remove these lines instead of commenting them out

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -13,10 +13,15 @@
 from imap_processing.glows.l1a.glows_l1a_data import DirectEventL1A, HistogramL1A
 
 
-def create_glows_attr_obj() -> ImapCdfAttributes:
+def create_glows_attr_obj(data_version: str) -> ImapCdfAttributes:
     """
     Load in 1la CDF attributes for GLOWS instrument.
 
+    Parameters
```
maxinelasp, line 20: nice

maxinelasp (approved):
Nice! One comment to address and then feel free to merge!

PR #692, opened 2024-07-09
CLI Type cleanup/standardization

maxinelasp:
# Change Summary
I noticed when making an update that the types in the CLI weren't quite correct. This DOES NOT make any changes to code, only to type returns to match up with what the code needs. There were places in Lo and MAG where the function is actually returning the wrong type, I did not update the code there (I will be making MAG changes and @sdhoyt will need to update Lo). 

## Overview

Correct type returns of processing, and standardize returns within do_processing.

## Future changes
- Fix MAG function returns to return list[xr.Dataset] and not list[Path]
- Fix Lo function returns to return list[xr.Dataset] and not list[Path]
- Fix codice type in one of the function definitions, currently marked with a mypy ignore


greglucas:
I noticed some of these failing mypy locally as well. Thank you for updating this! One thing that would be nice is if you could also update `pyproject.toml` to have the list of ignores, directories to use, etc... from the setup that pre-commit currently has within its call signature. That way I can run `mypy .` and get the same results as pre-commit.

daralynnrhode:
lgtm! Sucks that the cli return types were incorrect, I know how it feels having to make changes to the entire file, lol. In hindsight, probably should have occurred to me that the original `list` returns should have been `list[xr.Dataset]` returns instead of simply `xr.Dataset`. Thanks for making the necessary changes!

vmartinez-cu (approved):
Looks good!

maxinelasp:
@daralynnrhode 
> In hindsight, probably should have occurred to me that the original list returns should have been list[xr.Dataset] returns instead of simply xr.Dataset. Thanks for making the necessary changes!

This was confusing because we recently changed the way this code worked to be more consistent and not everyone changed over, so the types weren't aligned anyway. So, depending on which instrument you looked at, it would return different things.

PR #694, opened 2024-07-11
Bitflags

laspsandoval:
# Added Ability to Create Bitflags.

## Overview
This code introduces a mechanism for creating and managing bitflags. Bitflags are useful for representing multiple boolean options or statuses in a single integer value, which is efficient for storage and manipulation.

## New Files
- l1b_quality_flag_definitions.py
   - Defines specific quality flags for the L1b dataset using the QualityFlag class.
   - Includes both general and dataset-specific flags with associated messages.
- quality_flags.py
   - Introduced the QualityFlag class, a subclass of IntFlag, to manage bitflags with messages.
   - Added the FlagBit class to associate messages with integer values.
   - Implemented with_all_none to add pseudo-members NONE and ALL to flag enumerations.

## Testing
- test_quality_flags.py
   - Added tests for the new QualityFlag and FlagBit classes.
   - Included tests for flag decomposition, summary messages, and strict enforcement of valid flag values.
   - Verified the behavior of pseudo-members NONE and ALL.


greglucas (dismissed):
This is very complicated IMO, and I would like some more justification about _why_ we need all this extra complication. Overriding `__new__` is quite niche, adding decompose, etc... is all too much for what I would expect on a quality flag implementation.

Here is an example from another project where it is just a set of Enums. Then you can go and ask questions, Was the write successful? `my_output & OutputStatus.WRITE_ANY`, you can write a separate update function to update your status `current_status |= incoming_status` to update your flag.

```python
class OutputStatus(IntFlag):
    """Output status states.

    Notes
    -----
    Data flow:
    * (Existing)
        * Skip -> Generate
        * Pass -> Exit
        * Fail -> Exit
    * (Generate)
        * Skip -> Exit
        * Pass -> Save
        * Fail -> Recover
    * (Recover)
        * Skip -> Exit
        * Pass -> Save
        * Fail -> Exit
    * (Save)
        * New -> Exit
        * Overwrite -> Exit
        * Fail -> Exit
    * (Exit)
        * Final
        * Fatal
    """
    UNKNOWN = 0x0

    EXISTING_SKIP = 0x1  # Overwrite | no-storage
    EXISTING_NONE = 0x2  # No-exist
    EXISTING_PASS = 0x4  # Existed & no-overwrite
    EXISTING_FAIL = 0x8  # Error -> Fatal

    GENERATE_SKIP = 0x10  # Exist | Prev-error
    GENERATE_NONE = 0x20  # Empty
    GENERATE_PASS = 0x40  # Data
    GENERATE_FAIL = 0x80  # Error -> Recover

    RECOVER_SKIP = 0x100  # Generate succeeded.
    RECOVER_NONE = 0x200  # Empty
    RECOVER_PASS = 0x400  # Data
    RECOVER_FAIL = 0x800  # Error -> Fatal

    WRITE_SKIP = 0x1000  # No-storage | prev-error
    WRITE_NONE = 0x2000  # Empty (nothing to save)
    WRITE_PASS = 0x4000  # Pass
    WRITE_FAIL = 0x8000  # Error -> Fatal

    REMOTE_SKIP = 0x10000  # No-remote | prev-error
    REMOTE_NONE = 0x20000  # Empty (nothing to push)
    REMOTE_PASS = 0x40000  # Pass
    REMOTE_FAIL = 0x80000  # Error -> Fatal

    FINALIZED = 0x100000
    UNHANDLED_ERROR = 0x200000

    # Shorthands for mixed states.
    ANY_FAIL = EXISTING_FAIL | GENERATE_FAIL | RECOVER_FAIL | WRITE_FAIL | REMOTE_FAIL
    EXISTING_ANY = EXISTING_SKIP | EXISTING_NONE | EXISTING_PASS | EXISTING_FAIL
    GENERATE_ANY = GENERATE_SKIP | GENERATE_NONE | GENERATE_PASS | GENERATE_FAIL
    RECOVER_ANY = RECOVER_SKIP | RECOVER_NONE | RECOVER_PASS | RECOVER_FAIL
    WRITE_ANY = WRITE_SKIP | WRITE_NONE | WRITE_PASS | WRITE_FAIL
    REMOTE_ANY = REMOTE_SKIP | REMOTE_NONE | REMOTE_PASS | REMOTE_FAIL

    def __str__(self):
        return super().__str__().split('.', 1)[-1]
```

I'm going to requrest changes for now just so it isn't merged immediately.

laspsandoval:
> This is very complicated IMO, and I would like some more justification about _why_ we need all this extra complication. Overriding `__new__` is quite niche, adding decompose, etc... is all too much for what I would expect on a quality flag implementation.
> 
> Here is an example from another project where it is just a set of Enums. Then you can go and ask questions, Was the write successful? `my_output & OutputStatus.WRITE_ANY`, you can write a separate update function to update your status `current_status |= incoming_status` to update your flag.
> 
> ```python
> class OutputStatus(IntFlag):
>     """Output status states.
> 
>     Notes
>     -----
>     Data flow:
>     * (Existing)
>         * Skip -> Generate
>         * Pass -> Exit
>         * Fail -> Exit
>     * (Generate)
>         * Skip -> Exit
>         * Pass -> Save
>         * Fail -> Recover
>     * (Recover)
>         * Skip -> Exit
>         * Pass -> Save
>         * Fail -> Exit
>     * (Save)
>         * New -> Exit
>         * Overwrite -> Exit
>         * Fail -> Exit
>     * (Exit)
>         * Final
>         * Fatal
>     """
>     UNKNOWN = 0x0
> 
>     EXISTING_SKIP = 0x1  # Overwrite | no-storage
>     EXISTING_NONE = 0x2  # No-exist
>     EXISTING_PASS = 0x4  # Existed & no-overwrite
>     EXISTING_FAIL = 0x8  # Error -> Fatal
> 
>     GENERATE_SKIP = 0x10  # Exist | Prev-error
>     GENERATE_NONE = 0x20  # Empty
>     GENERATE_PASS = 0x40  # Data
>     GENERATE_FAIL = 0x80  # Error -> Recover
> 
>     RECOVER_SKIP = 0x100  # Generate succeeded.
>     RECOVER_NONE = 0x200  # Empty
>     RECOVER_PASS = 0x400  # Data
>     RECOVER_FAIL = 0x800  # Error -> Fatal
> 
>     WRITE_SKIP = 0x1000  # No-storage | prev-error
>     WRITE_NONE = 0x2000  # Empty (nothing to save)
>     WRITE_PASS = 0x4000  # Pass
>     WRITE_FAIL = 0x8000  # Error -> Fatal
> 
>     REMOTE_SKIP = 0x10000  # No-remote | prev-error
>     REMOTE_NONE = 0x20000  # Empty (nothing to push)
>     REMOTE_PASS = 0x40000  # Pass
>     REMOTE_FAIL = 0x80000  # Error -> Fatal
> 
>     FINALIZED = 0x100000
>     UNHANDLED_ERROR = 0x200000
> 
>     # Shorthands for mixed states.
>     ANY_FAIL = EXISTING_FAIL | GENERATE_FAIL | RECOVER_FAIL | WRITE_FAIL | REMOTE_FAIL
>     EXISTING_ANY = EXISTING_SKIP | EXISTING_NONE | EXISTING_PASS | EXISTING_FAIL
>     GENERATE_ANY = GENERATE_SKIP | GENERATE_NONE | GENERATE_PASS | GENERATE_FAIL
>     RECOVER_ANY = RECOVER_SKIP | RECOVER_NONE | RECOVER_PASS | RECOVER_FAIL
>     WRITE_ANY = WRITE_SKIP | WRITE_NONE | WRITE_PASS | WRITE_FAIL
>     REMOTE_ANY = REMOTE_SKIP | REMOTE_NONE | REMOTE_PASS | REMOTE_FAIL
> 
>     def __str__(self):
>         return super().__str__().split('.', 1)[-1]
> ```
> 
> I'm going to requrest changes for now just so it isn't merged immediately.

I like this much better. Time for an overhaul! I'll let you know when it is updated.

laspsandoval:
@greglucas take a look to see if this is what you have in mind.

tools/quality_flags/quality_flags.py
```diff
@@ -0,0 +1,47 @@
...
+    ALL = (
+        BASE_INF
+        | BASE_MISSING_TELEM
+        | BASE_NEG
+        | BASE_RES1
+        | ENA_RES1
+        | ULTRA_RES1
+        | ULTRA_RES2
+    )
+    BASE_ALL = BASE_INF | BASE_MISSING_TELEM | BASE_NEG
```
greglucas: Do you need to add the reserved here as well?
```suggestion
    BASE_ALL = BASE_INF | BASE_MISSING_TELEM | BASE_NEG | BASE_RES1
```

tools/quality_flags/quality_flags.py
greglucas, line 1: Why is this in the "tools" folder? Don't you want this included in the package itself so Ultra can make use of it?

greglucas (commented):
This is much better IMO, thanks for updating!

I have some hesitation for adding "dead code" that isn't used anywhere. In particular, this is in the tools folder, which doesn't get brought into the processing package, so it wouldn't be able to be used by Ultra. Maybe adding a specific location where it could be used in Ultra would help.

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,63 @@
+"""Bitwise flagging."""
+
+from enum import IntFlag
+
+
+class L1bQualityFlags(IntFlag):
```
tech3371: I will look at this first thing on Monday. Can you request input from Tim as well since he may have good idea on this topic?

greglucas:
This is now in the base `imap_processing` level. To me this seems very Ultra-specific with bits reserved for them. It would be helpful to have some use-case for when and where these are going to be used still I think.

imap_processing/quality_flags.py
tech3371: On first glance, it seems like it's not ultra specific but I see ultra specific stuff in it as well. If you are creating both common flags and ultra specific flags here, can we start with simple class with common flags and then may be another class for ultra specific?

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,63 @@
...
+        | ENA_RES1
+        | ULTRA_RES1
+        | ULTRA_RES2
+    )
+    BASE_ALL = BASE_INF | BASE_MISSING_TELEM | BASE_NEG | BASE_RES1
+
+    @property
+    def name(self) -> str:
+        """
+        Override the default name property to handle combined flags.
```
tech3371, line 12: I couldn't understand on first glance what this function's purpose was. Can you add some example of how this class will be called or used for better understanding? You can take example you had in test and add to this doc string. Same for `__str__` function

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __str__(self) -> str:
+        """
+        Override the default string representation.
+
+        Returns
+        -------
+        output_string : str
+            The modified string representation without the initial segment
+            before the first period.
```
tech3371: If I understood correctly, this function was removing classname in front of it and just return the flag. If you can add example of that, it would be helpful. Thank you!

tech3371 (commented):
Thank you for starting work on this!

imap_processing/quality_flags.py
laspsandoval: test_quality_flags.py:
assert combined_flags.name == "BASE_INF|ENA_RES1"

This does not work on Python < 3.11

imap_processing/quality_flags.py
laspsandoval: IntFlag class does not support inheritance the way regular classes do. Is that what you were thinking?

tmplummer: I would suggest that you create a generic base class, `ImapQualityFlag` for all project quality flags to inherit from that only defines the functionality not names or values. Then each specific flag class defines values. e.g.
```
class L1bEnaQualityFlag(ImapQualityFlag):
    """
    ENA L1b specific flags.
    """
    INF = 2**0
    MISSING_TELEM = 2**1
```

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,63 @@
...
+
+class L1bQualityFlags(IntFlag):
+    """Quality flags."""
+
+    NONE = 0x0
+
+    BASE_INF = 0x1  # bit 0, Infinite value
+    BASE_MISSING_TELEM = 0x2  # bit 1, Missing telemetry
+    BASE_NEG = 0x4  # bit 2, Negative value
+    BASE_RES1 = 0x8  # bit 3, Reserved 1
```
tmplummer: I find specifying the values using hex no clearer than decimal values. `2**n` where `n` is the bit number is clearer IMO.

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,63 @@
...
+
+from enum import IntFlag
+
+
+class L1bQualityFlags(IntFlag):
+    """Quality flags."""
+
+    NONE = 0x0
+
+    BASE_INF = 0x1  # bit 0, Infinite value
```
tmplummer: I don't think that `BASE_` or `ENA_` should be included in the names. More flexibility is afforded by not assuming any value/name overlap across data levels or products.

imap_processing/quality_flags.py
tmplummer, line 1: Overall, I like this implementation. The only additional functionality that I found useful to have on EMUS was an added `message` for each `name, value` pair that gave a more verbose description of what the flag indicates. That may or may not be useful here.

imap_processing/quality_flags.py
laspsandoval: I realized that it does the same as def name(self) -> str and is therefore not needed.

tmplummer (commented):
This is looking good. My main recommendation is for the creation of a project base QF class.

imap_processing/quality_flags.py
laspsandoval: I'm not certain if we need it. I'll put it as a TODO at the top in case we want to come back to that.

imap_processing/quality_flags.py
laspsandoval: I tried that and IntFlag does not allow it.



```
class BaseQualityFlags(IntFlag):
    """Base Quality flags."""

    NONE = 0x0
    BASE_INF = 0x1  # bit 0, Infinite value
    BASE_MISSING_TELEM = 0x2  # bit 1, Missing telemetry
    BASE_NEG = 0x4  # bit 2, Negative value
    BASE_RES1 = 0x8  # bit 3, Reserved 1

    BASE_ALL = BASE_INF | BASE_MISSING_TELEM | BASE_NEG | BASE_RES1


class UltraQualityFlags(BaseQualityFlags):
    """Ultra specific Quality flags."""

    ENA_RES1 = 0x10  # bit 4, ENA reserved 1
    ULTRA_RES1 = 0x20  # bit 5, Ultra reserved 1
    ULTRA_RES2 = 0x40  # bit 6, Ultra reserved 2

etc...
```

TypeError: <enum 'UltraQualityFlags'> cannot extend <flag 'BaseQualityFlags'>

laspsandoval: Have you done it before? I don't think there is any other way to do what you have suggested. But I could always be wrong.

tmplummer: The reason that didn't work is because you specified names and values in the base class. Try just adding the `name` property.

There is probably some pattern that allows for extending enum classes with names/values defined, but I don't know it off the top of my head.

laspsandoval: I mis-read your comment. Ok. Looks good now. I'll push tomorrow. I have to run.

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,48 @@
...
+        Returns
+        -------
+        combined_name : str
+            The combined name of the individual flags.
+        """
+        if self._name_ is not None:
+            return self._name_
+
+        members = [member for member in BaseQualityFlags if member & self == member]
+        return "|".join(str(m.name) for m in members if m != BaseQualityFlags.NONE)
```
greglucas: Should this be `self` instead of the subclass values? I don't think the Ultra one will appear right currently.

Suggest pushing the `self.NONE` up into the initial members check as well, which I also think can possibly be shortened to `if self` indicating if any bit is set.
```suggestion
        members = [member for member in self if (member & self == member) and self]
        return "|".join(str(m.name) for m in members)
```

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class BaseQualityFlags(ImapQualityFlag):
+    """Base Quality flags."""
+
+    NONE = 0x0
```
greglucas, line 29: Do we need "NONE", or can this be implicit? No strong preference either way, so fine to leave as well.

greglucas (commented):
I removed my block on this. It still isn't passing tests though.

I don't personally think we need to be in any rush on this, so I wouldn't stress out about getting this in ASAP. Someone else can pick it up if you leave it up for review and try to get it across the line as well so it isn't on your plate alone.

laspsandoval:
@greglucas @tech3371 @subagonsouth I considered all of your suggestions and at the end decided to go the simple route. Having different classes for different instruments complicates combining names and at the end I didn't really see a purpose to do that. It is up to you if you want me to continue to making changes, but I think it is fine.

imap_processing/quality_flags.py
tech3371: I meant if you can this example in this function's doc string :)

tech3371:
I wonder if it will be helpful to close(and not merge) this PR and transition this brainstorming discussion to Galaxy page for now for STM discussion purpose. Then once we have some idea of what flags they want, then we can revisit and reopen this PR. What do you think?

laspsandoval:
> I wonder if it will be helpful to close(and not merge) this PR and transition this brainstorming discussion to Galaxy page for now for STM discussion purpose. Then once we have some idea of what flags they want, then we can revisit and reopen this PR. What do you think?

This certainly isn't the final set of Quality Flags. It is an incremental change. We can always add QF but I don't see the reason to not merge.

tmplummer:
> Having different classes for different instruments complicates combining names and at the end I didn't really see a purpose to do that.

The reason to have different classes for different instruments/data levels is that IMO, we don't want to force a common set of QFs across all instruments. Ultras `BAD_SPIN` and `FOV` flags won't make sense for all instruments and rather than bloat the number of bits required in the QF variables to cover all flags across all instruments, we can create child classes that are tailored to the instrument, data-level, and even data-type. I think what you have is a great start, but we need the ability to have tailored sub-classes and we need to decide if we want to have any shared flags that are common across all products.

laspsandoval:
> > Having different classes for different instruments complicates combining names and at the end I didn't really see a purpose to do that.
> 
> The reason to have different classes for different instruments/data levels is that IMO, we don't want to force a common set of QFs across all instruments. Ultras `BAD_SPIN` and `FOV` flags won't make sense for all instruments and rather than bloat the number of bits required in the QF variables to cover all flags across all instruments, we can create child classes that are tailored to the instrument, data-level, and even data-type. I think what you have is a great start, but we need the ability to have tailored sub-classes and we need to decide if we want to have any shared flags that are common across all products.

I think the solution may be for each instrument to have their own quality flag class.

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,46 @@
...
+    NEG = 2**1  # bit 1, Negative value
+
+
+class ENAFlags(IntFlag):
+    """Common ENA flags."""
+
+    BADSPIN = 2**2  # bit 2, Bad spin
+
+
+class ImapUltraFlags(IntFlag):
```
tech3371: This is such a bummer that we can't inherit IntFlag enum using something like this without getting this error.
```
class ImapUltraFlags(CommonFlags, ENAFlags):
```
error:
```
E   TypeError: ENAFlags: cannot extend enumeration
```

imap_processing/quality_flags.py
```diff
@@ -0,0 +1,46 @@
...
+
+    NONE = 0x0
+    INF = 2**0  # bit 0, Infinite value
+    NEG = 2**1  # bit 1, Negative value
+
+
+class ENAFlags(IntFlag):
+    """Common ENA flags."""
+
+    BADSPIN = 2**2  # bit 2, Bad spin
```
tech3371, line 37: Just to be sure, this `2**2` is 4. If we continue using this kind of power increase, will we run out of values soon or is it ok?
```suggestion
    BADSPIN = 2**2  # 4, Bad spin
```

tech3371 (commented):
This is what I had in mind. It looks good to me. I will approve once tests are passing. But will be helpful to get another input from Greg or Tim.

tech3371 (approved):

maxinelasp (approved):
This is a good starting spot and I can see easy ways to extend it!

PR #713, opened 2024-07-25
[WIP] Mypy Check Bug Fix

### LINKED ISSUE #710: BUG - mypy should only be run over our actual code
### Description of the issue

Currently, we run mypy in the current working directory: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/bf31f33cdee695537a21ee3d89ecd88a9c4d2241/.pre-commit-config.yaml#L45-L49

If someone adds an extra file without type hints like `temp.py` as a testing script, or puts their virtual environment in the current directory then mypy will try to run over those files as well.

Create a temporary file or virtual environment without type hints in your local directory, and run `mypy .` and it will fail.

Only run over the project directory `imap_processing/`.

```
venv3.12/bin/jp.py:12: error: Function is missing a return type annotation [no-untyped-def]
venv3.12/bin/rst2odt_prepstyles.py:29: error: Function is missing a type annotation [no-untyped-def]
```

### Code Snippet:

```python
Code
```

Changing mypy to run only over `imap_processing/`. I also suggest adding the current mypy arguments from the pre-commit hook into the `pyproject.toml` configuration file instead. That way anyone could run `mypy` and get the proper ignores/additions without needing to add command line arguments.

daralynnrhode:
# A fix for the mypy bug and any needed changes

## Overview
Closes #710 

Ensures that mypy will only check files within `imap_processing/` etc.  Changes to make the pre-commit mypy check and regular mypy check align. 

Currently, I am being pushed in the direction to think it is a recursion difference that is causes the difference in errors for the two checks. I am trying to track down where this is occurring. 

For Reference:

`mypy .` ran from `imap_processing/` produces 
Screen Shot 2024-07-25 at 9 53 58 AM

`pre-commit run --all` ran from `imap_processing/` produces 
Screen Shot 2024-07-25 at 9 55 29 AM

To Note; all errors found by running the pre-check _are_ present when the mypy check is ran by itself

## Updated Files

- pyproject.toml
- .pre-commit-config.yaml
- codice_l1a.py


pyproject.toml
```diff
@@ -118,5 +118,15 @@ checks = ["all", #report on all checks, except the following
 exclude = ['__init__' ] # don't report on objects that match any of these regex
 
 [tool.mypy]
+disable_error_code = ['type-arg', 'import-untyped', 'import-not-found', 'no-untyped-call']
+strict = true
+explicit_package_bases = true
 follow_imports = 'skip'
-exclude = ["tests", "docs"]
+exclude = ["tests", "docs" , "__init__.py"]
+mypy_path = ["imap_processing/" , "tools/"]
```
greglucas: I think this should be under packages, but it was also a little unclear to me what `mypy_path` is supposed to do. It sounds more like it is trying to locate the configuration and mypy specific things rather than the packages to test.
```suggestion
packages = ["imap_processing" , "tools"]
```

pyproject.toml
```diff
@@ -118,5 +118,15 @@ checks = ["all", #report on all checks, except the following
 exclude = ['__init__' ] # don't report on objects that match any of these regex
 
 [tool.mypy]
+disable_error_code = ['type-arg', 'import-untyped', 'import-not-found', 'no-untyped-call']
+strict = true
+explicit_package_bases = true
 follow_imports = 'skip'
-exclude = ["tests", "docs"]
+exclude = ["tests", "docs" , "__init__.py"]
```
greglucas: I think you could remove this if you explicitly add the packages below.

Is there a reason we need to exclude `__init__.py` files?

pyproject.toml
```diff
@@ -118,5 +118,15 @@ checks = ["all", #report on all checks, except the following
 exclude = ['__init__' ] # don't report on objects that match any of these regex
 
 [tool.mypy]
+disable_error_code = ['type-arg', 'import-untyped', 'import-not-found', 'no-untyped-call']
```
greglucas: Very minor nit, but could you alphabetize the list, and the variables here would probably be helpful too.

.pre-commit-config.yaml
```diff
@@ -42,8 +42,5 @@ repos:
     hooks:
       - id: mypy
         pass_filenames: false
-        args: [ ., --strict, --explicit-package-bases,
-                  --disable-error-code, import-untyped,
-                  --disable-error-code, import-not-found,
-                  --disable-error-code, no-untyped-call,
-                  --disable-error-code, type-arg ]
+        args: [ . ]
```
greglucas: If you add `packages`, then I think you can remove this. At least then it works locally for me by just specifying `mypy` on the command line.

greglucas (commented):
Looks like you need to merge or rebase on `upstream/dev` to get the latest changes before the tests can run.

pyproject.toml
daralynnrhode: Nope, It was left over from when I was messing around.

daralynnrhode: Also, 'tests' need to be still excluded because test folders are located within both the imap_processing and tools folders.
It's a give and take, you could make packages more defined and therefor won't need to use exclude or keep it more generic and include the exclude section. 
In this case I feel like keeping it generic is the better route.

daralynnrhode:
With the most recent commit I am still having errors produced when just `mypy` is ran.

.pre-commit-config.yaml
```diff
@@ -42,8 +42,4 @@ repos:
     hooks:
       - id: mypy
         pass_filenames: false
```
greglucas: I think this line might be causing the difference. When I remove this line, then I get errors in both incantations. (although pre-commit raises errors in the tests directory, but when running locally I don't get the tests/ errors 🤔 )

It looks like there are a lot of errors, so I wonder if we weren't running mypy over all of these files before and they sort of went under the radar.

daralynnrhode: This is leading me back to thinking it is a recursion problem. If I remove the pass_filename then both calls return errors, but they are not the same errors. On top of that there is no overlap whatsoever between the errors. 
Funny enough all of the pre-commit errors that are shown are errors that should not have been returned, because they are all contained within a test folder/file or in the docs. So its like it ran the opposite of what the exclude statement in the toml dictated

daralynnrhode:
Now back to the point of pre-commit passing but the local mypy check produces errors. I am hedging toward this being caused because of a difference in the recursive calls for mypy when it is called individually versus in pre-commit.

In the mypy documentation it is stated that submodules found in what is listed in `packages` _will_ be checked, so the pyproject is covering all files within imap_processing and tools. Shouldn't the pre-commit-when it runs the mypy check-check the pyproject for how mypy should be configured?

daralynnrhode:
Turns out since the pre-commit runs its check in an isolated virtualenv some things may be lost. 
[https://github.com/pre-commit/mirrors-mypy](url) 

I believe the dependencies need to be listed for mypy but am at a loss as to what all those dependencies should be.

imap_processing/utils.py
```diff
@@ -299,9 +299,11 @@ def _create_minimum_dtype_array(values: list, dtype: str) -> np.ndarray:
         The array of values.
     """
     try:
-        return np.array(values, dtype=dtype)
+        dtype_array: np.ndarray = np.array(values, dtype=dtype)
+        return dtype_array
     except ValueError:
-        return np.array(values)
+        regular_array: np.ndarray = np.array(values)
+        return regular_array
```
greglucas: I think that for all of the numpy typing we need to import their typing module, and then use `npt.NDArray`. Saying that we are returning `np.ndarray` seems incorrect, so I think we should change those, rather than add these explicit type annotations throughout the code.

```
import numpy.typing as npt

def func() -> npt.NDArray
```

https://stackoverflow.com/questions/66349242/specific-type-annotation-for-numpy-ndarray-using-mypy

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -105,7 +105,7 @@ def decompress_count(
     # Decompress counts based on compression indicators
     # If 0, value is already decompressed. If 1, value is compressed.
     # If 1 and count is 0xFFFF, value is overflow.
-    new_count = copy.deepcopy(count_data).astype(np.int32)
+    new_count: np.ndarray = copy.deepcopy(count_data).astype(np.int32)
```
greglucas: I messed up when I added this earlier, could you change this while you're at it? We don't need to deepcopy, the default array constructor is to make a copy of the data. https://numpy.org/doc/stable/reference/generated/numpy.array.html

```suggestion
    new_count = np.array(count_data, dtype=np.int32)
```

PR #718, opened 2024-07-26
Refactor epoch attrs

### LINKED ISSUE #708: Clean up remaining CDF Attrs Dataclasses Implementation
- [ ] Remove CDF Attrs Dataclasses
- [ ] Fix any remaining code that uses above classes

tech3371:
# Change Summary

## Overview
Here, I removed duplicated definition of "epoch" CDF attrs and updated all instrument code to use shared "epoch" CDF attrs.

### DIFF
--- a/imap_processing/hi/l1a/science_direct_event.py
+++ b/imap_processing/hi/l1a/science_direct_event.py
@@ -3,7 +3,6 @@
 import numpy as np
 import xarray as xr
 
-from imap_processing.cdf import epoch_attrs
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.cdf.utils import met_to_j2000ns
 
@@ -267,6 +266,7 @@ def create_dataset(de_data_list: list, packet_met_time: list) -> xr.Dataset:
     # uncomment this once Maxine's PR is merged
     # attr_mgr.add_global_attribute("Data_version", data_version)
 
+    epoch_attrs = attr_mgr.get_variable_attributes("epoch")
     epoch_attrs["CATDESC"] = (
         "Direct Event time, number of nanoseconds since J2000 with leap "
         "seconds included"
--- a/imap_processing/hi/l1a/science_direct_event.py
+++ b/imap_processing/hi/l1a/science_direct_event.py
@@ -3,6 +3,7 @@
 import numpy as np
 import xarray as xr
 
+from imap_processing.cdf import epoch_attrs
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.cdf.utils import met_to_j2000ns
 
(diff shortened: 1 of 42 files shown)


imap_processing/hi/l1a/science_direct_event.py
```diff
@@ -266,11 +267,15 @@ def create_dataset(de_data_list: list, packet_met_time: list) -> xr.Dataset:
     # uncomment this once Maxine's PR is merged
     # attr_mgr.add_global_attribute("Data_version", data_version)
 
+    epoch_attrs["CATDESC"] = (
+        "Direct Event time, number of nanoseconds since J2000 with leap "
+        "seconds included"
+    )
```
greglucas, line 273: Should this be different? I think it'd be nice to be consistent everywhere and not have this one outlier.

greglucas (commented):
I like that we are able to remove the duplicated yaml definitions and ensure consistency, that seems nice.

It still requires an extra import everywhere though, which seems like it might be nice to avoid if we can. Sorry I haven't followed the attribute work in detail, but at a quick glance I think you've already enabled this actually and by default are loading the shared variables.

So, I think you can delete this.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/687354bf294a63ea3474fc78b1d684e5dd9c95c1/imap_processing/cdf/__init__.py#L1-L5

and then in everyone's attribute manager, it should already be present because of this in the base class initialization:
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/687354bf294a63ea3474fc78b1d684e5dd9c95c1/imap_processing/cdf/imap_cdf_manager.py#L39

So, I _think_ epoch should be available within each instrument by doing this:
```python
attrs=attr_mgr.get_variable_attributes("epoch"),
```

imap_processing/hi/l1a/science_direct_event.py
tech3371: I wondered that too but didn't want to assume. @subagonsouth may know more

tech3371:
> I like that we are able to remove the duplicated yaml definitions and ensure consistency, that seems nice.
> 
> It still requires an extra import everywhere though, which seems like it might be nice to avoid if we can. Sorry I haven't followed the attribute work in detail, but at a quick glance I think you've already enabled this actually and by default are loading the shared variables.
> 
> So, I think you can delete this.
> 
> https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/687354bf294a63ea3474fc78b1d684e5dd9c95c1/imap_processing/cdf/__init__.py#L1-L5
> 
> and then in everyone's attribute manager, it should already be present because of this in the base class initialization:
> 
> https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/687354bf294a63ea3474fc78b1d684e5dd9c95c1/imap_processing/cdf/imap_cdf_manager.py#L39
> 
> So, I _think_ epoch should be available within each instrument by doing this:
> 
> ```python
> attrs=attr_mgr.get_variable_attributes("epoch"),
> ```

That what I thought initially too but I didn't want to assume that everyone will use `ImapCdfAttributes` as their class. Some were using `CdfAttributeManager`. If we encourage everyone to use `ImapCdfAttributes`, then I think what you are thinking make sense.

greglucas:
> That what I thought initially too but I didn't want to assume that everyone will use ImapCdfAttributes as their class. Some were using CdfAttributeManager. If we encourage everyone to use ImapCdfAttributes, then I think what you are thinking make sense.

We are on IMAP ;) So, let us be opinionated here if we can.

A quick grep looks like `hi_l1c` and `ultra_l1_utils` are the two locations using the base version, and I think those can (should?) both be updated to use the Imap version.

greglucas (approved):
Thank you for doing this for everyone, I like this update!

Might want to have some of the more knowledgeable attribute people sign off on this approach too though.

bourque (approved):
Thanks for doing that!

imap_processing/hi/l1a/science_direct_event.py
tmplummer: Sorry that I missed this mention. It is probably fine to change. I just wanted to highlight that this epoch variable is storing the direct event times, not packet creation times like many other epoch variables in l1a.

PR #721, opened 2024-07-26
SWAPI l2 Implementation - counts to rate

### LINKED ISSUE #706: SWAPI L2 algorithm - Convert counts to rate
SWAPI L2 algorithm includes converting counts to rate and then add quality flags. This ticket only converts counts to rate. There will be another ticket to implement quality flags. I am doing this simple implementation first to help Menlo use this as bench mark for their L3 development.

tech3371:
# Change Summary

## Overview
Process SWAPI L1 to L2. This PR will close this ticket: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/706

## New Files
   - L2 code

## Updated Files
Added L2 updates to these files
- imap_processing/cdf/config/imap_swapi_global_cdf_attrs.yaml
- imap_processing/cli.py

## Testing
- imap_processing/tests/swapi/test_swapi_l1.py
- imap_processing/tests/swapi/test_swapi_l2.py

### DIFF
--- a/imap_processing/swapi/l2/swapi_l2.py
+++ b/imap_processing/swapi/l2/swapi_l2.py
@@ -1,0 +1,110 @@
+"""SWAPI L2 processing module."""
+
+import xarray as xr
+
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
+from imap_processing.swapi.swapi_utils import SWAPIAPID
+
+
+def process_science(l1_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
+    """
+    Produce science data to L2.
+
+    To process science data to L2, we need to:
+    - convert counts to rates. This is done by dividing the counts by the
+        t_bin time. t_bin is the exposure time per energy bin which is
+        obtained by dividing the time for one complete sweep
+        (12 s, coarse + fine sweep) by the total energy steps (72),
+        i.e., t_bin = 12/72 = 0.167 s. This will be constant.
+
+    - update uncertainty. Calculate new uncertainty value using
+        SWP_PCEM_ERR data from level one and divide by t_bin. Eg.
+            SWP_PCEM_UNC = SWP_PCEM_ERR / t_bin
+        Do the same for SCEM and COIN data.
+
+    Parameters
+    ----------
+    l1_dataset : xarray.Dataset
+        The L1 data input.
+    data_version : str
+        Version of the data product being created.
+
+    Returns
+    -------
+    data : xarray.Dataset
+        Processed data to L2.
+    """
+    # Load the CDF attributes
+    cdf_manager = ImapCdfAttributes()
+    cdf_manager.add_instrument_global_attrs("swapi")
+    cdf_manager.load_variable_attributes("imap_swapi_variable_attrs.yaml")
+
+    t_bin = 0.167  # seconds
+    l2_dataset = l1_dataset.copy()
+    l2_dataset.attrs = cdf_manager.get_global_attributes("imap_swapi_l2_sci")
+    l2_dataset.attrs["Data_version"] = data_version
+    # convert counts to rate
+    l2_dataset["swp_pcem_rate"] = l1_dataset["swp_pcem_counts"] / t_bin
+    l2_dataset["swp_scem_rate"] = l1_dataset["swp_scem_counts"] / t_bin
+    l2_dataset["swp_coin_rate"] = l1_dataset["swp_coin_counts"] / t_bin
+    # update attrs
+    l2_dataset["swp_pcem_rate"].attrs = cdf_manager.get_variable_attributes("pcem_rate")
+    l2_dataset["swp_scem_rate"].attrs = cdf_manager.get_variable_attributes("scem_rate")
+    l2_dataset["swp_coin_rate"].attrs = cdf_manager.get_variable_attributes("coin_rate")
+
+    # update uncertainty
+    l2_dataset["swp_pcem_unc"] = l1_dataset["swp_pcem_err"] / t_bin
+    l2_dataset["swp_scem_unc"] = l1_dataset["swp_scem_err"] / t_bin
+    l2_dataset["swp_coin_unc"] = l1_dataset["swp_coin_err"] / t_bin
+    # update attrs
+    l2_dataset["swp_pcem_unc"].attrs = cdf_manager.get_variable_attributes(
+        "pcem_uncertainty"
+    )
+    l2_dataset["swp_scem_unc"].attrs = cdf_manager.get_variable_attributes(
+        "scem_uncertainty"
+    )
+    l2_dataset["swp_coin_unc"].attrs = cdf_manager.get_variable_attributes(
+        "coin_uncertainty"
+    )
+
+    # drop the counts and err variables
+    l2_dataset = l2_dataset.drop_vars(
+        [
+            "swp_pcem_counts",
+            "swp_scem_counts",
+            "swp_coin_counts",
+            "swp_pcem_err",
+            "swp_scem_err",
+            "swp_coin_err",
+        ]
+    )
+
+    # TODO: add thruster firing flag
+    # TODO: add other flags
+
+    return l2_dataset
+
+
+def swapi_l2(l1_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
+    """
+    Generate science data to L2.
+
+    Parameters
+    ----------
+    l1_dataset : xarray.Dataset
+        The L1 data input.
+    data_version : str
+        Version of the data product being created.
+
+    Returns
+    -------
+    data : xarray.Dataset
+        Processed data to L2.
+    """
+    apid = int(l1_dataset.attrs["apid"])
+    # Right now, we don't process any other apid besides science packet
+    # to L2
+    if apid != SWAPIAPID.SWP_SCI:
+        raise ValueError(f"APID {apid} is not supported for L2 processing")
+
+    return process_science(l1_dataset, data_version)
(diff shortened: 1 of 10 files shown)


imap_processing/swapi/l2/swapi_l2.py
```diff
@@ -0,0 +1,115 @@
...
+def process_science(l1_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
+    """
+    Produce science data to L2.
+
+    To process science data to L2, we need to:
+    - convert counts to rates. This is done by dividing the counts by the
+        t_bin time. t_bin is the exposure time per energy bin which is
+        obtained by dividing the time for one complete sweep
+        (12 s, coarse + fine sweep) by the total energy steps (72),
+        i.e., t_bin = 12/72 = 0.167 s. This will be constant.
```
greglucas: Make this a global constant outside the function? `TIME_PER_BIN = 12 / 72`

Note that 12/72 is not equal to 0.167, it is 0.166666 repeating. Do you want to store 0.167 or 12/72?

imap_processing/swapi/l2/swapi_l2.py
```diff
@@ -0,0 +1,115 @@
...
+    data : xarray.Dataset
+        Processed data to L2.
+    """
+    # Load the CDF attributes
+    cdf_manager = ImapCdfAttributes()
+    cdf_manager.add_instrument_global_attrs("swapi")
+    cdf_manager.load_variable_attributes("imap_swapi_variable_attrs.yaml")
+
+    t_bin = 0.167  # seconds
+    l2_dataset = l1_dataset.copy()
```
greglucas: I'm somewhat hesitant of this since you're copying a lot of variables and also attributes over as well. You are cleaning it up below, but if we end up changing l1 data later, we may forget to come in here and drop those extra variables and such.

Would it be easy enough to "start fresh" in a sense, and only add to the dataset later rather than drop things? Something like `l2_dataset = xr.Dataset(l1_dataset["epoch"], ...)`

imap_processing/swapi/l2/swapi_l2.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+
+    # TODO: add thruster firing flag
+    # TODO: add other flags
+    logger.info("SWAPI L2 processing complete")
+
+    return l2_dataset
+
+
+def swapi_l2(l1_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
```
greglucas: This is extra indirection here by adding another function for 1 logic statement. Can you combine this apid validation into the process function, or bring that code into this function?

imap_processing/tests/swapi/test_swapi_l2.py
```diff
@@ -0,0 +1,31 @@
+from imap_processing import imap_module_directory
+from imap_processing.cdf.utils import load_cdf, write_cdf
+from imap_processing.swapi.l1.swapi_l1 import swapi_l1
+from imap_processing.swapi.l2.swapi_l2 import swapi_l2
+from imap_processing.swapi.swapi_utils import SWAPIAPID
+
+
+def test_swapi_l1_cdf():
```
greglucas:
```suggestion
def test_swapi_l2_cdf():
```

imap_processing/tests/swapi/test_swapi_l2.py
```diff
@@ -0,0 +1,31 @@
...
+    cdf_filename = "imap_swapi_l1_sci-1min_20100101_v001.cdf"
+
+    cdf_path = write_cdf(processed_data[0])
+    assert cdf_path.name == cdf_filename
+
+    # process science data to l2
+    l1_dataset = load_cdf(cdf_path)
+    assert l1_dataset.attrs["Apid"] == f"{SWAPIAPID.SWP_SCI}"
+    assert l1_dataset.attrs["Plan_id"] == "0"
+    assert l1_dataset.attrs["Sweep_table"] == "0"
```
greglucas: These assertions should be in the l1 tests area. You are testing l2 data in this file.

```suggestion
    l1_dataset = swapi_l1(l0_data_path, data_version="v001")
```

imap_processing/tests/swapi/test_swapi_l2.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # print("after ready l1 file ", l1_dataset.attrs.keys())
```
greglucas:
```suggestion
```

imap_processing/tests/swapi/test_swapi_l2.py
```diff
@@ -0,0 +1,31 @@
...
+    l2_processed_data = swapi_l2(l1_dataset, data_version="v001")
+    l2_cdf = write_cdf(l2_processed_data)
+    assert l2_cdf.name == "imap_swapi_l2_sci-1min_20100101_v001.cdf"
```
greglucas, line 16: Add some assertions on your counts turning into rates, uncertainty variables are as expected?

`np.testing.assert_array_equal(l2_dataset["swp_pcem_unc"], l1_dataset["swp_pcem_err"] / T_BIN)`

imap_processing/swapi/l2/swapi_l2.py
tech3371: Their algorithm document says to use `0.167`. That's why I used that. This comment was to give background understanding of where that value came from. I can ask SWAPI team this week.

imap_processing/swapi/l2/swapi_l2.py
tech3371: I thought of that too but was not sure. If you think it's better that way, then I am happy to.

imap_processing/tests/swapi/test_swapi_l2.py
```diff
@@ -0,0 +1,25 @@
...
+from imap_processing import imap_module_directory
+from imap_processing.cdf.utils import write_cdf
+from imap_processing.swapi.l1.swapi_l1 import swapi_l1
+from imap_processing.swapi.l2.swapi_l2 import TIME_PER_BIN, swapi_l2
+
+
+def test_swapi_l2_cdf():
+    """Test housekeeping processing and CDF file creation"""
+    l0_data_path = (
+        f"{imap_module_directory}/tests/swapi/l0_data/"
```
tmplummer: I suggest creating a fixture that returns a `Path` which points to the l0 test data path and then using that in any test function referencing that path. e.g...
```
@pytest.fixture
def swapi_l0_test_data_path():
    return imap_module_directory / "tests" / "swapi" / "l0_data"

def test_that_uses_swapi_l0_data(swapi_l0_test_data_path):
    test_file_path = swapi_l0_test_data_path / "this_l0_file.pkts"

File: imap_processing/tests/swapi/test_swapi_l2.py [hunk 9]
```diff
@@ -0,0 +1,25 @@
...
+        "imap_swapi_l0_raw_20231012_v001.pkts"
+    )
+    processed_data = swapi_l1(l0_data_path, data_version="v001")
+    l1_dataset = processed_data[0]
+
+    l2_dataset = swapi_l2(l1_dataset, data_version="v001")
```
tmplummer, line 14: It is worth thinking about if the input `l1_dataset` could be synthetic or come from an l1 file. At some point, it is nice to not need to run the l1 processing to generate test data for l2. That is up to you as long as you keep in mind that avoiding adding tests that take a long time to run is good.

tmplummer (commented):
This looks good to me. I left some comments that are more advice about changes that can avoid future challenges.

tmplummer (approved):

imap_processing/swapi/l2/swapi_l2.py
```diff
@@ -0,0 +1,105 @@
...
+    Returns
+    -------
+    data : xarray.Dataset
+        Processed data to L2.
+    """
+    apid = int(l1_dataset.attrs["Apid"])
+    # Right now, we don't process any other apid besides science packet
+    # to L2
+    if apid != SWAPIAPID.SWP_SCI:
+        raise ValueError(f"APID {apid} is not supported for L2 processing")
```
greglucas: Thinking about this more. Why do we have an apid check at level 2, shouldn't this actually be a "level 1 filename check" instead?

All this to say, I'm not actually sure we need this check, or if it really should be up in the CLI dependencies side of things when you are checking the length of dependencies you could also check the filename there if you wanted.

imap_processing/swapi/l2/swapi_l2.py
```diff
@@ -0,0 +1,105 @@
...
+        "epoch",
+        "energy",
+        "energy_label",
+        "swp_pcem_flags",
+        "swp_scem_flags",
+        "swp_coin_flags",
+    ]
+    l2_dataset = l1_dataset[l1_data_keys].copy()
+    # Copy over attributes
+    l2_dataset.attrs = l1_dataset.attrs
```
greglucas: Nice, I like this approach of keeping only what you need first rather than deleting what you don't. Thanks for updating it.

I don't think we need the copy, but can just use the view into it?
```suggestion
    l2_dataset = l1_dataset[l1_data_keys]
```
The attributes should already be brought along as well I thought.

greglucas (approved):
Looks good to me. Some minor nits that don't impact the actual calculations.

imap_processing/tests/swapi/test_swapi_l2.py
tech3371: I will think about it. May not do that in this PR but will keep in mind.

imap_processing/swapi/l2/swapi_l2.py
tech3371: True, this will be prevent by dependency length. Ok. I can remove it.

imap_processing/swapi/l2/swapi_l2.py
tech3371: yeah, that worked. I was having issue before and I think it was because of `.copy()`. Thanks!

PR #723, opened 2024-07-26
MNT: Minimize the datatypes created from space_packet_parser

### LINKED ISSUE #722: Infer the proper numpy data types when reading packets
### Summary

When creating the xarray datasets naively numpy infers 64bit sizes. We do have a little more knowledge because `space_packet_parser` has the packet definition with the specific types defined. We can map those to the proper numpy types and use those directly.

### Requirements

Add a function to calculate the numpy datatype from `space_packet_parser` and use that when creating the xarray datasets from packet files.

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview

Previously all datatypes from numpy were created with int64/float64 encodings. `np.array([1, 2, 3])` isn't minimized. We know the expected datatype from space packet parser's xtce definition, so we can infer a good numpy datatype to minimize the size.

closes #722 

## Testing

I added a test within SWAPI's area because it is using the function already and the data is there to test with. I'm not sure if we want something more standalone to test the functionality or if this is OK, let me know if you have preferences either way.

## Extra

cc @medley56 and @BStoneLASP as I talked to both of you about this earlier. @medley56 let me know if something like this would be nice to add to `space_packet_parser`, or if we should look at doing the "simpler" thing and just add a reference to the `DataEncoding` within the `ParsedDataItem`. Happy to PR something over there if it'd be useful more broadly.

### DIFF
--- a/imap_processing/tests/test_utils.py
+++ b/imap_processing/tests/test_utils.py
@@ -5,7 +5,7 @@ import pandas as pd
 import pytest
 import xarray as xr
 
-from imap_processing import utils
+from imap_processing import imap_module_directory, utils
 
 
 def test_convert_raw_to_eu(tmp_path):
@@ -69,3 +69,38 @@ def test_convert_raw_to_eu(tmp_path):
         eu_dataset = utils.convert_raw_to_eu(
             dn_dataset.copy(), test_csv.absolute(), "PACKET_1", comment="#"
         )
+
+
+@pytest.mark.parametrize(
+    "use_derived_value, expected_mode", [(True, "HVENG"), (False, 2)]
+)
+def test_packet_file_to_datasets(use_derived_value, expected_mode):
+    """
+    Test that all datatypes aren't all int64 and that we get
+    uint8/uint16 from header items as expected.
+
+    Test that we get multiple apids in the output.
+    """
+    test_file = "tests/swapi/l0_data/imap_swapi_l0_raw_20231012_v001.pkts"
+    packet_files = imap_module_directory / test_file
+    packet_definition = (
+        imap_module_directory / "swapi/packet_definitions/swapi_packet_definition.xml"
+    )
+    datasets_by_apid = utils.packet_file_to_datasets(
+        packet_files, packet_definition, use_derived_value=use_derived_value
+    )
+    # 3 apids in the test data
+    assert len(datasets_by_apid) == 3
+    data = datasets_by_apid[1188]
+    assert data["sec_hdr_flg"].dtype == np.uint8
+    assert data["pkt_apid"].dtype == np.uint16
+    np.testing.assert_array_equal(data["mode"], [expected_mode] * len(data["mode"]))
+
+
+def test__create_minimum_dtype_array():
+    """Test expected return types for minimum data types."""
+    result = utils._create_minimum_dtype_array([1, 2, 3], "uint8")
+    assert result.dtype == np.dtype("uint8")
+    # fallback to a generic array if the requested dtype can't be satisfied
+    result = utils._create_minimum_dtype_array(["a", "b", "c"], "uint8")
+    assert result.dtype == np.dtype("<U1")
--- a/imap_processing/utils.py
+++ b/imap_processing/utils.py
@@ -226,6 +226,84 @@ def create_dataset(
     return dataset
 
 
+def _get_minimum_numpy_datatype(  # noqa: PLR0912 - Too many branches
+    name: str, definition: xtcedef.XtcePacketDefinition
+) -> str:
+    """
+    Get the minimum datatype for a given variable.
+
+    Parameters
+    ----------
+    name : str
+        The variable name.
+    definition : xtcedef.XtcePacketDefinition
+        The XTCE packet definition.
+
+    Returns
+    -------
+    datatype : str
+        The minimum datatype.
+    """
+    data_encoding = definition.named_parameters[name].parameter_type.encoding
+
+    if isinstance(data_encoding, xtcedef.NumericDataEncoding):
+        nbits = data_encoding.size_in_bits
+        if isinstance(data_encoding, xtcedef.IntegerDataEncoding):
+            datatype = "int"
+            if data_encoding.encoding == "unsigned":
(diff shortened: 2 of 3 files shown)


tmplummer (commented):
This is great! I had identified this as something that needed to be worked on, but you got it implemented before I could even bring it up.

One nit pick on test coverage.

greglucas:
> One nit pick on test coverage.

@subagonsouth I don't see that, did you forget to submit that comment?

imap_processing/tests/test_utils.py
```diff
@@ -7,6 +7,10 @@
 
 from imap_processing import utils
 
+pytest_plugins = [
+    "imap_processing.tests.swapi.test_swapi_l1",
+]
```
greglucas: @subagonsouth, I used your plugins idea here to bring in an external semi-unrelated fixture that probably shouldn't be brought up into the main `conftest.py` level. Does this make sense to you here, or is there a different/better way to bring in external fixtures? Maybe you have an even better way of addressing the test coverage here though.

imap_processing/utils.py
```diff
@@ -226,6 +226,84 @@ def create_dataset(
...
+
+    return datatype
+
+
+def _create_minimum_dtype_array(values: list, dtype: str) -> np.ndarray:
+    """
+    Create an array with the minimum datatype.
+
+    If it can't be coerced to that datatype, fallback to general array creation
+    without a specific datatype. This can happen with derived values.
```
greglucas, line 287: Note that I added a parametrization over `use_derived_value` above, that actually found that generically assuming the XTCE datatype doesn't work when we go to derived values because we can go from `uint2` to `str` for enumerations and other cases. This function does the "simple" thing and just falls back to a default array creation if we can't say anything about what it will be derived to.

imap_processing/tests/test_utils.py
greglucas: So, this apparently didn't actually work the way I thought it did... It did work locally for me, but not remotely. Reading more about it on pytest it looks like it may not be supported to redefine `pytest_plugins` outside of the root `conftest.py`?

I ended up calling the function directly now instead of the fixture, which also allowed me to parametrize the test. Still curious about your thoughts on making this better though.

imap_processing/tests/test_utils.py
```diff
@@ -69,3 +69,29 @@ def test_convert_raw_to_eu(tmp_path):
...
+def test_packet_file_to_datasets(use_derived_value, expected_mode):
+    """
+    Test that all datatypes aren't all int64 and that we get
+    uint8/uint16 from header items as expected.
+
+    Test that we get multiple apids in the output.
+    """
+    test_file = "tests/swapi/l0_data/imap_swapi_l0_raw_20231012_v001.pkts"
+    packet_files = imap_module_directory / test_file
+    packet_definition = (
```
tmplummer, line 86: Suggestion for making this platform independent.
```suggestion
    packet_definition = str(
        imap_module_directory / "swapi/packet_definitions/swapi_packet_definition.xml"
        )
```

imap_processing/tests/test_utils.py
```diff
@@ -69,3 +69,29 @@ def test_convert_raw_to_eu(tmp_path):
         eu_dataset = utils.convert_raw_to_eu(
             dn_dataset.copy(), test_csv.absolute(), "PACKET_1", comment="#"
         )
+
+
+@pytest.mark.parametrize(
+    "use_derived_value, expected_mode", [(True, "HVENG"), (False, 2)]
+)
+def test_packet_file_to_datasets(use_derived_value, expected_mode):
```
tmplummer, line 77: Thanks for adding this test. My nit pick that got lost in the ether was suggesting that this test belonged in this file rather than in the swapi test. One additional improvement would be to have unit test coverage for `_get_minimum_numpy_datatype`.

tmplummer (approved):
This looks good. I commented about one additional unit test that could be added.

imap_processing/tests/test_utils.py
```diff
@@ -69,3 +69,29 @@ def test_convert_raw_to_eu(tmp_path):
...
+    """
+    Test that all datatypes aren't all int64 and that we get
+    uint8/uint16 from header items as expected.
+
+    Test that we get multiple apids in the output.
+    """
+    test_file = "tests/swapi/l0_data/imap_swapi_l0_raw_20231012_v001.pkts"
+    packet_files = imap_module_directory / test_file
```
tmplummer, line 85: Question: Does this work on Windows? I can't specifically find if dividing by a string with hard coded forward-slashes works is handled by pathlib or not.

greglucas: Yes, I believe it should. I of course don't have a Windows machine to test this on, but I believe I've run into this in CI before and as long as you start from a `Path` instance the forward slashes in strings are respected across OS' properly. Here is what our AI overlords tell me:

On Windows, Path.cwd() will give you the current working directory in a format like C:\Users\YourUsername\YourDirectory. When you divide this by "a/b/c.txt", it will result in a path like C:\Users\YourUsername\YourDirectory\a\b\c.txt.

On POSIX systems (Linux, macOS), Path.cwd() will give you the current working directory in a format like /home/YourUsername/YourDirectory. When you divide this by "a/b/c.txt", it will result in a path like /home/YourUsername/YourDirectory/a/b/c.txt.

greglucas: and speaking of which, we are testing it in CI, so I guess it is working as expected :)

imap_processing/tests/test_utils.py
greglucas: Good catch. Updated.

imap_processing/tests/test_utils.py
greglucas: I looked into this quick and couldn't see a great way of doing it. It'd be a deeply nested mock currently and basically just hard-coding expected results rather than actually testing against XTCE instance types. So, I'm going to punt on this because we are testing at least uint8 / uint16 as having multiple explicit return types in the current test.

I did do the easier version and test the other private function `_create_minimum_dataset` for the two cases there.

PR #725, opened 2024-07-30
Glows cdf attrs update

anamanica:
# Change Summary

## Overview
Changed glows l1a cdf variable attributes yaml file as requested by Marek and GLOWS team.

## New Dependencies

None

## New Files

None

## Deleted Files

None

## Updated Files

- imap_glows_l1a_variable_attrs.yaml
   - Updated the information in comparison to the file Marek and the GLOWS team sent to us.
- glows_l1a.py
   - Updated some changed variable names.

## Testing
None (Already covered in test_glows_l1a.py)


imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
-# Info found in cdf.defaults.py
-# Info defined from glows_cdf_attrs
 int_fillval: &int_fillval -9223372036854775808
+int_uint32: &int_uint32 4294967295
+int_uint32_min_one: &int_uint32_min_one 4294967294
```
maxinelasp: Rather than this, I would just use `*int_uint32 - 1` just because it seems a little overkill when `int_unit32 -1` makes a lot of sense, but that's a nitpick so feel free to ignore

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
+int_uint16: &int_uint16 65535
```
maxinelasp: I also think all these variables should be `max_uint16` etc

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
 min_epoch: &min_epoch -315575942816000000
 max_epoch: &max_epoch 3155630469184000000
 
-#Information taken from CDF warnings of missing attributes
 default_attrs: &default_attrs
-  # Resolution is optional?
   RESOLUTION: ' '
   DISPLAY_TYPE: no_plot
   TIME_BASE: J2000
```
maxinelasp, line 12: Could you add a TODO to remove these unneeded attributes once SAMMI is fixed?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
-  # Guess from other l1a sections, prob wrong
   REFERENCE_POSITION: Rotating Earth Geoid
-  # Guess from other l1a sections
   UNITS: ' '
   FILLVAL: *int_fillval
   VALIDMIN: *min_epoch
 
-# Information from glows_cdf_attrs.py
+epoch:
+  <<:  *default_attrs
```
maxinelasp: Do we need default_attrs on this? Actually, Tenzin had a PR recently about epoch - can you look at that and see if we should use that style instead of manually defining it here?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
-  DISPLAY_TYPE: time_series
   LABLAXIS: Counts
+  FILLVAL: -32768
+  MONOTON: INCREASE
+  SCALETYP: linear
 
-per_second_attrs:
+within_the_second: # Used to be per_second_attrs
   <<:  *default_attrs
   VALIDMMIN: 0
```
maxinelasp: Still a typo here

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
   FORMAT: I10
   VAR_TYPE: support_data
   DISPLAY_TYPE: time_series
-  LABLAXIS: Counts
+  LABLAXIS: Direct Events
 
-event_attrs:
+direct_event_components_attrs:
   <<:  *default_attrs
   VALIDMMIN: 0
```
maxinelasp: typo

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
+  FORMAT: I2
+  VAR_TYPE: support_data
+  LABLAXIS: Components
+  FILLVAL: 255
 
-direct_event_attrs:
+direct_events:
   <<:  *default_attrs
   DEPEND_O: epoch
   DEPEND_1: per_second
```
maxinelasp: Should be `within_the_second`

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
   FIELDNAM: Histogram of photon counts
-  # FORMAT is different from glows_cdf_attrs.py
-  # Changed this in order to avoid warnings
-  # Float with 4 digits
-  FORMAT: F7.4
+  FORMAT: I4
   DISPLAY_TYPE: time_series
   LABLAXIS: Counts
-  FILL_VAL: *int_fillval
+  FILL_VAL: *int_uint16
```
maxinelasp: We need to update the code to match this, did you do it or add a TODO comment?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
 imap_start_time:
   <<: *metadata_attrs
-  CATDESC: Start time of the block, in seconds, in IMAP spacecraft clock time
-  FIELDNAM: Start time in seconds in IMAP time
-  # FORMAT is different from glows_cdf_attrs.py
-  # Changed this in order to avoid warnings
-  # Can range from F0 to F0.9
+  CATDESC: Histogram start time, IMAP-clock seconds
+  FIELDNAM: Histogram start time, IMAP-clock seconds
+  FILLVAL: *int_fillval
```
maxinelasp, line 119: Is this right? It is a float

maxinelasp: Maybe -1?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
   <<: *metadata_attrs
-  CATDESC: End time of the block, in seconds, in IMAP spacecraft clock time
-  FIELDNAM: End time in seconds in IMAP time
-  # FORMAT is different from glows_cdf_attrs.py
-  # Changed this in order to avoid warnings
-  # Can range from F0 to F0.9
-  FORMAT: F16.6
+  CATDESC: Accumulation time in seconds for GLOWS histogram
+  FIELDNAM: Histogram accumulation time
+  FILLVAL: *int_fillval
```
maxinelasp, line 131: Same comment

maxinelasp: Actually, it might be good to make some kind of `time_defaults` for the shared bits here

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
 flags_set_onboard:
   <<: *metadata_attrs
-  CATDESC: FLags for bad data
-  FIELDNAM: Flags for missing or bad data
+  CATDESC: Binary mask with histogram flags set onboard
+  FIELDNAM: Mask with histogram flags set onboard
+  FILLVAL: *int_uint32
+  FORMAT: I6
+  LABLAXIS: Mask value
+  VALIDMAX: *int_uint16
```
maxinelasp: Max is uint16, fillval is uint32 - is this what they had in their doc? that seems weird to me. Could you add a todo to validate that with GLOWS?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
+  CATDESC: Number of spins per block during accumulation of histogram
   FIELDNAM: Number of spins per block
+  FILLVAL: 65535
+  FORMAT: I4
+  LABLAXIS: Num of spins
+  VALIDMAX: 256
+  VALIDMIN: 1
 
-number_of_spins_per_histogram:
+number_of_spins_per_histogram: # Should this be removed?
```
maxinelasp: If this isn't used anywhere it should be removed, I think it's duplicated from number_of_bins_per_histogram

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
-  FORMAT: F7.4
+  FORMAT: I4
   DISPLAY_TYPE: time_series
   LABLAXIS: Counts
-  FILL_VAL: *int_fillval
+  FILL_VAL: *int_uint16
   UNITS: counts
   VAR_TYPE: data
 
 metadata_attrs: &metadata_attrs
```
maxinelasp: can you a. move this up to the rest of the defaults and b. rename it to `support_data_defaults` to be more clear?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
 
 filter_temperature_variance:
   <<: *metadata_attrs
-  CATDESC: Filter temperature variance, per block, and unit encoded.
-  FIELDNAM: Uint encoded filter temperature variance
+  CATDESC: Uint-encoded spin-block-averaged variance of filter temperature
+  FIELDNAM: Variance of filter temperature
+  FILLVAL: *int_uint32
+  FORMAT: I6
+  LABLAXIS: Variance
```
maxinelasp, line 223: Can you change this to "Tmp Variance?"

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
```diff
@@ -1,358 +1,523 @@
...
 hv_voltage:
   <<: *metadata_attrs
-  CATDESC: Uint encoded voltage
-  FIELDNAM: Uint encoded voltage
+  CATDESC: Uint encoded CEM voltage
+  FIELDNAM: Uint encoded CEM voltage
+  FILLVAL: *int_uint32
+  FORMAT: I6
+  LABLAXIS: Voltage
+  VALIDMAX: 65635 # Is this a typo?
```
maxinelasp: I think this should should be the same as the rest of the UINT values. It might be worth pulling them out as a default grouping, but you already have them in each specific variable so it's up to you if that's worth doing

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -258,7 +260,7 @@ def generate_de_dataset(
             "per_second": per_second,
```
maxinelasp: per_second should be within_the_second here too

maxinelasp (commented):
Sorry for the delay in reviewing, I forgot 😅 some minor requested changes from me. Thanks for doing this!

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: No, I totally agree with you, it is quite redundant. Is there a way I can do this? Chat gpt informed me that .yaml files don't support operations like that (which could totally be wrong of course) 😅

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: Sure thing! It looks like we just need to remove the epoch variable dimension, and ImapCdfAttriutes will load in epoch from imap_constant_attrs.yaml.

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: I thought that we wanted to change `per_second_attrs` to `within_the_second`, and `direct_event_attrs` to just `direct_events`. Is this wrong?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: How do we need to change the code to match this exactly?

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: It does appear that this is the number Marek wrote in the file you sent me, but there is a little comment next to it. I added a TODO here instead of changing it.

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
anamanica: Yeah, I just double checked and that is in the document. I added that comment!

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
maxinelasp: Ah, I meant this specific line - instead of `DEPEND_1: per_second` it should be `DEPEND_1: within_the_second` to point to our new dimension

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -244,20 +244,20 @@ def generate_de_dataset(
...
+        name="within_the_second",
+        dims=["within_the_second"],
         attrs=glows_cdf_attributes.get_variable_attributes("within_the_second"),
     )
 
     de = xr.DataArray(
         direct_events,
         name="direct_events",
-        dims=["epoch", "per_second", "direct_event"],
+        dims=["epoch", "within_the_second", "direct_event"],
```
maxinelasp, line 257: So since this is updated here, the `DEPEND_n` values that depend on this coordinate also need to be updated to match.

maxinelasp (commented):
One minor change but otherwise this is looking pretty good!

imap_processing/cdf/config/imap_glows_l1a_variable_attrs.yaml
maxinelasp: I'm not sure, if it's not possible then it's fine how it is

maxinelasp (approved):
Looks good, thanks for doing this!

PR #726, opened 2024-07-31
MNT: Change to a period separator for XTCE definitions

greglucas:
# Change Summary

## Overview

This updates the XTCE generator to produce a period between the packet name and the mnemonic name, which helps for subsequent parsing.

Additionally, use that helpful fact to strip that leading packet name in the packet_file_to_datasets routine.

## Testing

I did test this out manually by modifying an XTCE file to ensure that the stripping works as expected, but there aren't any other packet definitions with a packet name to make this easily testable yet. Maybe we'll update some of the packet definitions soon to exercise this.


tech3371 (approved):
Thank you for adding that. I will try using XTCE that this produces in SWAPI after this PR.

tmplummer (approved):
LGTM

PR #731, opened 2024-08-06
Mypy Update

### LINKED ISSUE #710: BUG - mypy should only be run over our actual code

daralynnrhode:
# 

## Overview

Kinda long, so sorry for that. Feel free to only check the files that concern your instrument.

Ensuring that both pre-commit mypy and a local mypy check will return the same outputs. In doing so, more errors were found and will be fixed within this PR.

numpy is weird:
As a way to avoid redeclaring a type, ie. `indices: np.ndarray = np.where(valid)[0]` that should be inferred to be a `np.ndarray` because of what the `np.where()` returns, but is not. Instead function returns of `np.ndarray` were changed to `npt.NDArray` from the numpy typing library. I believe in most instance it will represent the same thing. 
This did however cause the build the docs to fail to recognize the type for some functions, but not all. Just added an ignore to the `config.py` file.

*`np.array` is not an object type, it is a function that returns an `np.ndarray`. `object: np.array = ` is not valid.

Closes #710

## Updated Files

- pyproject.toml
- pre-commit-config.yaml
- docs/source/conf.py

-

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -57,7 +57,9 @@ def met_to_j2000ns(
     time_array = (np.asarray(met, dtype=float) * 1e9).astype(np.int64)
     # Calculate the time difference between our reference system and J2000
     if isinstance(reference_epoch, np.datetime64):
-        place_holder: np.timedelta64 = (reference_epoch - J2000_EPOCH).astype("timedelta64[ns]")
+        place_holder: np.timedelta64 = (reference_epoch - J2000_EPOCH).astype(
+            "timedelta64[ns]"
+        )
         j2000_offset: np.int64 = place_holder.astype(np.int64)
         return j2000_offset + time_array
 
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -3,7 +3,6 @@
 import logging
 import re
 from pathlib import Path
-from typing import Optional
 
 import imap_data_access
 import numpy as np
@@ -25,8 +24,8 @@ J2000_EPOCH = np.datetime64("2000-01-01T11:58:55.816", "ns")
 
 def met_to_j2000ns(
     met: np.typing.ArrayLike,
-    reference_epoch: Optional[np.datetime64] = IMAP_EPOCH,
-) -> np.typing.ArrayLike:
+    reference_epoch: np.datetime64 = IMAP_EPOCH,
+) -> np.typing.NDArray[np.int64]:
     """
     Convert mission elapsed time (MET) to nanoseconds from J2000.
 
@@ -56,12 +55,10 @@ def met_to_j2000ns(
     #       to 32bit and overflow due to the nanosecond multiplication
     time_array = (np.asarray(met, dtype=float) * 1e9).astype(np.int64)
     # Calculate the time difference between our reference system and J2000
-    if isinstance(reference_epoch, np.datetime64):
-        place_holder: np.timedelta64 = (reference_epoch - J2000_EPOCH).astype(
-            "timedelta64[ns]"
-        )
-        j2000_offset: np.int64 = place_holder.astype(np.int64)
-        return j2000_offset + time_array
+    j2000_offset: np.typing.NDArray[np.datetime64] = (
+        reference_epoch - J2000_EPOCH
+    ).astype("datetime64[ns]")
+    return j2000_offset.astype(np.int64) + time_array
 
 
 def load_cdf(
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -56,10 +56,10 @@ def met_to_j2000ns(
     #       to 32bit and overflow due to the nanosecond multiplication
     time_array = (np.asarray(met, dtype=float) * 1e9).astype(np.int64)
     # Calculate the time difference between our reference system and J2000
-    j2000_offset = (
-        (reference_epoch - J2000_EPOCH).astype("timedelta64[ns]").astype(np.int64)
-    )
-    return j2000_offset + time_array
+    if isinstance(reference_epoch, np.datetime64):
+        place_holder: np.timedelta64 = (reference_epoch - J2000_EPOCH).astype("timedelta64[ns]")
+        j2000_offset: np.int64 = place_holder.astype(np.int64)
+        return j2000_offset + time_array
 
 
 def load_cdf(
--- a/imap_processing/mag/l0/mag_l0_data.py
+++ b/imap_processing/mag/l0/mag_l0_data.py
@@ -110,7 +110,7 @@ class MagL0:
         # big-endian bytes
         self.VECTORS = np.frombuffer(
             int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"),  # type: ignore[arg-type]
-            # MYPY Error: Argument 1 to "int" has incompatible type
+            # TODO Check MYPY Error: Argument 1 to "int" has incompatible type
             # "Union[ndarray[Any, Any], str]"; expected "Union[str, bytes, bytearray]"
             dtype=np.dtype(">b"),
         )
--- a/imap_processing/mag/l0/mag_l0_data.py
+++ b/imap_processing/mag/l0/mag_l0_data.py
@@ -109,7 +109,9 @@ class MagL0:
         # Convert string output from space_packet_parser to numpy array of
         # big-endian bytes
         self.VECTORS = np.frombuffer(
-            int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"),
+            int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"), #type: ignore[arg-type]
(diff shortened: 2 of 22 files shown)


imap_processing/cdf/utils.py
```diff
@@ -56,10 +56,12 @@ def met_to_j2000ns(
...
-    j2000_offset = (
-        (reference_epoch - J2000_EPOCH).astype("timedelta64[ns]").astype(np.int64)
-    )
-    return j2000_offset + time_array
+    if isinstance(reference_epoch, np.datetime64):
+        place_holder: np.timedelta64 = (reference_epoch - J2000_EPOCH).astype(
+            "timedelta64[ns]"
+        )
+        j2000_offset: np.int64 = place_holder.astype(np.int64)
+        return j2000_offset + time_array
```
greglucas: This seems like type checking imposing its will incorrectly, because if we pass in something that isn't an instance of datetime64 then we get a `None` return. I think it would be better to error actually. I have a suggestion below to make it a little nicer, although still not as nice as it was before IMO 🤷 .

```diff
diff --git a/imap_processing/cdf/utils.py b/imap_processing/cdf/utils.py
index fc2e46b..17a75e0 100644
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -25,8 +25,8 @@ J2000_EPOCH = np.datetime64("2000-01-01T11:58:55.816", "ns")
 
 def met_to_j2000ns(
     met: np.typing.ArrayLike,
-    reference_epoch: Optional[np.datetime64] = IMAP_EPOCH,
-) -> np.typing.ArrayLike:
+    reference_epoch: np.datetime64 = IMAP_EPOCH,
+) -> np.typing.NDArray[np.int64]:
     """
     Convert mission elapsed time (MET) to nanoseconds from J2000.
 
@@ -56,12 +56,10 @@ def met_to_j2000ns(
     #       to 32bit and overflow due to the nanosecond multiplication
     time_array = (np.asarray(met, dtype=float) * 1e9).astype(np.int64)
     # Calculate the time difference between our reference system and J2000
-    if isinstance(reference_epoch, np.datetime64):
-        place_holder: np.timedelta64 = (reference_epoch - J2000_EPOCH).astype(
-            "timedelta64[ns]"
-        )
-        j2000_offset: np.int64 = place_holder.astype(np.int64)
-        return j2000_offset + time_array
+
+    j2000_offset: np.typing.NDArray[np.datetime64] = \
+        (reference_epoch - J2000_EPOCH).astype("datetime64[ns]")
+    return j2000_offset.astype(np.int64) + time_array
 
 
 def load_cdf(

```

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -138,11 +139,12 @@ def find_sweep_starts(packets: xr.Dataset) -> np.ndarray:
...
     Returns
     -------
-    numpy.ndarray
+    indices_start : numpy.ndarray
         Array of indices of start cycle.
     """
     if packets["epoch"].size < 12:
-        return np.array([], np.int64)
+        indices_start = np.array([], np.int64)
+        return indices_start
```
greglucas: I think this can be reverted.
```suggestion
        return np.array([], np.int64)
```

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -172,10 +174,11 @@ def find_sweep_starts(packets: xr.Dataset) -> np.ndarray:
         & ione[9:-1]
         & ione[10:]
     )
-    return np.where(valid)[0]
+    indices = np.where(valid)[0]
+    return indices
```
greglucas: This one can be reverted too.
```suggestion
    return np.where(valid)[0]
```

imap_processing/swe/l1a/swe_science.py
```diff
@@ -112,7 +112,7 @@ def swe_science(decom_data: list, data_version: str) -> xr.Dataset:
     science_array = []
     raw_science_array = []
 
-    metadata_arrays: np.array = collections.defaultdict(list)
+    metadata_arrays: np.array = collections.defaultdict(list)  # type: ignore[valid-type]
```
greglucas: This was an incorrect type previously, if you change this to:
```suggestion
    metadata_arrays: dict[list] = collections.defaultdict(list)
```
Then you can remove a few of the other type ignores later in the file too.

imap_processing/swe/l1a/swe_science.py
```diff
@@ -133,7 +133,8 @@ def swe_science(decom_data: list, data_version: str) -> xr.Dataset:
         # where 1260 = 180 x 7 CEMs
         # Take the "raw_counts" indices/counts mapping from
         # decompression_table and then reshape the return
-        uncompress_data = np.take(decompression_table, raw_counts).reshape(180, 7)
+        place_holder: np.ndarray = np.take(decompression_table, raw_counts)
+        uncompress_data = place_holder.reshape(180, 7)
```
greglucas: Rather than the placeholder variables to appease the type checkers, I think I'd prefer to just ignore them. Not sure how you and others feel about that though, but I personally find this harder to read than just ignoring the type checking error.

`# type: ignore[attr-defined]`

greglucas (commented):
🎉 This is fantastic! Thank you for updating all of this and the persistence to clean it all up.

I have a few minor nitpicks where I don't like what the type checker made you do and added some suggestions for what I think might be easier/better.

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -109,7 +109,9 @@ def __post_init__(self) -> None:
         # Convert string output from space_packet_parser to numpy array of
         # big-endian bytes
         self.VECTORS = np.frombuffer(
-            int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"),
+            int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"),  # type: ignore[arg-type]
+            # MYPY Error: Argument 1 to "int" has incompatible type
```
bourque:
```suggestion
            # TODO: MYPY Error: Argument 1 to "int" has incompatible type
```

The "todo" keyword gets specially highlighted in PyCharm (and hopefully other editors too?), so I think its good to include it (like you have in other modules)! That way it should pop out to the appropriate developer.

pyproject.toml
```diff
@@ -118,6 +118,13 @@ checks = ["all", #report on all checks, except the following
...
+disable_error_code = ['import-not-found', # Unable to locate module or package specified
+                'import-untyped', # The import lacks type information
+                'no-untyped-call', # Function calls are only made to functions that are fully typed
+                'type-arg' ] # Requires type arguments to be specified in list[] and dict[]
+strict = true
+explicit_package_bases = true
+follow_imports = 'skip' #may want to remove
+exclude = ["tests"]
+packages = ["imap_processing" , "tools"]
+plugins = 'numpy.typing.mypy_plugin'
```
bourque, line 130: That's great that you were able to define these here instead of `pre-commit-config`. I think its this team's preference to have all of these configurations under the one `pyproject.toml` umbrella!

bourque (approved):
Just a few non-blocking comments/suggestions from me. Thank you for doing this work -- you really had to get into the weeds in this one! It is much appreciated!

greglucas (approved):

imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -109,7 +109,9 @@ def __post_init__(self) -> None:
         # Convert string output from space_packet_parser to numpy array of
         # big-endian bytes
         self.VECTORS = np.frombuffer(
-            int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"),
+            int(self.VECTORS, 2).to_bytes(len(self.VECTORS) // 8, "big"),  # type: ignore[arg-type]
+            # TODO Check MYPY Error: Argument 1 to "int" has incompatible type
```
maxinelasp, line 113: I can check on this, I'm updating things related to this code soon

maxinelasp (approved):
Looks good, thanks for doing this!

PR #733, opened 2024-08-06
Define the imap_processing spice structure

### LINKED ISSUE #732: Create module struture for spice functions
### Algorithm Description:
In the vein of breaking things up into small PRs, this is simply adding some files that define the organization of shared spice related functions. These files include: kernel_utils.py, time.py, and geometry.py

tmplummer:
Add files defining the structure of the spice package within imap_processing

# Change Summary

## Overview
This PR adds some empty files which define how SPICE related functions will be organized.

## New Dependencies
None

## New Files
   - Functions for computing geometry using SPICE.
- imap_processing/spice/kernels.py
   - Functions for furnishing and tracking SPICE kernels.
   - Time conversion functions that rely on SPICE.
   - Tests coverage for imap_processing/spice/geometry.py
- imap_processing/tests/spice/test_kernels.py
   - Tests coverage for imap_processing/spice/kernels.py
   - Tests coverage for imap_processing/spice/time.py

## Deleted Files
None

## Updated Files
None

## Testing
None

Closes #732


greglucas (approved):
I don't know enough about SPICE to know that this is the directory structure we want or not. Seems reasonable though.

My only comment is that I would actually prefer slightly larger PRs that make use of these files :) Or maybe one-by-one adding the `kernels.py` and `test_kernels.py` in separate PRs. But, I do appreciate the thought in trying to break it up.

greglucas:
As a quick comment: we have some work already done in `tools/` https://github.com/IMAP-Science-Operations-Center/imap_processing/tree/dev/tools/spice
where does that work fit into this? Are we going to be merging everything together and only having things in the deployed package, or are those good examples to keep separate?

tmplummer:
@greglucas 
> As a quick comment: we have some work already done in `tools/` https://github.com/IMAP-Science-Operations-Center/imap_processing/tree/dev/tools/spice
> where does that work fit into this? Are we going to be merging everything together and only having things in the deployed package, or are those good examples to keep separate?

Right... I am not sure about why all of that code was located there. It looks to me like some of that code can be pulled in to the imap_processing package. Is the `tools` directory intended to contain development tools that are not needed/wanted in the deployed package?


> My only comment is that I would actually prefer slightly larger PRs that make use of these files :) Or maybe one-by-one adding the `kernels.py` and `test_kernels.py` in separate PRs. But, I do appreciate the thought in trying to break it up.

As I was working on the kernel handling decorator, I realized that I am working on one piece, Laura is working on another and Veronica is working on a third. I thought would be prudent to define an organization for SPICE related functions that has worked on other projects so that the three of us can be consistent. That is the main reason that I broke this PR out and added files with nothing in them.

greglucas:
That all makes sense to me and you still have my approval :)

> Is the tools directory intended to contain development tools that are not needed/wanted in the deployed package?

Correct. Not bringing unnecessary dependencies.

PR #734, opened 2024-08-06
728 spice kernel furnishing decorator

### LINKED ISSUE #728: SPICE kernel furnishing decorator
### Algorithm Description:
Implement a decorator that will handle the furnishing of spice kernels. Base this on EMUS but strip out the SPICE API code. This will start fairly basic to just furnish spice metakernel/kernels that we have.

Also implement a decorator that will set the environment variable to point to the test SPICE metakernel that we have from Nick.

tmplummer:
# Change Summary

## Overview
Sorry for the size of this PR. I don't think that it is possible to break it up any smaller.  The most important piece is the `ensure_spice` decorator. I think that the docstrings are descriptive enough for you all to understand what it does without additional clarification, but let me know if there are any questions. All other changes have to do with setting up testing infrastructure needed to test the decorator.

## New Dependencies
None

## New Files
- imap_processing/tests/spice/test_data/imap_sclk_0000.tsc
   - test sclk kernel
- imap_processing/tests/spice/test_data/imap_test_metakernel.template
   - test metakernel template that is used in the fixture that inserts the local absolute path so that tests work anywhere
- imap_processing/tests/spice/test_data/naif0012.tls
   - test leapsecond kernel

## Deleted Files
None

## Updated Files
- imap_processing/spice/kernels.py
   - Added ensure_spice decorator which will handle automatic kernel furnishing when needed
   - Added several testing fixtures used to set up spice kernel needs
   - Of particular note is the autouse=true fixture that was added to ensure spice kernels are always cleared after each test
-  imap_processing/tests/spice/test_kernels.py
   - Added tests coverage for ensure_spice decorator with different use cases

## Testing
Test coverage for `ensure_spice` decorator

closes #728


imap_processing/spice/kernels.py
```diff
@@ -0,0 +1,158 @@
...
+        ----------
+        func : Callable
+            The function to be decorated/wrapped.
+
+        Returns
+        -------
+        Callable
+            If used as a function wrapper, the decorated function is returned.
+        """
+
```
greglucas, line 100: There are a lot of paths in here, so I'm not sure I totally follow all the flow, but I _think_ all of this can be removed and dedented and the wrapper returned directly? So your final lines of the decorator would be this:
```python
    return func(*args, **kwargs)
return wrapper_ensure_spice
```
without needing to check for callables because your returns would take care of them directly.

imap_processing/spice/kernels.py
```diff
@@ -0,0 +1,158 @@
...
+            """
+            try:
+                # Step 1.
+                return func(
+                    *args, **kwargs
+                )  # Naive first try. Maybe SPICE is already furnished.
+            except SpiceyError as spicey_err:
+                try:
+                    # Step 2.
+                    metakernel_path = os.environ["SPICE_METAKERNEL"]
```
greglucas, line 126: Is this standard? Does this mean we are _only_ allowing metakernels and not any other way of loading individual kernels?

imap_processing/tests/conftest.py
```diff
@@ -15,6 +19,119 @@ def _set_global_config(monkeypatch, tmp_path):
...
 
+@pytest.fixture(scope="session")
+def monkeypatch_session():
+    from _pytest.monkeypatch import MonkeyPatch
+
+    m = MonkeyPatch()
+    yield m
+    m.undo()
+
+
```
greglucas: This isn't being used currently, can we remove it? It looks like you are using the monkeypatch fixture directly below instead of the "session" based one here.
\

imap_processing/tests/conftest.py
```diff
@@ -15,6 +19,119 @@ def _set_global_config(monkeypatch, tmp_path):
...
+def furnish_test_lsk(spice_test_data_path):
+    """Furnishes (temporarily) the testing LSK"""
+    test_lsk = spice_test_data_path / "naif0012.tls"
+    spice.furnsh(test_lsk)
+    yield test_lsk
+    spice.kclear()
+
+
+@pytest.fixture()
+def furnish_sclk(spice_test_data_path):
```
greglucas, line 53: Looks like this fixture isn't used, do you want to use it here, planning on needing it later? (I assume it might come in future PR of `test_time`, but just wanted to call it out)

imap_processing/tests/conftest.py
```diff
@@ -15,6 +19,119 @@ def _set_global_config(monkeypatch, tmp_path):
...
+    """Furnishes (temporarily) the SCLK for JPSS stored in the package data directory"""
+    test_sclk = spice_test_data_path / "imap_sclk_0000.tsc"
+    spice.furnsh(test_sclk)
+    yield test_sclk
+    spice.kclear()
+
+
+@pytest.fixture()
+def use_test_metakernel(tmpdir_factory, monkeypatch, spice_test_data_path):
```
greglucas: Do we want to make use of any of the `IMAP_DATA` directory locations for spice file paths?
https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/ff6c637533d39a8c5359f2e456e918f2495b0871/imap_data_access/file_validation.py#L354

You might be able to use that location as your tmpdir by accessing `imap_data_access.config["DATA_DIR"] / imap/spice/...` .
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/b9abcb35e0dbc20a6afc5d80f6bf114200f804d3/imap_processing/tests/conftest.py#L10-L15

imap_processing/tests/spice/test_kernels.py
```diff
@@ -0,0 +1,63 @@
...
+        double_wrap_et2utc,
+        double_wrap_et2utc_tk_only,
+    ],
+)
+def test_ensure_spice_emus_mk_path(func, use_test_metakernel):
+    """Test functionality of ensure spice with SPICE_METAKERNEL set"""
+    assert func(577365941.184, "ISOC", 3) == "2018-04-18T23:24:31.998"
+
+
+def test_ensure_spice_time_kernels():
```
greglucas, line 52: Are these next two tests failing because we aren't using the fixture `use_test_metakernel`? It might be good to add a quick comment about that if so.

imap_processing/spice/kernels.py
tmplummer: If kernels are already furnished, then the `Step 1` code will call the function and proceed. This allows for custom/manual furnishing of whatever kernels you want. If they are not furnished, then the secondary default behavior is to look for a metakernel pointed to by this environment variable. We can certainly tweak this functionality as we define details about how we want kernels to be furnished in the pipeline.

It seems like I should put some slides together and set up a meeting to discuss this decorator and the logic flow with the team.

imap_processing/tests/conftest.py
tmplummer: I do think that it will be useful later, especially in `test_time`. Mostly wanted to put it in to help others with testing fixtures that they might need.

imap_processing/tests/spice/test_data/imap_sclk_0000.tsc
tmplummer, line 1: There is already a clock kernel in the tools directory. I need to compare to this one and keep only one.

imap_processing/tests/spice/test_data/naif0012.tls
tmplummer, line 1: There is already a leapsecond kernel checked in in the tools directory. We should have only on in the repo. I will either move the other one or point to it in the metakernel template.

imap_processing/spice/kernels.py
laspsandoval: +1 I think an overview of what is happening in ppt would be helpful. 

Would we not know if kernels were already furnished? Could we use this

`with spice.KernelPool(kernels):`

Instead of this?
 `spice.furnsh(test_sclk)
    yield test_sclk
    spice.kclear()`

laspsandoval (commented):
Tenzin is using sds-data-manager/lambda_code/efs_lambda/lambda_function.py to figure out which attitude and ephemeris kernels to use for the symlink. I think getting together to discuss your code and hers and how they work together will be important for this PR.

Another thing that I have been thinking about lately is what about the additional kernels that are not part of the metakernels?
- de430.bsp for example I don't think will be delivered to us by the MOC. Something to think about.

It looks good though. A very nice start. Looking forward to discussing. You have boldly gone where none of us dared to go. Nicely done.

imap_processing/spice/kernels.py
greglucas: Sorry for the noise here. I see that I missed the docstring about two paths with keyword arguments or without. Resolving this, but if we don't need the `time_kernels_only` argument on IMAP, then it could be simplified to avoid so much nesting.

To me, the `time_kernels_only` seems like you might be doing too much in this one function and a separate function specifically for `retrieve_time_kernels()` could be warranted in that case to avoid any confusion about "ensuring spice" doing more than just ensuring, it would also be downloading.

tmplummer: Do we need the`time_kernels_only` argument on IMAP? In l1a processing, we need to be able to convert SHCOARSE to J2000 seconds which will require us to furnish an sclk and leapsecond kernel. We could achieve that by furnishing a full metakernel that also provides all the kernels required for computing geometry but which will not be used in l1a. Doing it that way makes is slightly harder to document the SPICE kernels that were used in processing because unused kernels are being furnished. I can certainly see that we could write a separate decorator that is specific to functions that only require the time kernels. I don't know if it is preferable to have two double nested decorators instead of one triple nested decorator.

There are so many interconnected decisions on implementation that ultimately link to the decisions we make about how we will manage and use kernels. I need to present the two main options that I am aware of to the team so that you all can help with that decision. In the meantime, this decorator is meant to be a starting point that will allow us to develop functions that use SPICE without having to think about  the intricacies of furnishing required kernels.

imap_processing/tests/conftest.py
tmplummer: Hmmm, maybe. I would like to add this to a discussion. I'm not sure I fully grok what you are suggesting but think that it might work better than using this template metakernel.

imap_processing/spice/kernels.py
greglucas: That all makes sense to me. I honestly just don't know enough about any of it, so your recommendations / suggestions are fantastic. I see your point about only needing the time kernels for some levels and only limiting the documentation to that. I was thinking that we were just documenting the "metakernel" in a sense and dumping everything. Does it also make it slower to load all the geometry/attitude kernels if they aren't used?

I also agree that this is a really nice starting point 👍

imap_processing/tests/conftest.py
greglucas: Sorry, I was not very clear above. I meant to write your template metakernel to that `DATA_DIR` location (so just pointing to a different tmpdir location, not changing any logic). The `DATA_DIR` in tests is already set to a temporary directory, so it'd just avoid another temporary directory/location and possibly be closer to where we want the kernels during the mission I think.

imap_processing/spice/kernels.py
```diff
@@ -1 +1,158 @@
...
+    3. An explicit wrapper function, providing a dynamically set value for
+    parameters, e.g. time_kernels_only
+        >>> wrapped = ensure_spice(spicey_func, time_kernels_only=True)
+        ... result = wrapped(*args, **kwargs)
+    """
+    if f_py and not callable(f_py):
+        raise ValueError(
+            f"Received a non-callable object {f_py} as the f_py argument to"
+            f"ensure_spice.  f_py must be a callable object."
+        )
```
greglucas, line 84: FYI that my editor says this is unreachable code, so I think it could be removed. i.e. you don't decorate something that isn't callable.

imap_processing/spice/kernels.py
greglucas: We have also been putting our global configuration / env variables in the `imap-data-access` repository. I'm not sure if the SPICE configuration options should go over there too.

imap_processing/tests/conftest.py
```diff
@@ -18,3 +22,107 @@ def _set_global_config(monkeypatch, tmp_path):
...
+    """Automatically clears out all SPICE remnants after every single test to
+    prevent the kernel pool from interfering with future tests. Option autouse
+    ensures this is run after every test."""
+    yield
+    spice.kclear()
+
+
+@pytest.fixture(scope="session")
+def spice_test_data_path(imap_tests_path):
+    return imap_module_directory.parent / "tools/tests/test_data/spice"
```
greglucas: I _think_ we might want to move these data files into the main package. I'm pretty sure if we released this, the tests would get included in the release, but the test data wouldn't so you wouldn't be able to test the actual sdist because the test data would be missing.

greglucas (approved):

imap_processing/spice/kernels.py
vmartinez-cu: +1 on an overview. That would be really helpful

vmartinez-cu:
Nice job on the detailed docstrings!

imap_processing/spice/kernels.py
tmplummer: Loading kernels has very little overhead. SPICE really just reads some metadata from the kernel that is needed to know which kernel to use when it comes time.

Documenting kernels is another interconnected decision to make. If we just want to furnish the metakernels (we will have one attitude metakernel and one ephemeris metakernel) and document that those were furnished, that works. It is possible to identify exactly which kernels affect each product and document just that subset of kernels too. On EMM, the SDC built the capability to generate a custom metakernel for a given time range that only included the specific kernels which provide coverage for that time range.

PR #743, opened 2024-08-07
packet_file_to_datasets *NOT* in IDEX

anamanica:
# Change Summary

## Overview

Originally, this PR was intended to switch over IDEX to using the `packet_file_to_datasets` function established in utils.py, however it was decided that this would be more work than it is worth. Due to the nature of the `.xml` file for IDEX, using this function created dimension issues. For now, IDEX will continue to use PacketParser, and RawDustEvents

I added a comment in the doc-string for `packet_file_to_datasets` to explain this.

## New Dependencies

None

## New Files

None

## Deleted Files

None

## Updated Files

- utils.py
   - Added comment to `packet_file_to_dataset` docstring.

## Testing

None

closes 687 for IDEX only!

### DIFF
--- a/imap_processing/utils.py
+++ b/imap_processing/utils.py
@@ -319,6 +319,11 @@ def packet_file_to_datasets(
     packet definition, the ``derived_value`` will be the converted value.
     The dimension of the dataset will be the time field in J2000 nanoseconds.
 
+    Note: While this is a general function intended to work for all
+    instruments, this cannot be used for IDEX due to the complex nature of
+    the xml file. IDEX will continue to use PacketParser and RawDustEvents
+    found in idex_l1.py.
+
     Parameters
     ----------
     packet_file : str
--- a/imap_processing/utils.py
+++ b/imap_processing/utils.py
@@ -319,11 +319,6 @@ def packet_file_to_datasets(
     packet definition, the ``derived_value`` will be the converted value.
     The dimension of the dataset will be the time field in J2000 nanoseconds.
 
-    Note: While this is a general function intended to work for all
-    instruments, this cannot be used for IDEX due to the complex nature of
-    the xml file. IDEX will continue to use PacketParser and RawDustEvents
-    found in idex_l1.py.
-
     Parameters
     ----------
     packet_file : str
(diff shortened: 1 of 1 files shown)


bourque (approved):
Thank you for trying to make this work! It sounds like you learned some things along the way so that is good!

imap_processing/utils.py
```diff
@@ -319,6 +319,11 @@ def packet_file_to_datasets(
     packet definition, the ``derived_value`` will be the converted value.
     The dimension of the dataset will be the time field in J2000 nanoseconds.
 
+    Note: While this is a general function intended to work for all
+    instruments, this cannot be used for IDEX due to the complex nature of
+    the xml file. IDEX will continue to use PacketParser and RawDustEvents
+    found in idex_l1.py.
```
greglucas: I think this should be either a `Notes` section in numpydoc, https://numpydoc.readthedocs.io/en/latest/format.html#notes or a `.. note::` rst callout.
https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

Some minor wordsmithing, where I'd avoid saying what we'll do with the specific classes in `idex_l1.py` in case someone refactors that later and forgets to update this docstring.
```suggestion
    .. note::
      This function only handles packet definitions with the same variable structure across
      all packets with the same ApId. For example, this cannot be used for IDEX due to the
      conditional XML structure defined for their science packet.
```

One other thought that occurred to me is whether we could catch this and print out a nice warning message here so others don't also hit this roadblock. Something like creating a new dictionary that keeps track of `apid: variables`. Around l365 below I think something like this might work?

```python
            if apid not in data_dict:
                # This is the first packet for this APID
                data_dict[apid] = collections.defaultdict(list)
                datatype_mapping[apid] = dict()
                variable_mapping[apid] = packet.data.keys()
             if packet.data.keys() != apid_variable_mapping[apid]:
                 raise ValueError(f"Cannot parse apid {apid} which has different variables defined across the same packet type")
```

greglucas (approved):
Looks good to me. Some very minor nits, that could be follow-on PRs if desired too. I think its great to add documentation about the limitations, so thank you for adding this!

imap_processing/utils.py
anamanica: Sweet! Yeah, I agree with you. I changed the comment to a "Notes" section in the doc-string. 

I didn't end up implementing the dimension check in this PR (I just don't quite understand the code yet, and I feel like I should before I add it and test it). Would you like me to write another issue ticket?

greglucas: Yeah, that'd be great. I think the code above is close to working, so if you copy that over to another issue and work from that it might be a good thing. Additionally, add a test trying to parse IDEX and showing that your `ValueError` is raised with `pytest.raises(ValueError, match="Cannot prase apid")`.

PR #750, opened 2024-08-14
Add Pulse Height Calculations

laspsandoval:
# Change Summary

## Overview
Adds calculations for PH for TOF and back positions.

## Updated Files
- ultra_l1b_extended.py
   - Calculate back xb, yb position and tof.
- lookup_utils.py
   - Minor formatting.

## Testing
 test_ultra_l1b_extended.py
   - Test data for xb, yb position. No TOF test data available.

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
@@ -5,6 +5,7 @@ import pandas as pd
 import pytest
 
 from imap_processing.ultra.l1b.ultra_l1b_extended import (
+    StopType,
     get_front_x_position,
     get_front_y_position,
     get_path_length,
@@ -77,8 +78,10 @@ def test_get_ph_tof_and_back_positions(
     )
 
     ph_indices = np.where(
-        (de_dataset["STOP_TYPE"] == 1) | (de_dataset["STOP_TYPE"] == 2)
+        np.isin(de_dataset["STOP_TYPE"], [StopType.Top.value, StopType.Bottom.value])
     )[0]
+
     selected_rows = df_filt.iloc[ph_indices]
+
     np.testing.assert_array_equal(ph_xb, selected_rows["Xb"].astype("float"))
     np.testing.assert_array_equal(ph_yb, selected_rows["Yb"].astype("float"))
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
@@ -1,5 +1,6 @@
 """Tests Extended Raw Events for ULTRA L1b."""
 
+import numpy as np
 import pandas as pd
 import pytest
 
@@ -7,6 +8,7 @@ from imap_processing.ultra.l1b.ultra_l1b_extended import (
     get_front_x_position,
     get_front_y_position,
     get_path_length,
+    get_ph_tof_and_back_positions,
 )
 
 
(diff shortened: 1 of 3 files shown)


imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -111,6 +123,115 @@ def get_front_y_position(start_type: ndarray, yb: ndarray) -> tuple[ndarray, nda
...
+    tof : np.array
+        Time of flight (tenths of a nanosecond).
+    t2 : np.array
+        Particle time of flight from start to stop (tenths of a nanosecond).
+    xb : np.array
+        Back positions in x direction (hundredths of a millimeter).
+    yb : np.array
+        Back positions in y direction (hundredths of a millimeter).
+    """
+    indices = np.where(
```
tmplummer: numpy recommends using np.nonzero here.
```suggestion
    indices = np.nonzero(
```

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -111,6 +123,115 @@ def get_front_y_position(start_type: ndarray, yb: ndarray) -> tuple[ndarray, nda
...
+
+    # particle_tof (t2) used later to compute etof
+    t2 = np.zeros(len(de_dataset["STOP_TYPE"]))
+    tof = np.zeros(len(de_dataset["STOP_TYPE"]))
+
+    # Stop Type: 1=Top, 2=Bottom
+    # Convert converts normalized TDC values into units of
+    # hundredths of a millimeter using lookup tables.
+    index_top = indices[de_dataset["STOP_TYPE"].data[indices] == StopType.Top.value]
+    stop_type_top = de_dataset["STOP_TYPE"].data[indices] == StopType.Top.value
```
tmplummer: I'm trying to understand the difference between `index_top` and `stop_type_top`.
It seems like `index_top` contains indices to where `STOP_TYPE == StopType.Top.value`. i.e. `index_top = np.nonzero(de_dataset["STOP_TYPE"].data == StopType.Top.value`?
And `stop_type_top` is a mask of where indices points to locations with `StopType.Top`.

I dunno maybe I am going too deep but I was trying to see if this code could be simplified to be more clear.

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -59,3 +62,26 @@ def test_get_path_length(de_dataset, yf_fixture):
...
+    """Tests get_ph_tof_and_back_positions function."""
+
+    df = pd.read_csv(events_fsw_comparison_theta_0)
+    df_filt = df[df["StartType"] != -1]
+
+    _, _, ph_xb, ph_yb = get_ph_tof_and_back_positions(
+        de_dataset, df_filt.Xf.astype("float").values, "ultra45"
+    )
+
+    ph_indices = np.where(
```
tmplummer:
```suggestion
    ph_indices = np.nonzero(
```

tmplummer (approved):
LGTM. This is a complex function! Looks like you have good test coverage for it. Couple of suggestions about using `np.nonzero` instead of `np.where`.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -111,6 +123,115 @@ def get_front_y_position(start_type: ndarray, yb: ndarray) -> tuple[ndarray, nda
     return np.array(d), np.array(yf)
 
 
+def get_ph_tof_and_back_positions(
+    de_dataset: xarray.Dataset, xf: np.ndarray, sensor: str
+) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
```
greglucas, line 128: No need to do this, but thought I'd throw this suggestion out in case you like it.

Would it be useful to stick to datasets since these are all the same size and in a sense grouped together? Then you could return a single object that contains all of these arrays.

```suggestion
def get_ph_tof_and_back_positions(
    de_dataset: xarray.Dataset, xf: np.ndarray, sensor: str
) -> xarray.Dataset:
```

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -111,6 +123,115 @@ def get_front_y_position(start_type: ndarray, yb: ndarray) -> tuple[ndarray, nda
...
+    back of the sensor are measured using the timing of the pulses.
+    Further description is available on pages 32-33 of
+    IMAP-Ultra Flight Software Specification document
+    (7523-9009_Rev_-.pdf).
+
+    Parameters
+    ----------
+    de_dataset : xarray.Dataset
+        Data in xarray format.
+    xf : np.array
```
greglucas, line 146: What is "xf" here. Is it a fixed value of the front position? Is it a part of the original de_dataset and you could index into that? How do we know it is the same shape as de_dataset (or maybe it isn't).

imap_processing/ultra/l1b/ultra_l1b_extended.py
greglucas: Agreed with nonzero instead of where.

But, I'll leave another suggestion as well, which is to filter the dataset to create a subset dataset and work on that. https://docs.xarray.dev/en/latest/generated/xarray.Dataset.isel.html#xarray.Dataset.isel

```python
de_filtered = de_dataset.isel(STOP_TYPE=[StopType.Top.value, StopType.Bottom.value])
```

Then work on  `de_filtered` below without needing to worry about the indices because it is already subset.

```python
sp_n_norm = get_norm(de_dataset["STOP_NORTH_TDC"], "SpN", sensor)
# or add this into the dataset as a new variable (maybe keep the same variable name and it has just been normalized?)
de_filtered["SP_N_NORM"] = get_norm(de_filtered["STOP_NORTH_TDC"], "SpN", sensor)
```

PR #757, opened 2024-08-19
I-ALiRT - Test decom for HIT CCSDS

### LINKED ISSUE #740: I-ALiRT: HIT CCSDS Packet Generation
### Requirements:

> **Specify: (optional)**

### Code:

```python
# Code
```

### LINKED ISSUE #752: I-ALiRT: HIT Algorithm Lambda
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Test decom for HIT CCSDS

## New Files

imap_processing/ialirt/packet_definitions/ialirt_hit.xml
 - packet definitions for hit packet that is generated onboard the spacecraft

imap_processing/ialirt/l0/decom_ialirt.py
 - name change for time_keys based on new packet definitions

imap_processing/ialirt/packet_definitions/ialirt.xml
 - update to packet definitions

## Testing
imap_processing/tests/ialirt/unit/test_decom_ialirt_hit.py
 - Verification that test data was properly parsed.

Test data:
hit_ialirt_sample.ccsds
hit_ialirt_sample.csv
hit_ialirt_sample_header.csv


laspsandoval:
I think that the next step for this is to populate HIT test data into the I-ALiRT SC packet. That way I can use it for testing architecture. Please comment on this idea.

imap_processing/tests/ialirt/unit/test_decom_ialirt_hit.py
```diff
@@ -0,0 +1,144 @@
...
+        imap_module_directory
+        / "tests"
+        / "ialirt"
+        / "test_data"
+        / "l0"
+        / "hit_ialirt_sample.csv"
+    )
+    data = pd.read_csv(
+        data_path, header=None, names=header, na_values=[" ", ""]
+    ).astype("float")
```
greglucas: * Can you manually add the header onto the sample data? This is a bit confusing to read in two different files, grabbing the columns from one dataset and adding it to another.
* I don't think we want "float", these are integer values being read in. pandas read_csv should auto-infer the types of columns for you, so can we just let that do its thing?

imap_processing/tests/ialirt/unit/test_decom_ialirt_hit.py
```diff
@@ -0,0 +1,144 @@
...
+    )
+    prefix_slow_rate.extend([f"ERATE_{i}" for i in range(1, 6)])
+    prefix_slow_rate.extend(["L12A", "L123A", "PENA", "L12B", "L123B", "PENB"])
+    prefix_slow_rate.extend(
+        ["SLOW_RATE_51", "SLOW_RATE_52", "SLOW_RATE_53", "SLOW_RATE_54"]
+    )
+    prefix_slow_rate.extend(["H_06_08", "H_12_15", "H_15_70", "HE4_06_08", "HE4_15_70"])
+
+    # The sequence begins where "HIT_MET" != 0
+    start_index = hit_test_data.index[hit_test_data["MET"] != 0][0]
```
greglucas: Why only test the first sequence?

greglucas (commented):
This is an interesting case because we won't ever get these packets, these are only for the spacecraft.

You currently have all of your decoding logic in the tests, you should move the calculations into the actual code and then only call the functions within the test.

Can you switch over to using the `packet_file_to_datasets()` helper function here as well?

greglucas:
> I think that the next step for this is to populate HIT test data into the I-ALiRT SC packet. That way I can use it for testing architecture. Please comment on this idea.

This seems like it is a good idea, but could also be complicated because you're having to inject specific bytes into another specific bytes object. If it is easy, it is likely worth doing, but I wouldn't spend too much time on it either.

More important in my opinion is writing a generic function that can grab the proper HIT values you need.
```python
def process_hit(xarray_dataset):
    # do something with ds["HIT_EVENT_DATA_00"], ...
    # This should be agnostic of whether you are passing in only the HIT CCSDS decoded dataset or the I-ALiRT decoded dataset.
```

imap_processing/tests/ialirt/test_data/l0/hit_ialirt_sample_header.csv
```diff
@@ -0,0 +1 @@
...T_TRIG_08,B_EVNT_TRIG_08,L3B_TRIG_08,L1B_TRIG_09,IB_EVNT_TRIG_09,B_EVNT_TRIG_09,L3B_TRIG_09,L1B_TRIG_10,IB_EVNT_TRIG_10,B_EVNT_TRIG_10,L3B_TRIG_10,L1B_TRIG_11,IB_EVNT_TRIG_11,B_EVNT_TRIG_11,L3B_TRIG_11,L1B_TRIG_12,IB_EVNT_TRIG_12,B_EVNT_TRIG_12,L3B_TRIG_12,L1B_TRIG_13,IB_EVNT_TRIG_13,B_EVNT_TRIG_13,L3B_TRIG_13,L1B_TRIG_14,IB_EVNT_TRIG_14,B_EVNT_TRIG_14,L3B_TRIG_14,L1B_TRIG,IB_EVNT_TRIG,B_EVNT_TRIG,L3B_TRIG,SLOW_RATE,L1A,L2A,L3A,L1A0AHG,L1B0AHG,L1C0AHG,L4IAHG,L4OAHG,SLOW_RATE_08,SLOW_RATE_09,SLOW_RATE_10,L1A0BHG,L1B0BHG,L1C0BHG,L4IBHG,L4OBHG,IALRT_RATE_1,IALRT_RATE_2,IALRT_RATE_3,IALRT_RATE_4,IALRT_RATE_5,IALRT_RATE_6,IALRT_RATE_7,IALRT_RATE_8,IALRT_RATE_9,IALRT_RATE_10,IALRT_RATE_11,IALRT_RATE_12,IALRT_RATE_13,IALRT_RATE_14,IALRT_RATE_15,IALRT_RATE_16,IALRT_RATE_17,IALRT_RATE_18,IALRT_RATE_19,IALRT_RATE_20,TRIG_IA_EVNT,TRIG_IB_EVNT,NASIDE_IALRT,NBSIDE_IALRT,ERATE_1,ERATE_2,ERATE_3,ERATE_4,ERATE_5,L12A,L123A,PENA,L12B,L123B,PENB,SLOW_RATE_51,SLOW_RATE_52,SLOW_RATE_53,SLOW_RATE_54,H_06_08,H_12_15,H_15_70,HE4_06_08,HE4_15_70,EVENT_DATA_00,EVENT_DATA_01,EVENT_DATA_02,EVENT_DATA_03,EVENT_DATA_04,EVENT_DATA_05,EVENT_DATA_06,EVENT_DATA_07,EVENT_DATA_08,EVENT_DATA_09,EVENT_DATA_10,CHECKSUM
```
sdhoyt: For the other HIT data I ended up copying the header into the validation data csv. That should make it easier to read in the data

laspsandoval:
> > I think that the next step for this is to populate HIT test data into the I-ALiRT SC packet. That way I can use it for testing architecture. Please comment on this idea.
> 
> This seems like it is a good idea, but could also be complicated because you're having to inject specific bytes into another specific bytes object. If it is easy, it is likely worth doing, but I wouldn't spend too much time on it either.
> 
> More important in my opinion is writing a generic function that can grab the proper HIT values you need.
> 
> ```python
> def process_hit(xarray_dataset):
>     # do something with ds["HIT_EVENT_DATA_00"], ...
>     # This should be agnostic of whether you are passing in only the HIT CCSDS decoded dataset or the I-ALiRT decoded dataset.
> ```

imap_processing/tests/ialirt/unit/test_decom_ialirt_hit.py
```diff
@@ -0,0 +1,144 @@
...
+        flat_test_fast_rate_1[~np.isnan(flat_test_fast_rate_1)],
+    )
+
+    # Test Fast Rate 2
+    ccsds_fast_rate_2 = fast_rate_2[start_index : start_index + 60]
+    test_fast_rate_2 = hit_test_data.loc[
+        start_index : start_index + 59,
+        hit_test_data.columns.str.startswith(tuple(prefix_fast_rate_2)),
+    ]
+    flat_test_fast_rate_2 = test_fast_rate_2.to_numpy().flatten()
```
vmartinez-cu: Minor suggestion - you can combine the slicing and conversion to flattened numpy array in one step. This would simplify the tests to just compare the ccsds_fast_rate_# and test_fast_rate_# values directly without needing the flat_test_flat_rate_# variable. It's a little more readable, but just personal preference.

`test_fast_rate_2 = hit_test_data.loc[
    start_index : start_index + 59,
    hit_test_data.columns.str.startswith(tuple(prefix_fast_rate_2))
].to_numpy().flatten()`

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -1,19 +1,15 @@
-"""Decommutates i-alert packets and creates L1 data products."""
+"""Decommutates i-alirt packets and creates L1 data products."""
```
greglucas:
```suggestion
"""Decommutates I-ALiRT packets and creates L1 data products."""
```

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -0,0 +1,210 @@
...
+        "L4OAHG",
+        "SLOW_RATE_08",
+        "SLOW_RATE_09",
+        "SLOW_RATE_10",
+        "L1A0BHG",
+        "L1B0BHG",
+        "L1C0BHG",
+        "L4IBHG",
+        "L4OBHG",
+        *[f"IALRT_RATE_{i}" for i in range(1, 21)],
```
sdhoyt, line 40: Are you going to need to decom the HIT science packet for IALRT RATE and ERATEs?

imap_processing/ialirt/l0/decom_ialirt.py
```diff
@@ -1,19 +1,15 @@
...
-from imap_processing.decom import decom_packets
+from imap_processing.utils import packet_file_to_datasets
 
 logger = logging.getLogger(__name__)
 
 
-def generate_xarray(
-    packet_file: str, xtce: str, time_keys: Optional[dict] = None
-) -> xr.Dataset:
+def generate_xarray(packet_file: str, xtce: str) -> dict[int, xr.Dataset]:
```
greglucas: Can we just delete this file entirely? It looks like you are just returning the `packet_file_to_datasets()` routine.

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -0,0 +1,210 @@
...
+from enum import Enum
+from typing import ClassVar
+
+import numpy as np
+import xarray as xr
+
+logger = logging.getLogger(__name__)
+
+
+class HITPrefixes(Enum):
```
greglucas: This seems more like a mapping/lookup instead of an Enum to me, but I'm not really sure. I generally think of Enums as one-to-one mappings. Could you add an extra sentence explaining what this is used for too. "Create Enum for each rate type" doesn't explain what the rate types are or what it is doing.

```python
HIT_PREFIX_TO_RATE_TYPE = {"FAST_RATE_1": ..., "FAST_RATE_2": ..., "SLOW_RATE": ...}
```

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -0,0 +1,210 @@
...
+        "H_12_15",
+        "H_15_70",
+        "HE4_06_08",
+        "HE4_15_70",
+    ]
+
+
+def find_groups(data: xr.Dataset) -> xr.Dataset:
+    """
+    Find groups in the data.
```
greglucas: What does "groups" mean here?

Find all occurrences of the sequential set of 60 values 0 through 59. If a value is missing, or we are starting/ending in the middle of a sequence we do not count that as a valid group.

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -0,0 +1,210 @@
...
+    grouped_data = data.where(
+        (data["hit_sc_tick"] >= start_sc_tick) & (data["hit_sc_tick"] <= last_sc_tick),
+        drop=True,
+    )
+
+    # Create group labels
+    group_labels = np.searchsorted(
+        start_sc_ticks, grouped_data["hit_sc_tick"], side="right"
+    )
+    grouped_data["group"] = ("group", group_labels)
```
greglucas: Could you add an example of what `grouped_data` looks like? I'm not following all of the logic here and think you might just be getting one "group"? Or are you labeling every item in the array with a "group value", like this set of `[0-59]` all has an attached "1" label, the next set of `[0-59]` has a "2" label, etc.

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -0,0 +1,210 @@
...
+
+    Returns
+    -------
+    hit_data : dict
+        Dictionary final data product.
+    """
+    hit_data = []
+    grouped_data = find_groups(xarray_data)
+
+    for group in np.unique(grouped_data["group"]):
```
greglucas: I think this might be xarray's groupby function: https://docs.xarray.dev/en/stable/user-guide/groupby.html

```suggestion
    for group in grouped_data.groupby("group"):
```

Then you wouldn't have to do the `grouped_data["group"] == group` below, you'd already have the subset entries.

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -0,0 +1,210 @@
...
+        fast_rate_2 = grouped_data["hit_fast_rate_2"][
+            (grouped_data["group"] == group).values
+        ]
+        slow_rate = grouped_data["hit_slow_rate"][
+            (grouped_data["group"] == group).values
+        ]
+        met = int(grouped_data["hit_met"][(grouped_data["group"] == group).values][0])
+
+        # Verify that each group has 60 datapoints
+        if len(slow_rate.data) != 60:
```
greglucas: We shouldn't need to do `.data` here I don't think. We should avoid accessing the `.data` and `.values` attributes if possible and stick to the dataarray/dataset when we can.

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -0,0 +1,210 @@
...
+            logger.info(
```
greglucas: Didn't your grouping already take care of this so we shouldn't ever encounter this situation? Should we raise here instead?

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -0,0 +1,210 @@
...
+                f"for met {met}. Skipping."
+            )
+            continue
+
+        l1 = create_l1(fast_rate_1, fast_rate_2, slow_rate)
+
+        hit_data.append(
+            {
+                "met": met,
+                "hit_lo_energy_e_A_side": float(
```
greglucas: Why are we casting to Python ints/floats? The dataarrays should be able to be added together I think.

imap_processing/tests/ialirt/unit/test_process_hit.py
```diff
@@ -0,0 +1,159 @@
...
+def test_find_groups(xarray_data):
+    """Tests find_groups"""
+
+    filtered_data = find_groups(xarray_data)
+
+    assert filtered_data["hit_subcom"].values[0] == 0
+    assert filtered_data["hit_subcom"].values[-1] == 59
+    assert len(filtered_data["hit_subcom"]) / 60 == 15
+    assert len(filtered_data["hit_sc_tick"][filtered_data["hit_subcom"] == 0]) == 15
+    assert len(filtered_data["hit_sc_tick"][filtered_data["hit_subcom"] == 59]) == 15
```
greglucas: Would this be better as a test of the exact arrays expected?

```python
np.testing.assert_array_equal(filtered_data["hit_sc_tick"], np.tile(np.arange(60), 15))
```

imap_processing/tests/ialirt/test_data/l0/hit_ialirt_sample.csv
greglucas, line 1: It doesn't look like you are using this file in any of the assertions? This should give you a pretty good dataset to assert against I think rather than individual values.

```python
ds = pd.read_csv().to_xarray()
```

imap_processing/ialirt/l0/process_hit.py
greglucas: Does the science packet have those values in it? The science packet won't ever make it over into the I-ALiRT stream, so this should be separate. But perhaps if you're planning on doing this on the science side as well (if they have duplicated that data into the science packet in addition to the I-ALiRT packet) you could use the I-ALiRT implementation here for that work.

imap_processing/ialirt/l0/process_hit.py
laspsandoval: The latter: Every item in the array with a "group value", like this set of [0-59] all has an attached "1" label, the next set of [0-59] has a "2" label, etc. 

1. I used hit_subcom == 0 to define the beginning of the group. I then found the hit_sc_tick at this index and used it as the beginning time for the group.
2. I then assigned labels based on the hit_sc_tick start times.

Example:
grouped_data.coords
Coordinates:
    * epoch    (epoch) int64 7kB 315922822184000000 ... 315923721184000000
    * group    (group) int64 7kB 1 1 1 1 1 1 1 1 1 ... 15 15 15 15 15 15 15 15 15

imap_processing/ialirt/l0/process_hit.py
laspsandoval: I added a better check.

```# Subcom values for the group should be 0-59 with no duplicates.
subcom_values = grouped_data["hit_subcom"][(grouped_data["group"] == group).values]

# Ensure no duplicates and all values from 0 to 59 are present
if not np.array_equal(subcom_values, np.arange(60)):
    raise ValueError(f"Group {group} does not contain all values from 0 to 59 without duplicates.")```

laspsandoval: I hope that we will never encounter this situation where there are multiple subcom=2 in a group, for example, or missing subcom values, but it is good to have the extra check.

imap_processing/ialirt/l0/process_hit.py
sdhoyt: The science packet has fields called IALIRTRATES_00 to _19, so I'm not sure if that's needed for I-ALiRT. We're not using them for anything on our end, so I thought they'd be used by I-ALiRT. I imagine we won't be unpacking at a high enough frequency for I-ALiRT to use what we have, but we can share some code to help unpack those values from the science packet if those are needed. But it sounds like those shouldn't be needed on this end.

vmartinez-cu (approved):
Looks good to me!

sdhoyt (approved):

PR #758, opened 2024-08-20
MNT: Enable numpy 2+ installs

greglucas:
# Change Summary

## Overview

Numpy 2 is more strict about the overflow math calculations, so we needed to update a few cases to handle that as well.

## New Dependencies

Bumping numpy to allow 2+

### DIFF
--- a/imap_processing/mag/l1a/mag_l1a_data.py
+++ b/imap_processing/mag/l1a/mag_l1a_data.py
@@ -336,6 +336,10 @@ class MagL1a:
         primary_vectors = []
         secondary_vectors = []
 
+        # To avoid overflows, we need to cast the potentially 8 bit signed integers to
+        # int32 before the bitshifting operations below.
+        vector_data = vector_data.astype(np.int32)
+
         # Since the vectors are stored as 50 bit chunks but accessed via hex (4 bit
         # chunks) there is some shifting required for processing the bytes.
         # However, from a bit processing perspective, the first 48 bits of each 50 bit
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ space_packet_parser = ">=4.2.0"
 spiceypy = ">=6.0.0"
 xarray = '>=2023.0.0'
 pyyaml = "^6.0.1"
-numpy = "^1.26.4"
+numpy = "<=3"
 
 # Optional dependencies
 numpydoc = {version="^1.5.0", optional=true}
--- a/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
@@ -287,7 +287,6 @@ hi_de_nominal_bin:
   <<: *default_uint8
   CATDESC: Corresponding histogram angle bin for this Direct Event
   FIELDNAM: Histogram Bin Number
-  FILLVAL: 511
   FORMAT: I2
   LABLAXIS: Hist Bin \#
   VALIDMIN: 0
--- a/poetry.lock
+++ b/poetry.lock
@@ -1747,4 +1747,4 @@ tools = ["openpyxl", "pandas"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.9,<4"
-content-hash = "28cbb877e15c362c80541a8cebd35f12575cd86802c902505b4db1f65640d37e"
+content-hash = "f750033025b765826c827adb1142fd59444e13a985755c2a6efc3a207348a959"


imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -336,6 +336,10 @@ def to_signed16(n: int) -> int:
         primary_vectors = []
         secondary_vectors = []
 
+        # To avoid overflows, we need to cast the potentially 8 bit signed integers to
+        # int32 before the bitshifting operations below.
+        vector_data = vector_data.astype(np.int32)
```
greglucas, line 341: @maxinelasp, I'm not sure how you want to handle this, so let me know if you want something else. You are calling `astype(int32)` in the code calling section, but in the tests you pass in a `frombuffer()` call which is only int8. We could get rid of the type-cast in the calling routine and rely on this, or we could immediately cast the frombuffer in the test too if you'd prefer.

maxinelasp: Annoyingly, this is also complicated by the compression code that I'm working on, which also takes in vector data. Is the problem that the 8 bit ints are signed?

maxinelasp: If this change passes the tests then I'm ok with it. The bitshifting is unfortunately complicated with different types, but I am also looking to rework the types to be more consistent in my newest change, so that it works with the uncompressed and compressed algorithms.

tmplummer (approved):
Looks good to me.

imap_processing/mag/l1a/mag_l1a_data.py
greglucas: > Is the problem that the 8 bit ints are signed?

The problem is here:
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/4a2642b09956be882f278b38175660a4b2e670da/imap_processing/mag/l1a/mag_l1a_data.py#L349-L352

where you are doing math with `0xFFFF` which is out of range for those types, so numpy2+ is warning and saying we need to be explicit with what types we want to do the math on.

Within the code you are already casting here:
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/4a2642b09956be882f278b38175660a4b2e670da/imap_processing/mag/l1a/mag_l1a.py#L185

But in the tests, you aren't:
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/4a2642b09956be882f278b38175660a4b2e670da/imap_processing/tests/mag/test_mag_l1a.py#L63-L65

So I was saying that we could also update this in those two locations if you wanted as well.

maxinelasp: Yeah, ok, in that case either change is fine, and I'll go back once the compression algorithm is done and make sure it's consistent with the types I'm using there.

pyproject.toml
```diff
@@ -38,7 +38,7 @@ space_packet_parser = ">=4.2.0"
 spiceypy = ">=6.0.0"
 xarray = '>=2023.0.0'
 pyyaml = "^6.0.1"
-numpy = "^1.26.4"
+numpy = "<=3"
```
maxinelasp, line 41: We probably want to have this pinned to a minor version, no? `numpy = "^2.1.0` I'm not sure we want random bumps up to, for example, `2.2.0`.

greglucas: I don't think we should require numpy 2+ because that will likely cause issues with other packages only supporting <=2 because those packages maybe haven't done the fixes themselves yet to be compatible.

I'm personally a fan of more open-ended ranges for packages so that we don't get pinned back and then have to do a massive update later on. When deploying/installing the package as a user, then we can/should use explicit package locks, but in general I have a preference for leaving the testing of the package open so that we don't get stuck way behind the times and miss things.

maxinelasp: Ok, since this uses a lockfile I'm alright with that. Did you see if requiring numpy 2+ actually caused issues with packages?

maxinelasp (approved):

pyproject.toml
greglucas: Yes, it does not cause issues right now, we are able to install numpy 2+. I think this is a difference of whether we are the final "user" (what I think you're getting at with the lockfile for install) or if we expect this package to be used by others (my assumption) and potentially installing other packages besides the ones we require. Pinning to numpy 2+ will affect others and not just ourselves. If we work with numpy 1.26 right now we should let that keep working IMO.

greglucas:
Merging now, we can follow up on installation pins in subsequent PRs.

Issue #739, opened 2024-08-07
Pointing Frame Kernel Generation Algorithm - per pointing kernel


laspsandoval:
Update the pointing frame kernel generation code to use the repoint table to identify each pointing and generate a per-pointing kernel file. Compare each averaged quaternion to existing pointing kernel for that pointing (if it exists) and determine if a new pointing kernel should be written.

PR #759, opened 2024-08-20
Pointing frame code

### LINKED ISSUE #760: Pointing Frame Kernel Generation Algorithm (1) - Lambda and Test Only
### Requirements:

> **Specify: (optional)**

### Code:

```python
# Code
```

### LINKED ISSUE #764: ensure_spice functionality for pointing kernel generation
### Algorithm Description:
Furnishing of kernels for the pointing (cough, despun) kernel generation code requires unique handling. This ticket is to implement that logic or make sure existing logic correctly handles this case.

### Requirements:
- I think that we only want to furnish the single CK that the pointing kernel is being generated for.
- We should identify the minimum set of other kernels that need to be furnished. For example, do we even need sclk or leapsecond kernels?

### Ideas for implementation:
- One possibility is to add another keyword argument to the `ensure_spice` decorator to handle the pointing kernel case.
- Another possibility is to just make sure that the needed kernels are furnished before calling the function that makes the kernel.

### LINKED ISSUE #765: Add ensure spice to pointing frame and move to kernels.py
### Requirements:
* The functions in pointing_frame_handler.py should be in the imap_processing/spice/kernels.py module.
* Add ensure spice decorator instead of mount path to efs and in other places currently using:

 with spice.KernelPool(kernels):

The ensure_spice decorator is in place and I think should be used here. I pictured the input of this function to be a file path to the CK that should be used and that it would not contain any kernel furnishing code.

For your test, you can manually furnish only the kernels needed using fixture(s) similar to https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/conftest.py#L44

laspsandoval:
# Change Summary

## Overview
Provides algorithm that mocks Nick Duttons pointing frame (DPS frame) kernel generation. Note that I have TODOs that will be included in the next tickets.

## New Files
- pointing_frame_handler.py
   - writes a new C-kernel file containing the averaged pointing data

## Testing
- test_pointing_frame_handler.py
   - tests pointing_frame_handler.py
- imap_science_0001.tf 
   - pointing frame kernel used for test data
-  imap_sclk_0000.tsc
   - spacecraft clock kernel used for test data
- imap_wkcp.tf 
   - spacecraft frame kernel used for test data


imap_processing/pointing_frame_handler.py
tmplummer, line 1: IMO, the functions in this file should be in the `imap_processing/spice/kernels.py` module.

imap_processing/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+    -------
+    q_avg : np.ndarray
+        Average quaternion.
+    z_eclip_time : list
+        Z-axis of the ECLIPJ2000 frame. Used for plotting.
+    """
+    z_eclip_time = []
+    aggregate = np.zeros((4, 4))
+
+    for tdb in et_times:
```
tmplummer: I would like to keep Nick's comment in here about this method. Personally, I find it odd that we are not getting the quaternions directly from the CK, but instead we are resampling.
```suggestion
    for tdb in et_times:
        # we use a quick and dirty method here for grabbing the quaternions
        # from the attitude kernel.  Depending on how well the kernel input
        # data is built and sampled, there may or may not be aliasing with this
        # approach.  If it turns out that we need to pull the quaternions
        # direclty from the CK there are several routines that exist to do this
        # but it's not straight forward.  We'll revisit this if needed.
        #
        # CKLPF - open the ck
        # DAFBBS - begin backwards order search for segments
        # DAFGS - gets the segment summary
        # CKNR03 - gets the number of records in a type 3 CK segment
        # CKGR03 - retrieves an individual record from a type 3 CK segment
        # CKUPF - close the file
        #
```

imap_processing/tests/pointing_frame/test_data/imap_science_0001.tf
tmplummer, line 1: Along with the above comment about location of the code, this should be located in `imap_processing/tests/spice/test_data`.

imap_processing/tests/pointing_frame/test_data/imap_sclk_0000.tsc
tmplummer, line 1: There is already a sclk kernel checked in to `imap_processing/tests/spice/test_data`

imap_processing/tests/pointing_frame/test_data/imap_wkcp.tf
tmplummer, line 1: Same test data location comment as above. This should be in `imap_processing/tests/spice/test_data`.

imap_processing/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
+"""
+Generate Pointing Frame.
+
+Notes
+-----
+Kernels that are required to run this code:
+1. imap_science_0001.tf - pointing frame kernel
+2. imap_sclk_0000.tsc - spacecraft clock kernel
+3. imap_wkcp.tf - spacecraft frame kernel
+4. de430.bsp - standard SPICE planetary ephemeris kernel
```
tmplummer: I don't think that de430.bsp is really needed for this code, is it?

imap_processing/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+
+    for tdb in et_times:
+        # Rotation matrix from IMAP spacecraft frame to ECLIPJ2000.
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.pxform
+        body_rots = spice.pxform("IMAP_SPACECRAFT", "ECLIPJ2000", tdb)
+        # Convert rotation matrix to quaternion.
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.m2q
+        body_quat = spice.m2q(body_rots)
+        # z-axis of the ECLIPJ2000 frame.
+        z_eclip_time.append(body_rots[:, 2])
```
tmplummer: While it is unfortunate that spiceypy doesn't offer much vectorization of CSPICE functions, one thing that could be done to speed up this for loop is to preallocate `z_eclip_time` to the known size rather than using `append`.

imap_processing/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Returns
+    -------
+    path_to_pointing_frame : Path
+        Path to dps frame.
+    """
+    # Mount path to EFS.
+    mount_path = Path(os.getenv("EFS_MOUNT_PATH", ""))
+
+    # TODO: this part will change with ensure_spice decorator.
```
tmplummer: The `ensure_spice` decorator is in place and I **think** should be used here. I pictured the input of this function to be a file path to the CK that should be used and that it would not contain any kernel furnishing code.

For your test, you can manually furnish only the kernels needed using fixture(s) similar to https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/conftest.py#L44

imap_processing/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+# TODO : Add multiple pointings to the pointing frame.
+
+
+def get_coverage(ck_kernel: str) -> tuple[float, float, np.ndarray]:
+    """
+    Create the pointing frame.
```
tmplummer: Please update this description.

imap_processing/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+    ck_kernel : str
+        Path of ck_kernel used to create the pointing frame.
+
+    Returns
+    -------
+    et_start : float
+        Start time of ck_kernel.
+    et_end : float
+        End time of ck_kernel.
+    et_times : numpy.ndarray
```
tmplummer: I think that this output should be left out. It has nothing to do with the actual kernel coverage.

imap_processing/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.gipool
+    id_imap_spacecraft = spice.gipool("FRAME_IMAP_SPACECRAFT", 0, 1)
+
+    # TODO: Queried pointing start and stop times here.
+
+    # Get the coverage window
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.ckcov
+    cover = spice.ckcov(ck_kernel, int(id_imap_spacecraft), True, "SEGMENT", 0, "TDB")
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.wnfetd
+    et_start, et_end = spice.wnfetd(cover, 0)
```
tmplummer: I'm not totally clear on the spice terminology here, but why do we only care about the first interval of the window? Maybe a question for Nick.

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,136 @@
...
+    """Create kernel lists."""
+    kernels = [str(file) for file in kernel_path.iterdir()]
+    ck_kernel = [
+        str(file) for file in kernel_path.iterdir() if file.name == "imap_spin.bc"
+    ]
+
+    return kernels, ck_kernel
+
+
+@pytest.mark.xfail(reason="Will fail unless kernels in pointing_frame/test_data.")
```
tmplummer: I'm not sure I understand why this is marked `xfail` if the CK is checked in, why can't this test always pass?

imap_processing/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+import numpy as np
+import spiceypy as spice
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+# TODO : Add multiple pointings to the pointing frame.
+
+
+def get_coverage(ck_kernel: str) -> tuple[float, float, np.ndarray]:
```
tmplummer: This function should either allow spk kernels as input too or be renamed to be ck specific.

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,136 @@
...
+def test_create_pointing_frame(monkeypatch, kernel_path, create_kernel_list):
+    """Tests create_pointing_frame function."""
+    monkeypatch.setenv("EFS_MOUNT_PATH", str(kernel_path))
+    _, ck_kernel = create_kernel_list
+    create_pointing_frame()
+
+    # After imap_dps.bc has been created.
+    kernels = [str(file) for file in kernel_path.iterdir()]
+
+    with spice.KernelPool(kernels):
```
tmplummer: I would expect this to do a targeted furnish of the output from `create_pointing_frame()` and any other needed kernels (are there any?) rather than everything in the directory.

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,136 @@
...
+    kernels, ck_kernel = create_kernel_list
+
+    with spice.KernelPool(kernels):
+        et_start, et_end, et_times = get_coverage(str(ck_kernel[0]))
+
+    assert et_start == 802008069.184905
+    assert et_end == 802094467.184905
+
+
+@pytest.mark.xfail(reason="Will fail unless kernels in pointing_frame/test_data.")
```
tmplummer: Again, I don't understand why this is expected to fail.

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,136 @@
...
+
+    with spice.KernelPool(kernels):
+        et_start, et_end, et_times = get_coverage(str(ck_kernel[0]))
+
+        # Converts rectangular coordinates to spherical coordinates.
+        q_avg, z_eclip_time = average_quaternions(et_times)
+        z_avg_expected = spice.q2m(list(q_avg))[:, 2]
+        _, z_avg = create_rotation_matrix(et_times)
+
+        assert np.array_equal(z_avg, z_avg_expected)
```
tmplummer: Can this be moved into another test and this one removed. I generally think that plots are an algorithm validation step that you could save and add to a ticket rather than doing plotting in automated tests.

tmplummer (commented):
This is a great start. I have some comments on organization and use of kernels.

imap_processing/pointing_frame_handler.py
laspsandoval: I can move it to the spice directory. It would seem odd to add it to kernels.py. The purpose of that program is:

"""Functions for furnishing and tracking SPICE kernels."""

which does seem to fit what I am doing w/ averaging quaternions.

imap_processing/pointing_frame_handler.py
laspsandoval: Sounds good.

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
laspsandoval: imap_spin.bc is a very large file and so I didn't commit it to the repo.

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
laspsandoval: I listed the kernels to furnish here, but I suspect this will change with your ensure spice code anyways.

    kernels = [
        "naif0012.tls",
        "imap_science_0001.tf",
        "imap_sclk_0000.tsc",
        "imap_wkcp.tf",
        "imap_spin.bc",
    ]

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
laspsandoval: If imap_spin.bc is not in the test directory it fails:

imap_spin.bc - test attitude kernel available at:
   https://lasp.colorado.edu/galaxy/display/IMAP/Data

imap_processing/pointing_frame_handler.py
tmplummer: Maybe the docstring should be updated to be `"""Functions for furnishing, tracking, and generating SPICE kernels."""`. In my mind, this code would definitely be categorized into kernel related SPICE code.

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
laspsandoval: I find it helpful to have it in one place but I'll do a @pytest.mark.skip

imap_processing/pointing_frame_handler.py
laspsandoval: Eventually this will be the place where we query the pointing csv for start/end times. I'll rename though.

imap_processing/spice/pointing_frame_handler.py
greglucas, line 1: I agree with @subagonsouth that this would go nicely in the `kernels.py` file. A function called something like `create_pointing_frame_kernel()` would be the logical public function, with all these helpers being private (or even inlined into the one function honestly since it is all pretty serial logic and not shared anywhere else).

imap_processing/spice/pointing_frame_handler.py
```diff
@@ -0,0 +1,245 @@
...
+    # TODO: Queried pointing start and stop times here.
+
+    # Get the coverage window
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.ckcov
+    cover = spice.ckcov(ck_kernel, int(id_imap_spacecraft), True, "SEGMENT", 0, "TDB")
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.wnfetd
+    et_start, et_end = spice.wnfetd(cover, 0)
+    # Each spin is 15 seconds. We want 10 quaternions per spin.
+    # duration / # samples (nominally 15/10 = 1.5 seconds)
+    et_times = np.arange(et_start, et_end, 1.5)
```
greglucas: linspace per the other review.

imap_processing/spice/pointing_frame_handler.py
```diff
@@ -0,0 +1,245 @@
...
+    # that has the most influence.
+    # The eigenvector corresponding to the largest
+    # eigenvalue points in the direction that has the most
+    # combined rotation influence.
+    eigvals, eigvecs = np.linalg.eig(aggregate)
+    # q0: The scalar part of the quaternion.
+    # q1, q2, q3: The vector part of the quaternion.
+    q_avg = eigvecs[:, np.argmax(eigvals)]
+
+    return q_avg, z_eclip_time
```
greglucas: My personal opinion is that I don't think it is useful to return tuples here just for nice-to-have things and debugging. We should have functions that return a single item if that is all that is being used. We can step in with a debugger if we want to get the other data out. (this applies to most of the functions in this file, just return the name of the function, not ancillarly data)

imap_processing/tests/pointing_frame/test_pointing_frame_handler.py
tmplummer: I'm not sure why it is so big. Can you try to use `IMAP_spacecraft_attitude.bc` which Nick delivered with the simulated data? That file is only 55 KB.

imap_processing/spice/pointing_frame_handler.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Create the pointing frame.
+
+    Returns
+    -------
+    path_to_pointing_frame : Path
+        Path to dps frame.
+    """
+    # Mount path to EFS.
+    mount_path = Path(os.getenv("EFS_MOUNT_PATH", ""))
```
greglucas: This should be obtained from `imap_data_access`, or I think it is actually called `IMAP_SPICE_DIR` I think? EFS is an implementation detail, if I have it locally I don't have an EFS MOUNT PATH, I have a SPICE DIR. This is where the `ensure_spice` decorator should point to this location and not require you to manage it here I think.

imap_processing/tests/spice/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,177 @@
+"""Tests Pointing Frame Generation."""
+
+import shutil
+import sys
+from pathlib import Path
+
+import matplotlib.pyplot as plt
```
greglucas:
```suggestion
```
We don't want matplotlib as any dependency in my opinion. It can be an optional one if we want, but we shouldn't bring it in as a top-level import.

imap_processing/tests/spice/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,177 @@
...
+    """Tests get_et_times function."""
+    kernels, ck_kernel = create_kernel_list
+
+    with spice.KernelPool(kernels):
+        et_start, et_end, et_times = get_et_times(str(ck_kernel[0]))
+
+    return et_times
+
+
+@pytest.mark.xfail(reason="Will fail unless kernels in pointing_frame/test_data.")
```
greglucas: This should be a `skipif`, because you want it to actually pass/fail if you have the data there, whereas it will be XPASS in this case but still not count for/against your test coverage.

You can make a decorator at the top of the file that is something like `requires_spice_data = pytest.mark.skipif(not Path(file).exists())` and then mark your tests here with `@requires_spice_data`

imap_processing/tests/spice/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+    np.testing.assert_allclose(rotation_matrix_1, rotation_matrix_expected, atol=1e-4)
+
+    # Verify imap_dps.bc has been created.
+    assert (kernel_path / "imap_dps.bc").exists()
+
+
+@pytest.mark.skip(reason="Plotting only")
+def test_declination_plot(create_kernel_list, et_times):
+    """Tests Inertial z axis and provides visualization."""
```
greglucas: I don't think think this is "testing", it is a helper plot function, but there are no image comparisons or assertions going on.

imap_processing/tests/spice/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,177 @@
...
+    kernels, ck_kernel = create_kernel_list
+
+    with spice.KernelPool(kernels):
```
greglucas: Can you have this be the fixture itself that yields these needed kernels?

pyproject.toml
```diff
@@ -52,11 +52,12 @@ ruff = {version="==0.2.1", optional=true}
 sphinx = {version="*", optional=true}
 sphinxcontrib-openapi = {version="^0.8.3", optional=true}
 mypy = {version="^1.10.1", optional=true}
+matplotlib = {version="^3.7.1", optional=true}
```
greglucas:
```suggestion
```

pyproject.toml
```diff
@@ -52,11 +52,12 @@ ruff = {version="==0.2.1", optional=true}
...
 
 [tool.poetry.extras]
 dev = ["pre-commit", "ruff", "mypy"]
 doc = ["numpydoc", "pydata-sphinx-theme", "sphinx", "sphinxcontrib-openapi"]
-test = ["openpyxl", "pytest", "pytest-cov"]
+test = ["openpyxl", "pytest", "pytest-cov", "matplotlib"]
```
greglucas:
```suggestion
test = ["openpyxl", "pytest", "pytest-cov"]
```

imap_processing/spice/pointing_frame_handler.py
laspsandoval: Ok, but this way:
    # 1 spin/15 seconds * 86400 seconds/pointing * 10 quaternions / spin
    et_times = np.linspace(et_start, et_end, 57600)

imap_processing/tests/spice/test_pointing_frame_handler.py
laspsandoval: What about this?
[tool.poetry.extras]
dev = ["pre-commit", "ruff", "mypy"]
doc = ["numpydoc", "pydata-sphinx-theme", "sphinx", "sphinxcontrib-openapi"]
test = ["openpyxl", "pytest", "pytest-cov"]
tools= ["openpyxl", "pandas"]
plotting = ["matplotlib"]

pyproject.toml
laspsandoval: It is just so helpful to be able to plot and to have the plot in a place where I know all of the tests are.

imap_processing/tests/spice/test_pointing_frame_handler.py
greglucas: Sure, but then you have to move this import inside your helper function so our tests don't fail when not finding matplotlib. IMO this shouldn't be in the "tests" area though, it should be in the main code or tools area. Tests should be assertions. I wouldn't look in test directories to find helper routines.

imap_processing/spice/pointing_frame_handler.py
laspsandoval: I will put it as a TODO to use ensure_spice.

greglucas: Why not use it now? It is a helper function that should make our lives easier and to me this is currently not correct because it is using an undocumented environment variable. If the decorator isn't helping we should add to it so it is more helpful. @subagonsouth do you have any pointers for how to explicitly use it here?

imap_processing/spice/pointing_frame_handler.py
laspsandoval: Let's get through this PR and then we can migrate if we want. I started this before we had ensure spice so I think that should be another PR.

imap_processing/spice/pointing_frame_handler.py
tmplummer: Sure...
```
@ensure_spice()
def create_pointing_frame() -> Path:
```
Then just remove all of the kernel and furnishing code from this function. For now, you will need to furnish the needed kernels in your test functions before calling. I am going to write up a ticket for making sure that we correctly handle furnishing kernels for this specific case in the decorator.

imap_processing/spice/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+# TODO : Add multiple pointings to the pointing frame.
+
+
+def get_et_times(ck_kernel: str) -> tuple[float, float, np.ndarray]:
+    """
+    Create the pointing frame.
```
tmplummer: Please update this docstring.

imap_processing/spice/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+    et_end : float
+        End time of ck_kernel.
+    et_times : numpy.ndarray
+        Array of times between et_start and et_end.
+    """
+    # Get the spacecraft ID.
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.gipool
+    id_imap_spacecraft = spice.gipool("FRAME_IMAP_SPACECRAFT", 0, 1)
+
+    # TODO: Queried pointing start and stop times here.
```
tmplummer: Can you please write a ticket for this with the SPICE label?

imap_processing/spice/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # Get the coverage window
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.ckcov
+    cover = spice.ckcov(ck_kernel, int(id_imap_spacecraft), True, "SEGMENT", 0, "TDB")
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.wnfetd
+    et_start, et_end = spice.wnfetd(cover, 0)
+    # Assumes a pointing is ~ 1 day.
+    # 1 spin/15 seconds * 86400 seconds/pointing * 10 quaternions / spin
+    et_times = np.linspace(et_start, et_end, 57600)
```
tmplummer: Can you make this dynamic instead of assuming 24 hour pointing...
```suggestion
    # Sample the attitude ~10 times per 15-second spin
    num_samples = (et_end - et_start) / 15 * 10
    et_times = np.linspace(et_start, et_end, )
```

imap_processing/spice/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+    ----------
+    et_times : numpy.ndarray
+        Array of times between et_start and et_end.
+
+    Returns
+    -------
+    q_avg : np.ndarray
+        Average quaternion.
+    """
+    z_eclip_time = []
```
tmplummer: It looks like this is no longer used. Can it be removed?

imap_processing/spice/pointing_frame_handler.py
```diff
@@ -0,0 +1,242 @@
...
+        sclk_begtim = spice.sce2c(int(id_imap_sclk), et_start)
+        sclk_endtim = spice.sce2c(int(id_imap_sclk), et_end)
+
+        # Get the pointing frame ID.
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.gipool
+        id_imap_dps = spice.gipool("FRAME_IMAP_DPS", 0, 1)
+
+        # Create the pointing frame kernel.
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.ckw02
+        spice.ckw02(
```
tmplummer: Cool! Nice job using the lower level function rather than the executable. I missed this on my first pass.

imap_processing/tests/spice/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,182 @@
...
+        "naif0012.tls",
+        "imap_science_0001.tf",
+        "imap_sclk_0000.tsc",
+        "imap_wkcp.tf",
+        "imap_spin.bc",
+    ]
+
+    for file in test_dir.iterdir():
+        if file.name in kernels:
+            shutil.copy(file, tmp_path / file.name)
```
tmplummer: Along with the suggested change to using `@ensure_spice()`, you could avoid copying the kernels to a temp directory and just furnish them directly from the `tests/spice/test_data` directory. There is already a fixture that defines that location. It is called `spice_test_data_path`.

imap_processing/tests/spice/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+@pytest.fixture()
+def ck_kernel(kernel_path):
+    """Create ck kernel."""
+    ck_kernel = [
+        str(file) for file in kernel_path.iterdir() if file.name == "imap_spin.bc"
+    ]
+
+    return ck_kernel
```
tmplummer: There is a fixture that defines the spice test data path. This function would be clearer if you did:
```suggestion
def ck_kernel(spice_test_data_path):
    """Location of ck kernel to create pointing kernel from."""
    return spice_test_data_path / "imap_spin.bc"
```

tmplummer (commented):
Ugg, sorry for requesting yet more changes. Thanks for sticking with this. I wonder if doing an early WIP PR would be helpful for critical code like this in the future?

imap_processing/spice/pointing_frame_handler.py
greglucas: I don't think we need to be in a rush to get this in though? I realize you started before, but now we have it so we should use it IMO. Things are constantly evolving and it is easier to update now, similar to having to deal with merge and rebase conflicts.

imap_processing/tests/spice/test_pointing_frame_handler.py
```diff
@@ -0,0 +1,136 @@
...
+
+def test_average_quaternions(et_times, kernels):
+    """Tests average_quaternions function."""
+
+    with spice.KernelPool(kernels):
+        q_avg = average_quaternions(et_times)
+
+    # Generated from MATLAB code results
+    q_avg_expected = np.array([-0.6838, 0.5480, -0.4469, -0.1802])
+    np.testing.assert_allclose(q_avg, q_avg_expected, atol=1e-1)
```
greglucas: FYI that this is a 10% error, which seems quite high. Do you want to compare the digits? You might be able to use `siginficant=5` with this function: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_approx_equal.html

imap_processing/pointing_frame_handler.py
laspsandoval: As I understand it there are sometimes segments within the kernel so the start and end time would be for each segment. That isn't the case here since we generated the test data. Also this will all go away when we use the et_start and et_end of the queried csv for pointing.

imap_processing/spice/pointing_frame_handler.py
laspsandoval: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/739

imap_processing/tests/spice/test_pointing_frame_handler.py
laspsandoval: It was closer before I changed the test data.

greglucas: Ahh I see. I think you want to update the expected values to the actual new data then (maybe leave the MATLAB results as a comment). Because you want it to raise an error if someone changes some other subtle thing like using a different kernel as you have. So we don't want it to pass if we use the actual CK that produced the data anymore.

imap_processing/tests/spice/test_kernels.py
```diff
@@ -1,33 +1,65 @@
...
+    _average_quaternions,
+    _create_rotation_matrix,
+    _get_et_times,
+    create_pointing_frame,
+    ensure_spice,
+)
+
+
+@pytest.fixture()
+def kernels(spice_test_data_path):
```
tmplummer: This fixture seems sensitive to extra kernels being added to the test data path. Each test should have a distinct set of kernels it requires rather than just furnishing everything in the test data path. Perhaps this fixture could be called `pointing_frame_kernels` and return a hard coded list of those kernels?

imap_processing/tests/spice/test_kernels.py
```diff
@@ -1,33 +1,65 @@
...
+    """List SPICE kernels."""
+    # TODO: remove this fixture after ensure_spice update.
```
tmplummer: I'm not sure this TODO is real. We may add a test that checks the `ensure_spice` functionality when wrapping the `create_pointing_frame` function, but I don't anticipate changes that affect how kernels are furnished for tests.

imap_processing/tests/spice/test_kernels.py
```diff
@@ -1,33 +1,65 @@
...
+    kernels = [str(file) for file in spice_test_data_path.iterdir()]
+    return kernels
+
+
+@pytest.fixture()
+def et_times(kernels):
+    """Tests get_et_times function."""
+    # TODO: remove spice.furnsh(kernels) after ensure_spice update.
```
tmplummer: Same comment as the TODO as above.

imap_processing/spice/kernels.py
```diff
@@ -156,3 +159,215 @@ def wrapper_ensure_spice(*args: Any, **kwargs: Any) -> Any:
...
+    # combined rotation influence.
+    eigvals, eigvecs = np.linalg.eig(aggregate)
+    # q0: The scalar part of the quaternion.
+    # q1, q2, q3: The vector part of the quaternion.
+    q_avg = eigvecs[:, np.argmax(eigvals)]
+
+    return q_avg
+
+
+@ensure_spice
```
tmplummer: I don't think that create rotation matrix needs to be wrapped with `@ensure_spice` the only spice call in this function is `q2m` which does not require kernels be furnished.

imap_processing/spice/kernels.py
```diff
@@ -156,3 +159,215 @@ def wrapper_ensure_spice(*args: Any, **kwargs: Any) -> Any:
...
+        np.array([1.0]),
+    )
+
+    # Close CK file.
+    spice.ckcls(handle)
+
+    return path_to_pointing_frame
+
+
+@ensure_spice
```
tmplummer, line 272: Can you add a TODO for removing the `@ensure_spice` decorator when this function is modified to use the repointing table?

imap_processing/tests/spice/test_kernels.py
```diff
@@ -1,33 +1,65 @@
...
+@pytest.fixture()
+def kernels(spice_test_data_path):
+    """List SPICE kernels."""
+    # TODO: remove this fixture after ensure_spice update.
+    kernels = [str(file) for file in spice_test_data_path.iterdir()]
+    return kernels
+
+
+@pytest.fixture()
+def et_times(kernels):
```
tmplummer: This is kind of a weird combo of a fixture and test. It isn't really clear that it is a test by the name and in fact I don't think `pytest` will pick this up as a test. Can this just be a fixture and write a separate test for `_et_times`?

imap_processing/tests/spice/test_kernels.py
```diff
@@ -61,9 +93,65 @@ def test_ensure_spice_time_kernels():
...
     # The ensure_spice decorator should raise a SpiceyError when all attempts to
     # furnish a set of kernels with sufficient coverage for the spiceypy
     # functions that it decorates.
     with pytest.raises(SpiceyError):
         _ = wrapped(577365941.184, "ISOC", 3) == "2018-04-18T23:24:31.998"
+
+
+def test_average_quaternions(et_times, kernels):
+    """Tests average_quaternions function."""
+    # TODO: remove spice.furnsh(kernels) after ensure_spice update.
```
tmplummer: Remove this TODO. This is a good way to test this code independent of the `ensure_spice` decorator.

imap_processing/tests/spice/test_kernels.py
```diff
@@ -61,9 +93,65 @@ def test_ensure_spice_time_kernels():
...
+    q_avg = _average_quaternions(et_times)
+
+    # Generated from MATLAB code results
+    q_avg_expected = np.array([-0.6611, 0.4981, -0.5019, -0.2509])
+    np.testing.assert_allclose(q_avg, q_avg_expected, atol=1e-4)
+
+
+def test_create_rotation_matrix(et_times, kernels):
+    """Tests create_rotation_matrix function."""
+    # TODO: remove spice.furnsh(kernels) after ensure_spice update.
```
tmplummer: Remove this TODO. This is a good way to test this code independent of the ensure_spice decorator.

imap_processing/tests/spice/test_kernels.py
```diff
@@ -61,9 +93,65 @@ def test_ensure_spice_time_kernels():
...
+    )
+    z_avg_expected = np.array([0.4136, 0.9104, 0.0000])
+
+    np.testing.assert_allclose(z_avg, z_avg_expected, atol=1e-4)
+    np.testing.assert_allclose(rotation_matrix, rotation_matrix_expected, atol=1e-4)
+
+
+def test_create_pointing_frame(spice_test_data_path, kernels, tmp_path):
+    """Tests create_pointing_frame function."""
+    # TODO: remove spice.furnsh(kernels) after ensure_spice update.
```
tmplummer: Remove this TODO. This is a good way to test this code independent of the ensure_spice decorator.

imap_processing/tests/spice/test_data/imap_spin.bc
tmplummer, line 1: Can this file be renamed to something less generic? Something like `imap_sim_ck_24hr_10secsampling_with_nutation.bc`? Maybe that is too verbose but you get the idea.

tmplummer (commented):
Nice, Laura. This is looking great! I just have a few requests for minor changes 😬

imap_processing/spice/kernels.py
```diff
@@ -156,3 +159,215 @@ def wrapper_ensure_spice(*args: Any, **kwargs: Any) -> Any:
...
+
+    # Convert the rotation matrix to a quaternion.
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.m2q
+    q_avg = spice.m2q(rotation_matrix)
+
+    # TODO: come up with naming convention.
+    if pointing_frame_dir is None:
+        path_to_pointing_frame = directory / "imap_dps.bc"
+    else:
+        path_to_pointing_frame = pointing_frame_dir / "imap_dps.bc"
```
greglucas: Do we need this to be optional / configurable, or can we say something like "write the frame kernel to the same location as the input CK with the name ..."

I'd suggest changing to the full path to be passed in if we do want to go that route rather than a directory since that is more explicit for end users (bringing your `directory` variable inline here directly too)

```suggestion
    # TODO: come up with naming convention.
    if pointing_frame_path is None:
        pointing_frame_path = Path(ck_kernel).parent / "imap_dps.bc"
```

imap_processing/spice/kernels.py
```diff
@@ -156,3 +159,215 @@ def wrapper_ensure_spice(*args: Any, **kwargs: Any) -> Any:
...
+        q_avg,
+        # 0.0 Angular rotation terms.
+        np.array([0.0, 0.0, 0.0]),
+        # Rates (seconds per tick) at which the quaternion and
+        # angular velocity change.
+        np.array([1.0]),
+    )
+
+    # Close CK file.
+    spice.ckcls(handle)
```
greglucas: Do we get into any weird situations/leaks if there is an error above this and we aren't able to close the file? I'm wondering if we need a context manager around this file opening/closing.

imap_processing/spice/kernels.py
```diff
@@ -156,3 +159,215 @@ def wrapper_ensure_spice(*args: Any, **kwargs: Any) -> Any:
...
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.m2q
+        body_quat = spice.m2q(body_rots)
+
+        # Standardize the quaternion so that they may be compared.
+        body_quat = body_quat * np.sign(body_quat[0])
+        # Aggregate quaternions into a single matrix.
+        aggregate += np.outer(body_quat, body_quat)
+
+    # Reference: "On Averaging Rotations"
+    # Link: https://link.springer.com/content/pdf/10.1023/A:1011129215388.pdf
```
greglucas: All of these comments are fantastic, thank you!

This is a bit buried here, do you want to add this to the references section of the main docstring?
https://numpydoc.readthedocs.io/en/latest/format.html#references

poetry.lock
greglucas, line 1: Can you revert this update? It should be created by pre-commit automatically which is on version 1.8.0, so you must have updated this with a local poetry version I think.

imap_processing/tests/spice/test_kernels.py
```diff
@@ -61,9 +93,65 @@ def test_ensure_spice_time_kernels():
...
+    spice.furnsh(kernels)
+    ck_kernel, _, _, _ = spice.kdata(0, "ck")
+    et_start, et_end, et_times = _get_et_times(ck_kernel)
+    create_pointing_frame(pointing_frame_dir=tmp_path)
+
+    # After imap_dps.bc has been created.
+    dps_kernel = str(tmp_path / "imap_dps.bc")
```
greglucas:
```suggestion
    dps_kernel = create_pointing_frame(pointing_frame_dir=tmp_path)
```

greglucas (approved):
Thank you for taking the time to address all of the feedback, I know there was a lot of it which can sometimes be hard to keep track of. I think this reads really nicely now, great work on this!

I just have a few minor nitpicks and I agree with all of Tim's comments so those should be addressed before merging too.

greglucas:
Just as a note that we should discuss in a future tag-up. Because @subagonsouth had comments I think his approval should have been waited for. I approved suggesting my comments were sufficient, but thought maybe Tim should take another look before merging.

PR #762, opened 2024-08-20
CoDICE L1 decompression and unpacking data updates

### LINKED ISSUE #767: CoDICE: Implement proper unpacking and restructuring of data arrays for CoDICE-lo
Prior to SIT-3, the CoDICE pipeline essentially created bogus data arrays in order to produce CDFs in time for the integration test. The pipeline should be improved to properly unpack the decommutated science data and structure the data arrays in the CDF data variables for CoDICE-lo.

bourque:
# Change Summary

## Overview
This PR makes improvements to how CoDICE data is unpacked and structured as data arrays in the CDF data products. Prior to SIT-3, this step of the algorithm was mostly glossed over in order to get CDFs created to support the test, but I am now going back to properly implement how the data arrays get stored. This PR also makes improvements to the decompression algorithm, namely making it so that an entire binary string can be decompressed at once, instead of individual values.

## Updated Files

- `imap_processing/cdf/config/imap_codice_l1a/b_variable_attrs.yaml`
  - Added attribute definition for `spin_sector` coordinate
  -  Various fixes and restructuring of data arrays, now utilizing `spin_sector` coordinate; split `unpack_science_data` into two functions, one for each `hi` and `lo`, since they have slightly differently ways of upacking data.
  - Modified dataset configurations to support `spin_sector`; removed Collapse Table lookups, as they are not needed because I am defining configurations via the `DATA_PRODUCT_CONFIGURATIONS` dictionaries.  
- `imap_processing/codice/decompress.py`
  - Refactored module so that an entire binary string can be decompressed, as opposed to individual integer values.
  - Updated tests to reflect new data array structure 
- `imap_processing/tests/codice/test_decompress.py`
  - Updated tests to reflect new way of decompressing data

### DIFF
--- a/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
@@ -36,6 +36,17 @@ energy_attrs:
     VALIDMAX: 127
     VAR_TYPE: support_data
 
+spin_sector_attrs:
+    <<: *default
+    CATDESC: Spin sector indicating range of spin angles
+    FIELDNAM: Spin sector
+    FORMAT: I4
+    LABLAXIS: spin sector
+    UNITS: ' '
+    VALIDMIN: 0
+    VALIDMAX: 1152
+    VAR_TYPE: support_data
+
 # <=== Labels ===>
 energy_label:
     CATDESC: Energy per charge (E/q) sweeping step
--- a/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
@@ -36,15 +36,26 @@ energy_attrs:
     VALIDMAX: 127
     VAR_TYPE: support_data
 
+inst_az_attrs:
+    <<: *default
+    CATDESC: Azimuth
+    FIELDNAM: Azimuth
+    FORMAT: I2
+    LABLAXIS: Azimuth
+    UNITS: ' '
+    VALIDMIN: 0
+    VALIDMAX: 31
+    VAR_TYPE: support_data
+
 spin_sector_attrs:
     <<: *default
     CATDESC: Spin sector indicating range of spin angles
     FIELDNAM: Spin sector
-    FORMAT: I4
+    FORMAT: I2
     LABLAXIS: spin sector
     UNITS: ' '
     VALIDMIN: 0
-    VALIDMAX: 1152
+    VALIDMAX: 11
     VAR_TYPE: support_data
 
 # <=== Labels ===>
--- a/imap_processing/codice/codice_l1a.py
+++ b/imap_processing/codice/codice_l1a.py
@@ -101,8 +101,10 @@ class CoDICEL1aPipeline:
         self.num_counters = config["num_counters"]
         self.num_energy_steps = config["num_energy_steps"]
         self.num_spin_sectors = config["num_spin_sectors"]
+        self.num_positions = config["num_positions"]
         self.variable_names = config["variable_names"]
         self.dataset_name = config["dataset_name"]
+        self.instrument = config["instrument"]
 
     def create_science_dataset(self, met: np.int64, data_version: str) -> xr.Dataset:
         """
@@ -135,6 +137,12 @@ class CoDICEL1aPipeline:
             dims=["epoch"],
             attrs=cdf_attrs.get_variable_attributes("epoch"),
         )
+        inst_az = xr.DataArray(
+            np.arange(self.num_positions),
+            name="inst_az",
+            dims=["inst_az"],
+            attrs=cdf_attrs.get_variable_attributes("inst_az_attrs"),
+        )
         spin_sector = xr.DataArray(
             np.arange(self.num_spin_sectors),
             name="spin_sector",
@@ -160,6 +168,7 @@ class CoDICEL1aPipeline:
         dataset = xr.Dataset(
             coords={
                 "epoch": epoch,
+                "inst_az": inst_az,
                 "spin_sector": spin_sector,
                 "energy": energy_steps,
                 "energy_label": energy_label,
@@ -169,21 +178,34 @@ class CoDICEL1aPipeline:
 
         # Create a data variable for each counter
         for variable_data, variable_name in zip(self.data, self.variable_names):
-            variable_data_arr = np.array(variable_data).reshape(
-                1, self.num_spin_sectors, self.num_energy_steps
-            )
+            # Data arrays are structured depending on the instrument
+            if self.instrument == "lo":
+                variable_data_arr = np.array(variable_data).reshape(
+                    1, self.num_positions, self.num_spin_sectors, self.num_energy_steps
+                )
+                dims = ["epoch", "inst_az", "spin_sector", "energy"]
+            elif self.instrument == "hi":
+                variable_data_arr = np.array(variable_data).reshape(
+                    1, self.num_energy_steps, self.num_positions, self.num_spin_sectors
+                )
+                dims = ["epoch", "energy", "inst_az", "spin_sector"]
+
+            # Get the CDF attributes
             cdf_attrs_key = (
                 f"{self.dataset_name.split('imap_codice_l1a_')[-1]}-{variable_name}"
             )
+            attrs = cdf_attrs.get_variable_attributes(cdf_attrs_key)
+
+            # Create the CDF data variable
             dataset[variable_name] = xr.DataArray(
                 variable_data_arr,
                 name=variable_name,
-                dims=["epoch", "spin_sector", "energy"],
-                attrs=cdf_attrs.get_variable_attributes(cdf_attrs_key),
+                dims=dims,
+                attrs=attrs,
             )
 
         # Add ESA Sweep Values and acquisition times (lo only)
-        if "_lo_" in self.dataset_name:
+        if self.instrument == "lo":
             self.get_esa_sweep_values()
             self.get_acquisition_times()
             dataset["esa_sweep_values"] = xr.DataArray(
@@ -274,43 +296,15 @@ class CoDICEL1aPipeline:
         sweep_table = sweep_data[sweep_data["table_idx"] == sweep_table_id]
         self.esa_sweep_values = sweep_table["esa_v"].values
 
-    def unpack_hi_science_data(self, science_values: str) -> None:
+    def unpack_science_data(self, science_values: str) -> None:
         """
-        Decompress, unpack, and restructure CoDICE-Hi data arrays.
+        Decompress, unpack, and restructure science data arrays.
 
         The science data within the packet is a compressed, binary string of
-        values.
-
-        Parameters
-        ----------
-        science_values : str
-            A string of binary data representing the science values of the data.
-        """
-        self.compression_algorithm = constants.HI_COMPRESSION_ID_LOOKUP[self.view_id]
-
-        # Decompress the binary string
-        science_values_decompressed = decompress(
-            science_values, self.compression_algorithm
-        )
-
-        # Divide up the data by the number of priorities or species
-        chunk_size = len(science_values_decompressed) // self.num_counters
-        science_values_unpacked = [
-            science_values_decompressed[i : i + chunk_size]
-            for i in range(0, len(science_values_decompressed), chunk_size)
-        ]
-
-        # TODO: Determine how to properly divide up hi data. For now, just use
-        #       arrays for each counter
-        self.data = science_values_unpacked
-
-    def unpack_lo_science_data(self, science_values: str) -> None:
-        """
-        Decompress, unpack, and restructure CoDICE-Lo data arrays.
-
-        The science data within the packet is a compressed, binary string of
-        values. These data need to be divided up by species or priorities,
-        and re-arranged into 2D arrays representing energy and spin angle.
+        values. These data need to be divided up by species or priorities (or
+        what I am calling "counters" as a general term), and re-arranged into
+        3D arrays representing spin sectors, positions, and energies (the order
+        of which depends on the instrument).
 
         Parameters
         ----------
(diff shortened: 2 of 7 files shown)


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -36,6 +36,17 @@ energy_attrs:
...
+spin_sector_attrs:
+    <<: *default
+    CATDESC: Spin sector indicating range of spin angles
+    FIELDNAM: Spin sector
+    FORMAT: I4
+    LABLAXIS: spin sector
+    UNITS: ' '
+    VALIDMIN: 0
+    VALIDMAX: 1152
+    VAR_TYPE: support_data
```
bourque: I have plans for a future PR that updates these yaml files to adhere to the proper nomenclature (https://lasp.colorado.edu/galaxy/display/IMAP/IMAP+Nomenclature+of+Science+Data+Products), so this may change, but works for now.

imap_processing/codice/codice_l1a.py
```diff
@@ -385,13 +427,15 @@ def create_hskp_dataset(
...
-    # packets/validation data that match the latest telemetry definitions
+    #       Once housekeeping packet definition file is re-generated with
+    #       updated version of space_packet_parser, can get fieldname and
+    #       catdesc info via:
+    #           for key, value in (packet.header | packet.data).items():
+    #               fieldname = value.short_description
+    #              catdesc = value.long_description
+    #       I am holding off making this change until I acquire updated
+    #       housekeeping packets/validation data that match the latest telemetry
+    #       definitions
```
bourque, line 440: Just doing some indenting here so that this appears nicer in PyCharm

imap_processing/codice/codice_l1a.py
```diff
@@ -457,8 +501,6 @@ def process_codice_l1a(file_path: Path, data_version: str) -> xr.Dataset:
     dataset : xarray.Dataset
         The ``xarray`` dataset containing the science data and supporting metadata.
     """
-    # TODO: Use new packet_file_to_dataset() function to simplify things
```
bourque, line 460: Moved this TODO to the top of the module with the other TODOs

imap_processing/codice/constants.py
```diff
@@ -179,25 +194,6 @@
...
-    1: 9,
-    2: 10,
-    3: 0,
-    4: 1,
-    5: 2,
-    6: 4,
-    7: 5,
-    8: 6,
-    9: 7,
-}
```
bourque, line 199: I have found that I can avoid using lookup tables for these because of the way I am defining `DATA_PRODUCT_CONFIGURATIONS`

imap_processing/codice/codice_l1a.py
```diff
@@ -262,33 +274,73 @@ def get_esa_sweep_values(self) -> None:
...
+        # Further divide up the data by energy levels
+        # The result is a [12,128] array representing 12 spin angles and 128
+        # energy levels
+        self.data = []
+        for counter_data in science_values_unpacked:
+            data_array = [
+                counter_data[i : i + self.num_energy_steps]
+                for i in range(0, len(counter_data), self.num_energy_steps)
+            ]
+            self.data.append(data_array)  # type: ignore[arg-type]
```
greglucas: Are you able to allocate the `(12, 128)` array initially and fill that in?

I'd find this a bit easier to follow as a double for loop over `(i, j)` and grabbing the data from the proper location as you go.

```python
data = np.empty(12, 128, dtype=...)
for i in range(12):
    chunk = science_values_decompressed[i*chunk_size:(i+1)*chunk_size]
    for j in range(128):
        # Turn it into an integer value as well?
        data[i, j] = int(chunk[j*num_energy_steps : (j+1)*num_energy_steps], 2)
```

imap_processing/codice/codice_l1a.py
greglucas: :+1: I like this style of indenting as well.

As an FYI, this actually isn't taken care of by the `packet_files_to_dataset` function as we aren't adding the short/long description to the output arrays I don't think.

imap_processing/codice/constants.py
```diff
@@ -76,73 +79,85 @@
 DATA_PRODUCT_CONFIGURATIONS = {
     CODICEAPID.COD_HI_INST_COUNTS_AGGREGATED: {
         "num_counters": 1,
-        "num_energy_steps": 128,
+        "num_energy_steps": 1,
+        "num_spin_sectors": 6,
```
greglucas: Are all of the data products going to have these same dimensions? I would have assumed each product would have the same "spin_sectors" (or at least consistent between hi/lo)

imap_processing/codice/constants.py
```diff
@@ -76,73 +79,85 @@
...
+        "num_energy_steps": 1,
+        "num_spin_sectors": 60,  # TODO: Double-check this
         "variable_names": HI_OMNI_SPECIES_NAMES,
         "dataset_name": "imap_codice_l1a_hi_omni",
     },
     CODICEAPID.COD_HI_SECT_SPECIES_COUNTS: {
         "num_counters": 4,
-        "num_energy_steps": 128,
+        "num_energy_steps": 1,
+        "num_spin_sectors": 1152,  # TODO: Double-check this
```
greglucas: This seems very large, the time resolution for that would need to be very high.

imap_processing/codice/constants.py
```diff
@@ -796,4 +793,5 @@
     252: 6815744,
     253: 7340032,
     254: 7864320,
+    255: 999999,
```
greglucas: One more 9 so it is always increasing? Is there a way to make these a fill_value instead if that is needed?

imap_processing/codice/decompress.py
```diff
@@ -26,115 +26,117 @@
...
 
     Returns
     -------
-    decompressed_value : int
-        The 24- or 32-bit decompressed value.
+    decompressed_values : list[int]
+        The 24- or 32-bit decompressed values.
     """
+    # Convert the binary string to a byte stream
+    compressed_bytes = bitarray.bitarray(compressed_binary).tobytes()
```
greglucas: Does this work instead?

```suggestion
    compressed_bytes = int(compressed_binary, 2).to_bytes((len(compressed_binary) + 7) // 8, byteorder='big')
```

greglucas: I'm trying to avoid bitarray/bitstring if possible as I think you're a bit lucky here because it is brought in via space_packet_parser, but that might be removed in the future (I hope): https://github.com/medley56/space_packet_parser/pull/51

imap_processing/codice/codice_l1a.py
greglucas: I had a thought for something even potentially cleaner/easier, which I think you're already doing above. Create a flat numpy array and then reshape it.

```python
np.array([int(science_values_decompressed[i*stepsize:(i+1)*stepsize], 2) for i in range(12*128)]).reshape(12, 128)
```

imap_processing/codice/codice_l1a.py
```diff
@@ -27,16 +27,22 @@
...
 from imap_processing.utils import group_by_apid, sort_by_time
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
-# TODO: Decom data arrays need to be decompressed
 # TODO: In decommutation, how to have a variable length data and then a checksum
 #       after it? (Might be fixed with new XTCE script updates)
 # TODO: Add support for decomming multiple APIDs from a single file
+# TODO: Add these as variables in CDF: SPIN_PERIOD, ST_BIAS_GAIN_MODE,
```
tmplummer, line 40: I'm curious if you write tickets for all of your TODOs. I know that I should be better about forcing myself to do that.

tmplummer (commented):
This all looks good to me. I like how you have chosen to structure your code. I will hold off on approving since there are some open items from Greg's feedback.

imap_processing/codice/codice_l1a.py
bourque: Historically I haven't been creating tickets for these, but I think I am going to start doing that, especially since we are now trying to have more incremental progress / smaller PRs / project boards tied to tickets.

imap_processing/codice/codice_l1a.py
bourque: I think this is not quite right because we first need to chunk of the data by the number of counters (a variable number) before re-arranging the data by number of spin sectors (12) and number of energy levels (128). But I see what you are saying and I will play around with this.

To be honest, I find your solution to be harder to read/understand :joy: but maybe we can find some middle ground.

bourque: Also I don't believe the values need to be converted to integers here. They should already be converted to a list of integers from the decompression algorithm.

greglucas: I might not be following what is going on here then. I thought you were grabbing ranges out of a string to try and put groups of "101010" together appropriately.

I hear you on some of it being personal preference with how to write nested for loops, so no worries to keep it how it makes the most sense to you, mine are purely suggestions.

> I think this is not quite right because we first need to chunk of the data by the number of counters (a variable number) before re-arranging the data by number of spin sectors (12) and number of energy levels (128). But I see what you are saying and I will play around with this.

It might help to put in the input shape / an example of what you are getting, and what you are trying to turn it into as a comment above this. If the number of counters is variable, does that mean that you have a 3D array coming out? So a `data.reshape(-1, 12, 128)` would get you an (ncounters, 12, 128) shaped array output.

bourque: > I might not be following what is going on here then. I thought you were grabbing ranges out of a string to try and put groups of "101010" together appropriately.

That was true recently, but with this PR I changed the decompression algorithm to spit out lists of (decompressed) integers instead, so that is now taken care of in the `decompress()` function call.

I took your latest suggestion as inspiration and came up with this:

```python
    def unpack_lo_science_data(self, science_values: str) -> None:
        """
        Decompress, unpack, and restructure CoDICE-Lo data arrays.

        The science data within the packet is a compressed, binary string of
        values. These data need to be divided up by species or priorities (or
        what I am calling "counters" as a general term), and re-arranged into
        2D arrays representing spin sector and energy.

        Parameters
        ----------
        science_values : str
            A string of binary data representing the science values of the data.
        """
        self.compression_algorithm = constants.LO_COMPRESSION_ID_LOOKUP[self.view_id]

        # Decompress the binary string into a list of integers
        science_values_decompressed = decompress(
            science_values, self.compression_algorithm
        )

        # This will hold the data to be later stored in the CDF. Each element
        # of the list will store the data for a particular counter
        self.data = []

        # Divide up the data by the number of counters
        chunk_size = len(science_values_decompressed) // self.num_counters
        science_values_per_counter = [
            science_values_decompressed[i : i + chunk_size]
            for i in range(0, len(science_values_decompressed), chunk_size)
        ]

        # Re-arrange the counter data to be a 2D array with a shape representing
        # [<num_spin_sectors>,<num_energy_steps]
        for counter_data in science_values_per_counter:
            reshaped_data = np.array(counter_data, dtype=np.int64).reshape(self.num_spin_sectors, self.num_energy_steps)

            # Add the re-arranged data to the list
            self.data.append(reshaped_data)
```

I also added some clarifying in-line comments and improved some variable names. Does this read well to you?

imap_processing/codice/constants.py
bourque: Good catch, i did intend for this to be `9999999`. I need to ask Joey about what to do here, so this is a temporary solution this this problem. I added a TODO here to make sure I don't forget.

imap_processing/codice/codice_l1a.py
greglucas: Yes, I like that 👍 good idea!

I'm not sure about the unpacking order but I still think you might be able to make `self.data` an array instead of a list of arrays? 
```
np.array(science_values_decompressed, dtype=np.int64).reshape(self.num_counters, self.num_spin_sectors, self.num_energy_steps)
```

Do you want `int64` here or are the counts uint and smaller size? (Can be updated later and use Tim's dtype idea within the metadata too if that is helpful)

imap_processing/codice/decompress.py
bourque: Yes, that works! I'll update that.

imap_processing/codice/codice_l1a.py
bourque: My motivation for making `self.data` a list here is because when I go to create the CDF, I then iterate over the list to create separate data variables for each counter. Is it better to iterate over a list or over a dimension of a `np.array`?

Good call on the `int64`, I think I can use `uint` here instead. I had just used `int64` without thinking much about it.

greglucas: Ahh, I see what you're saying now that they are all separate "variables" in your output file.
> Is it better to iterate over a list or over a dimension of a np.array

It is almost always better/faster to go to numpy arrays if possible (here you'd be creating one instead of ncounter arrays). So writing out the counters in your future steps I'd do something like:
```python
for i in range(self.num_counters):
    counter_variable = self.data[i, ...]  # (12, 128)
```
which should be very similar to your current way with lists.

bourque: This got me thinking, could this whole logic just be reduced to:

```python
self.data = np.array(science_values_decompressed, dtype=np.uint).reshape(self.num_counters, self.num_spin_sectors, self.num_energy_steps)
```

And avoid having to go over the for loops/chunking to restructure the data? I don't have a full understanding of how `np.reshape` is working under-the-hood, and if it is actually re-arranging the values how I would want it to.

imap_processing/codice/constants.py
bourque: Thank you, your comment made me realize that I messed up the logic a bit here. What I am calling "spin sectors" here is actually a representation of "number of positions with 12 spin angles each", and so I need to restructure the data a bit differently. Will address this in an upcoming commit.

imap_processing/codice/codice_l1a.py
greglucas: Yes, I believe you could!
https://numpy.org/doc/stable/reference/generated/numpy.reshape.html

I _think_ it is doing what you want by default with C-order because the furthest right dimension is changing fastest, and the counters is changing slowest (inside dimension). You can change the order if you want. Or you could also read it in with one order and then swap dimensions if needed too. But, I think you're good.

bourque: Thanks for the validation here. I am glad I spent an hour this morning accidentally re-implementing `numpy.reshape` from scratch 😅 What a fun exercise.

greglucas: > I am glad I spent an hour this morning accidentally re-implementing numpy.reshape from scratch 😅 What a fun exercise.

We've all been there! 😄  I don't think it was wasted time, because now I'm guessing you have a better appreciation of what is all going on under the hood and how the data is organized internally as well.

bourque:
Closes #767

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -36,6 +36,28 @@ energy_attrs:
     VALIDMAX: 127
     VAR_TYPE: support_data
 
+inst_az_attrs:
+    <<: *default
+    CATDESC: Azimuth
+    FIELDNAM: Azimuth
+    FORMAT: I2
+    LABLAXIS: Azimuth
```
bourque, line 44: I'm planning to provide better labels/descriptions for these in an upcoming PR where I address nomenclature

tmplummer (approved):
This all looks good to me!

imap_processing/codice/codice_l1a.py
```diff
@@ -264,31 +296,46 @@ def get_esa_sweep_values(self) -> None:
...
+        # Decompress the binary string into a list of integers
+        science_values_decompressed = decompress(
+            science_values, self.compression_algorithm
+        )
+
+        # Re-arrange the counter data
+        # For CoDICE-lo, data are a 3D arrays with a shape representing
+        # [<num_positions>,<num_spin_sectors>,<num_energy_steps>]
+        if self.instrument == "lo":
+            self.data = np.array(science_values_decompressed, dtype=np.uint).reshape(
```
tech3371: do you mean to add `dtype=np.uint32` or 64?

imap_processing/codice/decompress.py
```diff
@@ -26,115 +26,117 @@
...
+    compressed_values = list(compressed_bytes)
+    decompressed_values = [LOSSY_A_TABLE[item] for item in compressed_values]
+    return decompressed_values
 
 
-def _apply_lossy_b(compressed_value: int) -> int:
+def _apply_lossy_b(compressed_bytes: bytes) -> list[int]:
     """
-    Apply 8-bit to 32-bit Lossy B decompression algorithm.
+    Apply 8-bit to 32-bit Lossy A decompression algorithm.
```
tech3371:
```suggestion
    Apply 8-bit to 32-bit Lossy B decompression algorithm.
```

tech3371 (approved):
LGTM

imap_processing/codice/codice_l1a.py
bourque: Yes, I think this can be `uint32`. This prompted me to read up on what the difference between `np.uint` and `np.uint32` was, and it sounds like `uint32` is a better choice:

```
np.uint:

    This is an alias for the default unsigned integer type in NumPy, which is platform-dependent. On most platforms, np.uint is equivalent to np.uint32, meaning it is a 32-bit unsigned integer. However, this can vary based on the platform and version of NumPy.

np.uint32:

    This explicitly defines a 32-bit unsigned integer. It can store values from 0 to 232−1232−1, which is 0 to 4,294,967,295. This is a fixed-size type and is not platform-dependent.
```

imap_processing/codice/decompress.py
bourque: Thanks for catching that! It was a bad copy/paste

PR #763, opened 2024-08-21
MNT: Add useful error message when packet to datasets fails

greglucas:
# Change Summary

## Overview

This occurs when there is a nested packet file definition with different variables for packets with the same apid. For example, with data spread across multiple packets and that definition being written into the XTCE logic.

Implementing the suggestion from here:
https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/743#discussion_r1707691379

@anamanica let me know if you think there is a better wording for this or if there is some other error message that would have helped in your debugging.


tmplummer (approved):
Quick and easy... LGTM.

greglucas:
Force-pushing a new commit trying to retrigger pre-commit since I couldn't see a button to just re-run that single failing job :(

PR #781, opened 2024-08-26
FIX: Avoid loss of precision when casting in packet loading

### LINKED ISSUE #780: BUG - Reading in derived float values loses precision
### Description of the issue

The data onboard the spacecraft can be packed as unsigned integers, but then a linear conversion factor can be applied on the ground to turn it into a float. This is an issue in the `packet_file_to_datasets()` routine when we try to infer that the variable is a UINT and cast the value to integer even though it has been derived/calibrated into a float. Causing the value 2.1 to silently become 2.

Noticed this while look at SWE's decommutation routines, but don't think we have any explicitly in the project currently.

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

greglucas:
# Change Summary

## Overview

When using derived values there can be situations where a linear conversion factor is applied to a uint8 value to turn a raw measurement into a float temperature value for instance. These are represented as a small uint datatype onboard, but need to be represented as a float or larger integer datatype on the ground so we don't lose precision. Previously we were getting 2.1 cast to 2 after the derived types were attempted to be cast to their onboard types.

closes #780

### DIFF
--- a/imap_processing/utils.py
+++ b/imap_processing/utils.py
@@ -299,10 +299,18 @@ def _create_minimum_dtype_array(values: list, dtype: str) -> npt.NDArray:
     array : np.array
         The array of values.
     """
+    # Create an initial array and then try to safely cast it to the desired dtype
+    x = np.asarray(values)
     try:
-        return np.array(values, dtype=dtype)
+        # ValueError: when trying to cast strings (enum states) to ints
+        y = x.astype(dtype, copy=False)
+        # We need to compare the arrays to see if we trimmed any values by
+        # casting to a smaller datatype (e.g. float64 to uint8, 2.1 to 2)
+        if np.array_equal(x, y):
+            return y
     except ValueError:
-        return np.array(values)
+        pass
+    return x
 
 
 def packet_file_to_datasets(
--- a/imap_processing/tests/test_utils.py
+++ b/imap_processing/tests/test_utils.py
@@ -97,10 +97,26 @@ def test_packet_file_to_datasets(use_derived_value, expected_mode):
     np.testing.assert_array_equal(data["mode"], [expected_mode] * len(data["mode"]))
 
 
-def test__create_minimum_dtype_array():
+@pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


imap_processing/utils.py
```diff
@@ -299,10 +299,18 @@ def _create_minimum_dtype_array(values: list, dtype: str) -> npt.NDArray:
...
     """
+    # Create an initial array and then try to safely cast it to the desired dtype
+    x = np.asarray(values)
     try:
-        return np.array(values, dtype=dtype)
+        # ValueError: when trying to cast strings (enum states) to ints
+        y = x.astype(dtype, copy=False)
+        # We need to compare the arrays to see if we trimmed any values by
+        # casting to a smaller datatype (e.g. float64 to uint8, 2.1 to 2)
+        if np.array_equal(x, y):
```
greglucas, line 309: This is all somewhat unfortunate in how many "arrays" we'll be creating here. If anyone has ideas for a better way to do this I'm interested in hearing it.

We are potentially creating 3 separate arrays:
1. initial asarray from a list -> creates an int64
2. Trying to cast to a smaller dtype -> creates a uint8
3. Doing this array comparison in a higher-order dtype space.

They are all happening in numpy c-level functions so shouldn't be a huge hit, but it would be nice to not do this array comparison just to check loss of precision.

tmplummer: My thought would be to modify the `_get_minimum_numpy_datatype` function to take the derived value logic into account and return the appropriate datatype.

greglucas: Really good suggestion, thanks.

I opened up: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/786
with an alternative approach, which is actually able to get rid of the numpy try/except array creation too. The one downside is there isn't a great way to test all of the XTCE stuff currently :-/ I did test it on the new SWE XTCE file that was failing and it looks good, but I don't have a great way to parameterize that PR like we do with this one.

greglucas:
Closing in favor of the other PR

PR #783, opened 2024-08-26
ENH: Add Enumeration states to XTCE parser

### LINKED ISSUE #782: ENH - Add enumeration states to XTCE parser
### Summary

In the spreadsheets we receive, there is a `convertAs` column that we are using to do analog conversions, but it also has the `STATE` option as well to provide enumerations. SWE's packet definition has these in it and the packet parser is currently unable to represent these states. We should add those in.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/87ede4dfdfac521b478cf611e9e8fa0084ba2f1d/imap_processing/swe/packet_definitions/swe_packet_definition.xml#L54-L59

### Requirements

The more specific the better.
  - [ ] Add Enumerations to the XTCE parser based on the convertAs column and States tab.

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview

This adds the ability to include derived enumeration states from the "States" tab in the spreadsheets.

## Testing

Manual testing with SWE's packet definition. Unfortunately it looks like we don't have any tests for this utility and it is a pain to mock up the spreadsheets. This should be an action item to do in the future though.

closes #782


imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -329,6 +328,10 @@ def _add_parameter(self, row: pd.Series, total_packet_bits: int) -> None:
             # Go look up the conversion in the AnalogConversions tab
             # and add it to the encoding
             self._add_analog_conversion(row, encoding)
+        elif row["convertAs"] == "STATE":
```
tech3371, line 331: I didn't notice that this column and that tab was connected. I knew that they stored enums(derived) values in that tab but didn't connect those two together. Nice!

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -363,6 +366,34 @@ def _add_analog_conversion(self, row: pd.Series, encoding: Et.Element) -> None:
...
+        # Lookup the enumeration states for this parameter from the States sheet
+        state_sheet = self.sheets["States"]
+        state_sheet = state_sheet.loc[
+            (state_sheet["packetName"] == row["packetName"])
+            & (state_sheet["mnemonic"] == row["mnemonic"])
+        ]
+        for _, state_row in state_sheet.iterrows():
+            enumeration = Et.SubElement(enumeration_list, "xtce:Enumeration")
+            enumeration.attrib["value"] = str(state_row["value"])
+            enumeration.attrib["label"] = str(state_row["state"])
```
tech3371, line 395: I am wondering if this function produces something like this?
```
	        <xtce:EnumeratedParameterType name="QUARTER_CYCLE_ENUM" signed="false">
                <xtce:IntegerDataEncoding sizeInBits="5" encoding="unsigned"/>
                <xtce:EnumerationList>
                    <xtce:Enumeration label="FIRST" value="0"/>
                    <xtce:Enumeration label="SECOND" value="1"/>
		   <xtce:Enumeration label="THIRD" value="2"/>
                    <xtce:Enumeration label="FORTH" value="3"/>
                </xtce:EnumerationList>
```

I saw most of it but didn't see it add this line. I think we need that.
```
                <xtce:IntegerDataEncoding sizeInBits="5" encoding="unsigned"/>
```

greglucas: That is taken care of in the above `_add_parameter_type()`, which adds the encodings. Then here we modify that parameter type to be `EnumeratedParameterType` **instead of** `IntegerParameterType`. It should be there if you run this on the swapi or swe definitions (could you verify it works for you as well). This is where tests would be nice...

tech3371: I ran locally and it looks correct.

tech3371 (approved):

greglucas:
> Unfortunately it looks like we don't have any tests for this utility and it is a pain to mock up the spreadsheets. This should be an action item to do in the future though.

Of course, I totally missed that we had an Excel sheet and expected output in the `test_data` folder! 🐑 

I've changed from that Excel sheet to a dynamically generated Excel file instead as I suggested initially. The expected output file has a minimal diff with the new enumeration updates.

greglucas:
@bourque , I am pinging you to see if you agree with updating/changing the tests here now.

Codecov is now 100% for this diff.

bourque (approved):
Yes, looks good to me!

PR #786, opened 2024-08-26
FIX: Avoid loss of precision when casting in packet loading (alternative)

### LINKED ISSUE #780: BUG - Reading in derived float values loses precision

greglucas:
# Change Summary

## Overview

When using derived values there can be situations where a linear conversion factor is applied to a uint8 value to turn a raw measurement into a float temperature value for instance. These are represented as a small uint datatype onboard, but need to be represented as a float or larger integer datatype on the ground so we don't lose precision. Previously we were getting 2.1 cast to 2 after the derived types were attempted to be cast to their onboard types.

closes #780

### DIFF
--- a/imap_processing/utils.py
+++ b/imap_processing/utils.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from typing import Optional, Union
 
 import numpy as np
-import numpy.typing as npt
 import pandas as pd
 import xarray as xr
 from space_packet_parser import parser, xtcedef
@@ -228,8 +227,8 @@ def create_dataset(
 
 
 def _get_minimum_numpy_datatype(  # noqa: PLR0912 - Too many branches
-    name: str, definition: xtcedef.XtcePacketDefinition
-) -> str:
+    name: str, definition: xtcedef.XtcePacketDefinition, use_derived_value: bool = True
+) -> Optional[str]:
     """
     Get the minimum datatype for a given variable.
 
@@ -239,6 +238,8 @@ def _get_minimum_numpy_datatype(  # noqa: PLR0912 - Too many branches
         The variable name.
     definition : xtcedef.XtcePacketDefinition
         The XTCE packet definition.
+    use_derived_value : bool, default True
+        Whether or not the derived value from the XTCE definition was used.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


imap_processing/utils.py
```diff
@@ -388,9 +378,7 @@ def packet_file_to_datasets(
             {
                 key.lower(): (
                     "epoch",
-                    _create_minimum_dtype_array(
-                        list_of_values, dtype=datatype_mapping[apid][key]
-                    ),
+                    np.asarray(list_of_values, dtype=datatype_mapping[apid][key]),
```
tmplummer, line 391: Does numpy default to 64-bit floats? And if so, do we have any way to get the CDF products to use 32-bit floats where we want them?

tmplummer (approved):
Looks good to me. Just one question about 64 vs 32 bit floats.

imap_processing/utils.py
greglucas: Yes and yes.

I am actually getting 32-bit floats in another packet definition where the floats are brought down directly in the telemetry stream, so `space_packet_parser` emits the Python floats and then we can use the bits to determine 32/64 bits correctly: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/87ede4dfdfac521b478cf611e9e8fa0084ba2f1d/imap_processing/utils.py#L264-L269
This situation is purely for when a conversion factor gets applied and we just shrug and say, not much I can know about it from the packet definition since you told me it was unsigned 8 bits in the packet, but then you multiplied it by 2.5 and all bets are off for what you want me to turn that into (float, new int, string enumeration, ...).

Ultimately, this gets to your earlier suggestions though that this is really just a "first cut" and we should be putting `dtype` into the CDF metadata attributes yaml definitions and doing the casting there when creating the datasets and saving the products.

PR #791, opened 2024-08-28
SPICE geometry enums and IMAP state function

### LINKED ISSUE #787: Add some SPICE infrastructure and simple geometry function
### Algorithm Description:
This is mostly about getting some of the convenience pieces in place similar to what was done on EMUS. One addition could be a function for computing a simple geometry value such as basis vectors of a given frame in another frame.

tmplummer:
# Change Summary

## Overview
- Add spice infrastructure code for spice bodies and frames
- Add geometry functions for computing IMAP spacecraft state (position, velocity)
- Add IMAP SPK kernel from Nick's 366 day simulation
- Add test coverage for new geometry functions

## New Files

- imap_processing/tests/spice/test_data/imap_spk_demo.bsp
   - 366 day simulated IMAP ephemeris kernel

## Deleted Files
N/A

## Updated Files
   - Add spice infrastructure code for spice bodies and frames
   - Add geometry functions for computing IMAP spacecraft state (position, velocity)
- imap_processing/tests/spice/test_data/imap_test_metakernel.template
   - Add the new SPK kernel to this test template
   - Add test coverage for new geometry functions

## Testing
2 Unit tests added

closes #787


imap_processing/spice/geometry.py
```diff
@@ -1 +1,128 @@
...
+import spiceypy as spice
+
+from imap_processing.spice.kernels import ensure_spice
+
+
+class SpiceId(NamedTuple):
+    """Class that represents a unique identifier in the NAIF SPICE library."""
+
+    strid: str
+    numid: int
```
greglucas: I think we can get rid of this NamedTuple and just use your Enums directly below. Maybe use IntEnum instead. But then you'd just have `SpiceBody.IMAP.name` which would still return the string IMAP. (The one difference you have is SSB which I think could just be written out in the long form)
```suggestion
```

imap_processing/spice/geometry.py
```diff
@@ -1 +1,128 @@
...
+
+
+class SpiceBody(Enum):
```
greglucas: This is great, I really like this style of defining these as Enums.

Do we ever run into a situation of possibly getting out of sync with the SPICE kernels? Should these be dynamically looked up from the Kernels themselves somehow?

imap_processing/spice/geometry.py
```diff
@@ -1 +1,128 @@
...
+    IMAP_CODICE = SpiceId("IMAP_CODICE", -43400)
+    IMAP_HIT = SpiceId("IMAP_HIT", -43500)
+    IMAP_IDEX = SpiceId("IMAP_IDEX", -43700)
+    IMAP_GLOWS = SpiceId("IMAP_GLOWS", -43750)
+
+
+def imap_state(
+    et: Union[np.ndarray, float],
+    ref_frame: Optional[SpiceFrame] = None,
+    observer: Optional[SpiceBody] = None,
```
greglucas: You can put the defaults directly in the call.
```suggestion
    ref_frame: SpiceFrame = SpiceFrame.ECLIPJ2000,
    observer: SpiceBody = SpiceBody.SUN,
```

imap_processing/spice/geometry.py
```diff
@@ -1 +1,128 @@
...
+        ref_frame = SpiceFrame.ECLIPJ2000
+    if observer is None:
+        observer = SpiceBody.SUN
+    state, light_time = ensured_spkezr(
+        SpiceBody.IMAP.name, et, ref_frame.name, "NONE", observer.name
+    )
+    return state
+
+
+def ensured_spkezr(
```
greglucas: How would you feel about making these more human readable? I think this is actually one of the things I don't like about `spiceypy`, where they just reiterated the low-level spice function names. So then I need to know SPICE to be able to use it.

If you make this: `calc_spacecarft_state()` I would find that much easier to parse what is going on. Then only you and other SPICE experts need to know this is ensured and using the underlying spkezr function under the hood.

Also, how many of these arguments will we actually want to change? Maybe limit the number to the relevant ones we'll modify and chosen defaults for the others.

imap_processing/tests/spice/test_geometry.py
```diff
@@ -1 +1,35 @@
...
+    # The queried ET, 798033670 is ~2025-04-16T00:00:00.0
+    state, lt = ensured_spkezr(
+        SpiceBody.IMAP.name,
+        798033670,
+        SpiceFrame.ECLIPJ2000.name,
+        "NONE",
+        SpiceBody.EARTH.name,
+    )
+    assert state.shape == (6,)
+    assert isinstance(lt, float)
```
greglucas: Maybe parametrize this test to input a list and array of quantities as well to make sure you get back the expected iterables?

imap_processing/spice/geometry.py
tmplummer: Yeah, good question. It would be great to parse certain kernels for this information. I'd have to think about how to approach that. For example, do we check those kernels into the repo?  The likelihood that any of the spacecraft or instrument values change should be small. I can write a ticket up for this but think that for now it is a nice to have that would take too much development time.

imap_processing/spice/geometry.py
tmplummer: Hahaha... good point

imap_processing/spice/geometry.py
tmplummer: Perhaps I just need to make this a private function? The intent here is only to wrap the spicey function with `ensure_spice`. The general paradigm to follow when using `ensure_spice` is to wrap the lowest level functions you can. That reduces ensure spice recursion and lowers the processing done on the wrapped function retries.
I provided the above function `imap_state()` that does what your `calc_spacecraft_state()` function would do and is intended to be used by the team rather than this function.

imap_processing/spice/geometry.py
tmplummer: The only issue I see with that is that some SPICE names have spaces in them.

greglucas: Oh interesting... and unfortunate. It would be much cleaner as a normal Enum if we could somehow manage that.

imap_processing/spice/geometry.py
greglucas: :+1: Completely agree that this is very minor. I was just curious, so definitely punt on it into the very distant future.

imap_processing/spice/geometry.py
greglucas: Ahh, I see, I don't know how I missed that you actually were making the nicer name. Thank you!!

This is where I think the global state and loading of kernels would be beneficial, and then you could wrap your original function with `@ensure_spice` rather than the low-level versions as a two-step process. No need to do the global thing here, I'm going to give that some thought and try to propose something.

> Perhaps I just need to make this a private function?

Yes, I think that might be good. The other possible thing to do is just inline the ensure_spice into your original function? It is only a few lines.

imap_processing/spice/geometry.py
```diff
@@ -1 +1,128 @@
...
+    """Class that represents a unique identifier in the NAIF SPICE library."""
+
+    strid: str
+    numid: int
+
+
+class SpiceBody(Enum):
+    """Enum containing SPICE IDs for bodies that we use."""
+
+    # A subset of IMAP Specific bodies as defined in imap_wkcp.tf
```
laspsandoval, line 24: Make certain to put a TODO for change them to positive once IMAP is launched. I believe that's the standard.

imap_processing/spice/geometry.py
```diff
@@ -1 +1,128 @@
...
+from enum import Enum
+from typing import NamedTuple, Optional, Union
+
+import numpy as np
+import spiceypy as spice
+
+from imap_processing.spice.kernels import ensure_spice
+
+
+class SpiceId(NamedTuple):
```
laspsandoval: Would this not work here?

https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/bodn2c_c.html
sun_id = spice.bodn2c("SUN")

imap_processing/spice/geometry.py
```diff
@@ -1 +1,128 @@
...
+    IMAP_GLOWS = SpiceId("IMAP_GLOWS", -43750)
+
+
+def imap_state(
+    et: Union[np.ndarray, float],
+    ref_frame: Optional[SpiceFrame] = None,
+    observer: Optional[SpiceBody] = None,
+) -> Union[np.ndarray, Iterable[np.ndarray]]:
+    """
+    Get the state (position and velocity) of the IMPA spacecraft.
```
laspsandoval: Get the state (position and velocity) of the **IMAP** spacecraft.

laspsandoval (approved):
Looks good to me. Thanks for the addition. Just a few changes along with Greg's.

imap_processing/spice/geometry.py
tech3371:
```suggestion
    Get the state (position and velocity) of the IMAP spacecraft.
```

tech3371 (approved):
With little I know, it looks good to me. I was able to follow it on high level.

imap_processing/spice/geometry.py
tmplummer: I don't see any spaces in IMAP names and don't know of any SPICE names with spaces so I will implement it with just `IntEnum` for now.

imap_processing/spice/geometry.py
tmplummer: Not according to the documentation:

> THE SPICE convention uses negative integers as spacecraft ID codes. The code assigned to interplanetary spacecraft is normally the negative of the code assigned to the same spacecraft by JPL's Deep Space Network (DSN) as determined the NASA control authority at Goddard Space Flight Center.

The current SPICE vehicle code assignments:

      NAIF ID     NAME
      ________    ____________________
      -1          'GEOTAIL'
      -3          'MOM'
      -3          'MARS ORBITER MISSION'
        ...
      -41         'MARS EXPRESS'
      -43         'IMAP'
       ...
https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/naif_ids.html#Spacecraft

Issue #727, opened 2024-08-05
MAG L1A compression algorithm


maxinelasp:
### Algorithm Description:
MAG has requested a new compression algorithm for their L1A processing, which uses Fibonacci encoding and zig-zag encoding to compress vectors.

### Other Notes:
The description for the algorithm is in the updated algorithm document. I also got some code.

PR #801, opened 2024-08-30
Mag l1a compression

maxinelasp:
# Change Summary
This is a pretty involved PR. I would like to thank in advance anyone who reviews. I don't think breaking this code up further would make sense, or I would have done so.

I'd recommend starting by reading the docstring for "process_compressed_vectors" in mag_l1a_data.py, as it gives a good overview on the overall algorithm.

Note: "process_uncompressed_vectors" was moved into a new method but otherwise unchanged. 

## Overview

- Added MAG decompression option
- Added tests

## Updated Files

- mag_l1a_data.py
   - Added 7 new methods to decode compressed data

- test_mag_l1a
   - Added tests
- mag_l0_data
  - Minor fix for mypy

## Testing

Testing is pretty extensive for this change, since it's so complex. I generated test data primarily using a piece of code that the MAG team provided for me.

Closes issue #727


maxinelasp:
pre-commit.ci autofix

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,463 @@ def to_signed16(n: int) -> int:
...
+        -------
+        (primary, secondary): (numpy.ndarray, numpy.ndarray)
+            Two arrays, each containing tuples of (x, y, z, sample_range) for each
+            vector sample.
+        """
+        bit_array = np.unpackbits(vector_data)
+        # The first 8 bits are a header - 6 bits to indicate the compression width,
+        # 1 bit to indicate if there is a range data section, and 1 bit spare.
+        compression_width = int("".join([str(i) for i in bit_array[:6]]), 2)
+        has_range_data_section = int(str(bit_array[6]), 2)
```
alastairtree:
```suggestion
        has_range_data_section = (bit_array[6] == 1)
```
be nicer if this was typed as a bool?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,463 @@ def to_signed16(n: int) -> int:
...
+
+        # The full vector includes 3 values of compression_width bits, plus 2 bits for
+        # the range
+        uncompressed_vector_size = compression_width * 3 + 2
+        # plus 8 to get past the compression width and range data section
+        first_vector_width = uncompressed_vector_size + 8
```
alastairtree: I dont understand the +8 here - you are aready skipping 8 bits in bit array on line 546?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,463 @@ def to_signed16(n: int) -> int:
...
+        """
+        bit_array = np.unpackbits(vector_data)
+        # The first 8 bits are a header - 6 bits to indicate the compression width,
+        # 1 bit to indicate if there is a range data section, and 1 bit spare.
+        compression_width = int("".join([str(i) for i in bit_array[:6]]), 2)
+        has_range_data_section = int(str(bit_array[6]), 2)
+
+        # The full vector includes 3 values of compression_width bits, plus 2 bits for
+        # the range
+        uncompressed_vector_size = compression_width * 3 + 2
```
alastairtree: Would be tempted to 3 as a constant NUMBER_OF_AXIS and 2 as BIT_WIDTH_OF_VECTOR_RANGE for reader sanity

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,463 @@ def to_signed16(n: int) -> int:
...
+        # plus 8 to get past the compression width and range data section
+        first_vector_width = uncompressed_vector_size + 8
+
+        first_vector = MagL1a.unpack_one_vector(
+            bit_array[8:first_vector_width], compression_width, True
+        )
+
+        end_vector = (
+            len(bit_array)
+            - (primary_count + secondary_count) * 2 * has_range_data_section
```
alastairtree: range_data_section has one less range than total as first range is in the header

imap_processing/mag/constants.py
```diff
@@ -55,3 +55,49 @@ class PrimarySensor(Enum):
...
+    9227465,
+    14930352,
+    24157817,
+    39088169,
+    63245986,
+    102334155,
+    165580141,
+]
+
+MAX_FINE_TIME = 65535  # maximum 16 bit unsigned int
```
tech3371: ?
```suggestion
MAX_FINE_TIME = np.iinfo(np.uint16).max 
```

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -42,6 +42,10 @@ def mag_l1a(packet_filepath: Path, data_version: str) -> list[xr.Dataset]:
     norm_data = packets["norm"]
     burst_data = packets["burst"]
 
+    print(np.unpackbits(norm_data[0].VECTORS))
+    print(len(np.unpackbits(norm_data[0].VECTORS)))
+    print(len(norm_data[0].VECTORS))
```
tech3371:
```suggestion
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -287,10 +296,50 @@ def calculate_vector_time(
...
+        secondary_count : int
+            Count of the number of secondary vectors.
+        compression : int
+            Flag indicating if the data is compressed (1) or uncompressed (0).
+
+        Returns
+        -------
+        (primary, secondary): (numpy.ndarray, numpy.ndarray)
+            Two arrays, each containing tuples of (x, y, z, sample_range) for each
+            vector sample.
```
tech3371, line 332: very helpful comments!

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,467 @@ def to_signed16(n: int) -> int:
...
+
+    @staticmethod
+    def process_compressed_vectors(  # noqa: PLR0912, PLR0915
+        vector_data: np.ndarray, primary_count: int, secondary_count: int
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """
+        Given raw compressed packet data, process into Vectors.
+
+        To do this, we need to decode the compressed data. The compressed data starts
+        with an 8 bit header that defines the width of the uncompressed vectors and
```
tech3371, line 498: I think we decided to use decompressed instead of uncompressed. Can you change it in other places as well?
```suggestion
        with an 8 bit header that defines the width of the decompressed vectors and
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,467 @@ def to_signed16(n: int) -> int:
...
+        if there is a range data section. Then, the vector data follows, then the range
+        data section if it exists.
+
+        To decode, we start by decoding the first compression_width bits. This is an
+        uncompressed primary vector with range. Then, we proceed through the compressed
+        data, where each value is fibonacci and zig-zag encoded. This means each value
+        ends in 2 sequential ones (11). We split the data along these numbers until
+        we reach primary_count.
```
tech3371: Do you mean this?
```suggestion
        we reach last primary_count.
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,467 @@ def to_signed16(n: int) -> int:
...
+
+        Returns
+        -------
+        (primary, secondary): (numpy.ndarray, numpy.ndarray)
+            Two arrays, each containing tuples of (x, y, z, sample_range) for each
+            vector sample.
+        """
+        bit_array = np.unpackbits(vector_data)
+        # The first 8 bits are a header - 6 bits to indicate the compression width,
+        # 1 bit to indicate if there is a range data section, and 1 bit spare.
```
tech3371, line 541: These doc strings and comments are helping a lot to follow with your code.

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,467 @@ def to_signed16(n: int) -> int:
...
 
         return (
-            np.array(primary_vectors, dtype=np.int64),
-            np.array(secondary_vectors, dtype=np.int64),
+            np.array(primary_vectors, dtype=np.int32),
+            np.array(secondary_vectors, dtype=np.int32),
         )
+
+    @staticmethod
+    def process_compressed_vectors(  # noqa: PLR0912, PLR0915
```
tech3371: I would really feel good if MAG reviewed this function. I am only understanding small portion and very high level. I apologize that I am not much of help here.

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Good find, thank you!

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: I would use "decompressed" to mean the data after I had finished with the algorithm to turn it from compressed -> no longer compressed. In this case, the data was never compressed to start. I can change it if that's what we have decided "decompressed" means, but I want to call that out. 

Also, MAG uses "uncompressed" which is where that word came from. Ultimately, the decompressed/uncompressed word would only be used in this code and not in any of the outputs - so while I normally would be all the way for using the same word, in this case, I'd prefer to match the MAG team and not the other instruments.... but I am definitely open to being convinced. Where else are we using "decompressed?"

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: This is an awkward sentence, and I will rephrase it, thanks! It should be "until we reach primary_count vectors"

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Good! I'm glad. I would have been lost without them 😅

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Thank you for looking, you have been helpful! I have asked them for a review, particularly of the tests.

imap_processing/mag/constants.py
maxinelasp: This change actually randomly fixed a mypy error 🤷  so thanks for pointing it out!

imap_processing/mag/l1a/mag_l1a_data.py
tech3371: I don't feel strongly about this since it's just in the doc string. I did look at your function with name `uncompressed` and it mades sense to keep it `uncompressed`. May be it's better to keep the comment to match that. Ok, ignore my comment since the data is not compressed.

maxinelasp: Cool, thanks for bringing it up! It's good to get the check to make sure my words are good 😄

imap_processing/mag/l1a/mag_l1a_data.py
laspsandoval: Is there a way to break this up at all?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,467 @@ def to_signed16(n: int) -> int:
...
+        code : numpy.ndarray
+            The code to decode, in the form of an array of bits (eg [0, 1, 0, 1, 1]).
+            This should always end in 2 ones (which indicates the end of a fibonacci
+            encoding).
+
+        Returns
+        -------
+        value: int
+            Signed integer value, with fibonacci and zig-zag encoding removed.
+        """
```
laspsandoval, line 994: Could there be a case in which code is too short?
if len(code) < 2:
        raise ValueError(f"Input code too short: {code}")

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,467 @@ def to_signed16(n: int) -> int:
...
+        secondary_count : int
+            Count of the number of secondary vectors.
+
+        Returns
+        -------
+        (primary, secondary): (numpy.ndarray, numpy.ndarray)
+            Two arrays, each containing tuples of (x, y, z, sample_range) for each
+            vector sample.
+        """
+        bit_array = np.unpackbits(vector_data)
```
laspsandoval, line 539: Is there a reason to pack the bits and then unpack them in the same program?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +477,467 @@ def to_signed16(n: int) -> int:
...
+        vector_data : numpy.ndarray
+            Vector data for the vector to unpack. This is uncompressed data as a numpy
+            array of bits (the output of np.unpackbits).
+        width : int
+            The width of each vector component in bits. This needs to be a multiple of
+            8 (including only whole bytes).
+        has_range : int
+            1 if the vector data includes range data, 0 if not. The first vector always
+            has range data, it is only if the compression fails and we revert to
+            uncompressed data partway through that we skip the range.
```
laspsandoval: So would it either be one of three scenarios:
1. compressed - range data
2. compressed + uncompressed - range data
3. uncompressed - no range data

What do you mean by "partway through"? This is more for my own understanding than anything.

laspsandoval (commented):
Interesting concept. Only a few suggestions.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: It's already fairly broken up, personally I don't think there are other obvious things to pull out. But, if you'd prefer, I can break it up further. I know it's a bit of a monster method.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: I think it is ok for the length to be equal to 2 (this is how you represent 0.) If it's shorter than two, then it will be caught with the existing check... although without the nice error message. I'll add that in, good suggestion

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: The packing in the tests is to get the test data to line up with the real output data. It's basically convenient to have the test data explicitly in bits, and then pack that into bytes to match what the expected packet data looks like, because the bytes are very obscure. In the actual steps, this is the only unpack step, and then later they are re-packed in the uncompressed case, but it's still more convenient to use the output of `np.unpack`. 

Long story short, in the actual data, the data is already packed, so there isn't a packing then unpacking step like there is in the test data.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: You're correct. I'll reword this sentence because I understand your confusion. The `has_range` section is totally separate from the compressed/uncompressed bits. What this sentence is trying to say, is that the uncompressed data does not include a range value. But that should probably be said somewhere else. 

It is also possible to have compressed + uncompressed with no range data. If the range stays the same for the entire packet, then we do not have a range data section. Otherwise, we retrieve all the range data for all the vectors from the range data, regardless if the vectors are compressed or uncompressed. 

I'll rewrite this sentence because it is confusingly tying things together when they should be separate, thanks for calling that out!

vmartinez-cu:
Nice job with your doc strings and comments. The details are helpful

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """
+        Given raw compressed packet data, process into Vectors.
+
+        To do this, we need to decode the compressed data. The compressed data starts
+        with an 8 bit header that defines the width of the uncompressed vectors and
+        if there is a range data section. Then, the vector data follows, then the range
+        data section if it exists.
+
+        To decode, we start by decoding the first compression_width bits. This is an
```
sdhoyt, line 502: >To decode, we start by decoding the first compression_width bits. This is an
         uncompressed primary vector with range.

I'm having trouble understanding this part. Looking at the code, the compression_width is uncompressed 6 bits of the header. What does it mean that this is the "uncompressed primary vector with range"? Is this giving the bit length of the compressed data in the primary vector?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+    @staticmethod
+    def process_compressed_vectors(  # noqa: PLR0912, PLR0915
+        vector_data: np.ndarray, primary_count: int, secondary_count: int
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """
+        Given raw compressed packet data, process into Vectors.
+
+        To do this, we need to decode the compressed data. The compressed data starts
+        with an 8 bit header that defines the width of the uncompressed vectors and
+        if there is a range data section. Then, the vector data follows, then the range
```
sdhoyt, line 499: It looks like range data is a kind of MAG data that may or may not be present,  not something used specifically for the decompression scheme, correct?

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+
+    expected_range_primary = [3, 0, 0, 0, 0, 0, 0, 0, 1, 1, 2, 2, 3, 3, 3, 3]
+
+    range_secondary = "000000000000000101101011111101"
+    expected_range_secondary = [3, 0, 0, 0, 0, 0, 0, 0, 1, 1, 2, 2, 3, 3, 3, 1]
+    # 16 bit width with range section
+    headers = "01000010"
+    # TODO: if there is a number of vectors that is not a multiple of 8,
+    #  is there also padding at the end of the range section?
+    input_data = np.array(
```
alastairtree, line 485: Yes the range section ends on the byte boundary with padding if needed but as it is the end of the packet it can be ignored. Guess this TODO can be removed/resolved.

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+    (primary, secondary) = MagL1a.process_compressed_vectors(input_data, 16, 16)
+
+    assert np.array_equal(primary[0], primary_expected[0])
+    assert np.array_equal(secondary[0], secondary_expected[0])
+
+    assert primary.shape[0] == 16
+    assert secondary.shape[0] == 16
+
+    assert primary.shape[1] == 4
+    assert secondary.shape[1] == 4
```
alastairtree, line 470: The 16 and 4 here are a bit magical. I think you are checking 16 vectors per packet, and 4 fields (x,y,z,r). Maybe add a comment? Also I think this is kinda covered already by the `np.array_equal` assertion above?

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+        dtype=np.uint8,
+    )
+
+    input_data = np.packbits(input_data)
+
+    for i in range(len(expected_range_primary)):
+        primary_expected[i][3] = expected_range_primary[i]
+        secondary_expected[i][3] = expected_range_secondary[i]
+
+    (primary_with_range, secondary_with_range) = MagL1a.process_compressed_vectors(
```
alastairtree, line 506: I think this test should be broken out into 2 separate tests, one for with range data and one without. The one without should check the generated range data all matches the first vectors range.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Yeah, sorry, this sentence is kind of weird. The compression width is the first 6 bits of the header, which corresponds to a bit count that indicates the width of uncompressed vectors. 

So, if that compression width was `01 0000`, then we have uncompressed vectors values of a width of 16 bits. The full width of the first vector is 16 * 3 + 2 bits, because there are three values per vector and 2 bits for the range value.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Yes, the range data is the last piece of information about the vector. In the compressed case, we assume that the range is the same for all vectors unless there is a range data section, in which case we read the range data section at the end to update the ranges.

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+    assert secondary.shape[1] == 4
+
+    assert np.array_equal(primary, primary_expected)
+    assert np.array_equal(secondary, secondary_expected)
+
+    # range data has 2 bits per vector, primary and then secondary in sequence.
+    # It excludes the first vector, making the length (primary_count - 1) * 2
+    range_primary = "000000000000000101101011111111"
+
+    expected_range_primary = [3, 0, 0, 0, 0, 0, 0, 0, 1, 1, 2, 2, 3, 3, 3, 3]
```
alastairtree, line 480:
```suggestion
    first_primary_vector_range = 3  # This is not included in the range data section
    expected_range_primary = [first_primary_vector_range, 0, 0, 0, 0, 0, 0, 0, 1, 1, 2, 2, 3, 3, 3, 3]
```
Just for readability

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+
+    # 4 uncompressed vectors from uncompressed_vector_bytearray
+    uncompressed_bits = (
+        "000000100000010000001000000100000001000000100000"
+        "000000100000011100001000000111010001000000111010"
+        "000000100000101000001000001010100001000001010100"
+        "000000100000110100001000001101110001000001101111"
+    )
+
+    # TODO fix test
```
alastairtree: Still To do?

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+
+    input_data = np.packbits(input_data)
+    (primary, secondary) = MagL1a.process_compressed_vectors(input_data, 20, 20)
+
+    # The 16th compressed vector is bad because it needs to be >60 bits
+    assert np.array_equal(primary[:15], expected_vectors[0][:-1])
+    assert np.array_equal(primary[16:], uncompressed_expected_vectors)
+
+    assert np.array_equal(secondary[:15], expected_vectors[1][:-1])
+    assert np.array_equal(secondary[16:], uncompressed_expected_vectors)
```
alastairtree, line 574: Both primary and secondary have the same expected vectors for 16th onwards?

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+            + secondary_compressed
+        ],
+        dtype=np.uint8,
+    )
+    input_data = np.packbits(input_data)
+
+    (primary, secondary) = MagL1a.process_compressed_vectors(input_data, 6, 16)
+    assert len(primary) == 6
+    assert np.array_equal(primary[0], expected_vectors[0][0])
+    assert np.array_equal(primary[2:], uncompressed_expected_vectors)
```
alastairtree, line 599: This seems like a second test and I can't work out what it is testing? What does "Test if first primary vector is too long" mean?

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+
+
+def test_different_compression_width():
+    # Compression headers - indicating a 12 bit width and no range section
+    headers = "00110000"
```
alastairtree, line 604: Would be tempted to have a method to build this to make it readable and reduce duplication:
```python
headers=generate_mag_science_header(compression=False, bit_width=12)
```

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+
+
+def test_accumulate_vectors():
+    start_vector = np.array([1, 2, 3, 4], dtype=np.uint)
+
+    diff_vectors = [1, 1, 1, 3, 0, -3, -1, -10, 1]
+
+    expected_vectors = np.array(
+        [[1, 2, 3, 4], [2, 3, 4, 4], [5, 3, 1, 4], [4, -7, 2, 4]]
+    )
```
alastairtree, line 662:
```python
   range=3 
   start_vector = np.array([1, 2, 3, range], dtype=np.uint)

    diff_vectors = [
        1, 1, 1, 
        3, 0, -3, 
       -1, -10, 1]

    expected_vectors = np.array(
        [[1, 2, 3, range], [2, 3, 4, range], [5, 3, 1, range], [4, -7, 2, range]]
    )
```
Something like this makes it easier to read/review?

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+
+    test_vectors = MagL1a.accumulate_vectors(start_vector, diff_vectors, 4)
```
alastairtree: Would be tempted to rename this something like `convert_vectors_from_diffs_to_eng_units()` but naming is hard and i have a preference for long_but_very_readable_method_names :-)

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -47,7 +305,364 @@ def test_compare_validation_data():
...
+
+    assert np.array_equal(test_vectors, expected_vectors)
+
+
+def test_unpack_one_vector(uncompressed_vector_bytearray, expected_vectors):
```
alastairtree: Might be better as a parameterised test rather than multiple tests in one. Our vectors will always be between 12 and 20 bits when compressed, and always 16bit when uncompressed

imap_processing/tests/mag/test_mag_l1a.py
```diff
@@ -61,7 +676,7 @@ def test_process_vector_data():
 
     # 36 bytes
     (primary_vectors, secondary_vectors) = MagL1a.process_vector_data(
-        input_data, total_primary_vectors, total_secondary_vectors
+        input_data, total_primary_vectors, total_secondary_vectors, 0
```
alastairtree, line 738: Maybe use a constant or enum here
```suggestion
        input_data, total_primary_vectors, total_secondary_vectors, UNCOMPRESSED
```

imap_processing/tests/mag/test_mag_l1a.py
alastairtree, line 1: These tests look good thanks, have added a few inline style things. 
Missing cases I can see it might be worth covering with tests:
- When there is no range section and so all vectors in the same range based on the first vector
- having different rates in primary and secondary, like in Burst(64,8) mode say
- different cases for the padding between sections, like when the end of the primaries falls on the byte boundary and so there is no padding

imap_processing/mag/l1a/mag_l1a_data.py
alastairtree: Each sensors are in one of 4 ranges 0,1,2,3 at any given moment, based on the size of the field. A tiny field would be range 3, and then when a big flare passes imap it might switch into range 2 for a while, and then go back to range 3. On earth is range 0. 
99% of the time it does not change range so you can assume every vector is in the same range as the first vector. When it switches range you need to know that say the first 5 vectors were in range 3, and vectors 6-128 are in range 2 so we include the range array only when you need that data as an extra way to save bandwidth.

imap_processing/mag/l1a/mag_l1a_data.py
alastairtree: Say header compression_width=16, then you know the first vector is length (x=16bits,y=16bit,z=16bits,rng=2bits)=50bits
After that the rest is compressed with varying bit width, but will decompress into 16bit x/y/z values.

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+            )
+            secondary_vectors = MagL1a.process_range_data_section(
+                bit_array[
+                    end_vector + (primary_count - 1) * 2 : end_vector
+                    + (primary_count + secondary_count - 2) * 2
+                ],
+                secondary_vectors,
+            )
+
+        return primary_vectors, secondary_vectors
```
alastairtree, line 720: I would be tempted to put in some defensive assertions here? E.g. that you have the desired number of pri/sec vectors, and that each has a valid range 0-3.

imap_processing/mag/l1a/mag_l1a_data.py
alastairtree: Yeah it is hard to get it more split up, but still probably a worth endeavour if you can as it is very long.

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+                    secondary_boundaries += [seq_val+1]
+                    vector_count += 1
+
+            # If we're greater than primary_count, we are in the secondary vectors.
+            # Like before, we skip indices with 3 ones.
+            if (
+                vector_count > primary_count
+                and seq_val - secondary_boundaries[-1] > 0
+            ):
+                secondary_boundaries.append(seq_val + 1)
```
alastairtree, line 652: The logic for primary and secondary processing of boundaries is duplicated and a good candidate for refactoring out to a method because the primary and secondary sections are processed/packed the same way just with different start/end locations in the bit stream

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+        # For example: [0 0 1 0 1 1] + [1 0 0 1 0 1] = [1 0 1 1 1 2], so the last index
+        # has 2 ones in a row.
+        # The first bit is invalid, so we remove it at the end.
+        sequential_ones = np.where(
+            np.add(
+                vector_bits,
+                np.roll(vector_bits, 1),
+            )[1:]
+            == 2
+        )[0]
```
alastairtree, line 579: This is a much fancier method to split the bits into individual vectors all in one go, nice! I just iterated the array and chopped after sequential 1s!

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+                # 3 boundaries equal one vector
+                if len(primary_boundaries) % 3 == 0:
+                    vector_count += 1
+                    # If the vector length is >60 bits, we switch to uncompressed.
+                    # So we skip past all the remaining seq_ones.
+                    if (
+                        (len(primary_boundaries) > 4)
+                        and (primary_boundaries[-1] - primary_boundaries[-4] > 60)
+                        or (vector_count == 2 and primary_boundaries[-1] > 60)
+                    ):
```
alastairtree, line 620: The 60 here should probably be refactored out to a constant EXIT_COMPRESSION_MAX_BITS_THRESHOLD or something.

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+
+                # 3 boundaries equal one vector
+                if len(primary_boundaries) % 3 == 0:
+                    vector_count += 1
+                    # If the vector length is >60 bits, we switch to uncompressed.
+                    # So we skip past all the remaining seq_ones.
+                    if (
+                        (len(primary_boundaries) > 4)
+                        and (primary_boundaries[-1] - primary_boundaries[-4] > 60)
+                        or (vector_count == 2 and primary_boundaries[-1] > 60)
```
alastairtree:
```
                       (len(primary_boundaries) > 4)
                        and (primary_boundaries[-1] - primary_boundaries[-4] > 60)
                        or (vector_count == 2 and primary_boundaries[-1] > 60)
```
I have read this a few times and cannot quite work out why len > 4 or count == 2 for this so i think it needs a refactor or explanation

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+        vector_count = 1
+        end_primary_vector = 0
+        for seq_val in sequential_ones:
+            if vector_count > primary_count + secondary_count:
+                break
+            # Add the end indices of each primary vector to primary_boundaries
+            # If we have 3 ones in a row, we should skip that index
+            if vector_count < primary_count and (
+                seq_val - primary_boundaries[-1] > 0
+            ):
```
alastairtree: huh?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+                primary_boundaries.append(seq_val + 1)
+
+                # 3 boundaries equal one vector
+                if len(primary_boundaries) % 3 == 0:
```
alastairtree:
```suggestion
                if len(primary_boundaries) % NO_OF_AXIS == 0:
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+        # The first bit is only needed for the np.roll step, so now we remove it.
+        # we are left with compressed primary vectors, and all the secondary vectors.
+        vector_bits = vector_bits[1:]
+        # print(f"Vector bits: {vector_bits}")
+        # which indices within vector_bits are the end of a vector
+        primary_boundaries = [sequential_ones[0] + 1]
+        secondary_boundaries = []
+        vector_count = 1
+        end_primary_vector = 0
+        for seq_val in sequential_ones:
```
alastairtree, line 596: It might be easier to read/follow this if you iterated in jumps of three here, and then process the x/y/z values as a whole vector?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+            uncompressed_vectors = vector_bits[primary_boundaries[-1] : primary_end]
+            for i in range(0, len(uncompressed_vectors), uncompressed_vector_size):
+                decoded_vector = MagL1a.unpack_one_vector(
+                    uncompressed_vectors[i : i + uncompressed_vector_size],
+                    compression_width,
+                    False,
+                )
+                primary_vectors[vector_index] = decoded_vector
+                primary_vectors[vector_index][3] = first_vector[3]
+                vector_index += 1
```
alastairtree: Sorry I cant really follow this and don't understand why any vectors would ever be "missing"?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+                bit_array[end_vector : end_vector + (primary_count - 1) * 2],
+                primary_vectors,
+            )
+            secondary_vectors = MagL1a.process_range_data_section(
+                bit_array[
+                    end_vector + (primary_count - 1) * 2 : end_vector
+                    + (primary_count + secondary_count - 2) * 2
+                ],
+                secondary_vectors,
+            )
```
alastairtree, line 719: Not really clear from this that if `has_range_data_section=0` then all vectors have the same range as their sensors first vector. Refactor the processing of the range section out to a method and make that logic explicit?

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+        """
+        Given a range data section and vectors, return an updated vector array.
+
+        Each range value has 2 bits. range_data will have a length of n*2, where n is
+        the number of vectors in vectors.
+
+        Parameters
+        ----------
+        range_data : numpy.ndarray
+            Array of range values, where each value is one bit.
```
alastairtree:
```suggestion
            Array of range data section in binary. Each range is 2 bits long.
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+            raise ValueError(
+                "Incorrect length for range_data, there should be two bits per vector, "
+                "excluding the first."
+            )
+
+        updated_vectors: np.ndarray = np.copy(vectors)
+        range_str = "".join([str(i) for i in range_data])
+        for i in range(len(vectors) - 1):
+            range_int = int(range_str[i * 2 : i * 2 + 2], 2)
+            updated_vectors[i + 1][3] = range_int
```
alastairtree, line 823: It bugged me how this went from a array of bits to a string to ints. Best i could do to simplify was like this into array of 2bit range values. Will be faster but not sure if it is any better?

```python
arr = np.array([0,0,0,1,1,0,1,1]) #[0,1,2,3] as 2 bit numbers in binary array
result = [(x<<1)+y for x,y in np.array_split(arr,len(arr)/2)]
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+            A numpy array of shape (expected_vector_count, 4) of signed integers,
+            representing the calculated vectors.
+        """
+        vectors: np.ndarray = np.empty((vector_count, 4), dtype=np.int32)
+        vectors[0] = first_vector
+        if len(vector_differences) % 3 != 0:
+            logger.error(f"Error! Computed compressed vector differences are not "
+                         f"divisible by 3 - meaning some data is missing. "
+                         "Expected length: %s, actual length: "
+                         "%s", vector_count * 3, len(vector_differences))
```
alastairtree: If this happened we would probably say abandon the packet as something very bad has happened

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+        index = 0
+        vector_index = 1
+        for diff in vector_differences:
```
alastairtree, line 874: Another readability thing but how about in chunks of 3 something like this?
```suggestion
        for vector in [(x,y,z) for x,y,z in np.array(vector_differences).reshape(-1,3)] :
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+                "unpack_one_vector method is expecting an array of bits as" "input."
+            )
+
+        if len(vector_data) != width * 3 + 2 * has_range:
+            logger.error(
+                f"Invalid length {len(vector_data)} for vector data. Expected "
+                f"{width * 3} or {width * 3 + 2} if has_range."
+            )
+            return np.zeros(4, dtype=np.int32)
+            # TODO: should this raise an error? Update flags?
```
alastairtree: Should fail the whole packet if this errors

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -429,6 +479,478 @@ def to_signed16(n: int) -> int:
...
+
+        # take slices of the input data and pack from an array of bits to an array of
+        # uint8 bytes
+        x = np.packbits(np.concatenate((padding, vector_data[:width])))
+        y = np.packbits(np.concatenate((padding, vector_data[width : 2 * width])))
+        z = np.packbits(np.concatenate((padding, vector_data[2 * width : 3 * width])))
+
+        range_string = "".join([str(i) for i in vector_data[-2:]])
+
+        rng = int(range_string, 2) if has_range else 0
```
alastairtree, line 933: This range logic is duplicated above and could be a method

alastairtree (commented):
Epic effort maxine, thank you!

I have left you lots of comments (sorry!). Lots are just style and readability stuff - i know how hard it is to make this compression logic comprehendable so I feel your pain.

Think my only other query is on error handling. I think it should be a bit more defensive and if you get any bad unexpected values it means something awful has happened and the whole packet should be considered a stream of NaNs and an error raised somewhere so we can investigate..

Thank you!

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: I emailed you about this - this was originally an error, but I changed it to a log because the test packets you sent me had several packets where this error cropped up. 

In the future, would you want this to generate a CDF file that is empty, or skip generating the CDF file altogether? Skip that packet only, but include the other packets if they're ok? Pass in the data we have but mark a quality flag? Something else....?

Thanks for your review!

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: This is the index of the end of the first vector, so it's uncompressed_vector size long, but we need to skip past the first 8 bits - so the end index is uncompressed_size plus 8. I can move the `+8` into the slice in line 546, if that makes more sense?

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Good catch! Fixed this

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: I dug deep into my days of hackerrank to get this weird algorithm 😅 in the end I'm not totally convinced it's better than your design.... but it is quite fast, which is good considering we will probably need to process this code a couple of times for the perfect 24 hour files.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: I'll give it a shot and see if that helps make things more readable!

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: with the np.roll method, the indices will be duplicated for 3 ones in a row. This is allowed, but we need to skip the indices for the second pair of ones. 

eg: [0, 1, 1, 1, 0, 0, 1, 1] is valid and should be split into [0, 1, 1] and [1, 0, 0, 1, 1], along the indices 1 and 6. 

However, there are also 2 sequential ones at index 3. So, the output of np.roll is [0, 1, 1, 1, 0, 0, 1, 1] + [1, 1, 1, 0, 0, 1, 1, 0] = [1, 2, 2, 1, 0, 1, 2, 1] with indices [1, 2, 6]. So we need to drop that 2.

I'll update the comment to be more clear. This is the main reason why I think this code is... not ideal.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: This line caused me much trouble. The beginning and ends of the sequence are kind of weird, since each index indicates the end of the value. This means we need to compare the most recent value to 3 indices ago - except for the first vector, which does not have  3 prior indices, only 2. So in that case we just look at the index of the last value to see if it's > 60. 

I will update the comment to explain this a bit more, thanks for calling it out.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Yeah, bad phrasing. They aren't missing from the data, but rather missing from the compressed vectors. If we finish with the compressed vectors, and we don't have the number that we expect, then this code assumes that they are missing because they are in uncompressed form. So, it's more like they're missing from what has been processed so far. I will try to rename this and expand my comments.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Agreed, and I've had to update two places a few times now, so I'll pull this out into another method.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: So, do you mean that I should always call `process_range_data_section` and pass `has_range_data_section` through, or should I just be more explicit within this if statement?

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Sure, I can add those checks

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: This was a slow bit in my profiling. But that change is less intuitive.... I'll contemplate it, thanks for the suggestion!

imap_processing/tests/mag/test_mag_l1a.py
maxinelasp: Your first case is covered under the first part of `test_compressed_vector_data`, but your other tests are good and I will definitely add them, thanks!

maxinelasp:
@alastairtree @tech3371 @sdhoyt @laspsandoval Thank you so much for your reviews! I have addressed all the comments on this PR, added some additional tests, and had the output validated by the MAG team. Could I get a re-review and maybe an approval from you?

sdhoyt (approved):
It looks good to me. Nice work!

imap_processing/mag/l1a/mag_l1a_data.py
alastairtree: I think maybe calling a method called `unpack_vector_ranges` that either copies it from the first vector, or reads the end section would be more explicit/easier to understand. ATM there are 2 really separate bits of code, (line 770 `vectors[i + compressed_count][3] = vectors[0][3]`) and here that generate the range for each vector, depending on the value of `has_range_data_section` and it is non-obvious that they are connected

alastairtree: Also seems like the logic is on line 865 as well `vectors[vector_index][3] = vectors[0][3]`

alastairtree:
There are a couple of comments of mine that you "thumbs up'd" / replied to but didn't actually change, which is fine but might be worth checking you have pushed all your changes if not. Otherwise looks good!

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Ok yeah, that makes sense. I will rework that, I think it will be more clear. Thanks!

maxinelasp:
@alastairtree Thanks for that final review! I will double check that I addressed everything, there were a few comments that I preferred to leave as-is, but I definitely want to make sure I decided on everything. If there are any comments that you definitely want to see changed, please let me know.

imap_processing/tests/mag/test_mag_l1a.py
maxinelasp: Yes, I just used the same bits for primary and secondary

imap_processing/tests/mag/test_mag_l1a.py
maxinelasp: This is checking some weird edge cases in my code. The first part of the test is checking to see if a vector partway through is > 60 bits, this part checks if the first vector is >60 bits.

imap_processing/tests/mag/test_mag_l1a.py
maxinelasp: I don't love this variable name, but your suggestion is confusing to me, because that is not how I think about this method. But, I will rename it

maxinelasp:
@alastairtree Thanks for pointing out those comments, I guess I resolved a bunch of them in some kind of fugue state on Friday afternoon and didn't actually fix the things you were pointing out! I went back in and I'm addressing your comments, including a bunch on the tests that I missed. Sorry about that, I got a little over eager to finish this up 😅

PR #803, opened 2024-08-30
MNT/TST: Update SWAPI packet definition and tests

greglucas:
This removes some unnecessary utils functions and moves over to `packet_file_to_datasets()`

Note that updating the packet definition also changed the `plan_id_science` back to `plan_id` which was in the original spreadsheet definition. This is because they can now have different bit lengths between the packets.

The packet definition also does not have the `AUT` packet defined in it anymore, do we want to add that one to the definition, or leave it out? I left it out for now and just removed that test.

### DIFF
--- a/imap_processing/tests/swapi/test_swapi_decom.py
+++ b/imap_processing/tests/swapi/test_swapi_decom.py
@@ -10,8 +10,8 @@ from imap_processing.utils import packet_file_to_datasets
 
 @pytest.fixture(scope="session")
 def decom_test_data(swapi_l0_test_data_path):
-    """Read test data from file as raw values"""
-    test_file = "imap_swapi_l0_raw_20231012_v001.pkts"
+    """Read test data from file with derived values"""
+    test_file = "imap_swapi_l0_raw_20240924_v001.pkts"
     packet_file = imap_module_directory / swapi_l0_test_data_path / test_file
     packet_definition = (
         f"{imap_module_directory}/swapi/packet_definitions/swapi_packet_definition.xml"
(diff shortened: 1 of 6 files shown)


imap_processing/tests/swapi/test_swapi_decom.py
```diff
@@ -2,77 +2,81 @@
...
-from imap_processing.decom import decom_packets
 from imap_processing.swapi.l1.swapi_l1 import (
     SWAPIAPID,
 )
-from imap_processing.utils import group_by_apid
+from imap_processing.utils import packet_file_to_datasets
+
+
+@pytest.fixture(scope="session")
+def decom_test_science_data(swapi_l0_test_data_path):
```
greglucas: It is interesting that they gave you "derived_values" for the science data, but didn't use any "derived_values" for the housekeeping. That is why we have these two different fixtures here now.

greglucas:
The failing tests are due to an issue upstream with space_packet_parser that has already been fixed, but not released yet. https://github.com/medley56/space_packet_parser/pull/54

I'm not sure there is much we can do about this until the fix is out.

imap_processing/tests/swapi/test_swapi_decom.py
tech3371: yes. That's why I had to add if and else in test but your approach is much better.

tech3371 (approved):
Looks great! Thank you again!

greglucas:
Converting to draft since this requires changes to Space Packet Parser that will be released in the next version. Need to revisit this then when we upgrade that dependency version.

PR #804, opened 2024-08-30
Implement packet_file_to_datasets function for CoDICE

### LINKED ISSUE #771: CoDICE: Implement new packet_file_to_datasets function in L1 pipeline
The new `utils.packet_file_to_dataset()` function can be implemented in the CoDICE L1 pipeline and help simplify the code.

bourque:
# Change Summary

## Overview
This PR implements the new `utils.packet_file_to_datasets()` function in the CoDICE pipelines.

## Updated Files

  - Moved XTCE file mapping to `constants.py` since it is used in multiple modules; implemented `packet_file_to_datasets` function
  - Implemented `packet_file_to_datasets` function and made a few other cleanup changes
-  `imap_processing/codice/constants.py`
  - Moved XTCE file mapping here since it is used in multiple modules now
- `imap_processing/tests/codice/test_codice_l0.py` and `imap_processing/tests/codice/test_codice_l1a.py`  
  - Updated tests to reflect new changes 

Closes #771


imap_processing/codice/codice_l1a.py
```diff
@@ -179,12 +175,22 @@ def create_science_dataset(self, met: np.int64, data_version: str) -> xr.Dataset
             # Data arrays are structured depending on the instrument
             if self.instrument == "lo":
                 variable_data_arr = np.array(variable_data).reshape(
-                    1, self.num_positions, self.num_spin_sectors, self.num_energy_steps
+                    (
+                        1,
+                        self.num_positions,
+                        self.num_spin_sectors,
+                        self.num_energy_steps,
+                    )
```
bourque, line 184: This change is to avoid 'unexpected argument' warnings from `np.reshape()`

imap_processing/codice/codice_l1a.py
```diff
@@ -309,49 +315,49 @@ def unpack_science_data(self, science_values: str) -> None:
         science_values : str
             A string of binary data representing the science values of the data.
         """
-        self.compression_algorithm = constants.LO_COMPRESSION_ID_LOOKUP[self.view_id]
+        compression_algorithm = constants.LO_COMPRESSION_ID_LOOKUP[self.view_id]
```
bourque, line 319: When I originally wrote this, I thought that I might be using the `compression_algorithm` in various places in the module, but turns out I only need it here, so no need to make it a class attribute.

imap_processing/codice/codice_l1a.py
```diff
@@ -309,49 +315,49 @@ def unpack_science_data(self, science_values: str) -> None:
...
-    met : list[int]
-        The Mission Elapsed Time of the data.
-    event_data : str
-        A string of binary numbers representing the event data.
-    dataset_name : str
-        The name for the dataset.
+    apid : int
+        The APID of the packet.
+    packet : xarray.Dataset
+        The packet to process.
```
greglucas, line 361: full_dataset ?

I think this is the entire packet file you are passing in and not an individual packet.

imap_processing/codice/codice_l1a.py
```diff
@@ -360,14 +366,25 @@ def create_event_dataset(
...
+
     cdf_attrs = ImapCdfAttributes()
     cdf_attrs.add_instrument_global_attrs("codice")
     cdf_attrs.add_instrument_variable_attrs("codice", "l1a")
     cdf_attrs.add_global_attribute("Data_version", data_version)
 
     # Define coordinates
     epoch = xr.DataArray(
-        met_to_j2000ns(met),  # TODO: Fix after SIT-3 (see note below)
+        met_to_j2000ns([met]),
```
greglucas: This is already calculated in the dataset routine. So `epoch` in the incoming dataset should be what you want here and just need to add on the attributes like you're doing.

```suggestion
        packet["epoch"],
```

imap_processing/codice/codice_l1a.py
```diff
@@ -408,14 +425,9 @@ def create_hskp_dataset(
...
 
-    metadata_arrays: dict = collections.defaultdict(list)
-
-    for packet in packets:
-        add_metadata_to_array(packet, metadata_arrays)
-
     epoch = xr.DataArray(
         met_to_j2000ns(
-            metadata_arrays["SHCOARSE"],
+            packet.shcoarse.data,
```
greglucas: This should already be taken care of in the "epoch" variable. (I wonder if we will run into issues with you here and the different reference_epoch...

imap_processing/codice/codice_l1a.py
```diff
@@ -438,19 +450,21 @@ def create_hskp_dataset(
     #       I am holding off making this change until I acquire updated
     #       housekeeping packets/validation data that match the latest telemetry
     #       definitions
-    for key, value in metadata_arrays.items():
+    for variable in packet:
```
greglucas, line 448: Thanks for changing from metadata to variable!

imap_processing/codice/codice_l1a.py
```diff
@@ -479,10 +493,10 @@ def get_params(packet: space_packet_parser.parser.Packet) -> tuple[int, int, int
     view_id : int
         Provides information about how data was collapsed and/or compressed.
     """
-    table_id = packet.data["TABLE_ID"].raw_value
-    plan_id = packet.data["PLAN_ID"].raw_value
-    plan_step = packet.data["PLAN_STEP"].raw_value
-    view_id = packet.data["VIEW_ID"].raw_value
+    table_id = packet.table_id.data[0]
```
greglucas: Are you able to remove any of the `[0]` indexes now and keep the entire array as you go through? I think we'd want each epoch associated with its own table_id and not just the first one.

imap_processing/codice/codice_l1a.py
```diff
@@ -504,48 +518,30 @@ def process_codice_l1a(file_path: Path, data_version: str) -> xr.Dataset:
         The ``xarray`` dataset containing the science data and supporting metadata.
     """
     # Decom the packets, group data by APID, and sort by time
-    packets = decom_packets(file_path)
-    grouped_data = group_by_apid(packets)
+    xtce_packet_definition = Path(
+        f"{imap_module_directory}/codice/packet_definitions/{constants.PACKET_TO_XTCE_MAPPING[file_path.name]}"
+    )
```
greglucas, line 518: This can be saved for a follow-up PR, but you should be able to get to a single XML definition file now as well.

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -35,7 +35,7 @@
     (1, 128),  # lo-pha
 ]
 EXPECTED_ARRAY_SIZES = [
-    123,  # hskp
+    129,  # hskp
```
greglucas: Interesting, you got a few more packets now?

greglucas (approved):
Overall, I think this looks great! A few minor comments is all.

imap_processing/codice/codice_l1a.py
bourque: Thanks. I have a few TODOs regarding updates to XTCE, I will tackle this then.

imap_processing/tests/codice/test_codice_l1a.py
bourque: Your comment here made me realize that the test for this is poorly named, because these values are actually the expected number of variables in the CDF (i.e. `len(dataset)`), not the data array sizes. I will update this in a new commit.

imap_processing/codice/codice_l1a.py
bourque: These four parameters are grabbed from a single APID/packet, so I think these will always be a single value.  However, I can avoid the `[0]` by using `int(packet.table_id.data)`

imap_processing/codice/codice_l1a.py
bourque: Thanks, good catch. So if I am understanding correctly, this can be changed from:

```python
    epoch = xr.DataArray(
        met_to_j2000ns(
            packet.shcoarse.data,
            reference_epoch=np.datetime64("2010-01-01T00:01:06.184", "ns"),
        ),
        name="epoch",
        dims=["epoch"],
        attrs=cdf_attrs.get_variable_attributes("epoch"),
    )
```

to:

```python
    epoch = xr.DataArray(
        packet.epoch,
        name="epoch",
        dims=["epoch"],
        attrs=cdf_attrs.get_variable_attributes("epoch"),
    )
```

imap_processing/codice/codice_l1a.py
bourque: No I think this is just an individual packet? Here is how it is called:

```python
    packets = packet_file_to_datasets(file_path, xtce_packet_definition)

    for apid in packets:
        packet = packets[apid]
        logger.info(f"\nProcessing {CODICEAPID(apid).name} packet")

        if apid == CODICEAPID.COD_NHK:
            dataset = create_hskp_dataset(packet, data_version)

        elif apid in [CODICEAPID.COD_LO_PHA, CODICEAPID.COD_HI_PHA]:
            dataset = create_event_dataset(apid, packet, data_version)
```

PR #807, opened 2024-09-04
Added new tof test data

laspsandoval:
# Change Summary

## Overview
Added test data for tof that I recently received + another function.

## Deleted Files
- FM45_40P_Phi28p5_BeamCal_LinearScan_phi28.50_theta-0.00_Ultra_Image_Raw_Event_20240207T102746_withFSWcalcs.csv
   - Replaced by new test data.

## Updated Files
- ultra_l1b_extended.py
   - Added calculations for coincidence position and "etof"
- conftest.py
   - Filename update
- lookup_utils.py
   - Docstring update
- FM45_Startup1_ULTRA_IMGPARAMS_20240719.csv
   - Parameter update

## Testing
- imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
   - Test ultra_l1b_extended.py
- FM45_40P_Phi28p5_BeamCal_LinearScan_phi28.50_theta-0.00_Ultra_Image_Raw_Event_20240207T102746_withFSWcalcs.csv
   - New test data that includes tof and d values.


imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -315,3 +319,96 @@ def get_ssd_back_position_and_tof_offset(
...
+        coin_s_norm_bottom - coin_n_norm_bottom
+    ) + get_image_params("XCOINBTOFF")  # millimeter
+
+    # Time for the electrons to travel back to coincidence anode.
+    t2[index_bottom] = get_image_params("ETOFSC") * t1[index_bottom] + get_image_params(
+        "ETOFBTOFF"
+    )
+
+    # Multiply by 10 to convert to tenths of a nanosecond.
+    etof[index_bottom] = t2[index_bottom] * 10 - particle_tof[index_bottom]
```
tmplummer: This code section is essentially a duplicate of lines 377-392. Consider extracting into a function.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -315,3 +319,96 @@ def get_ssd_back_position_and_tof_offset(
...
+    Returns
+    -------
+    etof : np.ndarray
+        Time for the electrons to travel back to
+        coincidence anode (tenths of a nanosecond).
+    xc : np.ndarray
+        X coincidence position (hundredths of a millimeter).
+    """
+    index_top = np.nonzero(np.isin(de_dataset["COIN_TYPE"], CoinType.Top.value))[0]
+    de_top = de_dataset.isel(epoch=index_top)
```
tmplummer, line 404: I don't know xarray, but this seems like an unintuitive way to get a filtered dataset. I get the first line, but don't understand why `de_dataset.isel(epoch=index_top)` is used. I would think that the indices can be used directly?

Might be worth looking at `xarray.Dataset.filter_by_attrs()`: https://docs.xarray.dev/en/stable/generated/xarray.Dataset.filter_by_attrs.html

tmplummer (commented):
Overall, looks good. I would recommend trying to reduce the duplicated code in the new `get_coincidence_positions()` function.

imap_processing/ultra/l1b/ultra_l1b_extended.py
greglucas: I think the incantation might be:
```python
de_dataset.sel(COIN_TYPE=CoinType.Top.value)
```

`sel` is used for selecting against values, and `isel` for the indices. You have the values above, so you can just do the comparison within `sel`.

laspsandoval: Thank you for the suggestion! I understand that sel() is generally preferred for selecting data based on values, and I’ve considered using it here. However, since COIN_TYPE is a data variable and not a coordinate, I can't directly use sel() nor filter_by_attrs since it isn't an attribute either. 

 I’m opting to use np.nonzero() for indexing..

tmplummer (approved):
Thanks for the refactor. This looks good to me. I was also forced to learn that `xarray.dataset.isel()` is for "index select" instead of my ignorant assumption that it stood for "is element" 🤦

PR #811, opened 2024-09-04
Create Ultra pset energy bins

laspsandoval:
# Change Summary

## Overview
Simple code to create energy bins for pset.

## New Files
- ultra_l1c_pset_bins.py
   - Create energy bins for pset

## Testing
test_ultra_l1c_pset_bins.py

### DIFF
--- a/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
+++ b/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
@@ -1,0 +1,35 @@
+"""Creates bins for pointing sets"""
+
+import numpy as np
+
+
+def build_energy_bins():
+    """
+    Build energy bin boundaries and midpoints for the
+    resolution (alpha) and lower and upper limits on energy (energy_bounds).
+
+    Returns
+    -------
+    energy_bin_start : np.ndarray
+        Array of energy bin start values.
+    energy_bin_end : np.ndarray
+        Array of energy bin end values.
+    """
+    alpha = 0.05  # deltaE/E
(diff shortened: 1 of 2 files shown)


imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -0,0 +1,34 @@
...
+    # Create bins.
+    energy_bin_start = np.array([energy_bounds[0]])
+    while energy_bin_start[-1] * energy_step <= energy_bounds[-1]:
+        energy_bin_start = np.append(
+            energy_bin_start, energy_bin_start[-1] * energy_step
+        )
+
+    # Compute the end values for the bins
+    energy_bin_end = energy_bin_start[1:] * energy_step
+    energy_bin_end = np.insert(energy_bin_end, 0, energy_bounds[0] * energy_step)
```
tmplummer: I don't see why this can't just be `energy_bin_end = energy_bin_start * energy_step`?

tmplummer (approved):
I was trying to figure out if there was a clever way to generate the energy bins, but didn't come up with anything better than your while loop. Looks good to me with one nit-pick.

greglucas (commented):
I have a few quick suggestions to think about. Do you know the number of bins and the start? I'd start with that rather than the while < value version.

I think this one-liner might give you what you are after:
```python
3.5 * energy_step ** np.arange(91)
```

That will also return all bin edges. I'd return that single array of the edges rather than a tuple, and then use it within the code as `bin_start = bin_edges[:-1]`, `bin_end = bin_edges[1:]`.

PR #812, opened 2024-09-04
Updated CoDICE XTCE files from imap_xtce tool

### LINKED ISSUE #768: CoDICE: Add support for variable length data in XTCE definition files
Currently the CoDICE XTCE definition files have hard-coded data bit lengths that work with the current set of simulated/test data. However, in flight, we will be receiving variable length data packets, so the XTCE definition files need to be updated to support that.  This may be an easy fix with the recent updates to the XTCE scripts?

bourque:
# Change Summary

## Overview
This PR updates the CoDICE XTCE files from running the new `imap_xtce` command line tool. As such, there is now a single telemetry definition file for all packet types. However, I need to retain the old `P_COD_NHK.xml` file until I can sort out a discrepancy with one of the fields that is causing an error (marked as a `TODO` in the code).

## New Files

- `codice_packet_definition.xml`
  - The new, combined CoDICE packet definition, output from the `imap_xtce` tool 

## Deleted Files

- A whole bunch of individual packet definition files for each packet type

## Updated Files

  - Updated code to use new packet definition file
  - Removed redundant code that decoms packets, using `codice_lo.decom_packets()` instead  
  - Removed the hard-coded XTCE file mapping, no longer needed

Closes #768


imap_processing/codice/codice_l0.py
```diff
@@ -37,9 +36,20 @@ def decom_packets(packet_file: Path) -> dict[int, xr.Dataset]:
     packets : dict[int, xarray.Dataset]
         Mapping from apid to ``xarray`` dataset, one dataset per apid.
     """
+    # TODO: Currently need to use the 'old' packet definition for housekeeping
+    #       because the test housekeeping packet has LAST_OPCODE=0, and the
+    #       telemetry definition "STATES" tab doesn't allow for a 0 value for
+    #       this. Need to figure out if the test packet is bad, or the telemetry
+    #       definition is wrong.
```
greglucas: This is an interesting case. Does an Enumeration need to exist for every possible value or should `space_packet_parser` just return the raw_value here and not decode anything rather than raising an error? Gavin might have some insight here as well.

greglucas (approved):
Looks good. I like the combination of these into a single file.

imap_processing/codice/codice_l0.py
```diff
@@ -37,9 +36,20 @@ def decom_packets(packet_file: Path) -> dict[int, xr.Dataset]:
...
     xtce_packet_definition = Path(
-        f"{imap_module_directory}/codice/packet_definitions/{constants.PACKET_TO_XTCE_MAPPING[packet_file.name]}"
+        f"{imap_module_directory}/codice/packet_definitions/{xtce_filename}"
+    )
+    packets: dict[int, xr.Dataset] = packet_file_to_datasets(
+        packet_file, xtce_packet_definition
     )
-    packets = packet_file_to_datasets(packet_file, xtce_packet_definition)
 
     return packets
```
tech3371: can you return the dataset now without assigning since it's not packets anymore? or this file itself might go away like mine did for SWE and SWAPI once your HK issue is fixed.

imap_processing/codice/packet_definitions/codice_packet_definition.xml
```diff
@@ -0,0 +1,5072 @@
...
+			<xtce:IntegerParameterType name="COD_NHK.CMDEXE" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="8" encoding="unsigned" />
+			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="COD_NHK.CMDRJCT" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="8" encoding="unsigned" />
+			</xtce:IntegerParameterType>
+			<xtce:EnumeratedParameterType name="COD_NHK.LAST_OPCODE" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="16" encoding="unsigned" />
+				<xtce:EnumerationList>
+					<xtce:Enumeration value="4096" label="COD_NOOP" />
```
tech3371, line 49: Can you add this to get around your above issue temporarily or will this cause issue later on?
```suggestion
                                          <xtce:Enumeration value="0" label="TESTING" />
					<xtce:Enumeration value="4096" label="COD_NOOP" />
```

tech3371:
Nice consolidation!

imap_processing/codice/codice_l0.py
bourque: From what I understand from Gavin's comment in the `xtce` slack channel, it sounds like his preference would be to explicitly define what a value of `0` means in the packet definition file. I will ask Joey about this in our next tagup.

imap_processing/codice/codice_l0.py
bourque: Good call, this should really be called `datasets` now instead of `packets`. But I think I will keep the module around, at least for now, as a way to distinguish L0 code from L1 code.

imap_processing/codice/packet_definitions/codice_packet_definition.xml
bourque: Based on Gavin's comments I think this will likely be the workaround I go with, but I want to ask Joey about this first.

bourque:
After discussing with Joey, a `LAST_OPCODE` value of 0 is unexpected to happen in-flight, and is likely in there now because Greg Dunn has been filling some values with zeros for simulated data. For now, I added `<xtce:Enumeration value="0" label="ERROR" />` to the packet definition file as a workaround, but this should be revisited once I acquire updated simulated housekeeping data (mentioned in the `TODO`).

PR #813, opened 2024-09-05
Updated various CDF attributes/variable names to match expected nomenclature

### LINKED ISSUE #774: CoDICE: Ensure CDF attributes/variables match the expected nomenclature
Go through the CoDICE data products and make sure that everything matches the expected nomenclature described here: https://lasp.colorado.edu/galaxy/pages/viewpage.action?spaceKey=IMAP&title=IMAP+Nomenclature+of+Science+Data+Products

bourque:
# Change Summary

## Overview
This PR updates some of the CoDICE CDF attribute and variable names to match the expected nomenclature that is documented here: https://lasp.colorado.edu/galaxy/display/IMAP/IMAP+Nomenclature+of+Science+Data+Products. I also changed a few values in order to be more consistent with the CoDICE L1a validation CDFs I received.

## Updated Files

   - Updated various names and values to match explected nomenclature and also be more consistent with the validation CDFs.

Closes #774

### DIFF
--- a/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
@@ -39,7 +39,7 @@ esa_step:
 
 inst_az:
     <<: *default
-    CATDESC: Azimuth Index
+    CATDESC: Instrument Azimuth Index
     FIELDNAM: Azimuth Index
     FORMAT: I2
     LABLAXIS: " "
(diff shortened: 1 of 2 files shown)


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -25,94 +25,83 @@ codice_support_attrs: &support_default
...
-    FIELDNAM: Energy step
+    FIELDNAM: Energy Index
     FORMAT: I3
-    LABLAXIS: energy step
+    LABLAXIS: Energy Index
+    SI_CONVERSION: " > "
     UNITS: ' '
     VALIDMIN: 0
-    VALIDMAX: 127
+    VALIDMAX: 128
```
tech3371: Are you sure you want this to be 128? Just double checking

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -25,94 +25,83 @@ codice_support_attrs: &support_default
...
     VAR_TYPE: support_data
 
-inst_az_attrs:
+inst_az:
     <<: *default
-    CATDESC: Azimuth
-    FIELDNAM: Azimuth
+    CATDESC: Azimuth Index
```
```suggestion
    CATDESC: Instrument Azimuth Index
```

imap_processing/codice/codice_l1a.py
```diff
@@ -165,8 +157,7 @@ def create_science_dataset(
                 "epoch": epoch,
                 "inst_az": inst_az,
                 "spin_sector": spin_sector,
-                "energy": energy_steps,
-                "energy_label": energy_label,
+                "esa_step": esa_step,
```
tech3371, line 160: If you remove those labels, you are gonna get error for data whose dimension have more than `epoch` as dimension when we need to update `cdflib`. To test it, some of us have been updating `cdflib` locally and then run the same test and it will raise those helpful errors. Can you do the same?

Eg.
```
dims = ["epoch", "inst_az", "spin_sector", "esa_step"]
```

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: I pulled this from the L1A validation CDFs that joey gave me but I do think it should be 127. I will double check with Joey.

imap_processing/codice/codice_l1a.py
bourque: Ah yes, I forgot that I added `energy_label` in anticipation of the `cdflib` upgrade in the first place. But I think I found a new (better) solution, which is to define the `LABL_PTR_[1|2|3]`s in the attribute YAML file. It appears to work in both `cdflib` `1.3` and `1.2`.

tech3371: Good to know. I am surprised that you were able to reuse dimension cdf attrs for labels like this
```
    LABL_PTR_1: inst_az
    LABL_PTR_2: spin_sector
    LABL_PTR_3: esa_step
```
I thought it has to be metadata type.

tech3371 (approved):

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: Joey confirmed that this should be `127` (and that the exact values in his validation CDFs should not be considered the strict requirement).

PR #814, opened 2024-09-06
Pointing frame updates

laspsandoval:
# Change Summary

## Overview
Adds ability to combine different pointings into the frame kernel.

## Updated Files
- kernels.py
   - Adds ability to combine different pointings into the frame kernel.
   - Results in imap_pointing_frame.bc which will be used for ENA teams for l1b.

## Testing
- test_kernels.py
   - Tests kernels.py

Necessary test data:
IMAP_spacecraft_attitude.bc
imap_pointing_frame.bc


vmartinez-cu:
What does "et_" refer to in many of the variable names. Is this ephemeris time?

laspsandoval:
> What does "et_" refer to in many of the variable names. Is this ephemeris time?

Yes. That is correct.

imap_processing/spice/kernels.py
```diff
@@ -173,7 +173,7 @@ def wrapper_ensure_spice(*args: Any, **kwargs: Any) -> Any:
 
 
 @contextmanager
-def spice_ck_file(pointing_frame_path: str) -> Generator[int, None, None]:
+def spice_ck_file(pointing_frame_path: Path) -> Generator[int, None, None]:
```
tmplummer: Can you rename this to something with "open" so we know that it is opening a file for editing?

imap_processing/spice/kernels.py
```diff
@@ -187,15 +187,18 @@ def spice_ck_file(pointing_frame_path: str) -> Generator[int, None, None]:
...
         yield handle
     finally:
         spice.ckcls(handle)
 
 
 @ensure_spice
-def create_pointing_frame(pointing_frame_path: Optional[Path] = None) -> Path:
+def create_pointing_frame(pointing_frame_path: Path) -> None:
     """
     Create the pointing frame.
```
tmplummer, line 206: I would like to see more detailed documentation for this function. For example, what kernels are required to be furnished? Are there other assumptions made upon calling this function? How many CK kernels can be furnished? etc.

imap_processing/spice/kernels.py
```diff
@@ -204,116 +207,123 @@ def create_pointing_frame(pointing_frame_path: Optional[Path] = None) -> Path:
...
-            # Rates (seconds per tick) at which the quaternion and
-            # angular velocity change.
-            np.array([1.0]),
+    # Get IDs.
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.gipool
+    id_imap_dps = spice.gipool("FRAME_IMAP_DPS", 0, 1)
+    id_imap_sclk = spice.gipool("CK_-43000_SCLK", 0, 1)
+
+    # If the pointing frame kernel already exists, find the last time.
+    if pointing_frame_path.exists():
```
tmplummer, line 250: This is confusing. I think that the docstring is wrong:
```
pointing_frame_path : Path
        Directory of where pointing frame will be saved.
```
But here it looks like it is the path to the file, not a directory.

imap_processing/spice/kernels.py
```diff
@@ -204,116 +207,123 @@ def create_pointing_frame(pointing_frame_path: Optional[Path] = None) -> Path:
...
-            np.array([0.0, 0.0, 0.0]),
-            # Rates (seconds per tick) at which the quaternion and
-            # angular velocity change.
-            np.array([1.0]),
+    # Get IDs.
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.gipool
+    id_imap_dps = spice.gipool("FRAME_IMAP_DPS", 0, 1)
+    id_imap_sclk = spice.gipool("CK_-43000_SCLK", 0, 1)
+
+    # If the pointing frame kernel already exists, find the last time.
```
tmplummer, line 249: I think that we want a one-to-one relationship between spacecraft CKs and pointing frame CKs. This would mean that anytime we get a new spacecraft CK, we generate a new pointing frame CK and there will never be one that already exists.

imap_processing/spice/kernels.py
```diff
@@ -204,116 +207,123 @@ def create_pointing_frame(pointing_frame_path: Optional[Path] = None) -> Path:
...
+        et_end_pointing_frame = None
 
-    return pointing_frame_path
-
-
-@ensure_spice
-def _get_et_times(ck_kernel: str) -> tuple[float, float, np.ndarray]:
+    # TODO: Query for .csv file to get the pointing start and end times.
+    # TODO: Remove next four lines once query is added.
+    ck_kernel, _, _, _ = spice.kdata(0, "ck")
```
tmplummer: I think that it would "harden" this code against errors if the path of the CK was passed in rather than just assumed to be the first CK in the kernel pool.

Along those lines, it would also be good to check that it is the only CK that has been furnished.

imap_processing/spice/kernels.py
```diff
@@ -204,116 +207,123 @@ def create_pointing_frame(pointing_frame_path: Optional[Path] = None) -> Path:
...
+    id_imap_spacecraft = spice.gipool("FRAME_IMAP_SPACECRAFT", 0, 1)
+    ck_cover = spice.ckcov(
+        ck_kernel, int(id_imap_spacecraft), True, "INTERVAL", 0, "TDB"
```
tmplummer: Not 100% sure, but shouldn't this be retrieving the "SEGMENT" coverage. The way I read intervals is that it has to do with interpolation intervals rather than intervals of coverage segments. This is to be verified in SPICE definition of CK segment and interpolation interval.

imap_processing/tests/spice/test_kernels.py
```diff
@@ -159,12 +175,77 @@ def test_et_times(pointing_frame_kernels):
...
 
-    assert et_start == 802008069.184905
-    assert et_end == 802015267.184906
     assert et_times[0] == et_start
     assert et_times[-1] == et_end
 
     return et_times
+
+
+@ensure_spice
```
tmplummer: Tests should not be decorated with @ensure_spice

tmplummer (commented):
I have reviewed most of the code and have some feedback. I ran out of time and haven't done an in-depth review of the tests. I will plan to come finish my review after school pickup or sometime this evening.

imap_processing/tests/spice/test_data/IMAP_spacecraft_attitude.bc
tmplummer, line 1: I think we should rename this file to something more descriptive. `sim_1yr_imap_attitude.bc`?

imap_processing/tests/spice/test_data/imap_pointing_frame.bc
tmplummer, line 1: I'm curious how long this took to generate. It might be worth holding off on checking it in?

imap_processing/tests/spice/test_kernels.py
```diff
@@ -159,12 +175,77 @@ def test_et_times(pointing_frame_kernels):
...
+    assert num_intervals == 1
+
+
+@ensure_spice
+def test_multiple_pointings(multiple_pointing_kernels, spice_test_data_path):
+    """Tests create_pointing_frame function with multiple pointing kernels."""
+    spice.furnsh(multiple_pointing_kernels)
+
+    create_pointing_frame(
+        pointing_frame_path=spice_test_data_path / "imap_pointing_frame.bc"
```
tmplummer: Wait, why are you creating the pointing frame in the `imap_test_data_path`? This should be going into a temp directory. 
```suggestion
        pointing_frame_path=imap_data_access.config["DATA_DIR"] / "imap_pointing_frame.bc"
```

imap_processing/tests/spice/test_kernels.py
```diff
@@ -159,12 +175,77 @@ def test_et_times(pointing_frame_kernels):
...
+
+    create_pointing_frame(pointing_frame_path=tmp_path / "imap_dps.bc")
+    ck_cover = spice.ckcov(
+        str(tmp_path / "imap_dps.bc"), -43901, True, "INTERVAL", 0, "TDB"
+    )
+    num_intervals = spice.wncard(ck_cover)
+    assert num_intervals == 1
+
+
+@ensure_spice
```
tmplummer: Remove `@ensure_spice` decorator.

imap_processing/spice/kernels.py
laspsandoval: I think that makes sense. That could be handled by how we name the pointing frame ck. We can keep to the same convention as the spacecraft ck.

imap_processing/tests/spice/test_data/imap_pointing_frame.bc
laspsandoval: Approximately 5-10 minutes. Hence already checking it in.

imap_processing/spice/kernels.py
laspsandoval: If I use "INTERVAL" then the number of intervals = 366.
If I use "SEGMENT" then the number of segments = 1.

I think it depends on how the ck kernel is defined. Our test ck kernel uses intervals and not segments. We will delete this line once we have the .csv, but it would be nice to know when interval vs segment should be used. 

From my quick search:
Segments can be broken into intervals, but intervals cannot be broken into segments. The interval is time-based only. I'll ask Nick Dutton tomorrow at our meeting for further clarification on when each is used.

imap_processing/spice/kernels.py
laspsandoval: Let's talk about this at tomorrow's 1pm meeting.

tmplummer (approved):
Looks good! Thanks for addressing comments.

PR #815, opened 2024-09-06
HIT sci decom - group packets & parse count rates

### LINKED ISSUE #820: HIT: Group L0 science packets into science frames
### Algorithm Description:

Group L0 binary science data packets into groups of 20 to create science frames. Use group flag and sequence counts to assemble the frames.

### LINKED ISSUE #821: HIT: Unpack L0 count rates data
Unpack and organize binary count rates science data into data variables needed for the L1A product.

vmartinez-cu:
This code begins the work of decommutating and decompressing HIT L0 science data. 

Background info: 
`packet_file_to_datasets` in utils.py performs an initial decom of HIT ccsds files. This function returns a dictionary with an xarray dataset per APID. For HIT science data (APID = 1251), the xarray dataset contains the unpacked CCSDS headers and the science data in binary to be manually unpacked in the code (this is necessary due to how the TLM file is organized). The science dataset is the input for this code and will return a dataset with decommutated and decompressed science data. So far, this code performs the following tasks:

- Assemble science frames - a complete set of data (i.e. science frame) consists of science data from 20 packets. This code groups science data into science frames. Checks are made to ensure the packets in a science frame belong together. Invalid science frames are skipped. For each valid science frame, data is further categorized by the l1a products that are needed (i.e. count rates and event data). The first 6 packets in the frame are count rates and the remaining 14 packets are event data. These get added to the dataset as new data variables and are still in binary.
- Parse count rates - The count rates data per science frame need to be manually unpacked. The code handles this and adds the data to the dataset as new data variables

The `decom_hit` function is the starting point for processing the L0 data. This function will be called from the hit_l1a.py file (to be implemented in a future PR)

## New Files

- decom_hit.py
   - L0 data processing
- hit_packet_definitions.xml
   - combines definitions for all APIDs
- sci_sample.ccsds
   - sample science data

## Deleted Files
- P_HIT_SCIENCE.xml
   - replaced with hit_packet_definitions.xml

## Testing
- test_hit_decom.py


imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+    # science frame header
+    "hdr_unit_num": HITPacking(2, 2, (1,)),
+    "hdr_frame_version": HITPacking(6, 6, (1,)),
+    "hdr_status_bits": HITPacking(8, 8, (1,)),
+    "hdr_minute_cnt": HITPacking(8, 8, (1,)),
+    # ------------------------------------------
+    # spare
+    "spare": HITPacking(24, 24, (1,)),
+    # ------------------------------------------
+    # erates
```
tech3371: can you spell these out in comments? Just the `erates` or `evrates` and so on.

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+
+from imap_processing import imap_module_directory
+from imap_processing.utils import packet_file_to_datasets
+
+HITPacking = namedtuple(
+    "HITPacking",
+    [
+        "bit_length",
+        "section_length",
+        "shape",
```
tech3371, line 18: can you add comment about this?

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+    "livetime": HITPacking(16, 16, (1,)),
+    "num_trig": HITPacking(16, 16, (1,)),
+    "num_reject": HITPacking(16, 16, (1,)),
+    "num_acc_w_pha": HITPacking(16, 16, (1,)),
+    "num_acc_no_pha": HITPacking(16, 16, (1,)),
+    "num_haz_trig": HITPacking(16, 16, (1,)),
+    "num_haz_reject": HITPacking(16, 16, (1,)),
+    "num_haz_acc_w_pha": HITPacking(16, 16, (1,)),
+    "num_haz_acc_no_pha": HITPacking(16, 16, (1,)),
+    # -------------------------------------------
```
tech3371, line 53: can you add comment here what this section is? Above abbr. comments were still helpful!

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+    "nl2": HITPacking(16, 16, (1,)),
+    "nl3": HITPacking(16, 16, (1,)),
+    "nl4": HITPacking(16, 16, (1,)),
+    "npen": HITPacking(16, 16, (1,)),
+    "nformat": HITPacking(16, 16, (1,)),
+    "naside": HITPacking(16, 16, (1,)),
+    "nbside": HITPacking(16, 16, (1,)),
+    "nerror": HITPacking(16, 16, (1,)),
+    "nbadtags": HITPacking(16, 16, (1,)),
+    # -------------------------------------------
```
tech3371, line 78: same here

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+
+    # L0 file path
+    packet_file = Path(imap_module_directory / "tests/hit/test_data/sci_sample.ccsds")
+
+    datasets_by_apid = packet_file_to_datasets(
+        packet_file=packet_file,
+        xtce_packet_definition=packet_definition,
+    )
+
+    science_dataset = datasets_by_apid[1252]
```
tech3371: You can create ENUM for apid like I have done for other instruments.

tech3371 (commented):
I skimmed it and so far so good. I will look at remaining function first thing on Monday. Nice work!

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: Good suggestion! I currently have this in hit_l1a.py, which will call this file to perform the decom work on the science packets. I might move the ENUM for apid to this file instead, but haven't decided yet

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+        from a CCSDS file.
+
+    Returns
+    -------
+    sci_dataset : xr.Dataset
+        Updated xarray dataset.
+    """
+    # sc_tick contains spacecraft time per packet
+    sci_dataset.coords["sc_tick"] = sci_dataset["sc_tick"]
+    sci_dataset = sci_dataset.swap_dims({"epoch": "sc_tick"})
```
tech3371, line 316: I am not sure if you need this function. I believe `packet_file_to_datasets` already creates `epoch` using `sc_tick`.

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+    should have a sequence flag equal to 2.
+
+    Valid science frame sequence flags
+    [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 2]
+
+    Additionally, packets have a sequence counter. This function also
+    checks if the counters are in sequential order.
+
+    Both conditions need to be met for a science frame to be considered
+    valid.
```
tech3371: Really helpful doc strings!

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+    #  incomplete science frames in the middle of the CCSDS file.
+    #  The code currently skips all incomplete science frames.
+
+    # Initialize lists to store data from valid science frames
+    count_rates_binary = []
+    pha_binary = []
+    epoch_science_frame = []
+
+    science_frame_start = 0
+    while science_frame_start + 20 <= len(sci_dataset.epoch):
```
tech3371: I feel like you may want to approach this differently. You may want to find indices of all 1s and 2 and then loop through those and check if those are valid science frame or not. If you start incrementing with 20 from beginning, you may miss good packets due to being off from where you should start. I am happy to jump on call if it helps.

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+            epoch_science_frame.append(epoch_data_chunk[0])
+            science_frame_start += 20  # Move to the next science frame
+        else:
+            print(
+                f"Invalid science frame found with starting packet index = "
+                f"{science_frame_start}"
+            )
+            # Skip science frame and move on to the next science frame packets
+            # Get index for the first packet in the next science frame
+            start = get_starting_packet_index(
```
tech3371: you are kind of doing what I mention above but instead of looking for the next `1`, you can find all of 1s ahead and then check if it's valid science frame.

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,464 @@
...
+            science_frame_start : science_frame_start + 20
+        ]
+        epoch_data_chunk = sci_dataset.epoch[
+            science_frame_start : science_frame_start + 20
+        ]
+
+        if is_valid_science_frame(seq_flgs_chunk, src_seq_ctr_chunk):
+            # Append valid data to lists
+            count_rates_binary.append("".join(science_data_chunk.data[0:6]))
+            pha_binary.append("".join(science_data_chunk.data[6:]))
```
tech3371: can you add comment about what those .data[0:6]` is doing?

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: I replace the epoch dimension later to only contain one time per science frame rather than per packet since that is what's needed for the science data. However, I still need the per packet sc_tick times for the ccsds header fields since those are not being grouped by science frame. Essentially, I need two time dimensions. One per packet for the ccsds header fields and one per science frame for the science data. I can explain this better in the docstring

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: Yes, this would be helpful! I'll find a time to chat more about this

vmartinez-cu: I updated this to use numpy.where and numpy.diff. Please take a look when you get a chance!

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,463 @@
...
+            pha_binary.append("".join(science_data_chunk[6:]))
+            # Just take first packet's epoch for the science frame
+            epoch_science_frame.append(epoch_data_chunk[0])
+            last_index_of_frame = start + packets_in_frame
+        else:
+            # TODO: log issue
+            # Skip invalid science frame and move on to the next one
+            print(
+                f"Invalid science frame found with starting packet index = " f"{start}"
+            )
```
tech3371: That makes sense to me. Nice work!

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,463 @@
...
+    epoch_science_frame = np.array(epoch_science_frame)
+    sci_dataset = sci_dataset.drop_vars("epoch")
+    sci_dataset.coords["epoch"] = epoch_science_frame
+    sci_dataset["count_rates_binary"] = xr.DataArray(
+        count_rates_binary, dims=["epoch"], name="count_rates_binary"
+    )
+    sci_dataset["pha_binary"] = xr.DataArray(
+        pha_binary, dims=["epoch"], name="pha_binary"
+    )
+    return sci_dataset
```
tech3371: Now, you need to figure out how to filter rest of data variables in `sci_dataset` to filter out only good science frame's data. Otherwise you will have different data shape between `count_rates_binary`, `pha_binary`, and `epoch` vs CCSDS header variables. You can use `.isel` on `sci_dataset`.

vmartinez-cu:
Added unit tests and updated this PR from draft to ready for review. There are pre-commit errors I still need to fix but the code is ready for review

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,484 @@
...
+
+    # Convert sequence flags and counters to NumPy arrays for vectorized operations
+    seq_flgs = sci_dataset.seq_flgs.values
+    seq_ctrs = sci_dataset.src_seq_ctr.values
+    science_data = sci_dataset.science_data.values
+    epoch_data = sci_dataset.epoch.values
+
+    # Number of packets in the file
+    total_packets = len(epoch_data)
+    frame_size = 20
```
sdhoyt: This is just an idea that I'm not sure is really better, so feel free to ignore it: you could make the flag pattern a global variable and use the length of that to get the frame_size instead of hardcoding that. Another option is to pull the APID Enum out into a hit config type file and the flag pattern could be added as a variable in that file. The flag pattern could then be imported into this file.

sdhoyt (approved):
Looks good to me! Great work handling the frames!

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,485 @@
...
+    packet_file = Path(imap_module_directory / "tests/hit/test_data/sci_sample.ccsds")
+
+    datasets_by_apid = packet_file_to_datasets(
+        packet_file=packet_file,
+        xtce_packet_definition=packet_definition,
+    )
+
+    science_dataset = datasets_by_apid[1252]
+    science_dataset = decom_hit(science_dataset)
+    print(science_dataset)
```
tech3371: remove print?

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,485 @@
...
+    # Parse count rates data from binary and add to dataset
+    parse_count_rates(sci_dataset)
+
+    # TODO:
+    #  Parse binary PHA data and add to dataset (function call)
+
+    return sci_dataset
+
+
+# TODO: remove main after code is finalized
```
tech3371: remove this now? The way I have been getting around this is to write a test to call these exact things and it will help both the test and seeing our code work which you are already doing. After that, I call test like this to see temporary print statement, `pytest imap_processing/tests/spice/test_geometry.py -s`. `-s` will print those print statement.

imap_processing/hit/l0/decom_hit.py
tech3371: I think global variable could be helpful. Then you can use that global variable in `line 246` and `line 283` since they should be equal.

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,485 @@
...
+            f"from previous day's ccsds file"
+        )
+    last_index_of_last_frame = starting_indices[-1] + frame_size
+    if last_index_of_last_frame:
+        remaining_packets = total_packets - last_index_of_last_frame
+        if 0 < remaining_packets < frame_size:
+            print(
+                f"{remaining_packets} packets at end of file belong to science frame "
+                f"from next day's ccsds file"
+            )
```
tech3371, line 384: I like this check a head of all

imap_processing/hit/l0/decom_hit.py
```diff
@@ -0,0 +1,485 @@
...
+    # 20, from the flags array and is used to slide the "window" across the array
+    # and compare the sub-arrays with the predefined pattern.
+    window_size = len(flag_pattern)
+    windows = np.lib.stride_tricks.sliding_window_view(flags, window_size)
+    # Find where the windows match the pattern
+    matches = np.all(windows == flag_pattern, axis=1)
+    # Get the starting indices of matches
+    match_indices = np.where(matches)[0]
+    # Filter for only indices from valid science frames with sequential counters
+    valid_indices = get_valid_indices(match_indices, counters, window_size)
```
tech3371: I wonder if you can combine remaining two functions into one to reduce code tracing. You can ignore for now but something to think about going forward.

tech3371 (approved):
Nice work!

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: Great idea. I've added a TODO for now so I don't forget

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: I like this suggestion! I've added a global variable for now and a TODO to consider moving it to a config file in the future

PR #816, opened 2024-09-06
Added FIELDNAMs, LABLAXISs, and CATDESCs for housekeeping attributes

### LINKED ISSUE #776: CoDICE: Add catdescs and fieldnames to housekeeping CDF attributes
Currently the housekeeping data uses "TBD" for many of the `catdesc` and `fieldname` fields in the CDF attributes. These should be updated to actually store useful information. This could possibly be done programmatically by grabbing the `short_description` and `long_description` fields in the XTCE definition files. Perhaps something like:

```python
for key, value in (packet.header | packet.data).items():
    fieldname = value.short_description
    catdesc = value.long_description
```

bourque:
This PR adds `FIELDNAM`s, `CATDESC`s, and `LABLAXIS`s to the CoDICE L1 housekeeping CDFs.

Closes #776


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -392,4 +388,817 @@ lo_nsw_species-heplus:
...
+    <<: *default
+    CATDESC: Packet version - this will be incremented each time the format of the packet changes.
+    FIELDNAM: Packet Version
+    LABLAXIS: PACKET_VERSION
+    VAR_TYPE: support_data
+
+cmdexe:
+    <<: *default
+    lablaxis: CMDEXE
+    CATDESC: Number of commands that have been executed. Counts 0-255, then rolls over to 0.  Reset via CLR_LATCHED_SINGLE(COMMAND_COUNTS) [also resets cmdjrct, cmdacc, itf_error counts)
```
tech3371: You will need to move ones with long description to VAR_NOTES. Andriy pointed that out.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -20,10 +20,6 @@ default_attrs: &default
     VALIDMAX: 9223372036854775807
     VAR_TYPE: data
```
tech3371, line 24: You can add `DEPEND_0` here and then you don't need to add it in `codice_l1a.py`.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: Good catch! It looks like the character limit for `CATDESC` is 80 characters according to the ISTP website, so I will go by that.

bourque:
I think this is good to go once I can get an approved review 😉

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -450,23 +450,23 @@ shcoarse:
 
 packet_version:
     <<: *default
-    CATDESC: Packet version - this will be incremented each time the format of the packet changes.
```
tech3371: I am pretty sure `CATDESC` is a required field but you may have set it to `' '` in your `default`. Later you might want to have short description in `CATDESC`

tech3371 (approved):
One comment. Besides, it looks good

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: Good to know, thanks! If I want to add `CATDESC`s for the ones that don't have them, I will have to come up with that myself most likely. But I will add a TODO for those and punt on that for a future PR.

PR #818, opened 2024-09-09
689 spice update met to j2000ns

### LINKED ISSUE #689: SPICE - Update met_to_j2000ns
### Summary

Update /Users/lasa6858/imap_processing/imap_processing/cdf/utils/met_to_j2000ns so that it is more accurate by using the spacecraft clock kernel.

MET - spacecraft ticks (close to a second)
SCLK - shows how to adjust clock ticks to a true time

kernel location: https://lasp.colorado.edu/galaxy/display/IMAP/Data

You will need to 
1. furnish the sclk ([imap_sclk_0000.tsc](https://lasp.colorado.edu/galaxy/download/attachments/173719425/imap_sclk_0000.tsc?version=1&modificationDate=1711400549607&api=v2)) and leapsecond ([naif0012.tls](https://lasp.colorado.edu/galaxy/download/attachments/173719425/naif0012.tls?version=1&modificationDate=1711400551089&api=v2)) kernels similar to how it works in spice_examples.py:
with spice.KernelPool(kernels): etc.
2. Use relevant tool/tools listed here to update the met: https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy

### Requirements

The more specific the better.
  - Implement SPICE SCLK in met_to_j2000ns

### Related tickets

_No response_

tmplummer:
# Change Summary

## Overview
Moved the `met_to_j2000ns` function from the cdf/utils.py to spice/time.py and re-implemented that function to use SPICE tools for the time conversion.

## Updated Files
There are lots of updated files as a result of moving the function `met_to_j2000ns`. I won't list all of those here. The most important changes are in the listed files below:
   - moved met_to_j2000ns out of this file
- imap_processing/spice/__init__.py
   - added global constant declaration of IMAP_SC_ID
   - Moved and reimplemented `met_to_j2000ns` in this file
   - Added wrapper for `spice.sct2e` that vectorizes and adds `@ensure_spice`
   - Made important changes to SPICE related fixtures that give all tests a metakernel pointed to by `SPICE_METAKERNEL` env var
   - Added test coverage for the new spice time functions

## Testing
One test is still broken. I need feedback from @tech3371 about why this change broke this test.

closes #689


imap_processing/spice/time.py
```diff
@@ -1 +1,72 @@
...
+from typing import Union
+
+import numpy as np
+import numpy.typing as npt
+import spiceypy as spice
+
+from imap_processing.spice import IMAP_SC_ID
+from imap_processing.spice.kernels import ensure_spice
+
+TICK_DURATION = 2e-5  # 20 microseconds
```
laspsandoval: Where did this value come from?

imap_processing/spice/time.py
```diff
@@ -1 +1,72 @@
...
+    There are two options when using SPICE to convert from SCLK time (MET) to
+    J2000. The conversion can be done on SCLK strings as input or using double
+    precision continuous spacecraft clock "ticks". The latter is more accurate
+    as it will correctly convert fractional clock ticks to nanoseconds. Since
+    some IMAP instruments contain clocks with higher precision than 1 SCLK
+    "tick" which is defined to be 20 microseconds, according to the sclk kernel,
+    it is preferable to use the higher accuracy method.
+    """
+    sclk_ticks = np.asarray(met, dtype=float) / TICK_DURATION
+    return np.asarray(_sct2e_wrapper(sclk_ticks) * 1e9, dtype=np.int64)
```
laspsandoval, line 44: Why do we want dtype np.int64 if _sct2e_wrapper returns a float?

laspsandoval (approved):
A couple of comments, but I think it is good.

imap_processing/spice/time.py
```diff
@@ -1 +1,72 @@
...
+    sclk_ticks : Union[float, Collection[float]]
+        Input sclk ticks value(s) to be converted to ephemeris time.
+
+    Returns
+    -------
+    ephemeris_time: np.ndarray
+        Ephemeris time, seconds past J2000.
+    """
+    if isinstance(sclk_ticks, Collection):
+        return np.array([spice.sct2e(IMAP_SC_ID, s) for s in sclk_ticks])
```
greglucas, line 70: 😮 This is not good if even a simple function like this isn't vectorized in the C-level for us :-/ I can only imagine that the ENA teams are going to complain about the slow lookups for every individual particle we are querying.

imap_processing/tests/conftest.py
```diff
@@ -119,9 +142,16 @@ def get_test_kernels_to_load():
...
     make_metakernel_from_kernels(metakernel_path, kernels_to_load)
-    monkeypatch.setenv("SPICE_METAKERNEL", str(metakernel_path))
+    monkeypatch_session.setenv("SPICE_METAKERNEL", str(metakernel_path))
     yield str(metakernel_path)
     spice.kclear()
+
+
+@pytest.fixture()
+def _metakernel_path_not_set(monkeypatch):
+    """Temporarily unsets the EMUS_METAKERNEL_PATH environment variable"""
```
```suggestion
    """Temporarily unsets the SPICE_METAKERNEL environment variable"""
```

imap_processing/tests/conftest.py
```diff
@@ -119,9 +142,16 @@ def get_test_kernels_to_load():
...
     kernels_to_load = get_test_kernels_to_load()
     make_metakernel_from_kernels(metakernel_path, kernels_to_load)
-    monkeypatch.setenv("SPICE_METAKERNEL", str(metakernel_path))
+    monkeypatch_session.setenv("SPICE_METAKERNEL", str(metakernel_path))
     yield str(metakernel_path)
     spice.kclear()
+
+
+@pytest.fixture()
+def _metakernel_path_not_set(monkeypatch):
```
greglucas: This reads more as a question / if-statement check to me. What about `_unset_metakernel_path` or "reset" or something similar?

imap_processing/tests/spice/test_time.py
```diff
@@ -1 +1,30 @@
...
+
+
+@pytest.mark.parametrize("sclk_ticks", [0.0, np.arange(10)])
+def test_sct2e_wrapper(sclk_ticks):
+    """Test for `_sct2e_wrapper` function."""
+    et = _sct2e_wrapper(sclk_ticks)
+    if isinstance(sclk_ticks, float):
+        assert isinstance(et, float)
+    else:
+        assert len(et) == len(sclk_ticks)
```
greglucas, line 35: Can you also check that these equal the offset from the EPOCH, or does that not work due to leap seconds / other SPICE details?

```python
np.testing.assert_array_equal(et, EPOCH + sclk_ticks)
```

imap_processing/spice/time.py
greglucas: We are always passing in an array of items too, so do we need the option of a float return and special casing that versus iterating over the array?

imap_processing/spice/time.py
tmplummer: That is from the SCLK kernel. I looked for spice functions to retrieve that value and it can be gotten programmatically, but had a small error associated with doing it that way.
```
print(spice.sct2e(-43, 1) - spice.sct2e(-43, 0))
print(20e-6)

2.002716064453125e-05
2e-05
```

imap_processing/spice/time.py
tmplummer: > Why do we want dtype np.int64 if _sct2e_wrapper returns a float?

This function is specific to returning nanoseconds since the J2000 epoch for use in CDF. The CDF datatype for epoch is int64. Paul Janzen has confirmed that integer nanoseconds is high enough precision.

> We are always passing in an array of items too, so do we need the option of a float return and special casing that versus iterating over the array?

I can certainly add that. I replicated exactly what was in the `met_to_j2000ns` function here but agree that returning a single `int64` instead of an array of length 1 is preferable. Is that what you meant?

imap_processing/spice/time.py
tmplummer: I think the real ENA concern is in calculating exposure times where the "correct" way to do it is to query every single clock tick. This is why it is important not to use SPICE for determining spin phase. We can precomupte all of the geometry of the pointing set ring and then bin clock ticks into that ring based on spin phase rather than going through spice.

imap_processing/tests/spice/test_time.py
tmplummer: That can be done for the test sclk kernel because the spacecraft clock is defined to tick at exactly one tick per second. Once we have real kernels, that is no longer true.

tmplummer: When implementing this test, I decided that checking values would actually be testing the `spice.sct2e` function rather than the wrapper that I implemented.

imap_processing/spice/time.py
greglucas: > I can certainly add that. I replicated exactly what was in the met_to_j2000ns function here but agree that returning a single int64 instead of an array of length 1 is preferable. Is that what you meant?

I actually meant to always return an array from the underlying wrapped function. You have an `isinstance()` check in there that returns a single float if it is only one value, but here we are always wrapping the input in `np.asarray()` and thus will never hit the single float path from this call location.

imap_processing/tests/spice/test_time.py
greglucas: Ahh, I see, that makes sense, thanks for clarifying. It might still be a better check here if we do have the known quantity in the tests so we are asserting that your TICK calculation was done correctly then I think, rather than just saying the returned arrays are the same shape.

imap_processing/spice/time.py
tmplummer: > I actually meant to always return an array from the underlying wrapped function.

Oh, Ok. I suspect that the underlying wrapper will be used in other time conversion functions, so I would propose that we leave it as is unless there is an argument for always returning an array. I try to follow rather than lead on function API design philosophy. I was thinking that there are numpy functions which return values that look like the input rather than always converting to an array, so I thought it made sense to do that here too.

imap_processing/tests/spice/test_time.py
tmplummer: The TICK calculation only comes into play in the higher level `met_to_j2000ns` function and I do check that the value returned from that function matches the et calculated purely by SPICE... Now that I look at it, I am using the same `TICK_DURATION` so that test is a bit circular. I will change that test to calculate the tick duration using SPICE.

imap_processing/tests/conftest.py
```diff
@@ -119,9 +142,16 @@ def get_test_kernels_to_load():
...
+    monkeypatch_session.setenv("SPICE_METAKERNEL", str(metakernel_path))
     yield str(metakernel_path)
     spice.kclear()
+
+
+@pytest.fixture()
+def _unset_metakernel_path(monkeypatch):
+    """Temporarily unsets the SPICE_METAKERNEL environment variable"""
+    if os.getenv("SPICE_METAKERNEL", None) is not None:
+        monkeypatch.delenv("SPICE_METAKERNEL")
```
greglucas, line 157: Looks like you can avoid the if check with a raising=False keyword.
https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.delenv
```suggestion
    monkeypatch.delenv("SPICE_METAKERNEL", raising=False)
```

greglucas (approved):

greglucas:
Looks great to me. Hopefully @tech3371 can respond about the SWAPI test failures.

tech3371:
I am looking into why SWAPI is failing. My hunch is that it's has to do with their sample data's SHCOARSE values being like this: 24, 25, 26, 27, and so on. That may be causing to have less full sweep dataset. Will update soon. If you like to merge this PR, don't hesitate to comment out that test as xfail and I can tackle it separately.

PR #824, opened 2024-09-10
Spin data and Spin Phase function setup

### LINKED ISSUE #806: SPICE - Spin phase query utils function - using spin start and spin rate data.
For a given input time, this function will need to return spin phase angle of spacecraft.

Function signature
```
def get_spacecarft_spin_phase(query_times: Union[float, np.ndarray]) -> Union[float, np.ndarray]:
"""
Compute the floating point spin phase using spin table [0, 1] at the input times.
"""
spin_table = get_spin_table()  # Start with this function?
spin_phase = (query_time - (spin_start_seconds + spin_start_subseconds)) / spin_period_sec
```

Could also add a function to get instrument spin phase (Put this in a new ticket?)
```
def get_instrument_spin_phase(query_times: Union[float, np.ndarray], instrument_id: IntEnum) -> Union[float, np.ndarray]:
instrument_spin_phase = get_spacecraft_spin_phase + get_instrument_phase_offset()
```

fixture function signature
```
def gen_spin_table(start_time: datetime, end_time: datetime = None) -> Path:
"""
Generate a pointing table CSV covering one or more days.

Parameters
-----------
start_time : provides the starting day of year
end_time : optionally specify end day of year. If not provided default to one day

Returns
--------
Path to CSV file
```

Tim's prototype code:
```
# Spin table contains the following fields:
# (spin_number, spin_start_sec, spin_start_subsec, spin_period_sec, spin_period_valid, spin_phas_valid, spin_period_source, thruster_firing)
seconds_per_point = 24 * 60 * 60

# get the very first spin start sclk
_, start_sclk_sec, start_sclk_ticks = parse_sclk_str(spice.sce2s(SC_ID, cov[0]))
_, end_sclk_sec, end_sclk_ticks = parse_sclk_str(spice.sce2s(SC_ID, cov[-1]))

# fill the whole year worth of spin data with assuming ideal 15 seccond spins
spin_dict = dict()
spin_start_sec = np.arange(start_sclk_sec, end_sclk_sec, 15, dtype=np.uint64)
spin_dict["spin_number"] = np.arange(spin_start_sec.size, dtype=np.uint64)
spin_dict["spin_start_sec"] = spin_start_sec
spin_dict["spin_start_subsec"] = np.full(spin_dict["spin_start_sec"].size, start_sclk_ticks * TICKS_TO_MS, dtype=np.uint64)
spin_dict["spin_period_sec"] = np.full(spin_dict["spin_start_sec"].size, 15.0, dtype=np.float64)
spin_dict["spin_period_valid"] = np.ones(spin_dict["spin_start_sec"].size, dtype=np.uint8)
spin_dict["spin_period_phase"] = np.ones(spin_dict["spin_start_sec"].size, dtype=np.uint8)
spin_dict["spin_period_source"] = np.zeros(spin_dict["spin_start_sec"].size, dtype=np.uint8)
spin_dict["thruster_firing"] = np.zeros(spin_dict["spin_start_sec"].size, dtype=np.uint8)

spin_start_float = np.arange(cov[0], cov[-1], 15)

# Set thruster firing flag. Intervals here indicate time during repointing
for i_pointing, interval in enumerate(cov[1:-1].reshape((-1, 2))):
    # set values where spin end time is greater than repoint start and spin start time is < repoint end
    firing_mask = np.logical_and(
        spin_start_float + spin_dict["spin_period_sec"] >= interval[0],
        spin_start_float < interval[1]
    )
    spin_dict["thruster_firing"][firing_mask] = 1
    spin_dict["spin_period_valid"][firing_mask] = 0

spin_df = pd.DataFrame.from_dict(spin_dict)
spin_df.to_csv("imap_2025_106_2026_105_v00.spin.csv", index=False)
```

tech3371:
# Change Summary

## Overview
Added function to create spin sample data for input date range. Created bare bone function to calculate spin phase and will extend it in future PR.

## Updated Files
   - function to calculate spin phase
   - Added fixture to generate spin sample data

## Testing
   - minimal tests


imap_processing/spice/geometry.py
```diff
@@ -87,3 +90,148 @@ def imap_state(
...
+    Get the spacecraft spin phase for the input query times.
+
+    Formula to calculate spin phase:
+        spin_phase = (
+            query_times - (spin_start_seconds + spin_start_subseconds)
+        ) / spin_period_sec
+
+    Parameters
+    ----------
+    query_times : float or np.ndarray[float]
```
tmplummer: Can we change this variable name to `query_mets` or just `met`. I think it is important to keep naming consistency across all functions.

imap_processing/spice/geometry.py
```diff
@@ -87,3 +90,148 @@ def imap_state(
...
+    -------
+    spin_phase : float or np.ndarray
+        Spin phase for the input query times.
+    """
+    # TODO: Write code to calculate spin phase for the input query times
+    if isinstance(query_times, float):
+        query_times = np.array([query_times])
+    # TODO: call get_spin_data function to get spin data for the input
+    # query times. And finish remaining implementation in upcoming PR.
+    return query_times
```
tmplummer: I think you should just return `np.zeros_like(query_times)` so the values make more sense.

imap_processing/spice/geometry.py
```diff
@@ -87,3 +90,148 @@ def imap_state(
...
+    ----------
+    start_met : float
+        Provide the start time in Mission Elapsed Time (MET).
+    end_met : float
+        Provide the end time in MET. If not provided, default to one day
+        from start time.
+
+    Returns
+    -------
+    spin_data : dict
```
tmplummer: Since the spin table is a CSV I would recommend returning a `pd.DataFrame` as I suspect we will read in the data using pandas.

tmplummer (commented):
This is looking good. I think that more can be filled in. If you have spin table data (whether it is from a file or generated), you should be able to calculate the spin phase for a set of times.

One piece that I think is missing is a function that takes a list of METs and returns tabular data from the spin table for each MET. Then the spin phase could be implemented.

Maybe this is growing into too big of a PR?

imap_processing/spice/geometry.py
tech3371: I had that originally as we discussed but then I was thinking about others who may want to use this function but don't want their data in pandas data form. Because of that, `dict` felt generic enough. But I can change it back to  return `pd.DataFrame` like we originally thought.

imap_processing/spice/geometry.py
tech3371: for sure!

imap_processing/spice/geometry.py
tech3371: ok

tech3371:
>If you have spin table data (whether it is from a file or generated), you should be able to calculate the spin phase for a set of times.
One piece that I think is missing is a function that takes a list of METs and returns tabular data from the spin table for each MET. Then the spin phase could be implemented.

@subagonsouth I do plan to tackle this in upcoming PR. I didn't want to make this PR too big. I want this PR to only generate spin data and then in upcoming PR, use that data to calculate spin phase which will include those that you are thinking. I also need to spend some time to figure out how to implement it. I will take your current ideas as I am thinking about it.

imap_processing/spice/geometry.py
```diff
@@ -87,3 +90,148 @@ def imap_state(
...
+
+    # Convert spin_start_sec to datetime to set repointing times flags
+    spin_start_dates = met_to_j2000ns(spin_start_sec)
+    spin_start_dates = cdflib.cdfepoch.to_datetime(spin_start_dates)
+
+    # Convert DatetimeIndex to Series for using .dt accessor
+    spin_start_dates_series = pd.Series(spin_start_dates)
+
+    # Find index of all timestamps that fall within 10 minutes after midnight
+    repointing_times = spin_start_dates_series[
```
laspsandoval: If we plan on using this test table w/ SPICE (and therefore use the test ck kernel IMAP_spacecraft_attitude.py) then here you might account for the test ck kernel's gaps that occur around midnight each day. I have been using the start/stop of the intervals within the kernel as the repointing times. I think you might eventually have to do that here. @subagonsouth what do you think?

laspsandoval (commented):
I want Tim's feedback on the repointing times. For now it is ok, but you might have to add the gaps in the test ck kernel in.

imap_processing/spice/geometry.py
tech3371: I am envisioning that in-situ will use this function but good point. I will need to think about how to alter `get_spacecraft_spin_phase` to use what you guys have vs this. I can think about this offline and that work will happen in upcoming PR.

imap_processing/spice/geometry.py
tech3371: I changed to return `pd.DataFrame` and if we decide to change down the road, we can do that.

imap_processing/tests/conftest.py
```diff
@@ -155,3 +159,79 @@ def _unset_metakernel_path(monkeypatch):
...
+        -------
+        spin_df : pd.DataFrame
+            Spin data.
+        """
+        if end_met is None:
+            # end_time is one day after start_time
+            end_met = start_met + 86400
+
+        # Create spin start second data of 15 seconds increment
+        spin_start_sec = np.arange(start_met, end_met, 15)
```
greglucas: Do we want end_met included?
```suggestion
        spin_start_sec = np.arange(start_met, end_met + 1, 15)
```

imap_processing/tests/conftest.py
```diff
@@ -155,3 +159,79 @@ def _unset_metakernel_path(monkeypatch):
...
+
+        spin_dict = {
+            "spin_number": np.arange(spin_start_sec.size, dtype=np.uint32),
```
```suggestion
        nspins = len(spin_start_sec)
            "spin_number": np.arange(nspins, dtype=np.uint32),
```

imap_processing/tests/conftest.py
```diff
@@ -155,3 +159,79 @@ def _unset_metakernel_path(monkeypatch):
...
+            "spin_start_sec": spin_start_sec,
+            "spin_start_subsec": np.full(spin_start_sec.size, 0, dtype=np.uint32),
```
```suggestion
            "spin_start_subsec": np.zeros(nspins, dtype=np.uint32),
```

imap_processing/tests/conftest.py
```diff
@@ -155,3 +159,79 @@ def _unset_metakernel_path(monkeypatch):
...
+            "spin_period_sec": np.full(spin_start_sec.size, 15.0, dtype=np.float32),
+            "spin_period_valid": np.ones(spin_start_sec.size, dtype=np.uint8),
+            "spin_phas_valid": np.ones(spin_start_sec.size, dtype=np.uint8),
```
```suggestion
            "spin_phase_valid": np.ones(spin_start_sec.size, dtype=np.uint8),
```

imap_processing/tests/conftest.py
```diff
@@ -155,3 +159,79 @@ def _unset_metakernel_path(monkeypatch):
...
+        spin_start_dates = cdflib.cdfepoch.to_datetime(spin_start_dates)
+
+        # Convert DatetimeIndex to Series for using .dt accessor
+        spin_start_dates_series = pd.Series(spin_start_dates)
+
+        # Find index of all timestamps that fall within 10 minutes after midnight
+        repointing_times = spin_start_dates_series[
+            (spin_start_dates_series.dt.time >= pd.Timestamp("00:00:00").time())
+            & (spin_start_dates_series.dt.time < pd.Timestamp("00:10:00").time())
+        ]
```
greglucas, line 347: Your "00" test will always be true, so you can either drop that one, or say +/- 10 minutes so you get before and after midnight.
```suggestion
        # Find index of all timestamps that fall within 10 minutes of midnight
        repointing_times = spin_start_dates_series[
            (spin_start_dates_series.dt.time > pd.Timestamp("23:50:00").time())
            & (spin_start_dates_series.dt.time < pd.Timestamp("00:10:00").time())
        ]
```

imap_processing/tests/conftest.py
```diff
@@ -155,3 +159,79 @@ def _unset_metakernel_path(monkeypatch):
...
+
+        repointing_times_index = repointing_times.index
+
+        # Use the repointing times to set thruster firing flag and spin period valid
+        spin_dict["thruster_firing"][repointing_times_index] = 1
+        spin_dict["spin_period_valid"][repointing_times_index] = 0
+        spin_dict["spin_phas_valid"][repointing_times_index] = 0
+
+        spin_df = pd.DataFrame.from_dict(spin_dict)
```
greglucas: I think you could make this directly above. You put `spin_dict = {}`, but you could just make that call: `spin_df = pd.DataFrame.from_dict({})` and start working with your dataframe right away. That might avoid the series creation and keep your indices all aligned.

imap_processing/tests/spice/test_geometry.py
```diff
@@ -23,3 +29,47 @@ def test_imap_state(et, use_test_metakernel):
...
+    """Test get_spacecraft_spin_phase() with generated spin data."""
+
+    start_time = 453051323.0
+
+    spin_phases = get_spacecraft_spin_phase(query_met_times=start_time)
+
+    # Uncomment this once spin phase calculation is implemented
+    # assert np.all(
+    #     (spin_phases >= 0) & (spin_phases <= 1)
+    # ), "Spin phases must be in [0, 1] range."
```
greglucas: Is this true? I would have expected spin phase to be in radians and thus a min/max around pi or 2pi.

imap_processing/tests/spice/test_geometry.py
```diff
@@ -23,3 +29,47 @@ def test_imap_state(et, use_test_metakernel):
...
+
+    # Ensure the length of spin phases matches the query times
+    assert len(spin_phases) == 1, "Spin phases length should match query times length."
+
+
+def test_get_spin_data(generate_spin_data, tmpdir):
+    """Test get_spin_data() with generated spin data."""
+
+    # SWE test data time minus 56120 seconds to get mid-night time
+    start_time = 453051323.0 - 56120
```
greglucas: Do you need to start at midnight? Or can you just let it go from wherever it was started and it will still be one-day just not centered on midnight.

imap_processing/tests/spice/test_geometry.py
```diff
@@ -23,3 +29,47 @@ def test_imap_state(et, use_test_metakernel):
...
+    spin_df = generate_spin_data(start_time)
+    spin_csv_file_path = Path(tmpdir) / "spin_data.spin.csv"
```
greglucas: I think tmpdir is already a Path?
```suggestion
    spin_csv_file_path = tmpdir / "spin_data.spin.csv"
```

imap_processing/tests/spice/test_geometry.py
```diff
@@ -23,3 +29,47 @@ def test_imap_state(et, use_test_metakernel):
...
+    spin_df.to_csv(spin_csv_file_path, index=False)
+    os.environ["SPIN_DATA_FILEPATH"] = str(spin_csv_file_path)
```
greglucas: This should be monkeypatched, otherwise it might stick around for future tests. See the `conftest` file for other environment monkeypatches.

imap_processing/tests/spice/test_geometry.py
```diff
@@ -23,3 +29,47 @@ def test_imap_state(et, use_test_metakernel):
...
+
+    spin_data = get_spin_data()
+
+    assert len(spin_data.keys()) == 8, "Spin data must have 8 fields."
```
greglucas: Which 8 fields? Can we assert against those exactly instead of just lengths. Edit: You are doing that below already on the columns, so this can be removed I think.

imap_processing/tests/spice/test_geometry.py
```diff
@@ -23,3 +29,47 @@ def test_imap_state(et, use_test_metakernel):
...
+    assert len(spin_data) == 5760, "One day should have 5,760 records of 15 seconds."
```
greglucas: Specific times or anything else associated with it?
```python
np.testing.assert_array_equal(spin_data["spin_number"], np.arange(5760))
```

imap_processing/spice/geometry.py
```diff
@@ -87,3 +90,69 @@ def imap_state(
...
+    """
+    spin_data_filepath = os.getenv("SPIN_DATA_FILEPATH")
+    if spin_data_filepath is not None:
+        path_to_spin_file = Path(spin_data_filepath)
+    else:
+        # Handle the case where the environment variable is not set
+        raise ValueError("SPIN_DATA_FILEPATH environment variable is not set.")
+
+    if path_to_spin_file.suffix != ".csv":
+        raise ValueError("Input file must be a CSV file.")
```
greglucas: read_csv will check the validity of the csv file for us.
```suggestion
```

greglucas (commented):
I think this seems like a good approach. Loading a spin file from somewhere on the disk into a pandas dataframe. In the tests, you're making it for the user similar to the SPICE metakernel. In processing we'll set this environment variable to wherever the actual data is.

Something to think about is that we are adding a lot of environment variables now, is this going to be in a separate location from `IMAP_DATA_DIR`, or is it going to be relative to that and we can leverage that environment variable that is already there?

imap_processing/tests/conftest.py
tech3371: I thought that was typo too but that's what we had in the document and sample.

imap_processing/tests/conftest.py
tech3371: I am assuming nominal repointing time is 10 minutes. With this, it will become 20 minutes. Is that right? Did you want this to five minutes before and after instead.

imap_processing/tests/spice/test_geometry.py
tech3371: I don't know this for sure. Let me remove this until we know.

imap_processing/tests/spice/test_geometry.py
tech3371: I just want to make sure that I am producing correct len and therefore, I used mid-night to mid-night. It doesn't have to be midnight.

imap_processing/tests/spice/test_geometry.py
tech3371: Without this, the call `.suffix` did not work but since I removed that from your above feedback, I can revert this back.

tech3371:
>Something to think about is that we are adding a lot of environment variables now, is this going to be in a separate location from IMAP_DATA_DIR, or is it going to be relative to that and we can leverage that environment variable that is already there?

@greglucas, @laspsandoval and @subagonsouth was thinking in similar manner. This will depend on where we decide to store spin data and how we get it. But I think having some fixed path for spin data in `imap-data-access` repo will be good if we decide to keep spin data in csv file and use download API to get it. Something to revisit in the near future.

imap_processing/tests/conftest.py
tech3371: I updated Galaxy page and in here as well. I will let Ian Murphy know of this change.

imap_processing/spice/geometry.py
```diff
@@ -87,3 +90,67 @@ def imap_state(
...
+    -------
+    spin_phase : float or np.ndarray
+        Spin phase for the input query times.
+    """
+    # TODO: Write code to calculate spin phase for the input query times
+    if isinstance(query_met_times, float):
+        query_met_times = np.array([query_met_times])
+    # TODO: call get_spin_data function to get spin data for the input
+    # query times. And finish remaining implementation in upcoming PR.
+    return np.zeros_like(query_met_times)
```
tmplummer: My last nit pick is that the type hint seems to indicate that if `query_met_times` is a float, then this function should return a float, not an array.

tmplummer (approved):
This looks good to go once my one nit-pick is addressed. Thanks for your persistence on this PR!

imap_processing/spice/geometry.py
tech3371: Where do you see this type hint? I am trying to find it but don't see it on my end.

tech3371: got it. I was not paying close attention to this function. I should have return float for float input value instead of np.array.

tmplummer: Sorry, I wasn't very clear. I think that if `get_spacecraft_spin_phase` gets a float for `query_met_times` that it is meant to return a float based on the function defined typing and the Returns doc string.

Issue #769, opened 2024-08-21
CoDICE: Add capability to decommutate multiple APIDs and multiple packets in a singe input file


bourque:
Currently, the CoDICE L1 pipeline assumes that input L0 data has one APID per file. However, we expect to receive L0 files from the POC that have multiple APIDs per file. The pipeline should be improved to support this.

PR #826, opened 2024-09-11
CoDICE data product organization refactor

bourque:
Apologies for the large PR here, I couldn't see a great way to split this up into smaller PRs. This PR is a somewhat large refactor of the CoDICE L1a processing pipeline in order to allow better flexibility in defining and processing the 18 different CoDICE L1a data products.  Before, I was treating all CoDICE-lo products and all CoDICE-hi products as if they could be built the same way, but I have found that there are enough subtle differences between all of these products that they should be built in a more individualized way. As such, I have introduced a few new 'data product configuration' variables (namely CDF coordinates, CDF dimensions, and CDF data variables) that can be defined on the per-product level. In doing so, I have created a few more focused methods and renamed a few things where I think it made sense to.


imap_processing/codice/codice_l1a.py
```diff
@@ -31,11 +31,8 @@
 logger.setLevel(logging.INFO)
 
 # TODO: Add support for decomming multiple APIDs from a single file
-# TODO: Add these as variables in CDF: SPIN_PERIOD, ST_BIAS_GAIN_MODE,
```
bourque, line 34: I moved this TODO to the parts of the code where they will be implemented

imap_processing/codice/codice_l1a.py
```diff
@@ -31,11 +31,8 @@
...
-#       SW_BIAS_GAIN_MODE, RGFO_HALF_SPIN, NSO_HALF_SPIN, DATA_QUALITY
 # TODO: Determine what should go in event data CDF and how it should be
 #       structured.
-# TODO: Make sure CDF attributes match expected nomenclature
```
bourque, line 38: This was taken care of in #774, I just forgot to remove this TODO

imap_processing/codice/codice_l1a.py
```diff
@@ -79,143 +84,143 @@ def __init__(self, table_id: int, plan_id: int, plan_step: int, view_id: int):
...
-                    (
-                        1,
-                        self.num_energy_steps,
-                        self.num_positions,
-                        self.num_spin_sectors,
-                    )
-                )
-                dims = ["epoch", "esa_step", "inst_az", "spin_sector"]
+            # Reshape to 4 dimensions to allow for epoch dimension
+            reshaped_variable_data = np.expand_dims(variable_data, axis=0)
```
greglucas, line 161: This doesn't seem quite right. This is adding an extra empty dimension, shouldn't we be reshaping the data you have into the proper dimensions? (Similar to what you had before I think)

Will this only ever have one epoch variable, or could there be more than that?

```python
variable_data.reshape(
                    (
                        len(dataset["epoch"],
                        self.num_energy_steps,
                        self.num_positions,
                        self.num_spin_sectors,
                    )
```

imap_processing/codice/codice_l1a.py
```diff
@@ -288,33 +305,23 @@ def get_energy_table(self) -> None:
 
         # Get the appropriate values
         sweep_table = sweep_data[sweep_data["table_idx"] == sweep_table_id]
-        self.energy_table = sweep_table["esa_v"].values
+        energy_table: list[float] = sweep_table["esa_v"].values
```
greglucas: Aren't these numpy arrays? I wonder if we need to bring in typechecking stubs from pandas and/or xarray to help with this... (For another time)
```suggestion
        energy_table: NDArray[float] = sweep_table["esa_v"].values
```

imap_processing/codice/codice_l1a.py
```diff
@@ -335,6 +342,42 @@ def unpack_science_data(self, science_values: str) -> None:
...
+        packet : xarray.Dataset
+            A packet for the APID of interest.
+        data_version : str
+            Version of the data product being created.
+        """
+        self.packet_dataset = packet
+
+        # Gather and set various configurations of the data product
+        config = constants.DATA_PRODUCT_CONFIGURATIONS.get(apid)  # type: ignore[call-overload]
+        self.coords_to_include = config["coords"]
```
greglucas: Would it make sense to set a `self.config = config`, and then lookup these values later since they are all named identically. `self.config["dataset_name"]` rather than `self.dataset_name` doesn't seem too bad.

imap_processing/codice/constants.py
greglucas, line 1: This is definitely a lot of data product definitions, nice work organizing it all :) This seems like something that would be nice to have within the yaml metadata files, but I don't know if that handles all of this extra stuff. I don't know how others are treating it, maybe it is easier because they have more like 1 or 2 data products instead of 20 😂

imap_processing/codice/constants.py
```diff
@@ -78,112 +78,271 @@
...
 DATA_PRODUCT_CONFIGURATIONS = {
     CODICEAPID.COD_HI_INST_COUNTS_AGGREGATED: {
+        "coords": [
+            "epoch",
+            "inst_az",
+            "spin_sector",
+            "esa_step",
+            "energy_label",
+        ],  # TODO: These will likely change
+        "dataset_name": "imap_codice_l1a_hi_counters_aggregated",
```
tech3371: This seems to follow filename convention. If so, can you change this and others to use `-` instead of `_` for the descriptor names?
```suggestion
        "dataset_name": "imap_codice_l1a_hi_counters-aggregated",
```

imap_processing/codice/constants.py
```diff
@@ -78,112 +78,271 @@
...
+            "esa_step",
+            "inst_az",
+            "spin_sector",
+        ],  # TODO: These will likely change
+        "instrument": "hi",
         "num_counters": 8,
         "num_energy_steps": 15,  # TODO: Double check with Joey
         "num_positions": 4,  # TODO: Double check with Joey
         "num_spin_sectors": 1,
+        "support_variables": ["data_quality", "spin_period"],
```
tech3371, line 152: Does this mean that there are these two extra data variable stored in CDF file and others only has science data stored?

imap_processing/codice/codice_l1a.py
```diff
@@ -509,9 +552,11 @@ def process_codice_l1a(file_path: Path, data_version: str) -> xr.Dataset:
```
tech3371, line 547: I just noticed that you are only processing first packet from `packet_dataset`. Is that right?

imap_processing/codice/codice_l1a.py
```diff
@@ -79,143 +84,143 @@ def __init__(self, table_id: int, plan_id: int, plan_step: int, view_id: int):
...
+            ``xarray`` dataset for the data product.
+
+        Returns
+        -------
+        dataset : xarray.Dataset
+            ``xarray`` dataset for the data product, with added support variables.
+        """
+        for variable_name in self.support_variables:
+            if variable_name == "energy_table":
+                variable_data = self.get_energy_table()
```
tech3371, line 201: Is this function getting energy value in engineering units or is this getting energy step value, Eg. [0,1,2,....127]?

imap_processing/codice/codice_l1a.py
```diff
@@ -79,143 +84,143 @@ def __init__(self, table_id: int, plan_id: int, plan_step: int, view_id: int):
...
 
             # Create the CDF data variable
             dataset[variable_name] = xr.DataArray(
-                variable_data_arr,
+                reshaped_variable_data,
                 name=variable_name,
-                dims=dims,
+                dims=self.dims,
                 attrs=attrs,
             )
```
tech3371: I was following pretty good up until this. I am not following how `self.data` is associated with `self.variable_names`.

tech3371 (commented):
This was a ton of good work. I followed almost of the change except the last one. Complicated but logic flows well IMO.

imap_processing/codice/constants.py
bourque: Yes, exactly!

imap_processing/codice/codice_l1a.py
bourque: `self.data` is a list of multi-dimensional numpy arrays, with each element of the list being the data for a given CDF variable. THe `self.variable_names` provides the list of CDF variable names (e.g. `h`, `he`, `cno`, etc.) to associate with each element of `self.data`.

This has got me realizing that I should change some of the variable names I am using in `constants.py` from something like `LO_SW_ANGULAR_NAMES` to `LO_SW_ANGULAR_VARIABLE_NAMES` to be more explicit.

imap_processing/codice/codice_l1a.py
bourque: It is getting the ESA sweep values technically, which I am calling the "energy table". This is what it is ultimately reading in: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/codice/data/esa_sweep_values.csv

imap_processing/codice/codice_l1a.py
bourque: Yes, good catch!

imap_processing/codice/codice_l1a.py
bourque: Yeah, that makes sense, especially now that I have a whole bunch of these parameters

imap_processing/codice/constants.py
bourque: Good idea to be consistent. I think I went down the road of using underscored because of how I defined things in the CDF attr config YAML files. Since the `dataset_name` is being used to find the appropriate key in those files, I will also need to update from underscore to hyphen there too.

Just pointing this out so that it is clear why those changes will be in the next commit to this PR.

imap_processing/codice/codice_l1a.py
bourque: I am quite confused here. Since `packet_file_to_datasets()` function (which is used within my `decom_packets()` function above) packages packets into distinct datasets, one dataset per APID, wouldn't I always expect just one packet within a particular `datasets[apid]`?

bourque: In general, I am still confused about what to expect in a typical CoDICE L0 file when we are in-flight.

My understanding is that I will receive a `imap_codice_l0_raw_YYYYMMDD_vXXX.pkts` file that will contain multiple APIDs (say, for example, `1136` (hskp) + `1155` (sw-priority) + `1160` (nsw-priority)), which will then be rearranged into:

```python
{1136 : <xr.Dataset for 1136>, 1155: <xr.Dataset for 1155>, 1160: <xr.Dataset for 1160>}
```

Where `<xr.Dataset for 1155>` for example, contains one packet with APID `1155` which has one `epoch` associated with it.

It confuses me how I could get multiple `epochs` for a single APID, as @greglucas eludes to in one of his comments above.

Am I misinterpreting anything here?

tech3371: You got it right up until this sentence `Where <xr.Dataset for 1155> for example, contains one packet with APID 1155 which has one epoch associated with it.`

`packet_file_to_datasets()` will unpack all packets you get in `imap_codice_l0_raw_YYYYMMDD_vXXX.pkts`. Let's say that your daily file has 100 packets for this particular 1155 apid, it will decom all those 100 packets per XTCE definition and store it as `xr.Dataset` where there is one variable in `xr.Dataset` for each field the XTCE definition. Therefore, all of your variable in the `xr.Dataset` has 100 element array. Therefore, your `epochs` has 100 epoch as well. I hope that helps. I am happy to jump on call if it helps.

bourque: Ah ok, thanks for clarifying! So I was correct in thinking one packet = one epoch, but I didn't realize that the `xr.Dataset` could have many packets. I hadn't encountered that before since the test data I am using is just a single file and a single packet.

My code will require some changes to support multiple packets, but I will take care of that in a future PR that addresses #769

imap_processing/codice/codice_l1a.py
bourque: Now that I understand this better (thank you @tech3371 for clarifying in the other comment!), you are right -- this will need to change.

I plan to make the code changes for this in a future PR that addresses #769

tech3371 (approved):
Since things I am concerned about will be addressed in future PR, it looks good! Thank you for refactoring. CoDICE is complicated for sure.

Issue #822, opened 2024-09-09
HIT: Refactor hit_l1a.py to use packet_file_to_datasets function


vmartinez-cu:
Refactor hit_l1a.py to use `packet_file_to_datasets` function to maintain consistency across instruments. Update unit tests accordingly

PR #828, opened 2024-09-11
Refactor hit_l1a.py to use packet_file_to_datasets function

vmartinez-cu:
Updated hit_l1a.py to use the packet_file_to_datasets function instead of using data classes and manually creating xarray datasets for products. Added some new functions to handle housekeeping data processing.

* uses packet_file_to_datasets function in hit_l1a.py
* New functions for processing housekeeping dataset
* New function to process science data (WIP)
* Updated unit tests

## New files
- hit_packet_definitions.xml
   - definitions for all APIDs. Replaced packet definition files per APID.

## Updated Files
- hit_l1a.py
   - uses packet_file_to_datasets function in hit_l1a.py
   - New functions for processing housekeeping dataset
   - New function to process science data (WIP)
- test_hit_l1a.py
   - updated unit tests to reflect changes in hit_l1a.py

## Deleted files
- test_housekeeping.py
   - tests the housekeeping data class. This class will be removed after hit_l1b is refactored.
- P_HIT_HSKP.xml
   - packet definition for housekeeping. Replaced by hit_packet_definitions.xml
- P_HIT_SCIENCE.xml
   - packet definition for science data. Replaced by hit_packet_definitions.xml

## Testing
- test_hit_l1a.py
   - added new fixtures and additional assertions for housekeeping data

Issue #822


imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
-    grouped_data = group_data(sorted_packets)
-
-    # create the attribute manager for this data level
+    # TODO add logging
+
+    # Unpack ccsds file
+    packet_definition = (
+        imap_module_directory / "hit/packet_definitions/" "hit_packet_definitions.xml"
+    )
+    datasets_by_apid = packet_file_to_datasets(
```
sdhoyt, line 59: the `use_dervied_values` flag needs to be set to `False` for L1A

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
     attr_mgr = ImapCdfAttributes()
     attr_mgr.add_instrument_global_attrs(instrument="hit")
     attr_mgr.add_instrument_variable_attrs(instrument="hit", level="l1a")
     attr_mgr.add_global_attribute("Data_version", data_version)
 
-    # Create datasets
-    datasets = create_datasets(grouped_data, attr_mgr)
+    # Process science to l1a.
+    processed_data = []
+    for apid in datasets_by_apid:
```
sdhoyt: Up to you if you want to change this, but you could avoid the loop by doing:

```python
if HitAPID.HIT_HSKP in datasets_by_apids:
     datasets_by_apid[HITAPID.HIT_HSKP] = process_housekeeping(
                 datasets_by_apid[HITAPID.HIT_HSKP], attr_mgr
             )
if HitAPID.HIT_SCIENCE in datasets_by_apids:
    print("Skipping science data for now")
```
I think our XTCE should stop us from ever seeing an unknown APID, but you could add another if to check for ones that don't match the APIDS contained in the APID Enums list

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
-    xtce_file = imap_module_directory / "hit/packet_definitions/P_HIT_HSKP.xml"
-    logger.debug(f"Unpacking {packet_file} using xtce definitions in {xtce_file}")
-    unpacked_packets: list = decom.decom_packets(packet_file, xtce_file)
-    logger.debug(f"{packet_file} unpacked")
-    return unpacked_packets
+    # Stack 64 leak variables (leak_00, leak_01, ..., leak_63)
+    leak_vars = [dataset[f"leak_i_{i:02d}"] for i in range(64)]
 
+    # Concatenate along 'leak_index' and reorder dimensions
+    stacked_leaks = xr.concat(leak_vars, dim="leak_index").transpose(
```
sdhoyt: should the dim here be adc_channels?

imap_processing/tests/hit/test_hit_l1a.py
```diff
@@ -1,99 +1,201 @@
...
+    assert len(processed_datasets) == 1
+    assert isinstance(processed_datasets[0], xr.Dataset)
+    assert processed_datasets[0].attrs["Logical_source"] == "imap_hit_l1a_hk"
+
+
+def test_concatenate_leak_variables(housekeeping_dataset):
+    """Test concatenation of leak_i variables"""
+
+    # Call the function
+    updated_dataset = concatenate_leak_variables(housekeeping_dataset)
```
sdhoyt: I think it'd be good to check that the values are correct. Maybe before you call this function, remove everything but the first couple packets for housekeeping and hardcode arrays for all the Leak values to check against

imap_processing/tests/hit/test_housekeeping.py
```diff
@@ -1,91 +0,0 @@
-import numpy as np
```
sdhoyt, line 1: You can also remove the Housekeeping data class

sdhoyt (commented):
Great work on this! Just a few comments

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
-    packets = decom_packets(packet_file)
-    sorted_packets = utils.sort_by_time(packets, "SHCOARSE")
-    grouped_data = group_data(sorted_packets)
-
-    # create the attribute manager for this data level
+    # TODO add logging
+
+    # Unpack ccsds file
+    packet_definition = (
+        imap_module_directory / "hit/packet_definitions/" "hit_packet_definitions.xml"
```
greglucas:
```suggestion
        imap_module_directory / "hit/packet_definitions/hit_packet_definitions.xml"
```

imap_processing/hit/l1a/hit_l1a.py
greglucas: :( Yeah, I should just set this as the default already!

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
-    logger.debug(f"{packet_file} unpacked")
-    return unpacked_packets
+    # Stack 64 leak variables (leak_00, leak_01, ..., leak_63)
+    leak_vars = [dataset[f"leak_i_{i:02d}"] for i in range(64)]
 
+    # Concatenate along 'leak_index' and reorder dimensions
+    stacked_leaks = xr.concat(leak_vars, dim="leak_index").transpose(
+        "epoch", "leak_index"
+    )
+    dataset["leak_i"] = stacked_leaks
```
greglucas, line 116: Do you want to drop the "i" here and just call it "leak"?

imap_processing/hit/l1a/hit_l1a.py
greglucas: Really nice use of the xarray functions to do this for you :)

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
-
-    skip_keys = [
-        "shcoarse",
-        "ground_sw_version",
-        "packet_file_name",
-        "ccsds_header",
-        "leak_i_raw",
+    logger.info("Creating HIT L1A housekeeping dataset")
+
+    logical_source = "imap_hit_l1a_hk"
```
greglucas, line 190: Should this be in the CDF metadata?

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
-                "adc_channels": adc_channels,
-                "adc_channels_label": adc_channels_label,
-            },
-            attrs=attr_mgr.get_global_attributes(logical_source),
+    # Drop variables not needed for CDF
+    dataset = dataset.drop_vars(drop_keys)
+
+    # Create data arrays for dependencies
+    adc_channels = xr.DataArray(
+        np.arange(64, dtype=np.uint16),
```
greglucas:
```suggestion
        np.arange(64, dtype=np.uint8),
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
+        name="adc_channels_label",
+        dims=["adc_channels_label"],
+        attrs=attr_mgr.get_variable_attributes("adc_channels_label"),
+    )
+
+    # Update dataset coordinates and attributes
+    dataset = dataset.assign_coords(
+        {
+            "adc_channels": adc_channels,
+            "adc_channels_label": adc_channels_label,
```
greglucas, line 232: I forget, do the "labels" need to be defined as coordinates as well, or are those variables that are dependent on the coordinates?

i.e. should `adc_channels_label` actually be in the `data_vars` section?

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -50,203 +47,207 @@ def hit_l1a(packet_file: str, data_version: str) -> list[xr.Dataset]:
...
+        }
+    )
+    dataset.attrs = attr_mgr.get_global_attributes(logical_source)
+
+    # Stack 64 leak variables (leak_00, leak_01, ..., leak_63)
+    dataset = concatenate_leak_variables(dataset)
+
+    # Assign attributes and dimensions to each data array in the Dataset
+    for field, data in dataset.data_vars.items():
```
greglucas: I think this will basically be re-creating all of your `DataArray`s and resetting the dataset.

Instead could you loop through just the field values and update the current dataset variable?

```python
dataset[field].attrs = attr_mgr.get_variable_attributes(field)
# Something like this where you can just set the name of existing coords
dataset[field].assign_coords(dims)
```

imap_processing/tests/hit/test_hit_l1a.py
```diff
@@ -1,99 +1,201 @@
...
-    unpacked_packets : list[space_packet_parser.parser.Packet]
-        A sorted list of decommutated packets
-    """
-    grouped_data = hit_l1a.group_data(unpacked_packets)
-    assert len(grouped_data.keys()) == 1
-    assert all([isinstance(i, Housekeeping) for i in grouped_data[1251]])
+def datasets(packet_filepath):
+    """Create datasets from packet file"""
+    packet_definition = (
+        imap_module_directory / "hit/packet_definitions/" "hit_packet_definitions.xml"
```
greglucas, line 28:
```suggestion
        imap_module_directory / "hit/packet_definitions/hit_packet_definitions.xml"
```

imap_processing/tests/hit/test_hit_l1a.py
```diff
@@ -1,99 +1,201 @@
...
+    }
+
+    # Define the coordinates and dimensions. Both have equivalent values
+    dataset_coords_dims = {"epoch", "adc_channels", "adc_channels_label"}
+
+    # Assertions
+    # ----------------
+    # Check that the dataset has the correct variables
+    assert valid_keys == set(processed_hskp_dataset.data_vars.keys())
+    assert set(dropped_keys).isdisjoint(set(processed_hskp_dataset.data_vars.keys()))
```
greglucas, line 220: Nice, I did not know the `isdisjoint` method existed, great use of it!

greglucas (commented):
Looks great to me! Just a few minor suggestions.

imap_processing/hit/l1a/hit_l1a.py
sdhoyt: @greglucas The variable is for "leakage current", so `i` here represents "current", not an index, so I think it's useful in the variable name. Maybe this should be a capital `I` and a comment somewhere so that's clear.

greglucas: Ahh, that makes sense once you know it :) I think it is fine to keep it as is, This is just housekeeping and they can do what they want with it. Thanks for clarifying!

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: This is a better approach. Thanks for the suggestion! If we won't ever see an unknown APID, I won't check for this then.

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Correct, adc_channels is the dim for the leak variable. Dims will be updated later in the process when dims for all the fields get assigned according to the cdf yaml file definitions. I just needed to put a dimension here that made sense as a placeholder.

Assign attributes and dimensions to each data array in the Dataset

```
    for field, data in dataset.data_vars.items():
        # Create a list of dimensions using the DEPEND_I keys in the
        # attributes
        dims = [
            value
            for key, value in attr_mgr.get_variable_attributes(field).items()
            if "DEPEND" in key
        ]
        dataset[field] = xr.DataArray(
            data,
            dims=dims,
            attrs=attr_mgr.get_variable_attributes(field),
        )
```

imap_processing/tests/hit/test_housekeeping.py
vmartinez-cu: I did initially, but it broke my hit l1b tests for housekeeping so to avoid putting those code updates in this PR, I just left the data classes for now.

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: The logical source variable is used to grab the HIT L1A instrument level attributes from imap_hit_global_cdf_attrs.yaml when updating the dataset attributes `dataset.attrs = attr_mgr.get_global_attributes(logical_source)` 

Attributes from imap_hit_global_cdf_attrs.yaml for L1A housekeeping:

```
imap_hit_l1a_hk:
    <<: *instrument_base
    Data_level: 1A
    Data_type: L1A_HK>Level-1A Housekeeping
    Logical_source: imap_hit_l1a_hk
    Logical_source_description: IMAP Mission HIT Instrument Level-1A Housekeeping Data.
```


Would it help to rename this variable or plug in the value directly into the line that updates the dataset attributes? `dataset.attrs = attr_mgr.get_global_attributes(imap_hit_l1a_hk)`

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: I think it needs to be added to the coordinates as well because I initially didn't have it in there but then had to add it. I don't remember why though. I'll get clarification on this.

vmartinez-cu: I checked with Tenzin and she explained that cdflib updates requires this for data with greater than 1 dimensions

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Good catch. Updated!

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Tenzin explained that it’s because cdflib updates requires labels in the coordinates for data with dimensions greater than 1

imap_processing/tests/hit/test_hit_l1a.py
vmartinez-cu: Added!

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: I updated the function to pass in adc_channels and assign it as a dimension

sdhoyt (approved):
I recommend running the cli.py as an extra test if you haven't done that already. But I think everything looks good!

PR #829, opened 2024-09-12
Ultra l1c bin data

laspsandoval:
# Change Summary

## Overview
Add spatial and energy binning equations.

## Updated Files
- ultra_l1c_pset_bins.py
   - Added spatial and energy binning mechanisms.

## Testing
- test_ultra_l1c_pset_bins.py
   - Tests ultra_l1c_pset_bins.py

### DIFF
--- a/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
+++ b/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
@@ -26,10 +26,12 @@ def build_energy_bins() -> tuple[np.ndarray, np.ndarray]:
     energy_bin_start = bin_edges[:-1]
     energy_bin_end = bin_edges[1:]
 
-    return energy_bin_start, energy_bin_end
+    return energy_bin_start, energy_bin_end, az_bin_midpoints, el_bin_midpoints
 
 
-def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
+def build_spatial_bins(
+    spacing: float = 0.5,
+) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
     """
     Build spatial bin boundaries for azimuth and elevation.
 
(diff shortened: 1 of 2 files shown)


laspsandoval:
Hold off on reviewing for a couple of hours.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +46,131 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
 
     # Elevation bins from -90 to 90 degrees.
     el_bin_edges = np.arange(-90, 90 + spacing, spacing)
+    el_bin_midpoints = el_bin_edges[:-1] + spacing / 2  # Midpoints between edges
 
-    return az_bin_edges, el_bin_edges
+    return az_bin_edges, el_bin_edges, az_bin_midpoints, el_bin_midpoints
+
+
+def cartesian_to_spherical(
```
tmplummer, line 65: This seems like a common function that everyone may need. I will comment that I would think a more canonical input signature would be to accept a single 3 element vector or an array-like set of 3 element vectors. Cartesian to spherical should really output the radius as well.

There are SPICE functions for these common conversions:
https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/recazl_c.html
https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/reclat_c.html
https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/recrad_c.html
https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/recsph_c.html
Though I am not against the idea that implementing them in vectorized numpy may be faster than wrapping the SPICE functions in a loop.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +46,131 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+    # Azimuth angle (angle in the xy-plane, range: [0, 2*pi])
+    az = np.arctan2(vhat_y, vhat_x)
+
+    # Ensure azimuth is from 0 to 2PI
+    az = az % (2 * np.pi)
+
+    return az, el
+
+
+def bin_space(
```
tmplummer, line 111: You might consider checking out the 2d histogram function. I believe it would be ideal for this binning:
https://numpy.org/doc/stable/reference/generated/numpy.histogram2d.html

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +46,131 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+
+    Returns
+    -------
+    energy_start : np.ndarray
+        Start of energy bin.
+    """
+    energy_bin_edges = build_energy_bins()
+
+    # Find the appropriate bin index.
+    energy_bin_idx = np.searchsorted(energy_bin_edges, energy, side="right") - 1
```
tmplummer, line 172: Any reason not to use `np.histogram` here? https://numpy.org/doc/stable/reference/generated/numpy.histogram.html#numpy.histogram

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +46,131 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+    vx : np.ndarray
+        The x-components of the velocity vector.
+    vy : np.ndarray
+        The y-components of the velocity vector.
+    vz : np.ndarray
+        The z-components of the velocity vector.
+
+    Returns
+    -------
+    az_midpoint : np.ndarray
```
tmplummer: I guess that I am missing something. Is there a reason that this function doesn't just produce the 2d histogram data?

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +46,131 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+
+    Parameters
+    ----------
+    energy : np.ndarray
+        Particle energy.
+
+    Returns
+    -------
+    energy_start : np.ndarray
+        Start of energy bin.
```
tmplummer, line 167: Same question here... When I think of a function that "bins" data, it returns a histogram rather than the bin center location for each input.

tmplummer (commented):
I have a few comments about implementation things and recommendations on functions to look at to see if they might help.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
laspsandoval: Nice ideas. I made changes except for the use of SPICE.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
laspsandoval: I made some changes, but the reason I can't do a 2d histogram is that it would result in 259200 bins since the Ultra has 0.5 x 0.5 degree bins. I want my xarray to be like this:

coordinate: spatial_bin_id
coordinate: energy_bin_id
variable (spatial_bin_id): az_midpoint
variable (spatial_bin_id): el_midpoint
variable (energy_bin_id): energy_edge_start
variable (spatial_bin_id, energy_bin_id): counts

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
laspsandoval: Let me know what you think regarding comment below.

PR #847, opened 2024-09-16
Added SWAPI L3 levels to the list of valid levels

tech3371:
# Change Summary

## Overview
small update to add SWAPI L3 levels


maxinelasp (approved):
Does this need to be updated in imap-data-access? Good catch!

tech3371:
> Does this need to be updated in imap-data-access? Good catch!

@maxinelasp I looked on `imap-data-access` before but looks like we don't need to update there. There, we only stored all valid levels whereas here, we go more into instrument specific valid levels.

PR #848, opened 2024-09-16
FIX: Update SWAPI quality flags to work with numpy 2+

greglucas:
# Change Summary

## Overview

SWAPI tests were failing locally for me because I had numpy2+ installed. It looks like we need to bitwise or the actual Enum value rather than the Enum itself. It might be worth upstreamining this issue to Numpy, but this fixes the issue here for now.

Additionally, regenerate the `poetry.lock` file from the pre-commit rather than the local environment. @subagonsouth as an FYI, it is good to let pre-commit take care of the updates to the lock file because otherwise it is constantly getting changed by people with different versions of poetry it seems like.


tech3371 (approved):
Thank you!

tmplummer (approved):

PR #851, opened 2024-09-18
Ultra l1b species + start to l1b Annotated

laspsandoval:
# Change Summary

## Overview
I added calculations for corrected time of flight (tof) and a placeholder for looking up the species. I also added a start to getting the annotated events.

## New Files
- ultra_l1b_annotated.py
   - Start to using SPICE to get particle velocities in different reference frames.

## Updated Files
- ultra_l1b_extended.py
   - Calculates corrected tof and has a placeholder for species lookup table.

## Testing
- test_ultra_l1b_annotated.py
- test_ultra_l1b_extended.py

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
@@ -37,7 +37,7 @@ def test_get_particle_velocity(kernels):
     spice.furnsh(kernels)
     import pickle
 
-    with open("../tests/test_data/spice/spice_examples/direct_events.pkl", "rb") as file:
+    with open("/Users/lasa6858/Desktop/directEvents.pkl", "rb") as file:
         directEvents = pickle.load(file)
 
     time = directEvents['tdb']
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
@@ -43,6 +43,6 @@ def test_get_particle_velocity(kernels):
     time = directEvents['tdb']
     ultra_velocity = directEvents["vultra"]
 
-    velocity = get_particle_velocity(time, ultra_velocity)
+    velocities = get_particle_velocity(time, ultra_velocity)
 
     print("hi")
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
@@ -4,6 +4,7 @@ import numpy as np
 import pytest
 import spiceypy as spice
 
+from imap_processing.spice.geometry import SpiceFrame
 from imap_processing.ultra.l1b.ultra_l1b_annotated import get_particle_velocity
 
 
@@ -27,10 +28,9 @@ def test_get_particle_velocity(spice_test_data_path, kernels):
     """Tests get_particle_velocity function."""
     spice.furnsh(kernels)
 
-    id_imap_dps = spice.gipool("FRAME_IMAP_DPS", 0, 1)
     pointing_cover = spice.ckcov(
         str(spice_test_data_path / "sim_1yr_imap_pointing_frame.bc"),
-        int(id_imap_dps),
+        SpiceFrame.IMAP_DPS.value,
         True,
         "SEGMENT",
         0,
@@ -43,8 +43,12 @@ def test_get_particle_velocity(spice_test_data_path, kernels):
 
     instrument_velocity = np.array([[41.18609, -471.24467, -832.8784]])
 
-    velocity_45 = get_particle_velocity(times, instrument_velocity, "IMAP_ULTRA_45")
-    velocity_90 = get_particle_velocity(times, instrument_velocity, "IMAP_ULTRA_90")
+    velocity_45 = get_particle_velocity(
+        times, instrument_velocity, SpiceFrame.IMAP_ULTRA_45, SpiceFrame.IMAP_DPS
+    )
+    velocity_90 = get_particle_velocity(
+        times, instrument_velocity, SpiceFrame.IMAP_ULTRA_90, SpiceFrame.IMAP_DPS
+    )
 
     # Compute the magnitude of the velocity vectors in both frames
     magnitude_45 = np.linalg.norm(velocity_45[0])
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
@@ -40,8 +40,8 @@ def test_get_particle_velocity(kernels):
     with open("/Users/lasa6858/Desktop/directEvents.pkl", "rb") as file:
         directEvents = pickle.load(file)
 
-    time = directEvents['tdb']
-    ultra_velocity = directEvents["vultra"]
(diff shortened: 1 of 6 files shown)


imap_processing/ultra/l1b/ultra_l1b_annotated.py
```diff
@@ -0,0 +1,50 @@
...
+import numpy as np
+import spiceypy as spice
+from numpy.typing import NDArray
+
+from imap_processing.spice.kernels import ensure_spice
+
+
+@ensure_spice
+@typing.no_type_check
+def get_particle_velocity(
```
tmplummer, line 15: I'm think that this function may be useful for other instruments. Would you consider moving this into the `spice/geometry.py` module, renaming it to reflect that it is really just rotating a vector from one reference frame to another, and allowing for any `from_frame` and `to_frame`? Then you could add a simple wrapper function that is specifically useful for rotating from a specified instrument frame into the IMAP_DPS frame.

imap_processing/ultra/l1b/ultra_l1b_annotated.py
```diff
@@ -0,0 +1,50 @@
...
+    time: np.ndarray,
+    instrument_velocity: np.ndarray,
+    instrument_frame: str,
```
tmplummer: This should require the frame input to be `SpiceFrame` type as defined in spice/geometry.py. It looks like IMAP_DPS is defined in the `SpiceBody` enum but not `SpiceFrame`.

imap_processing/ultra/l1b/ultra_l1b_annotated.py
tmplummer: I might walk this comment back. Maybe what we should do is create a ticket that is for adding the SPICE frame transform code in one PR and then modify this function once that is in place to use that function. There is something that I can pull from EMUS that is close to this implementation.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -649,3 +649,111 @@ def get_energy_ssd(de_dataset: xarray.Dataset, ssd: np.ndarray) -> NDArray[np.fl
...
+    Returns
+    -------
+    ctof : np.array
+        Corrected TOF.
+    bin : np.array
+        Species bin.
+    """
+    z_dstop = 2.6 / 2  # position of stop foil on Z axis (mm)
+    z_ds = 46.19 - z_dstop  # position of slit on Z axis (mm)
+    df = 3.39  # distance from slit to foil (mm)
```
tmplummer: I see these values defined in multiple functions. It seems like you really should have one place where all of the instrument constants are defined. For EMUS I did something like this using a YAML file and wrote an accessor function that would retrieve instrument parameters from that file. It seems like others define things like this in an `__inti__.py` file. I don't know if there is consistency across the project, but defining the same constant in multiple locations should definitely be avoided.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -649,3 +649,111 @@ def get_energy_ssd(de_dataset: xarray.Dataset, ssd: np.ndarray) -> NDArray[np.fl
...
+
+    # SSD event TOF normalization to Z axis
+    dmin = z_ds - np.sqrt(2) * df  # (mm)
+    dmin_ssd_ctof = dmin**2 / (dmin - z_dstop)  # (mm)
+    # Multiply times 100 to convert to hundredths of a millimeter.
+    ctof = tof * dmin_ssd_ctof * 100 / r  # (tenths of a ns)
```
tmplummer: Calculation of corrected TOF should be extracted into a function since both `determine` functions are doing the same calculation.

imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
```diff
@@ -0,0 +1,53 @@
...
+        "TDB",
+    )
+    # Get start and end time of first interval
+    start, _ = spice.wnfetd(pointing_cover, 0)
+
+    times = np.array([start])
+
+    instrument_velocity = np.array([[41.18609, -471.24467, -832.8784]])
+
+    velocity_45 = get_particle_velocity(times, instrument_velocity, "IMAP_ULTRA_45")
```
tmplummer: I recommend trying to avoid hard coding names even in tests by using the `SpiceFrame` enum class.

tmplummer (commented):
Overall, looks great. I would like to see this broken into two PRs in the future as the SPICE related function is totally separate from the species related functions.

I think that the only thing I would like to see addressed is regarding the duplicated code in the species functions. Perhaps writing a ticket to address the comment about locating all of the instrument constants in one place.

imap_processing/ultra/l1b/ultra_l1b_extended.py
vmartinez-cu: I suggest updating the variable names to be more descriptive. So instead of 'r' use 'path_length' and instead of 'tof' use 'time_of_flight'. This will help improve readability of the code and can replace some of the comments.

imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
```diff
@@ -0,0 +1,53 @@
...
+    velocity_90 = get_particle_velocity(times, instrument_velocity, "IMAP_ULTRA_90")
+
+    # Compute the magnitude of the velocity vectors in both frames
+    magnitude_45 = np.linalg.norm(velocity_45[0])
+    magnitude_90 = np.linalg.norm(velocity_90[0])
+
+    assert np.allclose(magnitude_45, magnitude_90, atol=1e-6)
```
vmartinez-cu, line 57: Nice use of np.allclose! I haven't used that before

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -289,3 +291,31 @@ def test_get_energy_pulse_height(de_dataset, yf_fixture):
     test_energy = df_ph["Energy"].astype("float")
 
     assert np.array_equal(test_energy, energy[ph_indices])
+
+
+def test_determine_species_ph(yf_fixture):
```
vmartinez-cu, line 315: Can you add an explanation of what 'yf' is in the fixture description? I assume it's front y position but I'm not sure

vmartinez-cu (commented):
Nice work!

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
vmartinez-cu: Nvm, I see that it's front y position. Spelling this out in the fixture description would be helpful but not necessary

imap_processing/ultra/l1b/ultra_l1b_extended.py
laspsandoval: Good catch.

laspsandoval: For sure regarding "r".

imap_processing/ultra/l1b/ultra_l1b_annotated.py
laspsandoval: Reviewed your code and I agree. If ok, I'll leave it for now, but create a ticket to modify this function

tmplummer (approved):
Thanks for addressing my comments. Looks good.

PR #852, opened 2024-09-18
Spin phase calculation using spin data

### LINKED ISSUE #806: SPICE - Spin phase query utils function - using spin start and spin rate data.

tech3371:
# Change Summary
closes #806 

## Overview
Calculate spin phase

## Updated Files
   - Code to calculate spin phase

## Testing

### DIFF
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -202,7 +202,7 @@ def get_spacecraft_spin_phase(
     # 1. Check that the spin phase is in valid range, [0, 1].
     # 2. Check invalid spin phase using spin_phase_valid,
     #   spin_period_valid columns.
-    invalid_spin_phase_range = (spin_phases < 0) | (spin_phases > 1)
+    invalid_spin_phase_range = (spin_phases < 0) | (spin_phases >= 1)
 
     invalid_spins = (spin_df["spin_phase_valid"].values[last_spin_indices] == 0) | (
         spin_df["spin_period_valid"].values[last_spin_indices] == 0
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -142,9 +142,7 @@ def get_spacecraft_spin_phase(
     Get the spacecraft spin phase for the input query times.
 
     Formula to calculate spin phase:
-        spin_phase = (
-            query_met_times - (spin_start_seconds + spin_start_subseconds / 1e3)
-        ) / spin_period_sec
+        spin_phase = (query_met_times - spin_start_time) / spin_period_sec
 
     Parameters
     ----------
@@ -165,6 +163,9 @@ def get_spacecraft_spin_phase(
         # Force scalar to array because np.asarray() will not
         # convert scalar to array
         query_met_times = np.atleast_1d(query_met_times)
+    # Empty array check
+    if query_met_times.size == 0:
+        return query_met_times
 
     # Create an empty array to store spin phase results
     spin_phases = np.zeros_like(query_met_times)
@@ -179,10 +180,18 @@ def get_spacecraft_spin_phase(
     last_spin_indices = np.searchsorted(
         spin_df["spin_start_time"], query_met_times, side="right"
     )
-    # Use np.clip to ensure that the last_spin_indices are within the range.
-    # >>> np.clip(x, 0, len(df['a']) -1 )
-    # array([1, 1, 2, 3, 4])
-    last_spin_indices = np.clip(last_spin_indices, 0, len(spin_df) - 1)
+    # Make sure input times are within the bounds of spin data
+    spin_df_start_time = spin_df["spin_start_time"].values[0]
+    spin_df_end_time = (
+        spin_df["spin_start_time"].values[-1] + spin_df["spin_period_sec"].values[-1]
+    )
+    input_start_time = query_met_times.min()
+    input_end_time = query_met_times.max()
+    if input_start_time < spin_df_start_time or input_end_time > spin_df_end_time:
+        raise ValueError(
+            f"Query times, {query_met_times} are outside of the spin data range, "
+            f"{spin_df_start_time, spin_df_end_time}."
+        )
 
     # Calculate spin phase
     spin_phases = (
@@ -190,22 +199,15 @@ def get_spacecraft_spin_phase(
     ) / spin_df["spin_period_sec"].values[last_spin_indices]
 
     # Check for invalid spin phase using below checks:
-    # 1. Check that the spin period is valid. To do that, calculate the
-    #   time difference between the query_met_times and the spin start time.
-    #   Check that the time difference is within the spin period.
-    # 2. Check that the spin phase is valid range, [0, 1].
-    # 3. Check invalid spin phase using spin_phase_valid,
+    # 1. Check that the spin phase is in valid range, [0, 1].
+    # 2. Check invalid spin phase using spin_phase_valid,
     #   spin_period_valid columns.
-    time_diff = query_met_times - spin_df["spin_start_time"].values[last_spin_indices]
-    invalid_spin_period = (
-        time_diff > spin_df["spin_period_valid"].values[last_spin_indices]
-    )
     invalid_spin_phase_range = (spin_phases < 0) | (spin_phases > 1)
 
     invalid_spins = (spin_df["spin_phase_valid"].values[last_spin_indices] == 0) | (
         spin_df["spin_period_valid"].values[last_spin_indices] == 0
     )
-    bad_spin_phases = invalid_spin_period | invalid_spin_phase_range | invalid_spins
+    bad_spin_phases = invalid_spin_phase_range | invalid_spins
     spin_phases[bad_spin_phases] = np.nan
 
     if is_scalar:
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -189,12 +189,24 @@ def get_spacecraft_spin_phase(
         query_met_times - spin_df["spin_start_time"].values[last_spin_indices]
     ) / spin_df["spin_period_sec"].values[last_spin_indices]
 
-    # Check for spin phase that are not valid spin phase.
-    # Right now, we are checking that the spin phase is between 0 and 1.
-    # TODO: filter out more invalid spin phase using spin_phase_valid,
-    # spin_period_valid columns.
-    invalid_spin_phase = (spin_phases < 0) | (spin_phases > 1)
-    spin_phases[invalid_spin_phase] = np.nan
+    # Check for invalid spin phase using below checks:
+    # 1. Check that the spin period is valid. To do that, calculate the
+    #   time difference between the query_met_times and the spin start time.
+    #   Check that the time difference is within the spin period.
+    # 2. Check that the spin phase is valid range, [0, 1].
+    # 3. Check invalid spin phase using spin_phase_valid,
+    #   spin_period_valid columns.
+    time_diff = query_met_times - spin_df["spin_start_time"].values[last_spin_indices]
+    invalid_spin_period = (
+        time_diff > spin_df["spin_period_valid"].values[last_spin_indices]
+    )
+    invalid_spin_phase_range = (spin_phases < 0) | (spin_phases > 1)
+
+    invalid_spins = (spin_df["spin_phase_valid"].values[last_spin_indices] == 0) | (
+        spin_df["spin_period_valid"].values[last_spin_indices] == 0
+    )
+    bad_spin_phases = invalid_spin_period | invalid_spin_phase_range | invalid_spins
+    spin_phases[bad_spin_phases] = np.nan
 
     if is_scalar:
         return spin_phases[0]
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -199,7 +199,7 @@ def get_spacecraft_spin_phase(
     ) / spin_df["spin_period_sec"].values[last_spin_indices]
 
     # Check for invalid spin phase using below checks:
-    # 1. Check that the spin phase is in valid range, [0, 1].
+    # 1. Check that the spin phase is in valid range, [0, 1).
     # 2. Check invalid spin phase using spin_phase_valid,
     #   spin_period_valid columns.
     invalid_spin_phase_range = (spin_phases < 0) | (spin_phases >= 1)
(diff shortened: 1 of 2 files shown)


imap_processing/spice/geometry.py
```diff
@@ -149,13 +149,45 @@ def get_spacecraft_spin_phase(
...
+
+    # Combine spin_start_sec and spin_start_subsec to get the spin start
+    # time in seconds. The spin start subseconds are in milliseconds.
+    # TODO: Decide if we should do this calculation in the data itself
+    # or in get_spin_data function.
+    spin_df["spin_start_time"] = (
+        spin_df["spin_start_sec"] + spin_df["spin_start_subsec"] / 1e3
+    )
+
     if isinstance(query_met_times, float):
```
greglucas: This is the same code duplicated in both paths essentially. You should be able to do something like:
1. see what we got as input `is_scalar = isinstance(query_met_times, numbers.Number)`
2. convert met_times to an array regardless `query_met_times = np.asarray(query_met_times)` and use that for the calculations
3. special-case the return value `if is_scalar: convert_to_scalar, ...`

imap_processing/spice/geometry.py
```diff
@@ -149,13 +149,45 @@ def get_spacecraft_spin_phase(
...
+
+        if mask.any():
+            # Get the last spin before or at the query time
+            last_spin = spin_df[mask].iloc[-1]
+            # Calculate spin phase using the formula
+            spin_phases[i] = (data_time - last_spin["spin_start_time"]) / last_spin[
+                "spin_period_sec"
+            ]
+        else:
+            raise ValueError(f"No spin data found for this data time: {data_time}")
```
greglucas: I think this is `searchsorted` to get the indices?
https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html

```python
# Filter out values above/below, raise, ...?
query_met_times = np.clip(query_met_times, spin_df["spin_start_time"][0], spin_df["spin_start_time"][-1])
last_spin_indices = np.searchsorted(spin_df["spin_start_time"], query_met_times)
spin_df_indexed = spin_df.iloc[last_spin_indices]
spin_phases = (query_met_times - spin_df_indexed["spin_start_time"]) / spin_df_indexed["spin_period_sec"]
```

greglucas: Do you need to multiply by `2*pi` here? Are we expecting spin phase in radians?

imap_processing/spice/geometry.py
```diff
@@ -149,13 +149,45 @@ def get_spacecraft_spin_phase(
...
+            )
+
+    # Create an empty array to store spin phase results
+    spin_phases = np.zeros_like(query_met_times)
+
+    # Loop through each query time and find the corresponding spin start time
+    for i, data_time in enumerate(query_met_times):
+        # Find the row where the spin start time is less than or
+        # equal to the query time
+        mask = spin_df["spin_start_time"] <= data_time
```
tech3371: Add a check if an entry missing spin table. Add a check to make sure spin phase is less than 1 since it should be between 0-1.

imap_processing/spice/geometry.py
```diff
@@ -149,13 +149,45 @@ def get_spacecraft_spin_phase(
     spin_phase : float or np.ndarray
         Spin phase for the input query times.
     """
+    spin_df = get_spin_data()
+
+    # Combine spin_start_sec and spin_start_subsec to get the spin start
+    # time in seconds. The spin start subseconds are in milliseconds.
+    # TODO: Decide if we should do this calculation in the data itself
```
tmplummer: It makes sense to me to add a floating point MET column in the `get_spin_data()` function.

imap_processing/spice/geometry.py
tmplummer: The suggested use of `np.clip` scares me because it will silently modify the query times and cause this function to return incorrect spin phase rather than raising an exception.

imap_processing/spice/geometry.py
```diff
@@ -149,13 +149,45 @@ def get_spacecraft_spin_phase(
...
-        return 0.0  # Replace this with actual logic to calculate spin phase
-
-    # Return an ndarray of the same shape, filled with 0.0 for
-    # now (dummy implementation)
-    # TODO: Replace with actual logic to calculate spin phase
-    return np.array(query_met_times, dtype=float)
+        # calculate spin phase for a single query time
+        mask = spin_df["spin_start_time"] <= query_met_times
+        if mask.any():
+            last_spin = spin_df[mask].iloc[-1]
```
tmplummer: As we discussed in the ENA tagup, we need a check to make sure this last_spin index doesn't index a previous spin due to a missing spin in the spin table. It also occurs to me that we may want/need to filter out bad spins using the `valid` flags present in the table. Since we are still gathering information about what those flags mean, perhaps this should just be captured in a TODO and github ticket.

tmplummer (commented):
Nice work. I spent a bit more time reviewing and added a couple of comments.

imap_processing/spice/geometry.py
tech3371: About your second point about filtering bad spin, I was just thinking about this. I was thinking of returning pd.DataFrame with spin information and spin phase calculated for the input times and let user decide what they want to do with spin info or filter spin phase based on bad spin or just use spin phase column from returned pd.DataFrame. What do you guys think?

imap_processing/spice/geometry.py
tech3371: I was just reading on those two and I don't understand `np.clip` but `np.searchsorted` seems like a good approach. I am still reading though :)

imap_processing/spice/geometry.py
tmplummer: I would vote for just returning the spin phase for now. I think that later we might want to add in returning a quality flag along with each of the spin phase values. I think that returning too much will promote duplicated code each place this function is called rather than keeping the code common.

tech3371: Will do. That will make things easier here. user can call `get_spin_data` directly too if needed without making this function too complicated. Good thinking!

imap_processing/spice/geometry.py
greglucas: > The suggested use of np.clip scares me because it will silently modify the query times and cause this function to return incorrect spin phase rather than raising an exception.

Fair enough, we should never encounter that situation anyways, so I am fine with raising an error instead of removing the input requests (probably smarter for shapes of arrays as well!). The issue is you do still need to add some check for out of bounds because I think `searchsorted` will return 0 even if it is outside of the leftmost value.

tech3371: Look like np.clips does help in some things. Eg.
```
>>> x = np.searchsorted(df['a'], [0, 17, 19])
>>> x
array([0, 5, 5])
>>> np.clip(x,0, len(df['a'])-1)
array([0, 4, 4])
```

imap_processing/spice/geometry.py
tech3371: created this issue for future work: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/863

imap_processing/spice/geometry.py
```diff
@@ -136,7 +143,7 @@ def get_spacecraft_spin_phase(
 
     Formula to calculate spin phase:
         spin_phase = (
-            query_met_times - (spin_start_seconds + spin_start_subseconds)
+            query_met_times - (spin_start_seconds + spin_start_subseconds / 1e3)
```
greglucas:
```suggestion
            (query_met_times - spin_start_time) / spin_period_sec
```

imap_processing/spice/geometry.py
```diff
@@ -149,13 +156,58 @@ def get_spacecraft_spin_phase(
...
+    # array([0, 15, 30, 45, 60])
+    # >>> np.searchsorted(df['a'], [0, 13, 15, 32, 70], side='right')
+    # array([1, 1, 2, 3, 5])
+    last_spin_indices = np.searchsorted(
+        spin_df["spin_start_time"], query_met_times, side="right"
+    )
+    # Use np.clip to ensure that the last_spin_indices are within the range.
+    # >>> np.clip(x, 0, len(df['a']) -1 )
+    # array([1, 1, 2, 3, 4])
+    last_spin_indices = np.clip(last_spin_indices, 0, len(spin_df) - 1)
```
greglucas: I thought we were going to raise if this happened?

```suggestion
    # Make sure we aren't going out of the bounds
    if query_met_times.min() < spin_df["spin_start_time"][0] or query_met_times.max() > spin_df["spin_start_time"][-1] + spin_df["spin_period_sec"][-1]:
        raise ValueError
```

imap_processing/spice/geometry.py
```diff
@@ -149,13 +156,58 @@ def get_spacecraft_spin_phase(
...
+    #   time difference between the query_met_times and the spin start time.
+    #   Check that the time difference is within the spin period.
+    # 2. Check that the spin phase is valid range, [0, 1].
+    # 3. Check invalid spin phase using spin_phase_valid,
+    #   spin_period_valid columns.
+    time_diff = query_met_times - spin_df["spin_start_time"].values[last_spin_indices]
+    invalid_spin_period = (
+        time_diff > spin_df["spin_period_valid"].values[last_spin_indices]
+    )
+    invalid_spin_phase_range = (spin_phases < 0) | (spin_phases > 1)
```
greglucas: I think this is a redundant check? Your `invalid_spin_phase_range` should be this same calculation.
```suggestion
    invalid_spin_phase_range = (spin_phases < 0) | (spin_phases > 1)
```

greglucas (commented):
I don't think we have a spin phase check yet, could you add one where you make a mock spin data product with all of these potential cases in it and make sure we get the right spin phases as you expect?

tech3371:
> I don't think we have a spin phase check yet, could you add one where you make a mock spin data product with all of these potential cases in it and make sure we get the right spin phases as you expect?

Let me ask this in Slack as I have some detail question.

imap_processing/spice/geometry.py
tech3371: That works. I hope we add checks in dependency to avoid this value error.

imap_processing/spice/geometry.py
tech3371: That works for me. I tend to over complicate things sometimes.

imap_processing/spice/geometry.py
```diff
@@ -136,7 +143,7 @@ def get_spacecraft_spin_phase(
```
laspsandoval, line 143: Tim had brought up a good point in a meeting which is: what would happen if a spin failed to be recorded in the .csv. I don't immediately see how this would be handled here, but I might just be missing it.

laspsandoval (commented):
Just a double check that you are accounting for the potential for missing spins in the .csv.

imap_processing/spice/geometry.py
tech3371: Laura, I believe this check `invalid_spin_phase_range = (spin_phases < 0) | (spin_phases > 1)` will capture for this. If a spin data is missing, it will result in spin phase being less than or greater than 1. @subagonsouth correct me if that's not case.

tmplummer: Yes, checking that the spin_phase is in the range [0, 1) is a good way to do this check, IMO.

tech3371: I pushed changes to correct what I had to `[0, 1)`.

tmplummer (approved):
LGTM. Thanks!

tech3371:
I am going to merge this to keep it moving forward. I think I address all of PR comments. Don't hesitate to comment. I am happy to open new PR with your requests.

PR #857, opened 2024-09-19
MNT: Change default to use raw value in  packet_file_to_datasets

### LINKED ISSUE #856: Make raw output the default for packet parsing
### Summary

There have been several cases where people wanted to just have the "raw" values output from the packet parser instead of the derived values. It would be more helpful to just stick with the raw values for the default.

### Requirements

The more specific the better.
  - [ ] Change the default to `False` https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/2f79f29d89ea397865cd16cb7a468c879b57e63c/imap_processing/utils.py#L301

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview

Looks like no test updates which is nice! Hope this helps for future work as well since I think all of you have mentioned this would be a better default value.

closes #856

### DIFF
--- a/imap_processing/utils.py
+++ b/imap_processing/utils.py
@@ -298,7 +298,7 @@ def _get_minimum_numpy_datatype(  # noqa: PLR0912 - Too many branches
 def packet_file_to_datasets(
     packet_file: Union[str, Path],
     xtce_packet_definition: Union[str, Path],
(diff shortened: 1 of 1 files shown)


tech3371 (approved):

sdhoyt (approved):
Thanks for making this update!

PR #858, opened 2024-09-19
ENH: Add "Parents" global attribute to datasets

### LINKED ISSUE #849: Include parent data products in the created output CDF files
### Summary

https://spdf.gsfc.nasa.gov/istp_guide/gattributes.html

> Parents --- optional
This attribute lists the parent CDF(S) for files of derived and merged data sets. Subsequent entry values are used for multiple parents. The syntax for a CDF parent would be e.g. "CDF>logical_file_id".

### Requirements

  - [ ] All downloaded input files should be added to the parents metadata attributes of produced data files.

This can likely be done during the write_cdf() routine, when we inject things because we can keep track of what was downloaded as dependencies coming into the CLI. I wonder if any of the instruments will be downloading calibration files or other things outside of the initial CLI download routines though?

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview

This injects the Parent files used to create this data product when writing the CDF file. This takes all of the dependencies input during the processing run and saves them to the processing runner class. Then adds a new path to the `write_cdf()` routine to inject the parents attribute when writing the dataset, similar to how we are adding other global attributes like logical file id.

Note that we can add other dependencies like calibration files and the like in the processing routines directly if we want to do that too.

closes https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/849

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -133,7 +133,7 @@ def write_cdf(
         # Include the current files if there are any and include just the filename
         # [file1.txt, file2.cdf, ...]
         dataset.attrs["Parents"] = dataset.attrs.get("Parents", []) + [
-            p.name for p in parent_files
+            parent_file.name for parent_file in parent_files
         ]
 
     # Convert the xarray object to a CDF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -3,6 +3,7 @@
 import logging
 import re
 from pathlib import Path
+from typing import Optional
 
 import imap_data_access
 import numpy as np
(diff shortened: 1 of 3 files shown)


imap_processing/cdf/utils.py
```diff
@@ -121,6 +128,13 @@ def write_cdf(dataset: xr.Dataset, **extra_cdf_kwargs: dict) -> Path:
     dataset.attrs["Logical_file_id"] = file_path.stem
     # Add the processing version to the dataset attributes
     dataset.attrs["ground_software_version"] = imap_processing._version.__version__
+    # Add any parent files to the dataset attributes
+    if parent_files:
+        # Include the current files if there are any and include just the filename
+        # [file1.txt, file2.cdf, ...]
+        dataset.attrs["Parents"] = dataset.attrs.get("Parents", []) + [
+            p.name for p in parent_files
```
bourque:
```suggestion
            parent_file.name for parent_file in parent_files
```

My preference here is to be more explicit.  But you can take it or leave it!

bourque (approved):
Good idea! I recall discussion of this being something we should implement but I had forgotten all about it.

This looks good to me!

tmplummer (approved):
Looks great!

PR #860, opened 2024-09-20
Lo L1A CDF YAML Reorganization

### LINKED ISSUE #843: Improve Lo L1A CDF YAML Definitions
### Summary

The Lo L1A CDF YAML definitions should be re-worked in the following ways
- make better use of shared attributes
- change to use int32 for fields currently using int64
- add datatype field to specify the data type to be used when creating the data array

### Requirements

_No response_

### Related tickets

_No response_

sdhoyt:
# Change Summary

Reorganized the Lo L1A CDF YAML to make better use of default values and added data type to use in data arrays to match what Hi is currently doing

Closes #843 

# Testing
- Ran existing unit tests
- Ran cli.py to create L1A histogram CDF and spot checked contents

### DIFF
--- a/imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
@@ -1,56 +1,42 @@
 default_attrs: &default
   # Assumed values for all variable attrs unless overwritten
   DISPLAY_TYPE: time_series
-  DEPEND_0: epoch
   FILLVAL: -9223372036854775808
   FORMAT: I12
   VALIDMIN: 0
   VALIDMAX: 9223372036854775807
-  VAR_TYPE: data
   UNITS: ' '
+
+coord_default_attrs: &coord_default
+  <<: *default
+  VAR_TYPE: support_data
 
 de_default_attrs: &de_default
-  DISPLAY_TYPE: time_series
-  FILLVAL: -9223372036854775808
-  FORMAT: I12
-  VALIDMIN: 0
-  VALIDMAX: 9223372036854775807
+  <<: *default
   VAR_TYPE: support_data
   DEPEND_1: direct_events
-  UNITS: ' '
+
+hist_az_60_default: &hist_az_60_default
+  <<: *default
+  VAR_TYPE: data
+  DEPEND_0: epoch
+  DEPEND_1: azimuth_60
+  DEPEND_2: esa_step
+
+hist_az_6_default: &hist_az_6_default
+  <<: *default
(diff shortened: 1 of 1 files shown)


maxinelasp (approved):
Looks good, one minor thing that might not even be wrong. Did you double check that you can stack defaults like that and get the expected result?

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -1,133 +1,182 @@
...
+
+default_uint32_attrs: &default_uint32
+  <<: *default
+  FILLVAL: 4294967295
+  FORMAT: I10
+  VALIDMIN: 0
+  VALIDMAX: 4294967295
+  dtype: uint32
+
+default_uint64_attrs: &default_uint64
```
tmplummer: I found out the hard way that CDF doesn't have a CDF_UINT8 datatype, so I think that we need to use `int64` only, not `uint64`.

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -1,133 +1,182 @@
...
+
+spin_default_attrs: &spin_default
+  <<: *default
+  VAR_TYPE: support_data
+  DEPEND_1: spin
+
 
 # Re-used Coordinate
+epoch:
+  <<: *default_uint32
```
tmplummer: Everything I have seen indicates that we want to use `int64` for epoch.

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -1,133 +1,182 @@
...
+
 esa_step_label:
    CATDESC: ESA Steps
    FIELDNAM: ESA Steps
    FORMAT: A1
    VAR_TYPE: metadata
 
 esa_step:
+  <<: *coord_default
+  <<: *default_uint8
```
tmplummer, line 93: Oh, interesting! I hadn't thought about multiple inheritance like this. Good pattern to keep in mind.

tmplummer (commented):
I made two comments worth looking at before merging. Thanks for keeping this to a single file change rather than making the code changes in the same PR!

sdhoyt:
> Looks good, one minor thing that might not even be wrong. Did you double check that you can stack defaults like that and get the expected result?

I spot checked one of the stacked fields in the CDF and it has all the attribute keys with the values I expected, so I think it works as you'd expect.

tmplummer (approved):

PR #861, opened 2024-09-20
859 spice frame transform function

### LINKED ISSUE #859: SPICE: Frame transform function
### Algorithm Description:

Implement a function for transforming vectors from one frame to another.

From EMUS:

``` Python
def get_vec_pxform():
    """
    Return a vectorized form of spice.pxform function.

    "Return the matrix that transforms position vectors from one specified frame to another at a specified epoch."
    https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/pxform_c.html
    """
    return np.vectorize(spice.pxform,
                        excluded=['fromstr', 'tostr'],
                        signature='(),(),()->(3,3)',
                        otypes=[np.float64])


@ensure_spice
def frame_transform(from_frame: SpiceFrame, to_frame: SpiceFrame, et: float or np.ndarray, position: np.ndarray,
                    normalize: bool = False) -> np.ndarray:
    """
    Transform a position <x, y, z> vector between reference frames, optionally normalizing the result.

    :param from_frame: SpiceFrame, required, Reference frame of position
    :param to_frame: SpiceFrame, required, Reference frame of output
    :param et: np.float64 or np.ndarray with dtype np.float64, required, Ephemeris time(s) corresponding to position(s)
    :param position: np.ndarray, required, <x, y, z> vector or array of vectors in reference frame `from_frame`
    :param normalize: bool, optional, Optionally normalize the output vector
    :return: np.ndarray, 3d position vector(s) in reference frame `to_frame`
    """
    if position.ndim == 1:
        assert len(position) == 3
        assert isinstance(et, float)
    elif position.ndim == 2:
        assert len(position) == len(et)
        assert len(position[0]) == 3

    vec_pxform = get_vec_pxform()
    rotate = vec_pxform(from_frame.value.strid, to_frame.value.strid, et)

    if hasattr(rotate[0][0], '__len__'):
        result = np.array([np.matmul(rotate, pos).astype(np.float64) for rotate, pos in zip(rotate, position)])
        if normalize:
            result = normalize_vectors(result)
    else:
        result = np.matmul(rotate, position).astype(np.float64)
        if normalize:
            result = result / np.linalg.norm(result)

    return result
```

tmplummer:
# Change Summary

## Overview
Add SPICE functions for transforming positions from one reference frame to another. This is useful for things like transforming particle velocity vectors.

## Updated Files
   - Added function: `frame_transform`
   - Added function: `get_rotation_matrix`
   - Added test coverage for new geometry functions
   - Added general fixture/context manager for furnishing a list of kernels.
 
closes #859

### DIFF
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -12,7 +12,7 @@ import os
 import typing
 from enum import IntEnum
 from pathlib import Path
-from typing import Union
+from typing import Any, Callable, Union
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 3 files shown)


imap_processing/spice/geometry.py
```diff
@@ -159,3 +161,102 @@ def get_spacecraft_spin_phase(
...
+
+    rotate = get_rotation_matrix(from_frame, to_frame, et)
+
+    if hasattr(rotate[0][0], "__len__"):
+        result = np.array(
+            [
+                np.matmul(rotate, pos).astype(np.float64)
+                for rotate, pos in zip(rotate, position)
+            ]
+        )
```
greglucas: Could you add a comment about what the shapes of the arrays you're expecting in these two cases are?

I think this is the array path and that you could broadcast the matmul operations, expanding/shrinking extra dimensions as needed to avoid the for loop here.

```python
# (i, n, k) @ (i, k, m) -> (i, n, m)
np.array(rotate, dtype=np.float64) @ np.array(position, dtype=np.float64)
```

imap_processing/spice/geometry.py
```diff
@@ -159,3 +161,102 @@ def get_spacecraft_spin_phase(
...
     return np.array(query_met_times, dtype=float)
+
+
+@typing.no_type_check
+@ensure_spice
+def frame_transform(
+    from_frame: SpiceFrame,
+    to_frame: SpiceFrame,
+    et: Union[float, npt.NDArray],
+    position: npt.NDArray,
```
greglucas: I'm not sure if we have any standard way of handling argument order, but this stuck out to me at a quick glance expecting the arrays to be first in the list and the from/to to be last. Not a major preference from me other than to say we should be consistent if we've been doing it this way then we should stick to it.

imap_processing/spice/geometry.py
tmplummer: Yes! Thanks for commenting on this. I had identified this as something to investigate when I copied the code but forgot to do that. I also learned about the `@` operator!

imap_processing/spice/geometry.py
tmplummer: Good point. I was keeping arguments the same as SPICE uses but I'll think a bit about if one way or the other makes more sense.

imap_processing/spice/geometry.py
```diff
@@ -58,6 +58,8 @@ class SpiceFrame(IntEnum):
     IMAP_HIT = -43500
     IMAP_IDEX = -43700
     IMAP_GLOWS = -43750
+    # IMAP Pointing Frame (Despun) as defined in iamp_science_0001.tf
```
laspsandoval, line 61: Typo: imap_science_0001.tf

imap_processing/spice/geometry.py
```diff
@@ -159,3 +161,102 @@ def get_spacecraft_spin_phase(
...
+                for rotate, pos in zip(rotate, position)
+            ]
+        )
+    else:
+        result = np.matmul(rotate, position).astype(np.float64)
+
+    return result
+
+
+def get_rotation_matrix(
```
laspsandoval, line 287: Nice way to use a numpy array w/ pxform

imap_processing/spice/geometry.py
```diff
@@ -159,3 +161,102 @@ def get_spacecraft_spin_phase(
...
+    et : float or npt.NDArray
+        Ephemeris time(s) corresponding to position(s).
+    position : npt.NDArray
+        <x, y, z> vector or array of vectors in reference frame `from_frame`.
+
+    Returns
+    -------
+    result : npt.NDArray
+        3d position vector(s) in reference frame `to_frame`.
+    """
```
laspsandoval, line 253: In the docstring could you add a comment that this performs the same operation as spice.mxv except that it handle multiple transformations in a single, unified step?

laspsandoval (approved):
Looks good. Thanks Tim.

PR #872, opened 2024-09-24
Convert all hi instances of uint64 CDF variables to int64

### LINKED ISSUE #871: Hi product definition update: convert uint64 to int64
### Algorithm Description:
Convert all Hi datatypes from uint64 to int64. No uint64 exists in CDF.

tmplummer:
# Change Summary

## Overview
This is a simple change converting all Hi CDF variables that were defined as uint64 to int64 since uint64 does not exist in CDF.

## Updated Files
   - search and replace uint64 -> int64

Closes #871

### DIFF
--- a/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
@@ -31,13 +31,13 @@ default_uint32_attrs: &default_uint32
   VALIDMAX: 4294967295
   dtype: uint32
 
-default_uint64_attrs: &default_uint64
+default_int64_attrs: &default_int64
   <<: *default
   FILLVAL: 18446744073709551615
   FORMAT: I20
   VALIDMIN: 0
   VALIDMAX: 18446744073709551615
-  dtype: uint64
+  dtype: int64
 
 default_int32_attrs: &default_int32
   <<: *default
@@ -91,7 +91,7 @@ hi_de_esa_stepping_num:
   <<: *hi_esa_stepping_num
 
 hi_de_event_met:
-  <<: *default_uint64
+  <<: *default_int64
   CATDESC: Mission Elapsed Time (MET) of Direct Event
   DISPLAY_TYPE: no_plot
   FIELDNAM: Time since (TODO - what is the def of MET?)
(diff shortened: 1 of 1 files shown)


imap_processing/cdf/config/imap_hi_variable_attrs.yaml
```diff
@@ -91,7 +91,7 @@ hi_de_esa_stepping_num:
   <<: *hi_esa_stepping_num
 
 hi_de_event_met:
-  <<: *default_uint64
+  <<: *default_int64
```
greglucas, line 94: Do any of these need to be `uint32`? I think that MET has typically been 32 bit unsigned so wondering if this should be the same as that.

greglucas (approved):

imap_processing/cdf/config/imap_hi_variable_attrs.yaml
tech3371: I believe `Hi` needs int64 because `Hi` converts `met_time` to nanosecond accuracy. https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/d24478fafa7f477a8fb394945057baef78ffb7c5/imap_processing/hi/l1a/science_direct_event.py#L243

imap_processing/cdf/config/imap_hi_variable_attrs.yaml
```diff
@@ -31,13 +31,13 @@ default_uint32_attrs: &default_uint32
   VALIDMAX: 4294967295
   dtype: uint32
 
-default_uint64_attrs: &default_uint64
+default_int64_attrs: &default_int64
   <<: *default
   FILLVAL: 18446744073709551615
   FORMAT: I20
   VALIDMIN: 0
   VALIDMAX: 18446744073709551615
```
tech3371: Are these still good with change from `uint64` to `int64`?

tech3371 (approved):

imap_processing/cdf/config/imap_hi_variable_attrs.yaml
tmplummer: Correct. The Direct Event MET is in nanoseconds. Also in L1A is the CCSDS MET which is a uint32 and stores seconds.

imap_processing/cdf/config/imap_hi_variable_attrs.yaml
tmplummer: Ah... Good catch! Nope, those are not correct.

PR #875, opened 2024-09-24
Add Ultra spatial and energy binning equations.

laspsandoval:
# Change Summary

## Overview
Add spatial and energy binning equations.

## Updated Files
- ultra_l1c_pset_bins.py
   - Added spatial and energy binning mechanisms.

## Testing
- test_ultra_l1c_pset_bins.py
   - Tests ultra_l1c_pset_bins.py


imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +48,221 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+    hist, _ = np.histogramdd(
+        sample=(az, el, energy), bins=[az_bin_edges, el_bin_edges, energy_bin_edges]
+    )
+
+    non_zero_counts = hist[hist > 0]
+
+    return hist, non_zero_counts
+
+
+def create_unique_identifiers(
```
tmplummer: I don't understand what this function does. Is the point that you only want to store data in the file for spatial bins that have counts in them at one or more energy levels? So your counts array in the CDF is shaped like `n_b x n_e `where `n_b` is the number of spatial bins with at least one count in at least one energy bin and `n_e` is the number of energy bins?

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +48,221 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+
+    # Get the starting edges for energy bins
+    energy_bin_start = energy_bin_edges[:-1]
+
+    # Create unique identifiers and map non-zero counts to 2D histogram array
+    counts, bin_id, midpoints = create_unique_identifiers(
+        hist, az_bin_midpoints, el_bin_midpoints, non_zero_counts
+    )
+
+    _, _, energy_bin = np.argwhere(hist > 0).T
```
tmplummer: I see this use of `np.argwhere(hist > 0).T` in 2 places and also `non_zero_counts = hist[hist>0]` in another place. It would be good to consolidate those to a single place so that there aren't hidden inter-dependencies between these three new functions.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +48,221 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+    -------
+    hist : np.ndarray
+        A 3D histogram array.
+    non_zero_counts : np.ndarray
+        An array of non-zero counts.
+    """
+    az, el, _ = cartesian_to_spherical(v)
+
+    # 3D binning.
+    hist, _ = np.histogramdd(
```
tmplummer, line 143: Yay! This was drastically simplified by using `np.histogramdd`.

tmplummer (commented):
It might be worth getting in a meeting to talk this through. I am having a little trouble understanding what the goal of some of the code is.

I haven't used them extensively, but some of the code made me think that numpy MaskedArray might be useful. Especially the linked section:
https://numpy.org/doc/stable/reference/routines.ma.html#finding-masked-data

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
laspsandoval: hist.size = 6220800 and that doesn't yet include the bin for species. So what I am doing is eliminating the bins that do not have counts.

This is where I get the indices for the nonzero bins:
az_bin_id, el_bin_id, energy_bin_id = np.argwhere(hist > 0).T
az_bin_id = array([150, 269, 269])
el_bin_id = array([99, 78, 78])
energy_bin_id=array([ 0,  1, 21])

We have 4 counts from our test data. We can use these indices to find the values in the histogram:
hist[150,99,0]
np.float64(1.0)
hist[269,78,1]
np.float64(1.0)
hist[269,78,21]
np.float64(2.0)

But as you can see the bins are sometimes the same so we just need to know the unique ones:
bin_id = (array([150, 269]), array([78, 99]), array([ 0,  1, 21]))

Then we have to map the original bin indices to the unique bin indices:
az_bin_id
array([150, 269, 269])
bin_id[0]
array([150, 269])
az_positions
array([0, 1, 1])

We know that the counts will have the dimension of these unique ids:
counts = np.zeros((len(bin_id[0]), len(bin_id[1]), len(bin_id[2])))
counts[az_positions, el_positions, energy_positions] = non_zero_counts

az_positions
array([0, 1, 1])
el_positions
array([1, 0, 0])
energy_positions
array([0, 1, 2])
non_zero_counts
array([1., 1., 2.])

counts[0, 1, 0] = 1
counts[1, 0, 1] = 1
counts[1, 0, 2] = 2

So now we have
bin_id
(array([150, 269]), array([78, 99]), array([ 0,  1, 21]))
counts
array([[[0., 0., 0.],
        [1., 0., 0.]],
       [[0., 1., 2.],
        [0., 0., 0.]]])

And we can use it:
For the target bin combination [150, 99, 0]:
150 in bin_id[0] corresponds to index 0 because 150 is the first element.
99 in bin_id[1] corresponds to index 1 because 99 is the second element.
0 in bin_id[2] corresponds to index 0 because 0 is the first element.

counts[0, 1, 0] = 1.0

laspsandoval:
**Update: asked Ultra if they want:**
1. a full dataset w/ all bins and then compressed cdf OR
2. a dataset w/ unpopulated bins removed

Ultra said they will need the whole grid eventually. So let's do the cdf compression. It makes the code easier.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +48,101 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+
+def cartesian_to_spherical(
+    v: tuple[np.ndarray, np.ndarray, np.ndarray],
+) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """
+    Convert cartesian coordinates to spherical coordinates.
+
+    Parameters
+    ----------
+    v : tuple[np.ndarray, np.ndarray, np.ndarray]
```
tmplummer: Is this a tuple, or a ndarray with shape = (3, n)? See below comment.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +48,101 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+    Returns
+    -------
+    az : np.ndarray
+        The azimuth angles in degrees.
+    el : np.ndarray
+        The elevation angles in degrees.
+    r : np.ndarray
+        The radii, or magnitudes, of the vectors.
+    """
+    vx, vy, vz = np.hsplit(v, 3)
```
tmplummer: I'm confused by the use of `np.hsplit`. If the input is a tuple or ndarray, can't you get the vector components just using indexing?
```
vx = v[0]
vy = v[1]
vz = v[2]
```
FWIW, we may want to pull in the fast vector normalization routine from EMUS. Then this code would become:
`vhat = -1 * normalize(v)`

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -44,11 +48,101 @@ def build_spatial_bins(spacing: float = 0.5) -> tuple[np.ndarray, np.ndarray]:
...
+    # Magnitude of the velocity vector
+    magnitude_v = np.sqrt(vx**2 + vy**2 + vz**2)
+
+    vhat_x = -vx / magnitude_v
+    vhat_y = -vy / magnitude_v
+    vhat_z = -vz / magnitude_v
+
+    # Convert from cartesian to spherical coordinates (azimuth, elevation)
+    # Radius (magnitude)
+    r = np.sqrt(vhat_x**2 + vhat_y**2 + vhat_z**2)
```
tmplummer: Since `v_hat` is normalized, `r` should/will always be 1. The correct radius should be `magnitude_v`, no?

tmplummer (commented):
I have a couple of comments on the Cartesian to spherical function that I missed the first time through. Sorry about that.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
laspsandoval: NIce catch.

tmplummer (approved):
Looks good. I will write a SPICE ticket to document that the cartesian to spherical function is here and that we should consider moving it to the geometry module.

PR #878, opened 2024-09-27
REL/CI: Add wheel build to release

greglucas:
# Change Summary

## Overview
Historically we've only been uploading [source distribution tarballs](https://pypi.org/project/imap-processing/#files), but we should actually be releasing the sdist **and** the wheel. This updates that (by removing the `--sdist` only flag) and updates to the latest packaging recommendation workflow examples here: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,5 @@
 name: Build and upload to PyPI
+# https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -10,47 +11,41 @@ on:
       - published
 
 jobs:
-
-  build_sdist:
-    name: Build source distribution
+  build:
+    name: Build distribution 📦
```
bourque, line 15: Emojis in code??? 🤯 🤯 🤯

bourque (approved):
Thanks for updating this, LGTM, and I should be able to test this out today when I make a release.

bourque:
@greglucas I hope you don't mind, I am going to merge this now so that I can get the next release made and we can test this out.

PR #881, opened 2024-09-27
SWE L1 Improvement to store acquisition time of each measurement

### LINKED ISSUE #880: SWE L1 Improvement for L2 purposes
When I met with Ruth and went through SWE L2 algorithm, we realized that the algorithm document didn't capture information about how to calculate acquisition time of each measurement. This information is needed to calculate spin phase in L2. Therefore, we discussed and I proposed a formula based on discussion. Ruth reviewed and gave feedback. This is formula we are using to calculate acquisition of each measurement.

```
sci_step_acquisition_time_sec = ACQ_START_COARSE + ACQ_START_FINE / 1000000 + (step * ( ACQ_DURATION + SETTLE_DURATION) / 1000 )
```
where step goes from 0 to 179, ACQ_START_COARSE is in seconds and ACQ_START_FINE is in microseconds and ACQ_DURATION and  SETTLE_DURATION are in milliseconds.

tech3371:
# Change Summary

## Overview
closes #880

## Updated Files
- imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
   - Added attrs for new data variable
   - Fills in checker board with acquisition time of each measurement to be used in L2

## Testing
imap_processing/tests/swe/test_swe_l1b.py

### DIFF
--- a/imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
@@ -105,6 +105,24 @@ science_data:
     Metadata field acq_duration is 17 uint. Max value of uint17 is 131071.
     Dividing max counts by acq_duration gave validmax
 
+sci_acquisition_time:
+  <<: *default
+  CATDESC: Acquisition time organized by voltage step and spin sector and CEM
+  DEPEND_1: energy
+  DEPEND_2: angle
+  DEPEND_3: cem
+  LABL_PTR_1: energy_label
+  LABL_PTR_2: angle_label
+  LABL_PTR_3: cem_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Acquisition by volt step and spin sector and CEM
+  FILLVAL: -1.0000000E+31
+  LABLAXIS: Count Acq Time
+  UNITS: sec
+  VAR_TYPE: support_data
+  VAR_NOTES: >
+    This stores acquisition times of each count data.
+
 shcoarse:
   <<: *default
   DEPEND_1: cycle
--- a/imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
@@ -105,7 +105,7 @@ science_data:
     Metadata field acq_duration is 17 uint. Max value of uint17 is 131071.
     Dividing max counts by acq_duration gave validmax
 
-sci_acquisition_time:
+sci_step_acq_time_sec:
   <<: *default
   CATDESC: Acquisition time organized by voltage step and spin sector and CEM
   DEPEND_1: energy
--- a/imap_processing/swe/l1b/swe_l1b_science.py
+++ b/imap_processing/swe/l1b/swe_l1b_science.py
@@ -219,6 +219,8 @@ def populate_full_cycle_data(
     -------
     numpy.ndarray
         Array with full cycle data populated.
+    numpy.ndarray
+        Array with acquisition times.
     """
     esa_lookup_table = read_lookup_table(esa_table_num)
 
@@ -228,6 +230,11 @@ def populate_full_cycle_data(
     if esa_table_num == 0:
         # create new full cycle data array
         full_cycle_data = np.zeros((24, 30, 7))
+        # SWE needs to store acquisition time of each count data point
+        # to use in level 2 processing to calculate
+        # spin phase. This is done below by using information from
+        # science packet.
+        acquisition_times = np.zeros((24, 30, 7))
 
         # Initialize esa_step_number and column_index.
         # esa_step_number goes from 0 to 719 range where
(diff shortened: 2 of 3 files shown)


imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
```diff
@@ -105,6 +105,24 @@ science_data:
...
+  CATDESC: Acquisition time organized by voltage step and spin sector and CEM
+  DEPEND_1: energy
+  DEPEND_2: angle
+  DEPEND_3: cem
+  LABL_PTR_1: energy_label
+  LABL_PTR_2: angle_label
+  LABL_PTR_3: cem_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Acquisition by volt step and spin sector and CEM
+  FILLVAL: -1.0000000E+31
```
laspsandoval, line 119: We still need a conversation with our whole group about how we want fill values. Didn't you use np.iinfo(np.int64).min previously for a FILLVAL?

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -233,6 +232,11 @@ def populate_full_cycle_data(
     if esa_table_num == 0:
         # create new full cycle data array
         full_cycle_data = np.zeros((24, 30, 7))
+        # SWE needs to store acquisition time of each count data point
+        # to use in level 2 processing to calculate
+        # spin phase. This is done below by using information from
+        # science packet.
+        acquisition_times = np.zeros((24, 30, 7))
```
laspsandoval: What are each of these dimensions?

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -265,6 +285,11 @@ def populate_full_cycle_data(
                     column_index += 1
                 # Put counts rate in full cycle data array
                 full_cycle_data[esa_voltage_row_index][column_index] = counts_rate[step]
+                # Put acquisition time in acquisition_times array
+                acquisition_times[esa_voltage_row_index][column_index] = (
```
laspsandoval, line 288: Why divide by 1000?

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -224,6 +221,8 @@ def populate_full_cycle_data(
     -------
     numpy.ndarray
         Array with full cycle data populated.
+    numpy.ndarray
```
laspsandoval: full_cycle_data : numpy.ndarray
Array with full cycle data populated.
acquisition_times : numpy.ndarray
Array with acquisition times.

laspsandoval (commented):
It looks like a pretty straightforward change. Just a few comments/suggestions.

imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
tech3371: Yeah, we had constants for fill values and others in a `.py` file before and then we transition to YAML. But now I don't know how to incorporate into YAML. YAML is tricky.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Because those times are in milliseconds and I need to convert it to seconds to be consistent with others.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: (energy steps, angle, cem detectors)

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -108,12 +108,9 @@ def deadtime_correction(counts: np.ndarray, acq_duration: int) -> npt.NDArray:
...
         Corrected counts.
     """
-    # deadtime will be constant once it's defined.
-    # This deadtime value is from previous mission. SWE
-    # will give new one once they have it ready.
-    # TODO: update deadtime when we get new number
-    deadtime = 1.5e-6
-    correct = 1.0 - (deadtime * counts / acq_duration)
+    # deadtime is 360 ns
+    deadtime = 360e-9 / 1e9  # convert to seconds
```
greglucas: I think this should just be, don't divide it by 1e9 again. This should get a test so we know it is being applied correctly.
```suggestion
    deadtime = 360e-9
```

imap_processing/swe/l1b/swe_l1b_science.py
greglucas: I think two separate return items is a pattern to avoid if we can. Can you make a dataset that you are passing around as a single object? Or are these two distinct functions that each return one array?

tech3371: I thought about this but I want to keep them in same function because both of them are populated same way and populating them in checker box pattern is complicated. That's why it's helpful to keep them together. I can consolidate return value into new xr.Dataset like you first suggestion. Does that sound ok?

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Thank you for catching this. I will check with Ruth about this again. She sent me this `SWE deadtime: 360 ns` and now when I am calculating corrected counts, I need to convert deadtime and acquisition times to seconds to be consistent. That's why I think I did that. But good point. I will check with Ruth.

greglucas: Yes, 360e-9 is 360 ns. You could write it as 360 / 1e9 too. I don't think you need to check this I'm confident it should be this.

imap_processing/swe/l1b/swe_l1b_science.py
greglucas: Yes, I think keeping structures that are used together in a common "container" like a Dataset makes sense here. Then you can add more named arrays to that in other routines as needed as well but keep passing around one object.

tech3371: I pushed new changes. Can you check if that's what you thought?

imap_processing/swe/l1b/swe_l1b_science.py
laspsandoval: Could you add that to the docstring just to make it easy to read?

tech3371: Yes. For sure. My last push should have made this more readable.

laspsandoval (approved):
I think you addressed Greg and my comments. I think it's good.

PR #888, opened 2024-09-27
calculate partical energy

### LINKED ISSUE #883: Calculate particle energy
Volts to Energy: Multiply ESA voltage in Volts by 4.75 to get electron energy in eV

tech3371:
# Change Summary
closes #883 

## Overview
Small PR that calculates particle energy.

## New Files
   - File to add L2 code
   - Added function to calculate energy

## Updated Files
   - minor type update
- updated file 2

## Testing
imap_processing/tests/swe/test_swe_l2.py

### DIFF
--- a/imap_processing/swe/l2/swe_l2.py
+++ b/imap_processing/swe/l2/swe_l2.py
@@ -1,0 +1,31 @@
+"""SWE L2 processing module."""
+
+import numpy.typing as npt
+
+from imap_processing.swe.l1b.swe_l1b_science import read_lookup_table
+
+
+def calculate_particle_energy(esa_table_num: int) -> npt.NDArray:
+    """
+    Calculate particle energy.
+
+    To convert Volts to Energy, multiply ESA voltage in Volts by 4.75 to get
+    electron energy in eV.
+
+    Parameters
+    ----------
+    esa_table_num : int
+        ESA table number.
+
+    Returns
+    -------
+    energy : np.ndarray
+        720 step energy values.
+    """
+    # The lookup table gives voltage applied to analyzers.
+    esa_table = read_lookup_table(table_index_value=esa_table_num)
+
+    # Convert voltage to electron energy in eV by apply conversion factor.
+    conversion_factor = 4.75
+    energy = esa_table["esa_v"] * conversion_factor
+    return energy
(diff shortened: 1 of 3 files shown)


imap_processing/swe/l2/swe_l2.py
```diff
@@ -0,0 +1,31 @@
...
+    -------
+    energy : np.ndarray
+        720 step energy values.
+    """
+    # The lookup table gives voltage applied to analyzers.
+    esa_table = read_lookup_table(table_index_value=esa_table_num)
+
+    # Convert voltage to electron energy in eV by apply conversion factor.
+    conversion_factor = 4.75
+    energy = esa_table["esa_v"] * conversion_factor
```
sdhoyt: It looks like this is applying a constant conversion to the esa table values. If that's the case, these values could be added as a new column to the table and pulled directly during processing instead of multiplying this constant conversion factor every time. Although I guess the problem with changing the file is needing to document that somewhere to do it again if the tables change some day.

sdhoyt (approved):
Looks good to me! Just left one comment that most likely won't be useful

imap_processing/swe/l2/swe_l2.py
tech3371: Good point. Unfortunately, they will likely add or change or update lookup tables and `esa_v` after launch. I was told that `esa_v` could be ordered differently and so on. If that, then keeping this here makes it easier at the moment.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -0,0 +1,31 @@
...
+    Returns
+    -------
+    energy : np.ndarray
+        720 step energy values.
+    """
+    # The lookup table gives voltage applied to analyzers.
+    esa_table = read_lookup_table(table_index_value=esa_table_num)
+
+    # Convert voltage to electron energy in eV by apply conversion factor.
+    conversion_factor = 4.75
```
tmplummer: Is this a calibration factor? I have been recommending that folks put all calibration related constants in a single place for their instrument. Just something to think about.

tmplummer (approved):
One minor comment to think about. Looks good to me though!

imap_processing/swe/l2/swe_l2.py
tech3371: Good point. This would be perfect in instrument status summary file. I will add that as TODO in there.

PR #901, opened 2024-10-01
Fixed Lo DE unpacking scheme 

### LINKED ISSUE #900: BUG - Lo Unpacking Scheme Definitions
### Description of the issue

Found issues with the Lo unpacking scheme defined in decompression_tables.py

- For coincidence type 6, mode 1: TOF3 is not transmitted, but Position is
- The checksum is now 4 bits instead of 3 bits
- Position should be 2 bits, not 1

sdhoyt:
# Change Summary

## Overview
Found issues with the Lo unpacking scheme defined in decompression_tables.py

- For coincidence type 6, mode 1: TOF3 is not transmitted, but Position is
- The checksum is now 4 bits instead of 3 bits
- Fixed mistake with the position bits saying 1 instead of 2

Closes #900

### DIFF
--- a/imap_processing/lo/l0/decompression_tables/decompression_tables.py
+++ b/imap_processing/lo/l0/decompression_tables/decompression_tables.py
@@ -6,7 +6,7 @@ from collections import namedtuple
 # must be shifted by 1 bit to the left
 DE_BIT_SHIFT = 1
 
-# This named tupled will be used to store the bit length
+# This named tuple will be used to store the bit length
 # of each TOF field. This length will be used for unpacking
 # the TOF data.
 DataFields = namedtuple(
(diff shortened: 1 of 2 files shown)


tech3371 (approved):

greglucas (approved):
How did you determine this? It looks like you're updating the tests to pass this new version as well, but I'm wondering if there is a way to have more robust tests here.

sdhoyt:
@greglucas This change is a result of updates Lo made to their packing scheme which are reflected in the document linked on this page: https://lasp.colorado.edu/galaxy/display/IMAP/Science+Direct+Events+Unpacking+Algorithm

One of those changes is the checksum will be 4 bits instead of 3, so I updated the checksum input in the unit test to be 4 bits.

The tests should be more robust once I refactor and get the new sample data working in a test. The true/false changes were because I didn't have tests for those coincidence types. The validation data should test all of them though. I ran through the validation data on a branch and found the mismatches between my TOF tuple and what's in the document from that.

PR #903, opened 2024-10-01
795 spice instrument pointing at specified time

### LINKED ISSUE #795: SPICE - Instrument pointing at specified time
### Algorithm Description:
Write code query spice in order to get an instrument's look direction at a specified time. To do this, we need to understand each instrument's frame definition and how it relates to the boresight. That info may be in the frame kernel imap_wkcp.tf

GLOWS requests this in "Cartesian heliocentric coordinates in ecliptic reference system"

tmplummer:
# Change Summary

## Overview
Add preliminary function for computing instrument pointing using spice.

## Updated Files
   - Modify frame_transform function to allow multiple time inputs with a single position vector.
   - Add simple lookup for getting instrument boresight vectors.
   - Add instrument pointing function that can return Lat/Lon or Cartesian coordinates
   - Parameterize test for frame_transform and add coverage for multiple time inputs with single position vector
   - Add test coverage for new `instrument_pointing` function

Closes: #795

### DIFF
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -241,6 +241,8 @@ def frame_transform(
         Ephemeris time(s) corresponding to position(s).
     position : npt.NDArray
         <x, y, z> vector or array of vectors in reference frame `from_frame`.
+        A single position vector may be provided for multiple `et` query times
+        but only a single position vector can be provided for a single `et`.
     from_frame : SpiceFrame
         Reference frame of input vector(s).
     to_frame : SpiceFrame
@@ -249,16 +251,12 @@ def frame_transform(
     Returns
     -------
     result : npt.NDArray
-        3d position vector(s) in reference frame `to_frame`.
+        3d Cartesian position vector(s) in reference frame `to_frame`.
     """
     if position.ndim == 1:
         if not len(position) == 3:
             raise ValueError(
                 "Position vectors with one dimension must have 3 elements."
-            )
-        if not isinstance(et, float):
-            raise ValueError(
-                "Ephemeris time must be float when single position vector is provided."
             )
     elif position.ndim == 2:
         if not position.shape[1] == 3:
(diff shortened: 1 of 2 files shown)


imap_processing/spice/geometry.py
```diff
@@ -320,3 +337,42 @@ def get_rotation_matrix(
...
+        The instrument pointing at the specified times.
+    """
+    pointing = frame_transform(
+        et, BORESIGHT_LOOKUP[instrument.name], instrument, to_frame
+    )
+    if cartesian:
+        return pointing
+    if isinstance(et, typing.Collection):
+        return np.array([spice.reclat(vec)[1:] for vec in pointing])
+    return np.array(spice.reclat(pointing)[1:])
```
tmplummer: `spice.reclat()` returns radians. These should probably be converted to degrees.

imap_processing/spice/geometry.py
```diff
@@ -62,6 +62,24 @@ class SpiceFrame(IntEnum):
     IMAP_GLOWS = -43750
 
 
+# TODO: Update boresight for in-situ instruments
+# TODO: Confirm ENA boresight vectors
+BORESIGHT_LOOKUP = {
```
tmplummer, line 67: This is my low effort initial approach to defining instrument boresights. I am open to suggestions of how to make this better, where to move this definition, etc. I still don't know if we should eventually create an instrument kernel that defines these. I need to get a better understanding of all of the requirements from all of the instruments before making that decision.

imap_processing/spice/geometry.py
greglucas:
```suggestion
        return np.rad2deg([spice.reclat(vec)[1:] for vec in pointing])
    return np.rad2deg(spice.reclat(pointing)[1:])
```

imap_processing/spice/geometry.py
greglucas: Do we have to use `.name` or can it just be the enumeration itself as the lookup?

greglucas (approved):

tech3371 (approved):
LTGM with my limited knowledge.

PR #921, opened 2024-10-02
Updated New Lo Histogram Decompression Tables 

### LINKED ISSUE #920: Update Lo to use new decompression tables
### Summary

Lo delivered new decompression tables used with the histogram and star sensor packets. These need to be updated in the repo and minor changes need to be made to index the Mean column of the table to get the decompressed value.

sdhoyt:
# Change Summary

## Overview
Added new Lo histogram decompression tables

The tables are formatted slightly differently from before. There are now min, mean, and max columns after the compressed value column. The decompressed value we're using is the mean

Closes #920


imap_processing/tests/lo/test_bit_decompression.py
```diff
@@ -11,9 +11,9 @@
 @pytest.mark.parametrize(
     ("decompression", "compressed_value", "expected_decompressed_int"),
     [
-        (Decompress.DECOMPRESS8TO16, [209, 250], [9537, 53016]),
-        (Decompress.DECOMPRESS8TO12, [205], [1229]),
-        (Decompress.DECOMPRESS12TO16, [4068], [62908]),
+        (Decompress.DECOMPRESS8TO16, [209, 250], [13426, 54336]),
```
laspsandoval, line 14: Quick question: why did the values in the tests change?

laspsandoval (commented):
Quick question. Besides that looks good.

imap_processing/tests/lo/test_bit_decompression.py
sdhoyt: These tables are using a different compression than the previous tables so the decompressed values are different

tmplummer (approved):
LGTM

PR #971, opened 2024-10-07
Lo L1A Direct Event Refactor

### LINKED ISSUE #862: Refactor Lo L1A Direct Event Data Class to Function
### Summary

the Lo L1A Direct Events data class should be refactored and put into a function run on the dataset created by the packets_to_dataset function. To keep the resulting PR small, this change should only include the refactoring of the data class. 
Separate issues/PRs should cover running the new function in lo_l1a.py and adding the CDF attributes to the data arrays.

### Requirements

_No response_

### Related tickets

_No response_

sdhoyt:
# Change Summary

## Overview
This PR refactors the Lo L1A Direct Event algorithm from a data class to a function to later make use of the `packets_to_dataset()`. To reduce the PR size, the algorithm isn't  yet called by the `lo_l1a.py` script. That will be done in a separate PR.

Closes #862


imap_processing/lo/l0/lo_science.py
```diff
@@ -148,3 +153,167 @@ def decompress(
...
+    for field in de_fields:
+        dataset[field] = xr.DataArray(
+            np.full(num_de, attr_mgr.get_variable_attributes(field)["FILLVAL"]),
+            dims="direct_events",
+        )
+
+    # The DE index for the entire pointing
+    pointing_de = 0
+    # for each direct event packet in the pointing
+    for pkt_idx, de_count in enumerate(dataset["count"].values):
```
maxinelasp, line 196: I would prefer for this section to be in a different method - mainly for documentation. Overall it's pretty clear what's happening here, though! Is it possible to loop through the values of `DATA_BITS` rather than writing them all out here? You have a lot of `dataset["{val}"] = parse_de_bin(dataset, pkt_idx, bit_pos, {val}` that could reduce duplication. The last set of if statements can probably work the same way.

maxinelasp: If you do greatly simplify this for loop then it's fine in this method

sdhoyt: Agreed. I'll clean this up

maxinelasp (approved):
LGTM thanks for tidying!

sdhoyt:
> LGTM thanks for tidying!

@maxinelasp Thanks! I also just used your looping through the data bits fields idea and it's cleaned up even more now. Thanks for the comment!

imap_processing/tests/lo/test_lo_science.py
tmplummer, line 1: Really nice tests! Great job generating fake data that you can use for testing in a very deterministic way.

tmplummer (approved):
This all looks really good to me.

imap_processing/tests/lo/test_lo_science.py
sdhoyt: Thanks!

PR #991, opened 2024-10-08
MNT: Cdflib update

### LINKED ISSUE #606: Update cdflib version
### Algorithm Description:
Update to the latest version of cdflib. The major relevant changes to be aware of are:

### Other Notes:
~~- The code that sets the `to_datetime` keyword argument in `imap_processing.cdf.utils.load_cdf` should be removed.~~
`to_datetime` was already removed as part of another PR.

### Related Issues/PRs:
~~#595 - this was the ticket that added defaulting to `to_datetime = True`~~

tech3371:
# Change Summary
closes #606 

## Overview
Updated cdflib to it's latest version. Made changes where I can but some was outside of what I can do. For those, I marked pytest to skip and need to address in future PR.

## Updated Files
Instruments updated
- CoDICE
   - removed label axis because many of variable using the default has more than one dimension. As a result, it would error because ISTP doesn't allow to set both label axis and PTR label.
   - skipped few tests because sample data was not increasing monotonically.
- GLOWS
   - added changes to include LABL_PTR_1
 - IMAP-Hi
   - added changes to include LABL_PTR_1 
- MAG
   - added changes to include LABL_PTR_1
- SWE
   - removed label axis because ISTP doesn't allow to set both label axis and PTR label.
- imap_processing/cdf/config/shared/default_variable_cdf_attrs_schema.yaml
   - set some of attrs's `required` to `false` because some those are only required for `epoch` variable and not other data type. Also, `LABLAXIS` is not required if 'LABL_PTR_i` is set. These will be taken care of in `sammi`. Changes I made is temporary fixes.

## Testing
I had to skip few tests due to data not increasing monotonically. I will create issue for those that needs update.

### DIFF
--- a/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
@@ -1,11 +1,3 @@
-# <=== Label Attributes ===>
-# LABL_PTR_i expects VAR_TYPE of metadata with char data type
-hi_hist_angle_label:
-  CATDESC: Angle bin centers for histogram data.
-  FIELDNAM: ANGLE
-  FORMAT: A5
-  VAR_TYPE: metadata
-
 # ------- Default attributes section -------
 default_attrs: &default
   DEPEND_0: epoch
@@ -165,6 +157,13 @@ hi_de_tof_3:
     the direct event. 1023 is the value used to indicate no event was registered.
 
 # ======= L1A HIST Section =======
+# <=== Label Attributes ===>
+# LABL_PTR_i expects VAR_TYPE of metadata with char data type
+hi_hist_angle_label:
+  CATDESC: Angle bin centers for histogram data.
+  FIELDNAM: ANGLE
+  FORMAT: A5
+  VAR_TYPE: metadata
 
 hi_hist_angle:
   SCALE_TYPE: linear
--- a/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
@@ -1,3 +1,11 @@
+# <=== Label Attributes ===>
+# LABL_PTR_i expects VAR_TYPE of metadata with char data type
+hi_hist_angle_label:
+  CATDESC: Angle bin centers for histogram data.
+  FIELDNAM: ANGLE
+  FORMAT: A5
+  VAR_TYPE: metadata
+
 # ------- Default attributes section -------
 default_attrs: &default
   DEPEND_0: epoch
@@ -183,8 +191,8 @@ hi_hist_counters:
   FIELDNAM: "{counter_name} histogram"
   VALIDMAX: 4095
   DEPEND_1: angle
+  LABL_PTR_1: angle_label
   FORMAT: I4
-  LABLAXIS: "{counter_name}"
 
 # ======= L1B DE Section =======
 hi_de_coincidence_type:
--- a/imap_processing/hi/l1a/histogram.py
+++ b/imap_processing/hi/l1a/histogram.py
@@ -111,7 +111,7 @@ def allocate_histogram_dataset(num_packets: int) -> xr.Dataset:
     coords["angle_label"] = xr.DataArray(
         coords["angle"].data.astype(str),
         name="angle_label",
-        dims=["angle_label"],
+        dims=["angle"],
         attrs=attr_mgr.get_variable_attributes("hi_hist_angle_label"),
     )
 
--- a/imap_processing/hi/l1a/histogram.py
+++ b/imap_processing/hi/l1a/histogram.py
@@ -108,6 +108,13 @@ def allocate_histogram_dataset(num_packets: int) -> xr.Dataset:
         dims=["angle"],
         attrs=attr_mgr.get_variable_attributes("hi_hist_angle"),
     )
+    coords["angle_label"] = xr.DataArray(
+        coords["angle"].data.astype(str),
+        name="angle_label",
+        dims=["angle_label"],
+        attrs=attr_mgr.get_variable_attributes("hi_hist_angle_label"),
+    )
+
     data_vars = dict()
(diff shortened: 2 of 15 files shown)


maxinelasp (approved):
Why doesn't LABLAXIS work here? Is that an error out of cdflib? Overall it looks good to me, I'm just wondering why  LABLAXIS doesn't work...

tech3371:
> Why doesn't LABLAXIS work here? Is that an error out of cdflib? Overall it looks good to me, I'm just wondering why LABLAXIS doesn't work...

@maxinelasp There is two parts to this. If schema defines a required field, it will add that and fill it with empty string even if we don't include it our own YAML file. This happens from this [line](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/cdf/cdf_attribute_manager.py#L311). Since ISTP doesn't allow to set both LABLAXIS and LABL_PTR_i, cdflib will complain that we had set  LABLAXIS and LABL_PTR_i even though user didn't but schema did.

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -89,6 +89,7 @@ def test_l1a_data(request) -> xr.Dataset:
     return dataset
 
 
+@pytest.mark.xfail(reason="Epoch variable data needs to monotonically increase")
```
tech3371, line 92: I created new issue to capture this #992

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -167,6 +169,7 @@ def test_l1a_data_array_values(test_l1a_data: xr.Dataset, validation_data: Path)
             )
 
 
+@pytest.mark.xfail(reason="Epoch variable data needs to monotonically increase")
```
tech3371, line 172: captured this work in #993

maxinelasp:
But, if we mark both as `required: false` then do we need to update the yaml files? I thought that was the only change needed.

tech3371:
> But, if we mark both as `required: false` then do we need to update the yaml files? I thought that was the only change needed.

Copy of Slack conversation in case someone is wondering same thing.
> (Tenzin) If I understand what you are thinking, are you thinking that we set both false and then update our own attrs YAML to include LABLAXIS or LABL_PTR_i? If so, yes. That’s what I did in that PR with data variable who had more than one dimensions. cdflib requires to add LABL_PTR_i for data variable who had more than one dimension. Therefore, I had to make those additional changes to be ISTP compliant. For 2d or higher dimension data, we remove LABLAXIS and replace with LABL_PTR_i.

imap_processing/hi/l1a/histogram.py
```diff
@@ -108,6 +108,13 @@ def allocate_histogram_dataset(num_packets: int) -> xr.Dataset:
         dims=["angle"],
         attrs=attr_mgr.get_variable_attributes("hi_hist_angle"),
     )
+    coords["angle_label"] = xr.DataArray(
```
tmplummer: Label variables should not be added to the coordinates list. They are just variables.

imap_processing/cdf/config/imap_hi_variable_attrs.yaml
```diff
@@ -1,3 +1,11 @@
+# <=== Label Attributes ===>
```
tmplummer: Minor nit, but I would like to move this to the histogram section of this file.

imap_processing/hi/l1a/histogram.py
```diff
@@ -108,6 +108,13 @@ def allocate_histogram_dataset(num_packets: int) -> xr.Dataset:
...
+        coords["angle"].data.astype(str),
+        name="angle_label",
+        dims=["angle_label"],
```
tmplummer:
```suggestion
        dims=["angle"],
```

tmplummer (commented):
Thanks for taking this on! I have a couple of nit picks for Hi changes. One thing that we should make consistent across all instruments is that label variables should be variables, not coordinates. It would make sense to put that in it's own PR.

For my Hi nits, I am happy to make a commit with the changes since they are purely stylistic things.

imap_processing/hi/l1a/histogram.py
tech3371: This LABL_PTR_i confused me a lot. Changes in this file and XML file is related to that. This is something SPDF and ISTP caused. When your data has great than 1D array, it requires to have label for each `DEPEND_1` and onward. And we need to create xr.DataArray for those labels and need to add those to coordinate, otherwise `cdflib` will throw error.
```
FAILED imap_processing/tests/hi/test_l1a.py::test_app_hist_decom - cdflib.xarray.xarray_to_cdf.ISTPError: ISTP Compliance Warning: variable qual_ab listed angle_label as its LABL_PTR_1.  However, it was not found in the dataset.
```

tmplummer (approved):
Thanks for doing this!

imap_processing/hi/l1a/histogram.py
greglucas:
```suggestion
    data_vars["angle_label"] = xr.DataArray(
```
I think that is what @subagonsouth  is suggesting. cdflib fails with that as well? Coordinates should be optional, so this might be an issue to raise with cdflib.

tech3371: moving this to data variable. I will refactor other instrument's label in new PR. I do want to point out that other past mission did store their coordinate label in coordinate as well. I wonder if they did that to be able to select and filter data easier later on. That's something we should think about. I will check with Andriy first before I refactor other instrument.

imap_processing/tests/hi/test_l1a.py
```diff
@@ -62,6 +63,9 @@ def test_app_nhk_decom(hi_l0_test_data_path):
     assert cem_raw_cdf_filepath.name == "imap_hi_l1a_45sensor-hk_20100313_v001.cdf"
 
 
+@pytest.mark.skip(
```
tech3371, line 66: captured this work in #1009

PR #1007, opened 2024-10-10
MNT: space_packet_parser update

### LINKED ISSUE #1008: Update space_packet_parser to version 5.0.0
Update version and codes that was effect by new release.

tech3371:
# Change Summary
closes #1008 

## Overview
This PR updates `space_packet_parser` to new release of space_packet_parser==5.0.0. Gavin helped fix bug and release patch version `5.0.1`. This PR is large but very important to get this done early to clear roadblocks for all of us. Almost all instrument code had to change due to the new release.

This new release includes tons of breaking changes. Please see its [release notes](https://github.com/medley56/space_packet_parser/releases/tag/5.0.0) for more details. It's posted on `xtce` Slack channel as well. These changes effect us most:
- Create only a definition from your XTCE document and instead of `my_parser.generator`, use `my_packet_definition.packet_generator`
- The ``ParsedDataItem`` class has been removed and the derived values are being returned now.
  The ``raw_value`` is stored as an attribute on the returned object. **(My clarification of what this means to us) - The returned "value" defaults to raw value if there is no different derived value but if there is derived value then it uses that as default value. Therefore, we need to specific `value.raw_value` if we do want raw value of variable that also has derived value. Please reach out to Gavin and me if you have question.**
- Replace `bitstring` objects with native Python bytes objects. **(My clarification of how this effects us) - This means that it returns `<class bytes>` instead of binary string. Before it was returning binary string and many of our algorithm was written with that information. Now, we need to be aware of this new change. I tried to keep lot of your code same by adding below code blocks that converts bytes to binary string. We will need to revisit as needed. This allowed me to keep scope of this PR to only cdflib update.**
```
            # TODO: improve this as needed
            science_values = "".join(f"{byte:08b}" for byte in science_values)
```
- (This was not included in release note but important to us) Before, unpacked data can be accessed using `value.header` and `value.data`. In new release, it has changed to `value.user_data` instead of `value.data`.

## Updated Files
- updates every instruments files with changes mentioned above. I like request everyone to review this PR.


imap_processing/ccsds/ccsds_data.py
```diff
@@ -42,10 +42,7 @@ class CcsdsData:
     def __init__(self, packet_header: dict):
         attributes = [field.name for field in fields(self)]
 
-        for key, item in packet_header.items():
-            value = (
-                item.derived_value if item.derived_value is not None else item.raw_value
-            )
+        for key, value in packet_header.items():
             if key in attributes:
```
maxinelasp, line 46: I think we still need to get `value.raw_value` here... won't value be of type `packets.IntParameter` instead of `int`?

imap_processing/cdf/config/imap_ultra_l1a_variable_attrs.yaml
```diff
@@ -7,39 +7,42 @@ ultra_metadata_attrs:
   FORMAT: I19
   LABLAXIS: Metadata
   SCALE_TYP: linear
-  UNITS: " "
+  UNITS: ” ”
```
maxinelasp: Why is this changed?

imap_processing/codice/utils.py
```diff
@@ -88,7 +88,7 @@ def add_metadata_to_array(packet: space_packet_parser, metadata_arrays: dict) ->
     for key, value in packet.header.items():
         metadata_arrays.setdefault(key, []).append(value.raw_value)
 
-    for key, value in packet.data.items():
+    for key, value in packet.user_data.items():
```
maxinelasp: Similarly, are the types here correct?

imap_processing/glows/l0/decom_glows.py
```diff
@@ -49,39 +49,28 @@ def decom_packets(
...
                 histdata.append(hist_l0)
 
             if apid == GlowsParams.DE_APID.value:
-                values = [
-                    item.derived_value
-                    if item.derived_value is not None
-                    else item.raw_value
-                    for item in packet.data.values()
-                ]
+                values = [item for item in packet.user_data.values()]
```
maxinelasp:
```suggestion
                values = [item.raw_value for item in packet.user_data.values()]
```
Missed that one! Sorry about that

imap_processing/idex/idex_l1a.py
```diff
@@ -131,9 +131,9 @@ def __init__(self, packet_file: Union[str, Path], data_version: str) -> None:
 
         dust_events = {}
         for packet in decom_packet_list:
-            if "IDX__SCI0TYPE" in packet.data:
-                scitype = packet.data["IDX__SCI0TYPE"].raw_value
-                event_number = packet.data["IDX__SCI0EVTNUM"].derived_value
+            if "IDX__SCI0TYPE" in packet.user_data:
+                scitype = packet.user_data["IDX__SCI0TYPE"].raw_value
+                event_number = packet.user_data["IDX__SCI0EVTNUM"]
```
maxinelasp:
```suggestion
                event_number = packet.user_data["IDX__SCI0EVTNUM"].raw_value
```
?

imap_processing/idex/idex_l1a.py
```diff
@@ -312,9 +312,9 @@ def _set_impact_time(self, packet: space_packet_parser.parser.Packet) -> None:
               testing.
         """
         # Number of seconds since epoch (nominally the launch time)
-        seconds_since_launch = packet.data["SHCOARSE"].derived_value
+        seconds_since_launch = packet.user_data["SHCOARSE"]
```
maxinelasp: Do we need `.raw_value` or `.derived_value` here? I don't fully understand when the `_Parameter` type is returned...

imap_processing/tests/glows/test_glows_decom.py
```diff
@@ -64,51 +63,52 @@ def test_bad_header():
 
 
 def test_header(decom_test_data):
-    expected_hist = CcsdsData(
```
maxinelasp, line 67: We can fix this by adding a type check in `CcsdsData.__init__()` before it tries to retrieve the `raw_value`. Or, we might be able to update `CcsdsData` to remove the `raw_value` step altogether and then pass in a dict here? Again, I'm not clear when a `_Parameter` type is returned that needs to be opened up, vs just returning a direct python type (ie `int`)

imap_processing/utils.py
```diff
@@ -384,6 +386,23 @@ def packet_file_to_datasets(
         time_key = next(iter(data.keys()))
         # Convert to J2000 time and use that as our primary dimension
         time_data = met_to_j2000ns(data[time_key])
+        # data_dict = {}
```
maxinelasp: Remove?

maxinelasp (commented):
I have some questions about the types that are returned as part of `CcsdsPacket`....

imap_processing/ccsds/ccsds_data.py
tech3371: If I understood the original if statement, it seems like we want to get derived value if it's available or else get raw value. Now with new changes, we get derived value if it's available or else we get raw value. so getting the `value` will do what the if statement was doing before, if I understood correctly. That's how I understood at least when Gavin explain in the Slack thread. should we double check to be safe?

imap_processing/cdf/config/imap_ultra_l1a_variable_attrs.yaml
tech3371: I think it's from copy and paste between slack and Mac. Good catch. I will change it back.

imap_processing/codice/utils.py
tech3371: should we check with Gavin?

imap_processing/ccsds/ccsds_data.py
maxinelasp: If you could just print out the type and validate that it's a python type, that's good enough for me

imap_processing/codice/utils.py
maxinelasp: It would be nice to get clarification from him

imap_processing/ccsds/ccsds_data.py
greglucas: `packets.IntParameter` is a subclass of `int`

imap_processing/codice/utils.py
greglucas: Can't this be removed entirely because it is in `packet_file_to_datasets()` now? I thought codice was using that...

imap_processing/glows/l0/decom_glows.py
```diff
@@ -49,39 +49,28 @@ def decom_packets(
...
 
     filename = packet_file_path.name
 
     with open(packet_file_path, "rb") as binary_data:
-        glows_packets = glows_parser.generator(binary_data)
+        glows_packets = packet_definition.packet_generator(binary_data)
 
         for packet in glows_packets:
-            apid = packet.header["PKT_APID"].derived_value
+            apid = packet.header["PKT_APID"]
```
greglucas:
```suggestion
            apid = packet["PKT_APID"]
```

imap_processing/glows/l0/decom_glows.py
```diff
@@ -49,39 +49,28 @@ def decom_packets(
...
             # Do something with the packet data
             if apid == GlowsParams.HIST_APID.value:
-                values = [
-                    item.derived_value
-                    if item.derived_value is not None
-                    else item.raw_value
-                    for item in packet.data.values()
-                ]
+                values = [item.raw_value for item in packet.user_data.values()]
```
greglucas, line 66: Before this was getting the derived value if it is not None. The `if item.derived_value is not None else item.raw_value` is taken care of for you by space_packet_parser now.

I'm pretty sure this is the equivalent to before now:
```suggestion
                values = list(packet.user_data.values())
```

imap_processing/glows/l0/decom_glows.py
greglucas: See comment above. I think this shouldn't be `raw_value` based on the previous logic.

imap_processing/hit/l0/utils/hit_base.py
greglucas, line 1: @vmartinez-cu and @sdhoyt I thought this could be removed now that we have switched over to `packet_file_to_dataset` on HIT?

imap_processing/idex/idex_l1a.py
```diff
@@ -131,9 +131,9 @@ def __init__(self, packet_file: Union[str, Path], data_version: str) -> None:
 
         dust_events = {}
         for packet in decom_packet_list:
-            if "IDX__SCI0TYPE" in packet.data:
-                scitype = packet.data["IDX__SCI0TYPE"].raw_value
-                event_number = packet.data["IDX__SCI0EVTNUM"].derived_value
+            if "IDX__SCI0TYPE" in packet.user_data:
```
greglucas: We should look up in the packet in general, not just the user_data. The user_data adds extra overhead to the lookup because it creates a new dictionary under the hood.
```suggestion
            if "IDX__SCI0TYPE" in packet:
```

imap_processing/idex/idex_l1a.py
greglucas: The packet items are the derived value now. If there was no "derived" / "calibration" going on, then it is identical to the raw_value. So this should be left without the `raw_value` attribute.

imap_processing/idex/idex_l1a.py
greglucas: Basically, I don't think we should ever need to look at header/user_data. Or we should avoid it as much as possible. Just consider everything a part of the "packet" structure.
```suggestion
        seconds_since_launch = packet["SHCOARSE"]
```

greglucas: The `_Parameter` is the derived value. But, SHCOARSE is never derived it is just a counter, so it is the same as the raw value too.

imap_processing/mag/l0/decom_mag.py
```diff
@@ -41,24 +41,18 @@ def decom_packets(packet_file_path: str | Path) -> dict[str, list[MagL0]]:
...
-            apid = packet.header["PKT_APID"].derived_value
+            apid = packet.header["PKT_APID"]
             if apid in (Mode.BURST, Mode.NORMAL):
-                values = [
-                    item.derived_value
-                    if item.derived_value is not None
-                    else item.raw_value
-                    for item in packet.data.values()
-                ]
+                values = [item.raw_value for item in packet.user_data.values()]
```
greglucas, line 55:
```suggestion
                values = list(packet.user_data.values())
```

imap_processing/swe/l1a/swe_science.py
```diff
@@ -119,9 +119,7 @@ def swe_science(l0_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
     # 4. Reshape the data to 180 x 7
     raw_science_array = np.array(
         [
-            np.frombuffer(
-                int(binary_string, 2).to_bytes(1260, byteorder="big"), dtype=np.uint8
-            ).reshape(180, 7)
+            np.frombuffer(binary_string, dtype=np.uint8).reshape(180, 7)
```
greglucas, line 122: Nice! This is what I was hoping for with the binary parameter types, being able to avoid the conversion to string and back again.

imap_processing/tests/glows/test_glows_decom.py
```diff
@@ -64,51 +63,52 @@ def test_bad_header():
...
-            "SEC_HDR_FLG": ParsedDataItem("SEC_HDR_FLG", 1, unit=None),
-            "PKT_APID": ParsedDataItem("PKT_APID", 1481, unit=None),
-            "SEQ_FLGS": ParsedDataItem("SEQ_FLGS", 3, unit=None),
-            "SRC_SEQ_CTR": ParsedDataItem("SRC_SEQ_CTR", 0, unit=None),
-            "PKT_LEN": ParsedDataItem("PKT_LEN", 2775, unit=None),
-        }
-    )
-
-    assert expected_de == decom_test_data[1][0].ccsds_header
+    print(decom_test_data[0][0].ccsds_header)
```
greglucas: Remove the print and reinstate these tests?

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -88,11 +88,11 @@ def append_params(decom_data: dict, packet: Packet) -> None:
...
     decom_data : dict
         Dictionary to which the data is appended.
-    packet : space_packet_parser.parser.Packet
+    packet : space_packet_parser.packets.CCSDSPacket
         Individual packet.
     """
-    for key, item in packet.data.items():
-        decom_data[key].append(item.derived_value)
+    for key, item in packet.user_data.items():
+        decom_data[key].append(item.raw_value)
```
greglucas: There is a lot of switching from derived previously to raw now which I think is undesired.

```suggestion
    for key, value in packet.user_data.items():
        decom_data[key].append(value)
```

imap_processing/utils.py
```diff
@@ -342,34 +344,34 @@ def packet_file_to_datasets(
...
         for packet in packet_generator:
             apid = packet.header["PKT_APID"].raw_value
             if apid not in data_dict:
                 # This is the first packet for this APID
                 data_dict[apid] = collections.defaultdict(list)
                 datatype_mapping[apid] = dict()
-                variable_mapping[apid] = packet.data.keys()
-            if variable_mapping[apid] != packet.data.keys():
+                variable_mapping[apid] = packet.user_data.keys()
+            if variable_mapping[apid] != packet.user_data.keys():
```
greglucas:
```suggestion
                variable_mapping[apid] = packet.keys()
            if variable_mapping[apid] != packet.keys():
```

imap_processing/utils.py
```diff
@@ -342,34 +344,34 @@ def packet_file_to_datasets(
...
+                    f"\nExpected: {variable_mapping[apid]},\n"
+                    f"got: {packet.user_data.keys()}"
                 )
 
             # TODO: Do we want to give an option to remove the header content?
-            packet_content = packet.data | packet.header
+            packet_content = packet.user_data | packet.header
 
             for key, value in packet_content.items():
                 val = value.raw_value
```
greglucas: packet_content is just the packet here now, no need to combine them together anymore.
```suggestion
            for key, value in packet.items():
                val = value if use_derived_value else value.raw_value
```

imap_processing/utils.py
```diff
@@ -342,34 +344,34 @@ def packet_file_to_datasets(
...
                 if use_derived_value:
                     # Use the derived value if it exists, otherwise use the raw value
-                    val = value.derived_value or val
+                    val = value
```
greglucas: I think this needs to invert the logic and can be combined with the above logic.

greglucas (commented):
You should be able to remove almost all `.header` and `.user_data` attribute accessors and rely on just the `packet[lookup_value]` directly instead.

imap_processing/glows/l0/decom_glows.py
maxinelasp: This was giving me errors before because it was assigning everything as `packet.IntParameter` and messing with the types

imap_processing/hit/l0/utils/hit_base.py
vmartinez-cu: hit_base.py is still used by hit_l1b.py which still uses data classes to produce the l1b housekeeping product. I haven't had a chance to refactor l1b yet. I can do this in a couple of weeks since I'll be out 10/14-10/25

imap_processing/hit/l0/decom_hit.py
```diff
@@ -357,7 +357,11 @@ def assemble_science_frames(sci_dataset: xr.Dataset) -> xr.Dataset:
     # Convert sequence flags and counters to NumPy arrays for vectorized operations
     seq_flgs = sci_dataset.seq_flgs.values
     seq_ctrs = sci_dataset.src_seq_ctr.values
-    science_data = sci_dataset.science_data.values
+    # TODO: improve this as needed
+    binary_str_val = []
+    for data in sci_dataset.science_data.values:
+        binary_str_val.append("".join(f"{byte:08b}" for byte in data))
+    science_data = binary_str_val
```
vmartinez-cu: Just curious why this update is needed. Did changes to space packet parser result in returning a list of bytes instead of a full binary string?

pyproject.toml
```diff
@@ -31,8 +31,8 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-cdflib = "==1.3.1"
-imap-data-access = ">=0.5.0"
+cdflib = ">=1.3.1"
```
maxinelasp: Do we want to make this a karat dependency instead? That way, we won't get unexpected major version updates, since they tend to need some code changes.

```suggestion
cdflib = "^1.3.1"
```

imap_processing/hit/l0/decom_hit.py
maxinelasp: @vmartinez-cu  Yes, now it returns a list of bytes

imap_processing/ccsds/ccsds_data.py
tech3371: in this case, are we good with current change?

maxinelasp: Well, if it's a different type, that's going to mess with some code, right? Some of my code depends on the types being accurate because of MyPy

imap_processing/codice/utils.py
tech3371: yes. you are right Greg. I can delete this function. It's not used anymore. Sorry for including you to this PR by mistake.

imap_processing/ccsds/ccsds_data.py
maxinelasp: I guess, if you make the change and all the tests pass, then it's probably fine... but I'm just worried about changing all the types out from underneath everyone like this, even if they're basically the same as the original types. But perhaps that is just Python offending my delicate, type-loving sensibilities again

imap_processing/codice/utils.py
tech3371: To your question Maxine, data type of both `header` and `user_data` are dict from [this](https://github.com/medley56/space_packet_parser/blob/main/space_packet_parser/packets.py#L126)

imap_processing/glows/l0/decom_glows.py
tech3371: @maxinelasp what do you think here? I thought same as Greg but you helped me debug some of GLOWS' error and I saw that you had to get `raw_value`. Is that right?

imap_processing/glows/l0/decom_glows.py
tech3371: nvm, you did explain here.

imap_processing/hit/l0/decom_hit.py
tech3371: yes. you got it right. If your data is broken down in 8bits chunks, then you can use return value of `space_packet_parser` directly. But for HIT case, I think we do need to convert bytes back to binary strings since your unpacking is little complicated.

imap_processing/glows/l0/decom_glows.py
maxinelasp: Yes, when I printed out the types here it was all `IntParameter` (etc) instead of `int`.

imap_processing/codice/utils.py
maxinelasp: But what is inside the dicts? Python types? I did go look at that code but I'm still confused about the parameter types

imap_processing/idex/idex_l1a.py
tech3371: This is my understanding as well. That's why I removed `.derived_value` where it was used before.

imap_processing/idex/idex_l1a.py
tech3371: interesting...this does simplifies a lot. I am happy to do this change if others are ok with it.

imap_processing/mag/l0/decom_mag.py
tech3371: I will let @maxinelasp respond to this. I think she had reasons for doing this.

imap_processing/swe/l1a/swe_science.py
tech3371: It was nice for SWE but for some other instrument, unpacking scheme doesn't line up good with bytes. That's why I had to convert bytes to binary string for those.

imap_processing/utils.py
tech3371: I can't make all of these change because we had reason for flipping these two out. We use first key as `time_key` to calculate epoch time. I can make the if else statement change though.

imap_processing/ultra/l0/decom_ultra.py
tech3371: good catch!

imap_processing/cdf/config/imap_ultra_l1a_variable_attrs.yaml
tech3371, line 1: @laspsandoval After reverting some of `"` changes, I came across this error similar to what we saw couple days ago.
```
E                   ValueError: Unable to convert data to CDF format, data object cannot be of type string.
```
I did make changes to this file again since it seems like you did want your data in derived value. Please do revisit this in future PR if that's not what you want. It seems to involve more changes and I am worried I will mess something.

tech3371: Now I understand why it failed last time. It needed `support_data` type instead of `metadata` type since that's what it was. That's why I made that change in this PR.

imap_processing/idex/idex_l1a.py
tech3371: I did make changes where it was appropriate and left others as it is.

imap_processing/glows/l0/decom_glows.py
greglucas:
```python
class IntParameter(int):
```
`IntParameter` is a subclass of `int` so if you type it as `int` `IntParameter` should be acceptable as well? Alternatively, could you type it as `IntParameter` where this is needed?

`item.raw_value` is very different from `item`.

`item.derived_value` from v4.x is the same as `item` in v5.x

imap_processing/ccsds/ccsds_data.py
greglucas: Sure, then I think you can change the types to `IntParameter` now. But note that `raw_value` is not the same as `derived_value` which was being used before if it was present.

imap_processing/idex/idex_l1a.py
```diff
@@ -131,9 +131,9 @@ def __init__(self, packet_file: Union[str, Path], data_version: str) -> None:
 
         dust_events = {}
         for packet in decom_packet_list:
-            if "IDX__SCI0TYPE" in packet.data:
-                scitype = packet.data["IDX__SCI0TYPE"].raw_value
-                event_number = packet.data["IDX__SCI0EVTNUM"].derived_value
+            if "IDX__SCI0TYPE" in packet:
+                scitype = packet["IDX__SCI0TYPE"].raw_value
```
greglucas: I'm guessing you can remove some of these `raw_value`s as well here.
```suggestion
                scitype = packet["IDX__SCI0TYPE"]
```

imap_processing/idex/idex_l1a.py
```diff
@@ -488,18 +484,23 @@ def _calc_high_sample_resolution(self, num_samples: int) -> npt.NDArray:
...
             A single science data packet for one of the 6.
             IDEX observables.
         """
-        scitype = packet.data["IDX__SCI0TYPE"].raw_value
-        raw_science_bits = packet.data["IDX__SCI0RAW"].raw_value
+        scitype = packet["IDX__SCI0TYPE"].raw_value
+        # TODO: improve this as needed
+        raw_science_bits = "".join(
+            f"{byte:08b}" for byte in packet["IDX__SCI0RAW"].raw_value
+        )
```
greglucas: Rather than the iteration, I think we should try to do a direct conversion here (and all other bitstring conversions):
```suggestion
        raw_science_bits = f"{int.from_bytes(packet['IDX__SCI0RAW'], byteorder='big'):0{len(packet['IDX__SCI0RAW'])*8}b}"
```

I should have mentioned this before leaving, but I think the easier change everywhere would have been to put this string conversion in the `packet_file_to_datasets()` function to keep the same output there that everyone could use. Then do all of these individual changes in another follow-up. Sorry for all the work here!

imap_processing/codice/utils.py
tech3371: This may be a Gavin question :)

tech3371:
pre-commit.ci autofix

imap_processing/codice/utils.py
greglucas: subclasses of built-in types.

An `IntParameter` is simply an `int` with a `raw_value` attribute attached to it. You should be able to use it exactly like an int, it just has an extra attribute to help you get at the raw_value if you really want it.

imap_processing/idex/idex_l1a.py
tech3371: I did think about it but I don't know if I want to include this in `packet_file_to_datasets()` function though. It will complicate things more, I think. I could create new function in `imap_processing/utils.py` that take in bytes and return binary string if we want.

imap_processing/idex/idex_l1a.py
tech3371: I see your point but I don't want to assume and like to leave that to IDEX team to decide.

imap_processing/hit/l0/decom_hit.py
```diff
@@ -357,7 +359,11 @@ def assemble_science_frames(sci_dataset: xr.Dataset) -> xr.Dataset:
     # Convert sequence flags and counters to NumPy arrays for vectorized operations
     seq_flgs = sci_dataset.seq_flgs.values
     seq_ctrs = sci_dataset.src_seq_ctr.values
-    science_data = sci_dataset.science_data.values
+    # TODO: improve this as needed
+    binary_str_val = []
+    for data in sci_dataset.science_data.values:
+        binary_str_val.append(convert_to_binary(data))
+    science_data = binary_str_val
```
greglucas:
```suggestion
    science_data = [convert_to_binary(data)) for data in sci_dataset.science_data.values]
```

imap_processing/idex/idex_l1a.py
greglucas: I think you should be opinionated here and do things the "right" way because this was only done the other way due to the old space_packet_parser way of handling things. This is the new way and now that you are doing the update you should (IMO) update these for all the teams. (Otherwise it will never be done later)

imap_processing/idex/idex_l1a.py
greglucas: Did you want to use your new function here?

imap_processing/lo/l0/lo_science.py
```diff
@@ -70,7 +71,11 @@ def parse_histogram(dataset: xr.Dataset, attr_mgr: ImapCdfAttributes) -> xr.Data
     dataset : xr.Dataset
         Parsed and decompressed histogram data.
     """
-    hist_bin = dataset.sci_cnt
+    # TODO: improve this as needed
+    binary_str_val = []
+    for data in dataset.sci_cnt.values:
+        binary_str_val.append(convert_to_binary(data))
+    hist_bin = binary_str_val
```
greglucas: list comprehensions are faster and easier to read too without needing multiple variables.
```suggestion
    hist_bin = [convert_to_binary(data) for data in dataset.sci_cnt.values
```

imap_processing/utils.py
```diff
@@ -411,3 +323,22 @@ def packet_file_to_datasets(
...
+        List of bytes to convert.
+
+    Returns
+    -------
+    binary_data : str
+        The binary string data.
+    """
+    binary_data = "".join(f"{byte:08b}" for byte in bytes_data)  # type: ignore [str-bytes-safe]
+
+    return binary_data
```
greglucas: This still isn't the same as the changelog suggestion. Why does it accept list of bytes and not just bytes?
https://github.com/medley56/space_packet_parser/blob/502159d03b2f45e18ac0311f89d65108cd2a1f57/docs/source/changelog.md?plain=1#L32

Also, this should be named "binary_string" or something with string in it because binary is what we are coming _from_ not what we are going _to_

```suggestion
def convert_to_binary_string(data: bytes) -> str:
    """
    Convert bytes to a string representation.

    Parameters
    ----------
    data : bytes
        Bytes to convert

    Returns
    -------
    : str
        The binary data as a string
    """
    return f"{int.from_bytes(data, byteorder='big'):0{len(data)*8}b}"
```

imap_processing/idex/idex_l1a.py
tech3371: Yes. looks like I missed this one. Good catch!

imap_processing/utils.py
tech3371: Thank you! I did remember seeing Gavin's note but totally forgot when I refactored it.

tech3371: To you list question, good catch. I was in two places when I wrote that.

greglucas (approved):

PR #1010, opened 2024-10-11
Add spice module to documentation

tmplummer:
# Change Summary

## Overview
These are all changes that were needed to add the spice directory to the documentation build and fix some formatting issues.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -117,6 +117,8 @@ nitpick_ignore_regex = [
     (r"py:.*", r".*np.ndarray.*"),
     (r"py:.*", r".*numpy._typing._array_like._ScalarType_co.*"),
     (r"py:.*", r".*idex.l1a.TRIGGER_DESCRIPTION.*"),
+    (r"py:.*", r".*.spice.geometry.SpiceBody.*"),
+    (r"py:.*", r".*.spice.geometry.SpiceFrame.*"),
 ]
 
 # Ignore the inherited members from the <instrument>APID IntEnum class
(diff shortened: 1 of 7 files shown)


docs/source/conf.py
```diff
@@ -117,6 +117,9 @@
     (r"py:.*", r".*np.ndarray.*"),
     (r"py:.*", r".*numpy._typing._array_like._ScalarType_co.*"),
     (r"py:.*", r".*idex.l1a.TRIGGER_DESCRIPTION.*"),
+    (r"py:.*", r".*.spice.geometry.SpiceBody.*"),
```
tmplummer, line 120: More nitpick ignores for `IntEnum`s and some `numpy_typing`. What confuses me is that theses sphinx warnings would go away if I built the documentation a second time without cleaning. I simply couldn't figure out how to fix these issues without just adding the ignore regex here.

imap_processing/spice/geometry.py
```diff
@@ -96,10 +97,11 @@ def imap_state(
     ----------
     et : np.ndarray or float
         Epoch time(s) [J2000 seconds] to get the IMAP state for.
-    ref_frame : SpiceFrame, optional
```
tmplummer, line 99: I have no clue why sphinx was having problems with this `optional` here, which adheres to `numpydoc` syntax. I spent way too long trying to fix it but ended up just having to remove it.

laspsandoval (commented):
I actually don't see anything that needs to be changed. Looks good to me, but I'll let other reviewers chime in before I hit "approve".

maxinelasp (approved):
LGTM. Sphinx is... its own challenge... so nice job integrating this

PR #1016, opened 2024-10-11
Lo DE Segmented Packets

### LINKED ISSUE #1011: Combine Lo segmented direct event data
If there are too many bits of data needed to store Lo direct events for a single Aggregated Science Cycle into a single packet, the data will be segmented across multiple packets. The segments will be indicated by the seq_flag:

seq_flag 1 = start of segment
seq_flag 0 = one of the middle packet of a segment
seq_flag 2 = last packet of a segment
seq_flag 3 = unsegmented packet

The packets after the first will only contain the headers, shcoarse, binary data and checksum. They will not contain the count and passes fields.

sdhoyt:
# Change Summary

## Overview
This PR adds a function to handle Lo direct event segmented packets. This function will soon be added to Lo processing in combination with the event parsing to produce a L1A dataset for Lo direct events. 

If there are too many bits of data needed to store Lo direct events for a single Aggregated Science Cycle into a single packet, the data will be segmented across multiple packets. The segments will be indicated by the seq_flag:

seq_flag 1 = start of segment
seq_flag 0 = one of the middle packet of a segment
seq_flag 2 = last packet of a segment
seq_flag 3 = unsegmented packet

The packets after the first will only contain the headers, shcoarse, binary data and checksum. They will not contain the count and passes fields.

Closes #1011


imap_processing/lo/l0/lo_science.py
```diff
@@ -320,3 +321,87 @@ def parse_de_bin(
...
+        Updated dataset with any segmented direct events combined.
+    """
+    seq_flgs = dataset.seq_flgs.values
+    seq_ctrs = dataset.src_seq_ctr.values
+
+    # Find the start and end of each segment of direct events
+    # 1 = start of a group of segmented packet
+    # 2 = end of a group of segmented packets
+    # 3 = unsegmented packet
+    seg_starts = np.where((seq_flgs == 1) | (seq_flgs == 3))[0]
```
tmplummer: I nit-pick the use of `np.where()` a lot. This should really be `np.nonzero()`. See the `Note` at the top of the numpy documentation: https://numpy.org/doc/2.0/reference/generated/numpy.where.html

tmplummer (approved):
This looks good to me. Nice clean, easy to read code. I'll approve with the assumption that you address the comment on the use of `np.where()`.

imap_processing/lo/l0/lo_science.py
sdhoyt: Ah interesting. I wasn't aware of that. Thanks, I'll make that change

imap_processing/lo/l0/lo_science.py
```diff
@@ -320,3 +321,86 @@ def parse_de_bin(
...
+
+    Returns
+    -------
+    dataset : xr.Dataset
+        Updated dataset with any segmented direct events combined.
+    """
+    seq_flgs = dataset.seq_flgs.values
+    seq_ctrs = dataset.src_seq_ctr.values
+
+    # Find the start and end of each segment of direct events
```
tech3371, line 348: These short comments were very helpful!

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -61,6 +62,36 @@ def fake_de_dataset():
     return dataset
 
 
+@pytest.fixture()
+def segmented_pkts_fake_data():
+    dataset = xr.Dataset(
+        data_vars=dict(
+            seq_flgs=(["epoch"], np.array([1, 0, 0, 2, 3, 1, 0, 2, 1, 2])),
```
tech3371, line 69: can you test for bad scenarios? For example, missing start flag and missing end flag.
```suggestion
            seq_flgs=(["epoch"], np.array([1, 0, 0, 2, 2, 1, 0, 3, 1, 0, 2, 1, 2])),
```

tech3371 (commented):
One minor comment. Otherwise, LGTM

imap_processing/tests/lo/test_lo_science.py
sdhoyt: My check for sequential seq counter field should cover if one of the start/end packets are dropped. Or are you thinking of a different scenario that could happen?

tech3371: I may be over complicating it. I was thinking if there would be case when sequence counter continues to increment but your sequence flags would be out of order. That may be overkill. Do they reset sequence counter or does it continue incrementing? Just curious

sdhoyt: The sequence counter keeps incrementing, so I'm dropping any group of packets where there's some gap in the seq counter values. I believe the POC makes sure the seq values are in order when they are sent to us. So it sounds like you're concerned not that a packet is dropped, but that a packet gets tagged with a wrong seq field? Do you know how we might handle that situation? I'm not sure I'd have enough information to figure out which packet or group of packets caused the issue. I guess we wouldn't be able to create that CDF and would have to work with the instrument team to figure out where things may have gone wrong onboard the spacecraft?

tech3371: As I thought, I am overcomplicating. Ignore my comment :)

tech3371 (approved):

imap_processing/tests/lo/test_lo_science.py
sdhoyt: That could be something I should check for. I'll talk to Colin about it and see what he thinks. If that's something we could encounter I'll add that in a follow-up PR. Thanks!

sdhoyt:
I just made an update to this PR. I found out that padding is added to the binary data to make the packet divisible by 8, so I made a small change to help me identify where to look for padding when this is used in processing. The change adds a comma between segments when they're combined into a single binary string. I'm not crazy about that approach, but I should be able to use that with a find for the next comma or end of the string and determine which bits I need to throw way. I also thought about keeping the binary string separated in a list of lists, but that would require another for loop or passing around another variable to keep track of which segment I'm in. Adding the comma seems easier, but the binary string with commas isn't ideal. Hoping to get others' thoughts on this.

tech3371:
> I just made an update to this PR. I found out that padding is added to the binary data to make the packet divisible by 8, so I made a small change to help me identify where to look for padding when this is used in processing. The change adds a comma between segments when they're combined into a single binary string. I'm not crazy about that approach, but I should be able to use that with a find for the next comma or end of the string and determine which bits I need to throw way. I also thought about keeping the binary string separated in a list of lists, but that would require another for loop or passing around another variable to keep track of which segment I'm in. Adding the comma seems easier, but the binary string with commas isn't ideal. Hoping to get others' thoughts on this.

I wish I have ideas for this. This is where Greg is great at :).

tech3371 (approved):
Still looks good to me. You can revisit this again in the future if you think of another solution to `,`.

PR #1054, opened 2024-10-28
1045 hi l1b modify processing workflow to use xarraydatasetupdate

### LINKED ISSUE #1045: Hi L1B - modify processing workflow to use xarray.Dataset.update
### Algorithm Description:
The current implementation creates new variables and adds them to the dataset using `xarray.Dataset.assign()` which makes a copy of the dataset. An inplace modification would be to use the `update()` method. This will require the functions that create and populate the new variables to return dictionaries with the new `DataArray`s rather than returning a modified `Dataset`.

tmplummer:
# Change Summary

## Overview
These are all just changes to address a comment that @laspsandoval made in a recent Hi PR. I was using `xarray.Dataset.assign()` to add new variables to a dataset and Laura pointed out that I should use `update()` to do an in-place modification of the dataset rather than using `assign()` which does a `copy()` behind the scenes. I don't like to mutate inputs to functions, so I now output a dictionary containing the `xarray.Dataarray` objects to be added to the dataset and then call `.update()` in the outer function to add those to the output `xarray.Dataset`.

Closes: #1045

### DIFF
--- a/imap_processing/hi/l1b/hi_l1b.py
+++ b/imap_processing/hi/l1b/hi_l1b.py
@@ -155,7 +155,7 @@ def compute_coincidence_type_and_time_deltas(
     Returns
     -------
     new_vars : dict[str, xarray.DataArray]
-        Keys are variable names and values are new populated `xarray.DataArray`s.
+        Keys are variable names and values are new populated `xarray.DataArray`.
     """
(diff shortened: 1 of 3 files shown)


laspsandoval (approved):
I actually think this all looks good and have no suggestions. Nice work!

PR #1059, opened 2024-10-29
Ultra l1c pset updates

### LINKED ISSUE #904: SPICE - cartesian to spherical function
### Algorithm Description:

Laura has implemented this function in imap_processing/ultra/l1c/ultra_l1c_pset_bins.py. We should perhaps move that function into the geometry module?

### LINKED ISSUE #1060: Ultra l1c pset updates
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Responding to great PR comments by Tim that I missed the first time.

## Updated Files
- geometry.py
   - moved cartesian to spherical and vice versa functions to geometry.py and made recommended updates
- ultra_l1c_pset_bins.py
   - reformatting based on changes to cartesian to spherical functions
   - added clarifying docstrings

## Testing
- test_geometry.py
   - Tests for updates to geometry.py
- test_ultra_l1c_pset_bins.py
   - Uses metakernel now
- imap_ena_sim_metakernel.template
   - Update metakernel to include DPS frame and kernel

Closes: #904


demajistre:
These are comments to the last pull request that has already been merged, but these directed at the same files (most notably `ultra_l1c_pset_bins.py`). 

image

`build_energy_bins()`:
should `alpha, energy_start, n_bins` be set in default arguments?

`build_spatial_bins(spacing: float = 0.5,)`:
I'd recommend that there be separate spacing for az and el, which you can now default to 0.5.
I like how the representation of the bins matches `np.histogramdd()`

```
def get_histogram(
    v: tuple[np.ndarray, np.ndarray, np.ndarray],
    energy: np.ndarray,
    az_bin_edges: np.ndarray,
    el_bin_edges: np.ndarray,
    energy_bin_edges: np.ndarray,
):
```
Having both `v` and `energy` is potentially confusing, particularly because the comments do not say anything about the units of these quantities.   These are not independent quantities, i.e., the energy of a particle is (m*v*v)/2 where m is the particle mass and v*v is the vector dot product. I don't think you want to implement getting the energy from the velocity vector because the unit conversions are unnecessarily complicated.  HOWEVER, this routine will operate correctly if you define v here to be the **unit velocity vector**.  You don't have to change the code (except to maybe rename the variable `v` to `uv` or `vhat`)  and whoever is calling this routine will know that the magnitude of the velocity is going to be thrown away. 

`get_helio_exposure_times(time: np.ndarray,sc_exposure: np.ndarray,)`:
Reading through the code, this looks OK,  but there is a lot of opportunity for dyslexia. I assume that this is correct since your test routine looks like it matches Nick's standard tables. I should note that on line 207 that this is not a compton-getting transform - it is a Galilean Transform.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -175,7 +106,12 @@ def get_histogram(
     hist : np.ndarray
         A 3D histogram array.
     """
-    az, el, _ = cartesian_to_spherical(v)
+    spherical_coords = cartesian_to_spherical(v)
+    az, el, _ = (
+        spherical_coords[..., 0],
+        spherical_coords[..., 1],
+        spherical_coords[..., 2],
+    )
```
tmplummer: No need to include the radius in the right hand side:
```suggestion
    az, el = (
        spherical_coords[..., 0],
        spherical_coords[..., 1],
    )
```

imap_processing/tests/spice/test_geometry.py
```diff
@@ -306,3 +308,49 @@ def test_instrument_pointing(furnish_kernels):
             et, SpiceFrame.IMAP_HI_90, SpiceFrame.ECLIPJ2000, cartesian=True
         )
         assert ins_pointing.shape == (3, 3)
+
+
+def test_cartesian_to_spherical():
```
tmplummer, line 342: I would love to see this test bulked up by using a meshgrid of `x, y, z` values in the range `[-1, 1]` with steps of something like 0.05 so that all directions are covered. Results can be checked against the spiceypy function `recsph`. Note spice defines phi in the range `[-pi, pi]`.

imap_processing/tests/spice/test_geometry.py
```diff
@@ -306,3 +308,49 @@ def test_instrument_pointing(furnish_kernels):
...
+
+    np.testing.assert_allclose(
+        np.unique(np.radians(az_sc)), np.array([1.31300, 2.34891]), atol=1e-05, rtol=0
+    )
+    np.testing.assert_allclose(
+        np.unique(np.radians(el_sc)), np.array([-0.88901, -0.70136]), atol=1e-05, rtol=0
+    )
+
+
+def test_spherical_to_cartesian():
```
tmplummer, line 369: I would love to see this test bulked up by using a meshgrid of `theta, phi` values in the range `[0, pi]` and `[-pi, pi]` with steps of something like `pi/50` so that all directions are covered. Results can be checked against the spiceypy function `sphrec`.

imap_processing/spice/geometry.py
```diff
@@ -443,3 +444,74 @@ def instrument_pointing(
...
+        with x, y, z-components.
+
+    Returns
+    -------
+    spherical_coords : np.ndarray
+        Spherical coordinate.
+    """
+    vx = v[:, 0]
+    vy = v[:, 1]
+    vz = v[:, 2]
```
tmplummer: Is there a reason not to use the same indexing as you did in the `spherical_to_cartesian` function like:
```suggestion
    vx = v[..., 0]
    vy = v[..., 1]
    vz = v[..., 2]
```

imap_processing/spice/geometry.py
```diff
@@ -443,3 +444,74 @@ def instrument_pointing(
...
+
+    vhat_x = -vx / magnitude_v
+    vhat_y = -vy / magnitude_v
+    vhat_z = -vz / magnitude_v
+
+    # Elevation angle (angle from the z-axis, range: [-pi/2, pi/2])
+    el = np.arcsin(vhat_z)
+
+    # Azimuth angle (angle in the xy-plane, range: [0, 2*pi])
+    az = np.arctan2(vhat_y, vhat_x)
```
tmplummer: I'm wondering if some of this can be streamlined by:
```suggestion
    magnitude_v = np.sqrt(np.sum(v**2, axis=-1))
    vhat = -v / magnitude_v
    # Elevation angle (angle from the z-axis, range: [-pi/2, pi/2])
    el = np.arcsin(vhat[..., 2])
    # Azimuth angle (angle in the xy-plane, range: [0, 2*pi])
    az = np.arctan2(vhat[..., 1], vhat[..., 0])
```

imap_processing/spice/geometry.py
```diff
@@ -443,3 +444,74 @@ def instrument_pointing(
...
+
+    # Ensure azimuth is from 0 to 2PI
+    az = az % (2 * np.pi)
+    spherical_coords = np.stack((np.degrees(az), np.degrees(el), magnitude_v), axis=-1)
```
tmplummer: Spherical coordinates are standardized as being ordered like `(r, theta, phi)`.
```suggestion
    spherical_coords = np.stack(magnitude_v, (np.degrees(az), np.degrees(el)), axis=-1)
```

imap_processing/spice/geometry.py
```diff
@@ -443,3 +444,74 @@ def instrument_pointing(
...
+    ----------
+    v : np.ndarray
+        A NumPy array with shape (n, 3) where each
+        row represents a vector
+        with x, y, z-components.
+
+    Returns
+    -------
+    spherical_coords : np.ndarray
+        Spherical coordinate.
```
tmplummer: Can you define the order of coordinates output and their ranges here? Could lift some of the documentation from spice: https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/recsph_c.html

tmplummer (commented):
This is great! Thanks for doing this. I have some comments regarding improving test coverage and some suggestions to streamline code a bit.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -65,124 +73,71 @@ def build_spatial_bins(
...
-    # 3D binning.
-    hist, _ = np.histogramdd(
-        sample=(az, el, energy), bins=[az_bin_edges, el_bin_edges, energy_bin_edges]
+    # Initialize histogram
+    hist_total = np.zeros(
+        (len(az_bin_edges) - 1, len(el_bin_edges) - 1, len(energy_bin_edges))
     )
 
-    return hist
+    for i, (e_min, e_max) in enumerate(energy_bin_edges):
```
tmplummer, line 130: Are the energy bins such that they can't be passed directly to `np.histogramdd()` if they were made to be an array of edges like `az_bin_edges`? Like:
```
hist_total = histogramdd(
    (az.ravel(), el.ravel(), energy.ravel()),
    bins=[az_bin_edges, el_bin_edges, energy_bin_edges]
)
```
If there are no gaps between the energy bands, it seems like `energy_bin_edges` could be made to be an array of edges rather than a list of tuples.

tmplummer (approved):
This is great! Thanks for making those changes and adding those useful geometry functions. Just had the one further suggestion to consider.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
laspsandoval: That is how I had it before, but the Ultra team told me that there are **overlapping** energy bins so I had to make some changes.

PR #1079, opened 2024-10-29
Update Lo XTCE for Segmented DE packets

### LINKED ISSUE #1012: Update Lo XTCE Direct Events to group count and passing fields into the binary chunk
Because of the DE segmented packets, the Lo XTCE for direct events needs to be updated to include the count and passes fields into the binary chunk

Related issue: #1011

sdhoyt:
# Change Summary

## Overview
Updated the Lo XTCE file for the segmented DE packets. To handle segmented DE packets, the `counts` and `passes` fields were pulled into the `data` variable length binary chunk

There's also an issue with the EU conversion data in the telemetry sheet not working with space packet parser, so Housekeeping was removed for now.

Closes #1012


tech3371:
> There's also an issue with the EU conversion data in the telemetry sheet not working with space packet parser, so Housekeeping was removed for now.

Instead of removing your HK packet definition, we may want to fix the issue. I can take that on.

sdhoyt:
@tech3371 You're right I should resolve the issue here. I just started looking at this more now.

The Lo telemetry housekeeping engineering states have some in hexideicmal format: (0x0, 0x1, etc). When space packet parser reads the XTCE, it tries to convert those to integers in base 10 and it errors because those need to be converted in base 16. I'm not sure if manually changing these in the telemetry definition will work and I'd like to avoid doing that. I'm trying to figure out if this is something that needs to be handled in space packet parser or if there's something I can do on my end.

sdhoyt:
I made a couple updates to this PR. I pulled the DE checksum field into the de binary chunk and I removed the housekeeping xtce again. The Lo sample data file contains housekeeping packets that have values not matching the enumerations in the telemetry sheet, so this is causing errors in space packet parser when the ccsds file is unpacked. I will follow-up with the housekeeping issues in another PR.

tech3371 (approved):
I understand. LGTM

PR #1081, opened 2024-10-30
Ultra: Updates to annotated events and create l1b cdf.

laspsandoval:
# Change Summary

## Overview
Updates to annotated events and create l1b cdf.

## Updated Files
- de.py
   - Added way to create dataset
- ultra_l1b_annotated.py
   - Added velocity in the spacecraft frame
- ultra_l1b_extended.py
   - Minor edits necessary to make de.py
- constants.py
   -  Minor edits necessary to make de.py
   - Updated variable attributes to use defaults

## Testing
- test_ultra_l1b.py
   - Tests ultra_l1b.py
- test_ultra_l1b_annotated.py
   - Tests ultra_l1b_annotated.py
- test_ultra_l1b_extended.py
   - Tests ultra_l1b_extended.ply

### DIFF
--- a/imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
@@ -213,21 +213,21 @@ vz_sc:
 
 vx_dps_sc:
   <<: *default_float32
-  CATDESC: x-component of the velocity vector of the ena in the DPS frame at rest WRT spacecraft (velocity w/ spacecraft frame subtracted)
+  CATDESC: x-component of the velocity vector of the ena in the DPS frame at rest WRT spacecraft.
   FIELDNAM: vx_dps_sc
   LABLAXIS: vx dps sc
   UNITS: " "
 
 vy_dps_sc:
   <<: *default_float32
-  CATDESC: y-component of the velocity vector of the ena in the DPS frame at rest WRT spacecraft (velocity w/ spacecraft frame subtracted)
+  CATDESC: y-component of the velocity vector of the ena in the DPS frame at rest WRT spacecraft.
   FIELDNAM: vy_dps_sc
   LABLAXIS: vy dps sc
   UNITS: " "
 
 vz_dps_sc:
   <<: *default_float32
-  CATDESC: z-component of the velocity vector of the ena in the DPS frame at rest WRT spacecraft (velocity w/ spacecraft frame subtracted)
+  CATDESC: z-component of the velocity vector of the ena in the DPS frame at rest WRT spacecraft.
   FIELDNAM: vz_dps_sc
   LABLAXIS: vz dps sc
   UNITS: " "
--- a/imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
@@ -8,84 +8,122 @@ default_attrs: &default
   VALIDMAX: 9223372036854775807
   VAR_TYPE: data
 
+epoch:
+  <<: *default
+  CATDESC: Time, number of nanoseconds since J2000 with leap seconds included
+  FIELDNAM: epoch
+  LABLAXIS: epoch
+  UNITS: ns
+  VAR_TYPE: support_data
+  SCALETYP: linear
+  MONOTON: INCREASE
+  TIME_BASE: J2000
+  TIME_SCALE: Terrestrial Time
+  REFERENCE_POSITION: Rotating Earth Geoid
+
 x_front:
   <<: *default
   CATDESC: x front position
   FIELDNAM: Event x-position on front foil
   LABLAXIS: x front position
-  # TODO: come back to format
-  UNITS: mm
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: mm / 100
 
 y_front:
   <<: *default
   CATDESC: y front position
   FIELDNAM: Event y-position on front foil
   LABLAXIS: y front position
-  # TODO: come back to format
-  UNITS: mm
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: mm / 100
 
 x_back:
   <<: *default
   CATDESC: x back position
   FIELDNAM: x_back
   LABLAXIS: x back position
-  # TODO: come back to format
-  UNITS: mm
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: mm / 100
 
 y_back:
   <<: *default
   CATDESC: y back position
   FIELDNAM: y_back
   LABLAXIS: y back position
-  # TODO: come back to format
-  UNITS: mm
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: mm / 100
 
 x_coin:
   <<: *default
   CATDESC: x coincidence position
   FIELDNAM: x_coin
   LABLAXIS: x coincidence position
-  # TODO: come back to format
-  UNITS: mm
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: mm / 100
 
 tof_start_stop:
   <<: *default
   CATDESC: Particle time of flight from start to stop
   FIELDNAM: tof_start_stop
   LABLAXIS: tof start stop
-  # TODO: come back to format
-  UNITS: ns
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: nanosecond / 10
 
 tof_stop_coin:
   <<: *default
   CATDESC: Particle time of flight from stop to coincidence
   FIELDNAM: tof_stop_coin
   LABLAXIS: tof stop coin
-  # TODO: come back to format
-  UNITS: ns
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: nanosecond / 10
 
 tof_corrected:
   <<: *default
   CATDESC: Corrected time of flight
   FIELDNAM: tof_corrected
   LABLAXIS: tof corrected
-  # TODO: come back to format
-  UNITS: ns
+  VALIDMIN: -1e6
+  FORMAT: F12.6
+  UNITS: nanosecond / 10
 
-eventtype:
+event_type:
   <<: *default
+  VALIDMAX: 15
   CATDESC: Eventtype (1-2 top and bottom stop types; 8-15 ssd stop types)
-  FIELDNAM: eventtype
-  LABLAXIS: eventtype
-  # TODO: come back to format
-  UNITS: " "
+  FIELDNAM: Event type
+  FORMAT: I2
+  LABLAXIS: event type
+
+coincidence_type:
+  <<: *default
+  VALIDMAX: 2
+  CATDESC: Coincidence type for the direct event (1-2 top and bottom)
+  FIELDNAM: Coincidence type
+  FORMAT: I1
+  LABLAXIS: coincidence type
+
+start_type:
+  <<: *default
+  VALIDMAX: 2
+  CATDESC: Start type for the direct event (1-2 left and right)
+  FIELDNAM: Start type
+  FORMAT: I1
+  LABLAXIS: start type
 
 vx_ultra:
   <<: *default
   CATDESC: Normalized component of the velocity vector in x direction in the instrument frame.
   FIELDNAM: vx_ultra
   LABLAXIS: vx ultra
-  # TODO: come back to format
+  VALIDMIN: -1e6
+  FORMAT: F12.6
   UNITS: " "
 
 vy_ultra:
@@ -93,7 +131,8 @@ vy_ultra:
   CATDESC: Normalized component of the velocity vector in y direction in the instrument frame.
   FIELDNAM: vy_ultra
   LABLAXIS: vy ultra
-  # TODO: come back to format
+  VALIDMIN: -1e6
+  FORMAT: F12.6
   UNITS: " "
 
 vz_ultra:
@@ -101,7 +140,8 @@ vz_ultra:
   CATDESC: Normalized component of the velocity vector in z direction in the instrument frame.
   FIELDNAM: vz_ultra
   LABLAXIS: vz ultra
-  # TODO: come back to format
+  VALIDMIN: -1e6
+  FORMAT: F12.6
   UNITS: " "
 
 energy:
@@ -109,16 +149,33 @@ energy:
   CATDESC: Energy measured using the pulse height from the stop anode (DN) or Energy measured using the SSD (keV) depending on eventtype.
   FIELDNAM: energy
   LABLAXIS: energy
-  # TODO: come back to format
+  VALIDMIN: -1e6
+  FORMAT: F12.6
   UNITS: keV
 
 species:
   <<: *default
-  CATDESC: Species bin.
+  CATDESC: Species bin (0-255).
   FIELDNAM: species
   LABLAXIS: species
-  # TODO: come back to format
+  FORMAT: I2
   UNITS: " "
+
+front_back_distance:
+  <<: *default
+  CATDESC: Distance front to back.
+  FIELDNAM: front_back_distance
+  LABLAXIS: front_back_distance
+  FORMAT: F12.6
+  UNITS: mm / 100
+
+path_length:
+  <<: *default
+  CATDESC: Path length.
+  FIELDNAM: path_length
+  LABLAXIS: path_length
+  FORMAT: F12.6
+  UNITS: mm / 100
 
 event_efficiency:
   <<: *default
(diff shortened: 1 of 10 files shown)


imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -3,210 +3,293 @@ default_attrs: &default
...
+
+x_front:
+  <<: *default_float32
   CATDESC: x front position
   FIELDNAM: Event x-position on front foil
   LABLAXIS: x front position
-  # TODO: come back to format
-  UNITS: mm
+  UNITS: mm / 100
+  dtype: float32
```
tmplummer: This `dtype` is duplicated, no need to have it here. It is defined in `default_float32`. This goes for all of the variables below as well.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -3,210 +3,293 @@ default_attrs: &default
...
   LABLAXIS: vz sc
-  # TODO: come back to format
   UNITS: " "
+  dtype: float32
 
 vx_dps_sc:
-  <<: *default
-  CATDESC: Normalized component of the velocity vector in x direction in the DPS frame at rest WRT spacecraft (velocity w/ spacecraft frame subtracted)
+  <<: *default_float32
+  CATDESC: x-component of the velocity vector of the ena in the DPS frame at rest WRT spacecraft (velocity w/ spacecraft frame subtracted)
```
tmplummer: I'm not sure I understand this description. I understand up to the "velocity vector of the ena in the DPS frame". I don't know what "at rest WRT spacecraft" or "(velocity w/ spacecraft frame subtracted)" are meant to convey.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -3,210 +3,293 @@ default_attrs: &default
...
   LABLAXIS: vz dps sc
-  # TODO: come back to format
   UNITS: " "
+  dtype: float32
 
 vx_dps_helio:
-  <<: *default
-  CATDESC: Normalized component of the velocity vector in x direction in the DPS frame at rest WRT heliosphere.
+  <<: *default_float32
+  CATDESC: x-component of the velocity vector of the ena in the DPS frame at rest WRT heliosphere.
```
tmplummer, line 237: Could this be something like "Compton-getting corrected velocity vector"? If this is how the IT wants it described, that is fine, but I have a hard time understanding what "velocity vector of the ena in the DPS frame at rest WRT heliosphere" means.

imap_processing/ultra/l1b/de.py
```diff
@@ -23,36 +39,161 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
         Dataset containing the data.
     """
     de_dict = {}
+    sensor = name.split("_l1b_")[1].split("sensor-de")[0]
```
tmplummer: I wrote a function to parse things like this out of filename like strings such as `Logical_source`. https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/cdf/utils.py#L150

imap_processing/ultra/l1b/de.py
```diff
@@ -23,36 +39,161 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    ssd_r = get_path_length((xf[ssd_indices], ssd_yf), (ssd_xb, ssd_yb), ssd_d)
+    ssd_bin = determine_species_ssd(
+        ssd_energy,
+        ssd_tof,
+        ssd_r,
+    )
+    ssd_ctof = get_ctof(ssd_tof, ssd_r, "SSD")
+
+    # Combine ph_yb and ssd_yb along with their indices
+    combined_indices = np.argsort(np.concatenate((ph_indices, ssd_indices)))
```
tmplummer: I find this very confusing. Are you trying to accomplish something different than:
```
de_dict["y_back"] = np.full(len(epoch), np.nan)
de_dict["y_back"][ph_indices] = ph_yb
de_dict["y_back"][ssd_indices] = ph_ssd
```
and so forth?

Is there any guarantee that `ph_indices` is the compliment to `ssd_indices`? Otherwise, it seems like there could be duplicate or missing indices.

imap_processing/ultra/l1b/de.py
```diff
@@ -23,36 +39,161 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    )
+
+    vx_ultra, vy_ultra, vz_ultra = get_particle_velocity(
+        (de_dict["x_front"], de_dict["y_front"]),
+        (de_dict["x_back"], de_dict["y_back"]),
+        de_dict["front_back_distance"],
+        de_dict["tof_start_stop"],
+    )
+
+    # We need to fill velocities that have negative tof values.
```
tmplummer: This filling of velocities where tof is less than or equal to zero could be done in-place:
```
fill_velocity_mask = de_dict["tof_start_stop"] <= 0
vx_ultra[fill_velocity_mask] = np.finfo(np.float64).min
de_dict["vx_ultra"] = vx_ultra
...
```

imap_processing/ultra/l1b/de.py
```diff
@@ -23,36 +39,161 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    de_dict["vx_ultra"] = np.full_like(
+        vx_ultra, np.finfo(np.float64).min, dtype=np.float64
```
tmplummer: Is `np.finfo(np.float64).min` the same as `np.nan`? Might be worth checking. `np.nan` is much preferred IMO.

imap_processing/ultra/l1b/de.py
```diff
@@ -23,36 +39,161 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    # ultra_frame = getattr(SpiceFrame, f"IMAP_ULTRA_{sensor}")
+    # sc_velocity, sc_dps_velocity, helio_velocity = get_annotated_particle_velocity(
+    #     de_dataset.data_vars["EVENTTIMES"],
+    #     position,
+    #     ultra_frame,
+    #     SpiceFrame.IMAP_DPS,
+    #     SpiceFrame.IMAP_SPACECRAFT,
+    # )
+    # TODO: this is a temporary fix.
+    sc_velocity = np.zeros((len(de_dict["epoch"]), 3))
```
tmplummer: Fill with NaNs for now?

imap_processing/ultra/l1b/ultra_l1b_annotated.py
```diff
@@ -61,6 +74,12 @@ def get_particle_velocity(
 
     # Apply Compton-Getting.
     # Particle velocity in the DPS frame wrt to the heliosphere
-    particle_velocity_heliosphere = spacecraft_velocity + particle_velocity_spacecraft
+    particle_velocity_heliosphere = (
```
tmplummer, line 72: ~~I'm concerned about this calculation. I think that `spacecraft_velocity` should always be 0 in the `IMAP_DPS` frame because that frame moves with the spacecraft. I think that you need to get the spacecraft velocity in the heliosphere frame (`ECLIPJ2000`) and add that to the particle velocity expressed in the heliosphere frame. Something like:~~
```
particle_vel_helio = frame_transform(
    et=time,
    position=instrument_velocity,
    from_frame=instrument_frame,
    to_frame=SpiceFrame.ECLIP2000
)
# Spacecraft velocity in the pointing (DPS) frame wrt heliosphere.
state = imap_state(time, ref_frame=SpiceFrame.ECLIPJ2000)
# Extract the velocity part of the state vector
spacecraft_velocity = state[:, 3:6]
particle_vel_compton_corrected = spacecraft_velocity + particle_velocity_dps_spacecraft
```

tmplummer (commented):
I gave some feedback. This is a big PR. I'll try to come back to it next week.

imap_processing/ultra/l1b/de.py
tmplummer: Even better, why is filling with `np.finfo(np.float64).min` not done inside the `get_particle_velocity` function?

imap_processing/ultra/l1b/ultra_l1b_annotated.py
tmplummer: Nevermind, `imap_state` defaults to using the sun as the observing body.

imap_processing/ultra/l1b/de.py
```diff
@@ -23,36 +39,161 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    de_dict["front_back_distance"] = d[combined_indices]
+    r = np.concatenate((ph_r, ssd_r))
+    de_dict["path_length"] = r[combined_indices]
+    tof = np.concatenate((ph_tof, ssd_tof))
+    de_dict["tof_start_stop"] = tof[combined_indices]
+    etof = np.concatenate((ph_etof, np.zeros(len(ssd_indices))))
+    de_dict["tof_stop_coin"] = etof[combined_indices]
+
+    ctof = np.concatenate((ph_ctof, ssd_ctof))
+    de_dict["tof_corrected"] = ctof[combined_indices]
```
tmplummer: The main challenge that I am having in understanding all of this code relates to the subsetting of the dataset into PH and SSD. I think that it would be easier to follow if you did one of two things:

1. Instantiate the full array of output values for a variable ahead of time and then set the ph/ssd indices directly:
  ```
  # Instantiate full size arrays
  xf = np.full(len(epoch), np.nan, dtype=np.float64)
  ...
  # Set ph indexed values
  tof[ph_indices], t2[ph_indices], xb[ph_indices], yb[ph_indices] = get_ph_tof_and_back_positions(
        de_dataset, xf, f"ultra{sensor}"
    )
  # Set ssd indexed values
  tof[ssd_indices], t2[ssd_indices], xb[ssd_indices], yb[ssd_indices] = get_ssd_tof_and_back_positions(
        de_dataset, xf, f"ultra{sensor}"
    )
```
2. Have functions that just output the combined results already. i.e. within `get_tof_and_back_positions`, have the logic to do what suggestion 1 is doing.

**Comment by demajistre** [comment]
Quick non-urgent comment:the comments in calculate_de should be more clear.  let the user know that _de_dataset_ is the l1a dataset and _name_ is the name of the de dataset returned.

File: imap_processing/ultra/l1b/de.py [hunk 11]
```diff
@@ -23,36 +39,161 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    de_dict.update(
+        {key: de_dataset[dataset_key] for key, dataset_key in zip(keys, dataset_keys)}
+    )
+
+    vx_ultra, vy_ultra, vz_ultra = get_particle_velocity(
+        (de_dict["x_front"], de_dict["y_front"]),
+        (de_dict["x_back"], de_dict["y_back"]),
+        de_dict["front_back_distance"],
+        de_dict["tof_start_stop"],
+    )
```
demajistre, line 136: Note that the call to get_particle_velocity returns unit vectors, not full velocity vectors. These can't be used directly for the velocity transforms (when building annotated events) without including the magnitude (likely using the energy value from the extended set - checking that out now).

demajistre:
After review of the code and discussions with the Ultra team, I think a separate issue and PR needs to be open for the code that produces the L1B extended product.  I've got only minor comments on the code as it stands (as noted elsewhere in the review) and it seems pretty faithful to the FSW document algorithms.  The FSW algorithms need to be supplemented.

imap_processing/ultra/l1b/de.py
```diff
@@ -23,36 +39,161 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+        "event_type",
+        "de_event_met",
+    ]
+    dataset_keys = ["COIN_TYPE", "START_TYPE", "STOP_TYPE", "SHCOARSE"]
+
+    de_dict.update(
+        {key: de_dataset[dataset_key] for key, dataset_key in zip(keys, dataset_keys)}
+    )
+
+    vx_ultra, vy_ultra, vz_ultra = get_particle_velocity(
```
demajistre: Upon inspection the `get_particle_velocity` does not produce a velocity vector. In the flight software document it is labeled `v`, but I believe this is v for vector rather than v for velocity. The quantity in the FSW has units of length (not length/time).  `get_particle_velocity` does normalize the vector, and in doing so builds a **unit** velocity vector, but this is not the correct vector to be used for the annotated events downstream.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
laspsandoval: I can ask them again, but we've been through it a few times now and landed on this description.

imap_processing/ultra/l1b/de.py
demajistre: Low priority comment - the name of the method is now `get_unit_vector `, but I believe it should be `get_unit_velocity_vector `- since there are a lot of possible unit vectors that could be calculated.

imap_processing/ultra/l1b/de.py
```diff
@@ -13,46 +30,172 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    xb = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
+    yb = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
+    xc = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
+    d = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float64)
+    r = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
+    tof = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
+    etof = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
+    ctof = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
+    energy = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
+    species_bin = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.uint8)
```
tmplummer, line 62: I'm curious what value this fills with. `np.nan` should only be valid for floating point types AFAIK.

imap_processing/ultra/l1b/de.py
```diff
@@ -13,46 +30,172 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
+    # )
+    #
+    # ultra_frame = getattr(SpiceFrame, f"IMAP_ULTRA_{sensor}")
+    # sc_velocity, sc_dps_velocity, helio_velocity = get_annotated_particle_velocity(
+    #     de_dataset.data_vars["EVENTTIMES"],
+    #     position,
+    #     ultra_frame,
+    #     SpiceFrame.IMAP_DPS,
+    #     SpiceFrame.IMAP_SPACECRAFT,
+    # )
```
tmplummer: If I understand this, the problem is that you don't have test data from the same time as the coverage provided by the simulated pointing frame that we have?
I would recommend fixing that issue in the test rather than in this code. You could mock out the `get_annotated_particle_velocity` function in the test and have it return the full NaN ndarrays that you have coded in below.

imap_processing/ultra/l1b/ultra_l1b_annotated.py
```diff
@@ -14,12 +14,13 @@
 
 @ensure_spice
```
tmplummer: It looks like this function doesn't actually need the `ensure_spice` decorator since it now uses `frame_transform` and `imap_state` which are appropriately decorated.

tmplummer (commented):
Thanks for refactoring `calculate_de` it is much easier for me to follow now. I just have a couple more recommendations.

imap_processing/tests/ultra/unit/test_de.py
```diff
@@ -0,0 +1,119 @@
...
+    assert np.allclose(
+        dataset["vy_dps_helio"].data,
+        np.full(len(de_dataset["epoch"]), np.nan),
+        equal_nan=True,
+    )
+    assert np.allclose(
+        dataset["vz_dps_helio"].data,
+        np.full(len(de_dataset["epoch"]), np.nan),
+        equal_nan=True,
+    )
```
nkerman: Why are we testing these against all NaN arrays? Do we want them to populate with NaNs, or is it that we don't have valid test data, but still want to ensure the arrays are created with the proper shape?

If the latter, could we either use fake data which would result in round numbers for testing or just test the shape of these arrays?

imap_processing/ultra/l1b/de.py
nkerman: I was curious too - looks like np fills with 0s:
```
np.full((3,2), np.nan, dtype=np.uint8)
>>> array([[0, 0],
       [0, 0],
       [0, 0]], dtype=uint8)
```

laspsandoval: Good catch. I'll put a TODO for asking the Ultra team if zeros are acceptable. As of now they are still trying to figure out what they want in the species bin. They have said they might even want string values so this is all preliminary.

imap_processing/ultra/l1b/de.py
laspsandoval: I had never done that before. That came in handy. Thanks!

imap_processing/tests/ultra/unit/test_de.py
laspsandoval: These are all calculated here sc_velocity, sc_dps_velocity, helio_velocity = get_annotated_particle_velocity.

And since we don't have SPICE kernels for the timerange of the test data we can't test them. I can add a shape test instead.

nkerman: Thanks for clarifying - sounds good!

tmplummer (approved):
Nice. LGTM

PR #1082, opened 2024-10-30
Mag l1a compression fixes

### LINKED ISSUE #1006: Fix MAG Compression bug
### Summary

There was a missing piece from the MAG Compression algorithm - for bit widths of > 16, we need to scale the data down. This is described in the algorithm document

### Requirements

The more specific the better.
  - [ ] Add step to scale data
  - [ ] Update tests if needed

### Related tickets

_No response_

maxinelasp:
# Change Summary
This is part one of the fix to #1006 - updating MAG L1A to include the compression flag information and width for the L1B processing step which corrects compressed data.

## Overview

- Some minor updates to CDF files for L1A
- Added compression_width to the packet properties
- Added a step within the L1A data processing to change the compression flag and width to an array which can then be directly written to the CDF
- Updating and creating tests

~I don't think the LABL_PTR_1 attribute is required since it's 2D data - so we can just use LABLAXIS...? Not totally sure about that, would appreciate a look from @tech3371.~
Edit: I figured out why that change is required... as a timeseries it apparently requires it. Not sure if compression flags should just have type no_plot to avoid it, but the vectors need it.

Partly fixes #1006.

### DIFF
--- a/imap_processing/mag/l1a/mag_l1a.py
+++ b/imap_processing/mag/l1a/mag_l1a.py
@@ -159,7 +159,7 @@ def process_packets(
             )
         ).astype("datetime64[D]")
 
-        primary_packet_data = MagL1aPacketProperties(
+        primary_packet_properties = MagL1aPacketProperties(
             mag_l0.SHCOARSE,
             primary_start_time,
             mag_l0.PRI_VECSEC,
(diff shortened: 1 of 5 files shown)


imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -305,12 +318,40 @@ def generate_dataset(
...
+        attrs=attribute_manager.get_variable_attributes("compression_label"),
+    )
+
     output = xr.Dataset(
-        coords={"epoch": epoch_time, "direction": direction},
+        coords={
+            "epoch": epoch_time,
+            "direction": direction,
+            "compression": compression,
+            "direction_label": direction_label,
```
tmplummer, line 347: Just an FYI... I don't think it is a requirement that the label variables be specified as coordinates in the `Dataset`, they can go in the `data_vars`. IMO, they actually belong in the `data_vars` but it is still unclear if where they are in the `xarray.Dataset` actually affects the CDF in any meaningful way.

tmplummer (approved):
This all looks good to me. Good docstrings, easy to read code, and good test coverage. I just had the one FYI comment.

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Could you elaborate on `data_vars`? I'm not sure what that means. I agree they don't belong in the coordinates, but that may be more of a philosophical objection than any real complaint

tech3371: This is something Tim and I was discussing. Yes, it's a philosophical question at the moment. I don't think CDF will care but there are two approaches that we are following and Tim pointed out that we may want to converge on one approach. We didn't have enough people to discuss that this month.

maxinelasp: Sounds good, in that case I will leave it as is and we can move it later if we decide to go with the data_vars approach. It would be nice to have a standard

PR #1117, opened 2024-11-04
Updates to MAG L1B to scale based on compression flags

maxinelasp:
# Change Summary
This completes #1006 and finishes adding MAG compression to L1B. 

## Overview

L1B uses the compression_flags that I added in L1A to scale the output vectors. This adds another method and some new tests.


tmplummer (approved):
LGTM!

tech3371 (approved):
Nice doc string. I understand on high level and it makes sense to me.

PR #1118, opened 2024-11-05
HIT L1B - Create housekeeping dataset

### LINKED ISSUE #1065: HIT L1B - create housekeeping CDF
Use the `packet_file_to_datasets` function with the flag `use_derived_value=True` to create the L1B product in hit_l1a.py when the CCSDS file gets read in

vmartinez-cu:
Refactored hit_l1b.py to use the `packet_file_to_dataset()` function to create the L1B housekeeping dataset since that function provides the engineering units needed in an xarray dataset. This new approach requires a CCSDS packet file as a dependency for L1B processing. 

- Updated cli.py to handle a packet file as a dependency for HIT
- Created a hit_utils.py to provide common functions needed by more than one processing level. Added housekeeping data processing and packet file reading functions to this file since it's needed for both L1A and L1B.
- Updated hit L1B unit tests
- Deleted data classes that are no longer being used for HIT processing

Related to issue #1065 

## New Dependencies
A CCSDS file is needed for L1B housekeeping. Will put in a PR for this in sds_data_manager repo

## New Files
- hit_utils.py
   - Contains functions needed to process more than one data level

## Deleted Files
- housekeeping.py
   - data classes replaced by `packet_file_to_datasets` function
- science_packet.py
   - data classes replaced by `packet_file_to_datasets` function
- hit_base.py
   - data classes replaced by `packet_file_to_datasets` function 

## Updated Files
- hit_l1b.py
   - updated L1B housekeeping data processing
- test_hit_l1b.py
   - updated unit tests to reflect changes in hit_l1b.py
   - updated packet file path from "hk" to "raw"
- test_hit_l1a.py
   - updated packet file path from "hk" to "raw"

## Testing
Updated L1B unit tests to reflect changes in hit_l1b.py


imap_processing/cli.py
```diff
@@ -551,19 +551,21 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
...
-                    f"Unexpected dependencies found for HIT L1B:"
-                    f"{dependencies}. Expected only one dependency."
-                )
-            # process data and write all processed data to CDF files
-            l1a_dataset = load_cdf(dependencies[0])
-
-            datasets = hit_l1b(l1a_dataset, self.version)
+            data_dict = {}
+            for dependency in dependencies:
+                if "imap_hit_l0_raw" in str(dependency):
```
maxinelasp: Rather than checking the path, you can go to `self.dependencies` and get the level directly. I don't know if that would be cleaner than this, but it is an option. (see download_dependencies() method in `ProcessInstrument`). I think what you have is ok, since we aren't expecting the filenames to change, but just wanted to call it out as a more reliable method of accessing data level for dependencies.

imap_processing/hit/hit_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    # Create the attribute manager for this data level
+    attr_mgr = ImapCdfAttributes()
+    attr_mgr.add_instrument_global_attrs(instrument="hit")
+    attr_mgr.add_instrument_variable_attrs(instrument="hit", level=level)
+    attr_mgr.add_global_attribute("Data_version", data_version)
+    return attr_mgr
+
+
+def concatenate_leak_variables(
```
maxinelasp, line 93: Is this used in all levels?

imap_processing/hit/hit_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+    dataset["leak_i"] = stacked_leaks
+
+    # Drop the individual leak variables
+    updated_dataset = dataset.drop_vars([f"leak_i_{i:02d}" for i in range(64)])
+
+    return updated_dataset
+
+
+def process_housekeeping_data(
```
maxinelasp, line 131: I feel like this should go either in L0 or L1B

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -1,179 +1,60 @@
...
+    Processes dependencies needed to create L1B data products.
+
     Parameters
     ----------
-    l1a_dataset : xarray.Dataset
-        L1A data.
+    dependencies : dict
+        Dictionary of dependencies that are L1A xarray datasets
+        for science data and a file path string to a CCSDS file
+        for housekeeping data.
```
maxinelasp, line 30: Can you describe the expected structure of the dictionary here? Also, obligatory: if this dictionary should always have the same structure, why not a list of xarray datasets or a dataclass?

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -1,179 +1,60 @@
...
+        # Unpack ccsds file to xarray datasets
+        packet_file = dependencies["imap_hit_l0_raw"]
+        datasets_by_apid = get_datasets_by_apid(packet_file, derived=True)
+        # Process housekeeping to l1b.
+        datasets.append(
+            process_housekeeping_data(
+                datasets_by_apid[HitAPID.HIT_HSKP], attr_mgr, "imap_hit_l1b_hk"
+            )
+        )
+        logger.info("HIT L1B housekeeping dataset created")
```
maxinelasp, line 54: Do you need both of the logger statements here? I'd remove the earlier one and keep this one

imap_processing/hit/hit_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    dataset: xr.Dataset, attr_mgr: ImapCdfAttributes, logical_source: str
+) -> xr.Dataset:
+    """
+    Will process housekeeping dataset for CDF product.
```
sdhoyt, line 135: Can you add something to the docstring to say this function is used for both the L1A and L1B housekeeping?

imap_processing/hit/hit_utils.py
```diff
@@ -0,0 +1,220 @@
+"""
```
sdhoyt, line 1: I like that these were all pulled out into a utils file

imap_processing/tests/hit/test_hit_l1b.py
```diff
@@ -2,51 +2,174 @@
...
+        "active_heater",
+        "last_error_num",
+        "last_bad_cmd",
+        "ref_p5v",
+        "code_checksum",
+        "mode",
+    }
+    # Check that the dataset has the correct variables
+    assert valid_keys == set(hk_dataset.data_vars.keys())
+    assert set(dropped_keys).isdisjoint(set(hk_dataset.data_vars.keys()))
```
sdhoyt, line 116: Nice!

imap_processing/tests/hit/test_hit_l1b.py
```diff
@@ -2,51 +2,174 @@
...
 
-    l1b_hk_dataset = hit_l1b.create_hk_dataset(attr_mgr)
-    assert isinstance(l1b_hk_dataset, xr.Dataset)
+    # Define the coordinates and dimensions. Both have equivalent values
+    dataset_coords_dims = {"epoch", "adc_channels", "adc_channels_label"}
 
+    # Check that the dataset has the correct attributes, coordinates, and dimensions
+    assert hk_dataset.attrs == dataset_attrs
+    assert hk_dataset.coords.keys() == dataset_coords_dims
+    assert hk_dataset.sizes.keys() == dataset_coords_dims
```
sdhoyt: I'm a little confused about how the dataset_coords_dims dictionary tests the `.sizes`. Doesn't sizes return some number for the dataset size?

sdhoyt (commented):
Nice work! I really like how you organized the code for this. Just some minor comments

imap_processing/tests/hit/test_hit_l1b.py
```diff
@@ -2,51 +2,174 @@
...
 
-    # create the attribute manager for this data level
-    attr_mgr = ImapCdfAttributes()
-    attr_mgr.add_instrument_global_attrs(instrument="hit")
-    attr_mgr.add_instrument_variable_attrs(instrument="hit", level="l1b")
-    attr_mgr.add_global_attribute("Data_version", "001")
+def test_hit_l1b_hk_dataset_attributes(hk_dataset):
+    """Test the attributes, dims, and coords in the housekeeping dataset"""
+    # Define the housekeeping dataset attributes
+    dataset_attrs = {
```
maxinelasp, line 124: You can probably just spot-check this since the attribute management should be handled in imap_attribute_manager, you just want to make sure the attributes actually got added properly to your output dataset. IMO it will be annoying to upkeep this test if we ever change, for example, the global attribute variables.

maxinelasp (approved):
Overall looks good to me, I had a few questions/comments, but nothing that will block merging from my view!

imap_processing/hit/l1b/hit_l1b.py
maxinelasp: I actually reread this comment and now I would like to suggest pulling this out into two variables, one for L1A xarrays, one for filepath strings, and maybe make them optional or something. Mixing them together into one dictionary seems dangerous...

imap_processing/tests/hit/test_hit_l1b.py
vmartinez-cu: That shouldn't be there. Somehow the test passes with it. I'll remove this line

imap_processing/cli.py
vmartinez-cu: Great suggestion. I looked into this option earlier but found that `download_dependencies()` returns a list of file paths as PosixPath objects so I wasn't able to access the data level from the dependency. However, it does look like I can access the level through `self.dependencies` like you suggested.

```            
for i, dependency in enumerate(dependencies):
    if self.dependencies[i]['data_level'] == 'l0':
    data_dict["imap_hit_l0_raw"] = dependency
```

imap_processing/hit/hit_utils.py
vmartinez-cu: It's used by `process_housekeeping_data` in the same utils file which is called by l1a and l1b processing code. I pulled it out of `process_housekeeping_data` for readability, but maybe it's more confusing that way since it isn't used outside of housekeeping

imap_processing/hit/hit_utils.py
vmartinez-cu: This function is used by both l1a and l1b processing to create the housekeeping data products. It was originally in L1A code, but I pulled it out to be used by the L1B processing code and avoid redundancy

maxinelasp: K, if it's multi level then this is good! 👍

imap_processing/hit/hit_utils.py
maxinelasp: If it's multi level then I think it's fine. I think it's ok if it's only used in one spot

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: Yeah I don't love this either. It's a weird case because the housekeeping L1B product depends on the the raw ccsds packet file rather than an xarray dataset. I kind of want to see the ccsds packet reading code in cli.py since all instruments need it and that way the only data type getting passed into the different processing levels is xarray datasets. This would also reduce redundancy of the packet reading code being used in every instrument's L0 processing code. Something to discuss with the team

imap_processing/tests/hit/test_hit_l1b.py
vmartinez-cu: Good point. This was mainly to help me with initial validation of attributes for housekeeping, but probably isn't worth keeping in the test file

PR #1121, opened 2024-11-06
Add some more notes to the release workflow

bourque:
This PR updates the Release Workflow docs to add some notes on the expected cadences of the releases, as well as make some minor updates based on how we have been making releases.

### DIFF
--- a/docs/source/development-guide/release-workflow.rst
+++ b/docs/source/development-guide/release-workflow.rst
@@ -18,12 +18,13 @@ are described below. It is assumed that steps (1) through (4) in the
 
 The three repositories are expected to be released at different cadences:
 
-- The ``imap_processing`` repository is released on a monthly cadence (typically the last day of the month, but can
+* The ``imap_processing`` repository is released on a monthly cadence (typically the last day of the month, but can
 vary depending on the availability of developers time to make the release.
-- The ``sds_data_manager`` repository is released on a as-needed basis, typically when enough changes have been made
+* The ``sds_data_manager`` repository is released on a as-needed basis, typically when enough changes have been made
 since the last release to justify a new release.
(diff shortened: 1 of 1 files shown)


docs/source/development-guide/release-workflow.rst
```diff
@@ -27,16 +37,17 @@ Nominal releases
...
    may include commits that add release notes, or update version numbers in various configurations.
 #. Push the version branch to the main ``IMAP-Science-Operations-Center`` ``imap_processing`` repo (i.e. ``upstream``).
 #. Create a `new release <https://github.com/IMAP-Science-Operations-Center/imap_processing/releases>`_, using the
    version branch as the ``Target`` branch. Assign a new tag with the specific version number, including the patch
    number (e.g. ``v0.1.0``).
-#. In GitHub, create a pull request that merges the version branch into ``dev``. Proceed with the nominal review & merge
-   process described in steps (10) and (11) in the :ref:`git & GitHub Workflow <git-and-github-workflow>` section.
+#. If there have been release-specific commits, In GitHub, create a pull request that merges the version branch into
+   ``dev``. Proceed with the nominal review & merge process described in steps (10) and (11) in the :ref:`git & GitHub
+   Workflow <git-and-github-workflow>` section.
```
tech3371, line 47: I am not good with release. I tried to release last month to fix a bug and I think I had to merge PR before I create a `new release`. I tried to see if I can find notes but I couldn't. Is that what you do also? If so, should this step be moved one above.

bourque: Good point, I think you are right about the ordering. That is how I do it for `sds_data_manager`, which usually requires some release-specific commits to update version numbers. 

I updated the docs to reflect the correct order, and also added some more notes.

tech3371 (approved):
Thank you for updating this!

Issue #997, opened 2024-10-09
Move to use SAMMI from the official release


maxinelasp:
### Summary

Right now, we have a copy of SAMMI in imap-processing. To get the latest updates and switch to using the shared code base, we should instead install SAMMI from PyPi. This requires some minor updates to sync the new changes in SAMMI to the imap-processing repo. 

https://github.com/swxsoc/sammi
https://pypi.org/project/sammi-cdf/

### Requirements

  - [ ] Add SAMMI as a dependency
  - [ ] Remove cdf_attribute_manager.py from imap_processing
  - [ ] Update imap_attribute_manager as required to use the new SAMMI API
  - [ ] Remove the cdf/config/shared/ folder and import the schema from SAMMI - we might want to maintain our own copies of these config files, but I think it makes sense to try and use the official SAMMI ones
  - [ ] If needed - update instrument code - hopefully we can isolate changes in imap_attribute_manager for now and update instruments later to take advantage of new features. 

This is a best judgement ticket - if it makes sense to leave some pieces in the repo, we can. Ideally we just want to start moving away from our copy of SAMMI and towards the official release. imap_attribute_manager can act as a halfway point since everyone is pointing at that class. 

Maxine is point of contact for SAMMI questions.

### Related tickets

#606 ?

maxinelasp:
Closed in 1122

PR #1122, opened 2024-11-06
Switch to using SAMMI for CDF attribute management

maxinelasp:
# Change Summary
This removes our copy of code that has been moved to, and improved on, inside SAMMI. 

## Overview

Main changes are inside imap_cdf_manager since that acts as our abstraction layer. The only real thing that changed here is that the schema is overwritten for LABLAXIS because ISTP requires either LABLAXIS or LABL_PTR_i and SAMMI can't make that distinction yet between "required" and "one of two required." 

So, this marks both of them as not required. cdflib can act as a check on that for now. 

## New Dependencies

SAMMI

## Deleted Files

I deleted everything that is duplicated inside of SAMMI!

## Updated Files

I added a small piece in imap_cdf_manager to account for a case where an instrument attr file does not have a version number. 

Updates were either to:
A. Point towards the correct method inside imap_cdf_manager so I could control the flow of information and add IMAP specific assumptions in there
B. Fix ISTP bugs shaken out by SAMMI.

## Testing

I got everything working without skipping any tests or making any major updates. I did remove one of the imap_cdf_manager tests because it was no longer relevant. 

I requested a PR from everyone whose code or tests I touched, feel free to just review those pieces and comment.

Finishes issue #997.


tmplummer (approved):
Great! Looks good to me.

tech3371:
>The only real thing that changed here is that the schema is overwritten for LABLAXIS because ISTP requires either LABLAXIS or LABL_PTR_i and SAMMI can't make that distinction yet between "required" and "one of two required."

@maxinelasp Is SAMMI going to add that check in the future? Just curious

imap_processing/cdf/cdf_attribute_manager.py
bourque, line 1: That sweet blissful feeling of removing entire files from the repo 😀

bourque (approved):
Thanks for doing all this work to get SAMMI up and running!

imap_processing/cdf/cdf_attribute_manager.py
maxinelasp: Me looking at the lines added vs lines removed in this review: 

image

maxinelasp:
> @maxinelasp Is SAMMI going to add that check in the future? Just curious

@tech3371  Yes, we have an issue open to fix that problem, and there is some basic design work done for it already.

imap_processing/cdf/config/imap_variable_schema.yaml
tech3371, line 1: when SAMMI add checks for LABAXIS checks, will this file go away?

tech3371 (approved):
I had one question. Otherwise LGTM as well. Thank you for doing this and SAMMI!

imap_processing/cdf/config/imap_variable_schema.yaml
maxinelasp: Yes it will! Or it probably will, unless we want to override other variables for whatever reason. However, SAMMI is following ISTP, and we are following ISTP, so I expect us to just use the default schema.

vmartinez-cu (approved):

PR #1123, opened 2024-11-07
Small update CoDICE and IDEX data product naming conventions

bourque:
This PR adds the CoDICE `hi-priorities` data product in the list of valid descriptors and updates the `IDEX` data levels to distinguish between L2A and L2B.

Just noting here that the `VALID_DATALEVELS` list in `imap-data-access` will also need to be updated for IDEX, which will be done in a PR over in that repo (https://github.com/IMAP-Science-Operations-Center/imap-data-access/issues/98)


tech3371 (approved):
LGTM

Issue #1115, opened 2024-11-04
BUG - MAG RAW files have duplicate SHCOARSE values


maxinelasp:
### Description of the issue

Processing for L1A was failing due to duplicate epoch timestamps in the CDF files. I sourced this to duplicate SHCOARSE values in the packet. I'm working with the MAG team to diagnose this issue and see what the problem is, but I put in a temporary fix (#1114) which will need to be undone. 

### Steps to reproduce the issue

- Remove the temporary fix
- Run the CLI with the 20240313 file
- See the failure from cdflib: epoch variable is not monotonically increasing

### Expected behavior (What should happen)

Test data epoch values should be monotonically increasing with no duplicates. MAG team is investigating. 

### Code Snippet:

```python
Code
```

maxinelasp:
We determined this was an issue with the source/input files, since the test data was set to the same day for multiple tests. This results in duplicate timestamps and caused issues. To fix, we can query webPODA based on ERT to get only the tests we want and not the duplicate ones. I don't expect this to be an issue in production, because they won't be assigning fake timestamps.

PR #1142, opened 2024-11-12
Fixing mag tests, removing temp fix

maxinelasp:
# Change Summary
Remove the xfail from a MAG test, and fixing issue #1115.

This replaced the test file with a new output file that does not have the issue. More info is in the comment on #1115. TL;DR is that we need to query webPODA in a specific way, because they reassign test data to the same day sometimes.


imap_processing/tests/mag/test_mag_l1b.py
```diff
@@ -94,10 +95,10 @@ def test_mag_compression_scale(mag_l1a_dataset):
     mag_l1a_dataset["vectors"][2, :] = np.array([1, 1, 1, 0])
     mag_l1a_dataset["vectors"][3, :] = np.array([1, 1, 1, 0])
 
-    mag_l1a_dataset["compression_flags"][0, :] = np.array([1, 16])
-    mag_l1a_dataset["compression_flags"][1, :] = np.array([0, 0])
-    mag_l1a_dataset["compression_flags"][2, :] = np.array([1, 18])
-    mag_l1a_dataset["compression_flags"][3, :] = np.array([1, 14])
+    mag_l1a_dataset["compression_flags"][0, :] = np.array([1, 16], dtype=np.int8)
```
maxinelasp, line 98: This change was to more closely imitate the actual data, which caused a bug with raising integers to negative powers.

tech3371 (approved):

PR #1144, opened 2024-11-12
Removed Lo DE segmented packet string commas

### LINKED ISSUE #1143: Lo Segmented packet DE fields  can be spread across packets
For the Lo L0 Direct event segmented packets, we have a new understanding that an individual field within the direct event can be spread across packets. For example, the first 4 bits of a direct event time can be in packet 1 and the last 8 bits of the direct event time can be in packet 2. Commas were previously introduced to the combined binary string to mark where the segments end to help handle padding at the end of each packet to make them divisible by 8. However, since fields can be split across packets, there shouldn't be padding across packets and the comma is no longer needed.

sdhoyt:
# Change Summary

## Overview

Previously added commas to packet binary string to mark where packet segements occurred. Lo has recently communicated a different understanding of how the packets segment which has led to the commas no longer being needed.

Closes #1143


tmplummer (approved):
Easy enough.

tech3371 (approved):

PR #1159, opened 2024-11-13
Add Hexadecimal state handling to excel_to_xtce

### LINKED ISSUE #1158: excel to xtce should handle hexadecimal state values
### Summary

Some telemetry definition housekeeping state values are listed in hexadecimal form. `space-packet-parser` requires states to be in decimal form. This PR adds a function to ensure that the state value is an integer or a hexadecimal string is converted to an integer.

sdhoyt:
# Change Summary

## Overview
Some telemetry definition housekeeping state values are listed in hexadecimal form. `space-packet-parser` requires states to be in decimal form. This PR adds a function to ensure that the state value is an integer or a hexadecimal string is converted to an integer.

Closes #1158

### DIFF
--- a/imap_processing/ccsds/excel_to_xtce.py
+++ b/imap_processing/ccsds/excel_to_xtce.py
@@ -415,12 +415,12 @@ class XTCEGenerator:
             The dictionary for the state.
         """
         value = state["value"]
-        # convert hex string to int
-        if isinstance(value, str) and value.startswith("0x"):
-            state["value"] = int(value, 16)
-            return state
         # return if already an int
-        elif isinstance(value, int):
+        if isinstance(value, int):
+            return state
+        # convert hex string to int
+        elif isinstance(value, str) and value.startswith("0x"):
+            state["value"] = int(value, 16)
             return state
         # raise error if value is neither a hex string or integer
         else:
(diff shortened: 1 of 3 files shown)


imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -391,8 +391,40 @@ def _add_state_conversion(self, row: pd.Series, parameter_type: Et.Element) -> N
...
+        dict
+            The dictionary for the state.
+        """
+        value = state["value"]
+        # convert hex string to int
+        if isinstance(value, str) and value.startswith("0x"):
+            state["value"] = int(value, 16)
+            return state
+        # return if already an int
+        elif isinstance(value, int):
```
vmartinez-cu: If integers are more common than hex values, you could switch the if/elif logic to check if value is an integer first. There might be a slight performance gain there.

vmartinez-cu (approved):
Nice job! One minor comment, but not something that would block a merge

imap_processing/ccsds/excel_to_xtce.py
sdhoyt: Good point, I'll update that

imap_processing/tests/ccsds/test_excel_to_xtce.py
```diff
@@ -197,7 +197,7 @@ def xtce_excel_file(tmp_path):
     states = {
         "packetName": ["TEST_PACKET"] * 2,
         "mnemonic": ["VAR_STATE"] * 2,
-        "value": [0, 1],
+        "value": [0, "0x1"],
```
tech3371: can you add this hex test as another one without deleting that one? I think we need to test that also

tech3371 (commented):
Otherwise, LGTM!

tmplummer (approved):
LGTM

PR #1167, opened 2024-11-19
Update Lo DE Case Decoder for 4, 1 TOF2

### LINKED ISSUE #1165: fix Lo case 4 decompression table
The TOF2 value in the Lo L1A DE case decoder tuple for case 4, 1 should be True instead of False

sdhoyt:
# Change Summary

The Lo Direct Event case decoder for case 4, mode 1 TOF 2 was updated from False to True. This issue was found during validation testing which will be added in a follow-up PR.

Closes #1165


tech3371 (approved):
LGTM

Issue #1055, opened 2024-10-28
HIT L0 - Decompress science data


vmartinez-cu:
Write science packet decompression code

- Decompress science data. Event data do not require decompression
- Jon shared some code that might help with this
- Validate data (validation data is decompressed - Confirmed by Eric Smith 8/6/24)
- Write unit tests

PR #1169, opened 2024-11-20
HIT L0 - data decompression

vmartinez-cu:
This PR adds a function to decom_hit.py to decompress rates data from 16-bit integers to 32-bit integers

## Updated Files
- decom_hit.py
   - new function for decompressing rates data
   - consolidated functions that support finding valid indices for science frames
- constants.py
   - added constants needed for decompressing rates
-test_decom_hit.py
  - add test for decompression function

Closes issue [#1055](https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1055)

### DIFF
--- a/imap_processing/hit/l0/decom_hit.py
+++ b/imap_processing/hit/l0/decom_hit.py
@@ -420,11 +420,22 @@ def decompress_rates_16_to_32(packed: int) -> int:
     power = packed >> MANTISSA_BITS
 
     # Decompress the data depending on the value of the exponent
+    # If the exponent (power) extracted from the packed 16-bit integer is greater
+    # than 1, the compressed value needs to be decompressed by reconstructing the
+    # integer using the mantissa and exponent. If the condition is false, the
+    # compressed and uncompressed values are considered the same.
     if power > 1:
-        decompressed_int = (packed & (output_mask >> EXPONENT_BITS)) | (
-            0x0001 << MANTISSA_BITS
-        )
-        decompressed_int = decompressed_int << (power - 1)
+        # Retrieve the "mantissa" portion of the packed value by masking out the
+        # exponent bits
+        mantissa_mask = output_mask >> EXPONENT_BITS
+        mantissa = packed & mantissa_mask
+
+        # Shift the mantissa to the left by 1 to account for the hidden bit
+        # (always set to 1)
+        mantissa_with_hidden_bit = mantissa | (0x0001 << MANTISSA_BITS)
+
+        # Scale the mantissa by the exponent by shifting it to the left by (power - 1)
+        decompressed_int = mantissa_with_hidden_bit << (power - 1)
     else:
         # The compressed and uncompressed values are the same
         decompressed_int = packed
--- a/imap_processing/hit/l0/decom_hit.py
+++ b/imap_processing/hit/l0/decom_hit.py
@@ -424,6 +424,7 @@ def decompress_rates_16_to_32(packed: int) -> int:
     # than 1, the compressed value needs to be decompressed by reconstructing the
     # integer using the mantissa and exponent. If the condition is false, the
     # compressed and uncompressed values are considered the same.
+    decompressed_int: int
     if power > 1:
         # Retrieve the "mantissa" portion of the packed value by masking out the
         # exponent bits
@@ -440,7 +441,7 @@ def decompress_rates_16_to_32(packed: int) -> int:
         # The compressed and uncompressed values are the same
         decompressed_int = packed
 
-    return int(decompressed_int)
+    return decompressed_int
 
 
 def decom_hit(sci_dataset: xr.Dataset) -> xr.Dataset:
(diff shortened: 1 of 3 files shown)


imap_processing/hit/l0/decom_hit.py
```diff
@@ -409,6 +385,53 @@ def assemble_science_frames(sci_dataset: xr.Dataset) -> xr.Dataset:
...
+    if power > 1:
+        decompressed_int = (packed & (output_mask >> EXPONENT_BITS)) | (
+            0x0001 << MANTISSA_BITS
+        )
+        decompressed_int = decompressed_int << (power - 1)
+    else:
+        # The compressed and uncompressed values are the same
+        decompressed_int = packed
+
+    return int(decompressed_int)
```
laspsandoval: I think decompressed_int is already an integer. Do you need to do int(decompressed_int)?

imap_processing/hit/l0/decom_hit.py
```diff
@@ -409,6 +385,53 @@ def assemble_science_frames(sci_dataset: xr.Dataset) -> xr.Dataset:
...
+    # the value) from the exponent (used for scaling).
+    # set to 16 bits
+    output_mask = 0xFFFF
+
+    # Packed is the compressed integer
+    # Right bit shift to get the exponent
+    power = packed >> MANTISSA_BITS
+
+    # Decompress the data depending on the value of the exponent
+    if power > 1:
```
laspsandoval, line 427: Add comment here to explain what you are doing. You might consider breaking it down into different steps for clarity.

laspsandoval (commented):
Nicely done. I have a couple of comments.

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: Not my preference either. I got a mypy error just leaving it as `decompressed_int` and declaring it an `int` fixes the issue. I'm not sure if there's a better solution to the mypy issue

imap_processing/hit/l0/decom_hit.py
```diff
@@ -409,6 +385,53 @@ def assemble_science_frames(sci_dataset: xr.Dataset) -> xr.Dataset:
...
+    """
+    # In compressed formats, the exponent and mantissa are tightly packed together.
+    # The mask ensures you correctly separate the mantissa (useful for reconstructing
+    # the value) from the exponent (used for scaling).
+    # set to 16 bits
+    output_mask = 0xFFFF
+
+    # Packed is the compressed integer
+    # Right bit shift to get the exponent
+    power = packed >> MANTISSA_BITS
```
sdhoyt, line 419: Based on this, it looks power will only be greater than 1 if packed is >=8192. From the algorithm if the power is less than 1 then the decompressed number = the compressed number. So that makes sense with what you're seeing in the sample data, despite what the algorithm document seems to say

imap_processing/hit/l0/decom_hit.py
```diff
@@ -177,6 +179,10 @@ def parse_count_rates(sci_dataset: xr.Dataset) -> None:
                 low_gain = data[1::2]  # Items at odd indices 1, 3, 5, etc.
                 parsed_data[i] = [high_gain, low_gain]
 
+        # Decompress data where needed
+        if all(x not in field for x in ["hdr", "spare", "pha"]):
+            parsed_data = np.vectorize(decompress_rates_16_to_32)(parsed_data)
```
sdhoyt, line 184: Nice use of vectorize!

sdhoyt (approved):

imap_processing/hit/l0/decom_hit.py
sdhoyt: Would initializing decompressed_int as `decompressed_int: int = ` fix it?

imap_processing/tests/hit/test_decom_hit.py
```diff
@@ -263,6 +240,24 @@ def test_subcom_sectorates(sci_dataset):
         )
 
 
+def test_decompress_rates_16_to_32():
```
sdhoyt: You could use `@pytest.mark.parametrize` here instead of defining that list of tuples and doing a for loop: https://docs.pytest.org/en/stable/how-to/parametrize.html

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: I just took this code from the algorithm doc where it was written in C and translated it to python. I'm not super familiar with bit shifting, but it seems to be working as expected in my tests. I broke down this section into smaller components for readability.  Let me know if this breakdown and the additional comments help.

imap_processing/tests/hit/test_decom_hit.py
vmartinez-cu: Thanks for the suggestion! I haven't used this pytest feature yet so it's great to learn a new approach

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: that worked! I thought that if I did that, I would need to set it to some default value. thanks for the suggestion!

PR #1173, opened 2024-11-20
Mag l1b CDF cleanup

maxinelasp:
# Change Summary
Fixing some ISTP and write_cdf bugs in MAG, shoring up the label variables, and generally making it so the code actually runs again. 

## Overview

Lots of fiddly CDF changes.

### DIFF
--- a/imap_processing/cdf/config/imap_mag_l1_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_mag_l1_variable_attrs.yaml
@@ -1,8 +1,5 @@
 # <=== Label Attributes ===>
 # LABL_PTR_i expects VAR_TYPE of metadata with char data type.
-# We need to define this if we have DEPEND_1 or more.
-# TODO: I am not sure what the FIELDNAM should be.
-# I tried best to match this: https://spdf.gsfc.nasa.gov/istp_guide/variables.html#Metadata_eg1
 default_attrs: &default
   # Assumed values for all variable attrs unless overwritten
   DEPEND_0: epoch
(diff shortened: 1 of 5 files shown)


imap_processing/cdf/config/imap_mag_l1_variable_attrs.yaml
```diff
@@ -3,18 +3,6 @@
 # We need to define this if we have DEPEND_1 or more.
 # TODO: I am not sure what the FIELDNAM should be.
 # I tried best to match this: https://spdf.gsfc.nasa.gov/istp_guide/variables.html#Metadata_eg1
```
tech3371:
```suggestion
```

imap_processing/mag/l0/decom_mag.py
```diff
@@ -129,7 +129,7 @@ def generate_dataset(
         direction.astype(str),
         name="direction_label",
         dims=["direction_label"],
-        attrs=attribute_manager.get_variable_attributes("direction_label"),
+        attrs=attribute_manager.get_variable_attributes("direction_label", check_schema=False),
```
tech3371: cool. Glad that is an option in SAMMI!

Do you think this special case of metadata type for `LABL_PTR_i` attribute will be added to SAMMI in the near future? Because I think that's a standard from SPDF and we use it heavily and new mission may need to in the future.

tech3371 (approved):
minor comments

imap_processing/mag/l0/decom_mag.py
maxinelasp: Yes, it's next up on my list to add, just have some other priorities to finish first! Long term I definitely want to fix up LABL_PTR_i.

Issue #1126, opened 2024-11-07
I-ALiRT: Create SPICE code that will be imported into the lambda


laspsandoval:
### Algorithm Description:

In this directory: /imap_processing/imap_processing/ialirt create code named process_ephemeris.py. We need to create code that **outputs  Time (UTC), Azimuth (UTC),  Elevation (deg.),  Doppler (km/s)**. We want the data to resemble what is shown in the text files here: https://stereo-ssc.nascom.nasa.gov/stations/kiel/

You will use the test data in /imap_processing/imap_processing/tests/spice/test_data for the required kernels. Watch out for gaps in the ephemeris kernel. If that becomes a problem let me know. Please note that some kernels are too large to keep in this repo (de430.bsp for example). Take a look at examples of in the repo of how to temporarily download this kernel (pytest.mark.external_kernel()) if it is needed.

The input for this code will be:
    longitude # longitude in degrees
    latitude # latitude in degrees
    altitude  # altitude in kilometers
    start_utc_input 
    stop_utc_input 
    time_interval # seconds between data points

You can use this as a good starting point:
chilean_partner_example branch in my repo: 
https://github.com/laspsandoval/imap_processing/branches
tools/spice/spice_examples_partners.py

Important:
We want to output the data and not a text file (so not exactly the same as the example). Also, we are using @ensure_spice to furnish the kernels now. Take some time to look at the SPICE use-cases because there are a lot of nuances there. So the starting point code, I expect, will change quite a bit based on the new SPICE tools.

PR #1174, opened 2024-11-21
I-ALiRT - Process Ephemeris

torimarbois:
# Change Summary

## Overview
Create the `process_ephemeris` file, meant to take in location data for a specific ground station partner and output pointing data (azmuth, elevation, doppler). Related issue can be found [here](https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1126).

## New Files

- process_ephemeris.py
   - finds azimuth, elevation, and doppler
- test_process_ephemeris.py
   - unit tests for process_ephemeris.py

## Updated Files

- spice/time.py
   - addition of spice wrapper for `str2et` method, used in process_ephemeris
   - addition of spice wrapper for `et2utc` method, used in process_ephemeris
- test_time.py
   - unit tests for new spice wrappers
- conftest.py
   - updates the external_kernel function (previously named `_download_de440s`) to download an additional external kernel from NAIF
- pyproject.toml
   - adds `convrt` to the list of ignored words for `codespell`, in order to get precommit hooks to pass

## Testing
There are new unit tests for every new method added in this PR, except for the updated fixture in `conftest`.

### Note
The code to calculate the doppler shift is not part of this PR, since it requires some more research. It'll be saved for a separate issue/PR.

### DIFF
--- a/imap_processing/ialirt/process_ephemeris.py
+++ b/imap_processing/ialirt/process_ephemeris.py
@@ -1,0 +1,199 @@
+"""
+Find azimuth (degrees), elevation (degrees), and doppler shift (Hz).
+
+Based on ephemeris data and ground station location (longitude, latitude, altitude).
+
+Reference: https://spiceypy.readthedocs.io/en/main/documentation.html.
+"""
+
+import logging
+import typing
+
+import numpy as np
+import spiceypy as spice
+from numpy import ndarray
+
+from imap_processing.spice.kernels import ensure_spice
+from imap_processing.spice.time import et_to_utc, str_to_et
+
+# Logger setup
+logger = logging.getLogger(__name__)
+
+
+def calculate_doppler() -> int:
+    """
+    Calculate the doppler shift. Placeholder for now.
+
+    Returns
+    -------
+    doppler : int
+        Doppler shift. Currently a throwaway value.
+    """
+    return 1
+
+
+@typing.no_type_check
+@ensure_spice
+def latitude_longitude_to_ecef(
+    longitude: float, latitude: float, altitude: float
+) -> ndarray:
+    """
+    Convert geodetic coordinates to rectangular coordinates.
+
+    Earth-Centered, Earth-Fixed (ECEF) coordinates are a Cartesian coordinate system
+    with an origin at the center of the Earth.
+
+    Parameters
+    ----------
+    longitude : float
+        Longitude in decimal degrees. Positive east of prime meridian, negative to west.
+    latitude : float
+        Latitude in decimal degrees. Positive north of equator, negative to south.
+    altitude : float
+        Altitude in kilometers.
+
+    Returns
+    -------
+    rect_coords : ndarray
+        Rectangular coordinates in kilometers.
+    """
+    latitude_radians = np.deg2rad(latitude)
+    longitude_radians = np.deg2rad(longitude)
+
+    # Retrieve Earth's radii from SPICE
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.
+    # (url cont.) bodvrd
+    radii = spice.bodvrd("EARTH", "RADII", 3)[1]
+    equatorial_radius = radii[0]  # Equatorial radius in km
+    polar_radius = radii[2]  # Polar radius in km
+    flattening = (equatorial_radius - polar_radius) / equatorial_radius
+
+    # Convert geodetic coordinates to rectangular coordinates
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.
+    # (url cont.) georec
+    rect_coords = spice.georec(
+        longitude_radians, latitude_radians, altitude, equatorial_radius, flattening
+    )
+
+    return rect_coords
+
+
+@typing.no_type_check
+@ensure_spice
+def calculate_azimuth_and_elevation(
+    longitude: float,
+    latitude: float,
+    altitude: float,
+    observation_time: float,
+    target: str = "IMAP",
+) -> tuple:
+    """
+    Calculate azimuth and elevation.
+
+    Parameters
+    ----------
+    longitude : float
+        Longitude in decimal degrees. Positive east of prime meridian,
+        negative to west.
+    latitude : float
+        Latitude in decimal degrees. Positive north of equator, negative
+        to south.
+    altitude : float
+        Altitude in kilometers.
+    observation_time : float
+        Time at which the state of the target relative to the observer
+        is to be computed. Expressed as ephemeris time, seconds past J2000 TDB.
+    target : str (Optional)
+        The target body. Default is "IMAP".
+
+    Returns
+    -------
+    azimuth : float
+        Azimuth in degrees.
+    elevation : float
+        Elevation in degrees.
+    """
+    observer_position_ecef = latitude_longitude_to_ecef(longitude, latitude, altitude)
+
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.
+    # (url cont.) azlcpo
+    azel_results = spice.azlcpo(
+        method="Ellipsoid",  # Only method supported
+        target=target,  # target ephemeris object
+        et=observation_time,  # time of observation
+        abcorr="LT+S",  # Aberration correction
+        azccw=False,  # Azimuth measured clockwise from the positive y-axis
+        # TODO: why not clockwise?
+        elplsz=True,  # Elevation increases from the XY plane toward +Z
+        obspos=observer_position_ecef,  # observer position relative to center of motion
+        obsctr="EARTH",  # Name of the center of motion
+        obsref="IAU_EARTH",  # Body-fixed, body-centered reference frame wrt observer's
+        # center
+    )
+
+    # codespell:ignore convrt
+    azimuth = spice.convrt(azel_results[0][1], "RADIANS", "DEGREES")
+    elevation = spice.convrt(azel_results[0][2], "RADIANS", "DEGREES")
+    # TODO: potentially use the velocity components returned from azlcpo to calculate
+    # TODO: doppler
+
+    return azimuth, elevation
+
+
+def build_output(
+    longitude: float,
+    latitude: float,
+    altitude: float,
+    time_endpoints: tuple[str, str],
+    time_step: float,
+) -> dict[str, list]:
+    """
+    Build the output dictionary containing time, azimuth, elevation, and doppler.
+
+    Parameters
+    ----------
+    longitude : float
+        Longitude in decimal degrees. Positive east of prime meridian, negative to west.
+    latitude : float
+        Latitude in decimal degrees. Positive north of equator, negative to south.
+    altitude : float
+        Altitude in kilometers.
+    time_endpoints : tuple[str, str]
+        Start and stop times in UTC.
+    time_step : float
+        Seconds between data points.
+
+    Returns
+    -------
+    output_dict: dict[str, list]
+        Keys are time, azimuth, elevation and doppler. Values are calculated for every
+        timestamp between start_utc_input and stop_utc_input, spaced by time_step.
+    """
+    output_dict: dict[str, list] = {
+        "time": [],
+        "azimuth": [],
+        "elevation": [],
+        "doppler": [],
+    }
+
+    start_et_input = str_to_et(time_endpoints[0])
+    stop_et_input = str_to_et(time_endpoints[1])
+
+    # For now, assume that kernel management will be handled by ensure spice
+
+    for obs_time in np.arange(start_et_input, stop_et_input, time_step):
+        azimuth, elevation = calculate_azimuth_and_elevation(
+            longitude, latitude, altitude, obs_time
+        )
+
+        output_dict["time"].append(et_to_utc(obs_time, format_str="ISOC"))
+        output_dict["azimuth"].append(azimuth)
+        output_dict["elevation"].append(elevation)
+        output_dict["doppler"].append(calculate_doppler())
+
+    logger.info(
+        f"Calculated azimuth, elevation and doppler for time range from "
+        f"{start_et_input} to {stop_et_input}."
+    )
+
+    return output_dict
--- a/imap_processing/ialirt/process_ephemeris.py
+++ b/imap_processing/ialirt/process_ephemeris.py
@@ -8,11 +8,13 @@ Reference: https://spiceypy.readthedocs.io/en/main/documentation.html.
 
 import logging
 import typing
+from typing import Union
 
 import numpy as np
 import spiceypy as spice
 from numpy import ndarray
 
+from imap_processing.spice.geometry import SpiceBody
 from imap_processing.spice.kernels import ensure_spice
 from imap_processing.spice.time import et_to_utc, str_to_et
 
@@ -20,16 +22,28 @@ from imap_processing.spice.time import et_to_utc, str_to_et
 logger = logging.getLogger(__name__)
 
 
-def calculate_doppler() -> int:
+@typing.no_type_check
+def calculate_doppler(
+    observation_time: Union[float, np.ndarray],
+) -> Union[int, ndarray[float]]:
     """
     Calculate the doppler shift. Placeholder for now.
 
+    Parameters
+    ----------
+    observation_time : float or np.ndarray
+        Time at which the state of the target relative to the observer
+        is to be computed. Expressed as ephemeris time, seconds past J2000 TDB.
+
     Returns
     -------
-    doppler : int
+    doppler : float or np.ndarray[float]
         Doppler shift. Currently a throwaway value.
     """
-    return 1
+    if isinstance(observation_time, np.ndarray):
+        return np.ones(len(observation_time), dtype=float)
+    else:
+        return 1
 
 
 @typing.no_type_check
@@ -61,16 +75,16 @@ def latitude_longitude_to_ecef(
     longitude_radians = np.deg2rad(longitude)
 
     # Retrieve Earth's radii from SPICE
-    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.
-    # (url cont.) bodvrd
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.bod
+    # (url cont.) vrd
     radii = spice.bodvrd("EARTH", "RADII", 3)[1]
     equatorial_radius = radii[0]  # Equatorial radius in km
     polar_radius = radii[2]  # Polar radius in km
     flattening = (equatorial_radius - polar_radius) / equatorial_radius
 
     # Convert geodetic coordinates to rectangular coordinates
-    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.
-    # (url cont.) georec
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.geo
+    # (url cont.) rec
     rect_coords = spice.georec(
         longitude_radians, latitude_radians, altitude, equatorial_radius, flattening
     )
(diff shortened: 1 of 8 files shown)


imap_processing/ialirt/process_ephemeris.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.
+    # (url cont.) azlcpo
+    azel_results = spice.azlcpo(
+        method="Ellipsoid",  # Only method supported
+        target=target,  # target ephemeris object
+        et=observation_time,  # time of observation
+        abcorr="LT+S",  # Aberration correction
+        azccw=False,  # Azimuth measured clockwise from the positive y-axis
+        # TODO: why not clockwise?
```
torimarbois: @laspsandoval I wanted to double check this Boolean parameter - its default is True, for measuring Azimuth clockwise, and I was wondering why you had set it to False for counterclockwise in your example code?

imap_processing/tests/spice/test_time.py
```diff
@@ -50,3 +56,47 @@ def test_sct2e_wrapper(sclk_ticks):
...
+    assert expected_et == actual_et
+
+    # Test list input
+    list_of_utc = [
+        "2017-08-14T19:46:00.000",
+        "2017-09-14T19:46:00.000",
+        "2017-10-14T19:46:00.000",
+    ]
+    # TODO: would this ^ be more of a realistic expectation as an array rather than a
+    # list?
```
torimarbois: Just a general question for people reviewing here. I'm not sure what type of input is more likely when an iterable needs to be passed into this function - an array or a list?

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    # Example usage for Longovilo-Hacienda, Chile
+    # https://www.findlatitudeandlongitude.com/l/LONGOVILO%2C+CHILE/
+    longitude = -71.41  # longitude in degrees
+    latitude = -33.94  # latitude in degrees
+    altitude = 0.157  # altitude in kilometers
+    time_endpoints = ("2025 AUG 17 00:00:00", "2025 AUG 18 00:00:00")
+    time_interval = int(1e3)  # seconds between data points
+
+    kernels = [
```
laspsandoval, line 91: Which of these kernels are necessary? For example, the de440s.bsp is used as a sanity check to compare with the sun's azimuth and elevation. Otherwise it isn't necessary for us.

imap_processing/ialirt/process_ephemeris.py
laspsandoval: I used that because I was doing a sanity check w/ pvlib.solarposition.get_solarposition. I think that it doesn't matter what we use as long as we are consistent.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -0,0 +1,199 @@
...
+        longitude_radians, latitude_radians, altitude, equatorial_radius, flattening
+    )
+
+    return rect_coords
+
+
+@typing.no_type_check
+@ensure_spice
+def calculate_azimuth_and_elevation(
+    longitude: float,
```
laspsandoval, line 98: We kind of talked about this. Can we put the longitude, latitude, and altitude in a config file and use the name of the station instead?

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -0,0 +1,199 @@
...
+        timestamp between start_utc_input and stop_utc_input, spaced by time_step.
+    """
+    output_dict: dict[str, list] = {
+        "time": [],
+        "azimuth": [],
+        "elevation": [],
+        "doppler": [],
+    }
+
+    start_et_input = str_to_et(time_endpoints[0])
```
laspsandoval, line 192: I think that you may want to do some error handling here to make certain that the time range that you are using is what is available in the ephemeris kernels (spkcov).

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -0,0 +1,199 @@
...
+        # TODO: why not clockwise?
+        elplsz=True,  # Elevation increases from the XY plane toward +Z
+        obspos=observer_position_ecef,  # observer position relative to center of motion
+        obsctr="EARTH",  # Name of the center of motion
+        obsref="IAU_EARTH",  # Body-fixed, body-centered reference frame wrt observer's
+        # center
+    )
+
+    # codespell:ignore convrt
+    azimuth = spice.convrt(azel_results[0][1], "RADIANS", "DEGREES")
```
laspsandoval: You used np.deg2rad earlier, which is probably more efficient here than a SPICE call.

imap_processing/tests/conftest.py
```diff
@@ -46,45 +46,48 @@ def _autoclear_spice():
 
 
 @pytest.fixture(scope="session")
-def _download_de440s(spice_test_data_path):
-    """This fixture downloads the de440s.bsp kernel into the
-    tests/spice/test_data directory if it does not already exist there. The
+def _download_external_kernels(spice_test_data_path):
```
laspsandoval, line 49: Do we need these for the tests? It is possible that we don't and this change isn't necessary.

laspsandoval (commented):
Nice work. Thanks for doing this. Just as a sanity check were you getting the same answer as what I had before?

I'll let Tim review time.py and test_time.py since those are his programs.

I had a handful of minor comments, but overall looks good.

laspsandoval:
One more thing that I thought of. If we DO end up needing pck00011.tpc make certain to include it in gitignore similar to de440.bsp:

Ignore specific SPICE kernels that get downloaded from NAIF automatically for tests
marked with @pytest.mark.external_kernel
**/de440*.bsp

imap_processing/ialirt/process_ephemeris.py
tmplummer: I haven't really considered the workflow for providing this data to the ground station partners. Have you, Laura? My assumption is that it will always be using predicted ephemeris and thus it would be sufficient to assume it will pretty much just want to furnish whatever files the ephemeris metakernel points to. I would advocate for just letting spiceypy raise an exception if the proper coverage isn't provided by the furnished kernels.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@typing.no_type_check
+@ensure_spice
+def calculate_azimuth_and_elevation(
+    longitude: float,
+    latitude: float,
+    altitude: float,
+    observation_time: float,
+    target: str = "IMAP",
```
tmplummer: There is a SpiceBody class in `imap_processing.spice.geometry.py` that should be used here.
```suggestion
    target: SpiceBody = SpiceBody.IMAP,
```

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -0,0 +1,199 @@
...
+    elevation : float
+        Elevation in degrees.
+    """
+    observer_position_ecef = latitude_longitude_to_ecef(longitude, latitude, altitude)
+
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.
+    # (url cont.) azlcpo
+    azel_results = spice.azlcpo(
+        method="Ellipsoid",  # Only method supported
+        target=target,  # target ephemeris object
```
tmplummer: If you use a `SpiceBody` for the target input type, this will need to be:
```suggestion
        target=target.name,  # target ephemeris object
```

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -0,0 +1,199 @@
...
+    return rect_coords
+
+
+@typing.no_type_check
+@ensure_spice
+def calculate_azimuth_and_elevation(
+    longitude: float,
+    latitude: float,
+    altitude: float,
+    observation_time: float,
```
tmplummer: I would suggest allowing for this to be a single time or a `np.ndarray` of times. Essentially moving the looping that is currently implemented in the outer `build_output` function into this function. `observer_position_ecef` will always be the same for each ground station and so the loop will only need to be implemented around the `spice.azlcpo` function.

imap_processing/spice/time.py
```diff
@@ -90,3 +90,75 @@ def _sct2e_wrapper(
         return np.array([spice.sct2e(IMAP_SC_ID, s) for s in sclk_ticks])
     else:
         return spice.sct2e(IMAP_SC_ID, sclk_ticks)
+
+
+@typing.no_type_check
+@ensure_spice
+def str_to_et(
+    time_str: Union[str, list[str]],
```
tmplummer: Changing from `list` to ~~`Collection`~~ `Iterable` makes this more generic so that it can accept things like a `ndarray` of strings.
```suggestion
    time_str: Union[str, Iterable[str]],
```
This change will need to be coupled with a flip of the `if` statement to check for a `str` instance instead of `list`.

imap_processing/spice/time.py
```diff
@@ -90,3 +90,75 @@ def _sct2e_wrapper(
...
+
+
+@typing.no_type_check
+@ensure_spice
+def et_to_utc(
+    et: Union[float, np.ndarray],
+    format_str: str = "ISOC",
+    precision: int = 3,
+    utclen: int = 24,
+) -> Union[str, Collection[str]]:
```
tmplummer: Flip the et type hint with the output type hint.
```suggestion
def et_to_utc(
    et: Union[float, Iterable[float]],
    format_str: str = "ISOC",
    precision: int = 3,
    utclen: int = 24,
) -> Union[str, np.ndarray]]:
```
BTW, I see that spiceypy uses `Iterable` rather than `Collection`. That is probably the correct type hint here though they are subtly different.

imap_processing/spice/time.py
```diff
@@ -90,3 +90,75 @@ def _sct2e_wrapper(
...
+
+    Returns
+    -------
+    utc_time : str or Collection[str]
+        UTC time(s).
+    """
+    if isinstance(et, np.ndarray):
+        return np.array([spice.et2utc(t, format_str, precision, utclen) for t in et])
+    else:
+        return spice.et2utc(et, format_str, precision, utclen)
```
tmplummer: A few of the spiceypy functions are already vectorized. This one is so, this can just be
```suggestion
    return spice.et2utc(et, format_str, precision, utclen)
```

imap_processing/tests/conftest.py
```diff
@@ -46,45 +46,48 @@ def _autoclear_spice():
...
+    """This fixture downloads the de440s.bsp and pck00010.tpc kernels into the
```
tmplummer:
```suggestion
    """This fixture downloads the de440s.bsp and pck00011.tpc kernels into the
```

imap_processing/tests/conftest.py
tmplummer: pck00011.tpc contains the Earth radii which is required by the `spice.bodvrd("EARTH", "RADII", 3)` call. The alternative would be that the radii or flattening could be hard coded as it should not change.

imap_processing/tests/conftest.py
```diff
@@ -46,45 +46,48 @@ def _autoclear_spice():
...
-                    f"Exception: {error}"
-                )
-                time.sleep(1)
-            else:
-                logger.error(
-                    f"Failed to download file after {allowed_attempts} "
-                    f"attempts, Final Error: {error}"
-                )
-                raise
+    kernel_urls = [
```
tmplummer, line 55: Thank you for making the changes to this function! Regardless of whether the new kernel gets kept or not, this implementation makes this fixture better.

imap_processing/tests/conftest.py
```diff
@@ -46,45 +46,48 @@ def _autoclear_spice():
...
+        "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/de440s.bsp",
+        "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/pck00011.tpc",
```
tmplummer, line 57: Please add `pck00011.tpc` to the `.gitignore` so that we don't have to worry about accidentally checking it in. There is and entry for `de440s.bsp` in there already to group it with.

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
```diff
@@ -0,0 +1,96 @@
...
+
+    kernels = ["pck00011.tpc"]
+    with furnish_kernels(kernels):
+        rect_coords = process_ephemeris.latitude_longitude_to_ecef(
+            longitude, latitude, altitude
+        )
+
+    assert np.round(rect_coords[0], 6) == -2595.359123
+    assert np.round(rect_coords[1], 6) == 4881.160589
+    assert np.round(rect_coords[2], 6) == 3170.373523
```
tmplummer: Using `np.testing.assert_allclose()` for comparison is a great way to not have to implement rounding. The benefit is that the output is really nice when it fails.
```suggestion
    np.testing.assert_allclose(rect_coords, [-2595.359123, 4881.160589, 3170.373523], atol=1e-6)
```

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
```diff
@@ -0,0 +1,96 @@
...
+def test_calculate_azimuth_and_elevation(furnish_kernels):
+    """
+    Test the calculate_azimuth_and_elevation() function.
+    """
+    longitude = -71.41  # longitude in degrees
+    latitude = -33.94  # latitude in degrees
+    altitude = 0.157  # altitude in kilometers
+    observation_time = 805794429.1837295  # "2025-07-14T19:46:00.000"
+
+    kernels = ["pck00011.tpc", "de440s.bsp"]
```
tmplummer: Just curious if you tried this test without the `de440s.bsp` kernel?

imap_processing/tests/spice/test_time.py
tmplummer: I commented above on the function. Allowing either would be great!

tmplummer (commented):
This looks really good. Very clean code with great documentation. I left some feedback to attend to.

imap_processing/ialirt/process_ephemeris.py
laspsandoval: Good point. You are probably right. It will likely always be ephemeris and we can add error handling later if it isn't the case.

imap_processing/tests/conftest.py
laspsandoval: Yep. Ok then leave it.

tech3371:
How soon do you guys think you will be able to finish this PR? The reason I ask is, I have some works that's waiting for this feature. I don't think I will get to it this month but love to get to next month(January 2025). It would be very helpful if we can have this code merged and included in December release so that it can be used later. These are the three functions I need for January work.
> met_to_J2000ns()
J2000ns_to_et()
et_to_UTC()

I think last two are works of this PR. Is that right @torimarbois @subagonsouth?

tmplummer:
> How soon do you guys think you will be able to finish this PR? The reason I ask is, I have some works that's waiting for this feature. I don't think I will get to it this month but love to get to next month(January 2025). It would be very helpful if we can have this code merged and included in December release so that it can be used later. These are the three functions I need for January work.
> 
> > met_to_J2000ns()
> > J2000ns_to_et()
> > et_to_UTC()
> 
> I think last two are works of this PR. Is that right @torimarbois @subagonsouth?

@tech3371 
The first two are already implemented. See:
 - https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/time.py#L17
 - https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/time.py#L47
    - Note that J2000 seconds is the same as ET.
Yes, the third is in this PR.

Since MET is sort of the standard time definition for this mission, I can also implement a `met_to_UTC` function. I'll write a ticket for that and make sure to get it done this month. Should be an easy one.

torimarbois:
> How soon do you guys think you will be able to finish this PR? The reason I ask is, I have some works that's waiting for this feature. I don't think I will get to it this month but love to get to next month(January 2025). It would be very helpful if we can have this code merged and included in December release so that it can be used later. These are the three functions I need for January work.
> 
> > met_to_J2000ns()
> > J2000ns_to_et()
> > et_to_UTC()
> 
> I think last two are works of this PR. Is that right @torimarbois @subagonsouth?

@tech3371 the goal is to get it merged sometime next week!

tech3371:
Thank you @torimarbois @subagonsouth for help on this! Really appreciate it

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
torimarbois: I was able to remove a few of those listed kernels with the tests continuing to pass, but the `azlcpo` spice function needs a few variables within `de440s.bsp`.

imap_processing/ialirt/process_ephemeris.py
torimarbois: I'm thinking I wait to do that until we have some more clarity on how this file will be used by the lambda function. That way we can make sure all of the input configuration is definitely what we need it to be. Or I could include that in the next chunk of work for adding the doppler calculation. Do you have a preference?

imap_processing/tests/conftest.py
torimarbois: Thanks for chiming in here Tim!

imap_processing/ialirt/process_ephemeris.py
torimarbois: This switch doesn't make sense to me, since SpiceBody.IMAP returns an integer and the `azlcpo` function expects a string. Maybe I'm misunderstanding the SpiceBody Enum? @subagonsouth

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
torimarbois: Yep, when I remove it the test fails with `Insufficient ephemeris data has been loaded to compute the state of 399 (EARTH) relative to 0 (SOLAR SYSTEM BARYCENTER) at the ephemeris epoch 2025 JUL 14 19:47:09.183.
 azlcpo_c --> AZLCPO --> SPKCPO --> SPKCVO --> ZZSPKFZO --> ZZSPKFAO --> SPKSSB --> SPKGEO`. So I'm going to leave it in!

torimarbois:
Also, I created plots to make sure that this new code created similar results to the sample code that Laura wrote. It does! Here are the plots produced by the sample code:
image
And here are the plots produced by the code in this PR:
image

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
laspsandoval: yep you are right.

imap_processing/ialirt/process_ephemeris.py
laspsandoval: You can include it in another PR. That's fine.

laspsandoval (approved):
Looks good to me. Thanks for the updates! Nice work.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -0,0 +1,226 @@
...
+                abcorr="LT+S",  # Aberration correction
+                azccw=False,  # Azimuth measured clockwise from the positive y-axis
+                elplsz=True,  # Elevation increases from the XY plane toward +Z
+                obspos=observer_position_ecef,  # observer pos. to center of motion
+                obsctr="EARTH",  # Name of the center of motion
+                obsref="IAU_EARTH",  # Body-fixed, body-centered reference frame wrt
+                # observer's center
+            )
+            azimuth.append(np.rad2deg(azel_results[0][1]))
+            elevation.append(np.rad2deg(azel_results[0][2]))
```
tmplummer: Does this function need to accept a `float` input for `observation_time`? I try to avoid repetition like is done here where there are a handful of hard coded values in 2 different locations. One possible fix I thought of would be to force `observation_time` to be an ndarray even if it only has one value. I'll leave this up to you to decide.

tmplummer (approved):
Great! Thanks for addressing all of those comments.

imap_processing/ialirt/process_ephemeris.py
torimarbois: At the time, when I changed this, I couldn't find a way to pass in a singular float for observation time without breaking it out into this `if not` logic. But my brain was helpful today and I was able to reduce the repetition by simply casting the single input value to a list.

PR #1191, opened 2024-11-27
Lo - Change esa_step coord name in L1A attrs

### LINKED ISSUE #1190: Rename ESA step coord in Lo L1A CDF Attributes
The Lo Direct Events and Histogram L1A data products both have ESA Step fields, however, the ESA step in the histogram product is a coordinate and the esa step in the direct events is data. Because of the differences, the ESA step field currently in the attributes yaml cannot be re-used for the direct events.

sdhoyt:
# Change Summary

## Overview

The Lo Direct Events and Histogram L1A data products both have ESA Step fields, however, the ESA step in the histogram product is a coordinate and the esa step in the direct events is data. Because of the differences, the ESA step field currently in the attributes yaml cannot be re-used for the direct events. This PR changes the name of the ESA step for the histogram to specify that it is a coordinate and adds a new ESA step field for the direct event data.

There were also missing fields (passes and direct events label) that were added and many of the histogram attributes were missing a label axis field

Finally, the name for the esa_step coordinate field was updated in the Lo L1A processing code.

Closes #1190

### DIFF
--- a/imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
@@ -58,6 +58,8 @@ hist_az_60_default: &hist_az_60_default
   DEPEND_0: epoch
   DEPEND_1: azimuth_60
   DEPEND_2: esa_step
+  LABL_PTR_1: azimuth_60_label
+  LABL_PTR_2: esa_step_label
 
 hist_az_6_default: &hist_az_6_default
   <<: *default_uint32
@@ -65,6 +67,8 @@ hist_az_6_default: &hist_az_6_default
   DEPEND_0: epoch
   DEPEND_1: azimuth_6
   DEPEND_2: esa_step
+  LABL_PTR_1: azimuth_6_label
+  LABL_PTR_2: esa_step_label
 
 spin_default_attrs: &spin_default
   <<: *default
@@ -88,7 +92,7 @@ esa_step_label:
    FORMAT: A1
    VAR_TYPE: metadata
 
-esa_step:
+esa_step_coord:
   <<: *coord_default
   <<: *default_uint8
   VALIDMIN: 1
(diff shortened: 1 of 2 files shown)


vmartinez-cu (approved):
LGTM!

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -129,6 +139,17 @@ de_time:
...
+esa_step:
+  <<: *de_default
+  <<: *default_uint8
+  VALIDMIN: 1
+  VALIDMAX: 7
+  FORMAT: I1
+  CATDESC: Energy Step
+  DEPEND_1: esa_step
+  FIELDNAM: Energy Step
+  LABLAXIS: ESA
```
tmplummer, line 151: It looks like the only difference between `esa_step_coord` and `esa_step` is that the former has `LABL_PTR_1: esa_step_label`? You could `esa_step` the basis for `esa_step_coord` so that it looks like:
```
esa_step: &esa_step
  <<: *de_default
  <<: *default_uint8
  VALIDMIN: 1
  VALIDMAX: 7
  FORMAT: I1
  CATDESC: Energy Step
  DEPEND_1: esa_step
  FIELDNAM: Energy Step
  LABLAXIS: ESA

esa_step_coord:
  <<: *esa_step
  LABL_PTR_1: esa_step_label
```

tmplummer (approved):
Looks good. Just the one suggestion.

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
sdhoyt: That's a good suggestion. The only issue is `esa_step` is using `de_default` while `esa_step_coord` is using `coord_default`. The `de_default` sets a `DEPEND_1` that the `esa_step_coord` isn't using

PR #1196, opened 2024-12-03
700 refactor hi l1a histogram

### LINKED ISSUE #700: Refactor Hi L1a histogram
### Algorithm Description:
Due to changes in PR #687, the `allocate_histogram_dataset` function is probably no longer needed. Check the TODOs in the `create_dataset` function and refactor the code as you see fit.

Note from the PR:

> The allocate_histogram_dataset function was really just my way of avoiding having an intermediate data storage. I wanted to go directly from the packet into the xr.DataSet arrays. With your new function, this is no longer achieving that goal. I will write a ticket to look at this Hi code and address the TODOs.

In addition to the above, this should also address the TODO on line ~60:
```
    # TODO: Look into avoiding the for-loops below
    #       It seems like we could try to reshape the arrays and do some numpy
    #       broadcasting rather than for-loops directly here. Ticket: #700
```

tmplummer:
# Change Summary

## Overview
This is a major overhaul of the Hi L1A hist processing. It was partially necessitated by an update to the SCI_CNT packet definition. The other changes were driven by the implementation of the `packet_file_to_datasets` function which was written after this code was first implemented.

## Updated Files
   - Added default entries for CCSDS header entries that were previously not decoded from the packets.
   - Added CCSDS header entries to the HIST CDF variable attributes.
   - No longer allocate Dataset from scratch. Instead, the L1A CDF is derived from the packet Dataset that is passed in.
   - Update attributes for variables already in the input dataset.
   - Only add data variables that are not already present in the input dataset.
   - Extract method for decoding the 90 element 12-bit counter arrays.
   - Remove test for removed function: `allocate_histogram_dataset`
   - Add test coverage for new function: `unpack_hist_counter`

Closes: #700

### DIFF
--- a/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
@@ -55,6 +55,54 @@ default_float32_attrs: &default_float32
   VALIDMAX: 3.4028235e+38
   dtype: float32
 
+default_ccsds_version: &default_ccsds_version
+  <<: *default_uint8
+  CATDESC: CCSDS packet version number
+  FIELDNAM: CCSDS version
+  VAR_TYPE: support_data
+
+default_ccsds_type: &default_ccsds_type
+  <<: *default_uint8
+  CATDESC: CCSDS packet type
+  FIELDNAM: CCSDS type
+  VAR_TYPE: support_data
+
+default_ccsds_sec_hdr_flg: &default_ccsds_sec_hdr_flg
+  <<: *default_uint8
+  CATDESC: CCSDS secondary header flag
+  FIELDNAM: CCSDS secondary header flag
+  VAR_TYPE: support_data
+
+default_ccsds_pkt_apid: &default_ccsds_pkt_apid
+  <<: *default_uint16
+  CATDESC: CCSDS application process ID
+  FIELDNAM: CCSDS APID
+  VAR_TYPE: support_data
+
+default_ccsds_seq_flgs: &default_ccsds_seq_flgs
+  <<: *default_uint8
+  CATDESC: CCSDS sequence flags
+  FIELDNAM: CCSDS sequence flags
+  VAR_TYPE: support_data
+
+default_ccsds_src_seq_ctr: &default_ccsds_src_seq_ctr
+  <<: *default_uint16
+  CATDESC: CCSDS source sequence counter
+  FIELDNAM: CCSDS sequence counter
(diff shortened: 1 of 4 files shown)


imap_processing/cdf/config/imap_hi_variable_attrs.yaml
```diff
@@ -55,6 +55,54 @@ default_float32_attrs: &default_float32
   VALIDMAX: 3.4028235e+38
   dtype: float32
 
+default_ccsds_version: &default_ccsds_version
+  <<: *default_uint8
+  CATDESC: CCSDS packet version number
+  FIELDNAM: CCSDS version
+  VAR_TYPE: support_data
+
```
vmartinez-cu, line 63: These look great! I'm borrowing these ccsds header definitions for HIT : )

vmartinez-cu (approved):
Great job!

imap_processing/hi/l1a/histogram.py
```diff
@@ -41,115 +41,113 @@ def create_dataset(input_ds: xr.Dataset) -> xr.Dataset:
...
         # Inject counter name into generic counter attributes
         counter_attrs = default_counter_attrs.copy()
         for key, val in counter_attrs.items():
             if isinstance(val, str) and "{counter_name}" in val:
                 counter_attrs[key] = val.format(counter_name=counter_name)
-        data_vars[counter_name] = xr.DataArray(
-            data=np.empty((num_packets, len(coords["angle"])), np.uint16),
+        # Instantiate the counter DataArray
+        new_vars[counter_name] = xr.DataArray(
+            data=unpack_hist_counter(input_ds[counter_name].data.sum()),
```
laspsandoval, line 103: So I am wondering if you might want to do this:

new_vars[counter_name] = xr.DataArray(
    data=np.array(
        [unpack_hist_counter(epoch_data) for epoch_data in input_ds[counter_name].data]
    ),
    dims=["epoch", "angle"],
    attrs=counter_attrs,
)

Does unpack_hist_counter depend on receiving one epoch's worth of data at a time to compute values correctly? Double check this because I am not certain either. But just check to make certain.

laspsandoval (commented):
I had one thing for you to check before you merge.

imap_processing/hi/l1a/histogram.py
tmplummer: It is correct as is. I wrote `unpack_hist_counter` to avoid that for loop. The docstring lets the user know that is the case.

PR #1201, opened 2024-12-03
Lo - Update XTCE to fix HK enum state issue

### LINKED ISSUE #1197: Update Lo XTCE to fix HK Enum state bug
The Lo telemetry definition sheet has an issue where the HK state enums do not cover all possible values for certain fields and fails on the sample data

sdhoyt:
# Change Summary

## Overview

The Lo telemetry sheet did HK state enums did not cover the full set of possible values resulting in errors from space packet parser when decommutating sample data. A new telemetry sheet has been delivered resolving this error.

Closes #1197

### DIFF
--- a/imap_processing/lo/packet_definitions/lo_xtce.xml
+++ b/imap_processing/lo/packet_definitions/lo_xtce.xml
@@ -24,6 +24,2555 @@
 			<xtce:IntegerParameterType name="PKT_LEN" signed="false">
 				<xtce:IntegerDataEncoding sizeInBits="16" encoding="unsigned" />
 			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="ILO_APP_SHK.SHCOARSE" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="32" encoding="unsigned" />
+			</xtce:IntegerParameterType>
+			<xtce:EnumeratedParameterType name="ILO_APP_SHK.BOARD_TYPE" signed="false">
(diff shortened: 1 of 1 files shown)


tmplummer (approved):
Do you decode any sample data in tests or is there any way to confirm this fix? I suppose that it can be done through the validation. Just a thought.

nkerman (approved):
The format and composition of the XTCE file looks good to me! I'm not entirely sure how to evaluate the content itself - i.e., if there's some ground truth of the packet definition which it would be good to compare this against?

sdhoyt:
> Do you decode any sample data in tests or is there any way to confirm this fix? I suppose that it can be done through the validation. Just a thought.

The original error was caught in a unit test which is now passing, so I believe the issue is fixed

bourque (approved):
LGTM!

imap_processing/lo/packet_definitions/lo_xtce.xml
tech3371, line 1: I have similar question as others. XTCE looks good to me. Are you planning to decom its data in upcoming PR? If so, ignore my comment. Right now, it seems like `lo_l1a.py` unpacks `LoAPID.ILO_SCI_CNT.

sdhoyt: The sample data I have for Lo contains all of the APIDs, so HK is being decommutated every time I use the sample data with lo_l1a.py, I'm just not writing the CDF for HK yet. But because it's always being decommutated when I use the sample data, I was getting errors when testing lo_l1a for the science counts and am no longer getting errors.

PR #1202, opened 2024-12-03
Draft: #1087 Create Ultra l2 pipeline structure and build exposure time maps, combined across multiple pointings

### LINKED ISSUE #1087: Ultra l2: Calculate the Total Exposure Time in the Spacecraft Frame for Nat
### Algorithm Description:
Using the MATLAB code as a guide (and test data) and the structure of the l1c pset counts in the spacecraft frame calculate the total exposure time for l2.

nkerman:
# Change Summary

## Overview

Create the basic structure of Ultra L2 processing and combine exposure time across pointings.

IMAP Ultra L2 Processing takes multiple L1C files, each with pointing sets containing counts, exposure times, sensitivities, etc. as a function of spatial coordinates (e.g. azimuth, elevation) and energy bins. The spatial coordinates are in different despun pointing set (DPS) frames for each different pointing (and thus each L1c file). The L2 pipeline projects these into an ecliptic, inertial frame (currently `J2000ECLIP`), where it combines counts, exposure times, etc. and calculates flux, etc.

Currently, this code only combines exposure time. It also combines counts, but in a simplified form. This code is not currently disambiguating between the L1C spacecraft frame values and the L1C heliocentric frame values.

Closes #1087

## New Dependencies

None, though I expect `Healpy` may become necessary in the near future.

## New Files

- new file 1

## Deleted Files

- deleted file 1

## Updated Files

- updated imap_processing/spice/geometry.py
    - Allow output of sph2cart in radians
- added imap_processing/tests/spice/test_data/imap_frames_demo_euler.tf
    - Test spice data used by Ultra instrument team
- added imap_processing/tests/spice/test_data/imap_ultra_test_metakernel.template
    - Metakernel which includes above frame kernel
- added imap_processing/ultra/l2/l2_utils.py
    - Util functions for use in L2: e.g., build_solid_angle_map
- added imap_processing/tests/ultra/unit/test_l2_utils.py
    - Tests for util functions
- added imap_processing/ultra/l2/ultra_l2.py
    - Main L2 processing file. Defines how to input a list of L1C files and combine values across pointings
- added imap_processing/tests/ultra/unit/test_ultra_l2.py
    - Tests for functions in ultra_l2
- updated poetry.lock
    - This was automatically updated by pre-commit hooks. Not sure if I need to change anything here.
- updated pyproject.toml
    - I had to add a pytest mark for use_test_metakernel to get it to work in my pytests... Not sure about this either.
## Testing

- Adds about a dozen new unit tests


imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+from imap_processing.ultra.l2 import l2_utils
+
+# Set default values:
+DEFAULT_SPACING_DEG = 0.5
+
+logger = logging.getLogger(__name__)
+logger.info("Importing ultra_l2 module")
+
+
+def is_ultra45(
```
tmplummer, line 25: Is this already implemented elsewhere in Ultra code? It seems like it is not specific to ultra L2 processing.

imap_processing/spice/geometry.py
```diff
@@ -494,6 +494,7 @@ def basis_vectors(
 
 def cartesian_to_spherical(
     v: NDArray,
+    degrees: bool = True,
```
tmplummer, line 502: Be careful changing the default output like this. Is this going to break code that uses this function assuming the output is in radians? Also, the docstring indicates the output is in radians.

imap_processing/tests/spice/test_data/imap_frames_demo_euler.tf
tmplummer, line 1: Where did this come from? It appears to be a really old (Nov. 2022) version of the frame kernel. Is there something here that is not contained in what I can only assume is the updated version of this file: `imap_wkcp.tf`?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+    # TODO: reset the frame definition for the appropriate Pointing
+    # I'm not sure where to get the various spice kernels
+
+    rotation_matrix_helio_to_dps = spice.pxform(
+        fromstr="ECLIPJ2000", tostr="IMAP_DPS", et=event_time
+    )
+
+    # cartesian (x,y,z) radii vectors in the DPS frame corresponding to
+    # the grid points in the helio inertial frame
+    radii_dps = np.dot(rotation_matrix_helio_to_dps, radii_helio.T).T
```
tmplummer: This can be done with an existing function in the spice.geometry module:
```suggestion
    from imap_processing.spice.geometry import frame_transform, SpiceFrame
    radii_dps = frame_transform(event_time, radii_helio, SpiceFrame.ECLIPJ2000, SpiceFrame.IMAP_DPS)
```
With this change, it looks like this function will no longer need the `@ensure_spice` decorator.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+
+    radii_helio = geometry.spherical_to_cartesian(
+        np.stack(
+            (np.ones_like(az_grid_raveled), az_grid_raveled, el_grid_raveled),
+            axis=-1,
+        )
+    )
+    flat_indices_helio = np.arange(az_grid_raveled.size)
+
+    # TODO: reset the frame definition for the appropriate Pointing
```
tmplummer: I'm not sure what this TODO means. The Pointing kernel (`sim_1yr_imap_pointing_frame.bc`) defines the Pointing frame (`IMAP_DPS`) for a full year taking into account the repointing each day.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+        radii_dps, degrees=False
+    )
+
+    hae_az_in_dps_az = hae_grid_in_dps_spherical_coord[:, 1]
+    hae_el_in_dps_el = hae_grid_in_dps_spherical_coord[:, 2]
+
+    # Wrap the azimuth to the range (0, 2pi) rad = (0, 360) degrees
+    hae_az_in_dps_az = np.mod(hae_az_in_dps_az, 2 * np.pi)
+
+    # Wrap the elevation to the range (-pi/2, pi/2) rad = (-90, 90) degrees
```
tmplummer, line 156: Documentation from `cartesain_to_spherical` says that the output azimuth is already [0, 2*pi] and elevation is already [-pi/2, pi/2]. Did you find this not to be true?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+    hae_az_in_dps_az_indices = np.floor(
+        (hae_az_in_dps_az / (2 * np.pi)) * (360 / spacing_deg)
+    ).astype(int)
+    hae_el_in_dps_el_indices = np.floor(
+        # NOTE: equiv to: (np.pi/2 - hae_el_in_dps_el) * ((180 / np.pi) / spacing_deg)
+        (0.5 - hae_el_in_dps_el / np.pi) * (180 / spacing_deg)
+    ).astype(int)
```
tmplummer, line 163: This assumes azimuth bin edges are always at 0 and 2*pi and elevation are always -pi/2 and pi/2 but in the `build_az_el_grid` function there is an option to define the bins differently. See np.digitize and np.searchsorted for alternative solutions.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+def project_inertial_frame_to_dps(
+    event_time: float,
+    existing_grids: typing.Optional[tuple[NDArray, NDArray]] = None,
+    # Only used if no existing grids are provided
+    spacing_deg: float = DEFAULT_SPACING_DEG,
+) -> tuple[NDArray]:
+    """
+    Project an az/el grid in an inertial frame (ECLIPJ2000) to the DPS frame.
+
+    # TODO: Likely replace J2000 with HAE frame (everywhere it occurs)
```
tmplummer, line 80: I believe that HAE is equivalent to ECLIPJ2000.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+    elif "90sensor" not in descriptor_str:
+        logger.warning(
+            f"Found neither 45, nor 90 in descriptor string: {descriptor_str}"
+        )
+    return False
+
+
+@typing.no_type_check
+@ensure_spice
+def project_inertial_frame_to_dps(
```
tmplummer, line 71: I don't have the full algorithm to understand why it might be done this way, but I want to comment on the choice to rotate the map grid to the Pointing frame instead of vice-versa...
Rotating the grid from each PSET into the common map frame would make the binning more straight forward. For example, a 2D histogram can be used to bin the pset data into the lat/lon map grid. Additionally, when considering HEALPix, I think that will require the PSET data be transformed into the frame where the healpix are defined.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+        flat_indices_dps,
+        hae_az_in_dps_az,
+        hae_el_in_dps_el,
+        hae_az_in_dps_az_indices,
+        hae_el_in_dps_el_indices,
+    )
+
+
+@typing.no_type_check
+@ensure_spice
```
tmplummer, line 186: I don't see any spice calls in this function. No need for these decorators.

Edit: I found the spice call. It can soon be replaced with an internal `imap_processing` call. Keep in mind that in general, `@ensure_spice` should only wrap very small atomic functions that are close to the spiceypy call.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+        else:
+            combined_exptime_90 += pointing_projected_exptime
+
+    combined_exptime_total = combined_exptime_45 + combined_exptime_90
+
+    return (combined_exptime_45, combined_exptime_90, combined_exptime_total)
+
+
+@typing.no_type_check
+@ensure_spice
```
tmplummer, line 271: Same comment as above. Unless you are directly calling a spiceypy function, within this function, you do not need these decorators.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+    # This will contain the counts for each energy bin
+    combined_counts = np.zeros((num_el * num_az, num_energy))
+
+    # Build the azimuth and elevation grid in the heliocentric ecliptic frame (HAE)
+    (az_range, el_range, az_grid, el_grid) = l2_utils.build_az_el_grid(
+        spacing=spacing_deg,
+        input_degrees=True,
+        output_degrees=False,
+        centered_azimuth=False,  # (0, 2pi rad = 0, 360 deg)
+        centered_elevation=True,  # (-pi/2, pi/2 rad = -90, 90 deg)
```
tmplummer, line 337: I see that these grids are build both in this function as well as in the `build_dps_combined_exposure_time` function. Can this duplication be removed somehow?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+
+    energy_delta = np.diff(energy_bin_edges)
+    if not np.all(energy_delta >= 0):
+        raise ValueError("Energy bin edges must be monotonically increasing.")
+
+    for prod_num, l1c_prod in enumerate(l1c_products):
+        time = float(l1c_prod.epoch.values)
+        logger.info(
+            f"Projecting exposure time from product at time index {prod_num} "
+            f"with epoch {time} = {spice.et2utc(time, 'ISOC', 3)}"
```
tmplummer, line 384: Oh... here is the spice call. This should be replaced with a `imap_processing.spice.time.et_to_utc` function that should show up soon (Tori has a PR up with it).

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+        (
+            az_grid_raveled,
+            el_grid_raveled,
+            flat_indices_helio,
+            flat_indices_dps,
+            hae_az_in_dps_az,
+            hae_el_in_dps_el,
+            hae_az_in_dps_az_indices,
+            hae_el_in_dps_el_indices,
+        ) = project_inertial_frame_to_dps(
```
tmplummer, line 396: Likewise, it would be great to reduce the number of times this is called for each L1C product.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,483 @@
...
+        output_degrees=False,
+        centered_azimuth=False,  # (0, 2pi rad = 0, 360 deg)
+        centered_elevation=True,  # (-pi/2, pi/2 rad = -90, 90 deg)
+    )
+
+    for prod_num, l1c_prod in enumerate(l1c_products):
+        time = float(l1c_prod.epoch.values)
+        logger.info(
+            f"Projecting exposure time from product at time index {prod_num} "
+            f"with epoch {time} = {spice.et2utc(time, 'ISOC', 3)}"
```
tmplummer, line 237: See comment below...

imap_processing/tests/ultra/unit/test_l2_utils.py
```diff
@@ -0,0 +1,102 @@
...
+    assert el_range.size == int(180 / spacing)
+    assert az_range.size == az_grid.shape[1]
+    assert el_range.size == el_grid.shape[0]
+
+    # Check grid values
+    expected_az_range = np.arange((spacing / 2), 360 + (spacing / 2), spacing)
+    expected_el_range = np.arange(-90 + (spacing / 2), 90 + (spacing / 2), spacing)[
+        ::-1
+    ]  # Note el order is reversed
+    assert np.allclose(az_range, expected_az_range), (
```
tmplummer, line 67: `np.testing.assert_array_equal` (https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html) and `np.testing.assert_allclose` are preferred here as you don't need to add the f-strings. They will automatically print out discrepancies.

tmplummer (commented):
I have gone through an initial review. This PR is really big and should be broken up into smaller pieces in the future. I think that I have given feedback on the major pieces.

Nice job on a complex algorithm. We really need to try to understand the commonality between this and the Hi/Lo L2 to see if we can pull out some shared functions. Also, I think there will need to be some modifications/additions to the map/grid generation functions to allow for the various maps that the ENAs are interested in.

imap_processing/tests/spice/test_data/imap_frames_demo_euler.tf
```diff
@@ -0,0 +1,364 @@
+KPL/FK
```
laspsandoval: I should have given you a heads up on this. The kernels that are most recent are here:
[imap_wkcp.tf](https://lasp.colorado.edu/galaxy/download/attachments/173719425/imap_wkcp.tf?version=5&modificationDate=1714687588699&api=v2) - spacecraft frame kernel

So this kernel (imap_frames_demo_euler.tf) should be replaced by imap_wkcp.tf.

imap_processing/tests/spice/test_data/imap_ultra_test_metakernel.template
```diff
@@ -0,0 +1,9 @@
+{SPICE_TEST_DATA_PATH}/imap_sclk_0000.tsc
```
laspsandoval: You can delete this metakernel and use imap_ena_sim_metakernel.template instead.

imap_processing/tests/ultra/unit/test_l2_utils.py
nkerman: Ah! That's super good to know, thank you!

imap_processing/ultra/l2/ultra_l2.py
nkerman: I can't seem to find an existing instance of this. @laspsandoval?

This could live in ultra utils if that's preferable.

nkerman:
> I have gone through an initial review. This PR is really big and should be broken up into smaller pieces in the future. I think that I have given feedback on the major pieces.

Thanks @subagonsouth! Yes, I agree it's too big of a PR for this project's guidelines - I'm still getting used to them, and couldn't figure out how to write a useful skeleton for the L2 code without it getting too long. I will keep future versions much smaller. 

> Nice job on a complex algorithm. We really need to try to understand the commonality between this and the Hi/Lo L2 to see if we can pull out some shared functions. 

Thanks, and yes I definitely want to understand the commonalities with Lo/Hi. I'm somewhat tempted to begin by building most of the Ultra code, since I only really understand what it's doing once I implement it, which leads me to wanting to do a massive PR... a vicious cycle!

> Also, I think there will need to be some modifications/additions to the map/grid generation functions to allow for the various maps that the ENAs are interested in.

We'll need to make the output frame more modular for sure, and allow the output bins to be different from the input bins. The Ultra team is now talking about healpix tessellations, too. I'm hopeful unwrapping the az/el grids and working with them as 1D indices on a sphere will be a good start to making this possible.

imap_processing/spice/geometry.py
nkerman: Hmmm - I think the original docstring was incorrect then, since it had been returning:

```python
    spherical_coords = np.stack(
        (np.squeeze(magnitude_v), np.degrees(az), np.degrees(el)), axis=-1
    )

```

By adding the `degrees` param, which would default to True, I thought it would default to the previous behavior. Does that sound okay? I'll update the docstring for sure to reflect the unit behavior!

imap_processing/tests/spice/test_data/imap_frames_demo_euler.tf
nkerman: This is from the Ultra team's MATLAB code. When I didn't include it in my metakernel (but did include `imap_wkcp.tf`), SPICE couldn't construct the rotation matrix. IIRC, it couldn't understand the `-43901` frame.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Ah! That's helpful, thanks!

imap_processing/ultra/l2/ultra_l2.py
nkerman: It may not be necessary. It was happening for each pointing in the MATLAB, and I didn't fully understand it yet, but I think that the MATLAB used separate pointing kernels.

imap_processing/ultra/l2/ultra_l2.py
nkerman: no, so maybe we can drop this. I was really just trying to replicate everything the MATLAB did.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Oh, really? That's great 😁

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,485 @@
...
+        frame_epochs,
+        energy_delta,
+        energy_bin_edges,
+        energy_midpoints,
+        solid_angle_grid,
+        developer_dict,
+    )
+
+
+def ultra_l2(l1c_products: list) -> xr.Dataset:
```
laspsandoval, line 436: If you look at examples of other instruments you will see that they all follow the same format for using this function:

def ultra_l1c(data_dict: dict, data_version: str) -> list[xr.Dataset]:

laspsandoval (commented):
Nice work! I added a few comments, but I think this is too much to review in a single go. Could you break this down into at least 3 PRs? It will help me to understand each piece of what you have written.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Good point! I can pass them into `build_dps_combined_exposure_time` from `build_flux_maps`. 

These functions have to replicate a lot of the same frame transforms, so I might end up combining that logic anyways into some future `match_helio_and_dps_indices()` function.

imap_processing/ultra/l2/ultra_l2.py
nkerman: I can drop this log statement entirely. Just found it helpful while developing.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Yes, I think I should probably eventually (maybe in this PR) refactor this to have a single call to `match_helio_and_dps_indices` per pointing, which is used by all the functions which need to project into the helio frame.

imap_processing/ultra/l2/ultra_l2.py
nkerman: I was also fairly confused by this, but I had basically figured they were equivalent, and that I should implement what the Ultra team had in their MATLAB code. That might not be the right mindset/place to go to determine ground truth on how to write these algorithms...

imap_processing/tests/spice/test_data/imap_frames_demo_euler.tf
nkerman: Ah, got it! Is this new version okay to replace the existing imap_wkcp.tf present in the spice test_data directory? I wouldn't want to mess up any other instruments' tests, if that's a risk.

nkerman: Ah, got it! Is this new version okay to replace the existing imap_wkcp.tf present in the spice test_data directory? I wouldn't want to mess up any other instruments' tests, if that's a risk.

imap_processing/ultra/l2/ultra_l2.py
nkerman: I'm not convinced that the same approach will work for L2. It looks like generally, the CLI passes in a `data_dict[Logical_source] = dataset`. But when passing multiple L1c datasets, won't their Logical_source values be non-unique? I could have it pass them in as `f"{Logical_source}_{i}"` to avoid that, but I wasn't sure what using a dict adds over a list here. 

I'll switch to a dict for consistency regardless.

imap_processing/spice/geometry.py
tmplummer: Yep, my bad. You are absolutely correct. Thanks for updating the docstring.

imap_processing/tests/spice/test_data/imap_frames_demo_euler.tf
tmplummer: I am not aware of a new version of the `imap_wkcp.tf` frame kernel. It looks like what is linked is the same as what is already checked in.

imap_processing/ultra/l2/ultra_l2.py
tmplummer: For sure ask Ultra. I think algorithmically, it makes more sense to do the transform the other way as long as there are no gotchas that I don't know about.

tmplummer (commented):
I am hope to have a rough draft of the L2/mapping algorithm design to share in the ENA Tag-up today. Perhaps that will help drive a direction to resolve some of the things in this PR.

imap_processing/spice/geometry.py
nkerman: No prob! Easy mistake to make.

It might happen in another PR, as I might need to break this one up.

imap_processing/tests/spice/test_data/imap_frames_demo_euler.tf
nkerman: I did a diff on that linked file vs the one in the repo and it's got a lot of new content.

imap_processing/ultra/l2/ultra_l2.py
nkerman: @subagonsouth - I don't actually know if I can use `frame_transform`, since it will only apply to a single vector for each et. I could pass it a whole list of all the same et, but then I think it would re-calculate the same rotation matrix n-times for each of the $360 * 180 / (spacing^2)$ vectors. 

How about I just create the rotation matrix using our geometry.get_rotation_matrix function, and apply the rotation matrix in the same way?

nkerman: Another option is altering `frame_transform` to allow multiple vectors for the same et. Given that the function explicitly forbids that, I figured there was a reason not to do so?

tmplummer: I think that altering `frame_transform` is a good idea. I think that it was my failure to be creative enough to think of that use case that caused that situation to be disallowed. I *think* that simply removing the `if` block that raises the ValueError will allow it to work correctly.

nkerman: Screenshot 2024-12-05 at 3 14 51 PM


I'm having it still raise the error, unless the len(et) == 1, because I think that's the only case where we can safely do this (e.g. if there are m `position`s and n `et`s, that's no good). I think that would fail within numpy broadcasting anyways, but no 100%.

imap_processing/ultra/l2/ultra_l2.py
nkerman: I could also just do away with that functionality in `build_az_el_grid`, as it's not used anywhere.

imap_processing/ultra/l2/ultra_l2.py
nkerman: We talked about this in our meeting, and I spoke to Nick at length about this. I think I've convinced Nick we should do it the way you suggested for the reasons you mentioned, and a few other challenges which would arise with the original method. 

I want to make all of these changes now, but I think I should break up the PR first into smaller pieces.

imap_processing/tests/spice/test_data/imap_ultra_test_metakernel.template
nkerman: Great! I got this working with the new version of imap_wkcp.tf - I had been confused because it doesn't cover the same time ranges as the old euler file.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Removed it, after convincing myself that cartesain_to_spherical does ensure this is met.

imap_processing/ultra/l2/ultra_l2.py
nkerman: In new branch/PR: 
Removed all direct use of spice (now uses spice.geometry.frame_transform). Resolving this and related threads.

nkerman:
NOTE: Keeping this PR open for a bit longer to access the last remaining threads on it.

PR #1206, opened 2024-12-06
Bring over l2_utils from large PR

### LINKED ISSUE #1207: Create rectangular (az/el) grids for Ultra L2 and calculate their solid angle.
Make functions to build az, el grids covering the sky for use in Ultra L2 Map projection. 

Make a function to calculate the solid angle covered by each pixel on this grid.

nkerman:
# Change Summary

Closes #1207

## Overview
3 Utility functions which will be needed for Ultra L2 Processing:
1. build_solid_angle_map to determine the solid angle for evenly angular spaced rectangular grid of the sky
2. build_az_el_grid to build evenly angular spaced rectangular grid of the sky
3. rewrap_even_spaced_el_az_grid to take an unwrapped (raveled) rectangular grid of the sky in 1D and turn it back into a 2D grid. 

## New Files

- imap_processing/ultra/l2/l2_utils.py
  - Utility functions for use in Ultra L2 processing. 
- imap_processing/tests/ultra/unit/test_l2_utils.py
  - Tests for above functions

## Testing

Adds unit tests for all new functions.


imap_processing/ultra/l2/l2_utils.py
```diff
@@ -0,0 +1,194 @@
...
+        solid_angle_lats[:, np.newaxis], (2 * np.pi) / spacing, axis=1
+    )
+
+    if output_degrees:
+        solid_angle_grid = solid_angle_grid * ((180 / np.pi) ** 2)
+
+    return solid_angle_grid
+
+
+def build_az_el_grid(
```
laspsandoval: This does pretty much exactly what ultra_l1c_pset_bins.py build_spatial_bins does. I am wondering if we could combine the two and put them under utils so we don't have to maintain basically the same function?

imap_processing/ultra/l2/l2_utils.py
```diff
@@ -0,0 +1,194 @@
+"""Build a solid angle map for a given spacing in degrees."""
+
+import typing
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+def build_solid_angle_map(
+    spacing: float, input_degrees: bool = False, output_degrees: bool = False
```
laspsandoval: A little nitpicky, but in the other code in geometry.py you had 

def cartesian_to_spherical(
    v: NDArray,
    **degrees: bool = True,**
    
 If possible be consistent.

imap_processing/ultra/l2/l2_utils.py
```diff
@@ -0,0 +1,194 @@
...
+            - np.sin(np.arange(start=(-np.pi / 2), stop=(np.pi / 2), step=spacing))
+        )
+    )
+
+    solid_angle_grid = np.repeat(
+        solid_angle_lats[:, np.newaxis], (2 * np.pi) / spacing, axis=1
+    )
+
+    if output_degrees:
+        solid_angle_grid = solid_angle_grid * ((180 / np.pi) ** 2)
```
laspsandoval: `solid_angle_grid *= (180 / np.pi) ** 2`

imap_processing/ultra/l2/l2_utils.py
```diff
@@ -0,0 +1,194 @@
...
+    if input_degrees:
+        spacing = np.deg2rad(spacing)
+
+    if spacing <= 0:
+        raise ValueError("Spacing must be positive valued, non-zero.")
+
+    if not np.isclose((np.pi / spacing) % 1, 0):
+        raise ValueError("Spacing must divide evenly into pi radians.")
+
+    solid_angle_lats = np.abs(
```
laspsandoval: For readability consider changing to something like:
`latitudes = np.arange(-np.pi / 2, np.pi / 2 + spacing, step=spacing)
solid_angle_lats = np.abs(
    spacing * (np.sin(latitudes[1:]) - np.sin(latitudes[:-1]))
)`

imap_processing/ultra/l2/l2_utils.py
```diff
@@ -0,0 +1,194 @@
...
+        az_range = np.rad2deg(az_range)
+        el_range = np.rad2deg(el_range)
+        az_grid = np.rad2deg(az_grid)
+        el_grid = np.rad2deg(el_grid)
+
+    return az_range, el_range, az_grid, el_grid
+
+
+@typing.no_type_check
+def rewrap_even_spaced_el_az_grid(
```
laspsandoval: I like this. Nice!

laspsandoval (commented):
Looks good to me. The biggest ask would be to move the build_az_el_grid to be a common util.

imap_processing/ultra/l2/l2_utils.py
nkerman: They do slightly different things - the elevation needs to be reversed in direction to follow the matlab, and we also tile the entire sphere with the grid, rather than just having 2 axis vectors. 

But I could change the l2 code so that it does the following rather than making a separate grid:

```Python
az_midpoints_deg, el_midpoints_deg = build_spatial_bins(
    az_spacing=spacing_deg, 
    el_spacing=spacing_deg
)[2:]
az_midpoints, el_midpoints = np.deg2rad(az_midpoints_deg), np.deg2rad(el_midpoints_deg)
az_grid = np.tile(az_midpoints, len(el_midpoints)).reshape(len(el_midpoints), len(az_midpoints))
el_grid = np.tile(el_midpoints, len(az_midpoints)).reshape(len(az_midpoints), len(el_midpoints)).T[::-1]
```

I'd probably lean away from that and keep them separate, but happy to make the change!

imap_processing/ultra/l2/l2_utils.py
nkerman: Good point! I prefer using radians, but had wanted to keep the default for the other function consistent with the existing code in degrees. I'll make this default to degrees as well and just override it

imap_processing/ultra/l2/l2_utils.py
nkerman: Thanks!

imap_processing/ultra/l2/l2_utils.py
nkerman: Good point! I broke it up into the constituent functions and used np.diff instead of the subtraction.

nkerman:
> Looks good to me. The biggest ask would be to move the build_az_el_grid to be a common util.

Thanks @laspsandoval! I made all the smaller changes - I think the amount of overlap with `build_spatial_bins` functionality is smaller than the differences between their outputs -  all that they both do is divide 2pi or 360 by a given spacing. 

I can definitely call `build_spatial_bins` at the start of `build_solid_angle_map`, but then I need to convert back and forth from degrees.

In terms of making these shared utils, how about I move this file to be in `ultra/utils/spatial_utils.py`, and bring `build_spatial_bins` there as well?

I did this in commit `fdfb2e9`, but we can roll that back if that's not the type of extraction you had in mind!

imap_processing/ultra/utils/spatial_utils.py
```diff
@@ -0,0 +1,221 @@
+"""Build a solid angle map for a given spacing in degrees."""
```
tmplummer: This docstring looks like it is specific just to the `build_solid_angle_map` function. Can you generalize it to this module?

tmplummer (commented):
I'm good with this as is. I don't have strong opinions about the possibility for extracting some common code between the three `build` functions but I'm not approving since that is an open discussion.

imap_processing/ultra/utils/spatial_utils.py
nkerman: Will do!

imap_processing/ultra/l2/l2_utils.py
laspsandoval: I agree with you here. Let's keep them separate.

laspsandoval (approved):

Issue #1208, opened 2024-12-06
SPICE - Error in computing filename date string in write_cdf


tmplummer:
### Algorithm Description:

While working on SPICE time functions, I discovered that the implementation in `imap_processing/cdf/utils.py::write_cdf` for generating the date string from the zeroth element of the Dataset epoch coordinate is incorrect by 5 seconds due to the occurrences of 5 leap-seconds since the J2000 epoch. This ticket it to evaluate and figure out a solution. Ideas:
- Hard code 5 second offset into calculation. The advantage of doing it this was is that writing the CDF will not require any SPICE kernels be furnished. The disadvantage is that it will be wrong if/when another leap-second occurs.
- Use SPICE. The advantage is it will always be correct assuming the correct leap-second kernel is furnished.

vineetbansal:
I think it's reasonable to package `naif0012.tls` in the `spice` submodule of `imap_processing` so we can use `spiceypy` for this stuff.  This only applies to the `leapseconds` kernel, not in general. It's updated rarely, with the last update in 2017, so it's not too onerous to include future latest leapseconds kernels. Also, my understanding is that the whole concept of leapseconds is being phased out anyway, so this will not turn into an ongoing maintenance headache.

@jtniehof - I'm wondering if you have thoughts on this?

tmplummer:
> It's updated rarely, with the last update in 2017, so it's not too onerous to include future latest leapseconds kernels.

If a leapsecond were to happen in two years who is going to remember to come add the new tls file to the `imap_processing` package and update those hard-coded references to the `naif0012.tls` kernel? I won't.

One other option that I didn't include in the issue description would be to use Astropy. For a long time, we didn't want to add Astropy as a dependency due to it's bloat, but it is a required dependency for `astropy-healpix` so maybe we should use that here instead. I don't know details of how Astropy handles leapseconds, but at least we won't need to remember to make a change.

vineetbansal:
Once dependencies are pinned (as they should be - see my open [issue](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/1252) on this), then using `astropy` for leap seconds doesn't actually eliminate the "remember" problem - it just moves it to having to remember to update `astropy`, whenever it incorporates that knowledge.

Also a leap second update in `astropy` won't come in isolation - it will be bundled with whatever else changed in that `astropy` release, opening up a new can of worms.

tmplummer:
[...]

PR #1210, opened 2024-12-06
Add functions to convert from met to ET and UTC

### LINKED ISSUE #1183: SPICE - met to utc
### Algorithm Description:
Write a function that converts from floating point MET to UTC timestamp.
Also useful would be converting MET to datetime

tmplummer:
# Change Summary

## Overview
Adds functions for converting from Mission Elapsed Time (MET) to ET and UTC

## Updated Files
   - Definition of `J2000_EPOCH` moved into the spice.time.py module
   - TODO added to address discovered error in converting epoch to date string for file naming. Ticket created: #1208
   - Definition of `J2000_EPOCH` moved into the spice.time.py module
   - Extracted `met_to_sclkticks` function for reuse
   - Added `met_to_utc` function
   - Added `met_to_datetime64` function with SPICE backed implementation.
   - Add test coverage for 3 new `spice/time.py` functions.
   - Don't import `spiceypy as spice` because it is confusing.

Closes: #1183


imap_processing/spice/time.py
```diff
@@ -13,6 +13,30 @@
 
 TICK_DURATION = 2e-5  # 20 microseconds as defined in imap_sclk_0000.tsc
 
+# Hard code the J2000 epoch. This allows for CDF epoch to be converted without
+# use of SPICE though it should be noted that this results in a 5-second error
+# due to the occurrence of 5 leap-seconds since the J2000 epoch.
```
nkerman, line 18: Arguably a **TODO** should go here, since this is where the epoch is defined, and where the 5 leapsecond fix will go, right?

nkerman (approved):
LGTM! Nice, thorough tests. 

Will the fix to the 5 second just be replacing this constant `J2000_EPOCH` with a spice call?

imap_processing/spice/time.py
```diff
@@ -13,6 +13,30 @@
...
+# The UTC string was generated by:
+# >>> spiceypy.et2utc(0, "ISOC", 9)
```
tech3371, line 22: Thanks for these doc strings. Helps someone like me.

imap_processing/spice/time.py
```diff
@@ -64,6 +88,51 @@ def j2000ns_to_j2000s(j2000ns: npt.ArrayLike) -> npt.NDArray[float]:
...
+    met : float, numpy.ndarray
+        Number of seconds since epoch according to the spacecraft clock.
+    precision : int
+        The number of digits of precision to which fractional seconds
+        are to be computed.
+
+    Returns
+    -------
+    numpy.ndarray[str]
+        The mission elapsed time converted to UTC string.
```
tech3371: Can you add an example about the format it returns in? If I understand correctly, `ISOC` format is something like `YYYY-MM-DDTHH:MM:SSZ`. Is that right? or is it like this `1987-04-12T16:31:12.814`?

imap_processing/spice/time.py
```diff
@@ -64,6 +88,51 @@ def j2000ns_to_j2000s(j2000ns: npt.ArrayLike) -> npt.NDArray[float]:
...
+        Number of seconds since epoch according to the spacecraft clock.
+
+    Returns
+    -------
+    numpy.ndarray[str]
+        The mission elapsed time converted to UTC string.
+    """
+    if isinstance(met, typing.Iterable):
+        return np.asarray([np.datetime64(utc) for utc in met_to_utc(met)])
+    return np.datetime64(met_to_utc(met))
```
tech3371, line 137: These two functions are gonna be so helpful! Thank you for adding those

imap_processing/tests/spice/test_time.py
```diff
@@ -42,6 +59,61 @@ def test_j2000ns_to_j2000s(furnish_time_kernels):
...
+        sclk_ticks = np.array([spiceypy.sce2c(IMAP_SC_ID, et) for et in et_arr])
+    else:
+        expected_dt64 = np.asarray(np.datetime64(utc))
+        et = spiceypy.str2et(utc)
+        sclk_ticks = spiceypy.sce2c(IMAP_SC_ID, et)
+    met = sclk_ticks * TICK_DURATION
+    dt64 = met_to_datetime64(met)
+    np.testing.assert_array_equal(
+        dt64.astype("datetime64[us]"), expected_dt64.astype("datetime64[us]")
+    )
```
tech3371, line 114: nice!

tech3371 (approved):
LGTM! I just had one minor doc string requests.

imap_processing/spice/time.py
tech3371: True Nat. I meant to ask Tim if there is a ticket to revisit for this?

tmplummer: I will add the TODO. Yes, there is a ticket for this. There are two ways to fix this each with a pro and con. See ticket: #1208

To clarify, this J2000EPOCH value is correct. The bug is in the code in the `cdf/utils.py` here:
```
  dt64 = J2000_EPOCH + dataset["epoch"].values[0].astype("timedelta64[ns]")
  start_time = np.datetime_as_string(dt64, unit="D").replace("-", "")
```

tmplummer:
> Will the fix to the 5 second just be replacing this constant `J2000_EPOCH` with a spice call?

No, actually, it is a bit more subtle. That epoch definition is correct in that it is the number of nanoseconds elapsed since the Unix epoch (00:00:00 UTC on 1 January 1970). The problem is that `np.datetime64` does not take leap-seconds into account so when an offset is added to this epoch and then converted to a string, if the offset passed any leap-seconds, the string will be incorrect by the number of leap-seconds passed by adding the offset.

PR #1212, opened 2024-12-06
Lo - Event Parsing Segmented Packets with Validation

### LINKED ISSUE #1211: Lo - Update event parsing to handle counts and passes in binary chunk
Lo event parsing needs to be updated to handle parsing the counts and passes from the binary chunk for segmented packets

sdhoyt:
# Change Summary

## Overview
Updates to handled Direct Event segmented packets which include the counts and passes field being part of the binary chunk rather than being directly decommutated in space packet parser. Validation and sample data were also added to validate the direct event parsing.

## Updated Files
- imap_processing/lo/l0/decompression_tables/decompression_tables.py
   - Added a packet field tuple to help with parsing the segmented packets
   - Added processing to handle counts and passes being parsed from binary chunk
   - Added logging

## Testing
Updated unit tests and added validation data test

Closes #1211


tmplummer (approved):
LGTM. Nice work!

vmartinez-cu (approved):
Nice work!

imap_processing/lo/l0/lo_science.py
```diff
@@ -1,5 +1,6 @@
 """Processing function for Lo Science Data."""
```
nkerman, line 1:
```suggestion
"""L0 processing functionality for Lo Science Data."""
```

nkerman (approved):
Looks great! My only comment (not particular to these changes) is that the name `lo_science.py` seems broader than the actual content of the file - which I think is only in charge of parsing and decomming the L0 packets. The name confused me a bit. It's less of an issue for that file, which is nested in an `l0` dir than for it's tests, which are just `test_lo_science.py`. 

Would `lo_l0.py` or something similar work?

imap_processing/lo/l0/lo_science.py
sdhoyt: Agreed that it's a little confusing. My plan was to put all the Lo science packet processing in lo_science.py, but it might make sense to rename this. I'll think about this more and try to come up with something less confusing.

PR #1223, opened 2024-12-13
SWE: Review and correct phase space density units

### LINKED ISSUE #1164: SWE: Fix phase space density units
Notes from Ruth:

> For the equation: *   fv(v,theta,phi) = 2*C(E,theta,phi)/(G*v^4*tau)
>  
> C should have units of Counts
> Tau should have units of seconds
> G has units of cm^-2 ster^-1
> So to make the units come out right, we need v in units of cm/s.
> Since we have E in units of eV, we first need to convert that to Joules, so
>       v = sqrt(2 E / m) = sqrt(2.0 * E(eV)*1.60219e-19(J/eV) / 9.10938e-31 kg)
> This equation gives v in units of (meters/second), so we now need to convert from meters to centimeters by multiplying by 100.
>  
> You should of course check my math, but from all that I get that v^4, in units of (cm/s)^4 is calculated by:
>    v^4 = 1.237e31 * Energy(eV) * Energy(eV)

tech3371:
# Change Summary
closes #1164 

## Overview
Reviewed and correct SWE phase space density units and conversion factors.

## Updated Files
   - update doc strings with notes on electron speed units ocnversion and how conversion factors was calculated.
- pyproject.toml
   - Added ignore for ruff

## Testing
    - updated tests

### DIFF
--- a/imap_processing/swe/l2/swe_l2.py
+++ b/imap_processing/swe/l2/swe_l2.py
@@ -22,7 +22,8 @@ GEOMETRIC_FACTORS = np.array(
 )
 ELECTRON_MASS = 9.10938356e-31  # kg
 
-VELOCITY_CONVERSION_FACTOR = np.sqrt(3.20438e-15 / ELECTRON_MASS)
+# See doc string of calculate_phase_space_density() for more details.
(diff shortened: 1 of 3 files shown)


tmplummer (approved):
LGTM!

PR #1224, opened 2024-12-16
IDEX L1b science data processing 

### LINKED ISSUE #833: Produce IDEX L1B CDFs
### Tasks

- [x] l1b code
- [x] l1b tests 
- [x] #1282

lacoak21:
# Change Summary

IDEX l1b science data processing and tests 

## Overview
Adds processing to unpack telemetry items using imap_processing/idex/idex_variable_unpacking_and_eu_conversion.csv, and converts all instrument setting variables and waveforms to engineering units. 

Addition of spice data and two more attributes will be added in another PR. 

closes #833
## New Files
- imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
   - L1b variable attrs  
   - Unpack instrument settings 
   - Convert instrument settings and waveforms to engineering units 
- imap_processing/idex/idex_variable_unpacking_and_eu_conversion.csv
   - Csv containing unpacking information and conversions 
   - Tests for l1b code
  
## Updated Files 

- imap_processing/cdf/config/imap_idex_global_cdf_attrs.yaml
   - Add l1b data level 
   - Add logger statement 
   - Fix dimension for science variable 
- imap_processing/cli.py
  - Fixed L1a cli handling 
  - Added L1b handling 

## Testing
- 5 l1b unit tests added


imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,330 @@
...
+  UNITS: pC
+  VAR_TYPE: data
+  CATDESC: ""
+  FIELDNAM: ""
+  FILLVAL: *int_fillval
+  LABLAXIS: ""
+
+l1b_tof_base: &l1b_tof_base
+  <<: *l1b_data_base
+  DEPEND_1: time_high_sr
```
tech3371, line 34: If you have a `DEPEND_1`, then you will need `LABL_PTR_1` instead of LABLAXIS. I tried to capture something about this here, https://imap-processing.readthedocs.io/en/latest/external-tools/cdf/cdf_requirements.html#metadata-data. But let me know if you like to chat about this.

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,330 @@
...
+  LABLAXIS: ""
+  FORMAT: I10
+  UNITS: " "
+
+
+tof_high:
+  <<: *l1b_tof_base
+  CATDESC: Time of flight waveform on the high-gain channel
+  FIELDNAM: High Gain Time of Flight
+  LABLAXIS: TOF High Ampl.
```
tech3371: remove period from all labelaxis?
```suggestion
  LABLAXIS: TOF High Ampl
```

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,330 @@
...
+  CATDESC: Last measurement in raw DN for LVPS Board signal - "P2.5V_IMON"
+  LABLAXIS: Current
+  UNITS: A
+
+current_neg2p5v_bus:
+  <<: *trigger_base
+  FIELDNAM: N2.5V_IMON
+  CATDESC: Last measurement in raw DN for LVPS Board signal - "N2.5V_IMON"
+  LABLAXIS: Current
+  UNITS: A
```
tech3371, line 326: In general, really good organization here!

imap_processing/idex/idex_variable_unpacking_and_eu_conversion.csv
tech3371, line 1: btw, I like this name better!

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+    idex_attrs : ImapCdfAttributes
+        CDF attribute manager object.
+
+    Returns
+    -------
+    telemetry_das : dict
+        A dictionary where the keys are the instrument setting array names and the
+        values are the unpacked xr.DataArrays.
+    """
+    telemetry_das = {}
```
tech3371: what is `_das`? should we pick a different name?

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # Create binary mask of the size of the variable in bits
+        mask = (1 << row["unsigned_nbits"]) - 1
+        # Determine the number of bits to shift
+        shift = row["starting_bit"] - row["nbits_padding_before"]
+        # Get the unpacked value by shifting the data to align the desired bits with
+        # the least significant bits and applying the mask to isolate the target bits
+        unpacked_val = (l1a_dataset[row["packetName"]].data >> shift) & mask
+
+        attrs = idex_attrs.get_variable_attributes(var_name) if idex_attrs else None
```
tech3371: Does this mean that if it's not set here, it's set somewhere else? can you add a note if that's the case?

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+    for _, row in var_information_df.iterrows():
+        var_name = row["mnemonic"]
+
+        # Create binary mask of the size of the variable in bits
+        mask = (1 << row["unsigned_nbits"]) - 1
+        # Determine the number of bits to shift
+        shift = row["starting_bit"] - row["nbits_padding_before"]
+        # Get the unpacked value by shifting the data to align the desired bits with
+        # the least significant bits and applying the mask to isolate the target bits
+        unpacked_val = (l1a_dataset[row["packetName"]].data >> shift) & mask
```
tech3371: I understand on high level but really helpful comments!

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+    -------
+    waveforms_converted : dict
+        A dictionary where the keys are the waveform array names and the values are
+        xr.DataArrays representing the waveforms transformed into picocoulombs.
+    """
+    waveforms_pc = {}
+
+    for var in ConvFactors:
+        waveforms_pc[var.name] = l1a_dataset[var.name] * var.value
+        waveforms_pc[var.name].attrs = idex_attrs.get_variable_attributes(
```
tech3371, line 188: yeah, looks like you set attrs here as well like I thought. Do you set some in above function and some in this function? Just curious

imap_processing/tests/idex/test_idex_l1b.py
```diff
@@ -0,0 +1,117 @@
...
+    -------
+    dataset : xr.Dataset
+        A ``xarray`` dataset containing the test data
+    """
+    dataset = idex_l1b(decom_test_data, data_version="001")
+    return dataset
+
+
+def test_l1b_cdf_filenames(l1b_dataset: xr.Dataset):
+    """Tests that the ``process_idex_l1b`` function generates datasets
```
tech3371: should this be this?
```suggestion
    """Tests that the ``idex_l1b`` function generates datasets
```

imap_processing/tests/idex/test_idex_l1b.py
```diff
@@ -0,0 +1,117 @@
...
+            "starting_bit": [0, 2],
+            "nbits_padding_before": [0, 0],
+            "unsigned_nbits": [2, 3],
+        }
+    )
+
+    unpacked_dict = unpack_instrument_settings(test_ds, test_cdf_defs_df)
+
+    assert np.all(unpacked_dict["test_var0"] == 2)
+    assert np.all(unpacked_dict["test_var1"] == 4)
```
tech3371, line 126: This test was very helpful to understand and check what you were doing in `unpack_instrument_settings()`. Nice work!

tech3371 (commented):
I had some non-blocking comments. Once you address those, it looks good to me!

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,330 @@
+int_fillval: &int_fillval -9223372036854775808
+
+max_uint16: &max_uint12 4095
```
tmplummer: I think that these names are intended to match?
```suggestion
max_uint12: &max_uint12 4095
```

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,330 @@
...
+
+data_min: &data_min 0
+data_max: &data_max 4096
```
tmplummer: 4096 is an uncommon number to assign to VALIDMAX. Is there a reason it isn't 4095?

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,330 @@
...
+  VALIDMIN: *data_min
+  VALIDMAX: *data_max
+  DISPLAY_TYPE: spectrogram
+  DEPEND_0: epoch
+  FORMAT: I12
+  UNITS: pC
+  VAR_TYPE: data
+  CATDESC: ""
+  FIELDNAM: ""
+  FILLVAL: *int_fillval
```
tmplummer: The defined `int_fillval` is only valid for 64 bit signed integers. Based on the `VALIDMIN` and `VALIDMAX`, it seems like a mismatch.

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -0,0 +1,330 @@
...
+  FIELDNAM: Ion Grid Signal
+  LABLAXIS: Ion Grid Ampl.
+  VAR_NOTES: This is the ion grid signal from IDEX.
+        Sampled at 3.75 Msps with 12-bit resolution.
+        Data is used to quantify dust charge.
+
+detector_voltage:
+  <<: *trigger_base
+  FIELDNAM: Detector Voltage
+  CATDESC: Last measurement in raw dN for processor board signal - "Detector Voltage"
```
tmplummer: All of these CATDESCs say "measurement in raw dN" but the UNITS say Voltage.

imap_processing/tests/idex/test_idex_l1b.py
```diff
@@ -0,0 +1,117 @@
...
+import pandas as pd
+import pytest
+import xarray as xr
+
+from imap_processing import imap_module_directory
+from imap_processing.cdf.utils import write_cdf
+from imap_processing.idex.idex_l1b import idex_l1b, unpack_instrument_settings
+
+
+@pytest.fixture()
```
tmplummer: I suggest changing the scope of this fixture to "module". That way the fixture is only executed once and the return value gets cached so that `idex_l1b` doesn't have to run for each test that calls this fixture.
```suggestion
@pytest.fixture(scope="module")
```
In case you want to know more about fixture scopes: https://docs.pytest.org/en/stable/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # TODO: Add TriggerMode and TriggerLevel attr
+    # TODO: Spice data?
+
+    logger.info("IDEX L1B science data processing completed.")
+
+    return l1b_dataset
+
+
+def unpack_instrument_settings(
```
tmplummer, line 117: Is there a reason that the fields that get unpacked are not defined in the xtce and unpacked by `space_packet_parser` rather than as done here at L1B?

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+    # Create l1b Dataset
+    l1b_dataset = xr.Dataset(
+        data_vars=processed_vars | waveforms_converted,
+        attrs=idex_attrs.get_global_attributes("imap_idex_l1b_sci"),
+    )
+    # Convert variables
+    l1b_dataset = convert_raw_to_eu(
+        l1b_dataset,
+        conversion_table_path=var_information_path,
+        packet_name=var_information_df["packetName"].to_list(),
```
tmplummer: Interesting... The docstring for `convert_raw_to_eu` says `packet_name` is a `str`, not a `list`. Is this working correctly? The function reduces the dataframe read in from the conversion table to only the entries where `dataframe["packetName"] == packet_name` which I would expect to be an array full of `False`, meaning no conversions get applied.

imap_processing/tests/idex/test_idex_l1b.py
```diff
@@ -0,0 +1,117 @@
...
+    Example
+    -------
+    In this example, we are using a test variable that has five bits
+    Idx__test_var01 = 0b10010
+
+    Int(0b10010) = 18
+
+    This should unpack into test_var0, and test_var1
+     - Test_var0 is two bits long and starts at 0, and the unpacked value should be 2
+     - Test_var1 is three bits long and starts at 3, and the unpacked value should be 4
```
bourque:
```suggestion
     - test_var0 is two bits long and starts at 0, and the unpacked value should be 2
     - test_var1 is three bits long and starts at 3, and the unpacked value should be 4
```

bourque: Thanks for providing this example, very helpful!

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
+"""IMAP-IDEX L1B processing module."""
```
bourque:
```suggestion
"""IDEX L1B processing module."""
```

It would also be great if you could add an example here of how the module is run.

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+import xarray as xr
+
+from imap_processing import imap_module_directory
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
+from imap_processing.utils import convert_raw_to_eu
+
+logger = logging.getLogger(__name__)
+
+
+class ConvFactors(float, Enum):
```
bourque:
```suggestion
class ConversionFactors(float, Enum):
```

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+        l1a_dataset["epoch"],
+        name="epoch",
+        dims=["epoch"],
+        attrs=idex_attrs.get_variable_attributes("epoch"),
+    )
+    # Create l1b Dataset
+    l1b_dataset = xr.Dataset(
+        data_vars=processed_vars | waveforms_converted,
+        attrs=idex_attrs.get_global_attributes("imap_idex_l1b_sci"),
+    )
```
tmplummer: I suggest including the `epoch` coordinate here rather than adding it later.
```suggestion
    l1b_dataset = xr.Dataset(
        coords={"epoch": epoch_da},
        data_vars=processed_vars | waveforms_converted,
        attrs=idex_attrs.get_global_attributes("imap_idex_l1b_sci"),
    )
```

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+    Returns
+    -------
+    waveforms_converted : dict
+        A dictionary where the keys are the waveform array names and the values are
+        xr.DataArrays representing the waveforms transformed into picocoulombs.
+    """
+    waveforms_pc = {}
+
+    for var in ConvFactors:
+        waveforms_pc[var.name] = l1a_dataset[var.name] * var.value
```
tmplummer, line 187: I *think* that numpy will convert the dtype to a `float` in this statement. Is that the desired dtype for the output CDF?

imap_processing/idex/idex_l1b.py
bourque: Agreed, it is a bit confusion. Perhaps use `telemetry_data`?

imap_processing/idex/idex_l1b.py
bourque: Agreed, your inline comment game is strong 😂

imap_processing/idex/idex_l1b.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+
+    return telemetry_das
+
+
+def convert_waveforms(
+    l1a_dataset: xr.Dataset, idex_attrs: ImapCdfAttributes
+) -> dict[str, xr.DataArray]:
+    """
+    Apply transformation from raw dn to picocoulombs (pC) for each of the six waveforms.
```
bourque:
```suggestion
    Apply transformation from raw DN to picocoulombs (pC) for each of the six waveforms.
```

tmplummer (commented):
Great work. Nice clean, well organized code. I left a few comments to consider.

bourque (approved):
Awesome work, you are making great progress! I have a few minor suggestions if you want to use them, otherwise this looks good.

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: Ah got it! Thanks for linking those docs. I think I need to do this for l1a as well and then can copy them to l1b.

tech3371: sounds good. If this is not blocking this PR, then you can do that in future PR if you like to not make this PR bigger.

imap_processing/idex/idex_l1b.py
lacoak21: Much better :)

imap_processing/idex/idex_l1b.py
lacoak21: I allowed for attrs to be none for testing purposes. I might remove this as optional and just add some example attrs in the test.

imap_processing/idex/idex_l1b.py
lacoak21: See related thread

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: Good catch

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: Oh yes those descriptions should be updated thanks

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: Nope! That should be 4095 - probably a typo

imap_processing/idex/idex_l1b.py
lacoak21: I was also confused about that. I talked to the IDEX team and they wanted l1a to have all the raw instrument settings so people could go back to l1a and look at them if need be.

imap_processing/idex/idex_l1b.py
lacoak21: Hm it is in fact working as expected. It is converting every value but I am definitely using "packetName" incorrectly. I will have to fix this.

imap_processing/idex/idex_l1b.py
lacoak21: Yes the output should now be a float64 (matching the l1b validation file)

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: Ok I went back to the docs and updated it to use 65535. I think that should be valid

tmplummer (approved):
LGTM, great work!

PR #1228, opened 2024-12-18
Ultra l1b: velocity magnitude and updated species functions

laspsandoval:
# Change Summary

## Overview
Added magnitude and updated species functions.

## Updated Files
- ultra_l1b_extended.py
   - Added velocity magnitude and species functions
- constants.py
   - Constants based on ctof ranges to determine species
   - Updated species attributes
- de.py
   - Updated equations
- ultra45_raw_sc_ultrarawimg_withFSWcalcs_FM45_40P_Phi28p5_BeamCal_LinearScan_phi2850_theta-000_20240207T102740.csv
   - Added test data from Bob DeMajistre

## Testing
- test_ultra_l1b_extended.py
- test_de.py


imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -163,10 +163,13 @@ energy:
...
 
 species:
-  <<: *default_uint8
-  CATDESC: Species bin (0-255).
-  FIELDNAM: species
-  LABLAXIS: species
+  <<: *default
+  DISPLAY_TYPE: no_plot
+  CATDESC: Label species type.
+  FIELDNAM: Label species type.
```
nkerman, line 169: Should these be redundant?

imap_processing/ultra/l1b/de.py
```diff
@@ -59,7 +58,7 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
     energy = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
     # TODO: uint8 fills with zeros instead of nans.
     #  Confirm with Ultra team what fill values and dtype we want.
-    species_bin = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.uint8)
+    species_bin = np.full(len(de_dataset["epoch"]), "UNKNOWN", dtype="U10")
```
nkerman: If I understand it correctly, we want an integer type with a clear NaN value. Could we  use a slightly larger signed int type and set NaNs as -1 or -max_value?

imap_processing/ultra/constants.py
```diff
@@ -67,3 +67,7 @@ class UltraConstants:
     ALPHA = 0.2  # deltaE/E
     ENERGY_START = 3.385  # energy start for the Ultra grids
     N_BINS = 23  # number of energy bins
+
+    # Constants for species determination
+    SPECIES_MIN = 50
+    SPECIES_MAX = 200
```
nkerman: Perhaps a more descriptive name or at least comment? These are min/max of CTOF to consider an event to be an ENA, right?

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -666,106 +668,62 @@ def get_ctof(tof: np.ndarray, path_length: np.ndarray, type: str) -> NDArray:
...
     type : str
-        Type of event, either "ph" or "ssd".
+        Type of event, either "PH" or "SSD".
 
     Returns
     -------
     ctof : np.ndarray
         Corrected TOF (tenths of a ns).
+    magnitude_v : np.ndarray
+        Magnitude of the particle velocity (tenths of a ns / mm).
```
nkerman: 1. mag_v is not currently used for anything aside from tests - I take it mag_v will be used in future functions?

2. Also I think it's just reversed in the comment not the code, but shouldn't this be `mm /tenths of a ns`? Not sure where the '10ths of a...' should come in, but the velocity should be a spatial (mm) over a temporal (ns) unit. 
Below, a comment says it's converted to km/s though, so maybe it should be km/s.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -648,7 +648,9 @@ def get_energy_ssd(de_dataset: xarray.Dataset, ssd: np.ndarray) -> NDArray[np.fl
     return energy_norm
 
 
-def get_ctof(tof: np.ndarray, path_length: np.ndarray, type: str) -> NDArray:
+def get_ctof(
+    tof: np.ndarray, path_length: np.ndarray, type: str
+) -> tuple[NDArray, NDArray]:
     """
     Calculate the corrected TOF.
```
nkerman: and the magnitude of the particle velocity.

nkerman (commented):
Looks good! I mostly had questions/comments about units in a docstring.

imap_processing/ultra/l1b/de.py
```diff
@@ -59,7 +58,7 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
     energy = np.full(len(de_dataset["epoch"]), np.nan, dtype=np.float32)
     # TODO: uint8 fills with zeros instead of nans.
     #  Confirm with Ultra team what fill values and dtype we want.
```
tmplummer: I assume that this TODO goes away with the below changes?

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -666,106 +668,62 @@ def get_ctof(tof: np.ndarray, path_length: np.ndarray, type: str) -> NDArray:
...
+        Magnitude of the particle velocity (tenths of a ns / mm).
     """
     dmin_ctof = getattr(UltraConstants, f"DMIN_{type}_CTOF")
 
     # Multiply times 100 to convert to hundredths of a millimeter.
     ctof = tof * dmin_ctof * 100 / path_length
 
-    return ctof
+    # Convert from mm/0.1ns to km/s.
+    magnitude_v = dmin_ctof / ctof * 1e4
```
tmplummer, line 686: Double check the conversion... 
mm -> km = 1e-3
1/0.1ns -> 1/s = 1e10 (?)
mm/0.1ns -> km/s = 1e-3 * 1e10 = 1e7. 
Maybe I'm doing it wrong but I thought that 1e4 seemed off.

tmplummer (commented):
Overall, looks good to me. Just a couple things to check.

imap_processing/ultra/l1b/ultra_l1b_extended.py
nkerman: @subagonsouth - you're missing a factor of 1000 converting mm ($10^{-3}$) to km ($10^{3}$).

$\frac{1}{0.1 } \frac{mm}{ns} = \frac{10^{-3}}{0.1 * 10^{-9}} \frac{m}{s} = \frac{10^{-3}}{ 10^{-10}}  \frac{m}{s} = 10^{7}  \frac{m}{s} = 10^{4}  \frac{km}{s}$ 

P.S. Clearly I have way too much fun with LaTEX 😁

demajistre:
Units are always tedious, and I'm providing a way to check the overall conversions here. For the level zero file we provided:
`ultra45_raw_sc_ultrarawimg_withFSWcalcs_FM45_40P_Phi28p5_BeamCal_LinearScan_phi2850_theta-000_20240207T102740.csv`
Note that these match the beam energy used in the calibration that generated the file

if you do the conversions correctly, you should get the following histograms for the velocity and energy
vplot

eplot

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
laspsandoval: Yes. I think it's fine. The description can be longer if we want, but I don't think that it needs more detail.

imap_processing/ultra/l1b/de.py
laspsandoval: @demajistre:

You had put that you would like fill type "UKNOWN" and species "H" so I put this as a string. Is that ok?

imap_processing/ultra/l1b/ultra_l1b_extended.py
laspsandoval: 1. Yes (short answer). I'll provide that info in the next PR.
2. Yep. Good catch.

laspsandoval:
Screenshot 2024-12-20 at 2 10 56 PM

laspsandoval:
Note: the rest of Bob's calculations will be incorporated into the next PR. So I'll match his energy plot there.

nkerman (approved):
LGTM!

PR #1229, opened 2024-12-18
SWE: spin phase

### LINKED ISSUE #886: SWE: Calculate spin phase
As described in the algorithm document, SWE measurements are made at 30 spin angles in each quarter-cycle measurement. Because the SWE measurements are not synced to the spacecraft spin, the physical angles for each measurement will vary over time. In Level 1 SWE data products, the data are organized by the spin sector of each measurement, which is simply an index from 1-15. As part of Level 2 processing, the spin sectors are converted to spin angles of the spacecraft by combining SWE timing information with knowledge of the spacecraft spin from spacecraft ephemeris data.

tech3371:
# Change Summary
closes #886 

## Overview
Setup L2 processing function and calculate spin phase in it.

## Updated Files
   - calculate spin phase using existing utils function from geometry.py
- pyproject.toml
   - added ignore to be able to reshape return value of float or numpy array return type
## Testing

### DIFF
--- a/imap_processing/swe/l2/swe_l2.py
+++ b/imap_processing/swe/l2/swe_l2.py
@@ -201,8 +201,7 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
     # calculate spin phase
     spin_phase = get_spacecraft_spin_phase(
         query_met_times=data_acq_time,
-    )
-    spin_phase = spin_phase.reshape(-1, 24, 30, 7)
+    ).reshape(-1, 24, 30, 7)
     # TODO: organize flux data by energy and spin_phase.
     # My understanding from conversation with Ruth is that this is the hardest part
     # and last part of the L2 processing.
--- a/imap_processing/swe/l2/swe_l2.py
+++ b/imap_processing/swe/l2/swe_l2.py
@@ -1,9 +1,14 @@
-"""SWE L2 processing module."""
+"""
+SWE L2 processing module.
+
+This module contains functions to process L1B data to L2 data products.
+"""
 
 import numpy as np
 import numpy.typing as npt
 import xarray as xr
 
+from imap_processing.spice.geometry import get_spacecraft_spin_phase
 from imap_processing.swe.utils.swe_utils import read_lookup_table
 
 # TODO: add these to instrument status summary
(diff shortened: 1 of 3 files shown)


tech3371:
pre-commit.ci autofix

imap_processing/swe/l2/swe_l2.py
```diff
@@ -169,3 +174,44 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    """
+    # calculate flux
+    flux = calculate_flux(l1b_dataset)
+
+    # Calculate spin phase using SWE sci_step_acq_time_sec calculated in l1b.
+    # L1B dataset stores it by (epoch, energy, angle, cem).
+    data_acq_time = l1b_dataset["sci_step_acq_time_sec"].data.flatten()
+
+    # calculate spin phase
+    spin_phase = get_spacecraft_spin_phase(
```
tmplummer, line 202: Does SWE want the spacecraft spin phase, or instrument spin phase. In case they want instrument spin phase, see this function: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/geometry.py#L237

tmplummer (commented):
Looks good. Just the one question about the spin phase function getting used.

imap_processing/swe/l2/swe_l2.py
tech3371: SWE algorithm document says
>As part of Level 2 processing, the spin sectors are converted to spin angles of the spacecraft by combining SWE timing information with knowledge of the spacecraft spin from spacecraft ephemeris data.

This says spacecraft spin angles. What do you think? And I can double check with Ruth when I meet her after new year.

tech3371: And I asked her if I can use formula we discussed a while ago for that `get_spacecraft_spin_phase` function and she said, yes. But I wanted to double check this week but she couldn't meet.

tmplummer: Good question. I think the only way to get it right is to explain the difference between spacecraft spin-phase and instrument spin-phase and ask what she wants. I'll point out that the quote you put in says "spin angles", so double check that phase rather than angle is OK as well.

tech3371: for sure. Will do!

tmplummer (approved):
Approving so that you can sort this out however you want. Merging and writing a ticket for follow-up is appropriate.

tech3371:
> Approving so that you can sort this out however you want. Merging and writing a ticket for follow-up is appropriate.

Thank you Tim. I created issue to address some of feedback from this and last PR. I will revisit in January. https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1232

PR #1233, opened 2024-12-20
I-ALiRT - Rewrite doppler calculation

torimarbois:
# Change Summary

## Overview
Replace the placeholder `calculate_doppler()` function with the actual calculation. GH issue found [here](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/399).

## Updated Files

- .gitignore
   - Added externally-downloaded earth_combined kernel to ignore list
- process_ephemeris.py
   - Changed calculate_doppler() to use the spice spkezr function
   - Renamed instances of the position location object to be more clear
- conftest.py
   - Added another external kernel that can be downloaded using the `_download_external_kernels` fixture
- test_process_ephemeris.py
   - Updated test for `calculate_doppler`

## Testing
I've updated the unit test associated with the function to furnish the necessary kernels and check that the function output `is not None`. Similar to the other functions in `process_ephemeris.py`, we don't have real expected values to compare results against at the moment.

## Sources
- Curryer code (big thanks to Brandon S. for pointing me towards this) that has a doppler correction (not quite the same as what we're doing, which is producing simply the doppler velocity): [Curry code](https://github.com/lasp/curryer/blob/main/curryer/compute/ephemeris.py#L78-L150)
- STEREO code that calculates doppler, azimuth and elevation: [STEREO code](https://soho.nascom.nasa.gov/solarsoft/stereo/ssc/idl/spice/ssc_write_station.pro)

### DIFF
--- a/imap_processing/ialirt/process_ephemeris.py
+++ b/imap_processing/ialirt/process_ephemeris.py
@@ -59,8 +59,7 @@ def latitude_longitude_to_ecef(
     flattening = (equatorial_radius - polar_radius) / equatorial_radius
 
     # Convert geodetic coordinates to rectangular coordinates
-    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.geo
-    # (url cont.) rec
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.georec
     rect_coords = spice.georec(
         longitude_radians, latitude_radians, altitude, equatorial_radius, flattening
     )
@@ -187,6 +186,7 @@ def calculate_doppler(
     doppler = []
 
     for timestamp in observation_time:
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.spkezr
         state, light_time = spice.spkezr(target, timestamp, "ITRF93", "LT+S", "EARTH")
         # re-orient spacecraft position relative to the ground station
         state = state - np.pad(ground_station_position_ecef, (0, 3), "constant")
--- a/imap_processing/ialirt/process_ephemeris.py
+++ b/imap_processing/ialirt/process_ephemeris.py
@@ -23,30 +23,6 @@ logger = logging.getLogger(__name__)
 
 
 @typing.no_type_check
-def calculate_doppler(
-    observation_time: Union[float, np.ndarray],
-) -> Union[int, ndarray[float]]:
-    """
-    Calculate the doppler shift. Placeholder for now.
-
-    Parameters
-    ----------
-    observation_time : float or np.ndarray
-        Time at which the state of the target relative to the observer
-        is to be computed. Expressed as ephemeris time, seconds past J2000 TDB.
-
-    Returns
-    -------
-    doppler : float or np.ndarray[float]
-        Doppler shift. Currently a throwaway value.
-    """
-    if isinstance(observation_time, np.ndarray):
-        return np.ones(len(observation_time), dtype=float)
-    else:
-        return 1
-
-
-@typing.no_type_check
 @ensure_spice
 def latitude_longitude_to_ecef(
     longitude: float, latitude: float, altitude: float
@@ -99,7 +75,7 @@ def calculate_azimuth_and_elevation(
     latitude: float,
     altitude: float,
     observation_time: Union[float, np.ndarray],
-    target: SpiceBody = SpiceBody.IMAP.name,
+    target: str = SpiceBody.IMAP.name,
 ) -> tuple:
     """
     Calculate azimuth and elevation.
@@ -127,7 +103,9 @@ def calculate_azimuth_and_elevation(
     elevation : np.ndarray
         Elevation in degrees.
     """
-    observer_position_ecef = latitude_longitude_to_ecef(longitude, latitude, altitude)
+    ground_station_position_ecef = latitude_longitude_to_ecef(
+        longitude, latitude, altitude
+    )
 
     if not isinstance(observation_time, np.ndarray):
         observation_time = [observation_time]
@@ -144,7 +122,7 @@ def calculate_azimuth_and_elevation(
             abcorr="LT+S",  # Aberration correction
             azccw=False,  # Azimuth measured clockwise from the positive y-axis
             elplsz=True,  # Elevation increases from the XY plane toward +Z
-            obspos=observer_position_ecef,  # observer pos. to center of motion
+            obspos=ground_station_position_ecef,  # observer pos. to center of motion
             obsctr="EARTH",  # Name of the center of motion
             obsref="IAU_EARTH",  # Body-fixed, body-centered reference frame wrt
             # observer's center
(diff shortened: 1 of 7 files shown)


imap_processing/ialirt/process_ephemeris.py
```diff
@@ -158,6 +135,66 @@ def calculate_azimuth_and_elevation(
...
+    ground_station_position_ecef = latitude_longitude_to_ecef(
+        longitude, latitude, altitude
+    )
+
+    if not isinstance(observation_time, np.ndarray):
+        observation_time = [observation_time]
+
+    # TODO: consider improving vectorizing this for loop by moving everything besides
+    #  the spkezr outside of the for loop,
+    # TODO: which would require the axis argument
```
torimarbois: Open to thoughts here about whether or not I should move the non-spkezr lines out of the for loop below, and instead append the state, light_time results to an array that gets modified in these moved lines outside of the loop. This might improve performance slightly, but I did not make this change because this felt simpler to me.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -158,6 +135,66 @@ def calculate_azimuth_and_elevation(
...
+    # TODO: error handling with a try/except block
```
torimarbois: @subagonsouth do I need to add a try/except block around this spice function?

tmplummer: Do you mean `@ensure_spice` ? If so, the answer is yes. In one form or another, you need `@ensure_spice` to wrap this spice call. There is a function that wraps `spice.spkezr` in the `spice.geometry` module called `imap_state` although, it will require modification to be used here as I didn't anticipate the need for doing the one-way light time correction.

imap_processing/ialirt/process_ephemeris.py
tmplummer: Actually, I don't think that you need any for loop as it looks like `spiceypy.spkezr` is already vectorized.

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
```diff
@@ -7,20 +7,46 @@
...
+    time_endpoints = ("2026 SEP 22 00:00:00", "2026 SEP 22 23:59:59")
+    time_interval = int(1e3)  # seconds between data points
+    observation_time = np.arange(
+        str_to_et(time_endpoints[0]), str_to_et(time_endpoints[1]), time_interval
+    )
+    with furnish_kernels(kernels):
+        doppler_result = process_ephemeris.calculate_doppler(
+            longitude, latitude, altitude, observation_time
+        )
+    assert doppler_result is not None
```
tmplummer, line 43: Is there any way to validate values?

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -158,6 +135,66 @@ def calculate_azimuth_and_elevation(
...
+    doppler = []
+
+    for timestamp in observation_time:
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.spkezr
+        state, light_time = spice.spkezr(target, timestamp, "ITRF93", "LT+S", "EARTH")
+        # re-orient spacecraft position relative to the ground station
```
tmplummer: Rather than saying this is re-orienting, I think it should say it is shifting the position by subtracting the ground station location relative to the center of the earth.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -158,6 +135,66 @@ def calculate_azimuth_and_elevation(
...
+        state = state - np.pad(ground_station_position_ecef, (0, 3), "constant")
```
tmplummer: The use of np.pad is unintuitive to me. Could this just be:
```suggestion
        state[0:3] -= ground_station_position_ecef
```

tmplummer (commented):
Looks good overall, just a couple of things to consider.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -158,6 +135,66 @@ def calculate_azimuth_and_elevation(
...
+        observation_time = [observation_time]
+
+    # TODO: consider improving vectorizing this for loop by moving everything besides
+    #  the spkezr outside of the for loop,
+    # TODO: which would require the axis argument
+    # TODO: error handling with a try/except block
+    doppler = []
+
+    for timestamp in observation_time:
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.spkezr
```
laspsandoval: Could you put a simple docstring saying this is finding the position and velocity relative to the center of the earth? It will help me remember, I think.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -158,6 +135,66 @@ def calculate_azimuth_and_elevation(
...
+        state, light_time = spice.spkezr(target, timestamp, "ITRF93", "LT+S", "EARTH")
+        # re-orient spacecraft position relative to the ground station
+        state = state - np.pad(ground_station_position_ecef, (0, 3), "constant")
+        doppler.append(np.sum(state[3:6] * state[0:3]) / np.linalg.norm(state[0:3]))
```
laspsandoval: Put a docstring above this line:
Calculate the radial velocity.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -158,6 +135,66 @@ def calculate_azimuth_and_elevation(
...
+
+    return np.asarray(doppler)
+
+
```
laspsandoval, line 191: Make certain to update build_output. Right now I get:
TypeError: calculate_doppler() missing 3 required positional arguments: 'latitude', 'altitude', and 'observation_time'. Which is why I think that your tests are failing.

laspsandoval (commented):
This is a very clean way to calculate the doppler. Nicely done.

I agree w/ Tim regarding using the tool in geometry.py and also removing the for loop. Don't forget to update build_output. And also add a unit docstring there too. 

Thanks for doing this. Excellent work.

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
torimarbois: Not yet - we're hopefully going to receive test data from STEREO which we can use to validate our code at some point soon. Right, @laspsandoval?

imap_processing/ialirt/process_ephemeris.py
torimarbois: If I remove the use of np.pad, I'll have to figure out a way to also do this position correction for a state of multiple rows. I haven't figured out how to condense this to one line, so it would be separated into some if/else logic. As it is, I can address both cases (one row vs. multiple rows) in one line.

imap_processing/tests/ialirt/unit/test_process_ephemeris.py
laspsandoval: That's correct.

laspsandoval (approved):
Looks good to me. I wasn't as thorough the second time around, but I think that you addressed all of the comments. Well done!

imap_processing/spice/geometry.py
```diff
@@ -84,7 +84,8 @@ class SpiceFrame(IntEnum):
 @ensure_spice
 def imap_state(
     et: Union[np.ndarray, float],
-    ref_frame: SpiceFrame = SpiceFrame.ECLIPJ2000,
+    ref_frame: str = SpiceFrame.ECLIPJ2000.name,
```
tmplummer: Rather than changing the input type for `ref_frame` from `SpiceFrame` to `str`, can you please add ITRF93 to the `SpiceFrame` enumeration definition. `ITRF93 = 13000`. I think that it is a bit more user friendly/less error prone for folks to have a restricted set of frame inputs to choose from. It is also in line with all of the other geometry related function APIs.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -144,20 +121,77 @@ def calculate_azimuth_and_elevation(
...
+    )
+
+    # find position and velocity relative to the center of the earth using spice spkezr
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.spkezr
+    state = imap_state(
+        et=observation_time, ref_frame="ITRF93", abcorr="LT+S", observer=SpiceBody.EARTH
+    )
+    # shifting position by subtracting ground station location relative to the center
+    # of the earth
+    state = state - np.pad(ground_station_position_ecef, (0, 3), "constant")
```
tmplummer: I'm fine with you leaving this as is, but the more pythonic way to do this is using the ellipsis to slice: `state[..., 0:3] -= ground_station_position`. The `[..., 0:3]` indexing will expand so that you are always indexing the first three elements of the final dimension. Numpy explanation of Ellipsis: https://numpy.org/doc/stable/user/basics.indexing.html#dimensional-indexing-tools

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -144,20 +121,77 @@ def calculate_azimuth_and_elevation(
...
+    ax = int(state.ndim == 2)
+    # calculate radial velocity
+    if isinstance(observation_time, np.ndarray):
+        doppler = np.sum(state[:, 3:6] * state[:, 0:3], axis=ax) / np.linalg.norm(
+            state[:, 0:3], axis=ax
+        )
+    else:
+        doppler = np.sum(state[3:6] * state[0:3], axis=ax) / np.linalg.norm(
+            state[0:3], axis=ax
+        )
```
tmplummer: I think that the ellipsis slicing also works here.
```suggestion
    doppler = np.sum(state[..., 3:6] * state[..., 0:3], axis=-1) / np.linalg.norm(
            state[..., 0:3], axis=-1
        )

```

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -144,20 +121,77 @@ def calculate_azimuth_and_elevation(
...
         azimuth.append(np.rad2deg(azel_results[0][1]))
         elevation.append(np.rad2deg(azel_results[0][2]))
 
-    # TODO: potentially use the velocity components returned from azlcpo to
-    # TODO: calculate doppler
-
     return np.asarray(azimuth), np.asarray(elevation)
 
 
+@typing.no_type_check
```
tmplummer, line 135: Can this be removed now that it is not wrapped with `@ensure_spice`?

tmplummer (commented):
I have one change request for how `imap_state` was modified and a couple other things to think about.
Thanks!

imap_processing/ialirt/process_ephemeris.py
torimarbois: I've never seen ellipsis slicing before, this is super handy. I like this a lot more and I'll replace the `pad` function with the ellipsis logic. Thanks Tim!

imap_processing/spice/geometry.py
torimarbois: Just to make sure I'm understanding @subagonsouth - you're saying to undo the type change here so that `ref_frame` is back to being a SpiceFrame object, and add the line `ITRF93 = 13000` to the SpiceFrame definition so that in `calculate_doppler` I'll pass in `SpiceFrame.ITRF93` for `ref_frame` rather than `ITRF93`?

tmplummer: Yes, that is what I was trying to get across.

tmplummer (approved):
LGTM. Thanks for making those changes!

PR #1234, opened 2024-12-20
Fix data array shapes for CoDICE-Lo products

bourque:
This PR fixes some issues with how CoDICE-Lo data arrays are shaped, as well as an indexing issue in Lossy A and B table lookup.

### DIFF
--- a/imap_processing/codice/codice_l1a.py
+++ b/imap_processing/codice/codice_l1a.py
@@ -129,12 +129,12 @@ class CoDICEL1aPipeline:
         for name in self.config["coords"]:
             if name == "epoch":
                 values = self.dataset.epoch.data
+            elif name == "esa_step":
+                values = np.arange(self.config["num_energy_steps"])
             elif name == "inst_az":
                 values = np.arange(self.config["num_positions"])
             elif name == "spin_sector":
                 values = np.arange(self.config["num_spin_sectors"])
-            elif name == "esa_step":
-                values = np.arange(self.config["num_energy_steps"])
             else:
                 # TODO: Need to implement other types of coords
                 continue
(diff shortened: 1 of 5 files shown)


imap_processing/codice/codice_l1a.py
```diff
@@ -129,12 +129,12 @@ def define_coordinates(self) -> None:
...
             if name == "epoch":
                 values = self.dataset.epoch.data
+            elif name == "esa_step":
+                values = np.arange(self.config["num_energy_steps"])
             elif name == "inst_az":
                 values = np.arange(self.config["num_positions"])
             elif name == "spin_sector":
                 values = np.arange(self.config["num_spin_sectors"])
-            elif name == "esa_step":
-                values = np.arange(self.config["num_energy_steps"])
```
bourque, line 137: I just changed the ordering here for consistency with the order of the array shapes

imap_processing/codice/decompress.py
```diff
@@ -50,7 +50,9 @@ def _apply_lossy_a(compressed_bytes: bytes) -> list[int]:
         The 24- or 32-bit decompressed values.
     """
     compressed_values = list(compressed_bytes)
-    decompressed_values = [LOSSY_A_TABLE[item] for item in compressed_values]
+    decompressed_values = [
+        LOSSY_A_TABLE[item - 1] if item > 0 else 0 for item in compressed_values
+    ]
```
tech3371, line 55: Why decompressed value is value of `item - 1` meaning the previous value?

tech3371 (commented):
I had one clarification question. Besides, LGTM

imap_processing/codice/decompress.py
bourque: I am using it as a way to implement the "<" condition for the lookup table. From Greg Dunn's instructions on this:

Screenshot 2024-12-20 at 15 49 25

So here I choose the previous index that corresponds to the value.

tech3371 (approved):

PR #1236, opened 2024-12-20
Binning prototype

demajistre:
# Change Summary

## Overview
This is pull request is NOT meant to be merged.  Rather it provides some example code for an abstract class for binning schemes and some examples of their use. 

## Overall structure
All the example code is in the ultra_user package, which contains code for us to execute the ultra codebase locally from some current state of the ISOC repo. The current package has a sub-package called 'energy_check', and these changes are not important to the binning schemes - they are just there to provide testing data and functionality (mostly because the initial version we pulled from did not have the velocity corrections necessary for testing the binning schemes). 

The code of interest here is in mapbin, which provides:
binscheme:  an abstract class for binning types
lonLatscheme: an example binning type - longitude latitude
hpscheme: an example binning type - using healpix
mapBinner: an example of the use of the abstract class
binfragments: some python that 'test drives' the modules in the package


nkerman:
Marking this as a draft to make it more clear that it's not to be merged/reviewed for merge.

laspsandoval:
Implemented healpix binning in another ticket.

PR #1239, opened 2025-01-02
Glows L2 Base processing

maxinelasp:
# Change Summary
This PR finishes most of the L2 processing for GLOWS. 

## Overview

As usual, glows_l2_data contains the structure and some key information on the L2 output file, while glows_l2 contains processing code. glows_l2 has the processing steps to split data up by spin count (called observation day in this code) and then run averaging on all the data. There is also some filtering of "bad times" and "bad angles" done at this stage. The bad times and bad flag filtering is complete, bad angle filtering is incomplete. 

Most of the incomplete pieces are related to SPICE (needs to be incorporated in L1B) and spin tables. 

## New Files

- glows_l2.py
   - GLOWS L2 processing 
- glows CDF attributes file
  -  best guess finished by me, runs without any errors. Still needs additional validation by ISTP and GLOWS team.


imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,358 @@
...
+
+spin_axis_orientation_std_dev:
+  <<: *support_data_defaults
+  CATDESC: obs day averaged std_dev of spin axis orientation
+  FIELDNAM: std_dev of spin axis orientation
+  FILLVAL: *max_uint16
+  FORMAT: I4
+  LABLAXIS: std_dev
+
+spin_axis_orientation_average:
```
tmplummer, line 251: Is this a 3 element (Cartesian) unit vector which defines the spin axis? It would be useful for users to document what reference frame the orientation is expressed in.

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,358 @@
...
+  <<: *support_data_defaults
+  CATDESC: obs day averaged spin axis orientation
+  FIELDNAM: Averaged spin axis orientation
+  FILLVAL: *max_uint16
+  FORMAT: I4
+  LABLAXIS: Avg spin axis orientation
+
+spacecraft_location_average:
```
tmplummer, line 259: I think it is worth bringing up with GLOWS that average location is problematic. The average will change depending on if you express the location in Cartesian or spherical coordinates. It also changes with how often you sample the s/c location. I think that the most useful value to store would be the s/c location at the center of the time window covered by the data product.

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,358 @@
...
+  <<: *support_data_defaults
+  DEPEND_1: ecliptic
+  CATDESC: obs day averaged spacecraft location
+  FIELDNAM: Averaged spacecraft location
+  FILLVAL: *max_uint16
+  FORMAT: I4
+  LABLAXIS: Avg spacecraft location
+
+spacecraft_location_std_dev:
```
tmplummer, line 268: S/C Location std_dev is even weirder than average location. I recommend getting explicit details of what is desired here and either extending the CATDESC or adding VAR_NOTES to define all of these geometry variables.

imap_processing/glows/ancillary/imap_glows_pipeline_settings_v001.json
tmplummer, line 1: Is this file used and is there an accessor function?

I like the use of a configuration file to store all of these configuration variables.

imap_processing/glows/l2/glows_l2.py
```diff
@@ -0,0 +1,323 @@
...
+    attrs : ImapCdfAttributes
+        CDF attributes for GLOWS L2.
+
+    Returns
+    -------
+    xarray.Dataset
+        L2 dataset for output to CDF file.
+    """
+    # Each L2 file only has one timestamp.
+    # TODO: If we want this to point to the start time, we need to set the attribute
```
tmplummer, line 195: Since this is L2, it must be ISTP compliant. ISTP says that the time must be the center of the time bin (see link). This is TBR by Andriy at SPDF because there is a `BIN_LOCATION` attribute that specifies the location in the bin of the stored value.
https://spdf.gsfc.nasa.gov/istp_guide/variables.html#Epoch
https://spdf.gsfc.nasa.gov/istp_guide/vattributes.html#Bin_Location

imap_processing/glows/l2/glows_l2.py
```diff
@@ -0,0 +1,323 @@
...
+    )
+
+    bins_label = xr.DataArray(
+        -1,
+        name="bins_label",
+        attrs=attrs.get_variable_attributes("bins_label", check_schema=False),
+    )
+
+    flags = xr.DataArray(
+        np.ones(17),
```
tmplummer: Why is this a 17 element array? Are you storing one value per flag instead of a bitmask?

Can the 17 be retrieved from a flag definition so that if a flag is added, you don't have to find this obscure code location to update this value?

imap_processing/tests/glows/test_glows_l2.py
```diff
@@ -0,0 +1,101 @@
...
+def test_glows_l2(l1b_hist_dataset):
+    l2 = glows_l2(l1b_hist_dataset, "v001")[0]
+    assert l2.attrs["Logical_source"] == "imap_glows_l2_hist"
+
+    assert np.allclose(l2["filter_temperature_average"].values, [57.6], rtol=0.1)
+
+
+@pytest.mark.xfail(reason="Spin table not yet complete")
+def test_split_by_observational_day(l1b_hist_dataset):
+    # TODO: Complete test when spin table is complete
```
tmplummer, line 40: There are spin table functions and fixtures available, but I think that you need the repointing table here. There are no repointing table functions yet.

tmplummer (commented):
Nice progress here. I left some feedback that is mostly just food for thought regarding some of the geometry related items.

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,358 @@
...
+  VAR_TYPE: data
+  LABL_PTR_1: bins_label
+
+bins_dim:
+  <<: *default_attrs
+  VALIDMIN: 0
+  VALIDMAX: 3599
+  CATDESC: Histogram bin number
+  FIELDNAM: Bin number
+  FORMAT: I5
```
bourque, line 61: Should this be `I4` since the `VALIDMAX` is less than 10,000?

bourque: I see now that `FILLVAL` has five digits so perhaps this is fine. I get confused when it comes to `FORMAT` -- should the `FORMAT` match _expected_ values, or _possible_ values?

imap_processing/tests/glows/test_glows_l2.py
```diff
@@ -0,0 +1,101 @@
...
+    input["histogram"] = hist
+
+    return input
+
+
+def test_glows_l2(l1b_hist_dataset):
+    l2 = glows_l2(l1b_hist_dataset, "v001")[0]
+    assert l2.attrs["Logical_source"] == "imap_glows_l2_hist"
+
+    assert np.allclose(l2["filter_temperature_average"].values, [57.6], rtol=0.1)
```
bourque, line 35: Cool, I didn't know about this `allclose` function. I assume this means that all of the values just need to be within `0.1` of `57.6`?

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -3,6 +3,139 @@
...
+        ecliptic longitude of bin centers [deg]
+    ecliptic_lat : numpy.ndarray
+        ecliptic latitude of bin centers [deg]
+    number_of_bins : int
+        number of bins in lightcurve
+    raw_uncertainties : numpy.ndarray
+        statistical uncertainties for raw histograms (sqrt of self.raw_histograms)
+    l1b_data : xarray.Dataset
+        L1B data filtered by good times, good angles, and good bins.
+    """
```
bourque, line 41: Excellent documentation and naming of variables here!

imap_processing/glows/ancillary/imap_glows_pipeline_settings_v001.json
maxinelasp: It is used, it will be moved into S3 ancillary files so it will be accessed via the API. That's not complete yet (though hopefully coming soon!)

imap_processing/glows/l2/glows_l2.py
maxinelasp: Yes, good plan. This specific shape was requested by the GLOWS team, so it is an array of flags rather than a bitmask.

bourque (approved):
Looks great, and nice to have the remaining work very thoroughly marked with TODOs!

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
maxinelasp: Yeah, I am not sure. I have assumed possible values, but it could go either way. I do plan on sending these attributes through the ISTP process after GLOWS gets a chance to validate, so we will find out!

imap_processing/tests/glows/test_glows_l2.py
maxinelasp: Yeah, this is only close because I previously was splitting up the data by a fake observation day and so the averages didn't quite line up. I will validate against the actual matches in my next PR (where I will be exactly on, not just close, haha)

In general it's a useful tool though! This way I can slightly modify `l1b_hist_dataset` and it doesn't break the tests.

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: Copied out of the GLOWS algorithm document 😅 but it is nice to have it all in one spot in the dataclass for easy reference!

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
mstrumik: Actually this was intended to be the longitude and latitude of the spin axis, not the Cartesian unit vector. The frame is to be the ecliptic J2000.

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
mstrumik: I fully agree that the average depends on Cartesian vs spherical coordinates and the sampling of the s/c location. I understand your concerns @subagonsouth, but we tried to build a consistent system for location avg and std_dev starting from out L1b. In L1b we have 2-minute time intervals for accumulating the histograms. We assumed that the avg and std_dev for the location can be determined from SPICE for the 2-minute histograms. Then we have a process of adding the 2-minute histograms to get the daily histogram. There are some bad-time 2-minute histograms that are excluded in this process. So we envisioned that we should build the daily avg and std_dev for the location taking into account the bad-time exclusions. This is why we prefer not to use the center of the time window.

Generally our approach to averages and std_devs is consistent for all averaged quantities (both onboard-collected and computed on the ground). It is described in our algorithm document.

There is one problem here, which will occur later in the processing pipeline in L3d, where we will be providing survival probabilities (SP) for ENA instruments. The SPs will be calculated for a given specified daily spin axis. It is important to ensure that the GLOWS team and the ENA teams use exactly the same "official" daily spin axis for SPs in L3d.

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
mstrumik: Please see my comment on the average above, I hope it makes the definitions clearer. I fully agree that the approach used in L1b and L2 should be well documented in CATDESC and/or VAR_NOTES to avoid confusions.

imap_processing/glows/ancillary/imap_glows_pipeline_settings_v001.json
```diff
@@ -0,0 +1,52 @@
...
+        "std_dev_threshold__volt": 50.0
+    },
+    "filter_based_on_spin_period_std_dev": {
+        "std_dev_threshold__sec": 0.033333
+    },
+    "filter_based_on_pulse_length_std_dev": {
+        "std_dev_threshold__usec": 1.0
+    },
+    "filter_based_on_maps": {
+        "angular_distance_for_flaging__deg": 2.5
```
mstrumik, line 25: Please note that the instrument team has recently provided a new version of this file v003, where angular_distance_for_flaging__deg has been replaced by angular_radius_for_excl_regions__deg. The new attribute angular_radius_for_excl_regions__deg is used in our code in a slightly different way as compared to angular_distance_for_flaging__deg.

imap_processing/glows/ancillary/imap_glows_pipeline_settings_v001.json
```diff
@@ -0,0 +1,52 @@
...
+        "is_pulse_length_std_dev_beyond_threshold": true,
+        "is_spin_period_difference_beyond_threshold": false
+    },
+    "active_bad_angle_flags": {
+        "is_close_to_uv_source": true,
+        "is_inside_excluded_region": true,
+        "is_excluded_by_instr_team": true,
+        "is_suspected_transient": true
+    }
+}
```
mstrumik, line 52: Following up on my previous comment regarding this file, in the new version v003 there is a new attribute number_of_good_histograms_at_night, which is used in our L2 code to say how many 2-minute histograms after raising the catbed_heater flag are considered as "good", even if they formally fall into the "night" category.

tmplummer (approved):
I'm happy with this. Let me know how I can help with geometry related stuff for GLOWS.

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
maxinelasp: Thanks for the comment @mstrumik! I agree adding a note into VAR_NOTES would help. I have added that to my notes.

imap_processing/glows/ancillary/imap_glows_pipeline_settings_v001.json
maxinelasp: Thank you for mentioning this! I will make that update in the next review.

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
maxinelasp: Thanks for clarifying @mstrumik! I may have further questions as I fully implement the SPICE code. I'll also work with Tim to ensure these variables make sense and are useful.

imap_processing/glows/l2/glows_l2.py
```diff
@@ -0,0 +1,324 @@
...
+    # per-timestamp.
+
+    # one dataset collects multiple epoch values which need to be averaged down into
+    # one value.
+    all_variables = dataclasses.fields(HistogramL1B)
+
+    daily_lightcurve = DailyLightcurve(good_data)
+
+    var_outputs = {
+        "total_l1b_inputs": len(good_data["epoch"]),
```
mstrumik, line 77: It looks like 'total_l1b_inputs" is set here to the same value as  "number_of_good_l1b_inputs".  If I correctly understand the code, "total_l1b_inputs" should be equal to len(l1b_dataset) (so the total number of l1b inputs, not only the good ones).

imap_processing/glows/l2/glows_l2.py
```diff
@@ -0,0 +1,324 @@
...
+            + l1b_dataset["imap_time_offset"].data[0]
+        )
+
+    for field in all_variables:
+        var_name = field.name
+        if "average" in var_name:
+            # This results in a scalar value, so `keepdims=True` ensures we keep the
+            # epoch dimension.
+            var_outputs[var_name] = (
+                l1b_dataset[var_name].mean(dim="epoch", keepdims=True).data
```
mstrumik, line 106: I am not well familiar with xarrays but it looks like averages here are computed over all l1b_dataset entries, while they should be computed using only good times.

imap_processing/glows/l2/glows_l2.py
```diff
@@ -0,0 +1,324 @@
...
+            )
+
+            var_outputs[var_name.replace("average", "std_dev")] = (
+                l1b_dataset[var_name].std(dim="epoch", keepdims=True).data
```
mstrumik, line 110: Similar remark here for std_dev as above for averages. The std_dev should be computed using only good times.

imap_processing/glows/l2/glows_l2.py
```diff
@@ -0,0 +1,324 @@
...
+    -------
+    numpy.ndarray
+        Histogram data with bad bins marked with -1.
+    """
+    # TODO: will need ancillary file imap_glows_exclusions_by_instr_team
+    # TODO: complete once unique_block_identifier is implemented
+    # file contains timestamp & bin filter array pairs. For the timestamp, the
+    # filter should be applied such that 1 excludes the bin.
+
+    # excluded bins can be marked with -1
```
mstrumik, line 141: There is no need to filter/exclude any bins in L2. The bad-angle flags in L1b should only propagate to bad-angle flags in L2, which then go to the L2 output. The L2 bad-angle flags will be used later in the L2->L3a processing. The handling of bad-angle flags for the L1b->L2 processing is shown in data_products_cbk_implementation_2024-12-11_3dayL0used/toolkit/l2_data_histogram.py#L358 and discussed in subsection "Bad-angle masking for Level-2" at the bottom of page 22 in the algorithm document v.4.3.7

imap_processing/glows/l2/glows_l2.py
```diff
@@ -0,0 +1,324 @@
...
+        "daily_lightcurve": daily_lightcurve,
+    }
+
+    if len(good_data["epoch"]) != 0:
+        # Generate outputs that are passed in directly from L1B
+        var_outputs["start_time"] = good_data["epoch"].data[0]
+        var_outputs["end_time"] = good_data["epoch"].data[-1]
+
+    else:
+        # No good times in the file
```
mstrumik, line 93: It seems that the algorithm document does not specify what to do if there are no good times in L1b. The code data_products_cbk_implementation_2024-12-11_3dayL0used/toolkit/ simply skips the generation of L2 if no good-time L1b data are found for a given observational day. Such approach is preferred by the instrument team if  possible from the point of view of ISTP and other standards.

imap_processing/glows/l2/glows_l2.py
```diff
@@ -0,0 +1,324 @@
...
+            "epoch": epoch_time,
+            "bins": bins,
+            "flags": flags,
+            "ecliptic": eclipic_data,
+        },
+        attrs=attrs.get_global_attributes("imap_glows_l2_hist"),
+    )
+
+    ecliptic_variables = [
+        "spacecraft_location_average",
```
mstrumik, line 251: I am not sure if it is relevant here, but spin_axis_orientation_average and spin_axis_orientation_std_dev are also "ecliptic variables". They contain the ecliptic longitude and latitude.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -3,6 +3,139 @@
...
+
+        Parameters
+        ----------
+        l1b_data : xarray.Dataset
+            L1B data filtered by good times, good angles, and good bins for one
+            observation day.
+        """
+        self.raw_histograms = self.calculate_histogram_sums(l1b_data["histogram"].data)
+
+        exposure_times_per_timestamp = (
```
mstrumik, line 70: These are formally exposure_times_per_bin. Actually per bin and per timestamp, but per timestamp simply stems implicitly from the processing of L1b here.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -3,6 +3,139 @@
...
+            l1b_data["spin_period_average"]
+            * l1b_data["number_of_spins_per_block"]
+            / 3600
```
mstrumik, line 73: The histogram resolution 3600 should not be hard coded here, because it is a configurable parameter that may vary depending on settings in the flight software of the instrument. The histogram resolution should be taken from L1b data. Generally, the histogram resolution will not be changed during a given observational day, it is planned to be changed only during "nights" or station keeping maneuvers, if necessary. So when adding together L1b histograms to get L2, the software may assume the same size of the arrays across entire observational day.

Perhaps it is convenient to move calculations of the exposure times further, to be done after setting self.number_of_bins. Or the other way around, move the definition of self.spin_angle and self.number_of_bins before the calculations of exposure times.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -3,6 +3,139 @@
...
+        self.photon_flux = np.zeros(len(self.raw_histograms))
+        self.flux_uncertainties = np.zeros(len(self.raw_histograms))
+
+        # TODO: Only where exposure counts != 0
+        if len(self.exposure_times) != 0:
+            self.photon_flux = self.raw_histograms / self.exposure_times
+            self.flux_uncertainties = self.raw_uncertainties / self.exposure_times
+
+        # TODO: Average this, or should they all be the same?
+        self.spin_angle = np.average(l1b_data["imap_spin_angle_bin_cntr"].data, axis=0)
```
mstrumik, line 89: All L1b histograms have the same spin_angle grid, where the "official" IMAP spin angle is used. So the averaging is not needed here. Then at L2 we would like to go to the spin angle measured from the northernmost point of the scanning circle.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -3,6 +3,139 @@
...
+
+        # TODO: is the first number here ok? Would it change mid-obs day?
+        self.number_of_bins = len(self.spin_angle)
```
mstrumik, line 92: As I wrote above, the first number is ok. No mid-obs-day changes expected.

imap_processing/mag/l1b/mag_l1b.py
mstrumik, line 1: I do not know why a MAG processing file ended up here.

imap_processing/tests/glows/test_glows_l2.py
mstrumik, line 35: I can see that this unit test (and many others here) check the values for a specific L0 synthetic data set. The instrument team is developing a number of testing L0 files with synthetic data. The SDC and GLOWS teams should presumably discuss how to organize the final testing campaign where many L0 files will be used.

mstrumik (approved):
Good job Maxine! Thank you. I included some comments throughout the code.

imap_processing/tests/glows/test_glows_l2.py
maxinelasp: Yes agreed! Since the unit tests run on every commit, ideally we would find a small but good variety to use for unit testing long term. Of course, we can also have tests which run semi-regularly against larger validation data sets. Once you have a more complete set of testing files, we can discuss.

imap_processing/mag/l1b/mag_l1b.py
maxinelasp: My mistake, I think it was some formatting issue.

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: Ok, so would that be the lowest value in L1B's "imap_spin_angle_bin_cntr" variable? Closest to 0? Also, to confirm, the `spin_angle` variable should be the same at all epoch values in level 2?

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: Yes, good catch. I should have access to the number of bins here, so I will use that value.

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: That makes sense, I will update the name.

imap_processing/glows/l2/glows_l2.py
maxinelasp: Here, "ecliptic" refers to the dimension used in the output. They have x, y, z coordinates. However, you are right that it is a little vague, I will think on different names for the variables.

imap_processing/glows/l2/glows_l2.py
maxinelasp: Sure, it's easy to skip processing. I can add that in the next chunk of code.

imap_processing/glows/l2/glows_l2.py
maxinelasp: Oh, good catch! I'll make that change

imap_processing/glows/l2/glows_l2.py
maxinelasp: This method is intended for the GLOWS team exclusions from the ancillary file "imap_glows_exclusions_by_instr_team_v002.dat", which I interpreted as a list of bins where the data is not counted as "good" and therefore should not be included in future calculations. The bad-angle flags are not implemented in this code yet. Will bins be excluded based on that file?

mstrumik: Bins will be excluded based on that file but not in L1b->L2 processing. We are doing it in the L2->L3a processing that has been delivered by the GLOWS team and incorporated into Menlo codes. But this exclusion at L2->L3a involves also other files and masks computed in the fly in the processing pipeline earlier.

In the code provided by the GLOWS team it is organized in the following way. In data_products_cbk_implementation_2024-12-11_3dayL0used/toolkit/l1b_data_histogram.py (so in L1a->L1b processing) in lines 307-316 there is a series of calls of _set_histogram_flag_array_*() methods:
self._set_histogram_flag_array_uv_sources(anc_input_from_instr_team, settings)
self._set_histogram_flag_array_excl_regions(anc_input_from_instr_team, settings)             self._set_histogram_flag_array_by_instr_team(anc_input_from_instr_team, 'exclusions_by_instr_team')
self._set_histogram_flag_array_by_instr_team(anc_input_from_instr_team, 'suspected_transients')
Each method sets a particular bit of the bad-angle flag as defined in L1B_L2_HIST_FLAG_FIELDS in constants.py. This particular file "imap_glows_exclusions_by_instr_team_v002.dat" you are asking about is handled by the third invocation, but it is only one from four bits that are set here. After those invocations, in the L1b JSON output we have "histogram_flag_array" for each 2-minute histogram.  Currently four bits are used as defined in  L1B_L2_HIST_FLAG_FIELDS mentioned above.

Then during the L1b->L2 processing, the 2-minute histograms are merged into one daily L2 histogram. Bad-angle flags from L1b are handled in data_products_cbk_implementation_2024-12-11_3dayL0used/toolkit/l2_data_histogram.py in lines 359-362. Basically, when processing 2-minute histograms, if any of them has a given bad-angle bit set to 1, the L2 daily histogram will have this bit set to 1 too. But additionally it takes into account the bad-angle mask 'active_bad_angle_flags' defined in 'imap_glows_pipeline_settings_v002.json' which simply says, which bits are "active" for the masking process. So the bad-angle flags in L1b only propagate to bad-angle flags in L2, which then simply go to the L2 output and are used later in the L2->L3a processing.

I suppose the confusion can be related to the changes that we have introduced in the meantime relative to the first version of our code shared with you. Please note that currently the method _set_histogram_flag_array_excl_by_instr_team() in l1b_data_histogram.py is not used at all. I should presumably have removed it as possibly introducing confusions.

imap_processing/glows/l2/glows_l2.py
mstrumik: Thanks for the explanation. Then perhaps "ecliptic_cartesian_variables"?

imap_processing/glows/l2/glows_l2_data.py
mstrumik: I am not sure if understand correctly the need to determine the lowest value. So my answer is: yes, closest to 0, but presumably we should discuss it during a Monday's tagup mtg.

Regarding the same spin_angle for all epoch values in level 2: no, it can change. The number of bins in the histograms producted by the instrument may vary from 225 to 3600. Then, in the ground processing each bin gets its spin angle at L1b (data_products_cbk_implementation_2024-12-11_3dayL0used/toolkit/l1b_data_histogram.py line 249). Then the spin angles from L1b should propagate to L2 but at some point 'position_angle_offset_average' must be accounted for.

imap_processing/glows/l2/glows_l2_data.py
mstrumik: I have just realized we should discuss during a tagup mtg what exactly the "first number" means.

maxinelasp: Yes, I will add it to the agenda so we remember to discuss

imap_processing/glows/l2/glows_l2.py
maxinelasp: I believe I understand what you are saying. I will look through and make some edits to this code, and we can discuss in the next tag-up meeting. Thank you for the thorough explanation.

PR #1242, opened 2025-01-03
Ultra l1b velocity revisions

### LINKED ISSUE #1140: Ultra l1b: Add velocity magnitude
### Algorithm Description:

Line 488:
annotated needs velocity (km/s)
get_particle_velocity routine: normalized velocity - good for direction
pulse height - larger for larger energies
species bin maps to mass

v=dmin/ctof (Bob to check)
energy_kev = 1/2*m(kg)*v(m/s)^2 (different from FSW results)
need to be added to extended l1b
Joules -> kev

def determine_species_pulse_height(
    pulse_height (name change): np.ndarray, tof: np.ndarray, path_length: np.ndarray
) -> NDArray:

def determine_species_ssd(
    ssd_energy: np.ndarray, tof: np.ndarray, path_length: np.ndarray
) -> NDArray:

Multiply the velocity magnitude by vx, vy, etc.

### LINKED ISSUE #1213: Ultra l1b: Put velocity as a tuple
### Summary

Also improve calculation based on Bob's comments

laspsandoval:
# Change Summary

## Overview
This is the (hopefully) last of the L1b direct event revisions. 
1. Calculate velocity magnitude 
2. New way to calculate energy using kinetic energy equation (in addition to what we have currently)
3. Calculate az and el
4. Update the velocity function
5. Incorporate the new values into the xarray

## Updated Files
- ultra_l1b_extended.py
   - Added new functions described above
- de.py
   - Added to the dictionary used to create the dataset
   - Updated cdf attributes 
- constants.py
   - Added a constant
- ultra_l1_utils.py
   - Made it to where the velocity is a tuple (instead of individual fields) in the xarray

## Testing
- test_ultra_l1b_extended.py
- test_de.py
- Updated test data (provided by Bob DeMajistre): ultra45_raw_sc_ultrarawimg_withFSWcalcs_FM45_40P_Phi28p5_BeamCal_LinearScan_phi2850_theta-000_20240207T102740.csv


imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -728,3 +750,34 @@ def determine_species(tof: np.ndarray, path_length: np.ndarray, type: str) -> ND
...
+        - azimuth : angle in the xy-plane
+          In radians:
+          output range=[0, 2*pi].
+        - elevation : angle from the z-axis
+          In radians:
+          output range=[-pi/2, pi/2].
+    """
+    # Compute azimuth (phi) angles and elevation (theta)
+    v_array = np.column_stack(v)
+    spherical_coords = cartesian_to_spherical(v_array, degrees=False)
```
demajistre: The documentation is incorrect here and in the header for `cartesian_to_spherical`.  The elevation angle (which is indeed the quantity calculated here) is not the "angle from the z-axis" (which is commonly called the "polar angle").  The quantity here is the "angle from the X/Y plane" (and this is the common definition for elevation or latitude).

imap_processing/ultra/constants.py
```diff
@@ -61,6 +61,7 @@ class UltraConstants:
 
     # Conversion factors
     KEV_J = 1.602180000000000e-16  # 1.6021766339999e-16 # keV to joules
+    J_KEV = 6.242e15  # joules to keV
     MASS_H = 1.6735575e-27  # Mass of a hydrogen atom in kilograms.
```
nkerman: Why this rounding of sig figs? To be specific:
1. If we have the precision in the comment, why not put it into the actual constant? 
2. Why the trailing 0s on `KEV_J`, implying further sig figs?
3. `KEV_J` has 6 sig figs, but its inverse `J_KEV` has only 4.

imap_processing/ultra/constants.py
```diff
@@ -61,6 +61,7 @@ class UltraConstants:
 
     # Conversion factors
     KEV_J = 1.602180000000000e-16  # 1.6021766339999e-16 # keV to joules
+    J_KEV = 6.242e15  # joules to keV
```
tmplummer: Is there any reason to define this rather than just using in `1/KEV_J`. At the least, it could be defined as `J_KEV = 1 / KEV_J`.

imap_processing/ultra/l1b/de.py
```diff
@@ -144,47 +151,49 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
         SpiceFrame.IMAP_DPS,
         SpiceFrame.IMAP_SPACECRAFT,
     )
 
-    de_dict["vx_sc"], de_dict["vy_sc"], de_dict["vz_sc"] = (
-        sc_velocity[:, 0],
-        sc_velocity[:, 1],
-        sc_velocity[:, 2],
+    de_dict["velocity_sc"] = np.column_stack(
+        (sc_velocity[:, 0], sc_velocity[:, 1], sc_velocity[:, 2])
```
tmplummer: I don't think that this column_stack call is doing anything.
```suggestion
    de_dict["velocity_sc"] = sc_velocity
```

imap_processing/ultra/l1b/de.py
```diff
@@ -144,47 +151,49 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
     )
-    de_dict["vx_dps_sc"], de_dict["vy_dps_sc"], de_dict["vz_dps_sc"] = (
-        sc_dps_velocity[:, 0],
-        sc_dps_velocity[:, 1],
-        sc_dps_velocity[:, 2],
+
+    de_dict["velocity_dps_sc"] = np.column_stack(
+        (sc_dps_velocity[:, 0], sc_dps_velocity[:, 1], sc_dps_velocity[:, 2])
```
tmplummer: Same as above. No need to column_stack these. It looks like values are already in the columns desired.

imap_processing/ultra/l1b/de.py
```diff
@@ -144,47 +151,49 @@ def calculate_de(de_dataset: xr.Dataset, name: str) -> xr.Dataset:
...
     )
-    de_dict["vx_dps_helio"], de_dict["vy_dps_helio"], de_dict["vz_dps_helio"] = (
-        helio_velocity[:, 0],
-        helio_velocity[:, 1],
-        helio_velocity[:, 2],
+
+    de_dict["velocity_dps_helio"] = np.column_stack(
+        (helio_velocity[:, 0], helio_velocity[:, 1], helio_velocity[:, 2])
```
tmplummer: Same as above. No need to column_stack these. It looks like values are already in the columns desired.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -437,20 +438,14 @@ def get_coincidence_positions(
     return etof, xc_array * 100
 
 
-def get_unit_vector(
+def get_de_velocity(
     front_position: tuple[NDArray, NDArray],
     back_position: tuple[NDArray, NDArray],
     d: np.ndarray,
     tof: np.ndarray,
 ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
```
tmplummer: IMHO, This would be more idiomatic if it returned an N x 3 array for N direct events. This makes is an array of Cartesian vectors and you don't have to carry each vector component around individually.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -186,74 +186,46 @@ path_length:
...
-  CATDESC: z-component of the velocity vector of the ena in the DPS frame at rest WRT spacecraft.
-  FIELDNAM: vz_dps_sc
-  LABLAXIS: vz dps sc
+  CATDESC: Estimated event efficiency for this path through the instrument.
+  FIELDNAM: event_efficiency
+  LABLAXIS: event efficiency
   UNITS: " "
 
-vx_dps_helio:
+velocity_sc:
```
tmplummer, line 211: Does this have a `DEPEND_1` for the second dimension?

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -213,40 +215,33 @@ def test_calculate_etof_xc(de_dataset, yf_fixture):
...
         (test_xb, test_yb),
         test_d,
         test_tof,
     )
-    # FSW test data should be negative and not have an analysis
-    # for negative tof values.
-    assert vhat_x[test_tof > 0] == pytest.approx(
-        -df_filt["vhatX"].iloc[ph_indices].astype("float").values[test_tof > 0],
+
+    assert v_x[test_tof > 0] == pytest.approx(
```
tmplummer: Is there a reason that you used `pytest.approx` in these tests rather than `numpy.testing.assert_allclose`?

tmplummer (commented):
Overall, looks great. I have a few comments of things that could be changed slightly.

imap_processing/ultra/l1b/ultra_l1b_extended.py
nkerman: I'd also lean towards a single N x 3 array.

imap_processing/ultra/constants.py
nkerman: I'd lean towards the second suggestion here (`J_KEV = 1 / KEV_J`), this defines J_KEV as a constant for clarity, but uses a single hardcoded value.

nkerman:
Looks good to me. Just had a couple questions about the constants, and I agree with Tim that it'd be good to replace the tuples of 1D arrays with a 2D array.

imap_processing/ultra/constants.py
demajistre: Sounds good to me.

imap_processing/ultra/l1b/de.py
demajistre: This is probably me just making sure python works for me. N00bish care.

imap_processing/ultra/l1b/de.py
demajistre: This is probably me just making sure python works for me. N00bish care.

imap_processing/ultra/l1b/de.py
demajistre: This is probably me just making sure python works for me. N00bish care.

imap_processing/ultra/l1b/ultra_l1b_extended.py
demajistre: Again, I'm not as familiar with standard practice.  I generally use 3XN because matrix math is more convenient- 
y=Bx
if B is 3x3 and x is 3xN, then Bx should be 3xN as well.  If x is Nx3, something needs to be transposed or re-ordered.
again - no idea how people do this stuff in standard practice.

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
demajistre: Only because I was making stuff up as I went along.

imap_processing/ultra/l1b/ultra_l1b_extended.py
tmplummer: @demajistre,
I'm far from an expert in performing matrix math in numpy but after reading the documents, the following code snippet is what I implemented in the geometry module for transforming vectors between frames using a rotation matrix. It might help understand the complexity of matrix multiplication with broadcasting using numpy. The `@` operator in this implementation is equivalent to `numpy.matmul`.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/geometry.py#L362

https://numpy.org/doc/stable/reference/generated/numpy.matmul.html

demajistre: Very good - I did not know about the broadcasting case, I was just forcing things into the form of the first example in the matmul documentation. You should implement it the way it works best (and the broadcasting seems perfectly reasonable).

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
laspsandoval: You are correct. Very nice catch.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -465,35 +460,29 @@ def get_unit_vector(
...
-    delta_x = front_position[0] - back_position[0]
-    delta_y = front_position[1] - back_position[1]
-
-    v_x = delta_x / tof
-    v_y = delta_y / tof
-    v_z = d / tof
+    # distances in .1 mm
+    delta_x = (front_position[0] - back_position[0]) * 0.1
+    delta_y = (front_position[1] - back_position[1]) * 0.1
+    delta_z = d * 0.1
```
tmplummer: This function would be considerably cleaner if you just defined a `delta_v` vector up front and use N x 3 arrays throughout.
```suggestion
    delta_v = np.empty((len(d), 3)  # Might want to specify a dtype here
    # distances in .1 mm
    delta_v[:, 0] = (front_position[0] - back_position[0]) * 0.1
    delta_v[:, 1] = (front_position[1] - back_position[1]) * 0.1
    delta_v[:, 2] = d * 0.1
```

tmplummer (approved):
LGTM. Just one comment on a potential cleanup that I saw. Take it or leave it.

PR #1249, opened 2025-01-07
Ultra L2: Calculate total exposure time

### LINKED ISSUE #1087: Ultra l2: Calculate the Total Exposure Time in the Spacecraft Frame for Nat

nkerman:
# Change Summary

## Overview
Create the basic structure of Ultra L2 processing and combine exposure time across pointings.

IMAP Ultra L2 Processing takes multiple L1C files, each with pointing sets containing counts, exposure times, sensitivities, etc. as a function of spatial coordinates (e.g. azimuth, elevation) and energy bins. The spatial coordinates are in different despun pointing set (`DPS`) frames for each different pointing (and thus each L1c file). The L2 pipeline projects these into an ecliptic, inertial frame (currently `J2000ECLIP` AKA `HAE`), where it combines counts, exposure times, etc. and calculates flux, etc.

- Currently, this code only combines exposure time. I removed the summed counts calculation, which had been in the draft PR (#1202), for a future PR. 
- This code is not currently disambiguating between the L1C spacecraft frame values and the L1C heliocentric frame values.
- Hopefully, the code mapping indices from DPS frames to HAE frames can eventually be reused or extracted to the shared mapper tool. 

Closes #1087.

## New Dependencies
None.

## New Files

- new file `imap_processing/ultra/l2/ultra_l2.py`
   - Main processing code for Ultra L2. 
   - Currently contains code to match indices from DPS frames (in L1C products) to a single HAE frame (for L2 output) and combine the exposure times across multiple DPS frames. 
- new file `imap_processing/tests/ultra/unit/test_ultra_l2.py`
   - Tests for functionality in `ultra_l2.py`

## Testing

- New tests for Ultra L2 index matching and exposure time summation.


imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+        )
+    )
+
+    # Project the grid points from the input frame to the projection frame
+    # radii_proj are cartesian (x,y,z) radii vectors in the projection frame
+    # corresponding to the grid points in the input frame
+    radii_proj = geometry.frame_transform(
+        et=[
+            event_time,
+        ],
```
tmplummer, line 151: According to the `frame_transform` docstring, you don't need to wrap this float in a list. Did you find otherwise?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+
+    # NOTE: This method of matching indices 1:1 is rectangular grid-focused
+    # It will not necessarily work with a tessellation (e.g. Healpix)
+    flat_indices_proj = np.ravel_multi_index(
+        multi_index=(input_az_in_proj_az_indices, input_el_in_proj_el_indices),
+        dims=(
+            int(360 // projection_frame_spacing_deg),
+            int(180 // projection_frame_spacing_deg),
+        ),
+        order="F",  # MATLAB uses Fortran order, which we replicate here
```
tmplummer, line 184: Maybe it won't matter because we will implement a companion binning function that uses the same ravel/reshape order but, I am concerned that this `order="F"` is hidden in the depths of this function and so a naive user might blindly use these indices without a `reshape(..., order="F")` and get the wrong results.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+    )
+
+    input_az_in_proj_az = input_grid_in_proj_spherical_coord[:, 1]
+    input_el_in_proj_el = input_grid_in_proj_spherical_coord[:, 2]
+
+    # Create bin edges for azimuth (0 to 2pi) and elevation (-pi/2 to pi/2)
+    az_bins = np.linspace(0, 2 * np.pi, int(360 / projection_frame_spacing_deg) + 1)
+    el_bins = np.linspace(
+        -np.pi / 2, np.pi / 2, int(180 / projection_frame_spacing_deg) + 1
+    )
```
tmplummer, line 170: In my mind, this really belongs as part of `build_az_el_grid_cached`

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+        az_grid_input_raveled,
+        el_grid_input_raveled,
+        input_az_in_proj_az,
+        input_el_in_proj_el,
+        input_az_in_proj_az_indices,
+        input_el_in_proj_el_indices,
+    )
+
+
+@typing.no_type_check
```
tmplummer, line 198: Why is this needed? Maybe a comment to justify it would be helpful?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+        raveled_l1c_exptime = l1c_prod.exposure_time.values.ravel(order="F")
+
+        # Correct but somewhat slow loop-based assignment of the exposure time values
+        for index_hae_proj, index_dps_input in zip(
+            matched_index_subdict["indices_hae_proj"],
+            matched_index_subdict["indices_dps_input"],
+        ):
+            pointing_projected_exptime[index_hae_proj] += raveled_l1c_exptime[
+                index_dps_input
+            ]
```
tmplummer, line 267: I believe that this can be done as a weighted sum similar to how it is done in `np.histogramdd`: https://github.com/numpy/numpy/blob/v2.1.0/numpy/lib/_histograms_impl.py#L1067

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+            indices_hae_proj,
+        ) = map_indices_frame_to_frame(
+            input_frame=geometry.SpiceFrame.IMAP_DPS,
+            projection_frame=geometry.SpiceFrame.ECLIPJ2000,
+            event_time=time,
+            input_frame_spacing_deg=l1c_spacing_deg,
+            projection_frame_spacing_deg=l2_spacing_deg,
+        )[:2]
+        product_index_map_dict["indices_hae_proj"] = indices_hae_proj
+        product_index_map_dict["indices_dps_input"] = indices_dps_input
```
tmplummer, line 368: I'm having trouble following this 'pull' and 'push' and what the indices variables contain.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+    # Build the combined exposure time arrays
+    (combined_exptime_45, combined_exptime_90, combined_exptime_total) = (
+        build_dps_combined_exposure_time(l1c_products, all_pointings_matched_indices)
+    )
+
+    ds_l2 = xr.Dataset(
+        {
+            "exposure": (
+                ["el", "az"],
+                spatial_utils.rewrap_even_spaced_el_az_grid(combined_exptime_total),
```
tmplummer, line 380: Is this any different than `combined_exptime_total.reshape((n_az, n_el), order="F")`

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+
+logger = logging.getLogger(__name__)
+logger.info("Importing ultra_l2 module")
+
+# Use lru_cache to wrap build_az_el_grid, allowing for efficient reuse of the same
+# az/el grid without passing through all functions
+build_az_el_grid_cached = lru_cache(maxsize=10)(spatial_utils.build_az_el_grid)
+
+
+def is_ultra45(
```
tmplummer, line 27: This doesn't seem like it is unique to L2. Does similar code exist elsewhere? This is great example of a function that could exist in it's own tiny PR.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+    projection_frame_spacing_deg : float, optional
+        The spacing of the projection frame grid in degrees,
+        by default DEFAULT_SPACING_DEG.
+
+    Returns
+    -------
+    tuple[NDArray]
+        Tuple of the following arrays:
+        - Flat indices of the grid points in the input frame.
+        - Flat indices of the grid points in the projection frame.
```
tmplummer, line 101: I am having trouble deciphering what these are. I think it would be more clear to use words like "indices that map input grid to projection grid" and add details like "size is equal to the number of pixels in the input frame grid".

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+    ].counts.shape
+    logger.info(
+        f"Number of DPS Pointings: {num_dps_pointings}. "
+        f"Shape is {num_el_bins_l1c} el x {num_az_bins_l1c} "
+        f"az x {num_energy_bins_l1c} energy"
+    )
+
+    # Setup the combined-across-pointings exp time arrays, in ecliptic inertial frame
+    combined_exptime_45 = np.zeros(num_el_bins_l1c * num_az_bins_l1c)
+    combined_exptime_90 = np.zeros(num_el_bins_l1c * num_az_bins_l1c)
```
tmplummer, line 244: I'm confused as to why these have the same number of elements as the input L1C files rather than the L2 grids.

imap_processing/ultra/l2/ultra_l2.py
tmplummer, line 1: My main concern is that this code doesn't really move toward the goal of having modular, shared L2 code between the ENA instruments. Perhaps I should start implementing some of the infrastructure for that shared code?

imap_processing/tests/ultra/unit/test_ultra_l2.py
```diff
@@ -0,0 +1,326 @@
...
+                f"imap_ultra_l1c_{head}sensor-pset_{timestr[:4]}"
+                f"{timestr[5:7]}{timestr[8:10]}-repointNNNNN_vNNN"
+            )
+        },
+    )
+
+    return pset_product
+
+
+class TestUltraL2:
```
tmplummer, line 112: I have some concerns about the scope of some of these tests. It would be nice to break out some smaller functions so that unit level tests can be written rather than all of these tests that I would consider integration level tests.

imap_processing/tests/ultra/unit/test_ultra_l2.py
```diff
@@ -0,0 +1,326 @@
...
+            "indices_dps_proj": indices_dps_proj,
+            "indices_hae_input": indices_hae_input,
+            "indices_hae_proj": indices_hae_proj,
+        }
+        self.all_dps_index_map_dict_simple = {
+            float(prod.epoch.values): pointing_indices_map_dict
+            for prod in self.l1c_products
+        }
+
+    def test_is_ultra45(self, caplog):
```
tmplummer, line 160: Is anything gained by having this test in the `TestUltraL2` class? It seems like a totally independent function to me.

imap_processing/tests/ultra/unit/test_ultra_l2.py
```diff
@@ -0,0 +1,326 @@
...
+
+        np.testing.assert_equal(combined_exptime_45[0], total_exposure_time_45)
+        np.testing.assert_array_equal(combined_exptime_45[1:], 0)
+
+        np.testing.assert_equal(combined_exptime_90[0], total_exposure_time_90)
+        np.testing.assert_array_equal(combined_exptime_90[1:], 0)
+
+    @pytest.mark.external_kernel()
+    @pytest.mark.use_test_metakernel("imap_ena_sim_metakernel.template")
+    def test_map_indices_frame_to_frame_same_frame(self):
```
tmplummer, line 226: By mocking the `frame_transform` function to return the input coordinates, I *think* you could remove the SPICE dependency.

imap_processing/tests/ultra/unit/test_ultra_l2.py
```diff
@@ -0,0 +1,326 @@
...
+            input_el_in_proj_el.reshape(
+                (180 // spacing_deg, 360 // spacing_deg), order="F"
+            ),
+            ultra_l2.build_az_el_grid_cached(spacing=np.deg2rad(spacing_deg))[3],
+        )
+
+    @pytest.mark.external_kernel()
+    @pytest.mark.use_test_metakernel("imap_ena_sim_metakernel.template")
+    @pytest.mark.parametrize("input_spacing_deg", [1 / 4, 1 / 3, 1 / 2, 1])
+    def test_map_indices_frame_to_frame_different_spacings(self, input_spacing_deg):
```
tmplummer, line 270: Same comment as above... I think you can easily remove the SPICE dependency from this test.

tmplummer (commented):
Lots of progress here with some good chunks of code that can/should be extracted into common L2 functions. I have a bit of a fear of saying "Hopefully, there is code that can be extracted from this..." and like to encourage a more proactive approach of "Hey, there is code that others can use here, I should extract it into common functions."
It might be a good idea to discuss how we should move forward with a more collaborative approach to L2.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Removed the list - I'm not sure why I had done this, maybe because of the previous version of `frame_transform`.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Ah thanks for calling this out! Looking back at it, I managed to fix the underlying thing Mypy didn't like and removed this.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+
+def map_indices_frame_to_frame(
+    input_frame: geometry.SpiceFrame,
+    projection_frame: geometry.SpiceFrame,
+    event_time: float,
+    input_frame_spacing_deg: float = DEFAULT_SPACING_DEG,
+    projection_frame_spacing_deg: float = DEFAULT_SPACING_DEG,
+) -> tuple[NDArray, NDArray, NDArray, NDArray, NDArray, NDArray, NDArray, NDArray]:
+    """
+    Match flattened rectangular grid indices between two frames.
```
laspsandoval, line 79: This is good. I am wondering if you could make it also match 1D grid indices? I recall that Lo and Hi are not using az,el and instead spin phase. Something to discuss at our next ENA meeting I think.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+
+# Set default values:
+DEFAULT_SPACING_DEG = 0.5
+
+logger = logging.getLogger(__name__)
+logger.info("Importing ultra_l2 module")
+
+# Use lru_cache to wrap build_az_el_grid, allowing for efficient reuse of the same
+# az/el grid without passing through all functions
+build_az_el_grid_cached = lru_cache(maxsize=10)(spatial_utils.build_az_el_grid)
```
laspsandoval, line 24: I've never seen this used before. Is it necessary if it is only used in a single function? Just a question because I am curious.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+            ]
+
+        if is_ultra45(l1c_prod):
+            combined_exptime_45 += pointing_projected_exptime
+        else:
+            combined_exptime_90 += pointing_projected_exptime
+
+    combined_exptime_total = combined_exptime_45 + combined_exptime_90
+
+    return (combined_exptime_45, combined_exptime_90, combined_exptime_total)
```
laspsandoval, line 276: Remove redundant parenthesis.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,409 @@
...
+        centered_elevation=True,  # (-pi/2, pi/2 rad = -90, 90 deg)
+    )
+
+    # Unwrap the grid to a 1D array for same interface as tessellation
+    az_grid_input_raveled = az_grid_in.ravel(order="F")
+    el_grid_input_raveled = el_grid_in.ravel(order="F")
+
+    # Get the flattened indices of the grid points in the input frame (Fortran order)
+    # TODO: Discuss with Nick/Tim/Laura: Should this just be an arange?
+    flat_indices_in = np.ravel(
```
laspsandoval, line 130: I think so. Im not certain why it would need to use Fortran ordering except maybe to match test results provided by Nick?

imap_processing/tests/ultra/unit/test_ultra_l2.py
```diff
@@ -0,0 +1,326 @@
...
+        ) = ultra_l2.map_indices_frame_to_frame(
+            input_frame=geometry.SpiceFrame.IMAP_DPS,
+            projection_frame=geometry.SpiceFrame.IMAP_DPS,
+            event_time=et,
+            input_frame_spacing_deg=input_spacing_deg,
+            projection_frame_spacing_deg=proj_spacing_deg,
+        )[:2]
+
+        # The input indices should still be the same length as the projection indices
+        assert len(flat_indices_in) == len(flat_indices_proj)
```
laspsandoval, line 286: Should it be the same length or just the same?

laspsandoval (commented):
Nice first pass! I do think we should talk strategies about how we should make this able to be used by Hi/Lo at the next ENA meeting.

imap_processing/ultra/l2/ultra_l2.py
nkerman: That makes sense since it's similar to getting the bin centers, as already done by `build_az_el_grid`. At present, I had only been calling `build_az_el_grid` for the input frame, and this actually would need the projection frame's spacing, so it will be a separate  call to `build_az_el_grid`

I'll add that into the version of this function I'm spinning out into the ENA Mapping Util PR we discussed today on Zoom.

imap_processing/ultra/l2/ultra_l2.py
nkerman: They're the same in this case, the function is just meant to be capable of rewrapping any unwrapped grid with minimal info. 

The benefits of `spatial_utils.rewrap_even_spaced_el_az_grid` are that it can rewrap an evenly spaced grid without being explicitly handed the number of az, el bins each time, and that it can handle if there's an extra axis (e.g. energy bins).

imap_processing/tests/ultra/unit/test_ultra_l2.py
nkerman: Huh, I'm just not sure where else `is_ultra45` and its test should go? It didn't seem to fit in anywhere obvious already, and this is a small and purpose-built utility function. 

Are you envisioning a new file like `imap_processing/ultra/utils/general.py`?

imap_processing/tests/ultra/unit/test_ultra_l2.py
nkerman: Ah, I think I see what you mean. I guess I felt it was safer to test that it works with frame_transform, but I understand that's more of an integration test, not a unit test.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Added clarification to docs in the PR I'm about to open!

imap_processing/ultra/l2/ultra_l2.py
nkerman: Yes! np.bincount seems to be a more efficient way of doing this. I've changed it in the upcoming `match_rect_indices_frame_to_frame` PR.

imap_processing/tests/ultra/unit/test_ultra_l2.py
nkerman: Done! I used your recent PR as a template to mock frame_transform - thanks!

imap_processing/tests/ultra/unit/test_ultra_l2.py
nkerman: Done! I used your recent PR as a template to mock frame_transform - thanks!

imap_processing/ultra/l2/ultra_l2.py
nkerman: We talked about this in person after the comment was left, where in the code is a good spot to describe the difference between push/pull? I'm thinking in the `bin_values` function?

imap_processing/ultra/l2/ultra_l2.py
nkerman: Tim and I spoke in person about how to do this better. I'm about to open 2 PRs extracting and generalizing the two main functions needed here:
1. `match_rect_indices_frame_to_frame` will replace `map_indices_frame_to_frame`. 
2. `bin_values_at_indices` or a similarly named function will take those matched indices and sum over them.

nkerman:
Closing old draft PRs, superceded by merged PRs.

PR #1250, opened 2025-01-08
WIP: Minimal updates for poetry 2.0 

maxinelasp:
# Change Summary
Updates to make imap_processing compatible with Poetry 2.0.


.pre-commit-config.yaml
```diff
@@ -27,11 +27,10 @@ repos:
       - id: codespell
         files: ^.*\.(py|md|rst|yml)$
   - repo: https://github.com/python-poetry/poetry
-    rev: '1.8.0'  # add version here
+    rev: '2.0.0'  # add version here
```
greglucas, line 30:
```suggestion
    rev: '2.0.1'  # add version here
```

.pre-commit-config.yaml
```diff
@@ -27,11 +27,10 @@ repos:
...
     hooks:
       - id: poetry-check
       - id: poetry-lock
-        args: [--no-update]
```
greglucas, line 34: Does this mean we'll get a lot more updates now in PRs? Is this the new default version?

pyproject.toml
```diff
@@ -106,7 +112,7 @@ imap_xtce = 'imap_processing.ccsds.excel_to_xtce:main'
 ignore-words-list = "livetime"
 
 [tool.poetry-dynamic-versioning]
-enable = true
+enable = false
```
greglucas, line 115: This should be reverted I think. When I checked this out I got a 0.0.0 version identifier installed. But enabling the dynamic versioning I then got the new hashed version.
```suggestion
enable = true
```

.pre-commit-config.yaml
maxinelasp: The new version does `--no-update` by default. So we will get the same behavior as before.

pyproject.toml
maxinelasp: Oh yeah, definitely. This PR is a hotfix type suggestion and I was having some minor issues with the dynamic versioning when testing it. But, I got a reply from the plugin maintainer about this which makes using it a lot simpler, so we can do it the new way instead https://github.com/mtkennerly/poetry-dynamic-versioning/issues/197#issuecomment-2576291495

greglucas:
Closing this because it is out of date and it will require significant work to actually update so should be done in a separate PR.

Issue #770, opened 2024-08-21
CoDICE: Some specific CDF attributes/variables to add to L1 data products


bourque:
The following CDF attributes need to be added to the L1 data products: `SPIN_PERIOD`, `ST_BIAS_GAIN_MODE`, `SW_BIAS_GAIN_MODE`, `RGFO_HALF_SPIN`, `NSO_HALF_SPIN`, and `DATA_QUALITY`. The values for these should be present in the packet data.

bourque:
Here specifically are the data products and the CDF variables that are missing:

- **`hi-counters-aggregated`**: `DCR`, `STO`, `SPO`, `Reserved1`, `MST`, `Reserved2`, `Reserved3`, `Reserved4`, `Reserved5`, `LowTOFCutoff`, `Reserved6`, `Reserved7`, `ASIC1FlagInvalid`, `ASIC2FlagInvalid`, `ASIC1ChannelInvalid`, `ASIC2ChannelInvalid`
- **`hi-counters-singles`**: `SSDID`
- **`hi-omni`**: Energy tables (see #809), `DataQuality`, `SpinPeriod`
- **`hi-sectored`**: Energy tables (see #809), `DataQuality`, `SpinPeriod`, `SpinSectorIndex`, `SSDIndex`
- **`lo-counters-aggregated`**:  `SpinSectorPairs`, `RGFOHalfSpin`, `NSOHalfSpin`, `SWBiasGainMode`, `STBiasGainMode`, `DataQuality`, `SpinPeriod`, `TCR`, `DCR`, `TOFPlusAPD`, `TOFOnly`, `PositionPlusAPD`, `PositionOnly`, `STAorSTBPlusAPD`, `STAorSTBOnly`, `Reserved1`, `Reserved2`, `SPOnly`, `APDOnly`, `LowTOFCutoff`, `STA`, `STB`, `SP`, `TotalPositionCount`, `InvalidPositionCount`, `ASIC1FlagInvalid`, `ASIC2FlagInvalid`, `ASIC1ChannelInvalid`, `ASIC2ChannelInvalid`
- **`lo-counters-singles`**: `PositionIndex`, `SpinSectorPairs`, `RGFOHalfSpin`, `NSOHalfSpin`, `SWBiasGainMode`, `STBiasGainMode`, `DataQuality`, `SpinPeriod`
[...]

PR #1254, opened 2025-01-08
CoDICE lo-(n)sw-angluar validation

### LINKED ISSUE #1192: CoDICE: Set up initial structure for L1A validation
The goal is to write unit test(s) that compare the values of the data arrays, variable names, attributes, etc. between the CDFs that I generate and the validation CDFs that Joey gave me.

bourque:
# Change Summary

## Overview
This PR contains changes to get the CoDICE `lo-nsw-angular` and `lo-sw-angular` data products in a state to be validated. More specifically, it adds more CDF data variables that are needed as support data (see #770) and adds a unit test for comparing data array values produced by the CoDICE pipeline against the validation CDFs (see #1192). This sets up a framework to validate other CoDICE data products down the line, but for now this PR just focuses on the `lo-(n)sw-angular` products.

Closes #1192 and #770

## New Files

- `imap_processing/tests/codice/data/validation/imap_codice_l1a_lo-sw-angular_20241110193700_v0.0.0.cdf`
- `imap_processing/tests/codice/data/validation/imap_codice_l1a_lo-nsw-angular_20241110193700_v0.0.0.cdf`
   - Added validation CDFs 

## Updated Files

  - Added CDF attribute definitions for the new support variables, and made a few other small tweaks to existing ones 
  - Added more support CDF variables
-  `imap_processing/codice/constants.py`
  - Now using all lowercase data variable names for consistency
-  `imap_processing/tests/codice/conftest.py`
  - Moved validation CDFs to their own directory 
   - Properly implemented test for validating data arrays (`test_l1a_data_array_values`), and other minor changes


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -80,27 +82,129 @@ counters_attrs: &counters
...
     SI_CONVERSION: " > "
     UNITS: eV
     VALIDMIN: 1
     VALIDMAX: 14100
     VAR_TYPE: support_data
 
+nso_half_spin:
+    <<: *default
+    CATDESC: Indicates the point when No Scan Operation (NSO) was activated. In NSO, the ESA voltage is set to the first step in the scan and remains fixed until the next cycle boundary.
+    DEPEND_0: epoch
```
maxinelasp: Why not put DEPEND_0, DISPLAY_TYPE, FILLVAL, and FORMAT in another shared group (similar to default) since they're the same through different variables?

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -80,27 +82,129 @@ counters_attrs: &counters
...
+    SCALETYP: linear
+    SI_CONVERSION: " > "
+    UNITS: half spin number
+    VALIDMIN: 0
+    VALIDMAX: 255
+    VAR_TYPE: data
+
+rgfo_half_spin:
+    <<: *default
+    CATDESC: Indicates the point when Reduced Gain Factor Operation (RGFO) was activated. In RGFO, the Entrance ESA voltage is reduced in order to limit the number of ions that reach the detectors.
```
maxinelasp: I think this CATDESC is too long. It should be something like 70 chars.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -80,27 +82,129 @@ counters_attrs: &counters
...
+    SCALETYP: linear
+    SI_CONVERSION: " > "
+    UNITS: s
+    VALIDMIN: 0.0
+    VALIDMAX: 16.0
+    VAR_TYPE: data
+
+st_bias_gain_mode:
+    <<: *default
+    CATDESC: Indicates whether FSW is tracking the Suprathermal High-Gain bias curve or the Suprathermal Low-Gain bias curve.
```
maxinelasp: May be too long. I think "Indicates whether" is superfluous.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -80,27 +82,129 @@ counters_attrs: &counters
...
+
+sw_bias_gain_mode:
+    <<: *default
+    CATDESC: Indicates whether FSW is tracking the Solarwind High-Gain bias curve or the Solarwind Low-Gain bias curve.
+    DEPEND_0: epoch
+    DISPLAY_TYPE: " "
+    FIELDNAM: Solarwind Bias Gain Mode
+    FILLVAL: 255
+    FORMAT: I1
+    LABLAXIS: " "
```
maxinelasp: Why not fill in LABLAXIS?

imap_processing/codice/codice_l1a.py
```diff
@@ -229,9 +230,39 @@ def define_support_variables(self, dataset: xr.Dataset) -> xr.Dataset:
...
-                # TODO: Need to implement methods to gather and set other
-                #       support attributes
+            # These variables can be gathered from the packet data
+            elif variable_name == "rgfo_half_spin":
+                variable_data = self.dataset.rgfo_half_spin.data
+                dims = ["epoch"]
+                attrs = self.cdf_attrs.get_variable_attributes("rgfo_half_spin")
+
+            elif variable_name == "nso_half_spin":
+                variable_data = self.dataset.nso_half_spin.data
```
maxinelasp: Could you remove these if statements? 

You have almost all the info you need, except for a special case in data_quality which can be one if statement.

```
for variable_name in self.config["support_variables"]:
  variable_data = self.dataset[variable_name].data
  dims = ["epoch"]
  attrs = self.cdf_attrs.get_variable_attributes(variable_name)
  dataset[variable_name] = xr.DataArray(variable_data, dims = dims, attrs = attrs)
```

imap_processing/codice/constants.py
```diff
@@ -278,10 +278,10 @@
         "support_variables": [
             "energy_table",
```
maxinelasp, line 279: Could these support variables get pulled out into a shared variable? Then you can add any extras but remove some of these duplicate lines

imap_processing/tests/codice/conftest.py
```diff
@@ -21,20 +21,24 @@
...
-    TEST_DATA_PATH / "validation_lo-counters-singles.cdf",
-    TEST_DATA_PATH / "validation_lo-sw-angular.cdf",
-    TEST_DATA_PATH / "validation_lo-nsw-angular.cdf",
-    TEST_DATA_PATH / "validation_lo-sw-priority.cdf",
-    TEST_DATA_PATH / "validation_lo-nsw-priority.cdf",
-    TEST_DATA_PATH / "validation_lo-sw-species.cdf",
-    TEST_DATA_PATH / "validation_lo-nsw-species.cdf",
-    TEST_DATA_PATH / "validation_lo-pha.cdf",
-]
+    TEST_DATA_PATH / "validation" / "imap_codice_l1a_hi-ialirt_20241110193700_v0.0.0.cdf",
```
maxinelasp, line 25: Are there any files in this folder that are not part of this VALIDATION_DATA variable? Could you automatically populate them?

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -172,14 +189,9 @@ def test_l1a_num_variables(test_l1a_data, index):
     assert len(processed_dataset) == EXPECTED_NUM_VARIABLES[index]
 
 
-@pytest.mark.skip("Awaiting validation data")
```
maxinelasp, line 175: Yay validation data!

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: Ah yes, good catch. The suggested limit by SPDF is 80 characters. I forgot that I can use `VAR_NOTES` instead for this!

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: I am taking the language here from what Joey is providing me in the validation data so I am inclined to keep it, but yes good point.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: I think I have a bunch of instances of these because I need to come up with these myself (they are not defined in the validation data). This is a good opportunity to fill these in, so I will do that!

imap_processing/codice/constants.py
bourque: I am torn on this. CoDICE has 13 data products, some of which have overlapping support variables, some don't, some have completely different ones. I could introduce shared variables for _some_ of them which would reduce some repeated lines. On the other hand, I could just explicitly define each of the 13 data products (which I would argue is easier to lookup/read when you are just interested in one particular data product configuration). Pros and cons to both ways I think, and I don't know what to do 😅

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: For the same reasons I mention in the comment for `constants.py`, I am torn on this as well -- being explicit vs introducing some shared variables.

imap_processing/tests/codice/conftest.py
bourque: Good call, now that I have put them in their own folder it will be easy to auto-populate this list.

bourque: On second thought, it might be a bit tricky, as I want to preserve a particular order to the list of files (and have that list correspond to the test L1a CDF files). I could do everything by alphabetical order, but it would change a lot of the ordering in the test code. I may address this in a future PR.

imap_processing/codice/constants.py
maxinelasp: How often do they get updated? If they really rarely get changed, then it's probably fine to leave them all. Or you could only define one set of shared variables and leave the rest as duplicates.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
maxinelasp: I tend to prefer shared variables, but it can also be case-by-case.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -139,11 +135,12 @@ nso_half_spin:
     UNITS: half spin number
     VALIDMIN: 0
     VALIDMAX: 255
+    VAR_NOTES: Indicates the point when No Scan Operation (NSO) was activated. In NSO, the ESA voltage is set to the first step in the scan and remains fixed until the next cycle boundary.
     VAR_TYPE: data
 
 rgfo_half_spin:
     <<: *default
-    CATDESC: Indicates the point when Reduced Gain Factor Operation (RGFO) was activated. In RGFO, the Entrance ESA voltage is reduced in order to limit the number of ions that reach the detectors.
+    CATDESC: See VAR_NOTES
```
maxinelasp: I don't think the ISTP people will go for this 😆 but worth a try!

maxinelasp (approved):
Looks good, thanks for making those changes

imap_processing/codice/constants.py
bourque: They should be set in stone at this point (if they are not I will be mad 😅)

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: Im really tempted to write some hidden messages in these CATDESCs

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -10,14 +10,10 @@
 default_attrs: &default
     CATDESC: " "
     DISPLAY_TYPE: no_plot
-    FIELDNAM: ""
+    FIELDNAM: " "
```
tech3371, line 5: Since this field is required, you may want to remove that space to allow cdflib ISTP checker or SAMMI to catch it just in case you forgot to set it below.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
tech3371: I agree with Maxine on putting `DEPEND_0` in default since the `VAR_TYPE: data` and it will always require epoch dimension. Looks like you did shared attrs for counters attrs and some other. Is there reason for these new variables to be explicit? Just curious.

tech3371 (commented):
On quick glance, it looks good to me. I haven't looked at tests change yet.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: If I put DEPEND_0=epoch, DISPLAY_TYPE=time_series, FILLVAL=255, FORMAT=I3, and VAR_TYPE=data as a shared collection of attributes, these would apply to `nso_half_spin` and `rgfo_half_spin` variables only, which is only two of the eight (what I am calling "support") data variables. I feel like introducing shared attributes to only two of eight variables is more confusing than just explicitly defining the eight variables.

bourque: Good point though Tenzin, I can at least put `DEPEND_0` in the `default` attrs since it will always be required for `VAR_TYPE=data`.

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -189,16 +201,34 @@ def test_l1a_data_array_values(test_l1a_data: xr.Dataset, index):
...
+            constants, f'{descriptor.upper().replace("-","_")}_VARIABLE_NAMES'
+        )
+        processed_dataset = test_l1a_data[index]
+        validation_dataset = load_cdf(VALIDATION_DATA[index])
+
+        # Joey says that the shape of the data arrays (i.e. how they are
+        # arranged) does not need to follow a specific order, and the SDC can
+        # decide to arrange them how we see fit. As such, the shape of the CDFs
+        # that SDC produces may not match the validation data. To get around
+        # this, compare the sum of the values of the data arrays.
```
vmartinez-cu, line 224: Comparing the sum of the values is a nice solution

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -189,16 +201,34 @@ def test_l1a_data_array_values(test_l1a_data: xr.Dataset, index):
...
+        for counter in counters:
             np.testing.assert_array_equal(
-                validation_dataset[variable].data, generated_dataset[variable].data[0]
+                getattr(processed_dataset, counter).data.sum(),
+                getattr(validation_dataset, counter).data.sum(),
             )
+    else:
+        pytest.xfail(f"Still need to implement validation for {descriptor}")
```
vmartinez-cu, line 231: This is a nice way to skip descriptors that aren't ready to be validated. I think I could use pytest.xfail for some of my tests so it's great to see this example in your code

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -7,13 +7,35 @@
...
+    "lo-nsw-angular",
+    "hi-counters-aggregated",
+    "hi-counters-singles",
+    "hi-omni",
+    "hi-sectored",
+    "hi-priority",
+    "lo-pha",
+    "hi-pha",
+]
+
```
vmartinez-cu, line 38: I like how you organized these definitions in your test file. It's easy to find and use in your tests

imap_processing/codice/constants.py
```diff
@@ -341,10 +341,10 @@
...
             "acquisition_time_per_step",
-            "RGFO_half_spin",
-            "NGO_half_spin",
-            "SW_bias_gain_mode",
-            "ST_bias_gain_mode",
+            "rgfo_half_spin",
+            "nso_half_spin",
+            "sw_bias_gain_mode",
+            "st_bias_gain_mode",
```
vmartinez-cu, line 347: good idea to make these lower case for consistency

vmartinez-cu (approved):
LGTM!

tech3371 (approved):

PR #1262, opened 2025-01-14
CoDICE: Calculate epoch values from acq_start_seconds instead of shcoarse

bourque:
# Change Summary

## Overview
This PR adds a function to calculate epoch values from `acq_start_seconds` field instead of the default `shcoarse` field, as per request from CoDICE instrument team.  The function may also be expanded in the future to support other time keys if needed.

## Updated Files


imap_processing/codice/codice_l1a.py
```diff
@@ -84,6 +87,22 @@ def __init__(self, table_id: int, plan_id: int, plan_step: int, view_id: int):
...
+
+        On CoDICE, the epoch values are derived from the `acq_start_seconds`
+        field in the packet.
+
+        Returns
+        -------
+        epoch : NDArray[int]
+            List of epoch values.
+        """
+        epoch = met_to_j2000ns(self.dataset.acq_start_seconds)
```
tmplummer: Just to double check... `acq_start_seconds` records the spacecraft clock coarse ticks (seconds) at acquisition start time and there is no sub-seconds field? If both of those things are true, this looks like the correct implementation to me.

tmplummer (approved):
I just have the one double check comment. Otherwise this is good to go!

imap_processing/codice/codice_l1a.py
bourque: After discussing with Joey, there is indeed a sub-seconds field that I will have to implement, so I will update that. `acq_start_seconds` does indeed record the S/C clock coarse ticks at acquisition time.

bourque: @subagonsouth Could you give this another look any make sure I am thinking about this correctly? It is a bit tricky since `acq_start_seconds` is in units of seconds and `acq_start_subseconds` is in units of microseconds, so I think I need to do a small conversion before calculating `epoch`.

imap_processing/codice/codice_l1a.py
```diff
@@ -84,6 +87,32 @@ def __init__(self, table_id: int, plan_id: int, plan_step: int, view_id: int):
...
+        On CoDICE, the epoch values are derived from the `acq_start_seconds`
+        field in the packet.
+
+        Returns
+        -------
+        epoch : NDArray[int]
+            List of epoch values.
+        """
+        # Convert the acq_start_subseconds field from microseconds to seconds
+        subseconds = [value / 1000000 for value in self.dataset.acq_start_subseconds]
```
greglucas: I tried to see what type `self.dataset` is and couldn't find where `dataset` is set because it isn't in the `__init__` function. Are we guaranteed to have a `self.dataset` attribute?

The reason I was wondering about the type is because I think this function could be simplified to directly work on the arrays of values rather than going between lists and numpy functions.

```python
return met_to_j2000ns(self.data.acq_start_seconds + self.dataset.acq_start_subseconds / 1e6)

bourque: Thanks @greglucas . To answer your questions, the type of `self.dataset` here is `xarray.core.dataset.Dataset` (I believe it is unchanged since being returned from `utils.packet_file_to_datasets`. Yes, we are guaranteed to have a `self.dataset` attribute at that point in the code.

Thanks for pointing that out, it is indeed simpler! I'll make that change.

greglucas (approved):

PR #1264, opened 2025-01-14
ENA Maps: Bin values at indices

### LINKED ISSUE #1263: L2 ENA Maps: Bin arrays based on matched indices
### Algorithm Description:

Once data value indices have been matched between two frames _(e.g. from an L1C DPS frame to the HAE frame in which the L2 map will be projected)_, a function must "bin" from the first input frame to the second output frame. "Binning" here means taking values from the indices in `frame 1`, and assigning them to the matched index in `frame 2`. 

### Requirements:

- There may be a non- `1:1` correspondence between the indices in frame 1 and the indices in frame 2. That is, multiple indices in one frame may map onto the same index in another frame. 
- Not all possible indices in each frame are guaranteed to be used. This can mean that not all input values are projected into the output frame. It can also mean that not all output frame indices receive any values from the input frame. 

#### Input data:
- input frame indices (ordered list)
- output frame indices (ordered list)
- input values to be binned
  - Allow multiple values to be binned on the same indices\* 
  - Must be at least as long on the axis to be indexed as input frame indices
  - May have an "extra axis" - e.g. a non-spatially-binned dimension like energy. Each bin of this axis will be treated independently. 
- Shape of output frame grid, or the number of elements in it, or some other way of determining how large the output value array must be. 
  - Must be at least as large as the output frame indices

\* Suggest 2 functions: one to bin a single array, and another to loop through a dictionary of: 
`{"array_name": NDarray}` and bin each. 

### Related Issues/PRs:
#1258 
#1260

nkerman:
# Change Summary

## Overview

Once data value indices have been matched between two frames _(e.g. from an L1C DPS frame to the HAE frame in which the L2 map will be projected)_, a function must "bin" from the first input frame to the second output frame. "Binning" here means taking values from the indices in `frame 1`, and assigning them to the matched index in `frame 2`. 

### Requirements:

- There may be a non- `1:1` correspondence between the indices in frame 1 and the indices in frame 2. That is, multiple indices in one frame may map onto the same index in another frame. 
- Not all possible indices in each frame are guaranteed to be used. This can mean that not all input values are projected into the output frame. It can also mean that not all output frame indices receive any values from the input frame. 
- Values are effectively summed together within a bin. 

Closes #1263 

## New Files

- imap_processing/ena_maps/map_utils.py
   - Adds ENA Map util functions for binning values on indices
     - Note this file is also added in #1257 , with other util functions which will create the indices needed for binning. 
  - Tests of above

## Testing

- 7 New unique unit tests of the binning functionality
  - 33 counting parameterizations
  - Tested against manually generated values and (paper + pencil) and programatically (loop through indices in a more explicit way from how the algorithm uses bincount) calculated values.

### DIFF
--- a/imap_processing/ena_maps/map_utils.py
+++ b/imap_processing/ena_maps/map_utils.py
@@ -47,16 +47,12 @@ def bin_single_array_at_indices(
     num_projection_indices = np.prod(projection_grid_shape)
 
     if value_array.ndim == 1:
-        extra_axis = False
-    elif value_array.ndim == 2:
-        extra_axis = True
-    else:
-        raise NotImplementedError(
-            "Only 1D and 2D arrays are supported for binning. "
-            f"Received array with shape {value_array.shape}."
+        binned_values = np.bincount(
+            projection_indices,
+            weights=value_array[input_indices],
+            minlength=num_projection_indices,
         )
-
-    if extra_axis:
+    elif value_array.ndim == 2:
         # Apply bincount to each row independently
         binned_values = np.apply_along_axis(
             lambda x: np.bincount(
@@ -67,14 +63,11 @@ def bin_single_array_at_indices(
             axis=0,
             arr=value_array,
         )
-
     else:
-        binned_values = np.bincount(
-            projection_indices,
-            weights=value_array[input_indices],
-            minlength=num_projection_indices,
+        raise NotImplementedError(
+            "Only 1D and 2D arrays are supported for binning. "
+            f"Received array with shape {value_array.shape}."
         )
-
     return binned_values
 
 
--- a/imap_processing/ena_maps/map_utils.py
+++ b/imap_processing/ena_maps/map_utils.py
@@ -1,0 +1,89 @@
+"""Utilities for generating ENA maps."""
+
+from __future__ import annotations
+
+import logging
+
+import numpy as np
+from numpy.typing import NDArray
+
+logger = logging.getLogger(__name__)
+
+
+def bin_values_at_indices(
+    input_values_to_bin: dict[str, NDArray],
+    input_indices: NDArray,
+    projection_indices: NDArray,
+    projection_grid_shape: tuple[int, int],
+) -> dict[str, NDArray]:
+    """
+    Project values from input grid to projection grid based on matched indices.
+
+    Parameters
+    ----------
+    input_values_to_bin : dict[str, NDArray]
+        Dict matching variable names to arrays of values to bin.
+    input_indices : NDArray
+        Ordered indices for input grid, corresponding to indices in projection grid.
+        1 dimensional. May be non-unique, depending on the projection method.
+    projection_indices : NDArray
+        Ordered indices for projection grid, corresponding to indices in input grid.
+        1 dimensional. May be non-unique, depending on the projection method.
+    projection_grid_shape : tuple[int, int]
+        The shape of the grid onto which values are projected (rows, columns).
+        This size of the resulting grid (rows * columns) will be the size of the
+        projected values contained in the output dictionary.
+
+    Returns
+    -------
+    dict[str, NDArray]
+        Dict matching the input variable names to the binned values
+        on the projection grid.
+
+    Raises
+    ------
+    ValueError
+        If the input indices are not 1D arrays with the same number of elements.
+    ValueError
+        If the number of i
+    """
+    # Both sets of indices must be 1D with the same number of elements
+    if input_indices.ndim != 1 or projection_indices.ndim != 1:
+        raise ValueError(
+            "Indices must be 1D arrays. "
+            "If using a rectangular grid, the indices must be unwrapped."
+        )
+    if input_indices.size != projection_indices.size:
+        raise ValueError(
+            "The number of input and projection indices must be the same. "
+            f"Received {input_indices.size} input indices and \
+                         {projection_indices.size} projection indices."
+        )
+
+    num_projection_indices = np.multiply(*projection_grid_shape)
+
+    binned_values_dict = {}
+    for value_name, value_array in input_values_to_bin.items():
+        if value_array.size != input_indices.size:
+            # TODO: This assumption doesn't hold for 'Pull' method. Re-evaluate.
+            # If the same size on axis 0, but values has extra axis (e.g. energy bin),
+            # we can carry the extra axis through, adding each bin separately
+            if value_array.shape[0] == input_indices.size:
+                # TODO: Allow for extra axis (e.g. energy bins)
+                # which is summed separately
+                extra_axis = True
+                logger.info(f"Extra axis detected for {value_name}.")
+            else:
+                raise ValueError(
+                    f"Value array for {value_name} "
+                    "must have the same size as the indices."
+                )
+        if extra_axis:
+            # TODO: how can we handle the bincount in case of extra axis?
+            pass
+        binned_values = np.bincount(
+            projection_indices, weights=value_array, minlength=num_projection_indices
+        )
+        binned_values_dict[value_name] = binned_values
+
+    return binned_values_dict
--- a/imap_processing/ena_maps/map_utils.py
+++ b/imap_processing/ena_maps/map_utils.py
@@ -45,11 +45,27 @@ def bin_single_array_at_indices(
 
     Raises
     ------
+    ValueError
+        If the input and projection indices are not 1D arrays
+        with the same number of elements.
     NotImplementedError
-        If the input value_array has more than 2 dimensions.
+        If the input value_array has dimensionality less than 1.
     """
     if input_indices is None:
         input_indices = np.arange(value_array.shape[0])
+
+    # Both sets of indices must be 1D with the same number of elements
+    if input_indices.ndim != 1 or projection_indices.ndim != 1:
+        raise ValueError(
+            "Indices must be 1D arrays. "
+            "If using a rectangular grid, the indices must be unwrapped."
+        )
+    if input_indices.size != projection_indices.size:
+        raise ValueError(
+            "The number of input and projection indices must be the same. \n"
+            f"Received {input_indices.size} input indices and {projection_indices.size}"
+            " projection indices."
+        )
 
     num_projection_indices = np.prod(projection_grid_shape)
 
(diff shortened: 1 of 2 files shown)


imap_processing/ena_maps/map_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    num_projection_indices = np.prod(projection_grid_shape)
+
+    if value_array.ndim == 1:
+        binned_values = np.bincount(
+            projection_indices,
+            weights=value_array[input_indices],
+            minlength=num_projection_indices,
+        )
+    elif value_array.ndim == 2:
```
tmplummer: I'm not sure I understand the need for the 2D case. Is this for a variable that has a non-spatial dimension such as energy step?

imap_processing/ena_maps/map_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    projection_grid_shape: tuple[int, int],
+) -> NDArray:
+    """
+    Bin an array of values at the given indices.
+
+    Parameters
+    ----------
+    value_array : NDArray
+        Array of values to bin.
+    input_indices : NDArray
```
tmplummer: Ok, I think that I finally understand that these indices give the ability to perform the "pull" binning. I would like to discuss this in the ENA L2/Mapping meeting because I am now of the opinion that not just agreeing to add HEALPix support is adding unnecessary complication (Assuming the need to do the push/pull goes away with HEALPix).

If we continue to support this way of binning in this function, I think that it should be made optional such that this function can be called without the input_indices.

imap_processing/ena_maps/map_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    projection_grid_shape: tuple[int, int],
+) -> dict[str, NDArray]:
+    """
+    Project values from input grid to projection grid based on matched indices.
+
+    Parameters
+    ----------
+    input_values_to_bin : dict[str, NDArray]
+        Dict matching variable names to arrays of values to bin.
+    input_indices : NDArray
```
tmplummer: Like my above comment, I think that providing `input_indices` should be optional.

imap_processing/ena_maps/map_utils.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+    else:
+        raise NotImplementedError(
+            "Only 1D and 2D arrays are supported for binning. "
+            f"Received array with shape {value_array.shape}."
+        )
+    return binned_values
+
+
+def bin_values_at_indices(
```
tmplummer, line 97: Is the goal of this to provide a single function that will do the binning for all variables in a single PSET?

imap_processing/tests/ena_maps/test_map_utils.py
```diff
@@ -0,0 +1,238 @@
...
+                projection_grid_size,
+                extra_axis_size,
+            ),
+        )
+
+        # Create the expected projection values by summing the input values in a loop
+        # This is different from the binning function, which uses np.bincount
+        expected_projection_values = np.zeros((projection_grid_size, extra_axis_size))
+        for ii, ip in zip(input_indices, projection_indices):
+            expected_projection_values[ip, :] += value_array[ii, :]
```
tmplummer, line 102: Cool, your other implementation was useful!

imap_processing/ena_maps/map_utils.py
```diff
@@ -0,0 +1,128 @@
...
+        raise ValueError(
+            "Indices must be 1D arrays. "
+            "If using a rectangular grid, the indices must be unwrapped."
+        )
+    if input_indices.size != projection_indices.size:
+        raise ValueError(
+            "The number of input and projection indices must be the same. \n"
+            f"Received {input_indices.size} input indices and {projection_indices.size}"
+            " projection indices."
+        )
```
tmplummer: I like these checks. Should they be in the `bin_single_array_at_indices` function in case someone wants to access that lower level function?

imap_processing/tests/ena_maps/test_map_utils.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+    def test_bin_single_array_at_indices_complex(
+        self, projection_grid_shape, input_grid_shape
+    ):
+        """Test coverage for bin_single_array_at_indices function w/ complex 2D input,
+        Corresponding to an extra axis that is not spatially binned.
+        Parameterized across different input and projection grid shapes.
+        """
+        np.random.seed(0)
+        extra_axis_size = 11  # Another axis which is not spatially binned, e.g. energy
```
tmplummer, line 76: Ah, maybe this answers my other comment about why the 2D case is being supported. I'd like to consider where that functionality really belongs and if it can be made more general. Hi and Lo (I think) both have PSET variables with 2 non-spatial dimensions: energy and coincidence type.

tmplummer (commented):
Really nice test coverage! I have a couple of suggestions and questions that may be worth discussing in a zoom call before moving forward. I think this is all supper close to being good to go.

imap_processing/ena_maps/map_utils.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Parameters
+    ----------
+    value_array : NDArray
+        Array of values to bin.
+    input_indices : NDArray
+        Ordered indices for input grid, corresponding to indices in projection grid.
+        1 dimensional. May be non-unique, depending on the projection method.
+    projection_indices : NDArray
+        Ordered indices for projection grid, corresponding to indices in input grid.
```
laspsandoval: In the case of using HEALPIX would these be the pixel ids?

And then could you just add this so that the input_indices aren't required?
if input_indices is None:
    input_indices = np.arange(value_array.shape[0])

imap_processing/ena_maps/map_utils.py
laspsandoval: I thought that would be for 2D spatial dimensions for Ultra rectangular coordinates (while Hi and Lo only require a single dimension since they are using phase angle). Is that right or am I mistaken?

laspsandoval (commented):
Thank you so much for keeping this small. I can focus on the details better. I think I've got it. The only question is related to the 1D vs 2D comment from Tim. I thought that this was only for spatial dimensions (and not energy steps). If this is for other non-spatial dimensions then I might have more questions.

Everything looks really good.

imap_processing/ena_maps/map_utils.py
nkerman: Yes, it's for that exact case. We want one spatial dimension for either the tesselated (healpix) or unwrapped rectangular tiling of the sky. But those pixels might have multiple energy bins. (They certainly will for Ultra).

imap_processing/ena_maps/map_utils.py
nkerman: Yes, I think it's important to allow for non-unique "input_indices", where input here is from the point of view of the binning function, not the projection. 

The other reason I had for passing input_indices is that, when using `ravel_multi_index` to go from your `digitize`d `az_bin`, and `el_bin`, it must leave the simple 1D tiling of the sky and (under-the-hood) unwrap a 2D rectangular grid. This unwrapping can be either "C" or "F" style, and if you used one style to unwrap your rectangular grid in the first place, you need to use the same style here. 

In the original PR with a single function for matching rectangular indices, I handled this by keeping track of the order the input indices should be in, if they were unwrapped. I'll be honest, I have a hard time keeping track of and explaining this issue, but I needed to do this so that when projecting a frame into the same frame and spacing, the input indices matched the output indices. If I didn't, I could end up with something like:
```
input indices >>>  [1,   2,   3, ..., 360, 361, ...]
output indices >>> [1, 361, 721, ..., 2  , 362, ...]
```

I think making it optional is a good idea. I _think_ the issue will go away when we can guarantee that the same ravel order is used everywhere.

imap_processing/ena_maps/map_utils.py
nkerman: Yes!

imap_processing/ena_maps/map_utils.py
nkerman: > In the case of using HEALPIX would these be the pixel ids?

Yep. 

> And then could you just add this so that the input_indices aren't required?
if input_indices is None:
input_indices = np.arange(value_array.shape[0])

Yes, I think that's exactly what I'll add!

imap_processing/ena_maps/map_utils.py
nkerman: Sure! I originally had it there, but then decided to move them up the hierarchy, since the indices are the same across all variables to be binned.

imap_processing/tests/ena_maps/test_map_utils.py
nkerman: Ah, interesting. I don't know for a fact that I couldn't extend the `apply_along_axis` call to 3D arrays, or perhaps `apply_over_axes`... I'll give that a try and test it as well.

imap_processing/tests/ena_maps/test_map_utils.py
nkerman: Thanks! Yeah, I like testing things with blunt, loopy logic - it convinces me that the more efficient way is identical too.

imap_processing/tests/ena_maps/test_map_utils.py
nkerman: Turns out `apply_over_axes` works perfectly well with more "extra axes". I've just pushed a new version, which now allows 1+ dimension. 

That makes it more important than ever to ensure that the one and only spatial dimension is along the 0th axis. I've clarified this in the docs.

nkerman: I'm testing it completely manually (hardcoded expected values) for 1-3 dimensional data, and in an automated way (looping through spatial bins) for a few higher dim cases. It gets slow to do too many dims, but it sounds like we'll only ever have 1-3D data.

imap_processing/ena_maps/map_utils.py
nkerman: Moved checks to bin_single_array_at_indices

nkerman:
> Thank you so much for keeping this small. I can focus on the details better. I think I've got it. The only question is related to the 1D vs 2D comment from Tim. I thought that this was only for spatial dimensions (and not energy steps). If this is for other non-spatial dimensions then I might have more questions.
> 
> Everything looks really good.

Thanks, Laura! Does this now make sense in light of the other threads on the MR? Only the spatial dimension should be binned on, but there may be multiple energy bins at each spatial bin. These get binned independently.

E.g. consider an input value array of shape `(4, 10)`, which has 4 spatial bins _(maybe it was a super coarse rectangular grid of 2 lats, 2 lons...)_ and 10 energy bins. It's being projected onto another grid. 
* Input pixel indices `(1,2,3,4)` correspond to projection pixel indices `(100, 100, 101, 102)`.
* At each of those pixels, there are multiple energy levels, all of which should be brought over to the corresponding projection pixel and summed, but they should be done independently  of all other energy bins. 

By using `bincount` in `apply_along_axis`, and stipulating that the 0th axis be the one and only spatial axis, we can do this efficiently for 1+ dim arrays, with no loops!

tmplummer (approved):
I am happy with this. Thanks!

laspsandoval (approved):
Nicely done!

nkerman:
Thank you @subagonsouth @laspsandoval for your reviews!

PR #1265, opened 2025-01-14
Ultra l1b - Extendedspin First Pass

### LINKED ISSUE #1243: Ultra l1b: Extended Spin
### Algorithm Description:

May be broken down further

laspsandoval:
# Change Summary

## Overview
What you will want to focus on is ultra_l1b_culling.py and test_ultra_l1b_culling.py. Everything else is just modified to get those programs working.

## New Files
- ultra_l1b_culling.py
   - Gets spin_number, spin_start_time, spin_duration based off of mission elapsed time (met) and Universal Spin Table
   - Bins based on spin number and energy level
   - Creates a quality flag for counts that exceed a given threshold at each energy level for each spin

## Updated Files
- conftest.py
   - moved the creation of l1b de dataset here
- quality_flags.py
   - defined a new "HIGHCOUNTS" flag for Ultra
   - added event_times attribute
- de.py
   - added event_times to dataset
- constants.py
   - added thresholds for counts based on each energy level
   
## Testing
- test_quality_flags.py
- test_ultra_l1b_culling.py
- test_de.py

## TODO
Todos are listed at the top of ultra_l1b_culling.py. I also need to confirm if we want an entire spin discarded or only the spins discarded at each energy level.

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
@@ -1,0 +1,50 @@
+"""Tests Culling for ULTRA L1b."""
+
+from imap_processing.ultra.l1b.ultra_l1b_culling import get_spin
+
+# @pytest.fixture()
+# def pointing_frame_kernels(spice_test_data_path):
+#     """List SPICE kernels."""
+#     required_kernels = [
+#         "imap_wkcp.tf",
+#         "imap_sclk_0000.tsc",
+#         "naif0012.tls",
+#     ]
+#     kernels = [str(spice_test_data_path / kernel) for kernel in required_kernels]
+#     return kernels
+
+# @pytest.mark.usefixtures("use_fake_spin_data_for_time")
+# def test_cull_something(use_fake_spin_data_for_time, l1c_culling_test_data,
+#                         pointing_frame_kernels):
+#
+#     df = pd.read_csv(l1c_culling_test_data, delim_whitespace=True,
+#                      comment='#', header=0)
+#
+#     spice.furnsh(pointing_frame_kernels)
+#     # Get IDs.
+#     # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.gipool
+#     id_imap_sclk = spice.gipool("CK_-43000_SCLK", 0, 1)
+#     # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.sce2c
+#     # Convert start and end times to SCLK.
+#     sclk_begtim = spice.sce2c(int(id_imap_sclk), df['tdb'].min())
+#     met_begtime = sclkticks_to_met(sclk_begtim)
+#
+#     # Do one pointing at a time
+#     use_fake_spin_data_for_time(met_begtime)
+
+
+def test_de_dataset(use_fake_spin_data_for_time, l1b_de_dataset):
+    l1b_de_dataset["de_event_met"]
+    spin_number = l1b_de_dataset["SPINNUMBER"].values
+
+    # Create fake universal spin table
+    use_fake_spin_data_for_time(
+        l1b_de_dataset["de_event_met"][0], l1b_de_dataset["de_event_met"][-1]
+    )
+    # From universal spin table
+    # From l1b de dataset
+    energy = de_dataset["ENERGY_PH"].values
+    # 0-10, 10-20, above 20 keV
+
+    get_spin(l1b_de_dataset["de_event_met"], l1b_de_dataset["energy"])
+    print("hi")
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
@@ -14,24 +14,24 @@ from imap_processing.ultra.l1b.ultra_l1b_culling import (
 def test_get_spin(use_fake_spin_data_for_time, l1b_de_dataset):
     """Tests get_spin function."""
     use_fake_spin_data_for_time(
-        l1b_de_dataset["de_event_met"][0], l1b_de_dataset["de_event_met"][-1]
+        l1b_de_dataset["event_times"][0], l1b_de_dataset["event_times"][-1]
     )
     spin_number, spin_start_time, spin_duration = get_spin(
-        l1b_de_dataset["de_event_met"]
+        l1b_de_dataset["event_times"]
     )
 
     assert len(np.unique(spin_number)) == len(np.unique(spin_start_time))
-    assert np.all(l1b_de_dataset["de_event_met"].values >= spin_start_time)
+    assert np.all(l1b_de_dataset["event_times"].values >= spin_start_time)
 
 
 def test_get_energy_histogram(use_fake_spin_data_for_time, l1b_de_dataset):
     """Tests get_energy_histogram function."""
 
     use_fake_spin_data_for_time(
-        l1b_de_dataset["de_event_met"][0], l1b_de_dataset["de_event_met"][-1]
+        l1b_de_dataset["event_times"][0], l1b_de_dataset["event_times"][-1]
     )
 
-    spin_number, _, _ = get_spin(l1b_de_dataset["de_event_met"])
+    spin_number, _, _ = get_spin(l1b_de_dataset["event_times"])
     hist, _ = get_energy_histogram(spin_number, l1b_de_dataset["energy"].values)
 
     assert hist.shape == (4, 15)
@@ -79,24 +79,23 @@ def test_flag_spin(use_fake_spin_data_for_time, l1b_de_dataset):
     """Tests flag_spin function."""
 
     use_fake_spin_data_for_time(
-        l1b_de_dataset["de_event_met"][0], l1b_de_dataset["de_event_met"][-1]
+        l1b_de_dataset["event_times"][0], l1b_de_dataset["event_times"][-1]
     )
 
-    spin_number, _, _ = get_spin(l1b_de_dataset["de_event_met"])
+    spin_number, _, _ = get_spin(l1b_de_dataset["event_times"])
     energy = l1b_de_dataset["energy"].values
     hist, spin_edges = get_energy_histogram(spin_number, energy)
 
-    quality_flags_data = flag_spin(l1b_de_dataset["de_event_met"], energy)
-    assert np.all(
-        quality_flags_data[energy < 0] & ImapUltraFlags.NEG.value
-        == ImapUltraFlags.NEG.value
-    )
+    quality_flags, spin, energy = flag_spin(l1b_de_dataset["event_times"], energy)
 
-    flag = ImapUltraFlags(quality_flags_data[0])
+    flag = ImapUltraFlags(quality_flags[0])
     assert flag.name == "HIGHCOUNTS"
+
+    spin_1 = quality_flags[spin == 1]
+
     components = []
     for flag in ImapUltraFlags:
-        if quality_flags_data[np.where(quality_flags_data == 10)][0] & flag.value:
+        if quality_flags[np.where(quality_flags == 10)][0] & flag.value:
             components.append(flag.name)
     assert components == ["NEG", "HIGHCOUNTS"]
 
@@ -109,6 +108,6 @@ def test_flag_spin(use_fake_spin_data_for_time, l1b_de_dataset):
         if hist[energy_idx][spin_idx] > UltraConstants.COUNTS_THRESHOLDS[energy_idx]:
             mask = (energy_bin_idx == energy_idx) & (spin_bin_idx == spin_idx)
             assert np.all(
-                (quality_flags_data[mask] & ImapUltraFlags.HIGHCOUNTS.value)
+                (quality_flags[mask] & ImapUltraFlags.HIGHCOUNTS.value)
                 == ImapUltraFlags.HIGHCOUNTS.value
             )
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
@@ -91,23 +91,10 @@ def test_flag_spin(use_fake_spin_data_for_time, l1b_de_dataset):
     flag = ImapUltraFlags(quality_flags[0])
     assert flag.name == "HIGHCOUNTS"
 
-    spin_1 = quality_flags[spin == 1]
+    flagged_indices = np.unique(spin)[hist[0, :] > 0]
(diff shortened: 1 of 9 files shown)


laspsandoval:
pre-commit.ci autofix

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    spin_df = get_spin_data()
+
+    last_spin_indices = (
+        np.searchsorted(spin_df["spin_start_time"], met, side="right") - 1
+    )
+    spin_number = spin_df["spin_number"].values[last_spin_indices]
+    spin_start_time = spin_df["spin_start_time"].values[last_spin_indices]
+    spin_duration = spin_df["spin_period_sec"].values[last_spin_indices]
+
```
demajistre, line 34: Question about the spin table - do we know that 
`spin_df["spin_start_time"].values[last_spin_indices+1] = spin_start_time+spin_duration`?
For this purpose, I think it would be OK, but it's not really clear if the period of the spin is exactly the duration (unless the period is defined this way)

tmplummer: @demajistre, I dug up some notes on confluence that says "Spin end time for one spin should be identical to the spin start time for the next spin." Reading between the lines, I take this to mean that your equation is True (within floating point precision).

imap_processing/ultra/constants.py
```diff
@@ -72,3 +73,8 @@ class UltraConstants:
     # Constants for species determination based on ctof range.
     CTOF_SPECIES_MIN = 50
     CTOF_SPECIES_MAX = 200
+
+    # TODO: this is a temporary place for this.
+    # Thresholds for culling based on counts.
+    CULLING_ENERGY_BIN_EDGES: ClassVar[list] = [-1e5, 0, 10, 20, 1e5]
+    COUNTS_THRESHOLDS: ClassVar[list] = [0, 100, 100, 100]
```
demajistre: Note for the future - some of these 'constants' are actually configuration values - there should be some standard method developed (maybe across the SDC?) for defining and using configuration files.

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -0,0 +1,111 @@
...
+import numpy as np
+from numpy.typing import NDArray
+
+from imap_processing.quality_flags import ImapUltraFlags
+from imap_processing.spice.geometry import get_spin_data
+from imap_processing.ultra.constants import UltraConstants
+
+
+def get_spin(met: NDArray) -> tuple[NDArray, NDArray, NDArray]:
```
tmplummer: It seems like this function is missing from geometry.py. While the implementation looks good, there is no error checking in case query times are outside the range of spin data available. See: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/geometry.py#L206

I'm going to write a ticket to pull this function into the shared module and add error handling.

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -0,0 +1,111 @@
...
+    spin_duration = spin_df["spin_period_sec"].values[last_spin_indices]
+
+    return spin_number, spin_start_time, spin_duration
+
+
+def get_energy_histogram(
+    spin_number: NDArray, energy: NDArray
+) -> tuple[NDArray, NDArray]:
+    """
+    Compute a 2D histogram of the counts.
```
tmplummer:
```suggestion
    Compute a 2D histogram of the counts binned by energy and spin number.
```

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -0,0 +1,111 @@
...
+    # Map data points to bins
+    energy_bin_idx = (
+        np.digitize(energy, bins=UltraConstants.CULLING_ENERGY_BIN_EDGES) - 1
+    )
+    spin_bin_idx = np.digitize(spin, bins=spin_edges) - 1
+
+    for energy_idx, spin_idx in np.ndindex(hist.shape):
+        if hist[energy_idx][spin_idx] > UltraConstants.COUNTS_THRESHOLDS[energy_idx]:
+            mask = (energy_bin_idx == energy_idx) & (spin_bin_idx == spin_idx)
+            quality_flags_data[mask] |= ImapUltraFlags.HIGHCOUNTS.value
```
tmplummer: Reading through this, I was thinking about how you are essentially having to redo some of the code that is performed as part of the `np.histogramdd` function, namely associating each event with a bin. I found the family of scipy functions that will allow you to avoid this if you want. I'm linking the 2D version below but there is also a 1D and DD version. I don't have a feel for if this optimization is worth it nor do I have a strong opinion that this code needs to be changed but wanted to point this out anyway.
https://docs.scipy.org/doc/scipy-1.15.0/reference/generated/scipy.stats.binned_statistic_2d.html

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    spin_edges = np.unique(spin_number)
+    spin_edges = np.append(spin_edges, spin_edges[-1] + 1)
+
+    # 2D binning.
+    hist, _ = np.histogramdd(
+        sample=(energy, spin_number),
+        bins=[UltraConstants.CULLING_ENERGY_BIN_EDGES, spin_edges],
+    )
+
```
demajistre, line 69: This is not quite the correct quantity - what we're looking for here is the Count Rate for the spin(i.e., [number of counts]/[spin period]), but what you are building here is the number of counts in a spin. If the spin period were constant, we could use the number of counts per spin here, but this may not be the case.  If things go as planned, this is not going to be a problem, but if the target spin rate gets adjusted during the mission, there will be a problem here.

imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def test_get_energy_histogram(use_fake_spin_data_for_time, l1b_de_dataset):
+    """Tests get_energy_histogram function."""
+
+    use_fake_spin_data_for_time(
+        l1b_de_dataset["de_event_met"][0], l1b_de_dataset["de_event_met"][-1]
+    )
+
+    spin_number, _, _ = get_spin(l1b_de_dataset["de_event_met"])
```
tmplummer: Since this test is covering `get_energy_histogram` and you pass in the `spin_number` I would suggest just creating the `spin_number` array that you want to remove any reliance that this test has on `get_spin`.

imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+    spin_bin_idx = np.digitize(spin_number, bins=spin_edges) - 1
+
+    for energy_idx, spin_idx in np.ndindex(hist.shape):
+        if hist[energy_idx][spin_idx] > UltraConstants.COUNTS_THRESHOLDS[energy_idx]:
+            mask = (energy_bin_idx == energy_idx) & (spin_bin_idx == spin_idx)
+            assert np.all(
+                (quality_flags_data[mask] & ImapUltraFlags.HIGHCOUNTS.value)
+                == ImapUltraFlags.HIGHCOUNTS.value
+            )
```
tmplummer: Having this code in the test that duplicates the code being tested should be avoided. Can you create a synthetic dataset that would allow you test this in a more deterministic manor?

tmplummer (commented):
Looks really good. I just have a couple of suggestions.

demajistre:
I don't think this is doing exactly what is necessary here.  When we actually start culling at this stage, we'll be throwing out entire spins - rather than individual events.  One of the necessary quantities calculated at L1C is the exposure time, for which we'll need the number of spins actually used.  My suggestion is that you want a set of flags that applies to each SPIN and not each EVENT.   This was the idea of having an 'extended spin table' - i.e., an instrument specific table of spins with data items used to decide whether we use the data from each spin. In a logical sense, we are basically adding columns (or extending) to the original spin table.  Further, I think we want either a flag or a value for each energy range, rather than a single flag that says the one or more of the count rates exceeded the threshold. 

It is probably helpful to add an event quality flag, but we also need the extended spin values.  It may or may not be the best place to accumulate the event quality flag here.  It is possible that it would be better to just build the extended spin table, and use it as part of quality flag generation later.

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -0,0 +1,111 @@
...
+    Returns
+    -------
+    hist : NDArray
+        A 2D histogram array.
+    spin_edges : NDArray
+        Edges of the spin number bins.
+    """
+    spin_edges = np.unique(spin_number)
+    spin_edges = np.append(spin_edges, spin_edges[-1] + 1)
```
nkerman: Is spin edges (from `np.unique(spin_number)`) guaranteed to be sorted in ascending order, such that you can append the last spin_edge + 1? Might be safer to append `spin_edges.max() + 1` or else sort it first?

imap_processing/ultra/l1b/ultra_l1b_culling.py
nkerman: We should be able to take array of binned counts (`hist`) and simply divide along the spin_number axis by the period of each spin, right?

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -0,0 +1,111 @@
...
+    spin, _, _ = get_spin(met)
+    hist, spin_edges = get_energy_histogram(spin, energy)
+
+    # Map data points to bins
+    energy_bin_idx = (
+        np.digitize(energy, bins=UltraConstants.CULLING_ENERGY_BIN_EDGES) - 1
+    )
+    spin_bin_idx = np.digitize(spin, bins=spin_edges) - 1
+
+    for energy_idx, spin_idx in np.ndindex(hist.shape):
```
nkerman: I thought from last week's meeting, that we needed to be able to filter based on total count rate in a spin (i.e., summed over all energy bins). Do we also want to do a total count threshold comparison? I might have misunderstood last week though.

imap_processing/ultra/l1b/ultra_l1b_culling.py
demajistre: Yep.  That's the rate.

imap_processing/ultra/l1b/ultra_l1b_culling.py
laspsandoval: From Bob's message above it has to be a filter based on count rate for spin/energy combination. So I think I'll address that in my revisions, but if the Ultra team wants all energies combined too and a threshold for that then I can add it. I'll see what they say.

laspsandoval: Ok. He says he wants it broken down by spin and energy. So what I have is ok.

demajistre:
Two overall comments - one for this PR and one more general
1) As built, this code builds a structure that flags spin/energy combinations that exceed a set of constant values.  This is only part of the job of preparing for culling, and only fits a special, simplified case.  This simple threshold may or may not be what we need and we will almost certainly need other quantities for culling.  The simplest change here would be to accumulate count rates in each of the spin/energy bins, which can be used with various different culling algorithms.  An example here (and a likely useful algorithm) would be to construct a threshold based on the average and standard deviation of the counts received in a pointing rather than a simple constant threshold. Also, there are other quantities that will be tabulated per spin that we may want to cull with (the easy example here is CODICE level 2 data).  My overall suggest here is to build a data structure (that can be made persistent) that extends the spacecraft spin table.  That is, something that basically adds columns to the spacecraft spin dataframe with the necessary quantities for culling out spins.  In this case that would involve adding the count rates in each energy bin to each entry in the spin table.  You may want to just make the 'ultra_extended_spin_ table' into a class, so you can keep the extended data frame in there and document things like the energy bin ranges within each object and more generally which data you have added to the dataframe for each object (e.g., has the CODICE data been added or not). The spin-level culling algorithms could then consume these objects to provide a list of spins to use for building L1C data.

2) More generally for the SDC - I strongly suggest that you adopt a single internal time representation across the SDC, or at least in the libraries that relate to instrument processing.  This need not be the same as whatever is represented in the CDF files, rather a single unambiguous time representation that is recognized across the codebase that can be easily converted to whatever you put in the CDF files.  A suggestion would be to define something like an sdc_time class that has constructors that use various types of time (met, TDB, segmented UT, etc.) and methods that can return a time in those various representations. Mixing time systems is an easy path to error and/or confusing conversions inside units of code that have to identify and convert times every time they see it. Further, this is something that should be defined and managed by the developers, this is not something that should be determined by consensus of instrument teams - they should review it to be sure it is sufficient for their application, but it is really and implementation decision best made by those using it.

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -0,0 +1,130 @@
...
+    # Counts per spin at each energy bin.
+    hist, _ = np.histogramdd(
+        sample=(energy, spin_number),
+        bins=[UltraConstants.CULLING_ENERGY_BIN_EDGES, spin_edges],
+    )
+
+    # Count rate per spin at each energy bin.
+    for i in range(hist.shape[1]):
+        spin_duration = spin_df.spin_period_sec[spin_df.spin_number == i]
+        hist[:, i] /= spin_duration.values[0]
```
nkerman, line 73: Can you do this with broadcasting rather than a for loop for efficiency? Something like:

```
spin_durations = spin_df.spin_period_sec[spin_df.spin_number == <all_relevant_spin_numbers>]
hist /= spin_durations.values[:, np.newaxis]
```

imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
```diff
@@ -0,0 +1,83 @@
...
+    """Tests flag_spin function."""
+
+    time, _, energy, expected_counts = test_data
+    quality_flags, spin, energy = flag_spin(time, energy)
+
+    flag = ImapHkUltraFlags(quality_flags[1])
+    assert flag.name == "HIGHCOUNTS"
+
+    flagged_indices = np.unique(spin)[expected_counts[0, :] / 15 > 0]
+    unflagged_indices = np.setdiff1d(np.unique(spin), flagged_indices)
```
nkerman, line 78: I've never used setdiff before - that's a useful function!

nkerman (approved):
Looks good to me! There's at least one for loop that I think could be slightly more efficient as a np broadcast, but given that we're only talking about a few thousand spins a day, and <100 energy bins, I'm not too anti-loop.

Issue #1232, opened 2024-12-19
SWE: Units and angle fix from follow up with Ruth


tech3371:
- [x] #1287

- [x]  Double check if function used to calculate spin phase is the correct one to use. Should the spin phase be in S/C or instrument frame?
Yes. We discussed that whether to calculate it in SC or instrument frame may be dependent on MAG. But Ruth mention to go with SC frame.
Ruth asked how is zero angle/phase is defined in the SC frame? I tried to answer that it may be related to spin data. Our assumption is that every spin start time is captured at spin angle zero. Then we use spin period to calculate other spin phase.

- [x]  Ask Ruth if spin phase and spin angle is same thing. The document says spin angle, but the code uses spin phase. The document says this:
      _the spin sectors are converted to spin angles of the spacecraft by
      combining SWE timing information with knowledge of the spacecraft
      spin from spacecraft ephemeris data._
Answer: Yes because we can convert spin phase into spin angle(either in degree or radian). We can multiple spin phase by  360 or 2pi to convert to angle. https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1306
- [x] should we add this check, in_flight_cal_df[in_flight_cal_df["met_time"] <= acquisition_time] as suggested on this PR, https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1267/files. Yes. Ruth agrees we don't need = comparison.
- [x] #1274

tech3371:
Sub-tasks have been done or in review. Create new issue as needed.

PR #1267, opened 2025-01-15
SWE: Adding ability to read in-flight calibration file

### LINKED ISSUE #114: Apply calibrations based on in-flight calibration

### LINKED ISSUE #893: Read and apply in-flight calibration factors to counts using simulated data
From document
 /*
    ** APPLY ELECTRON CALIBRATION FACTORS
    **
    ** These factors are used to account for changes in gain with time.
    **
    ** They are derived from the weekly electron calibration data.
    ** The procedure is to compare the counts obtained at the operating CEM 
    ** level (step 2 of the cal run) to the counts are the next higher CEM
    ** level (step 3 of the cal run).  We assume that the next higher 
    ** level is correct, and use this ratio to scale the counts.
    **
    */

    cal_factor = electron_cal(mystruct.sct);

    for (i=0; i<N_ENERGIES; i++){
      for (j=0; j<N_CEMS; j++){
	for (k=0; k<N_PHI; k++){
                  counts[i][j][k] = ccounts[i][j][k] * cal_factor[j];
	}
      }
    }

/*****************\
** electron_cal  **  given an S/C time, return interpolated electron cals
\*****************/
float *
electron_cal(uint32 sct)
{
  int i;
  static int last_i;			/* value of i at previous iteration */
  static float factor[7];

  /*
  ** First time through?  Read in the calibration data
  */
  if (cal == NULL) init_electron_cal();

  /*
  ** Iterate over all points, until we find ourselves within this range.
  **
  ** When we do, return a linear interpolation of the factors at the
  ** two measured points.
  **
  ** NOTE: since the cal data file is in sequential order, and the only
  ** code that invokes us also does so in sequence, we can optimize a
  ** step by preserving last_i and assuming we will never be invoked for
  ** a time prior to the last one.
  */
  for (i=last_i; i < npoints-1; i++){
    if (cal[i].sct <= sct  &&  sct < cal[i+1].sct) {
      int j;
      int run = cal[i+1].sct - cal[i].sct;
      int tdif= sct          - cal[i].sct;

      last_i = i;

      for (j=0; j<7; j++) {
	float slope = (cal[i+1].factor[j] - cal[i].factor[j]) / run;

	factor[j] = cal[i].factor[j] + slope * tdif;
      }
      return factor;
    }
  }

  printf("cannot extrapolate beyond last cal\n");
  exit(1);
}

tech3371:
# Change Summary
closes #893 

## Overview
SWE won't have real in-flight calibration file well after launch and commissioning and few months after that. I am writing ability to read and apply in-flight calibration factor the corrected counts data for future use. It will need some updates as we decide more on how ancillary or calibration files are read through cli.py.

## Updated Files
- imap_processing/cli.py
   - capturing TODO
   - main updates of this PR

## Testing
- imap_processing/tests/swe/test_swe_l1b_science.py


imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -171,14 +204,46 @@ def calculate_calibration_factor(time: int) -> None:
...
+    Returns
+    -------
+    cal_factor : numpy.ndarray
+        Calibration factor for each 7 CEM detectors. Array shape is (7,).
+    """
+    # Read in in-flight calibration data
+    in_flight_cal_df = read_in_flight_cal_data()
+    # For the given acquisition time, find it's nearest pre and post time
+    pre_cal_time = (
+        in_flight_cal_df[in_flight_cal_df["met_time"] < acquisition_time]
```
tmplummer: Should one of the checks include equivalence such as:
```suggestion
        in_flight_cal_df[in_flight_cal_df["met_time"] <= acquisition_time]
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -267,6 +341,22 @@ def populate_full_cycle_data(
...
+            # Apply calibration based on in-flight calibration. In-heritage mission,
+            # they didn't apply in-flight calibration until after certain date.
+            # For this mission, SWE want to set the date to be 6 months after
+            # commissioning. With assumption that that launch date is September 23,
+            # 2025 and it may take 3 months commissioning, we will set in-flight
+            # calibration date to be July 1, 2026. SWE will communicate if they
+            # want to change this date.
+            in_flight_cal_date = np.datetime64("2026-07-01")
+            data_time = met_to_datetime64(base_quarter_cycle_acq_time)
+            if data_time > in_flight_cal_date:
```
tmplummer: Will the cal file have entries prior to this date? I guess that I am just wondering why `calculate_calibration_factor` doesn't have this logic and catch the case where acquisition_time is before any calibration entries and return no correction for that case.

tmplummer (commented):
This all looks good to me and helps out with my current task of adding a Hi cal file requirement for L1C processing. Thanks!

Just had the two questions.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -134,13 +135,45 @@ def convert_counts_to_rate(data: np.ndarray, acq_duration: int) -> npt.NDArray:
...
+    has 8 numbers, with the first being a time stamp in MET, and the next
+    7 being the factors for the 7 detectors.
+
+    This file will be updated weekly with new calibration data. In other
+    words, one line of data will be added each week to the existing file.
+    File will be in CSV format. Processing won't be kicked off until there
+    is in-flight calibration data that covers science data. TODO: decide
+    filename convention given this information. This function is a
+    placeholder for reading in the calibration data until we decide on
+    how to read calibration data through dependencies list.
```
bourque:
```suggestion
    is in-flight calibration data that covers science data.
    
    TODO: decide filename convention given this information. This function is a
    placeholder for reading in the calibration data until we decide on
    how to read calibration data through dependencies list.
```

Just so the TODO is not buried within the paragraph and can be seen more easily

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -171,14 +204,46 @@ def calculate_calibration_factor(time: int) -> None:
...
+        in_flight_cal_df[in_flight_cal_df["met_time"] < acquisition_time]
+        .iloc[-1]
+        .values[0]
+    )
+    post_cal_time = (
+        in_flight_cal_df[in_flight_cal_df["met_time"] > acquisition_time]
+        .iloc[0]
+        .values[0]
+    )
+    run = post_cal_time - pre_cal_time
```
bourque:
```suggestion
    cal_time_length = post_cal_time - pre_cal_time
```

This variable name was a bit confusing to me so just suggesting something different here

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -267,6 +341,22 @@ def populate_full_cycle_data(
                 + l1a_data["acq_start_fine"].data[packet_index + index] / 1000000
             )
 
+            # Apply calibration based on in-flight calibration. In-heritage mission,
+            # they didn't apply in-flight calibration until after certain date.
+            # For this mission, SWE want to set the date to be 6 months after
+            # commissioning. With assumption that that launch date is September 23,
```
bourque:
```suggestion
            # commissioning. With the assumption that the launch date is September 23,
```

bourque (approved):
Just some small suggestions, otherwise it looks good to me!

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Based on my conversation with Ruth, it sounds like they won't have time that would equal pre or post time. They won't be collecting science data during calibration data collection. But I will double check with her. Capturing that question on this ticket https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1232

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: no. This will be updated once they know when they can start collecting calibration data. Then rest will be handle in dependency tree in batch starter. SWE doesn't want science processing to kick off until science data is covered by calibration data range.

tech3371: talking or writing it out loud, I am finding flaw in my code. thinking...

tech3371: Since this will be and need to be handled in batch starter, I wonder if I can remove this if statement all together. I can check with Ruth and upload a in-flight-calibration file that we should use using that will not apply any factor to the data until they know. For example, until they start collecting calibration data, we can have a in-flight calibration file like this, `imap_swe_in-flight-cal_<start-date>_<end-date>_v000.csv` with this content.
```
start_time 1 1 1 1 1 1 1
end_time  1 1 1 1 1 1 1
```
With that content, the calibration factor would result in 1 which won't change data at all. What do you think?

tmplummer: That could work. I think that it would be cleaner to simply default to correction factors of 1 when there is no calibration file but that may be difficult to disentangle from the batch starter dependency of needing a calibration file to run processing?

tech3371: We need to figure this out soon for all of us. It would be nice to know how we should handle that. So I am happy to be first test case. I have some ideas and will run by Ruth and then run with our team.

tmplummer (approved):

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -134,13 +135,45 @@ def convert_counts_to_rate(data: np.ndarray, acq_duration: int) -> npt.NDArray:
...
+
+
+def calculate_calibration_factor(acquisition_time: float) -> npt.NDArray:
     """
     Calculate calibration factor.
 
     Steps to calculate calibration factor:
 
     1. Convert input time to match time format in the calibration data file.
+        Both input time and calibration data time should be in S/C MET.
```
greglucas: Rather than input_time do you want to use acquisition_time since that is the variable name input to this function?
```suggestion
    1. Convert the input time (acquisition_time) to match the time format in the calibration data file.
       Both the input time and calibration data time should be in S/C MET.
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -171,14 +204,46 @@ def calculate_calibration_factor(time: int) -> None:
...
 
-def apply_in_flight_calibration(data: np.ndarray) -> None:
+    Returns
+    -------
+    cal_factor : numpy.ndarray
+        Calibration factor for each 7 CEM detectors. Array shape is (7,).
+    """
+    # Read in in-flight calibration data
+    in_flight_cal_df = read_in_flight_cal_data()
+    # For the given acquisition time, find it's nearest pre and post time
```
greglucas: Why do you need pre and post times? I think it will be a sorted list and therefore you should just need to find the first time and then index +1 to get the bounds.

In general though, there is the `np.interp()` function that would be useful to use here, or `searchsorted` as suggested in this issue, where you could apply this to many acquisition times in your dataset all at once in a vectorized manner.

https://stackoverflow.com/questions/43772218/fastest-way-to-use-numpy-interp-on-a-2-d-array

imap_processing/swe/l1b/swe_l1b_science.py
greglucas: If you use `np.interp()` you can specify what you want the left/right extrapolations to be as well. I'd also suggest pushing this logic into the other function so that if you are before any given time the calibration is `np.ones()` or whatever value you want.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Is this what you had in mind about first suggestion?
```
    pre_cal_time_idx = in_flight_cal_df[in_flight_cal_df['met_time'] < acquisition_time].index[-1]
    post_cal_time_idx = pre_cal_time_idx + 1
    pre_cal_time = in_flight_cal_df.loc[[pre_cal_time_idx]]
    post_cal_time = in_flight_cal_df.loc[[post_cal_time_idx]]
    cal_time_length = post_cal_time - pre_cal_time
```

greglucas: Yes, I think so.

Then I would write that all out into a single lookup as the SO link suggests.

```python
    i = np.arange(x.size)
    j = np.searchsorted(xp, x) - 1
    d = (x - xp[j]) / (xp[j + 1] - xp[j])
    return (1 - d) * fp[i, j] + fp[i, j + 1] * d
```

Would be something like translating to your case (you might need to get the array values out of the dataframe with `.values` or similar to get an array to index into, I'm not sure.

```python
x = acquisition_time
xp = in_flight_cal_df.index
fp = in_flight_cal_df
```

But then you could pass in a large array of acquisition times and do all of the calculations at once.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Ruth confirmed that we don't need to use `=` comparison given above reason.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Ruth and I discussed plan for removing this and do that through dependency. I captured notes in this ticket https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1274 and will change in infrastructure repo.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -166,83 +166,34 @@ def read_in_flight_cal_data() -> pd.DataFrame:
     return empty_df
 
 
-def calculate_calibration_factor(acquisition_time: float) -> npt.NDArray:
+def interp_func(x: np.ndarray, xp: np.ndarray, fp: np.ndarray) -> npt.NDArray:
```
greglucas: This is a small helper and should probably be inlined into the one function that uses it or renamed from the generic `interp_func`

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -254,21 +205,31 @@ def apply_in_flight_calibration(
...
-    # calculate calibration factor
-    cal_factor = calculate_calibration_factor(acquisition_time)
-    # Apply to current quarter cycle data
+    # Read in in-flight calibration data
+    in_flight_cal_df = read_in_flight_cal_data()
+    # calculate calibration factor.
+    # return shape of interp_func is (24, 30, 7, 7) where last 7 dimension contains
+    # calibration factor for each CEM detector.
+    cal_factor = interp_func(
+        acquisition_time,
```
greglucas, line 264: I assume the acquisition time for each CEM detector is the same? Otherwise I don't think the for loop below would work as expected since it is picking off the diagonal. I'd suggestion just taking the first acquisition time, or only passing in the shape (24, 30) to this function.
```suggestion
        acquisition_time[..., 0],
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -254,21 +205,31 @@ def apply_in_flight_calibration(
...
+        in_flight_cal_df["met_time"].values,
+        in_flight_cal_df.iloc[:, 1:].values,
+    )
+    # Apply to full cycle data
     corrected_counts = corrected_counts.astype(np.float64)
-    corrected_counts *= cal_factor
+    for idx in range(7):
+        cem_cal_factor = cal_factor[:, :, idx, idx]
+        corrected_counts[:, :, idx] *= cem_cal_factor
```
greglucas: This should be a completely broadcastable multiplication indicating you have one cal factor per value in the array.
```suggestion
    corrected_counts *= cem_cal_factor
```

You are also taking in `corrected_counts`, but I think that is the returned quantity? I'd separate out the input/output array names and avoid in-place modification, maybe calling the input out explicitly.
```python
return input_data * cem_cal_factor
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -385,10 +337,16 @@ def populate_full_cycle_data(
     # data. But for now, we are advice to continue with current setup and can
     # add/change it when we get real data.
 
+    # Apply calibration based on in-flight calibration.
+    corrected_counts = apply_in_flight_calibration(full_cycle_data, acquisition_times)
```
greglucas: Are there multiple "corrections"? Maybe call this explicitly `calibrated_counts`?
```suggestion
    corrected_counts = apply_in_flight_calibration(full_cycle_data, acquisition_times)
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -134,51 +134,103 @@ def convert_counts_to_rate(data: np.ndarray, acq_duration: int) -> npt.NDArray:
...
+        raise ValueError("Acquisition times are outside the calibration time range)")
+
+    # This line of code finds the indices of acquisition_times in cal_times where
+    # acquisition_times should be inserted to maintain order. As a result, it finds
+    # its nearest pre and post time from cal_times.
+    input_time_indices = np.searchsorted(cal_times, acquisition_times)
+    # This line of code clips the indices to be within the range of cal_times.
+    # We did len(cal_times) - 2 because we want to find the nearest pre and
+    # post time from cal_times.
+    valid_indices = np.clip(input_time_indices, 0, len(cal_times) - 2)
```
tech3371: @greglucas Can I remove this now given above if check?

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Implemented this approach now

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -134,51 +134,108 @@ def convert_counts_to_rate(data: np.ndarray, acq_duration: int) -> npt.NDArray:
...
+    # Define the column headers
+    columns = ["met_time", "cem1", "cem2", "cem3", "cem4", "cem5", "cem6", "cem7"]
 
-    | 3. **Calculating Factors**: Once it finds these two nearby calibration points,
-    |     the function calculates a correction factor by drawing a straight line
-    |     between them (linear interpolation). This factor helps adjust the measurement
-    |     to make it more accurate, considering how the instrument's sensitivity changed
-    |     between those two calibration points.
+    # Create an empty DataFrame with the specified columns
+    empty_df = pd.DataFrame(columns=columns)
```
greglucas, line 165: Do you want to return a DataFrame of ones?

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -134,51 +134,108 @@ def convert_counts_to_rate(data: np.ndarray, acq_duration: int) -> npt.NDArray:
...
+        acquisition_times.min() < cal_times.min()
+        or acquisition_times.max() > cal_times.max()
+    ):
+        logger.error(
+            f"Acquisition min/max times: {acquisition_times.min()} to "
+            f"{acquisition_times.max()}."
+            f"Calibration min/max times: {cal_times.min()} to {cal_times.max()}."
+            "Acquisition times should be within calibration time range."
+        )
+        raise ValueError("Acquisition times are outside the calibration time range)")
```
greglucas: Add all of the logger information into the ValueError text. I think the ValueError messages should get sent to the log stream too?

imap_processing/swe/l1b/swe_l1b_science.py
greglucas: I think so, but this might be something good to add a test for. Make sure that you raise with times requested out of the coverage span of the dataframe and don't return valid indices.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -134,51 +134,108 @@ def convert_counts_to_rate(data: np.ndarray, acq_duration: int) -> npt.NDArray:
...
+    #   array([2])
+    # We need to use - 1 to get the pre_time_indices.
+    pre_time_indices = valid_indices - 1
+    post_time_indices = valid_indices
+    slope = (acquisition_times - cal_times[pre_time_indices]) / (
+        cal_times[post_time_indices] - cal_times[pre_time_indices]
+    )
+    return cal_data[pre_time_indices] + slope[..., None] * (
+        cal_data[post_time_indices] - cal_data[pre_time_indices]
+    )
```
greglucas: I personally find this harder to read than `(x - x[i-1]) / (x[i] - x[i-1])` because the -1s indicate at a glance what is going on I think. This still looks right to me, but just noting that.

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -134,51 +134,108 @@ def convert_counts_to_rate(data: np.ndarray, acq_duration: int) -> npt.NDArray:
...
+        2. Find the nearest in time calibration data point.
+        3. Linear interpolate between those two nearest time and get factor for
+           input time.
 
     Parameters
     ----------
-    time : int
-        Input time.
+    acquisition_times : numpy.ndarray
+        Data points to interpolate. Shape is (24, 30, 7).
```
greglucas: Didn't you drop the 7 from the times variable? Is it also adding a sweep_number to the front too? `(n_sweeps, 24, 30)`
```suggestion
        Data points to interpolate. Shape is (24, 30).
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -188,12 +245,31 @@ def apply_in_flight_calibration(data: np.ndarray) -> None:
...
+    # last 7 dimension contains calibration factor for each CEM detector.
+    cal_factor = calculate_calibration_factor(
+        acquisition_time,
+        in_flight_cal_df["met_time"].values,
+        in_flight_cal_df.iloc[:, 1:].values,
+    )
+    # Apply to full cycle data
+    corrected_counts = corrected_counts.astype(np.float64)
+    corrected_counts *= cal_factor
+    return corrected_counts
```
```suggestion
    return corrected_counts.astype(np.float64) * cal_factor
```

imap_processing/tests/swe/test_swe_l1b_science.py
```diff
@@ -62,4 +36,51 @@ def test_get_full_cycle_data_indices():
...
+    )
+
+    # Check for value outside of calibration time range
+    input_time = 1.0
+    acquisition_time = np.full((24, 30), input_time)
+
+    with pytest.raises(
+        ValueError, match="Acquisition times are outside the calibration time range"
+    ):
+        apply_in_flight_calibration(one_full_cycle_data, acquisition_time)
```
greglucas: Test the boundaries here for robustness? What happens when you land on exactly the final MET time, or just 1 beyond it? I think that will answer the np.clip() question above if you codify those in tests here.

greglucas (approved):
Looks good to me. I have some minor nits/suggestions for code clarity and tests, but I like this move to more vectorization of applying the calibration factors.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: I kept this one for easier unittests testing. But I can test that value error is raised with starting message.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Yep. I added test that checks valid time range. cool. will remove this.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: sounds good. I will add what `i` means in comment instead then.

imap_processing/tests/swe/test_swe_l1b_science.py
tech3371: cool

tech3371: technically, we shouldn't have time that will be same between science data time and cal time. I am gonna hold off this.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: This is just a placeholder. I will need to remove this once we figure out how we can read and pass cal data from `cli.py`.

PR #711, opened 2024-07-24
Various CoDICE L1 Updates

bourque:
# Change Summary

## Overview

This PR contains various updates to the CoDICE L1 algorithm, mainly to clean some things up, add support for `lo-pha` and `hi-pha` data products (event data), and changes required to get recently-acquired simulated data working in the pipeline.

## Updated Files

- `imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml`
  - Added CDF attrs for event data products 
- `imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml`
  - Formatting improvements, added missing attributes that are now required
- `imap_processing/cdf/config/imap_codice_l1b_variable_attrs.yaml`
  - Formatting improvements, added missing attributes that are now required
- `imap_processing/codice/codice_l0.py`
  - Added XTCE mapping for event data 
- `imap_processing/codice/codice_l1a.py`
  - Added support for event data; moved housekeeping dataset creation into module to be consistent with where similar code is being defined; other various cleanup/organizing
- `imap_processing/codice/constants.py`
  - Moved hard-coded APID list from codice_l1a to constants file
- `imap_processing/codice/packet_definitions/*.xml` 
  - Updates from newer version of `space_packet_parser` and updates packet definition spreadsheet 
- `imap_processing/codice/utils.py`
  - Moved `create_hskp_dataset` to `codice_l1a.py` to be consistent with where the code to create other datasets resides 
- `imap_processing/tests/codice/data/imap_codice_l0_*.pkts`
  - Updated/new simulated data  
- `imap_processing/tests/codice/test_codice_l0.py`
  - Renamed functions with `hskp` to be consistent with naming convention 
- `imap_processing/tests/codice/test_codice_*.py`
  - Improved formatting and updated tests

PR #1268, opened 2025-01-15
CoDICE L1a CDF attrs improvements

bourque:
# Change Summary

## Overview
This PR makes various small improvements to CoDICE L1a CDF attribute definitions, namely:
- Introducing a shared `hskp` variable since all of those use the same `DEPEND_0` AND `VAR_TYPE`
- Add missing `CATDESC`s
- Removing empty/blank attributes that were needed to avoid bug before SAMMI was introduced (see https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/711#discussion_r1690212527)

## Updated Files


tech3371 (approved):
Since this is for L1 and housekeeping data, I think what you have is good.

PR #1279, opened 2025-01-17
Upgrade dependencies - cdflib and sammi

vmartinez-cu:
This PR upgrades dependencies, specifically cdflib to 1.3.3 and sammi-cdf to 1.0

## Updated Files
- poetry.lock
   - upgraded cdflib to 1.3.3. This version fixes a variety of bugs:
      xarray_to_cdf
         - ISTP compliance checker works better when "support_data" has its own "DEPEND_1"
         - np.uint64 inputs will now throw an error (they were never supported, but now an error will be thrown)
         - Fixed a bug that Ana found that would allow corrupt CDFs to be generated if you had a string attribute, but told cdflib 
            it was a number
         - Fully removed the flags for from_unixtime, from_datetime, unixtime_to_cdftt2000, datetime_to_cdftt2000, 
            datetime64_to_cdftt2000
      cdf_to_xarray
         - Fixed a bug that caused dimensions of size 1 to be removed when read in from the cdf
- pyproject.toml
   - upgraded sammi-cdf to ^1.0


bourque (approved):
Now that CoDICE tests are passing, and assuming the GLOWS error is a simple fix, this looks good to me! Thanks for doing this!

maxinelasp (approved):
I added a few changes to your branch to fix the GLOWS issues, so now the tests are passing! Thanks for doing this!

bryan-harter (approved):

PR #1284, opened 2025-01-21
Add SPICE data (ephemeris, attitude, and celestial coords) to IDEX l1b CDF

### LINKED ISSUE #1282: Spice data
Add spice data to L1b cdfs.

### LINKED ISSUE #1306: Util function to go from spin phase to angle
Write to common util function to convert spin phase [0, 1) to angle in degree or radian. This came out from SWE and IDEX work and may need it for others as well.

lacoak21:
# Change Summary

closes #1282
closes #1306

## Overview
Add SPICE data (ephemeris, attitude, and celestial coords) to IDEX l1b CDF using code defined in imap_processing/spice/geometry.py. (Thank you to whomever wrote very readable code!) 

## Updated Files
- imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
   - Add attributes for the new spice arrays. Will need to go back and get expected valid ranges for some of the arrays.
   - Add function to query all the necessary spice data using functions from imap_processing/spice/geometry.py
   - Create xr.DataArrays for the spice data and add to the output cdf.
   - Add wrapper function for m2eul https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/m2eul.html
   - Add wrapper function for https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/recrad_c.html
- imap_processing/tests/idex/conftest.py
   - Created a pytest 'side_effect' function to enable l1b testing without furnishing kernels.
   - Test idex l1b without having to furnish kernels 
   - Add tests for new spice functions (for vectorized or not)

### DIFF
--- a/imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
@@ -42,8 +42,15 @@ l1b_target_base: &l1b_target_base
 trigger_base: &trigger_base
   <<: *l1b_data_base
   DISPLAY_TYPE: no_plot
-  CATDESC: ""
   UNITS: " "
+
+spice_base: &spice_base
+  <<: *l1b_data_base
+  VAR_TYPE: data
+  VALIDMIN: -9223372036854775808 # TODO: Find actual validmin
+  VALIDMAX: 9223372036854775807 # TODO: Find actual validmax
+  DISPLAY_TYPE: time_series
+  UNITS: Radians
 
 # <=== Instrument Setting Attributes ===>
 trigger_mode:
@@ -334,3 +341,86 @@ current_neg2p5v_bus:
   CATDESC: Negative current reading on -2.5V bus on LVPS Board (ADC 1 channel 7)
   LABLAXIS: Current
   UNITS: A
+
+# <=== Spice Data Attributes ===>
+ephemeris_position_x:
+  <<: *spice_base
+  FIELDNAM: Position X
+  CATDESC: Cartesian coordinate X positions for the IMAP spacecraft in the ECLIPJ2000 frame.
+  LABLAXIS: Position X
+  UNITS: km
+
+ephemeris_position_y:
+  <<: *spice_base
+  FIELDNAM: Position Y
+  CATDESC: Cartesian coordinate Y positions for the IMAP spacecraft in the ECLIPJ2000 frame.
+  LABLAXIS: Position Y
+  UNITS: km
+
+ephemeris_position_z:
+  <<: *spice_base
+  FIELDNAM: Position Z
+  CATDESC: Cartesian coordinate Z positions for the IMAP spacecraft in the ECLIPJ2000 frame.
+  LABLAXIS: Position Z
+  UNITS: km
+
+ephemeris_velocity_x:
+  <<: *spice_base
+  FIELDNAM: Velocity X
+  CATDESC: Velocity X positions for the IMAP spacecraft in the ECLIPJ2000 frame.
+  LABLAXIS: Velocity X
+  UNITS: km/s
+
+ephemeris_velocity_y:
+  <<: *spice_base
+  FIELDNAM: Velocity Y
+  CATDESC: Velocity Y positions for the IMAP spacecraft in the ECLIPJ2000 frame.
+  LABLAXIS: Velocity Y
+  UNITS: km/s
+
+ephemeris_velocity_z:
+  <<: *spice_base
+  FIELDNAM: Velocity Z
+  CATDESC: Velocity Z positions for the IMAP spacecraft in the ECLIPJ2000 frame.
+  LABLAXIS: Velocity Z
+  UNITS: km/s
+
+right_ascension:
+  <<: *spice_base
+  VALIDMIN: 0
+  VALIDMAX: 6.283185307179586
+  FIELDNAM: Right Ascension
+  CATDESC: Right ascension of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
+  LABLAXIS: Right Ascension
+
+declination:
+  <<: *spice_base
+  VALIDMIN: -1.5707963267948966
+  VALIDMAX: 1.5707963267948966
+  FIELDNAM: Declination
+  CATDESC: Declination of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
+  LABLAXIS: Current
+
+attitude_roll:
+  <<: *spice_base
+  VALIDMIN: -3.141592653589793
+  VALIDMAX: 3.141592653589793
+  FIELDNAM: Roll
+  CATDESC: Roll angle (rotation around the X-axis) of the IMAP spacecraft in the ECLIPJ2000 reference frame.
+  LABLAXIS: Roll
+
+attitude_pitch:
+  <<: *spice_base
+  VALIDMIN: -1.5707963267948966
+  VALIDMAX: 3.141592653589793
+  FIELDNAM: Pitch
+  CATDESC: Pitch angle (rotation around the Y-axis) of the IMAP spacecraft in the ECLIPJ2000 reference frame.
+  LABLAXIS: Pitch
+
+attitude_yaw:
+  <<: *spice_base
+  VALIDMIN: -3.141592653589793
+  VALIDMAX: 3.141592653589793
+  FIELDNAM: Yaw
+  CATDESC: Yaw angle (rotation around the Z-axis) of the IMAP spacecraft in the ECLIPJ2000 reference frame.
+  LABLAXIS: Yaw
--- a/imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
@@ -47,10 +47,10 @@ trigger_base: &trigger_base
 spice_base: &spice_base
   <<: *l1b_data_base
   VAR_TYPE: data
-  VALIDMIN: -9223372036854775808 # TODO: Find actual validmin
-  VALIDMAX: 9223372036854775807 # TODO: Find actual validmax
+  VALIDMIN: -180
+  VALIDMAX: 360
   DISPLAY_TYPE: time_series
-  UNITS: Radians
+  UNITS: Degrees
 
 # <=== Instrument Setting Attributes ===>
 trigger_mode:
@@ -343,6 +343,7 @@ current_neg2p5v_bus:
   UNITS: A
 
 # <=== Spice Data Attributes ===>
+# TODO: Get actual validmin and vaildmax for ephemeris attrs
 ephemeris_position_x:
   <<: *spice_base
   FIELDNAM: Position X
@@ -388,39 +389,27 @@ ephemeris_velocity_z:
 right_ascension:
   <<: *spice_base
   VALIDMIN: 0
-  VALIDMAX: 6.283185307179586
   FIELDNAM: Right Ascension
   CATDESC: Right ascension of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
   LABLAXIS: Right Ascension
 
 declination:
   <<: *spice_base
-  VALIDMIN: -1.5707963267948966
-  VALIDMAX: 1.5707963267948966
   FIELDNAM: Declination
   CATDESC: Declination of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
-  LABLAXIS: Current
+  LABLAXIS: Declination
 
-attitude_roll:
+spin_phase:
   <<: *spice_base
-  VALIDMIN: -3.141592653589793
-  VALIDMAX: 3.141592653589793
-  FIELDNAM: Roll
-  CATDESC: Roll angle (rotation around the X-axis) of the IMAP spacecraft in the ECLIPJ2000 reference frame.
-  LABLAXIS: Roll
+  VALIDMIN: 0
+  FIELDNAM: Spin Phase
+  CATDESC: IMAP Spin Phase
+  LABLAXIS: Spin Phase
 
-attitude_pitch:
+solar_longitude:
   <<: *spice_base
-  VALIDMIN: -1.5707963267948966
-  VALIDMAX: 3.141592653589793
-  FIELDNAM: Pitch
-  CATDESC: Pitch angle (rotation around the Y-axis) of the IMAP spacecraft in the ECLIPJ2000 reference frame.
-  LABLAXIS: Pitch
+  VALIDMAX: 180
+  FIELDNAM: Solar Longitude
+  CATDESC: Solar Longitude of the IMAP spacecraft
+  LABLAXIS: Solar Longitude
 
-attitude_yaw:
-  <<: *spice_base
-  VALIDMIN: -3.141592653589793
-  VALIDMAX: 3.141592653589793
-  FIELDNAM: Yaw
-  CATDESC: Yaw angle (rotation around the Z-axis) of the IMAP spacecraft in the ECLIPJ2000 reference frame.
-  LABLAXIS: Yaw
--- a/imap_processing/idex/idex_l1b.py
+++ b/imap_processing/idex/idex_l1b.py
@@ -18,7 +18,6 @@ import logging
 from enum import Enum
 from typing import Union
 
-import numpy as np
 import pandas as pd
 import xarray as xr
 
@@ -28,9 +27,10 @@ from imap_processing.spice.geometry import (
     SpiceBody,
     SpiceFrame,
     cartesian_to_spherical,
-    get_rotation_matrix,
+    get_spacecraft_spin_phase,
     imap_state,
-    rotation_matrix_to_euler_angles,
+    instrument_pointing,
+    solar_longitude,
 )
 from imap_processing.spice.time import j2000ns_to_j2000s
 from imap_processing.utils import convert_raw_to_eu
@@ -122,7 +122,7 @@ def idex_l1b(l1a_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
     )
     # Get spice data and save them as xr.DataArrays in the output. Spice data is not
     # used for calculations yet but are saved in the CDF for reference.
-    spice_data = get_spice_data(epoch_da.data, idex_attrs)
+    spice_data = get_spice_data(l1a_dataset, idex_attrs)
 
     trigger_settings = get_trigger_mode_and_level(l1a_dataset)
     if trigger_settings:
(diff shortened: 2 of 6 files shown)


imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -334,3 +341,86 @@ current_neg2p5v_bus:
...
+  CATDESC: Velocity Z positions for the IMAP spacecraft in the ECLIPJ2000 frame.
+  LABLAXIS: Velocity Z
+  UNITS: km/s
+
+right_ascension:
+  <<: *spice_base
+  VALIDMIN: 0
+  VALIDMAX: 6.283185307179586
+  FIELDNAM: Right Ascension
+  CATDESC: Right ascension of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
```
tmplummer, line 393: This is interesting nomenclature. So far, the few instruments that I have seen these values computed for are calling them Azimuth and Elevation instead of RA and Dec.

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -334,3 +341,86 @@ current_neg2p5v_bus:
...
+  LABLAXIS: Right Ascension
+
+declination:
+  <<: *spice_base
+  VALIDMIN: -1.5707963267948966
+  VALIDMAX: 1.5707963267948966
+  FIELDNAM: Declination
+  CATDESC: Declination of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
```
tmplummer, line 399: This is interesting nomenclature. So far, the few instruments that I have seen these values computed for are calling them Azimuth and Elevation instead of RA and Dec.

imap_processing/idex/idex_l1b.py
```diff
@@ -324,3 +337,62 @@ def compute_trigger_values(
...
+    idex_attrs : ImapCdfAttributes
+        CDF attribute manager object.
+
+    Returns
+    -------
+    dict
+        Spice array names and xr.DataArrays.
+    """
+    # 'epoch' is in nanoseconds since j2000
+    # convert epoch times to seconds since j2000
```
torimarbois: Seems like these two comments might be duplicates of each other?

imap_processing/idex/idex_l1b.py
```diff
@@ -324,3 +337,62 @@ def compute_trigger_values(
...
+    # Reference frame to the IDEX instrument frame)
+    rotation = get_rotation_matrix(et, SpiceFrame.ECLIPJ2000, SpiceFrame.IMAP_IDEX)
+    # In order yaw, pitch, roll in radians
+    euler_angles = rotation_matrix_to_euler_angles(rotation)
+
+    # Get position and velocity of IMAP
+    ephemeris = imap_state(et, observer=SpiceBody.SUN)
+    imap_position = ephemeris[:3]
+
+    # get right ascension and declination of the IMAP spacecraft
```
torimarbois: Little nitpicky - this comment seems unnecessary to me since the following function is named so specifically.

imap_processing/spice/geometry.py
```diff
@@ -644,3 +684,32 @@ def spherical_to_cartesian(spherical_coords: NDArray, degrees: bool = False) ->
     cartesian_coords = np.stack((x, y, z), axis=-1)
 
     return cartesian_coords
+
+
+def get_right_ascension_and_declination(coords: npt.NDArray) -> npt.NDArray:
```
tmplummer: I think that this provides the same functionality as the `cartesian_to_spherical` function. Sorry for any confusion that the function name caused. Perhaps we should change it to be explicit?

Double check the docstring and see if you agree. To test, you could try calling this function in place of `cartesian_to_spherical` in the test coverage for `cartesian_to_spherical`.

Other than that slight issue, this is really well written!

imap_processing/spice/geometry.py
```diff
@@ -466,6 +466,46 @@ def get_rotation_matrix(
...
+    "Factor a rotation matrix as a product of three rotations about
+        specified coordinate axes."
+    https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/m2eul.html
+
+    Parameters
+    ----------
+    rotation_matrix : np.ndarray
+        Either single 2d rotation matrix of shape, `(3,3)` or 3d array
+        of rotation matrices of shape `(n, 3, 3)`.
+    axes : SpiceFrame
```
torimarbois: This parameter should also be marked as 'Optional' here, right? And looks like the default is None? (could it be set to [3, 2, 1] in the function definition instead of within the function?)

tmplummer: Please update this to a list.
```suggestion
    axes : list, optional
```

imap_processing/spice/geometry.py
```diff
@@ -466,6 +466,46 @@ def get_rotation_matrix(
...
+        Indices of third, second, and first rotation axes.
```
tmplummer: Maybe specify that the indices are 1-based or that they must be in the set of values: [1, 2, 3].

imap_processing/spice/geometry.py
```diff
@@ -466,6 +466,46 @@ def get_rotation_matrix(
...
+
+    This is a vectorized wrapper around `spiceypy.m2eul`
+    "Factor a rotation matrix as a product of three rotations about
+        specified coordinate axes."
+    https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/m2eul.html
+
+    Parameters
+    ----------
+    rotation_matrix : np.ndarray
+        Either single 2d rotation matrix of shape, `(3,3)` or 3d array
```
torimarbois: I wonder if it would be a good idea to add error handling around the number of dimensions that `rotation_matrix` is passed in with? Like, if it's not 2d or 3d, it returns an error? That might be unnecessary if `@ensure_spice` was in front of the function, though. @subagonsouth would know!

imap_processing/spice/geometry.py
```diff
@@ -644,3 +684,32 @@ def spherical_to_cartesian(spherical_coords: NDArray, degrees: bool = False) ->
...
+    """
+    Convert rectangular coordinates to range, right ascension, and declination.
+
+    This is a vectorized wrapper around `spiceypy.recrad`
+    https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/recrad_c.html
+
+    Parameters
+    ----------
+    coords : np.ndarray
+        Cartesian (rectangular) coordinates with shape `(3)` or `(n, 3)`,
```
torimarbois: Similar to my comment for the `rotation_matrix_to_euler_angles` function, I wonder if it would be a good idea to add error handling around the number of dimensions that coords is passed in with? Like, if it's not 1d or 2d, it returns an error?

imap_processing/tests/spice/test_geometry.py
```diff
@@ -357,6 +359,28 @@ def test_get_rotation_matrix(furnish_kernels):
         assert rotation.shape == (10, 3, 3)
 
 
+def test_rotation_matrix_to_euler_angles(furnish_kernels):
+    """rotation_matrix_to_euler_angles()."""
```
torimarbois: I think you might be missing a "Tests " at the beginning of this doc string.

torimarbois (commented):
A few comments about documentation/comments and error-handling! Nice job, these are cool functions.

imap_processing/idex/idex_l1b.py
```diff
@@ -324,3 +337,62 @@ def compute_trigger_values(
...
+    -------
+    dict
+        Spice array names and xr.DataArrays.
+    """
+    # 'epoch' is in nanoseconds since j2000
+    # convert epoch times to seconds since j2000
+    et = j2000ns_to_j2000s(epoch)
+    # Get IDEX rotation matrix (matrix to get transformation from the 'ECLIPJ2000'
+    # Reference frame to the IDEX instrument frame)
+    rotation = get_rotation_matrix(et, SpiceFrame.ECLIPJ2000, SpiceFrame.IMAP_IDEX)
```
tmplummer: This is interesting. You have the comment exactly correct about what rotation matrix is being retrieved. It does not provide the correct rotation matrix to use below in order to compute the yaw, pitch, and roll of IDEX pointing because the IDEX boresight vector is: `[0, 1, 0]` as expressed in the IDEX instrument frame (IMAP_IDEX).

imap_processing/tests/spice/test_geometry.py
```diff
@@ -357,6 +359,28 @@ def test_get_rotation_matrix(furnish_kernels):
...
+    kernels = [
+        "naif0012.tls",
+        "imap_wkcp.tf",
+        "imap_science_0001.tf",
+        "sim_1yr_imap_attitude.bc",
+        "sim_1yr_imap_pointing_frame.bc",
+    ]
```
tmplummer: SPICE shouldn't need any kernels to calculate Euler angles from a rotation matrix, this is strictly math. The kernels and the `with furnish_kernels()` context manager below can be removed.

imap_processing/tests/spice/test_geometry.py
```diff
@@ -357,6 +359,28 @@ def test_get_rotation_matrix(furnish_kernels):
...
+    rotation = np.array([[0, -1, 0], [1, 0, 0], [0, 0, 1]])
+
+    with furnish_kernels(kernels):
+        # test with one rotation matrix
+        euler_angles = rotation_matrix_to_euler_angles(rotation, axes=[3, 2, 1])
+        assert euler_angles.shape == (3,)
+        # Test with multiple rotation matrices
+        rotation = np.tile(rotation, (10, 1, 1))
+        euler_angles = rotation_matrix_to_euler_angles(rotation, axes=[3, 2, 1])
+        assert euler_angles.shape == (10, 3)
```
tmplummer: Great test coverage of both cases!

tmplummer (commented):
Really nice work! Just a couple of comments to double check.

imap_processing/spice/geometry.py
tmplummer: I suppose some error checking on the input dimension couldn't hurt. No need for `@ensure_spice` as SPICE doesn't need any kernels to be furnished to do these calculations.

imap_processing/idex/idex_l1b.py
lacoak21: They are not but definitely can be less wordy!

imap_processing/spice/geometry.py
lacoak21: Oops thanks good catch. I think Ill keep it in the function because im pretty sure its best practice to have a default parameter be immutable.

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: From my understanding RA and Dec are slightly different than Azimuth and Elevation. RA and dec seems to be global coords that do not change from any observing points on earth vs Azimuth and Elevations depends on the location of the reference point? IDEX specifically asked for RA and dec but I am definitely inexperienced in this area (as well as spice) and could be missing something big here.

imap_processing/spice/geometry.py
lacoak21: Similar to my comment above this seems to be doing something slightly different. Seems like `cartesian_to_spherical` uses spice.[recsph ](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/recsph_c.html) to test it while mine uses spice.[recrad](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/recrad_c.html). When I plug in my function instread of `cartesian_to_spherical` it fails ( I tried making sure they are both in the same units). Seems like you are able to convert from these spherical coords to RA and DEC but im not sure how. Again,  I might be confused about something here. 

I am going to try and look into this more!

imap_processing/spice/geometry.py
lacoak21: Good call

imap_processing/idex/idex_l1b.py
lacoak21: Ah ok this makes a lot of sense!! So in other words getting the rotation matrix to the IDEX frame is not enough because that provides no context of where it is actually pointing but when we take into account the boresight vector AND the rotation matrix we can get the attitude of the actual pointing of IDEX?  

Is that what you meant? If so, could I use `instrument_pointing` to do that instead?

imap_processing/tests/spice/test_geometry.py
lacoak21: I took a leaf out of your book! (test_get_rotation_matrix)

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
tmplummer: Ok, I was wondering about that. You will have to ask for clarification from IDEX. I am far from an expert on some of this terminology as well. If IDEX wants celestial RA and Dec, we can do that. I was initially going to comment that RA and Dec are not in the ECLIPJ2000 reference frame, they are in the J2000 or equatorial reference frame. Looking at the NAIF SPICE documentation of the `recrad` function had me second guessing that initial reaction because it just specifies that RA is the angular distance from the prime meridian and declination is angle from the xy-plane. Those angles can be computed in any reference frame, but I suspect that using RA and Dec has the implied connotation that they must be computed in the equatorial reference frame.

imap_processing/idex/idex_l1b.py
tmplummer: Assuming that IDEX wants the yaw, pitch, and roll of the IDEX boresight then yes, but `instrument_pointing` gives a Cartesian pointing vector and I don't yet know how to convert that to yaw, pitch, and roll.

imap_processing/spice/geometry.py
tmplummer: I tested it the other direction by putting the following code in the test for `cartesian_to_spherical`:
```
    for point in cartesian_points:
        r, az, el = cartesian_to_spherical(point)
        range, ra, dec = spice.recrad(point)
        np.testing.assert_allclose(r, range, atol=1e-5)
        np.testing.assert_allclose(az, np.degrees(ra), atol=1e-5)
        np.testing.assert_allclose(el, np.degrees(dec), atol=1e-5)
```
The test passed with this implementation. If you find the same, I would suggest modifying the `test_cartesian_to_spherical` test to use this implementation rather that the current one that requires modifying the output from `spiceypy.recsph` to match our implemented spherical coordinate definition.

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: OK thanks Tim - ill ask for more clarification on this.

imap_processing/spice/geometry.py
lacoak21: Oh nice! Thanks for writing that out - it worked for me as well.

imap_processing/idex/idex_l1b.py
lacoak21: OK Ill double check with IDEX about that but I would imagine they want it for the boresight as well.

imap_processing/spice/geometry.py
```diff
@@ -466,6 +466,46 @@ def get_rotation_matrix(
...
+    if not axes:
+        axes = [3, 2, 1]
+
+    # If rotation_matrix is 2d, add another dimension
+    while rotation_matrix.ndim < 3:
+        rotation_matrix = np.expand_dims(rotation_matrix, axis=0)
+
+    euler_angles = np.array([spice.m2eul(m, *axes) for m in rotation_matrix])
+    # Return array of euler angles and remove the first dimension if it is 1.
+    return np.squeeze(euler_angles)
```
laspsandoval: Will the first dimension always be 1?

imap_processing/spice/geometry.py
```diff
@@ -644,3 +684,32 @@ def spherical_to_cartesian(spherical_coords: NDArray, degrees: bool = False) ->
...
+        Returns np.ndarray containing range, right ascension and declination values.
+        If the input coords are shape `(3)`, the output array is of shape `(3)` and
+        if the input coords are shape `(n, 3)`,
+        the output array is of shape `(n, 3)` where n matches the number sets in coords.
+    """
+    # If coords is 1d, add another dimension
+    while coords.ndim < 2:
+        coords = np.expand_dims(coords, axis=0)
+
+    array = np.array([np.array(spice.recrad(coords)) for coords in coords])
```
laspsandoval: I think that you can remove the redundant array?
array = np.array([spice.recrad(coord) for coord in coords])

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -42,9 +42,16 @@ l1b_target_base: &l1b_target_base
...
   DISPLAY_TYPE: no_plot
-  CATDESC: ""
   UNITS: " "
 
+spice_base: &spice_base
+  <<: *l1b_data_base
+  VAR_TYPE: data
+  VALIDMIN: -9223372036854775808 # TODO: Find actual validmin
+  VALIDMAX: 9223372036854775807 # TODO: Find actual validmax
+  DISPLAY_TYPE: time_series
```
laspsandoval, line 52: I would think that the VALIDMIN and VALIDMAX would be 0 to 2pi, respectively, if the units are in radians?

laspsandoval (commented):
Looks good. Just a few comments.

imap_processing/spice/geometry.py
lacoak21: Not necessarily, but np.squeeze should only remove a dimension if it is one and otherwise it leaves it alone. The other dimension should never be one so this should be safe. ( I did end up removing this function)

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: OK I switched the default to degrees but applied the same idea.

imap_processing/idex/idex_l1b.py
```diff
@@ -324,3 +337,61 @@ def compute_trigger_values(
...
+    """
+    # convert 'epoch' from nanoseconds to seconds since j2000
+    et = j2000ns_to_j2000s(l1a_dataset["epoch"].data)
+    # Get 'shcoarse' (Mission Elapsed Time)
+    met = l1a_dataset["shcoarse"].data
+    # Get spacecraft spin phase in degrees
+    imap_spin_phase = get_spacecraft_spin_phase(query_met_times=met, degrees=True)
+    # Get position and velocity of IMAP in ecliptic frame
+    ephemeris = imap_state(et, observer=SpiceBody.SUN)
+    # Get Idex pointing in the j200 equatorial frame
```
tmplummer: Typo
```suggestion
    # Get Idex pointing in the j2000 equatorial frame
```

tmplummer (approved):
Very nice job. This all looks great!

imap_processing/spice/geometry.py
```diff
@@ -158,6 +158,35 @@ def get_spin_data() -> pd.DataFrame:
     return spin_df
 
 
+def get_spacecraft_spin_phase_angle(
```
lacoak21: @tech3371  Does this function look good for issue #1306?

tech3371: Yeah, that looks good but I am concerned what if user want their data in instrument frame but return value in degree also.

lacoak21: OK I just reworked the function to receive an array of spin phases instead.

tech3371 (approved):
`get_spin_angle()` looks good to me now

PR #1288, opened 2025-01-23
Ultra l1b - extendedspin data structure

### LINKED ISSUE #1283: Ultra l1b - extended spin data structure
### Summary

We need to add onto the first pass:
**Dictionary that will seed the cdf**
### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
This updates the structure of the ultra l1b extended spin table. However, one small change would propagate other changes in other files. Any major changes in files are listed as "Focus on these" and those are the ones you should focus on.

## Updated Files
**Focus on these**
- extended_spin.py
   - Added structure of what is expected for this data product. 
- ultra_l1b.py
   - The extended_spin data product is dependent on the de data product. And the culling and badtimes data products are both dependent on the extended_spin data product. So I changed the logical involved here.
- ultra_l1b_culling.py
   - I had to change the structure of quality flags for creating an xarray with two coordinates (spin and energy)
- ultra_l1_utils.py
   - Change the logic for creating the dataset based on which data product is being created.

**Not on these**
- quality_flags.py
   - Name change of the flag
   - Define quality_ena_rates attributes
- badtimes.py and cullingmask.py
   - Changed the data type for the input. Added data version as an input. 
- calculate_de.py, histogram.py, pset.py, ultra_l1c.py
   - added data_version as an input.
- conftest.py
   - added test data
  
## Testing
- test_de.py
- test_ultra_l1b.py
- test_ultra_l1b_culling.py
- test_ultra_l1c.py

### DIFF
--- a/imap_processing/ultra/l1b/badtimes.py
+++ b/imap_processing/ultra/l1b/badtimes.py
@@ -5,14 +5,14 @@ import xarray as xr
 from imap_processing.ultra.utils.ultra_l1_utils import create_dataset
 
 
-def calculate_badtimes(extended_spin_dict: dict, name: str) -> xr.Dataset:
+def calculate_badtimes(extended_spin_dict: xr.Dataset, name: str) -> xr.Dataset:
     """
     Create dataset with defined datatypes for Badtimes Data.
 
     Parameters
     ----------
-    extended_spin_dict : dict
-        L1b data dictionary.
+    extendedspin_dataset : xarray.Dataset
+        Dataset containing the data.
     name : str
         Name of the dataset.
 
--- a/imap_processing/ultra/l1b/badtimes.py
+++ b/imap_processing/ultra/l1b/badtimes.py
@@ -5,7 +5,9 @@ import xarray as xr
 from imap_processing.ultra.utils.ultra_l1_utils import create_dataset
 
 
-def calculate_badtimes(extended_spin_dict: xr.Dataset, name: str) -> xr.Dataset:
+def calculate_badtimes(
+    extendedspin_dataset: xr.Dataset, name: str, data_version: str
+) -> xr.Dataset:
     """
     Create dataset with defined datatypes for Badtimes Data.
 
@@ -15,12 +17,20 @@ def calculate_badtimes(extended_spin_dict: xr.Dataset, name: str) -> xr.Dataset:
         Dataset containing the data.
     name : str
         Name of the dataset.
+    data_version : str
+        Version of the data.
 
     Returns
     -------
     badtimes_dataset : xarray.Dataset
         Dataset containing the data.
     """
-    badtimes_dataset = create_dataset(extended_spin_dict, name, "l1b")
+    badtimes_dict = {}
+    badtimes_dict["spin_number"] = extendedspin_dataset["spin_number"]
+    badtimes_dict["median_rate_energy"] = extendedspin_dataset["median_rate_energy"]
+
+    # TODO: add more data to badtimes_dict.
+
+    badtimes_dataset = create_dataset(badtimes_dict, name, "l1b", data_version)
 
     return badtimes_dataset
--- a/imap_processing/ultra/l1b/ultra_l1b.py
+++ b/imap_processing/ultra/l1b/ultra_l1b.py
@@ -28,7 +28,7 @@ def ultra_l1b(data_dict: dict, data_version: str) -> list[xr.Dataset]:
     instrument_id = 45 if any("45" in key for key in data_dict.keys()) else 90
 
     if (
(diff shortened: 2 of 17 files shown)


imap_processing/ultra/l1b/badtimes.py
```diff
@@ -5,22 +5,32 @@
...
+        Version of the data.
 
     Returns
     -------
     badtimes_dataset : xarray.Dataset
         Dataset containing the data.
     """
-    badtimes_dataset = create_dataset(extended_spin_dict, name, "l1b")
+    badtimes_dict = {}
+    badtimes_dict["spin_number"] = extendedspin_dataset["spin_number"]
```
nkerman, line 29: I'm a bit confused about what this is doing. It seems to be copying over all spins from the extendedspin_dataset,  rather than actually filtering to some set of bad spins. 

Also at present, I don't understand the distinction between the culling mask dataset and the bad times dataset.

I imagine they will be fleshed out with more variables later, but do they both need to be created from scratch like this? Would it make more sense to copy one from another?

I might be missing something/this might just be meant as a placeholder.

nkerman: Ah, nevermind, I see that this is still a placeholder, and that you're focused on the other files in this PR.

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -99,32 +99,17 @@ def flag_spin(
...
-    )
+    quality_flags = np.full(hist.shape, ImapRatesUltraFlags.NONE.value, dtype=np.uint16)
 
-    appended_spin = np.empty(0, dtype=np.uint16)
-    appended_energy = np.empty(0, dtype=np.float64)
+    bin_edges = np.array(UltraConstants.CULLING_ENERGY_BIN_EDGES)
+    energy_midpoints = (bin_edges[:-1] + bin_edges[1:]) / 2
+    spin = np.unique(spin)
 
     for energy_idx in range(hist.shape[0]):
```
nkerman: LGTM! This could still be done without a loop probably\*, but there aren't a lot of energies, so it's not worth much thought/effort. 

\* something like the following, but I haven't tried it/don't know the exact dims of all the arrays at this point. 

```Python
# Indices where the counts exceed the threshold, done separately by energy (axis 0)
indices = spin_count_rates > UltraConstants.COUNT_RATES_THRESHOLDS[:, np.newaxis]
quality_flags[:, indices] |= ImapRatesUltraFlags.HIGHCOUNTS.value
```

imap_processing/ultra/utils/ultra_l1_utils.py
```diff
@@ -26,36 +30,64 @@ def create_dataset(data_dict: dict, name: str, level: str) -> xr.Dataset:
...
+
+    else:
+        epoch_time = xr.DataArray(
+            data_dict["epoch"],
+            name="epoch",
+            dims=["epoch"],
+            attrs=cdf_manager.get_variable_attributes("epoch"),
+        )
+        if name in {f"imap_ultra_l1b_{x}sensor-de" for x in [45, 90]}:
+            coords = {"epoch": epoch_time, "component": ["vx", "vy", "vz"]}
```
nkerman: Might be easier to work based on the suffix - e.g.:

```
if "sensor-de" in name:
```

nkerman (approved):
LGTM!

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -27,38 +27,36 @@ def ultra_l1b(data_dict: dict, data_version: str) -> list[xr.Dataset]:
     output_datasets = []
     instrument_id = 45 if any("45" in key for key in data_dict.keys()) else 90
 
-    if f"imap_ultra_l1a_{instrument_id}sensor-rates" in data_dict:
+    if (
```
tmplummer, line 30: So you will be doing the L1B de, extendedspin, culling_mask, and badtimes processing all in one batch job with one set of input dependencies? I would have thought that each of those would have their own entry in the dependency table and that you would specify that, for example, L1B extended spin is dependent upon L1B DE.

tmplummer (approved):
I really only reviewed the files in your list of files to focus on. Just had the one question about processing workflow everything else makes sens.

imap_processing/ultra/l1b/ultra_l1b.py
laspsandoval: Yes. I can see it going either way. To me it seemed simpler to keep them together.

PR #1291, opened 2025-01-23
MNT: Add direct casting for IntEnums to Numpy

### LINKED ISSUE #1290: BUG - Casting rules for Numpy integer types are failing on Hi
### Description of the issue

Numpy 2.2 is getting more strict with integer casting. A raw Python integer gets cast to np.int64, which then creates an error because we are bitwise combining it with an np.uint8 variables.

Install numpy 2.2+ and run the test suite.

No failures

Hi test failures

### Code Snippet:

```python
Code
```

_No response_

greglucas:
# Change Summary

The casting rules are becoming more strict in Numpy, this casts the IntEnum directly to np.uint8 to do the combinations better.

closes #1290


tmplummer (approved):
Thanks for finding this and fixing it. ~~I'm curious how you found this issue?~~ Maybe the ticket explained. Just install numpy 2.2 and run the tests.

greglucas:
Yeah, I like living with the latest versions of things 😂 But really, I think I just recreated everything recently when I came back.

PR #1293, opened 2025-01-23
1170 hi l1c define mapping from coincidence to calibration product

### LINKED ISSUE #1170: Hi L1C - Define mapping from coincidence to calibration product
### Algorithm Description:

In PSET processing, we need to know how to filter events by coincidence type for each calibration product. Define that with help from Paul and implement a mapping to use in code.

- Make sure to adjust the VALIDMAX in imap_processing/cdf/config/imap_hi_variable_attrs.yaml if it is not 4.
- Make sure to replace the hard coded value with getting the number of calibration products from the mapping in imap_processing/hi/l1c/hi_l1c.py::121

Calibration products have two ways that they filter events:

1. By coincidence type
2. By TOFs 

As part of this ticket, I need to design a calibration file that defines the filtering of DEs into calibration products and has the ability to be time varying. i.e. A set of date/time entries that specify the coincidence types and valid TOFs for each calibration product.

Notional starting definition:
0:  all quads plus all triples except BC1C2, no TOF filters
1:  BC1C2, and all doubles except BC, no TOF filters

tmplummer:
# Change Summary

## Overview
This PR adds a proposed ancillary file and associated class which defines how Hi DEs are mapped to calibration products. I am seeking input from Hi to make sure the file fits the need before implementing additional access functions for the CalibrationProductConfig class. I don't yet know if this ancillary file will be of the Universal or Time Window type as laid out in the following documentation page: https://imap-processing.readthedocs.io/en/latest/data-access/calibration-files.html

## New Dependencies
None

## New Files
- imap_processing/tests/hi/test_data/l1/imap_hi_pset-calibration-prod-config_20240101_v001.yaml
   - Defines the calibration file format and is used in tests

## Deleted Files
None

## Updated Files
- .gitignore
   - Fix the pattern match for the Earth combined .bpc kernel
   - Increase VALIDMAX in case of more calibration products
- imap_processing/cli.py
   - Temporarily, inject the test calibration file as a dependency.
   - Add the `CalibrationProductConfig` class and use it in PSET processing
   - Add test coverage for the `CalibrationProductConfig` class.

Closes: #1170


imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -45,8 +50,8 @@ def hi_l1c(dependencies: list, data_version: str) -> xr.Dataset:
 
     # TODO: I am not sure what the input for Goodtimes will be so for now,
     #    If the input is an xarray Dataset, do pset processing
-    if len(dependencies) == 1 and isinstance(dependencies[0], xr.Dataset):
-        l1c_dataset = generate_pset_dataset(dependencies[0])
+    if len(dependencies) == 2 and isinstance(dependencies[0], xr.Dataset):
```
greglucas, line 52: Do we need to assert on the instance type of all dependencies, or are we sure these are going to be in the right order when we request them?

greglucas (commented):
I am not up to speed on how these calibrations will work, how often they'll be updated etc. but this seems like a tough format to work with to me at first glance. yaml is nice for setting up configurations, but I feel like for this many values it can be a bit harder to follow than csv files. I'm also unclear about what index relates to/how it would be used, even with your nice documentation.

My naive recommendation would be to use multiple csv files with different product names in the "definition" field. I think I'm also uncertain how the nesting/depth occurs here so don't know whether you'd name your products `cal-idx0` or `cal-tof-ab`, which may be why you need to avoid csv and the flatter structure and use yaml like you are here?

tmplummer:
> I am not up to speed on how these calibrations will work, how often they'll be updated etc. but this seems like a tough format to work with to me at first glance. yaml is nice for setting up configurations, but I feel like for this many values it can be a bit harder to follow than csv files. I'm also unclear about what index relates to/how it would be used, even with your nice documentation.
> 
> My naive recommendation would be to use multiple csv files with different product names in the "definition" field. I think I'm also uncertain how the nesting/depth occurs here so don't know whether you'd name your products `cal-idx0` or `cal-tof-ab`, which may be why you need to avoid csv and the flatter structure and use yaml like you are here?

@greglucas I initially tired to imagine how I could put this data into a csv. The problematic part is that this essentially needs to represent 3 dimensional data. All direct events get combined into "calibration products" by filtering on coincidence type (which detectors were hit by the event) and the various times-of-flight. For example calibration product 0 contains all DEs of "ABC1C2", "ABC1", and "AC1C2" coincidence types where the TOFs are in a range specific to the ESA step that the DE is associated with. These parameters (coincidence type and TOF ranges) need to be allowed to be time varying.
I think that the time varying part can be taken care of by specifying the parameters in a time window ancillary file (I didn't know about this when I implemented this code: https://imap-processing.readthedocs.io/en/latest/data-access/calibration-files.html#file-overview) but that still leaves the challenge of representing 2D parameters in whatever file format is used.

greglucas:
> but that still leaves the challenge of representing 2D parameters in whatever file format is used

How many rows/columns would there be? I'm wondering if you'd want to make something like 20-30 calibration csv files to take care of that 2D version.

Another question is whether you would want to put the 3D quantity into a CDF product that you read in instead of these human-readable formats. I believe MAG is doing that and giving a calibration factor for every single time-varying vector quantity. It would be easier machine-loadable, but less user facing.

imap_processing/tests/hi/test_hi_l1c.py
```diff
@@ -95,3 +121,40 @@ def test_pset_geometry(mock_frame_transform, mock_geom_frame_transform, sensor_s
...
+class TestCalibrationProductConfig:
+    """A class containing test coverage for CalibrationProductConfig"""
+
+    def test_invalid_top_level_config(self):
+        """Test coverage for an invalid top level configuration dictionary"""
+        with pytest.raises(KeyError, match="Missing required key*"):
+            hi_l1c.CalibrationProductConfig([{"foo": 1}])
+
+    def test_invalid_cal_prod_config(self):
+        """Test coverage for an invalid top level configuration dictionary"""
```
lacoak21: Wouldn't this be a test for an invalid second level config? (Docstring might need updating)

imap_processing/tests/hi/test_hi_l1c.py
```diff
@@ -95,3 +121,40 @@ def test_pset_geometry(mock_frame_transform, mock_geom_frame_transform, sensor_s
...
+    def test_from_yaml(self, hi_test_cal_prod_config_path):
+        """Test class factory function from YAML file."""
+        config = hi_l1c.CalibrationProductConfig.from_yaml(hi_test_cal_prod_config_path)
+        assert config.number_of_products == 2
+
+    @mock.patch(
+        "imap_processing.hi.l1c.hi_l1c.CalibrationProductConfig._validate",
+        side_effect=KeyError("Test message"),
+    )
+    def test_from_yaml_exceptions(self, mock_validate, hi_test_cal_prod_config_path):
```
lacoak21: Cool usage of side_effect!

imap_processing/tests/hi/test_data/l1/imap_his_pset-calibration-prod-config_20240101_v001.csv
greglucas, line 1: I personally find this easier to read than yaml, but this is probably worth you and Paul / whoever else making sure you're OK with whatever format you choose because I assume these are going to be computer generated by someone else who should probably agree to whatever format you decide upon.

imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -271,3 +291,88 @@ def pset_geometry(pset_et: float, sensor_str: str) -> dict[str, xr.DataArray]:
...
+            Location of the Calibration Product configuration CSV file.
+
+        Returns
+        -------
+        dataframe : pandas.DataFrame
+            Validated calibration product configuration data frame.
+        """
+        return pd.read_csv(
+            path,
+            index_col=False,
```
greglucas: Would it be helpful to make this a multi-index, so you can lookup based on `(cal_prod, esa_step)` later on? If you read it in as a sequence a multi-index will be created.
```suggestion
            index_col=[0, 1],
```

imap_processing/tests/hi/test_data/l1/imap_his_pset-calibration-prod-config_20240101_v001.csv
tmplummer: Yes, I ended up liking this format better as well. I'll go over it with Paul today (1/29/2025).

imap_processing/hi/l1c/hi_l1c.py
tmplummer: Yes, great suggestion. Thanks!

imap_processing/tests/hi/test_data/l1/imap_his_pset-calibration-prod-config_20240101_v001.csv
tmplummer: Paul has approved of this format for the calibration file.

greglucas (approved):
Nice! If you and Paul agree on this format, this looks good to me.

PR #1295, opened 2025-01-24
1292 bug   epoch time should be tdt instead of tdb

### LINKED ISSUE #1292: BUG - Epoch time should be TDT instead of TDB
### Description of the issue

Currently, the `met_to_j2000ns` function is being used to compute values stored in CDF files with the CDF datatype CDF_TIME_TT2000. The implementation of `met_to_j2000ns` provides TDB nanoseconds but CDF_TIME_TT2000 is Terrestrial Time nanoseconds.

This ticket needs to fix that bug and add a function for converting TT nanoseconds to floating point TDB seconds (ET) for use with SPICE.

_No response_

_No response_

_No response_

### Code Snippet:

```python

```

_No response_

tmplummer:
# Change Summary

## Overview
Fix issue with spice time functions reporting TDB time instead of TT for CDF epoch values. This PR had to touch lots of files due to the function renaming. To make review easier, I suggest that you review in two segments. In the File Changes tab, select the first commit to view just the changes that were made to the `imap_processing.spice.time.py` module and associated tests. Once finished with that, view the other commit that pushed the function and global variable name changes in that module to all of the locations that they were used.

## Updated Files
I'm not going to list all of the files to save time. The important changes are detailed below.
   - Renamed global variable in spice.time from `J2000_EPOCH` to `J2000_TT_EPOCH` and updated value to TT timescale.
   - Renamed `met_to_j2000ns` function to `met_to_tt_ns` and modified it to return TT nanoseconds rather than TDB nanoseconds.
   - Renamed `j2000ns_to_j2000s` function to `tt_ns_to_et` and added conversion from TT to ET.
   - Added `sct_to_tt` function to aid conversion from spacecraft clock ticks directly to TT with one for loop instead of chaining together multiple looping functions.
   - Updated tests to correctly validate above changes.
   - Added functional test coverage for the new `sct_to_tt` function.

Closes: #1292


imap_processing/spice/time.py
```diff
@@ -40,11 +40,11 @@ def met_to_sclkticks(met: npt.ArrayLike) -> npt.NDArray[float]:
     return np.asarray(met, dtype=float) / TICK_DURATION
 
 
-def met_to_j2000ns(
+def met_to_tt_ns(
```
maxinelasp: I feel like you could keep the function name the same and just update the comment. I think most people think of the timestamp as "J2000" so it might be confusing to change the name. Also, TT has a different epoch (I think? it sounds like it shares the 1977 epoch with TDB and TAI from wikipedia) so J2000 is still a relevant piece of information (since we will be using TT seconds counted from the J2000 epoch.)

All that is pretty nitpick-y... just wanted to call it out as an option. It's also nice to formally define TT here because it differs from SPICE defaults, so it's also ok with me to leave the change in place.

imap_processing/spice/time.py
```diff
@@ -67,27 +68,35 @@ def met_to_j2000ns(
...
+@typing.no_type_check
+@ensure_spice
+def tt_ns_to_et(tt_ns: npt.ArrayLike) -> npt.NDArray[float]:
     """
-    Convert the J2000 epoch nanoseconds to J2000 epoch seconds.
+    Convert TT J2000 epoch nanoseconds to TDB J2000 epoch seconds.
 
-    The common CDF coordinate `epoch` stores J2000 nanoseconds. SPICE requires
-    J2000 seconds be used. This is a common function to do that conversion.
+    The common CDF coordinate `epoch` stores terrestrial time (TT) J2000
```
maxinelasp, line 80: Nice explanation!

imap_processing/hi/l1b/hi_l1b.py
```diff
@@ -288,7 +288,7 @@ def de_nominal_bin_and_spin_phase(dataset: xr.Dataset) -> dict[str, xr.DataArray
     # nominal_bin is the index number of the 90 4-degree bins that each DE would
     # be binned into in the histogram packet. The Hi histogram data is binned by
     # spacecraft spin-phase, not instrument spin-phase, so the same is done here.
-    met_query_times = j2000ns_to_j2000s(dataset.event_met.values)
+    met_query_times = tt_ns_to_et(dataset.event_met.values)
     imap_spin_phase = get_spacecraft_spin_phase(met_query_times)
```
maxinelasp: Is this `get_spacecraft_spin_phase` in ET then? Does it specify in the docs for that method (and all the SPICE geometry methods) that the inputs should use the `tt_ns_to_et` method? It might be overkill to be that specific, but time conversions are hard.

imap_processing/tests/ultra/unit/test_ultra_l1a.py
```diff
@@ -196,7 +196,7 @@ def test_xarray_events(decom_test_data, decom_ultra_aux, events_test_path):
 
     # # Spot check metadata data and attributes
     # j2000_time = (
-    #     np.datetime64("2024-02-07T15:28:37.184000", "ns") - J2000_EPOCH
+    #     np.datetime64("2024-02-07T15:28:37.184000", "ns") - J2000_TT_EPOCH
```
maxinelasp: What if we move these to point to the `J2000_TT_EPOCH` constant variable?

maxinelasp (approved):
Thanks for the quick update! Overall looks great, I had a few minor suggestions that you can take or leave.

imap_processing/spice/time.py
```diff
@@ -165,6 +174,38 @@ def _sct2e_wrapper(
...
+
+    Returns
+    -------
+    terrestrial_time: np.ndarray[float]
+        Terrestrial time, seconds past J2000.
+    """
+    if isinstance(sclk_ticks, Collection):
+        return np.array(
+            [spice.unitim(spice.sct2e(IMAP_SC_ID, s), "ET", "TT") for s in sclk_ticks]
+        )
```
greglucas, line 204: Is it worth inverting the `np.vectorize` you have in the function above (that one goes from TT to ET)? This is a for loop within Python currently.

imap_processing/hi/l1b/hi_l1b.py
greglucas: This seems like it should be the `met_to*` since you're inputting the met time. Or perhaps the event_met should be renamed?

greglucas (commented):
It seems like we have several functions with slightly different names all doing similar conversions. Would it be beneficial to combine these into a single function with keyword argument inputs so that we are just referencing this one function with different parameters as we need them? Just a thought I'm writing down in case this appeals to you, no need to do it here.

```python
def convert_time(my_time_in_whatever_frame, input_frame="TT", output_frame="ET"):
    # Some logic to select your input functions
    ....
    # Vectorize the spiceypy calls
    np.vectorize(func, ...)
```

imap_processing/spice/time.py
tmplummer: I'm happy to do that although my understanding of `np.vectorize` is that it is a python loop anyway. The main thing that I was trying to do here was avoid two for loops by doing these two steps of converting from s/c ticks to ET and then ET to TT each in their own loop. I could use `np.vectorize` on a `functools.partial` of the two chained spiceypy functions.

Side question: Do you think all of the manually "vectorized" functions in this module would benefit from using `np.vectorize`? My only complaint is that `np.vecotrize` returns an array with shape=() when a scalar is input.

greglucas: You are right, the documentation even says there is basically no speedup.
https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html

>The [vectorize](https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html#numpy.vectorize) function is provided primarily for convenience, not for performance. The implementation is essentially a for loop.

So I guess the answer is maybe it isn't worth it and just write the `np.array([value for value in list])` loop like you already are. I think the reason to use it is for the broadcasting types then? So if you wanted your function to accept a 0d, 1d, 2d arrays without having to do the for loops yourself for those cases.

imap_processing/spice/time.py
tmplummer: I don't find this nitpick-y. I think naming is important. The main reason that I wanted to change the name was to note the timescale in the function name. This is the first project that I have had to worry about multiple uniform timescales so I thought that it should be explicit in the function name. Good point about the epoch. What about `met_to_ttj2000ns`?

imap_processing/hi/l1b/hi_l1b.py
tmplummer: Good catch to both of you. I had been using the `j2000ns_to_j2000s` to convert from nanoseconds to seconds so this is indeed a bug.

tmplummer:
> It seems like we have several functions with slightly different names all doing similar conversions. Would it be beneficial to combine these into a single function with keyword argument inputs so that we are just referencing this one function with different parameters as we need them? Just a thought I'm writing down in case this appeals to you, no need to do it here.

@greglucas I agree. I'll write a new ticket to explore the idea of one time conversion function to rule them all. Brandon has done something quite complicated in the curreyer library to accomplish that goal. I am undecided if I like it or not.

imap_processing/spice/time.py
maxinelasp: I like that!

imap_processing/tests/ultra/unit/test_ultra_l1a.py
tmplummer: Not sure I understand. I don't know exactly what these commented out lines are for in this test, but they are pointing to the `J2000_TT_EPOCH` constant variable.

PR #1300, opened 2025-01-24
Move Spatial Utils, Fix bug in Spatial Utils with indexing of the azimuth and elevation grids

### LINKED ISSUE #1260: Move `build_az_el_grid` function and other spatial utils from `ultra.utils.spatial_utils` to `ena_maps`
### Algorithm Description:

As requested by @subagonsouth [here](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1258#discussion_r1913740873), simply move the build_az_el_grid function from `ultra.utils.spatial_utils` to `ena_maps.map_utils`.

### LINKED ISSUE #1297: Fix indexing in rectangular az_el grids
### Requirements:

Currently az and el grids are created with np meshgrid, using the "xy" indexing style. 
This causes issues when trying to find the nearest 1D index on the unwrapped az/el grids, based on a point's az and el, digitized in 1D. 

In effect, in the current configuration, the grids would need to be unwrapped in one order (C or F) and then the `ravel_multi_index` step performed in the other (F or C). 

Switch the indexing to "ij" on the meshgrid to resolve this issue.

### LINKED ISSUE #1298: Return az/el bin edges along with bin centers in `build_az_el_grid` utility function
Minor change: simply return the edges of the bins as well as their centers. The edges are needed for the digitization step of matching rectangular indices.

### LINKED ISSUE #1299: Allow spatial_utils.build_az_el_grid to return non-reversed elevation values
spatial_utils.build_az_el_grid was built with Ultra MATLAB code in mind, which reversed the order of elevation values from the increasing `(-90, -89,...89, 90)` to `(90, 89,...-89, -90)`. Allow this function to do either, set with a boolean flag.

nkerman:
# Change Summary

## Overview
* Moves the spatial_utils to a more generic location (Closes #1260 )
* In spatial_utils, changes the ordering of the az/el meshgrids from 'xy' to 'ij', which is more natural for ENA map index matching. The old method required using _different_ orders when `ravel`ling these grids and when calling `ravel_multi_index` to get 1D indices of a tiled sky. (Closes #1297 )
* Also return the az/el bin edges when creating these grids. (Closes #1298 )
* Allow for non-reversed elevations in these grids. (Closes #1299 )

## Updated Files

  - Ravel in "ij" order
  - Create az/el bin edges
  - Allow for returning reversed elevations `(90, 89,...-89, -90)` or non-reversed elevations `(-90, -89,...89, 90)`. 
    - Ultra may want reversed elevations, but it is probably more natural to work in non-reversed elevations.
    - Allowing this utility function to return either, allows us to create consistent bins and grids, rather than reversing each if/when we need to. 

## Testing

- Now also tests the creation of bin edges, and that the solid angles at equal elevation are equal.


tmplummer (commented):
This all looks reasonable to me. I have one concern.
The number of inputs/outputs from this function makes me think that it would make sense to put it in a Class. Maybe define a Grid or AngularGrid class that upon instantiation computes all the things that this single function does, but just stores the data in the class object. That will really simplify the code that uses the grid because you won't have to carry around 6 outputs from this function. You can just access the attributes of the class object to grab what you need. Maybe this is not needed because this grid generation function will only be called by either a Map class or a PSet class?
This isn't a blocker for this PR just wanted to put it out there to see what you think.

nkerman:
> This all looks reasonable to me. I have one concern. The number of inputs/outputs from this function makes me think that it would make sense to put it in a Class. Maybe define a Grid or AngularGrid class that upon instantiation computes all the things that this single function does, but just stores the data in the class object. That will really simplify the code that uses the grid because you won't have to carry around 6 outputs from this function. You can just access the attributes of the class object to grab what you need. Maybe this is not needed because this grid generation function will only be called by either a Map class or a PSet class? This isn't a blocker for this PR just wanted to put it out there to see what you think.

Thanks, Tim! 

I personally lean towards flexible functions and don't really think there's a problem with having a lot of inputs/outputs, as long as they're all doing different, well defined things. I'm not sure I see the benefit of having a class vs. function with the following inputs, but maybe the issue is more that this is just trying to be too flexible, and it ought not to have the options for radians/degrees, etc.

```
     spacing: float,
     input_degrees: bool = False,
     output_degrees: bool = False,
     centered_azimuth: bool = False,
     centered_elevation: bool = True,
     reversed_elevation: bool = False,
```

When in the future, I'm writing code that uses this, like the RectangularMap object, it's easier for me to think:

"Okay, I need to import this function `make_grid()` that will give me the pixel centers in az/el."

than to think:

"I need to import this class and make a new grid object, that will then own its grid information like the az/el."

---

That's just a depiction of my thought process, and I know I'm missing a lot of best practices experience. All said and done, if it makes more sense to you for this to live in a class, I will go ahead and change it! I think it's important to get this sort of support code in good shape, but I've just been thinking of it as a tool to get the ENA maps built, and mainly want to push those forward.

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,286 @@
...
+        # Ensure valid grid spacing (positive, divides evenly into pi radians)
+        if self.spacing_radians <= 0:
+            raise ValueError("Spacing must be positive valued, non-zero.")
+
+        if not np.isclose((np.pi / self.spacing_radians) % 1, 0):
+            raise ValueError("Spacing must divide evenly into pi radians.")
+
+        # Create the bin edges for azimuth and elevation.
+        # E.g. for spacing=1, az_bin_edges = [0, 1, 2, ..., 359, 360] deg.
+        self.az_bin_edges = np.linspace(
```
tmplummer: Is there a reason that bin edges and midpoints are calculated in their own way in this class rather than reusing `build_spatial_bins`?

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,286 @@
...
+
+        if angular_units == "deg":
+            self.az_range = np.rad2deg(self.az_range)
+            self.el_range = np.rad2deg(self.el_range)
+            self.az_grid = np.rad2deg(self.az_grid)
+            self.el_grid = np.rad2deg(self.el_grid)
+            self.az_bin_edges = np.rad2deg(self.az_bin_edges)
+            self.el_bin_edges = np.rad2deg(self.el_bin_edges)
+
+    def to_degrees(self) -> None:
```
tmplummer: I would lean toward always keeping the class az, el, spacing units in radians and writing accessor functions decorated with @property to specifically get degrees. For example:
```
    @property
    def az_range_degrees():
        return np.degrees(self.az_range)
```
There is maybe a way to do this generically across any of the angular attributes, but I am not sure.

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,286 @@
...
+            0, 2 * np.pi, int(2 * np.pi / self.spacing_radians) + 1
+        )
+        self.el_bin_edges = np.linspace(  # [-90, -89, ..., 89, 90] deg.
+            -np.pi / 2, np.pi / 2, int(np.pi / self.spacing_radians) + 1
+        )
+
+        # Create the 2D grid of azimuth and elevation angles at center of each bin.
+        # These ranges are offset by half the spacing and are
+        # one element shorter than the bin edges.
+        self.az_range = np.arange(  # [0.5, 1.5, ..., 359.5] deg.
```
tmplummer: I find the `az_range`, and `el_range` names confusing. These are bin centers as opposed to edges, right?

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,286 @@
...
+        if centered_azimuth:
+            self.az_range = self.az_range - np.pi
+        if centered_elevation:
+            self.el_range = self.el_range - np.pi / 2
+
+        # If desired, reverse the elevation range so that the grid is in the order
+        # defined by the Ultra prototype code (`build_dps_grid.m`).
+        if self.reversed_elevation:
+            self.el_range = self.el_range[::-1]
+            self.el_bin_edges = self.el_bin_edges[::-1]
```
tmplummer: Is this useful/needed. It would be nice if this could be removed or moved into a test rather than needing it in production code.

tmplummer (commented):
I gave some feedback on this PR with the assumption that the overall design is the direction that we want to go. Without sitting down and doing the architecture design for all of the moving pieces, I don't know how/if this fits. Hopefully Greg and I can hash out a more concrete design soon. Since this is a major roadblock for you, maybe we can also figure out a way to get some of these PRs approved while reducing impacts of future changes.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Yes, that's correct!

I wanted to distinguish them as the 1D, monotonically increasing range, on which the 2D meshgrid is based, (and then unwrapped into 1D non-monotonically-increasing/"zig-zagging" arrays of az, el). 

How about `"az_bin_centers_1d"`?

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: The Ultra prototype code uses it, so I wanted to at least have the option to do so. 

If we don't do it here, before making the meshgrid, it gets baked into the meshgrid, and hard to reverse later in Ultra-specific code.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: > There is maybe a way to do this generically across any of the angular attributes, but I am not sure.

Yeah, I didn't love the state approach, and I was thinking about how to do this generically, but I couldn't think of an obvious way. I'll try to poke around and find something!

imap_processing/ena_maps/utils/spatial_utils.py
tmplummer: Since this object doesn't have the unwrapped version of the 2D grid, I'd also be happy with `az_bin_centers`. Either way, as long as it is documented.

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,286 @@
...
+    az_bin_edges : np.ndarray
+        Array of azimuth bin boundary values.
+    el_bin_edges : np.ndarray
+        Array of elevation bin boundary values.
+    az_bin_midpoints : np.ndarray
+        Array of azimuth bin midpoint values.
+    el_bin_midpoints : np.ndarray
+        Array of elevation bin midpoint values.
+    """
+    # Azimuth bins from 0 to 360 degrees.
```
tmplummer, line 39: Ram and Anti-ram maps will need to have specific azimuth ranges, I think. So, for instance only have azimuth between 0 and 180 degrees. No need to modify this now, just keep it in mind that we may not always have full sky maps.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: I switched to using `build_spatial_bins`!

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Done! You can iterate through a list of attrs, and create properties from them:

```
    def __init_properties__(self) -> None:
        """Automatically generate degree properties for all radian attributes."""
        for name in [
            "spacing",
            "az_bin_midpoints",
            "el_bin_midpoints",
            "az_bin_edges",
            "el_bin_edges",
            "az_grid",
            "el_grid",
        ]:
            setattr(
                self.__class__,
                f"{name}_degrees",
                property(lambda self, n=name: np.rad2deg(getattr(self, n))),  # type: ignore[misc]
            )
```

Mypy doesn't love it (it doesn't seem to like lambdas, nor `numpy`'s loose return types from `rad2deg`). I tried spinning it out into functions, and it just got much too unwieldy. I personally think we should just ignore this particular type checking error.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Great! I opted for `az_bin_midpoints`, for consistency with how Laura had written about these 1D values in some L1 code.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Got it! We might just want to mask off part of the grid, but it'll require more thought.

nkerman:
@greglucas - this is the PR we were talking about. It's a couple bugfixes/bits of neatening up that spun out of previous PR discussions. Those PRs have since been marked as Drafts.

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    )
+
+    if output_degrees:
+        solid_angle_grid *= (180 / np.pi) ** 2
+
+    return solid_angle_grid
+
+
+@typing.no_type_check
+def rewrap_even_spaced_el_az_grid(
```
greglucas: In your functions above this you have `az_el`. Are we able to keep the order consistent everywhere?
```suggestion
def rewrap_even_spaced_az_el_grid(
```

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    Parameters
+    ----------
+    spacing : float
+        The bin spacing in the specified units.
+    input_degrees : bool, optional
+        If True, the input spacing is in degrees
+        (default is True for radians).
+    output_degrees : bool, optional
+        If True, the output solid angle map is in square degrees
+        (default is False for steradians).
```
greglucas: Do we need the options here? Sometimes it is helpful to limit our set of input parameters and document clearly what you are doing, then force the caller to do the transforms if they want something different.

Your docstring here says: build "solid angle map for a given spacing in degrees", but then you allow `input_degrees` to change that which is a bit confusing. You can change the first space parameter to be spacing_deg if you want to be explicit too.

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+    # Azimuth bins from 0 to 360 degrees.
+    az_bin_edges = np.arange(0, 360 + az_spacing, az_spacing)
+    az_bin_midpoints = az_bin_edges[:-1] + az_spacing / 2  # Midpoints between edges
+
+    # Elevation bins from -90 to 90 degrees.
+    el_bin_edges = np.arange(-90, 90 + el_spacing, el_spacing)
+    el_bin_midpoints = el_bin_edges[:-1] + el_spacing / 2  # Midpoints between edges
+
+    return az_bin_edges, el_bin_edges, az_bin_midpoints, el_bin_midpoints
```
greglucas: Do we want/need the midpoints? The edges should be good enough to constrain the problem and then let other functions get the midpoints if they need them later. Returning a lot of things from a function can be confusing.

Also, keep in mind the az/el ordering throughout, later you have `el` first.

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    raveled_values : NDArray
+        1D array of values to be reshaped into a 2D grid.
+    shape : tuple[int], optional
+        The shape of the original grid, if known, by default None.
+        If None, the shape will be inferred from the size of the input array.
+    extra_axis : bool, optional
+        If True, input is a 2D array with latter axis being 'extra', non-spatial axis.
+        This axis (e.g. energy bins) will be preserved in the reshaped grid.
+    order : {'C', 'F'}, optional
+        The order in which to rewrap the values, by default 'F' for Fortran order.
```
greglucas: This is different than numpy's default of C reshaping, is that expected?

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    Parameters
+    ----------
+    raveled_values : NDArray
+        1D array of values to be reshaped into a 2D grid.
+    shape : tuple[int], optional
+        The shape of the original grid, if known, by default None.
+        If None, the shape will be inferred from the size of the input array.
+    extra_axis : bool, optional
+        If True, input is a 2D array with latter axis being 'extra', non-spatial axis.
+        This axis (e.g. energy bins) will be preserved in the reshaped grid.
```
greglucas: Remove this and just say that the first dimension is always the raveled dimension and anything after that gets preserved? i.e. do this automatically without needing a keyword argument.

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    ):
+        raise ValueError("Input must be a 1D array or 2D array with extra axis.")
+
+    # We can infer the shape if its evenly spaced and 2D
+    if not shape:
+        spacing_deg = 1 / np.sqrt(raveled_values.shape[0] / (360 * 180))
+        shape = (int(180 // spacing_deg), int(360 // spacing_deg))
+
+    if extra_axis:
+        shape = (shape[0], shape[1], raveled_values.shape[1])
```
greglucas:
```suggestion
    if raveled_values.ndim == 2:
        shape = (shape[0], shape[1], raveled_values.shape[1])
```

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    spacing_deg : float, optional
+        Spacing of the grid in degrees, by default 0.5.
+    centered_azimuth : bool, optional
+        Whether the azimuth grid should be centered around 0 degrees/0 radians,
+        i.e. from -pi to pi radians, by default False.
+        If True, the azimuth grid will be from -pi to pi radians.
+    centered_elevation : bool, optional
+        Whether the elevation grid should be centered around 0 degrees/0 radians,
+        i.e. from -pi/2 to pi/2 radians, by default True.
+        If False, the elevation grid will be from 0 to pi radians.
```
greglucas: Should these be `center_azimuth : float` where a user would input what they want the center to be? The bool is slightly confusing to me.

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+        if self.spacing <= 0:
+            raise ValueError("Spacing must be positive valued, non-zero.")
+
+        if not np.isclose((np.pi / self.spacing) % 1, 0):
+            raise ValueError("Spacing must divide evenly into pi radians.")
+
+        # build_spacial_bins creates the bin edges and centers for azimuth and elevation
+        # E.g. for spacing=1, az_bin_edges = [0, 1, 2, ..., 359, 360] deg.
+        (az_bin_edges, el_bin_edges, az_bin_midpoints, el_bin_midpoints) = (
+            build_spatial_bins(az_spacing=spacing_deg, el_spacing=spacing_deg)
```
greglucas: Below this you call `np.deg2rad` on all these return values. Maybe the helper function should just return radians instead?

imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -0,0 +1,270 @@
...
+
+        # Keep track of number of points on the grid
+        self.grid_shape = self.az_grid.shape
+        self.grid_size = self.az_grid.size
+
+        # Create degree property attributes for all radian attributes
+        self.__init_properties__()
+
+    def __init_properties__(self) -> None:
+        """Automatically generate degree properties for all radian attributes."""
```
greglucas: This is a neat way to do this :)

I am worried about adding this just for the sake of adding it. I really find that swapping between degrees/radians leads to confusion and more options can be worse for the end user/easier to make a small mistake later on. So if we can work with only radians in all inputs/outputs of your helper functions etc, and only have the entrypoint to create a map in degrees or something like that it might make our lives easier. `create_my_map(degrees_inputs) -> create_bins(radians) -> more_radians_calls()`

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Yes!

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Done! I know I tend to design functions to be too flexible, with too many params, etc.

I chose to use `spacing_deg`, as you suggest, and always output steradians (while not the most self consistent, I think those are the right units, as the maps are built in degree increments, but square degrees are unpleasant).

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: The midpoints are the coordinates of the rectangular 'pixel' on the sky. We definitely need them for constructing the grid later in the AzEl class, and for building the map CDF. Also, some of Laura's L1 code uses them. 

I think it's probably more confusing to need to do something like:

`el_bin_midpoints = el_bin_edges[:-1] + el_spacing / 2  # Midpoints between edges`

in a lot of different locations.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: This all grew out of trying to replicate the Ultra prototype code, which uses 'F' style reshaping in MATLAB, and order really matters when unwrapping the sky like this. 

I can change the default to C.

imap_processing/ena_maps/utils/spatial_utils.py
greglucas: Alternatively, do we need the "edges" then, or could they be inferred from the spacing later when they are needed? What I'm getting at is that they are somewhat redundant with one another and it is often good to have one "source of truth" that you can then calculate the other from instead of possibly getting them out of sync at some later point in time.

imap_processing/ena_maps/utils/spatial_utils.py
greglucas: If you're going to always use "F" or that makes the most sense, then I'd say default to that, or even just make that a fixed parameter and not allow it to be input.

If these are all your functions, then do what is best for you. I don't think people are going to be flipping between C/F inputs and that might cause more confusion later if we are flexible.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Ah, thanks for the code suggestion below - that helped me understand how to do this simply.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Hmmm, I find that somewhat more confusing than just asking if a user wants their azimuth centered around 0 (-180˚ -> 180˚) or simply incrementing (0 -> 360˚). 

If this is confusing, I'd probably lean towards just making it always have centered elevation and non-centered azimuth. Thoughts?

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Yeah, that makes sense. I was trying to keep it consistent with how `build_spatial_bins` was made for Laura's L1 code, which uses degrees.

I can switch this to returning radians, and just adjust the L1a code to convert

imap_processing/ena_maps/utils/spatial_utils.py
greglucas: Sounds good, I would say fix it to one choice then and do the coordinates you think you'll be working in. Reduce the number of options for now, and then if you do need to change it later we can add that keyword argument then keeping the default at what you choose.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Good point to keep it consistent, and avoid the temptation to switch units. 

It just seemed much easier to get the units you want here, rather than having to call `rad2deg` on each individual value. Since you're right that it's not strictly necessary, I'll cut this for now.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: I think it's important to be able to switch the order as a parameter, but I'm glad you pointed this out, and I now think that it should default to the same order as numpy.

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: NOTE: We discussed this in person

greglucas (approved):

imap_processing/ena_maps/utils/spatial_utils.py
nkerman: Done!

PR #1301, opened 2025-01-24
Glows obs day updates

maxinelasp:
# Change Summary
I needed to add a change to split the validation data by observation data for GLOWS validation purposes in L2. While I was at it, I cleaned up a few other things and made the tests more thorough and resistant to future observation day changes.

## Overview

added a function to estimate observation days
added a function to decode flags (needed for previous function)
reworked some tests


vmartinez-cu (approved):

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -611,3 +611,26 @@ def output_data(self) -> tuple:
             A tuple containing each attribute value in the class.
         """
         return tuple(getattr(self, out.name) for out in dataclasses.fields(self))
+
+    @staticmethod
+    def deserialize_flags(raw: int) -> np.ndarray[int]:
```
lacoak21, line 616: return type -> np.ndarray[bool]?

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -88,6 +98,36 @@ def glows_l1a(packet_filepath: Path, data_version: str) -> list[xr.Dataset]:
     return output_datasets
 
 
+def determine_observational_day(hist_l0: list[HistogramL0]) -> list:
+    """
+    Find the timestamps for each observational day.
+
+    This function temporarily uses the is_night flag to determine the start of a new
+    observational day, but should eventually use the spin table APIs.
```
lacoak21, line 106: So this will eventually query to get the timestamp of the start of the spin (aka observational day)?

imap_processing/tests/glows/test_glows_l1a_cdf.py
```diff
@@ -16,13 +15,10 @@
 from imap_processing.glows.utils.constants import TimeTuple
 
 
-@pytest.fixture(scope="module")
-def l1a_data():
+@pytest.fixture()
```
lacoak21, line 18: Why did you remove 'scope="module"'? Seems like the object could be reused for both tests

lacoak21 (approved):

imap_processing/tests/glows/test_glows_l1a_cdf.py
maxinelasp: I moved all the module level tests into conftest, and this test is only used within this file. It also caused some error.

imap_processing/glows/l1a/glows_l1a.py
maxinelasp: Yes, eventually this method will be replace d with an API call

imap_processing/glows/l1b/glows_l1b_data.py
maxinelasp: I'm trying to figure out a mypy bug with this type return, so it's an int for now.

imap_processing/tests/glows/test_glows_l1a_cdf.py
lacoak21: Ah got it

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -88,6 +98,36 @@ def glows_l1a(packet_filepath: Path, data_version: str) -> list[xr.Dataset]:
...
+    Returns
+    -------
+    list
+        List of start times for each observational day.
+    """
+    prev_is_night = -1
+    obs_day_change = []
+    for hist in hist_l0:
+        flags = HistogramL1B.deserialize_flags(hist.FLAGS)
+        is_night = flags[6]
```
daralynnrhode: Mypy seems to be flagging `prev_is_night` as an `integer,` based on the -1 assignment to it. But `is_night` is being found to be an `ndarray` based on where you are getting the value from which is `flags` which is an `ndarry.` I think by adding a simple type hint to the line should fix the issue assuming both variable should be interpreted as integers.

```suggestion
        is_night: int  = int(flags[6])
```

PR #1302, opened 2025-01-24
HIT L1B standard rates

### LINKED ISSUE #1062: HIT L1B - write standard rates algorithm code
Write algorithm for standard rates

Apply equations

- livetime = livetime_counter / 270 
- count_rates = raw_counts / livetime

vmartinez-cu:
This PR updates HIT L1B processing to create a standard rates data product, which is one of three science products for this data level. It also adds a unit test for the new standard rates processing function and updates existing unit tests where needed. 

A minor update was also made to other files to change a data variable name as requested by the HIT team.

Closes #1062 

## Updated Files
   - added global attributes for L1B science products
- constants.py
   - updated data variable name
- test_decom_hit.py
   - updated data variable name
- test_hit_l1a.py
   - updated data variable name 
- hit_l1b.py
   - Added new function to create L1B science data products
   - Added new function to process standard rates data  
- test_hit_l1b.py
   - added unit test for new function that processes standard rates
   - added some pytest fixtures to support tests
   - updated test for hit_l1b function to include new standard rates dataset it processes
   - removed a unit test that checks attributes for hk dataset
   - cleaned up some existing tests  

## Testing
Unit test added and some existing tests were updated


imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
```diff
@@ -39,3 +39,25 @@ imap_hit_l1b_hk:
     Data_type: L1B_HK>Level-1B Housekeeping
     Logical_source: imap_hit_l1b_hk
     Logical_source_description: IMAP Mission HIT Instrument Level-1B Housekeeping Data.
+
+imap_hit_l1b_standard-rates:
+    <<: *instrument_base
+    Data_level: 1B
```
maxinelasp, line 45: I believe this should be "L1B"

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -26,34 +27,176 @@ def hit_l1b(dependencies: dict, data_version: str) -> list[xr.Dataset]:
...
+                # attributes
+                dims = {
+                    key: value
+                    for key, value in attr_mgr.get_variable_attributes(field).items()
+                    if "DEPEND" in key
+                }
+                dataset[field].attrs = attr_mgr.get_variable_attributes(field)
+                dataset[field].assign_coords(dims)
+            except KeyError:
+                print(f"Field {field} not found in attribute manager.")
```
maxinelasp, line 123: Do we need this in there?

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -26,34 +27,176 @@ def hit_l1b(dependencies: dict, data_version: str) -> list[xr.Dataset]:
...
+                logger.warning(f"Field {field} not found in attribute manager.")
+
+        dataset.epoch.attrs = attr_mgr.get_variable_attributes("epoch")
+        # Remove DEPEND_0 attribute from epoch variable added by attr_mgr.
+        # Not required for epoch
+        del dataset["epoch"].attrs["DEPEND_0"]
```
maxinelasp: I'd rather just pass in "check_schema=False" rather than deleting the depend_0 afterwards.

maxinelasp (approved):
Nice, looks like a clean implementation. I had a few comments on CDF attrs but other than that looks good!

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -26,34 +27,176 @@ def hit_l1b(dependencies: dict, data_version: str) -> list[xr.Dataset]:
...
+        "penfgrates_index",
+        "penbgrates_index",
+        "ialirtrates_index",
+    ]
+    l1b_standard_rates_dataset = l1b_standard_rates_dataset.assign_coords(
+        {coord: raw_counts_dataset.coords[coord] for coord in coords}
+    )
+
+    # Define fields from the raw_counts_dataset to calculate standard rates from
+    standard_rate_fields = [
```
sdhoyt, line 181: I think sectored rates should be part of this too?

sdhoyt (commented):
Nice work! Just one comment

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: Good question. This is a place holder and will eventually go away. I don't have attributes defined for L1B yet and I don't want this function to fail when attributes for the fields aren't found in the yaml.

imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
vmartinez-cu: Good catch!

vmartinez-cu: I looked at global attributes for other instruments and am also seeing "1A" and "1B" without the leading "L" so I wonder if it's not needed for this field?

maxinelasp: Hm, I have always used "L1B" but I'm not actually sure what the requirements are. We should match everyone though.

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: I don't see it listed in the attribute definitions HIT provided which is why I didn't include it. I'll double check with them in case it was omitted by mistake.

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: That's a much better approach. Thanks for the suggestion! I wasn't sure if check schema did other important checks but the attributes are the same using this flag vs. deleting DEPEND_0.

maxinelasp: No, it just checks for the existence of commonly accepted variable attributes as retrieved from the schema. In the future we hope to make it a bit more useful, but especially for the epoch variable any real errors are usually caught by cdflib anyway.

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: HIT confirmed that sectored rates don't need to be included in this product. It will be handled in another L1B product specific to sectored rates

PR #1303, opened 2025-01-27
SWE: Flux units and conversion factor correction

### LINKED ISSUE #1287: Correct Flux units and update conversion factors

tech3371:
# Change Summary
closes #1287 

## Overview
Small change to correct flux units and conversion factor.

## Updated Files
   - update doc strings and conversion factor

### DIFF
--- a/imap_processing/swe/l2/swe_l2.py
+++ b/imap_processing/swe/l2/swe_l2.py
@@ -58,33 +58,33 @@ def calculate_phase_space_density(l1b_dataset: xr.Dataset) -> xr.Dataset:
     """
     Convert counts to phase space density.
 
-    Calculate phase space density, fv, in units of s^3/ (cm^6 * ster).
-        fv = 2 * (C/tau) / (G * v^4)
-        where:
-            C / tau = corrected count rate. L1B science data.
-            G = geometric factor, in (cm^2 * ster). 7 CEMS value.
-            v = electron speed, computed from energy, in cm/s.
-                We need to use this formula to convert energy to speed:
-                    E = 0.5 * m * v^2
-                    where E is electron energy, in eV
-                    (result from get_particle_energy() function),
-                    m is mass of electron (9.10938356e-31 kg),
-                    and v is what we want to calculate. Reorganizing above
-                    formula result in v = sqrt(2 * E / m). This will be used
-                    to calculate electron speed, v.
+    Calculate phase space density is represented by this symbol, fv.
+    Its unit is s^3/ (cm^6 * ster).
 
-                Now to convert electron speed units to cm/s:
-                    v = sqrt(2 * E / m)
-                      = sqrt(2 * E(eV) * 1.60219e-19(J/eV) / 9.10938e-31 kg)
-                        where J = kg * m^2 / s^2.
-                      = sqrt(2 * 1.60219 * 10e−19 m^2/s^2 * E(eV) / 9.10938e-31)
-                      = sqrt(2 * 1.60219 * 10e−19 * 10e4 cm^2/s^2 * E(eV) / 9.10938e-31)
-                      = sqrt(3.20438 * 10e-15 * E(eV) / 9.10938e-31) cm/s
-                      = sqrt( (3.20438 * 10e-15 / 9.10938e-31) * E(eV) ) cm/s
-        fv = 2 * (C/tau) / (G * v^4)
-           = 2 * (C/tau) / (G * (sqrt( (3.20438 * 10e-15 / 9.10938e-31) * E(eV) ))^4)
-           = 2 * (C/tau) / (G * (sqrt(3.5176e16)^4 * E(eV)^2)
-           = 2 * (C/tau) / (G * 1.237e31 * E(eV)^2)
+    The formula to calculate phase space density,
+
+    Where:
+        C / tau = corrected count rate which in the input L1B science data.
+        G = geometric factor, in (cm^2 * ster). 7 CEMs geometric factor value.
+        E = Energy in Joules. eV * 1.60219e-19(J/eV). eV is result
+            from get_particle_energy() function.
+        m = mass of electron (9.10938356e-31 kg).
+        s = second.
+        v = sqrt(2 * E / m). Electron speed, computed from energy. In cm/s.
+        J = kg * m^2 / s^2. J for joules.
+        fv = phase space density.
+
+    v   = sqrt(2 * E / m)
+        = sqrt(2 * eV * 1.60219e-19(J/eV) / 9.10938e-31 kg)
+        = sqrt(2 * 1.60219 * 10e−19 m^2/s^2 * eV / 9.10938e-31)
+        = sqrt(2 * 1.60219 * 10e−19 * 10e4 cm^2/s^2 * eV / 9.10938e-31)
+        = sqrt(3.20438 * 10e-15 * eV / 9.10938e-31) cm/s
+        = sqrt((3.20438 * 10e-15 / 9.10938e-31) * eV) cm/s
+
+    fv  = 2 * (C/tau) / (G * v^4)
+        = 2 * (C/tau) / (G * (sqrt( (3.20438 * 10e-15 / 9.10938e-31) * eV ))^4)
+        = 2 * (C/tau) / (G * (sqrt(3.5176e16)^4 * eV^2)
+        = 2 * (C/tau) / (G * 1.237e31 * eV^2)
         Ruth Skoug also got the same result, 1.237e31.
 
     Parameters
@@ -113,7 +113,7 @@ def calculate_phase_space_density(l1b_dataset: xr.Dataset) -> xr.Dataset:
     particle_energy_data = particle_energy_data.reshape(-1, 24, 30)
 
     # Calculate phase space density using formula:
-    #   2 * (C/tau) / (G * 1.237e31 * E(eV)^2)
+    #   2 * (C/tau) / (G * 1.237e31 * eV^2)
     # See doc string for more details.
     density = (2 * l1b_dataset["science_data"]) / (
         GEOMETRIC_FACTORS[np.newaxis, np.newaxis, np.newaxis, :]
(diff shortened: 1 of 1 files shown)


imap_processing/swe/l2/swe_l2.py
```diff
@@ -136,22 +138,44 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
     """
     Calculate flux.
 
-    To get flux, j = 2 * m * E * f
+    As SWE stated, Flux, j, is calculated as
+        j = C / (G * E * tau)
+        where E is the particle energy, and is needed because
```
nkerman: You don't directly define what G, C, tau are here, but could reference the previous function which does. (e.g., see the function definition of `calculate_phase_space_density`)

imap_processing/swe/l2/swe_l2.py
```diff
@@ -136,22 +138,44 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    Or combining these two equations gives:
+        fv = 2 * j * E / v^4
+        or
+        j = (fv * v^4) / (2 * E)
+
+    To get flux factor, j = (fv * v^4) / (2 * E)
         where:
-        f = calculate_phase_space_density() result
-        m - mass of electron
+        fv = calculate_phase_space_density() result. Units are
```
nkerman:
```suggestion
        fv = the phase space density of ??? solar wind electrons ???
        given by the result of calculate_phase_space_density(). 
        Units are...
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -136,22 +138,44 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+            s^3 / (cm^6 * ster).
+        v = sqrt( (3.20438 * 10e-15 / 9.10938e-31) * E(eV) ) cm/s.
```
nkerman: In my opinion, when you write an equation, and then say "where variable 1 = ..., and variable 2 = ...", It should actually say what the variables are, rather than an equation for them, especially an equation of constants. 

To give an example, (but without actually knowing what this should be), I think it's better to say something like:

```Python
"""
where:
v = the velocity of a measured solar wind electron, calculated from its measured energy.
v_electron = sqrt(2 E_electron/m_electron)
"""
```
---

Separately, I'd suggest double checking the 10E-15, just since it's more common to work with "1E<exponent>", which means "10^<exponent>". 10E-15 actually means "10^16".

imap_processing/swe/l2/swe_l2.py
```diff
@@ -136,22 +138,44 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
-            (s^3 / (cm^6 * ster)
-          = 2 * 9.10938356e-31 * E(eV) * 1.60219e-19 * 10^4 ((kg^2 * cm^2 * s^3) /
-            (s^2 * cm^6 * ster))
-        TODO: ask Ruth Skoug what units to use for flux and work out remaining units.
+    To calculate flux units:
+        j = (fv * v^4) / (2 * E(eV))
+          = ((s^3 / (cm^6 * ster)) * (cm^4/s^4)) / (2 * E(eV))
+          = ((s^3 * cm^4) / (cm^6 * s^4 * ster)) / (2 * E(eV))
+          = (1 / (cm^2 * s * ster)) / (2 * E(eV))
+          = 1 / (2 * E(eV) * cm^2 * s * ster)
```
nkerman: I don't think 2 should be anywhere in the units calculation, and it'd be more consistent to just say the unit eV, rather than E(eV). 

For flux in energy bands, `1 / (cm^2 * s * steradian * eV)` makes sense to me though! That's saying 1 (or N) particles per bit of area, per second, per bit of solid angle, per energy.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -136,22 +138,44 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+
+    To get flux factor, j = (fv * v^4) / (2 * E)
         where:
-        f = calculate_phase_space_density() result
-        m - mass of electron
+        fv = calculate_phase_space_density() result. Units are
+            s^3 / (cm^6 * ster).
+        v = sqrt( (3.20438 * 10e-15 / 9.10938e-31) * E(eV) ) cm/s.
+            See calculate_phase_space_density() for calculation.
         E - energy in joules. E(eV) * 1.60219e-19(J/eV)
```
nkerman: In the equations, energy is written as E(eV), which I take to mean "Energy, in units of electron-Volts", but here it's described as input in Joules. The unit conversion written here is right, but I'm a bit confused about whether J or eV is being used in the equations and if it's consistent.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -136,22 +138,44 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+          = (1 / (cm^2 * s * ster)) / (2 * E(eV))
+          = 1 / (2 * E(eV) * cm^2 * s * ster)
+    To calculate flux conversion factor:
+        j = (fv * v^4) / (2 * E)
+          = ( fv * (sqrt( (3.20438 * 10e-15 / 9.10938e-31) * E(eV) )^4) ) / (2 * E(ev))
+          = ( fv * ((3.20438 * 10e-15 / 9.10938e-31) * E(eV))^1/2) ^ 4 ) / (2 * E(eV))
+          = ( fv * (3.20438 * 10e-15 / 9.10938e-31)^2 * E(eV)^2) ) / (2 * E(eV))
+          = ( fv * 1.237e31 * E(eV)^2) ) / (2 * E(eV))
+          = ( fv * 1.237e31 * E(eV) ) / 2
+          = (fv * 6.187e30 * E(eV)
```
nkerman: Nit: Missing a close paren on the last line

nkerman (commented):
The unit conversion factor change looks fine!

Most of this documentation seems to be deriving the actual equations in a bit more detail than I would usually expect to see in the code. 

Is this the best place for these multi-line equation derivations to live? I'd assume the derivations are all written out in the Algorithm doc, and that you could have a simplified description here, focused on what each variable _means_ and the final equation, with units.

imap_processing/swe/l2/swe_l2.py
tech3371: My bad. I directly took Ruth's response here. Will point to that function for more details.

imap_processing/swe/l2/swe_l2.py
tech3371: I am not familiar with this. When I wrote that, I was thinking from what inputs of those equations are coming from. Cool, I will rewrite it in that flow. Makes sense.

imap_processing/swe/l2/swe_l2.py
tech3371: Again, problem of copy paste.

tech3371: No. Actually, this is done because I need to read energy steps from lookup table and then convert to eV. Then Ruth mentioned she wants that in Joules.

imap_processing/swe/l2/swe_l2.py
tech3371: About the power notation, Ruth seems to use this notation
```
g = {255.6e-6, 511.0e-6, 633.4e-6, 659.5e-6, 733.5e-6, 540.3e-6, 272.9e-6};
```

Also, I sent her these workout. She confirmed that she got same. 🤞🏻

tech3371: btw, doc string much cleaner after rewriting in that format!

imap_processing/swe/l2/swe_l2.py
```diff
@@ -136,22 +138,37 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
-        j = 2 * m * E * f
-          = 2 * 9.10938356e-31 kg * E(eV) * 1.60219e-19(J/eV) * (s^3/ (cm^6 * ster)
-          = 2 * 9.10938356e-31 kg * E(eV) * 1.60219e-19 (kg * m^2 / s^2) *
-            (s^3 / (cm^6 * ster)
-          = 2 * 9.10938356e-31 * E(eV) * 1.60219e-19  kg^2 * ( 10^4 cm^2 / s^2) *
-            (s^3 / (cm^6 * ster)
-          = 2 * 9.10938356e-31 * E(eV) * 1.60219e-19 * 10^4 ((kg^2 * cm^2 * s^3) /
-            (s^2 * cm^6 * ster))
-        TODO: ask Ruth Skoug what units to use for flux and work out remaining units.
+    Flux is represented by this symbol, j. Its unit is
```
nkerman, line 141: Constants like 2 should not be present in units. If that's what's in the Algorithm Document, go ahead and keep it, but I find that strange/confusing.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -56,33 +58,33 @@ def calculate_phase_space_density(l1b_dataset: xr.Dataset) -> xr.Dataset:
...
-           = 2 * (C/tau) / (G * 1.237e31 * E(eV)^2)
+    Calculate phase space density is represented by this symbol, fv.
+    Its unit is s^3/ (cm^6 * ster).
+
+    The formula to calculate phase space density,
+
+    Where:
+        C / tau = corrected count rate which in the input L1B science data.
+        G = geometric factor, in (cm^2 * ster). 7 CEMs geometric factor value.
+        E = Energy in Joules. eV * 1.60219e-19(J/eV). eV is result
```
nkerman:
```suggestion
        eV = Energy in electron-volts, calculated by get_particle_energy().
        E = Energy in Joules. eV * 1.60219e-19(J/eV). 
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -136,22 +138,37 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    1 / (2 * eV * cm^2 * s * ster).
+
+    The formula to calculate flux,
+
+    Where:
+        fv = the phase space density of solar wind electrons
+            given by calculate_phase_space_density() result.
+        J = kg * m^2 / s^2. J for joules.
+        E = Energy in Joules. eV * 1.60219e-19(J/eV). eV is result
+            from get_particle_energy() function.
```
nkerman: You don't need this line defining Joules. 

In general, you don't really need to define SI units like Joules, seconds, centimeters, etc., but here it doesn't come into the equation, so you definitely don't need to. 

```suggestion
        eV = Energy in electron-volts, calculated by get_particle_energy().
        E = Energy in Joules. eV * 1.60219e-19(J/eV). 
```

imap_processing/swe/l2/swe_l2.py
nkerman: > About the power notation, Ruth seems to use this notation
> 
> ```
> g = {255.6e-6, 511.0e-6, 633.4e-6, 659.5e-6, 733.5e-6, 540.3e-6, 272.9e-6};
> ```
> 
> Also, I sent her these workout. She confirmed that she got same. 🤞🏻

Got it, I'm assuming she wanted to keep things easily comparable in some standard unit (if it's 1e-6 = 10^-6, that would be "micro-" somethings. micrometers for distance, microjoules for energy, etc.) I'm just careful/wary about writing 10eN because it's so easy to confuse with 10^N, which it is not equal to.

nkerman (approved):
Just a couple suggestions on the docstring, which I think would make it clearer/neater. I think I can approve at this point!

It still seems to me like these derivations don't need to live in code docstrings, but I don't mind keeping them.

imap_processing/swe/l2/swe_l2.py
tech3371: This is how Ruth sent me and I wanted to keep it same. It's for me more than anything :)

tech3371:
> Just a couple suggestions on the docstring, which I think would make it clearer/neater. I think I can approve at this point!
> 
> It still seems to me like these derivations don't need to live in code docstrings, but I don't mind keeping them.

Thank you @nkerman . I will ask Ruth to include these units and factors into the algorithm. I know she has plan to update algorithm doc. I will ask this too.

PR #1307, opened 2025-01-28
HIT L1A - add uncertainty data

### LINKED ISSUE #1280: HIT L1A - add uncertainty fields to count rate dataset
HIT updated the algorithm doc to include uncertainty calculations. These need to be added to the L1 products. Add the calculation and fields to the L1A processing code and create new L1A CDF products.

vmartinez-cu:
This PR adds uncertainty data fields to the L1A counts data product. A function calculating plus and minus uncertainties for rate fields was added to L1A processing and a unit test was added to test this function. Additionally, a L1B unit test was updated to account for the new data fields.

Closes #1280 

## Updated Files

- hit_l1a.py
   - new function calculates uncertainty and adds data to counts dataset
- test_hit_l1a.py
   - unit test added to test new function
- test_hit_l1b.py
   - unit test updated to include new uncertainty data fields and also adds the dynamic threshold field that needs to carry though to L2 products


imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -162,6 +162,69 @@ def subcom_sectorates(sci_dataset: xr.Dataset) -> None:
         )
 
 
+# Calculate uncertainties for count rates
+def calculate_uncertainties(dataset: xr.Dataset) -> xr.Dataset:
+    """
+    Calculate uncertainties for each counts data variable in the dataset.
+
+    Calculate the upper and lower uncertainties. The uncertainty for
```
sdhoyt, line 170: This is a really helpful docstring!

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -162,6 +162,69 @@ def subcom_sectorates(sci_dataset: xr.Dataset) -> None:
...
+        "cno_energy_min",
+        "cno_energy_max",
+        "nemgsi_energy_min",
+        "nemgsi_energy_max",
+        "fe_energy_min",
+        "fe_energy_max",
+    ]
+
+    # Calculate uncertainties for each counts data variable
+    for var in dataset.data_vars:
```
sdhoyt: Should you put this function in your utils file if this needs to be done again in L1B and on?

sdhoyt (approved):
This was fast! Great work! Just one question, but nothing blocking a merge

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: I checked with HIT and the calculation is only needed in L1A. The resulting uncertainty data variables will be passed into L1B and L2 products.

PR #1308, opened 2025-01-28
Decom and Organize Lo L1A Spin Data

sdhoyt:
# Change Summary

## Overview
This PR decommutates the Lo Spin packet in the Lo pipeline and organizes the data in the dataset into 1D arrays with a spin coordinate for the number of spins in the pointing. Attributes will be added in the next PR

## Updated Files
- lo_l1a.py
   - add check for spin packet in dependency file
- lo_science.py
   - adds function to organize packet data into 1D arrays
- lo_xtce.xml
   - regenerated xtce to define each spin field (original telemetry def format) rather than reading them in as a binary chunk (my previous plan for parsing and creating arrays).

## Testing
Added unit test for new function

### DIFF
--- a/imap_processing/lo/l0/lo_science.py
+++ b/imap_processing/lo/l0/lo_science.py
@@ -481,5 +481,4 @@ def organize_spin_data(dataset: xr.Dataset) -> xr.Dataset:
         # Drop the individual spin data fields
         dataset = dataset.drop_vars(packet_fields)
 
-    print(dataset.data_vars)
     return dataset
--- a/imap_processing/lo/l0/lo_science.py
+++ b/imap_processing/lo/l0/lo_science.py
@@ -439,3 +439,42 @@ def find_valid_groups(
     ]
     valid_groups = [is_sequential(seq_ctrs) for seq_ctrs in grouped_seq_ctrs]
     return valid_groups
+
+
+def organize_spin_data(dataset: xr.Dataset) -> xr.Dataset:
+    """
+    Organize the spin data for Lo.
+
+    Parameters
+    ----------
+    dataset : xr.Dataset
+        Lo science direct events from packets_to_dataset function.
+
+    Returns
+    -------
+    dataset : xr.Dataset
+        Updated dataset with the spin data organized.
+    """
+    # Get the spin data fields
+    spin_fields = [
+        "start_sec_spin",
+        "start_subsec_spin",
+        "esa_neg_dac_spin",
+        "esa_pos_dac_spin",
+        "valid_period_spin",
+        "valid_phase_spin",
+        "period_source_spin",
+    ]
+
+    for spin_field in spin_fields:
+        packet_fields = [f"{spin_field}_{i}" for i in range(1, 29)]
+        # Combine the spin data fields along a new dimension
+        combined_spin_data = xr.concat(
+            [dataset[field] for field in packet_fields], dim="spin"
+        )
+        # Assign the combined data back to the dataset
+        dataset[spin_field] = combined_spin_data
+        dataset = dataset.drop_vars(packet_fields)
+
+    print(dataset.data_vars)
+    return dataset
--- a/imap_processing/lo/l0/lo_science.py
+++ b/imap_processing/lo/l0/lo_science.py
@@ -473,7 +473,7 @@ def organize_spin_data(dataset: xr.Dataset) -> xr.Dataset:
             [dataset[field] for field in packet_fields], dim="spin"
         )
         # Assign the combined data back to the dataset
-        dataset[spin_field] = combined_spin_data
+        dataset[spin_field] = combined_spin_data.transpose()
         dataset = dataset.drop_vars(packet_fields)
 
     print(dataset.data_vars)
--- a/imap_processing/lo/l0/lo_science.py
+++ b/imap_processing/lo/l0/lo_science.py
@@ -445,10 +445,14 @@ def organize_spin_data(dataset: xr.Dataset) -> xr.Dataset:
     """
     Organize the spin data for Lo.
 
+    The spin data is spread across 28 fields. This function
+    combines each of those fields into 2D arrays for each
+    epoch and spin.
+
     Parameters
     ----------
     dataset : xr.Dataset
-        Lo science direct events from packets_to_dataset function.
+        Lo spin data from packets_to_dataset function.
 
     Returns
     -------
--- a/imap_processing/lo/l0/lo_science.py
+++ b/imap_processing/lo/l0/lo_science.py
@@ -478,6 +478,7 @@ def organize_spin_data(dataset: xr.Dataset) -> xr.Dataset:
         )
         # Assign the combined data back to the dataset
         dataset[spin_field] = combined_spin_data.transpose()
+        # Drop the individual spin data fields
         dataset = dataset.drop_vars(packet_fields)
 
     print(dataset.data_vars)
--- a/imap_processing/tests/lo/test_lo_science.py
+++ b/imap_processing/tests/lo/test_lo_science.py
@@ -143,9 +143,9 @@ def fake_spin_data():
         fake_array_val = 0
         for packet_field in packet_fields:
             dataset[packet_field] = xr.DataArray(
-                np.array([fake_array_val, fake_array_val + 1]), dims="epoch"
+                np.array([fake_array_val, fake_array_val + 28]), dims="epoch"
             )
-            fake_array_val += 2
+            fake_array_val += 1
     return dataset
 
 
@@ -304,6 +304,7 @@ def test_validate_parse_events(sample_data, attr_mgr):
 
 def test_organize_spin_data(fake_spin_data):
     # Arrange
+    data_by_epoch_spin = np.arange(0, 56).reshape(2, 28)
     expected_dataset = xr.Dataset(
         data_vars=dict(
             num_completed=(["epoch"], np.array([0, 0])),
@@ -313,31 +314,31 @@ def test_organize_spin_data(fake_spin_data):
             acq_end_subsec=(["epoch"], np.array([0, 0])),
             start_sec_spin=(
                 ["epoch", "spin"],
-                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+                np.array(data_by_epoch_spin),
             ),
             start_subsec_spin=(
                 ["epoch", "spin"],
-                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+                np.array(data_by_epoch_spin),
             ),
             esa_neg_dac_spin=(
                 ["epoch", "spin"],
-                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+                np.array(data_by_epoch_spin),
             ),
             esa_pos_dac_spin=(
                 ["epoch", "spin"],
-                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+                np.array(data_by_epoch_spin),
             ),
             valid_period_spin=(
                 ["epoch", "spin"],
-                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+                np.array(data_by_epoch_spin),
             ),
             valid_phase_spin=(
                 ["epoch", "spin"],
-                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+                np.array(data_by_epoch_spin),
             ),
             period_source_spin=(
                 ["epoch", "spin"],
-                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+                np.array(data_by_epoch_spin),
             ),
         ),
         coords=dict(epoch=(["epoch"], np.array([0, 1]))),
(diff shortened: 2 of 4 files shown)


imap_processing/lo/l0/lo_science.py
```diff
@@ -439,3 +439,47 @@ def find_valid_groups(
...
+        "esa_pos_dac_spin",
+        "valid_period_spin",
+        "valid_phase_spin",
+        "period_source_spin",
+    ]
+
+    for spin_field in spin_fields:
+        packet_fields = [f"{spin_field}_{i}" for i in range(1, 29)]
+        # Combine the spin data fields along a new dimension
+        combined_spin_data = xr.concat(
```
nkerman, line 476: To make sure I'm understanding what's going on: 

At this point, the dataset contains variables like:
* "start_sec_spin_1", "start_sec_spin_2",..."start_sec_spin_28"

And you're concatting all 28 into a new variable called "start_sec_spin" - right? 

Is there any info lost by not keeping track of _which_ of the 28 fields an entry started in?

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -266,3 +297,30 @@ def test_validate_parse_events(sample_data, attr_mgr):
 
     assert dataset["de_count"].values == 1998
     assert dataset["passes"].values == 8
+
+
+def test_organize_spin_data(fake_spin_data):
+    # Arrange
```
nkerman, line 306: This got me to look up "Arrange, Act, Assert" - that's a cool framework/mnemonic.

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -115,6 +116,36 @@ def segmented_pkts_fake_data():
...
+        "start_sec_spin",
+        "start_subsec_spin",
+        "esa_neg_dac_spin",
+        "esa_pos_dac_spin",
+        "valid_period_spin",
+        "valid_phase_spin",
+        "period_source_spin",
+    ]
+
+    for spin_field in spin_fields:
```
nkerman, line 141: If you could put `i` into the actual values rather than just 0s, you could test the ordering is done correctly.

imap_processing/lo/l0/lo_science.py
sdhoyt: That's correct and I don't believe so. If that information is needed, you can figure it out by the array index spin_(array index + 1)

imap_processing/tests/lo/test_lo_science.py
sdhoyt: Yeah good point. I'll update this

imap_processing/lo/l0/lo_science.py
sdhoyt, line 476: I think this needs to be reshaped to a 1D array, so the 28 spins recorded in next packet follow the 28 spins in the current packet along the same dimension. Need to confirm with Lo

imap_processing/lo/l0/lo_science.py
```diff
@@ -439,3 +439,47 @@ def find_valid_groups(
...
+            [dataset[field] for field in packet_fields], dim="spin"
```
greglucas, line 477: Can you do this directly rather than creating an intermediate by passing in the fields to the data_vars keyword?

```suggestion
            dataset, data_vars=packet_fields, dim="spin"
```

imap_processing/lo/l0/lo_science.py
```diff
@@ -439,3 +439,47 @@ def find_valid_groups(
...
+        )
+        # Assign the combined data back to the dataset
+        dataset[spin_field] = combined_spin_data.transpose()
+        # Drop the individual spin data fields
+        dataset = dataset.drop_vars(packet_fields)
+
+    print(dataset.data_vars)
```
greglucas:
```suggestion
```

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -115,6 +116,39 @@ def segmented_pkts_fake_data():
     return dataset
 
 
+@pytest.fixture()
+def fake_spin_data():
+    dataset = xr.Dataset(
+        data_vars=dict(
+            num_completed=(["epoch"], np.array([0, 0])),
+            acq_start_sec=(["epoch"], np.array([0, 0])),
```
greglucas, line 124: Should this be aligned with epoch? i.e. I thinik that epoch = acq_start_sec + acq_start_subsec

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -266,3 +300,51 @@ def test_validate_parse_events(sample_data, attr_mgr):
...
+def test_organize_spin_data(fake_spin_data):
+    # Arrange
+    expected_dataset = xr.Dataset(
+        data_vars=dict(
+            num_completed=(["epoch"], np.array([0, 0])),
+            acq_start_sec=(["epoch"], np.array([0, 0])),
+            acq_start_subsec=(["epoch"], np.array([0, 0])),
+            acq_end_sec=(["epoch"], np.array([0, 0])),
+            acq_end_subsec=(["epoch"], np.array([0, 0])),
+            start_sec_spin=(
```
greglucas, line 315: Is "spin" here a coordinate, or should it be a data variable? I sort of assumed that these were a 1-1 mapping. A spin has an associated epoch, and epoch is already our coordinate here. So you'd want one or the other being coordinates, but not both, otherwise we have an (n x n) matrix that we index into as (i, i) instead of just i.

imap_processing/lo/l0/lo_science.py
sdhoyt: Confirmed with Lo that this needs to be reshaped

greglucas: When you say needs to be reshaped, what do you mean? This is the correct way to do it, or you need to have a different dimensionality?

sdhoyt: There are 28 spins per packet and the way the code is currently written is those 28 spins have a dimension called "spin" and each packet has an index (epoch). Colin and I had discussed both options and I forgot we settled on instead having the spins all along one dimension. So the 28 spins in each packet would all be lined up along one dimension "spin" and it would be a support variable so it doesn't need to be tied to an epoch in the CDF

imap_processing/lo/l0/lo_science.py
sdhoyt: Is this what you were thinking?

```python
combined_spin_data = xr.concat(dataset, data_vars=packet_fields, dim="spin")
```

It doesn't look like that will work. This appears to work if you have multiple datasets and want to combine data vars named the same across those two datasets, but it doesn't look like it works in this scenario.

imap_processing/tests/lo/test_lo_science.py
sdhoyt: I don't think I understand. Do you mean epoch should be recalculated based on the acq_start in the packet?

imap_processing/tests/lo/test_lo_science.py
sdhoyt: Yes, the update I'm making to the data shape will use spin as the coordinate rather than epoch. Each epoch has 28 spins, but I'm going to flatten those out along one dimension called "spin" which will be an index coordinate for the number of spins in the pointing

imap_processing/lo/l0/lo_science.py
greglucas: I am still not sure I totally follow what you are trying to do here at a quick glance, could you add a comment with dimensions / structure of what this becomes? I would have thought you wanted to do `[spin_field1_vals, spin_field2_vals, ...]`, concatenating each spin one after the other. But, this is interleaving the values because you are stacking and then raveling.

I thought you might still have an `epoch` dimension on this variable and each spin would be the extended concatenation like you described. So the variables would have dimensions: `(epoch, spin_number)` So it would be shape: `(npackets, 28)`.

imap_processing/lo/l0/lo_science.py
greglucas: Yes, that is what I was thinking. Thanks for checking and sorry for sending you on that incorrect chase!

imap_processing/tests/lo/test_lo_science.py
greglucas: That is what other instruments are doing. If you want to stamp your data to a time, the acquisition start is usually what you want to use rather than the MET. I meant should these be `np.array([0, 1])` to align with the packet incrementing too.

imap_processing/tests/lo/test_lo_science.py
greglucas: OK, now I think I might be understanding what you're trying to do now. One question though: Do you want each packet concatenated after each spin within that packet: `packet1-spin_field1, packet1-spin_field2, ... packet1-spin_field29, packet2-spin_field1, packet2-spin_field2, ...`?

imap_processing/lo/l0/lo_science.py
sdhoyt: Yeah I do want an array of [spin_field1_vals_packet1, spin_field2_vals_packet1, ..., spin_field1_vals_packet2, ....]. It looks like that's what the code is doing based on my unit test. Originally I had dimensions (epoch, spin_number), but I talked to Lo and they want it all in one dimension, so I need to tie it to the overall spin number for the pointing rather than the spin number in the packet or the epoch.

imap_processing/tests/lo/test_lo_science.py
sdhoyt: Ok sounds good, I didn't think those adjustments were being made until L1B, but that makes sense to do it here since all the data is available in the packet. For example the direct events use SHCOARSE for the Epoch for L1A, but in L1B I'm calculating a new epoch from the spin start time + de time (which is relative to the spin start time).

imap_processing/tests/lo/test_lo_science.py
sdhoyt: Correct. Based on my unit test it looks like that is correctly being concatenated.

sdhoyt:
After talking with @greglucas, it makes more sense to use the (epoch, spin) dimensions in the original PR. The PR was reverted to the previous commit and now uses the dimensions (epoch, spin) with a data shape (num packets, 28).

imap_processing/lo/l0/lo_science.py
greglucas:
```suggestion
```

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -115,6 +116,39 @@ def segmented_pkts_fake_data():
...
+    ]
+
+    for spin_field in spin_fields:
+        packet_fields = [f"{spin_field}_{i}" for i in range(1, 29)]
+        fake_array_val = 0
+        for packet_field in packet_fields:
+            dataset[packet_field] = xr.DataArray(
+                np.array([fake_array_val, fake_array_val + 1]), dims="epoch"
+            )
+            fake_array_val += 2
```
greglucas: I think this might make it sequential in the output and make it easier to follow.
```suggestion
                np.array([fake_array_val, fake_array_val + 28]), dims="epoch"
            )
            fake_array_val += 1
```

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -266,3 +300,51 @@ def test_validate_parse_events(sample_data, attr_mgr):
...
+    expected_dataset = xr.Dataset(
+        data_vars=dict(
+            num_completed=(["epoch"], np.array([0, 0])),
+            acq_start_sec=(["epoch"], np.array([0, 0])),
+            acq_start_subsec=(["epoch"], np.array([0, 0])),
+            acq_end_sec=(["epoch"], np.array([0, 0])),
+            acq_end_subsec=(["epoch"], np.array([0, 0])),
+            start_sec_spin=(
+                ["epoch", "spin"],
+                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
```
greglucas: bring the definition out to a variable outside and re-use them for all of your variables?
`data_by_epoch_spin = np.arange(56).reshape((2, 28))`

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -266,3 +300,51 @@ def test_validate_parse_events(sample_data, attr_mgr):
...
+            valid_phase_spin=(
+                ["epoch", "spin"],
+                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+            ),
+            period_source_spin=(
+                ["epoch", "spin"],
+                np.array([np.arange(0, 56, 2), np.arange(1, 57, 2)]),
+            ),
+        ),
+        coords=dict(epoch=(["epoch"], np.array([0, 1]))),
```
greglucas, line 344: Should spin also be a coordinate? Or we should add it possibly?
```suggestion
        coords=dict(epoch=(["epoch"], np.array([0, 1])), spin=(["spin"], np.arange(28))),
```

greglucas (approved):
Minor nits at this point. I like this ordering :+1:

imap_processing/tests/lo/test_lo_science.py
sdhoyt: I add that as a coordinate in the CDF attribute part of the code, so I have that being done on another branch which will be my next PR

greglucas (approved):

PR #1319, opened 2025-01-29
ENA Maps: UltraPointingSet and RectangularMap objects

nkerman:
# Change Summary
Defines generic and IMAP-Ultra-specific PointingSet objects, and generic and Rectangular grid Map objects.

## Overview
After conversations with @subagonsouth on recent PRs, we determined that a system for making L2 maps from L1C Pointing Sets (PSETs) should have classes to represent each. The Map class could then be subclassed to `RectangularMap`, `HealpixMap`, etc., while the PSET class could be subclassed to suit Ultra, Lo, Hi PSETs. 

This PR creates the abstract `Map` and `PointingSet` classes, as well as `RectangularMap` and `UltraPointingSet` subclasses. 

The pointing set contains enough information to be transformed from the IMAP_DPS frame to an HAE (`ECLIPJ2000`) frame.

Then the `RectangularMap` object is able to match pixels in angular space from the PSET to the Map, and bin values measured in the PSET into the Map's frame and spacing. At present, this can only be done in the "Push" method of index matching (see documentation in `imap_processing/ena_maps/ena_maps.py`).

Author's note: _Apologies if this is another large PR. There's not too much new logic, but the abstract classes add a lot of boilerplate, and I tried to add a bunch of documentation._ 

### Plots (for big picture/motivation)

While the code used to produce these plots is just a sketch, it shows the general premise of these objects, and how they will be used going forward for Ultra L2. The 1st gif shows the naive sum of the counts from several mocked pointing sets, while the 2nd gif shows the resultant map's values, with all the L1C counts projected to L2. If they're not properly synched up, try reloading the page 🤞.

ultra_psets_animation_original_ 1, 3, 5, 7

ultra_psets_animation_Ultra90 only_ 1, 3, 5, 7

## New Files

   - Defines classes for PSETs and Maps
   - Mock simple and recognizable L1C PSET data for use in tests and diagnostic plots

## Testing
Added tests for `ena_maps.py` in `imap_processing/tests/ena_maps/test_ena_maps.py`


demajistre:
A quick but probably annoying comment - I recommend, please rename the abstract class `Map` to something more specific - maybe `SkyMap`? Map is an overloaded term; there is a python `map` builtin that has nothing to do with the types of maps we're talking about.

demajistre:
A quick but probably annoying comment - I recommend, please rename the abstract class `Map` to something more specific - maybe `SkyMap`? Map is an overloaded term; there is a python `map` builtin that has nothing to do with the types of maps we're talking about.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,536 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SkyTilingType(Enum):
+    """Enumeration of the types of tiling used in the ENA maps."""
+
+    RECTANGULAR = "Rectangular"
+    HEALPIX = "Healpix"
+    ABSTRACT = "Abstract"
```
greglucas: What is abstract? I think we are currently only using the two tilings above, so just remove this one until needed later.

More generally, do we need this TilingType? You declare a class later called `RectangularMap`, that should be enough to indicate the tiling type I believe.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,536 @@
...
+        If the azimuth or elevation bin centers do not match the constructed grid.
+        Or if the azimuth or elevation bin spacing is not uniform.
+    ValueError
+        If multiple epochs are found in the dataset.
+    """
+
+    def __init__(
+        self,
+        l1c_dataset: xr.Dataset | pathlib.Path | str,
+        reference_frame: geometry.SpiceFrame = geometry.SpiceFrame.IMAP_DPS,
```
greglucas: Call this `pset_frame` to be explicit? I was confused whether the reference was the map or the pointing set.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,536 @@
...
+        order: typing.Literal["C"] | typing.Literal["F"] = "F",
+    ):
+        # History of reference frames to which the pointing set has been projected
+        # Current frame is the last element in the list, accessed as @property
+        self.reference_frame_history = [
```
greglucas: Why do we need this history? I think it is always projecting from the pointing set frame to whatever desired map frame, and never using the intermediaries.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,536 @@
...
+                (key, IndexMatchMethod.PUSH)
+                for key in pointing_set.data.data_vars.keys()
+            ]
+
+        for value_key, _ in value_keys:
+            if value_key not in pointing_set.data.data_vars:
+                raise ValueError(f"Value key {value_key} not found in pointing set.")
+
+        # Determine the indices of the map grid that correspond to the pointing set
+        matched_indices = self.match_pset_coords_to_indices(pointing_set)
```
greglucas: I think this could be cached for each "project_values" we do. The coordinate mapping is always going to be the same and we are going to want to transform many different quantities.

greglucas (commented):
I haven't been in a lot of these discussions, so I may not be fully up to speed on decisions. Feel free to disagree or point me to any previous discussions I may have missed.

Right now, I think this is a little too specific to individual PointingSets. By that I mean that the classes you have don't need to have the pointingset data in them.
Going by this diagram: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/232912737/ENA+L2+Processing
I would interpret this as having a class to contain your transforms and indices calculations, but then you would input the pointing set data to a function on that transform class to do the regridding. The class would set up your transforms and calculate/cache any bin indices needed, then go and do the computations later on whatever input is desired.

In the `pyproj` (geographic transforms) this is a Transformer class that takes in two Coordinate Reference Systems (CRSs) `trans = Transformer(frame_from, frame_to)`
https://pyproj4.github.io/pyproj/stable/examples.html#step-2-create-transformer-to-convert-from-crs-to-crs
where then we would have our transformer set up to do whatever we wanted, so we can then call `transformer.transform(lon, lat)`.

If we want to follow a similar methodology to that, then we would set up a "Transformer" going from a pointing set reference frame to a map projection. You might have a bit more complication because you'd need the pixelation schemes of the input/output, but in general I think you could keep all of that information on the Transformer and then call `Transformer.project_values(my_pset_flux)` which would do the regridding based on the precalculated indices made when instantiating the Transformer.

I really don't think we should be raveling / re-ordering things so much throughout these routines. We should ravel and transform initially and then do computations and then unravel at the very end if needed. But we hopefully don't need to carry that information around with us.

imap_processing/ena_maps/ena_maps.py
nkerman: Good question, and at present, we definitely don't need it, which may be enough reason to scrap it. I have a habit of adding things in that I think might be useful later, but that can cause bloat/confusion. 

---

This was a minor consideration, but when experimenting with SPICE and spiceypy, I realized that in some circumstances rounding errors, etc. can compound, so that if you have some point, say at `lon,lat = (180, 0)` deg, and you transform it from `frame1 -> frame2` and then back `frame2 -> frame1`, you get out `(180.00000000000001, 359.9999999999999)` or something like that. They matter less with higher precision floats, but it still concerned me a bit for when we start binning them at precise boarders. 

This made me think that fewer transformations are better, and that if, for some reason, we needed to get the coordinates in the original frame, we don't want to go `frame1->2->1`, but just go back and get the original. If the object had an internal history, we could make sure we weren't doing some long chain of `frame1->2->3->4->1`. 

That's the background, but you're right that we're probably only ever doing one transform on any given PSET, so let me know if you think I should cut it.

tmplummer:
@greglucas 
> Right now, I think this is a little too specific to individual PointingSets. By that I mean that the classes you have don't need to have the pointingset data in them.

This is probably my fault. I think that the problem we were running into is that there isn't much commonality between the PSET data products aside from they store some counts data. For example, Hi is storing the HAE lat/lon coordinates of each PSET bin. This means that Hi doesn't need any of the utility code that Nat has implemented for building the Ultra PSET grids and calculating the DPS frame coordinates. I thought that we could perhaps have a PSET class per ENA instrument to give a common interface for getting PSET bin coordinates.

imap_processing/ena_maps/ena_maps.py
nkerman: I agree we don't want to recalculate the indices for each quantity! Can you elaborate on what you see as a good way to cache this? It belongs to the combination of a PSET with it's frame and time, and a Map, and so I'm not sure the best place for the index mapping to "live", aside from within this function. 

The way I'd been intending to handle projecting multiple quantities (e.g. exposure time, counts) is to pass them into this function as 'value_keys'. It then finds the matching indices once, and bins each value_key with those indices.

imap_processing/ena_maps/ena_maps.py
nkerman: Done - would you also want the `Map.reference_frame` to be `Map.map_frame`, or is it clear with this change?

imap_processing/ena_maps/ena_maps.py
nkerman: >What is abstract? I think we are currently only using the two tilings above, so just remove this one until needed later.

Done - I thought it was needed for the ABC, but it isn't at all! My bad. 

> More generally, do we need this TilingType? You declare a class later called RectangularMap, that should be enough to indicate the tiling type I believe.

So, a PSET, at least the Ultra PSET can also use different tiling types (Healpix, Rectangular). It's currently only rectangular, but it seems like that's changing imminently. I had considered just defining a string to indicate the tiling type, but I figured that an Enum was safer.

nkerman:
> A quick but probably annoying comment - I recommend, please rename the abstract class `Map` to something more specific - maybe `SkyMap`? Map is an overloaded term; there is a python `map` builtin that has nothing to do with the types of maps we're talking about.

Sure thing. I'm changing the class names to `AbstractSkyMap`, `RectangularSkyMap`, etc.

imap_processing/ena_maps/ena_maps.py
nkerman: I removed the history!

imap_processing/ena_maps/ena_maps.py
nkerman: Internally, they're both stored as `spice_reference_frame` to work interchangeably with the index matching function

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+        Must contain the Spice frame in which the pixel centers are defined.
+    spatial_object_output_frame : PointingSet | AbstractSkyMap
+        The object containing a grid or tessellation of spatial pixels
+        into which the input spatial pixel centers will 'land', and be matched to
+        corresponding pixel 1D indices in the output frame.
+    event_time : float, optional
+        Event time at which to transform the input spatial object to the output frame.
+        This can be manually specified, e.g., for converting between Maps which do not
+        contain an epoch value.
+        The default value is None, in which case the event time of the PointingSet
```
tmplummer, line 99: ~~It isn't clear to me what the `PointingSet` is since both input and output frames could be PointingSets or neither.~~ Reading on, I get it now. It might be worth describing the allowed type combinations above rather than only in the `Raises` section.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+    else:
+        obj1_az_el_points_frame2 = obj1_az_el_points_frame1
+
+    obj1_az_el_points_frame2 = geometry.frame_transform_az_el(
+        et=event_time,
+        az_el=obj1_az_el_points_frame1,
+        from_frame=spatial_object_input_frame.spice_reference_frame,
+        to_frame=spatial_object_output_frame.spice_reference_frame,
+        degrees=False,
+    )
```
tmplummer: I think that this was supposed to be removed?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+    ):
+        obj1_az_el_points_frame2 = geometry.frame_transform_az_el(
+            et=event_time,
+            az_el=obj1_az_el_points_frame1,
+            from_frame=spatial_object_input_frame.spice_reference_frame,
+            to_frame=spatial_object_output_frame.spice_reference_frame,
+            degrees=False,
+        )
+    else:
+        obj1_az_el_points_frame2 = obj1_az_el_points_frame1
```
tmplummer: I am finding the original naming of `input` and `output` followed by this naming of `obj1`, `frame1`, `obj2`, `frame2` a little confusing. I wonder if this could be simplified by something like:
```suggestion
    transformed_input_az_el = spatial_object_input_frame.az_el_points
    if (
        spatial_object_input_frame.spice_reference_frame
        is not spatial_object_output_frame.spice_reference_frame
    ):
        transformed_input_az_el = geometry.frame_transform_az_el(
            et=event_time,
            az_el=obj1_az_el_points_frame1,
            from_frame=spatial_object_input_frame.spice_reference_frame,
            to_frame=spatial_object_output_frame.spice_reference_frame,
            degrees=False,
        )
```

tmplummer (commented):
Wow, this is a ton of work! Great job. I like the direction this has gone. I left a couple of comments but need to come back to this to finish reviewing.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+
+from imap_processing.ena_maps.utils import map_utils, spatial_utils
+from imap_processing.spice import geometry
+
+logger = logging.getLogger(__name__)
+
+
+# Define ravel order for unravelling, multi-indexing of rectangular grids.
+# Must be done as explicit typing Literal for mypy to accept it.
+RAVEL_ORDER = typing.cast(typing.Literal["C", "F"], "C")
```
tmplummer: Is this being done this way so that you can mock/override it out in tests? Otherwise, I'm not sure I see the point of just removing this as well as removing it from the ravel/reshape calls. You probably talked this over with Greg, so please ignore this if it has already been discussed.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+    captured in the map, but does not ensure that all pixels in the map receive data.
+
+    **Pull Method**
+
+    The "pull" method takes each pixel in the map grid and transforms its coordinates
+    to the frame of the pointing set, then determines into which pixel in the
+    pointing set grid the transformed map pixel falls.
+    This method ensures that all pixels in the map receive data, but can result in
+    some pointing set pixels not being captured in the map, and others being captured
+    multiple times.
```
tmplummer, line 53: Great documentation!

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+        contain an epoch value.
+        The default value is None, in which case the event time of the PointingSet
+        object is used.
+
+    Returns
+    -------
+    flat_indices_input_grid_output_frame : NDArray
+        1D array of pixel indices of the output object corresponding to each pixel in
+        the input object. The length of the array is equal to the number of pixels in
+        the input object, and may contain 0 or >1 occurrences of the same output index.
```
greglucas: minor nit, but that seems to remove == 1. Maybe the following, or "any number of".
```suggestion
        the input object, and may contain 0, 1, or multiple occurrences of the same output index.
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+        If the event time is not specified and both objects are SkyMaps.
+    NotImplementedError
+        If the output tiling type is HEALPIX. Will be implemented in the future.
+    ValueError
+        If the tiling type of the output frame is not RECTANGULAR or HEALPIX.
+    """
+    if isinstance(spatial_object_input_frame, PointingSet) and isinstance(
+        spatial_object_output_frame, PointingSet
+    ):
+        raise ValueError("Cannot match indices between two PointingSet objects.")
```
greglucas: This gets me to thinking "why not"? Is there an inherent limitation to doing this, or just that it isn't recommended because the frames won't overlap more than a few bins?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+
+    # If event_time is not specified, use event_time of the PointingSet, if present.
+    if event_time is None:
+        if isinstance(spatial_object_input_frame, PointingSet):
+            event_time = spatial_object_input_frame.data["epoch"].values
```
greglucas: Does this need to get a singular time out of the `values[0]` array?

imap_processing/ena_maps/ena_maps.py
greglucas: Another naming suggestion would be to potentially shorten the input argument to `input_object` / `output_object` and remove the spatial and frame since those are implied by using this function.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+                "Event time must be specified if both objects are SkyMaps."
+            )
+
+    obj1_az_el_points_frame1 = spatial_object_input_frame.az_el_points
+
+    # If the two objects are not already in the same frame,
+    # transform the input pixel centers to the output frame.
+    if (
+        spatial_object_input_frame.spice_reference_frame
+        is not spatial_object_output_frame.spice_reference_frame
```
greglucas: Can you use `==` here instead? Otherwise I could see someone loading two different spice frames that may not point to the same place in memory but are equal in the transform sense we care about.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+    ):
+        obj1_az_el_points_frame2 = geometry.frame_transform_az_el(
```
greglucas: Actually, should we push this logic into the `geometry.frame_transform_az_el()` function, so that it is a no-op when the frames are identical? Maybe leaving the to_degrees if a user specified that.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+    obj1_az_el_points_frame2 = geometry.frame_transform_az_el(
+        et=event_time,
+        az_el=obj1_az_el_points_frame1,
+        from_frame=spatial_object_input_frame.spice_reference_frame,
+        to_frame=spatial_object_output_frame.spice_reference_frame,
+        degrees=False,
+    )
+
+    # The way indices are matched depends on the tiling type of the 2nd object
+    if spatial_object_output_frame.tiling_type is SkyTilingType.RECTANGULAR:
```
greglucas: Be careful with using `is` too frequently. I think we want the equals comparison here as well in case there is a different instance floating about.
```suggestion
    if spatial_object_output_frame.tiling_type == SkyTilingType.RECTANGULAR:
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+            )
+            - 1
+        )
+        flat_indices_input_grid_output_frame = np.ravel_multi_index(
+            multi_index=(az_indices, el_indices),
+            dims=(
+                len(spatial_object_output_frame.sky_grid.az_bin_midpoints),
+                len(spatial_object_output_frame.sky_grid.el_bin_midpoints),
+            ),
+            order=RAVEL_ORDER,
```
greglucas: Do you need to specify RAVEL_ORDER? I worry you are taking a module-level constant which someone could change on you. I still don't think this is necessary and is an implementation detail under the hood. It is only necessary because you are doing things in 1D rather than 2D, so how you go from 2D -> 1D -> 2D doesn't really matter as long as you are the one controlling the steps.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+        If multiple epochs are found in the dataset.
+    """
+
+    def __init__(
+        self,
+        l1c_dataset: xr.Dataset | pathlib.Path | str,
+        spice_reference_frame: geometry.SpiceFrame = geometry.SpiceFrame.IMAP_DPS,
+    ):
+        # Store the reference frame of the pointing set
+        self.spice_reference_frame = spice_reference_frame
```
greglucas, line 274: Out of curiosity (no need to do this here!), is there a way to store this within your `l1c_dataset`? It seems like it might be useful if we could somehow store these within the dataset and reference them together rather than separately.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+
+        # Read in the data and store the xarray dataset as data attr
+        if isinstance(l1c_dataset, (str, pathlib.Path)):
+            self.path: str | None = str(l1c_dataset)
+            self.data = xr.open_dataset(l1c_dataset)
+        elif isinstance(l1c_dataset, xr.Dataset):
+            self.path = None
+            self.data = l1c_dataset
```
greglucas: I don't think we need to store the path since you aren't using it later. Additionally, I thought `xr` couldn't open cdf files and we needed to use our cdflib helpers?

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/d9e683ceacc69c83c07330d9a19de572b99baae6/imap_processing/cdf/utils.py#L21-L25

```suggestion
        if isinstance(l1c_dataset, (str, pathlib.Path)):
            self.data = load_cdf(l1c_dataset)
        elif isinstance(l1c_dataset, xr.Dataset):
            self.data = l1c_dataset
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+
+        # Unwrap the az, el, solid angle grids to series of points tiling the sky
+        az_points = self.sky_grid.az_grid.ravel(order=RAVEL_ORDER)
+        el_points = self.sky_grid.el_grid.ravel(order=RAVEL_ORDER)
+        self.az_el_points = np.column_stack((az_points, el_points))
+        self.solid_angle_points = self.solid_angle_grid.ravel(order=RAVEL_ORDER)
+        self.num_points = self.az_el_points.shape[0]
+
+        # Initialize empty data dictionary to store map data
+        self.data_dict: dict[str, NDArray] = {}
```
greglucas, line 423: Do we know what we want the map dataset to look like at this point? It looks like we've defined all of the dimensions/coordinates above, so I'm curious if we should initialize the empty dataset with all of that right away here.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+
+    def project_pset_values_to_map(
+        self,
+        pointing_set: PointingSet,
+        value_keys: list[tuple[str, IndexMatchMethod]] | None = None,
+    ) -> None:
+        """
+        Project a pointing set's values to the map grid.
+
+        Here, the term "project" refers to the process of determining which pixels in
```
greglucas, line 434: I thought you decided to change everything to "transform"?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,543 @@
...
+        the map grid correspond to which pixels in the pointing set grid, and then
+        binning the values at those indices from the pointing set to the map.
+
+        Parameters
+        ----------
+        pointing_set : PointingSet
+            The pointing set containing the values to project to the map.
+        value_keys : list[tuple[str, IndexMatchMethod]] | None
```
greglucas, line 442: Do we need the `IndexMatchMethod` anymore? I thought we might be able to get rid of this with the input/output swapping from before.

greglucas (commented):
Really nice, I like this. I think there are still some simplifications that can be made, but some of them can be follow-ups as well. This is a great start of the organization!

I'm still wondering if we need to distinguish between the "PointingSet" vs "SkyMap" classes. I feel like these are "AbstractTiling"s and don't necessarily need to be separate. I can go from any tiling + spice_frame to any other tiling + spice_frame. You could have a classmethod `tiling = Tiling.from_pset(xr_dataset)` that would create the tiling associated with that dataset (or `Tiling.from_coords()` for a skymap). Then I'd view it as a function call `output_dataset += transform(xr_dataset[subset_of_vars], from_tiling, to_tiling)`

I think I am just getting hung up on how much is going on within the classes to store things that feel like they are already defined elsewhere, or perhaps don't need to be stored.

Again, though, I think that can be changed later if desired, so just view that as me putting thoughts down and not a requirement for this PR.

imap_processing/ena_maps/ena_maps.py
nkerman: I talked a bit more with Tim about this - In my opinion, transform and project mean different things. Transform is to apply a rotation to some vector or az/el coordinates and get the _transformed vector or coordinates_. Project is what you do to measured values defined at some point in space (e.g. counts) to get them onto a map in some frame. I think project is the right word to use here for taking values from a PSET and bringing them into a map.

imap_processing/ena_maps/ena_maps.py
nkerman: The Map would still need to know whether a given variable should be pushed or pulled (not the same for all values in the case of Rectangular Ultra PSETs). That will determine which order it passes the objects into the transformer function.

imap_processing/ena_maps/ena_maps.py
nkerman: We know the size the map (the number of pixels) but not which variables we're carrying over from the PSETs. E.g. Ultra will want to project "sensitivity" while that's not something Lo has to deal with.

imap_processing/ena_maps/ena_maps.py
nkerman: Good catch, and thanks for the suggestion!

imap_processing/ena_maps/ena_maps.py
greglucas: Can these be separate calls then? One for the Pull data, one for the push data `project_data(xr_dataset[variables], IndexMatchMethod)`

imap_processing/ena_maps/ena_maps.py
nkerman: Good question! I can ask Laura, since it would be at the L1C level. It would need to be stored as a string, and then read back into the SpiceFrame object, but that's certainly doable. It should always be IMAP_DPS for Ultra.

imap_processing/ena_maps/ena_maps.py
nkerman: > so how you go from 2D -> 1D -> 2D doesn't really matter as long as you are the one controlling the steps.

I agree it generally doesn't matter how you unwrap the PSETs, but I think it really matters when doing ravel_multi_index with Rectangular Maps. At the very least, it's vital to making sure the test values agree and aren't unwrapped/re-wrapped transpositions of the data. If you'd prefer, I could make it so that only the ravel_multi_index is specified. I could let it always use numpy's default, but I'm somewhat uncomfortable with that... maybe without good reason 🤷

imap_processing/ena_maps/ena_maps.py
nkerman: Sure!

imap_processing/ena_maps/ena_maps.py
nkerman: I read that when comparing Enums, you should use `is`, not `==`, but don't have a strong opinion. As long as it's safe, I can switch these.

imap_processing/ena_maps/ena_maps.py
nkerman: Sure - I can switch it in geometry.

imap_processing/ena_maps/ena_maps.py
nkerman: Ditto to the comment about is below. I thought you're 'supposed to' use is to compare enums?

imap_processing/ena_maps/ena_maps.py
nkerman: No, the geometry code is happy with a single value in the list.

imap_processing/ena_maps/ena_maps.py
nkerman: The problem is that it would be transforming from IMAP_DPS at time 1 to the same frame (IMAP_DPS) at time 2. As far as I can tell, this would never make any sense and picking one time or another for the frame_transformation could be dangerous.

imap_processing/ena_maps/ena_maps.py
greglucas: Yes, I agree with all of that :) I would prefer to default to numpy's decision unless there is a good reason not to and you need a specific ravel/reshape ordering. numpy's default is "C" in every function, if it was "K" (how it occurs in memory and dependent on the order of your array) then I think there could be issues with letting it default. But in our case, I think it will make the code much simpler if you don't have to keep track of that everywhere.

nkerman: Okay! I'll remove it!

imap_processing/ena_maps/ena_maps.py
greglucas: Nice explanation, that makes sense to me. Thanks!

imap_processing/ena_maps/ena_maps.py
greglucas: Oh cool. So this is specific to Enum's because they are singleton's and guaranteed to point to the same object. I'm fine with this then. Thanks for the explanation and teaching me what I should do :)

imap_processing/ena_maps/ena_maps.py
greglucas: As a quick note, if you haven't done this already. This might be a nice separate PR to just do this single case over there with a quick test to make sure you don't transform something that doesn't need transforming.

Then here, you could just call it regardless and not worry about what happens with that other PR.

nkerman: Okay - I'll leave this out of this PR! I'll try to address that one soon.

imap_processing/ena_maps/ena_maps.py
nkerman: Done! I switched to `"input_object"` and am now using the explicit formulation "`input_obj_az_el_output_frame`". It's a bit wordier but describes exactly what it is.

imap_processing/ena_maps/ena_maps.py
nkerman: Done!

nkerman:
Thanks for all your suggestions @subagonsouth @greglucas ! I think I've changed all the definite change requests, added documentation, etc. and fixed the broken tests.

imap_processing/tests/ultra/test_data/mock_data.py
```diff
@@ -0,0 +1,159 @@
...
+import spiceypy as spice
+import xarray as xr
+
+from imap_processing.spice.kernels import ensure_spice
+from imap_processing.ultra.l1c.ultra_l1c_pset_bins import build_energy_bins
+
+DEFAULT_SPACING_DEG_L1C = 0.5
+
+
+def mock_l1c_pset_product(
```
greglucas, line 13: fixture?

imap_processing/tests/ultra/test_data/mock_data.py
```diff
@@ -0,0 +1,159 @@
...
+    _, energy_bin_midpoints = build_energy_bins()
+    energy_bin_midpoints = energy_bin_midpoints[1:]
+    num_energy_bins = len(energy_bin_midpoints)
+
+    grid_shape = (num_lon_bins, num_lat_bins, num_energy_bins)
+
+    def get_binomial_counts(distance_scaling, lon_bin, central_lon_bin):
+        # Note, this is not quite correct, as it won't wrap around at 720
+        distance_lon_bin = np.abs(lon_bin - central_lon_bin)
+        return np.random.binomial(
```
greglucas: Is this reproducible? I think we'd want to seed the generator first so we can reproduce this every time we test. As an FYI, numpy suggests moving to the generator calls now, so initializing with `rng = np.random.default_rng(seed); rng.binomial(...)`

https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.binomial.html

imap_processing/ena_maps/ena_maps.py
```diff
@@ -0,0 +1,521 @@
...
+            )
+
+        for value_key in value_keys:
+            # If multiple spatial axes present
+            # (i.e (az, el) for rectangular coordinate PSET),
+            # flatten them in the values array to match the raveled indices
+            raveled_pset_data = np.reshape(
+                np.array(pointing_set.data[value_key]),
+                (pointing_set.num_points, -1),
+            )
```
greglucas: Can you reshape the data array directly to avoid creating copies here?
```suggestion
            raveled_pset_data = pointing_set.data[value_key].reshape(pointing_set.num_points, -1)
```

greglucas (approved):
Thanks for simplifying the function inputs, I find this much easier to follow!

I think this is good to go, but it would be good to hear from others too because it is going to be used so broadly. Making sure the function definitions / class layouts make sense to others too. I still don't personally love the PointingSet having class methods approach and would find it easier to follow with simpler class definitions, so I will put that down here for posterity for others. I think I'm the only one who has brought this up though and multiple others have suggested this approach, so I'm definitely not willing to die on this hill and you still have my approval :)

```python
class Tesselation:
    # No notion of data, just griddings
    grid_points
    spice_frame

class Transformer:
    input_tesselation
    output_tesselation
    # Calculate and cache the mapping of indices from the input tesselation to output tesselation
    _indices

    def project(arr):
        # Project input dataset / dataarray (required to be same dimensions as input_tesselation)
        # to an output dataset / dataarray with the same dimensions as the output_tesselation

# In my Ultra function code, use the generic transformer rather than in yet another class I have to keep track of.
output_dataset = create_empty_map_dataset()
for var in ["counts", "uncertainty"]:
    output_dataset[var] += transformer.project(pset[var])
```

imap_processing/tests/ultra/test_data/mock_data.py
nkerman: I don't see this as a fixture directly, since it's useful to be able to call it with different spacings, stripe center longitudes, etc. What I have is a fixture in `test_ena_maps.py` that calls this to generate multiple l1c products with stripes at different longitudes in the IMAP_DPS frame (see gifs above ^ )

I think that would be harder to do if I made this a fixture, right?

imap_processing/tests/ultra/test_data/mock_data.py
nkerman: Ah! Good to know about the generator. Switching now!

nkerman: (And adding a seed)

imap_processing/tests/ultra/test_data/mock_data.py
greglucas: Ahh, yes I missed that parametrization. Good call on not having it be a fixture. It is doable as a fixture but as you say it is harder and probably not a major benefit here.

imap_processing/ena_maps/ena_maps.py
nkerman: Sort of - There's no `xr.DataArray.reshape()` method, but I can do a `pointing_set.data[value_key]`**.data**`.reshape(pointing_set.num_points, -1)`. I'm not actually sure if that performs a copy. I think that it doesn't, since `.data` should just access the internal numpy representation.

nkerman:
> Thanks for simplifying the function inputs, I find this much easier to follow!
> 
> I think this is good to go, but it would be good to hear from others too because it is going to be used so broadly. Making sure the function definitions / class layouts make sense to others too. I still don't personally love the PointingSet having class methods approach and would find it easier to follow with simpler class definitions, so I will put that down here for posterity for others. I think I'm the only one who has brought this up though and multiple others have suggested this approach, so I'm definitely not willing to die on this hill and you still have my approval :)
> 
> ```python
> class Tesselation:
>     # No notion of data, just griddings
>     grid_points
>     spice_frame
> 
> class Transformer:
>     input_tesselation
>     output_tesselation
>     # Calculate and cache the mapping of indices from the input tesselation to output tesselation
>     _indices
> 
>     def project(arr):
>         # Project input dataset / dataarray (required to be same dimensions as input_tesselation)
>         # to an output dataset / dataarray with the same dimensions as the output_tesselation
> 
> # In my Ultra function code, use the generic transformer rather than in yet another class I have to keep track of.
> output_dataset = create_empty_map_dataset()
> for var in ["counts", "uncertainty"]:
>     output_dataset[var] += transformer.project(pset[var])
> ```

---

Hi @greglucas , thanks a ton for all this feedback! I really appreciate the similarities between the spatial objects: `PointingSet`s and `SkyMap`s that I think you're getting at here. This is definitely informed by my desire to avoid changing the structure again, but I don't think that the `Transformer` approach necessarily removes much of the complexity/passing around of multiple classes/objects. One thing I'm not seeing in this excellent pseudocode structure, is that you would need to build a Transformer for every combination of PointingSet and SkyMap - which would mean making one for each Pointing Set, since their transformations would be different as the DPS frame changes between Pointing Sets. 

With that in mind, I'm not sure how much simpler this actually ends up being. Here's what the entirety of projection looks like with the current approach (it's the code to make the map in those gifs ^), and this could be tightened up from where it is now. 

```python
ultra_psets = [
    ena_maps.UltraPointingSet(
        spice_reference_frame=geometry.SpiceFrame.IMAP_DPS, l1c_dataset=l1c_product,
    )
    for l1c_product in fake_l1c_products
]
# Make a rectangular map object
rectangular_map = ena_maps.RectangularSkyMap(
    spacing_deg=2,
    spice_frame=geometry.SpiceFrame.ECLIPJ2000,
)
# Loop through the UltraPointingSet objects and project their values to the rectangular map
for i, ultra_pset in enumerate(ultra_psets):
    rectangular_map.project_pset_values_to_map(
        ultra_pset,
        value_keys=[
            "counts", "exposure_time", "sensitivity"
        ],
        index_match_method=ena_maps.IndexMatchMethod.PUSH
    )
```

I think in the alternative approach it would need to look something like: 

```python
# You do still need to keep track of the type of tiling and spacing of the output Map
l2_map_of_sky_tiling = Tiling(tiling_type=Rectangular, spacing_deg=2)

variables_to_accumulate = ["counts", "exposure_time", "sensitivity"]
output_dataset = create_empty_map_dataset(size=l2_map_of_sky_tiling.num_points)

for l1c_product in fake_l1c_products:
    pset_specific_transformer = Transformer(
        input_tiling=Tiling(tiling_type=Rectangular, spacing_deg=0.5, event_time = ...),
        output_tiling=l2_map_of_sky_tiling,
    )
    push_indices = pset_specific_transformer.index_matching(index_match_type=PUSH)
    for variable in variables_to_accumulate:
        output_dataset[variable] += map_utils.bin_single_array_at_indices(
            l1c_product.data[variable].values.ravel(),
            push_indices,
        )
```

This seems pretty much equivalent to me in what they do and how they do it under the hood, but to me seems a bit more confusing.


> I still don't personally love the PointingSet having class methods approach 

On this, I'll note that, since extracting `match_coords_to_indices()` to a 'loose'/globally scoped function, the only method on either of the classes is `SkyMap.project_pset_values_to_map()`, (and the initialization/repr methods). Keeping them as different classes with their own, potentially instrument-specific initializations, lets us handle their quirks flexibly. E.g., the index-matching function can look for the `PointingSet` object to determine the event_time.

tmplummer (approved):
Thanks for pushing through on this Nat. It has been a challenging PR process.

imap_processing/ena_maps/ena_maps.py
nkerman: This was removed in a recent update.

imap_processing/ena_maps/ena_maps.py
nkerman: Thanks!

nkerman:
Thanks Tim! I really appreciate all your **many** detailed reviews, and I'm excited we can move forward with incorporating the maps into the main pipeline!

PR #1320, opened 2025-01-30
CoDICE: More flexible way to defining dimensions

bourque:
# Change Summary

## Overview
This PR makes changes to the CoDICE L1a pipeline to allow more flexibility in defining dimensions for the data product, which in turn will help me more easily validate L1a data products. The situation that this PR solves is that CoDICE L1a data products have varying amount and order of dimensions for the data arrays (e.g. some products use a `spin_sector` dimension, others do not; some products place the energy dimensions before position, others do the opposite, etc). These changes allow for me to define the configuration for these in the `DATA_PRODUCE_CONFIGURATIONS` dictionary and helps clean up the code a bit.

## Updated Files

  - Re-ordered dimensions for `lo-(n)sw-angular` and `lo-(n)sw-species` data products to match validation data 
- `imap_processing/codice/codice_l1a.py` and `imap_processing/codice/constants.py`
  - Changes to allow more flexibility in defining dimensions and reshaping the data arrays 
  - Updated tests to get `lo-(n)sw-angular` validation working with the new changes, and now comparing data array shapes in validation test.


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque, line 1: Changes in this file are to get some data products in the correct order for dimensions

imap_processing/codice/codice_l1a.py
```diff
@@ -149,17 +149,16 @@ def define_coordinates(self) -> None:
         """
         self.coords = {}
 
-        for name in self.config["coords"]:
+        coord_names = ["epoch", *list(self.config["dims"].keys())]
```
bourque, line 152: Since `self.config["dims"]` basically has the same values as `self.config["coords"]`, I was able to consolidate here and just use the `dims` list to define the coordinates.

imap_processing/codice/codice_l1a.py
```diff
@@ -193,26 +192,37 @@ def define_data_variables(self) -> xr.Dataset:
...
         ):
-            counter_data = all_data[:, counter, :, :, :]
+            # Dynamically determine the slicing required to extract the counter
+            # data, since the number of dimensions may vary. The counter
+            # dimension is always the first
+            slicing: list[Any] = [slice(None)] * all_data.ndim
+            slicing[1] = counter
+
+            # Extract the counter data
+            counter_data = all_data[tuple(slicing)]
```
bourque: This is pretty janky and I couldn't think of a better way to do this. Happy to hear if anyone has better ideas.

greglucas: I'm not sure what exactly you're trying to do here, but perhaps using an elipses?
`counter_data = all_data[:, counter, ...]`

imap_processing/codice/codice_l1a.py
```diff
@@ -391,45 +401,26 @@ def reshape_data(self) -> None:
...
+
+        # For each packet/epoch, reshape the data along these dimensions
+        for packet_data in self.raw_data:
+            if packet_data:
+                reshaped_packet_data = np.array(packet_data, dtype=np.uint32).reshape(
+                    reshape_dims
+                )
+                self.data.append(reshaped_packet_data)
+            else:
+                self.data.append(None)
```
greglucas: Do you want `None` here? If there is no data in the packet is that worth raising or just passing here to continue on?

imap_processing/codice/constants.py
```diff
@@ -12,8 +12,7 @@
 ESA = ElectroStatic Analyzer
 """
 
-# TODO: What to do in the case of a value of 255 in LOSSY_A and LOSSY_B
-#       compression? (Joey uses 0x100000000)
+from collections import OrderedDict
```
greglucas: Do you need an OrderedDict? Python dictionaries are now order by default.
https://docs.python.org/3/library/collections.html#ordereddict-objects

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -203,30 +203,34 @@ def test_l1a_data_array_values(test_l1a_data: xr.Dataset, index):
...
+            # Ensure the data array shapes are equal
+            assert (
+                processed_dataset[counter].data.shape
+                == validation_dataset[counter].data.shape
             )
+
+            # TODO: Once Joey and I figure out some small discrepancies with
+            #       some data products, we may be able to just compare the data
+            #       arrays directly (i.e. assert_array_equal on just the .data
+            #       attribute, instead of sum or shape.
```
greglucas: This doesn't make sense to me that the sum would be good but the array comparison not. Are you just getting very minimal differences or a 1 and -1 difference that cancel out in the sum?

There is also `xr.testing.assert_equal()` for comparing the datasets directly that I saw someone else using too.

imap_processing/codice/codice_l1a.py
bourque: Thanks, this is much better 😌

imap_processing/codice/codice_l1a.py
bourque: I now realize that these conditional statements are leftover from several months ago when I was dealing with decompression issues in the input data, which have since been resolved. So I think this can be removed altogether.

imap_processing/codice/constants.py
bourque: Ah, good to know! In that case I believe I can use just a regular dictionary

imap_processing/tests/codice/test_codice_l1a.py
bourque: Using `xr.testing.assert_equal()` on the entirety of the data arrays is what I am striving towards. For now, I split up the checks by `sum()` and `shape()` because i'm still sorting out some value differences in some of the data products with Joey. So it has been useful for me to easily identify issues. 

But yes, eventually I would like this test function to just compare the arrays, which should prove that I am producing valid data.

greglucas: My curiosity still stands though. How can your sum be valid but not the individual values? Do you really want an `assert_almost_equal()` instead of exactly equal?

bourque: Currently for some data products (not listed in this test because they would break, but I've been playing around with them locally) have the correct values but differ in shape from the validation data. I am still sorting out some of these differences with Joey. So basically, this is just a workaround / playground for testing as I am sorting these things out, hence why I put the `TODO` note there -- eventually this will all get moved to just comparing the data arrays with `assert_equal()`.

greglucas: > not listed in this test because they would break, but I've been playing around with them locally

My preference would be to leave these as the real test (not separate sum and shape assertions) for now then if they do pass. The testing assert_array_equal should tell you whether the shapes, specific elements etc are different in your local debugging, so should be nicer to debug output with too.

greglucas (approved):
I have a slight preference for changing to the actual array comparisons rather than the sum/shape, but this looks good other than that minor nit.

PR #1321, opened 2025-01-30
Excluding tests from build

maxinelasp:
# Change Summary
Exclude tests from build. Went from 36M to 14M when building locally. This excludes all unit tests and any large testing files.

## Overview

Just added the test path to the Poetry build config.


greglucas (approved):
Thanks! Looks good to me and really reduces the size of our distribution.

bourque (approved):

PR #1324, opened 2025-01-31
Make vectorization consistent across all time conversion functions

### LINKED ISSUE #1305: SPICE - use np.vectorize throughout
### Algorithm Description:

Convert existing loops wrapped with np.array() to use np.vectorize. Might need to figure out how to handle the scalar case better b/c np.vectorize returns an array with shape=().

tmplummer:
Fix bug causing iteration over a 0-d array when numpy.vectorize returns a 0-d array rather than a scalar

# Change Summary
This makes the vectorization of spiceypy time functions consistent.

It also fixes a bug that was caused by calling `et_to_utc` on a 0-d vector. The bug was caused by a 0-d vector being Iterable but not able to be used in list comprehension. This is ultimately due to `np.vectorize` undesirable behavior of returning a 0-d array instead of a scalar for scalar inputs. I find the fix to be a bit hacky, so if someone has an alternative, cleaner approach, let me know.

## Updated Files
   - Add the `_vectorize` method which wraps `np.vectorize` and returns that result with `[()]` which converts 0-d vectors to scalars and leaves everything else as is.
   - Use the `_vectorize` for all functions that were manually vectorizing calls to spiceypy functions.
   - Add checks that scalar input -> scalar output for vectorized functions.

Closes: #1305

### DIFF
--- a/imap_processing/spice/time.py
+++ b/imap_processing/spice/time.py
@@ -171,7 +171,7 @@ def met_to_datetime64(
     numpy.ndarray[str]
         The mission elapsed time converted to UTC string.
     """
-    return np.array(met_to_utc(met), dtype=np.datetime64)
+    return np.array(met_to_utc(met), dtype=np.datetime64)[()]
 
 
 @typing.no_type_check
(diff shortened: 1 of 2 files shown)


greglucas (approved):
I think the simpler way would be if these functions took arrays as inputs directly, but that would require adding that upstream in spiceypy.

PR #1335, opened 2025-02-04
Add decompression code for IDEX L1A. 

### LINKED ISSUE #1182: Add decompression logic to L1a processing
The attribute 'IDX__SCI0COMP' signifies whether the packet data needs to be decompressed using [RICE decoding](https://en.wikipedia.org/wiki/Golomb_coding). This needs to be implemented in l1a.

lacoak21:
# Change Summary

closes #1182

## Overview
Adds logic to check waveform compression status using 'IDX__SCI0COMP' instrument setting and implements Golomb-RICE decompression functionality. Original rice decode code was provided by the IDEX team, and written by Corinne Wuerthner for SUDA. 

## New Dependencies
This PR will add the bitstring package as a dependency 

## New Files
- imap_processing/idex/idex_rice_decode.py
   - Golomb-RICE decompression functionality.
- imap_processing/tests/idex/test_data/compressed_2023_102_14_24_55.pkts
   - Contains compressed waveform data. This is used to test the decompression functionality.
- imap_processing/tests/idex/test_data/non_compressed_2023_102_14_22_26.pkts
   - Contains the same uncompressed data as above.

## Updated Files
   - Decompress waveform data if need be. 
- imap_processing/tests/idex/conftest.py
   - Update filepath
- imap_processing/tests/idex/test_data/impact_14_tof_high_data.txt
   - Update filepath
- imap_processing/tests/idex/test_data/imap_idex_l0_raw_20231214_v001.pkts
   - Update filepath
- imap_processing/tests/idex/test_idex_l0.py
  - Update filepath
- imap_processing/tests/test_utils.py
  - Update filepath
- imap_processing/tests/idex/test_idex_l1a.p
   - Test idex_rice_decode code.


imap_processing/idex/idex_rice_decode.py
```diff
@@ -0,0 +1,173 @@
+"""
```
lacoak21: Most of the work here was to update the inputs and outputs to work with IDEX data, make some branching logic cleaner, and add comments, rename variables ect.

greglucas:
Are we able to avoid bitstring? It is really quite slow so we tried to remove it as a dependency in a few previous PRs. See here for an example rework within IDEX: https://github.com/IMAP-Science-Operations-Center/imap_processing/commit/0fa8bf2b1986faaca99cf8eaba3957390ecc17e8

imap_processing/idex/idex_l1a.py
```diff
@@ -391,22 +401,31 @@ def _parse_high_sample_waveform(self, waveform_raw: str) -> list[int]:
...
-                int(waveform_raw[i + 12 : i + 22], 2),
-                int(waveform_raw[i + 22 : i + 32], 2),
-            ]
-        return ints[:-4]  # Remove last 4 numbers
+        samples = self.MAX_HIGH_BLOCKS * self.NUMBER_SAMPLES_PER_HIGH_SAMPLE_BLOCK
+        if self.compressed.raw_value == 1:
+            ints: list[int] = idex_rice_decode(
+                waveform_raw, nbit10=True, sample_count=samples
+            )
+            return ints[:-3]
```
greglucas: Is there a reason not to trim the extra ints off in the `idex_rice_decode()` function? i.e. return `idex_rice_decode()` directly

imap_processing/idex/idex_l1a.py
```diff
@@ -391,22 +401,31 @@ def _parse_high_sample_waveform(self, waveform_raw: str) -> list[int]:
...
+        else:
+            ints = []
```
greglucas: Is there a name for this encoding / do you want to move this into a helper function like the rice decoding too? It is small enough to leave in-line here, but just curious.

imap_processing/idex/idex_rice_decode.py
```diff
@@ -0,0 +1,205 @@
...
+
+            sub_frame_data.append(d1)
+            sample_count += 1
+
+            # A 'psel' value of zero assumes that every sample in the frame is equal
+            # to the first sample in the frame. In this case, the first sample is
+            # stored as the original NBIT binary value
+            if psel == 0:
+                for _ in range(SUB_FRAME_SIZE - 1):
+                    sub_frame_data.append(d1)
```
greglucas: I think this might be equivalent and avoid the for-loop.
```suggestion
                sub_frame_data.extend([d1] * (SUB_FRAME_SIZE - 1))
```

imap_processing/idex/idex_rice_decode.py
```diff
@@ -0,0 +1,205 @@
...
+            # into two parts. The least significant k bits of the value are called
+            # the remainder (r). The other part of the value (not included in the
+            # remainder) is called the quotient (q).
+            # The remapped quotient is unary encoded by including a number of 0 bits
+            # equal to the value of the quotient, followed by a single '1' bit.
+            q = 0
+            while read_bits(bits, bp, 1)[0] == 0:
+                bp += 1
+                q += 1
+            bp += 1
```
greglucas: I'm not totally following this at a quick glance. We are reading incrementally more values each time (1, 2, 3, ... bits) but discarding the previous values? Do we need to look for zeros in the other slots as well to make sure it is full of zeros? It might help to add some examples with numbers here too.

(curious if you could use `bits[bp:].find("1")` or some other string lookup to get your first index rather than all these loops)

imap_processing/idex/idex_rice_decode.py
```diff
@@ -0,0 +1,205 @@
...
+            # If the value of the quotient is equal to or larger than 47, then a
+            # special symbol is used to denote that this particular residual value
+            # is not rice encoded, but that this special symbol is followed by the
+            # raw binary representation of the residual value using a (N_BITS+2)
+            # bit binary number. This special symbol is simply 47 zeros followed
+            # by a one.
+            if q == 47:
+                d1, bp = read_bits(bits, bp, n_bits + 2, True)
+            else:
+                if q & 0x1:
```
greglucas: Be consistent with hex vs int comparisons? `q == 1`

imap_processing/idex/idex_rice_decode.py
```diff
@@ -0,0 +1,205 @@
...
+                d1 = (
+                    d1
+                    + 2 * sub_frame_data[(sample_count - 1)]
+                    - sub_frame_data[(sample_count - 2)]
+                )
+
+            if (d1 > 2**n_bits) or (d1 < -(2**n_bits)):
+                raise ValueError(
+                    f"Overflow Error while decoding subframe "
+                    f"k = {k}, q = {q}, r = {r}, d1 = {d1}\n"
```
greglucas: FYI that this can be handled directly now with simpler syntax
https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging
```suggestion
                    f"{k=}, {q=}, {r=}, {d1=}\n"
```

imap_processing/idex/idex_rice_decode.py
```diff
@@ -0,0 +1,205 @@
...
+                    f"DataOut = {sub_frame_data}"
+                )
+
+            sub_frame_data.append(d1)
+            sample_count += 1
+
+    return sub_frame_data, sample_count, bp
+
+
+def idex_rice_decode(
```
greglucas: I find functions named the same as modules slightly confusing (which one are you using, the module or the function) `datetime.datetime` is a good example that is confusing IMO. Quick suggestion/thought: name the module `decoding.py`, and the function `rice_decoder()` (you are already in the idex module namespace).

pyproject.toml
```diff
@@ -56,6 +56,7 @@ sphinx = {version="*", optional=true}
 sphinxcontrib-openapi = {version="^0.8.3", optional=true}
 mypy = {version="1.10.1", optional=true}
 requests = {version = "^2.32.3", optional = true}
+bitstring = "^4.3.0"
```
greglucas: Wow, that was a super fast update!! Thanks for removing this :)
```suggestion
```

imap_processing/tests/idex/test_idex_l1a.py
```diff
@@ -89,9 +92,40 @@ def test_idex_tof_high_data_from_cdf(decom_test_data: xr.Dataset):
...
+        f"{imap_module_directory}/tests/idex/test_data/impact_14_tof_high_data.txt"
+    ) as f:
         data = np.array([int(line.rstrip()) for line in f])
 
     file_name = write_cdf(decom_test_data)
     l1_data = load_cdf(file_name)
     assert (l1_data["TOF_High"][13].data == data).all()
+
+
+def test_compressed_packet():
```
greglucas, line 107: Can you make unit tests for any of the decompression schemes? The new read_bits signed/unsigned should be tested as well. I think you could make some individual strings of 0s and 1s to pass into the functions to verify what you are expecting and give a better example of what we are anticipating as well. Right now, we are verifying it against CCSDS packets that someone else has given us and aren't entirely clear what is in those and how robust that data is against various cases.

greglucas (commented):
This is a ton of work here. Very nice job figuring it all out. Most of my suggestions are related to you having the knowledge of how this works now, but me coming in cold finds it quite confusing, so some more comments/examples would be helpful.

imap_processing/idex/idex_l1a.py
```diff
@@ -421,12 +440,18 @@ def _parse_low_sample_waveform(self, waveform_raw: str) -> list[int]:
...
-            ]
+        samples = self.MAX_LOW_BLOCKS * self.NUMBER_SAMPLES_PER_LOW_SAMPLE_BLOCK
+        if self.compressed.raw_value == 1:
+            ints: list[int] = idex_rice_decode(
+                waveform_raw, nbit10=False, sample_count=samples
+            )
+        else:
+            ints = []
+            for i in range(0, len(waveform_raw), 32):
+                ints += [
```
tmplummer: Adding a comment like you did on line 413 would be helpful for documenting how the integers are encoded into the 32 bits in this case.

tmplummer (commented):
This is great! I don't have much feedback. I agree with Greg on trying to add some additional tests of the lower level functions.

imap_processing/idex/idex_l1a.py
lacoak21: Im not exactly sure why but the IDEX team trims the extra ints because they are considered "bad" I can find out why. It is stated in the algorithm document that the high sample waveforms should be of length 1989 which is the length of idex_rice_decode() after it has been trimmed.

imap_processing/idex/idex_l1a.py
lacoak21: There doesn't seem to be a name for this in the algorithm document - its just mentioned that it has to be read in this specific pattern. I like the idea of extracting the logic into a helper function!

imap_processing/idex/idex_rice_decode.py
lacoak21: It should be reading one bit every time while the bit pointer is the one that is increasing. That is a great idea to use bits[bp:].find("1")! Seems much faster/simpler.

imap_processing/idex/idex_rice_decode.py
lacoak21: I think it would be q % 2 == 1?

imap_processing/idex/idex_rice_decode.py
lacoak21: Good idea

imap_processing/tests/idex/test_idex_l1a.py
lacoak21: Sure I can write tests for those! Maybe the original code (written for SUDA) has tests somewhere also...

lacoak21: Update: couldn't find any tests written for the original code but I wrote ones for each decompression schemes.

imap_processing/idex/decode.py
```diff
@@ -0,0 +1,241 @@
+"""
```
greglucas, line 1: This is fantastic documentation! Thank you for the details, can you add it to the docs pages? It doesn't look like it is showing up there yet.
https://imap-processing--1335.org.readthedocs.build/en/1335/code-documentation/idex.html

imap_processing/idex/decode.py
```diff
@@ -0,0 +1,241 @@
...
+        else:
+            # The rice parameter (k) is used to divide each re-mapped residual value
+            # into two parts. The least significant k bits of the value are called
+            # the remainder (r). The other part of the value (not included in the
+            # remainder) is called the quotient (q).
+            # The remapped quotient is unary encoded by including a number of 0 bits
+            # equal to the value of the quotient, followed by a single '1' bit.
+            ind = bits[bp:].find("1")
+            bp += ind + 1
+            q = ind
```
greglucas:
```suggestion
            q = bits[bp:].find("1")
            bp += q + 1
```

imap_processing/idex/idex_l1a.py
```diff
@@ -391,25 +445,21 @@ def _parse_high_sample_waveform(self, waveform_raw: str) -> list[int]:
...
-                int(waveform_raw[i + 12 : i + 22], 2),
-                int(waveform_raw[i + 22 : i + 32], 2),
-            ]
-        return ints[:-4]  # Remove last 4 numbers
+        samples = self.MAX_HIGH_BLOCKS * self.NUMBER_SAMPLES_PER_HIGH_SAMPLE_BLOCK
+        if self.compressed.raw_value == 1:
+            ints: list[int] = rice_decode(
+                waveform_raw, nbit10=True, sample_count=samples
+            )
+            return ints[:-3]
```
greglucas: Out of curiosity, do you know why you need to put in `list[int]` as the type hint here since it looks like that is already defined in rice_decode as what it returns? 🤔 

```suggestion
            return rice_decode(waveform_raw, nbit10=True, sample_count=sample)[:-3]
```

Here you return from in the if-blocks, and down below you set a variable and then return. I'd suggest just being consistent between the two places and go with one version or the other.

greglucas (approved):
Fantastic, these are awesome extra tests and give a great description of what is going on and what you're expecting, thank you for taking the time to add those.

Some minor nit suggestions, feel free to take or leave.

tmplummer (approved):
Really nice added tests!

imap_processing/idex/decode.py
lacoak21: Yes! Should have added that from the start :).

PR #1336, opened 2025-02-04
Download CoDICE L0 test file from AWS

bourque:
# Change Summary

## Overview
This PR adds code to the CoDICE test suite to download the (rather large) test L0 file from AWS via the API instead of storing the file in the repository.

## Deleted Files

- `imap_processing/tests/codice/data/imap_codice_l0_raw_20241110_v001.pkts`
   - Removed test file from repo

## Updated Files

- `imap_processing/tests/codice/conftest.py`
    - Added code to retrieve and store test L0 file from the AWS S3 bucket

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
           if [ "$RUN_EXTERNAL_KERNEL_MARK" = "true" ] || [ "$RUN_DOWNLOAD_TEST_DATA" = "true" ]; then
             poetry run pytest --color=yes --cov --cov-report=xml
           else
-            poetry run pytest --color=yes --cov --cov-report=xml -m "not external_kernel"
+            poetry run pytest --color=yes --cov --cov-report=xml -m "not external_kernel and not external_test_data"
           fi
 
       - name: Upload coverage reports to Codecov
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,11 +42,10 @@ jobs:
       - name: Testing
         id: test
         env:
-          RUN_EXTERNAL_KERNEL_MARK: ${{ contains(matrix.os, 'ubuntu') && matrix.python-version == '3.9' }}
-          RUN_DOWNLOAD_TEST_DATA: ${{ contains(matrix.os, 'ubuntu') && matrix.python-version == '3.9' }}
+          USE_EXTERNAL_DATA: ${{ contains(matrix.os, 'ubuntu') && matrix.python-version == '3.9' }}
         run: |
           # Ignore the network marks from the remote test environment
-          if [ "$RUN_EXTERNAL_KERNEL_MARK" = "true" ] || [ "$RUN_DOWNLOAD_TEST_DATA" = "true" ]; then
+          if [ "$USE_EXTERNAL_DATA" = "true" ]; then
             poetry run pytest --color=yes --cov --cov-report=xml
           else
             poetry run pytest --color=yes --cov --cov-report=xml -m "not external_kernel and not external_test_data"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,9 +43,10 @@ jobs:
         id: test
         env:
           RUN_EXTERNAL_KERNEL_MARK: ${{ contains(matrix.os, 'ubuntu') && matrix.python-version == '3.9' }}
+          RUN_DOWNLOAD_TEST_DATA: ${{ contains(matrix.os, 'ubuntu') && matrix.python-version == '3.9' }}
         run: |
           # Ignore the network marks from the remote test environment
-          if [ "$RUN_EXTERNAL_KERNEL_MARK" = "true" ]; then
+          if [ "$RUN_EXTERNAL_KERNEL_MARK" = "true" ] || [ "$RUN_DOWNLOAD_TEST_DATA" = "true" ]; then
             poetry run pytest --color=yes --cov --cov-report=xml
           else
             poetry run pytest --color=yes --cov --cov-report=xml -m "not external_kernel"
--- a/imap_processing/tests/codice/conftest.py
+++ b/imap_processing/tests/codice/conftest.py
@@ -1,4 +1,4 @@
-import requests
+import pytest
 
 from imap_processing import imap_module_directory
 
@@ -44,13 +44,21 @@ VALIDATION_DATA = [
     TEST_DATA_PATH / "validation" / "imap_codice_l1a_hi-pha_20241110193700_v0.0.0.cdf",
 ]  # fmt: skip
 
-# Download CoDICE L0 test data and write it to the appropriate directory
-response = requests.get(
-    "https://api.dev.imap-mission.com/download/test_data/imap_codice_l0_raw_20241110_v001.pkts"
-)
-if response.status_code == 200:
-    with open(TEST_L0_FILE, "wb") as file:
-        file.write(response.content)
-    print(f"Downloaded file: {TEST_L0_FILE}")
-else:
-    print(f"Failed to download file: {response.status_code}")
+# # Download CoDICE L0 test data if necessary and write it to the appropriate directory
+# if not TEST_L0_FILE.exists():
+#     response = requests.get(
+#         "https://api.dev.imap-mission.com/download/test_data/imap_codice_l0_raw_20241110_v001.pkts"
+#     )
+#     if response.status_code == 200:
+#         with open(TEST_L0_FILE, "wb") as file:
+#             file.write(response.content)
+#         print(f"Downloaded file: {TEST_L0_FILE}")
+#     else:
+#         print(f"Failed to download file: {response.status_code}")
+
+
+@pytest.fixture(scope="session")
+def download_codice_test_data(_download_test_data):
+    """"""
+
+    return "foo"
--- a/imap_processing/tests/codice/conftest.py
+++ b/imap_processing/tests/codice/conftest.py
@@ -43,9 +43,3 @@ VALIDATION_DATA = [
     TEST_DATA_PATH / "validation" / "imap_codice_l1a_lo-pha_20241110193700_v0.0.0.cdf",
     TEST_DATA_PATH / "validation" / "imap_codice_l1a_hi-pha_20241110193700_v0.0.0.cdf",
 ]  # fmt: skip
-
-
-@pytest.fixture(scope="session")
-def ensure_test_data_download(_download_test_data):
-    """"""
-    pass
--- a/imap_processing/tests/codice/conftest.py
+++ b/imap_processing/tests/codice/conftest.py
@@ -1,5 +1,3 @@
-import pytest
-
 from imap_processing import imap_module_directory
 
 TEST_DATA_PATH = imap_module_directory / "tests" / "codice" / "data"
--- a/imap_processing/tests/codice/conftest.py
+++ b/imap_processing/tests/codice/conftest.py
@@ -1,3 +1,5 @@
+import requests
+
 from imap_processing import imap_module_directory
 
 TEST_DATA_PATH = imap_module_directory / "tests" / "codice" / "data"
@@ -41,3 +43,14 @@ VALIDATION_DATA = [
     TEST_DATA_PATH / "validation" / "imap_codice_l1a_lo-pha_20241110193700_v0.0.0.cdf",
     TEST_DATA_PATH / "validation" / "imap_codice_l1a_hi-pha_20241110193700_v0.0.0.cdf",
 ]  # fmt: skip
+
+# Download CoDICE L0 test data and write it to the appropriate directory
+response = requests.get(
+    "https://api.dev.imap-mission.com/download/test_data/imap_codice_l0_raw_20241110_v001.pkts"
+)
+if response.status_code == 200:
+    with open(TEST_L0_FILE, "wb") as file:
+        file.write(response.content)
+    print(f"Downloaded file: {TEST_L0_FILE}")
+else:
+    print(f"Failed to download file: {response.status_code}")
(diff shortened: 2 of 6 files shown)


greglucas (changes requested):
You should check if the TEST_L0_FILE already exists, and only download it if it doesn't. If the download fails, do we need to pytest.skip() any tests? We should try and mark it as a network test too. Otherwise it is going to slam our downloads on every CI run. (See this for examples: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/ca73c8c86abbf47945997089251e4961199f220d/imap_processing/tests/conftest.py#L49-L117 )

As an FYI, I was reading through that L0 file in space packet parser tests and it contains many apids that aren't relevant to us, so could potentially be trimmed as well?

bourque:
@greglucas 

> You should check if the TEST_L0_FILE already exists, and only download it if it doesn't.

Good suggestion, I will add that!

> If the download fails, do we need to pytest.skip() any tests? We should try and mark it as a network test too. Otherwise it is going to slam our downloads on every CI run.

I am confused by this. If the download fails, wouldn't we want the affected tests to fail too? So that we can fix the issue and ensure the tests pass, instead of skipping them and having some issue fall through the cracks? What do you mean by "slam our downloads on every CI run"? I thought that every CI run is a build-from-scratch situation and thus the file would need to be downloaded on every CI session.

tmplummer:
>We should try and mark it as a network test too. Otherwise it is going to slam our downloads on every CI run.

I think that what @greglucas means is that you can put a `pytest.mark` on any test that downloads data and have it run only on a single CI test configuration. The SPICE stuff requiring external kernels does this by the following two entries:
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/ca73c8c86abbf47945997089251e4961199f220d/imap_processing/tests/spice/test_geometry.py#L48

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/ca73c8c86abbf47945997089251e4961199f220d/.github/workflows/test.yml#L44

bourque:
If I just need to download one file, once per PR build, to put in a directory to be used by some tests, would it suffice to just have a few lines of code in a `codice.conftests.py` fixture with `session=scope` to do that? My understanding is that the code within `conftest.py` gets evoked once in that case. I am not seeing why I would need to put in safeguards to prevent multiple downloads beyond that.

greglucas:
Yeah, we have a matrix of runs and we don't need to run all tests on all runners necessarily. Having just one do network I/O seems sufficient. Downloading a 45MB file repeatedly can add to our data egress fees in AWS, so limiting that is prefereable IMO. We have run 165 jobs through PR actions in the past 3 days, so that is ~2 GB/day just for that single file.

bourque:
Oh I think I am understanding now. A "network test" has shared resources across all matrixed CI configurations? Sorry that was not clear to me.

greglucas:
> Oh I think I am understanding now. A "network test" has shared resources across all matrixed CI configurations? Sorry that was not clear to me.

A "network test" would be skipped on 8/9 of the runners and only run on a single runner. So if you do want to test across all versions/operating systems, then a network test would not be the right thing here. It is definitely a trade-off in terms of what you want to test / how robust you want to be.

bourque:
> > Oh I think I am understanding now. A "network test" has shared resources across all matrixed CI configurations? Sorry that was not clear to me.
> 
> A "network test" would be skipped on 8/9 of the runners and only run on a single runner. So if you do want to test across all versions/operating systems, then a network test would not be the right thing here. It is definitely a trade-off in terms of what you want to test / how robust you want to be.

I think I got it. I am trading in being able to only download a large file once with the assumption that if my CoDICE L1a tests pass on say, ubuntu with Python 3.9, they will pass on Windows & Mac, and different python versions.

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -79,6 +79,7 @@
 ]
 
 
+@pytest.mark.download_test_data()
```
greglucas: I think you might be able to put this up at the module level if every test needs this data:
https://docs.pytest.org/en/stable/example/markers.html

```
pytestmark = pytest.mark.external_test_data
```

imap_processing/tests/conftest.py
```diff
@@ -102,6 +139,7 @@ def pytest_collection_modifyitems(items):
     +=====================+============================+
     | external_kernel     | _download_external_kernels |
     | use_test_metakernel | use_test_metakernel        |
+    | download_test_data  | _download_test_data        |
```
greglucas: The mark usually tells you what is going on like "requires X", "external data", then the function would do what you need to meet that requirement/expectation. For consistency, I'd suggest going with `external_test_data` indicating this test is marked as needing external test data.
```suggestion
    | external_test_data  | _download_test_data        |
```

imap_processing/tests/codice/conftest.py
```diff
@@ -41,3 +43,22 @@
...
+# if not TEST_L0_FILE.exists():
+#     response = requests.get(
+#         "https://api.dev.imap-mission.com/download/test_data/imap_codice_l0_raw_20241110_v001.pkts"
+#     )
+#     if response.status_code == 200:
+#         with open(TEST_L0_FILE, "wb") as file:
+#             file.write(response.content)
+#         print(f"Downloaded file: {TEST_L0_FILE}")
+#     else:
+#         print(f"Failed to download file: {response.status_code}")
```
greglucas: Remove the updates to this file now that it is all in the other download files?
```suggestion
```

imap_processing/tests/codice/conftest.py
```diff
@@ -41,3 +43,22 @@
...
+
+
+@pytest.fixture(scope="session")
+def download_codice_test_data(_download_test_data):
+    """"""
+
+    return "foo"
```
greglucas: What is this version doing? Is the download not happening with the marks in your other tests?

greglucas (commented):
:+1: I think this is the right idea. It would be good for @subagonsouth to comment on similarities / suggestions for what he previously did with the external kernel work and try to keep everything consistent between the approaches. Do we even want to have just a single mark for both SPICE kernels and external test data? (I don't have a strong preference either way)

imap_processing/tests/codice/conftest.py
bourque: This is just leftover as I am trying to get things working here. I will clean this up

bourque:
I think i've got most of the pieces in place for this now(?). It is not clear to me why the Ubuntu/Python3.9 build is not invoking the `_download_test_data()` fixture. It appears to be running as expected when I run the tests locally, and I see that the environment is set properly for that build:

```
Run # Ignore the network marks from the remote test environment
  # Ignore the network marks from the remote test environment
  if [ "$RUN_EXTERNAL_KERNEL_MARK" = "true" ] || [ "$RUN_DOWNLOAD_TEST_DATA" = "true" ]; then
    poetry run pytest --color=yes --cov --cov-report=xml
  else
    poetry run pytest --color=yes --cov --cov-report=xml -m "not external_kernel and not external_test_data"
  fi
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.21/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.9.21/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.21/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.21/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.21/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.21/x64/lib
    RUN_EXTERNAL_KERNEL_MARK: true
    RUN_DOWNLOAD_TEST_DATA: true
```

@greglucas @subagonsouth Does anything jump out to you here as being set up incorrectly?

greglucas:
> @greglucas @subagonsouth Does anything jump out to you here as being set up incorrectly?

Nothing immediate jumps out to me, it looks like you've set it up the same as the other external kernel download function to me at a quick glance. My only other potential thought would be that it is being downloaded to a different path than where you're accessing it later... It is looking for this path: `/home/runner/work/imap_processing/imap_processing/imap_processing/tests/codice/data/imap_codice_l0_raw_20241110_v001.pkts` which has 3 `imap_processing` nests in it, which I'm not sure if that is right or not 🤷

.github/workflows/test.yml
```diff
@@ -43,12 +43,13 @@ jobs:
         id: test
         env:
           RUN_EXTERNAL_KERNEL_MARK: ${{ contains(matrix.os, 'ubuntu') && matrix.python-version == '3.9' }}
+          RUN_DOWNLOAD_TEST_DATA: ${{ contains(matrix.os, 'ubuntu') && matrix.python-version == '3.9' }}
```
greglucas: Just do one environment variable here and remove the or below.
```suggestion
          USE_EXTERNAL_DATA: ${{ contains(matrix.os, 'ubuntu') && matrix.python-version == '3.9' }}
```

bourque:
> It is looking for this path: /home/runner/work/imap_processing/imap_processing/imap_processing/tests/codice/data/imap_codice_l0_raw_20241110_v001.pkts which has 3 imap_processing nests in it, which I'm not sure if that is right or not 🤷

I think this is right? I am defining everything relative to the `imap_module_directory` variable (just as Tim does for the spice kernel path definitions), and i double checked that I haven't hard-coded another nested `imap_processing` subdirectory somewhere. 🤔

bourque:
I got this working by using the `autouse=True` parameter in the fixture definition:

```python
@pytest.fixture(scope="session", autouse=True)
def _download_test_data(test_data_paths):
```

But I am not sure this is what we want. I think this would mean that if someone is running their (non-CoDICE) tests locally, they would also be downloading the CoDICE test file, which is not necessary for their tests.

greglucas:
@bourque, I just pushed a new commit up here to try and address the issue. This was a race condition between the fixtures. Your decom fixture was getting inserted into the list of fixtures before the download_data mark which was being appended to the list. So your tests were trying to call the load data before it would be downloaded. I added the _download_data as a dependency of the decom fixture now which causes those to be registered in order. We still have to make sure the mark is applied to all the tests using that fixture though, which is done through your module level pytestmark.

greglucas (approved):

imap_processing/tests/conftest.py
```diff
@@ -110,11 +150,16 @@ def pytest_collection_modifyitems(items):
...
+        "external_test_data": "_download_test_data",
+        "use_test_metakernel": "use_test_metakernel",
+    }
+
     for item in items:
-        if item.get_closest_marker("external_kernel") is not None:
-            item.fixturenames.append("_download_external_kernels")
-        if item.get_closest_marker("use_test_metakernel") is not None:
-            item.fixturenames.append("use_test_metakernel")
+        for marker, fixture in markers_to_fixtures.items():
```
tmplummer, line 166: Nice! I like this refactor.

imap_processing/tests/conftest.py
```diff
@@ -91,6 +91,45 @@ def _download_external_kernels(spice_test_data_path):
...
+                logger.info(f"Downloaded file: {source}")
+            else:
+                logger.error(f"Failed to download file: {response.status_code}")
+        else:
+            logger.info(f"File already exists: {destination}")
+
+
+@pytest.fixture(scope="session")
+def test_data_paths():
+    test_data_path_list = [
```
tmplummer, line 126: Do we want to have to add each file to this list? I don't know if we could just list the test data in s3 and mirror the bucket structure when downloaded. For example this file would go in `test_data/codice/data/imap_codice_l0_raw_20241110_v001.pkts`. I have no clue if the API would allow something like that. Another thought is to just put all of the test data in a single test_data directory when downloading?

Related... I don't see any addition to `.gitignore` for this test data file that will be downloaded into the repo. As instruments move data out of the repo and into s3 should they be adding a line to ignore everything in their test data directory?

tmplummer (commented):
This looks like a great solution to me. I just have a couple of questions that I left in one code comment to try to spur a little discussion about two details.

imap_processing/tests/conftest.py
greglucas: > Do we want to have to add each file to this list? I don't know if we could just list the test data in s3 and mirror the bucket structure when downloaded. For example this file would go in test_data/codice/data/imap_codice_l0_raw_20241110_v001.pkts. I have no clue if the API would allow something like that. Another thought is to just put all of the test data in a single test_data directory when downloading?

I like this idea :+1: My goal has been to have things be syncable so that you can either run things in AWS, on a local system, in GitHub and not have to think about things. Making sure the structure is consistent would be great and should be doable based on putting everything relative to IMAP_DATA_DIR (and swapping this for various test locations): https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/4c170c161ca0e880e7e66c2e82ca2fac3b897d7b/imap_data_access/__init__.py#L32
Our download capability likely needs updating to make that more usable for tests, so I think currently we would have to iterate over explicit paths. (Also note that I think SPICE should do the same thing here, so making these all equivalent and not special-casing the various versions)

> Related... I don't see any addition to .gitignore for this test data file that will be downloaded into the repo. As instruments move data out of the repo and into s3 should they be adding a line to ignore everything in their test data directory?

We have a general `data/` directory ignored, which I think gets any subpaths.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/5f72c4bfea95f3641118f8e034b2d0afc86c45ab/.gitignore#L178-L179

bourque: Since we want to fold SPICE into this and add the capability to dynamically generate the list of test files from the S3 bucket, can we punt that work to a future PR? I would like to get back to CoDICE development so I am hoping to wrap up this PR here.

Also thank you @greglucas for adding that fix for the race condition!

greglucas: Yes, definitely! I think that should be a follow-up and keep this isolated and contained for now since we don't have a clear path for that and it will be a little more work to get right.

PR #1339, opened 2025-02-04
Create Lo L1A Spin CDF

### LINKED ISSUE #1338: Create Lo L1A Spin CDF

sdhoyt:
# Change Summary

## Overview
Created and validated Lo L1A Spin CDF 

## Updated Files
- imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
   - renamed fields to match new telemetry definition
   - using dtype from yaml to set data type for dataset
- imap_processing/lo/l1a/lo_l1a.py
   - added CDF coordinates and attires

## Testing
Updated unit tests and added validation for spin data. Matching Lo provided spin validation data.

Closes #1338


imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -436,62 +443,87 @@ num_completed:
...
 
-acq_end:
+acq_start_sec:
+  <<: *default_uint32
+  DEPEND_0: epoch
+  VAR_TYPE: support_data
+  CATDESC: Acquisition Start
+  FIELDNAM: Acquisition Start Seconds
+  UNITS: ns
+  LABLAXIS: acquisition start seconds
```
greglucas: Inconsistent capitalization here. I guess lowercase "s" would probably be preferred? Also in other fields below this too.

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -436,62 +443,87 @@ num_completed:
...
+
+acq_start_subsec:
+  <<: *default_uint32
+  DEPEND_0: epoch
+  VAR_TYPE: support_data
+  CATDESC: Acquisition Start
+  FIELDNAM: Acquisition Start Sub-Seconds
+  LABLAXIS: acquisition start sub-seconds
```
greglucas: Do we need a UNITS: field to indicate that these are in `ms` (or whatever unit they are).

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -436,62 +443,87 @@ num_completed:
...
   LABLAXIS: number of spins completed
 
-acq_end:
+acq_start_sec:
+  <<: *default_uint32
+  DEPEND_0: epoch
+  VAR_TYPE: support_data
+  CATDESC: Acquisition Start
+  FIELDNAM: Acquisition Start Seconds
+  UNITS: ns
```
greglucas: Is this actually ns, or is it (s)? The variable is named "seconds"

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -436,62 +443,87 @@ num_completed:
...
 
-spin_seconds:
+start_sec_spin:
   <<: *spin_default
   <<: *default_uint32
   CATDESC: Spin Start Seconds Time
   FIELDNAM: Spin Start Seconds Time
   UNITS: ns
-  LABLAXIS: spin start seconds time
+  #LABLAXIS: spin start seconds time
```
greglucas: Remove these if they aren't needed?

imap_processing/lo/l0/lo_science.py
```diff
@@ -470,14 +473,27 @@ def organize_spin_data(dataset: xr.Dataset) -> xr.Dataset:
         "period_source_spin",
     ]
 
+    # Set epoch to the acq_start time
+    acq_start = dataset.acq_start_sec.values + (0.1 * dataset.acq_start_subsec.values)
+    epoch = met_to_ttj2000ns(acq_start)
```
greglucas: I've noticed that we do a lot of `.values` / `.data` on the project, does the `met_to_ttj2000ns()` function not take array-like and it has to be numpy specifically? In general, we should be able to use `dataset["acq_start_sec"]` and that will act like an array 🤔

imap_processing/lo/l1a/lo_l1a.py
```diff
@@ -126,8 +130,63 @@ def add_dataset_attrs(
...
+    #  own functions
+    # Get global attributes
+    dataset.attrs.update(attr_mgr.get_global_attributes(logical_source))
+    # Get attributes for shcoarse and epoch
+    dataset.shcoarse.attrs.update(attr_mgr.get_variable_attributes("shcoarse"))
+    dataset.epoch.attrs.update(attr_mgr.get_variable_attributes("epoch"))
+
+    if logical_source == "imap_lo_l1a_spin":
+        spin = xr.DataArray(
+            data=np.arange(0, 28, dtype=np.uint16),
```
greglucas: Is this supposed to be hard-coded? If so, you can get away with a uint8 here too.

imap_processing/lo/l1a/lo_l1a.py
```diff
@@ -126,8 +130,63 @@ def add_dataset_attrs(
...
+                "seq_flgs",
+                "src_seq_ctr",
+                "pkt_len",
+                "chksum",
+            ]
+        )
+        # An empty DEPEND_0 is being added to support_data
+        # variables that should only have DEPEND_1
+        # Removing Depend_0 here.
+        # Should look for a fix to this issue
```
greglucas: Add a NOTE or TODO for future searching.
```suggestion
        # TODO: Should look for a fix to this issue
```

imap_processing/lo/l1a/lo_l1a.py
```diff
@@ -231,5 +291,20 @@ def add_dataset_attrs(
...
+        del dataset["coincidence_type"].attrs["DEPEND_0"]
+        del dataset["de_time"].attrs["DEPEND_0"]
+        del dataset["mode"].attrs["DEPEND_0"]
+        del dataset["esa_step"].attrs["DEPEND_0"]
+        del dataset["tof0"].attrs["DEPEND_0"]
+        del dataset["tof1"].attrs["DEPEND_0"]
+        del dataset["tof2"].attrs["DEPEND_0"]
+        del dataset["tof3"].attrs["DEPEND_0"]
+        del dataset["pos"].attrs["DEPEND_0"]
+        del dataset["cksm"].attrs["DEPEND_0"]
```
greglucas: loop over the options
```suggestion
        for var in ["direct_events", ...]:
            dataset[var].attrs.pop("DEPEND_0")
```

imap_processing/tests/lo/test_lo_l1a.py
```diff
@@ -54,4 +59,56 @@ def test_lo_l1a_dataset():
...
+    for field in spin_fields:
+        # The validation data contains a duplicate set of columns for the same values.
+        # They are formatted as column_prefix_<spin> and column_prefix[<spin>]
+        # adding a condition to remove the one with [ when combining those columns into
+        # a list to combine the spins in the validation data into a single list
+        validation_fields = [
+            col
+            for col in validation_data.columns
+            if col.startswith(field.upper()) and "[" not in col
+        ]
```
greglucas: Can you move this up above the loop and drop the "bad" fields and then compare directly against the two datasets rather than looping/comparing with the branches?
```suggestion
        bad_fields = [
            col
            for col in validation_data.columns
            if "[" in col
        ]
        validation_data = validation_data.drop(bad_fields)
```

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
sdhoyt: Yeah I'll need to double check on what this should be

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
sdhoyt: I have ns on the page that Colin and I worked on together, but yeah the naming seems weird. I'll double check.

imap_processing/tests/lo/test_lo_l1a.py
sdhoyt: I would still need to combine all the spin columns (start_sec_spin_1, start_sec_spin_2, ...) into an array (start_sec_spin) which is partly what the validation field list is helping me do. I will try to think if there's an easier way to do that after removing the bad fields though

imap_processing/lo/l0/lo_science.py
sdhoyt: Yes, that does work. Not sure why I thought I had to do `.values`

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
sdhoyt: Discussed this with Lo today. The acq start seconds is in units of seconds and the acq start sub seconds is in units of microseconds. I was thinking about this more after our meeting and I will need to follow-up with them about where they want the EU conversion done because there is no L1B spin product.

imap_processing/lo/l0/lo_science.py
greglucas: I think we are doing this all over the place, so it is a pretty common pattern.

imap_processing/tests/lo/test_lo_l1a.py
greglucas: No worries, this is just me being too picky. This works, and I shouldn't get hung up on these little things in tests. Carry on!

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
greglucas: FYI that this other PR I was reviewing is what made me think of some of this too.
https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1346

There it looks like Tim is doing it in L1a right away. I would consider EU slightly different than a straight unit conversion with multiplying/dividing by a factor to get something from nano to micro or something like that 🤷 (EU is more like a calibration factor applied in my mind)

frahmanif: > Discussed this with Lo today. The acq start seconds is in units of seconds and the acq start sub seconds is in units of microseconds. I was thinking about this more after our meeting and I will need to follow-up with them about where they want the EU conversion done because there is no L1B spin product.

I emailed Chuck about this, and will let you know what he says as soon as I get a response.

sdhoyt: @greglucas Yeah, sorry my comment about the EU doesn't make a ton of sense with what I said here. There are EU conversions involving voltages in the spin packet which I was also thinking about but didn't mention in my comment. So I need to figure out if/where the voltage EU conversions should be put into a dataset.

@frahmanif Thanks for checking on this!

imap_processing/tests/lo/test_lo_science.py
```diff
@@ -302,16 +301,16 @@ def test_validate_parse_events(sample_data, attr_mgr):
...
     # Arrange
     data_by_epoch_spin = np.arange(0, 56).reshape(2, 28)
     expected_dataset = xr.Dataset(
         data_vars=dict(
             num_completed=(["epoch"], np.array([0, 0])),
-            acq_start_sec=(["epoch"], np.array([0, 0])),
-            acq_start_subsec=(["epoch"], np.array([0, 0])),
-            acq_end_sec=(["epoch"], np.array([0, 0])),
-            acq_end_subsec=(["epoch"], np.array([0, 0])),
+            acq_start_sec=(["epoch"], np.array([1000000, 2000000])),
```
greglucas, line 310: Thanks for updating these, this is easier for me to follow now!

greglucas (approved):

PR #1344, opened 2025-02-05
SWE: Adding initial L2 data variables and CDF attrs

### LINKED ISSUE #915: Create SWE L2 CDF attrs and file

tech3371:
# Change Summary

## Overview
Initial setup for L2 work and CDF

## Updated Files
- imap_processing/cdf/config/imap_swe_global_cdf_attrs.yaml and imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
   - L2 CDF attrs setup
   - Add initial L2 data variables
   - L2 nomenclature fix

## Testing
   - write to CDF

### DIFF
--- a/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
@@ -50,7 +50,7 @@ spin_sector:
 
 cycle:
   CATDESC: Full cycle data takes 4 spins' data
-  FIELDNAM: quater cycle
+  FIELDNAM: quarter cycle
   FILLVAL: -9223370000000000000
   FORMAT: I2
   LABLAXIS: Cycle
@@ -137,9 +137,9 @@ sci_step_acq_time_sec:
   LABL_PTR_1: esa_step_label
   LABL_PTR_2: spin_sector_label
   DISPLAY_TYPE: spectrogram
-  FIELDNAM: Acquisition by volt step and spin sector
+  FIELDNAM: Acquisition time by volt step and spin sector
   FILLVAL: -1.0000000E+31
   UNITS: sec
   VAR_TYPE: support_data
   VAR_NOTES: >
-    This stores acquisition times of each count data.
+    This stores the acquisition times of each measurement.
--- a/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
@@ -129,7 +129,7 @@ flux_spin_sector:
   VALIDMIN: 0
   VAR_TYPE: data
 
-sci_step_acq_time_sec:
+acquisition_time:
   CATDESC: Acquisition time organized by voltage step and spin sector
   DEPEND_0: epoch
   DEPEND_1: esa_step
@@ -142,4 +142,7 @@ sci_step_acq_time_sec:
   UNITS: sec
   VAR_TYPE: support_data
   VAR_NOTES: >
-    This stores the acquisition times of each measurement.
+    This stores the acquisition times of each measurement. This time is
+    calculated by combining ACQ_START_COARSE, ACQ_START_FINE,
+    acquisition duration and settle duration. It is time of each energy
+    step and each science measurement.
--- a/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
@@ -1,0 +1,145 @@
+# <=== Coordinates ===>
+esa_step:
+  CATDESC: Energy step id in lookup table
+  FIELDNAM: energy step
+  FILLVAL: -9223370000000000000
+  FORMAT: I2
+  LABLAXIS: Energy Step
+  SCALE_TYP: linear
+  UNITS: " "
+  VALIDMAX: 23
+  VALIDMIN: 0
+  VAR_TYPE: support_data
+
+spin_angle:
+  CATDESC: Spin angle. Angle resolution is 12 degree nominally
+  FIELDNAM: spin angle
+  FILLVAL: -9223370000000000000
+  FORMAT: I2
+  LABLAXIS: Angle
+  SCALE_TYP: linear
+  UNITS: "Degree"
+  VALIDMAX: 29
+  VALIDMIN: 0
+  VAR_TYPE: support_data
+
+cem_id:
+  CATDESC: Data of each CEM detectors
+  FIELDNAM: CEM data
+  FILLVAL: -9223370000000000000
+  FORMAT: E14.7
+  LABLAXIS: Rates
+  SCALE_TYP: linear
+  UNITS: " "
+  VALIDMAX: 0.000015514
+  VALIDMIN: 0.0
+  VAR_NOTES: " "
+  VAR_TYPE: support_data
+
+spin_sector:
+  CATDESC: Spin sector. 30 measurements nominally
+  FIELDNAM: Spin Sector
+  FILLVAL: -9223370000000000000
+  FORMAT: I2
+  LABLAXIS: Spin Sector
+  SCALE_TYP: linear
+  UNITS: " "
+  VALIDMAX: 29
+  VALIDMIN: 0
+  VAR_TYPE: support_data
+
+cycle:
+  CATDESC: Full cycle data takes 4 spins' data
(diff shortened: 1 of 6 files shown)


imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,145 @@
...
+  LABLAXIS: ' '
+  FILLVAL: -9223372036854775808
+  FORMAT: I19
+  UNITS: ' '
+  VALIDMIN: 0
+  VALIDMAX: 9223372036854769664
+  VAR_TYPE: support_data
+  SCALETYP: linear
+
+phase_space_density_spin_sector:
```
tech3371, line 98: I appreciate some better name suggestions. Ruth and I both struggle with good names.

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,145 @@
...
+  LABLAXIS: Spin Sector
+  SCALE_TYP: linear
+  UNITS: " "
+  VALIDMAX: 29
+  VALIDMIN: 0
+  VAR_TYPE: support_data
+
+cycle:
+  CATDESC: Full cycle data takes 4 spins' data
+  FIELDNAM: quater cycle
```
greglucas: Do you want to call this variable quarter_cycle_id or something similar to that to be in line with cem_id? Or is cycle descriptive enough?
```suggestion
  FIELDNAM: quarter cycle
```

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
greglucas: Just `phase_space_density` or `psd`? The "per spin_sector" can potentially be inferred by the depends I think.

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,145 @@
...
+  DEPEND_2: spin_sector
+  LABL_PTR_1: esa_step_label
+  LABL_PTR_2: spin_sector_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Acquisition by volt step and spin sector
+  FILLVAL: -1.0000000E+31
+  UNITS: sec
+  VAR_TYPE: support_data
+  VAR_NOTES: >
+    This stores acquisition times of each count data.
```
greglucas:
```suggestion
    This stores the acquisition time of each measurement.
```

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+sci_step_acq_time_sec:
+  CATDESC: Acquisition time organized by voltage step and spin sector
+  DEPEND_0: epoch
+  DEPEND_1: esa_step
+  DEPEND_2: spin_sector
+  LABL_PTR_1: esa_step_label
+  LABL_PTR_2: spin_sector_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Acquisition by volt step and spin sector
```
greglucas:
```suggestion
  FIELDNAM: Acquisition time by volt step and spin sector
```

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,145 @@
...
+  LABL_PTR_3: cem_id_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Flux
+  FILLVAL: -1.0000000E+31
+  UNITS: " "
+  VALIDMAX: 0.000015514
+  VALIDMIN: 0
+  VAR_TYPE: data
+
+sci_step_acq_time_sec:
```
greglucas: What does sci_step mean? Should this just be `acq_time`, or that should be before the sci_step qualifier?

imap_processing/swe/l2/swe_l2.py
```diff
@@ -125,8 +126,14 @@ def calculate_phase_space_density(l1b_dataset: xr.Dataset) -> xr.Dataset:
     # energy in eV value that flux calculation can use.
     phase_space_density_dataset = xr.Dataset(
         {
-            "phase_space_density": (["epoch", "energy", "angle", "cem"], density.data),
-            "energy_in_eV": (["epoch", "energy", "angle"], particle_energy_data),
+            "phase_space_density": (
+                ["epoch", "esa_step", "spin_sector", "cem_id"],
+                density.data,
+            ),
+            "energy_in_eV": (
```
greglucas, line 133: Do you want "in_eV" in the variable name, or is that stored in the units / metadata instead?

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
tech3371: Ruth does want phase_space_density calculated twice. One left in spin sector and the other binned in spin angle. That's why I added that `spin_sector` to the name. What should I name those two to distinguish them?
Screenshot 2025-02-05 at 12 40 39 PM

imap_processing/swe/l2/swe_l2.py
tech3371: This is an intermediate dataset. This doesn't get stored to CDF file. It's used in flux calculation.

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
tech3371: We did start as `acq_time`. That's simple and people seems to know what it means. Cool. I will change it to be `acquisition_time`

tech3371: I remember now, it was because packet variable also has these two
Screenshot 2025-02-05 at 12 49 31 PM

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
greglucas: What is the difference between spin_sector and spin_angle. Is one a mapping to the other? i.e. are they the same thing just one is an angle value and the other is the index. If so, I'd say that  the spin angle could be a variable based upon spin sector. `spin_angle(spin_sector)`

If they are two completely separate quantities, then I think what you have is good.

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
greglucas: I'm not sure I totally follow all of the variables above. Is epoch different between those things? There is a "packet epoch", but there is also a "data epoch" which may require combining packets.

tech3371: `epoch` is time calculated from `SHCOARSE`. `sci_step_acquisition_time` is time calculated by combining `ACQ_START_COARSE`, `ACQ_START_FINE`, acquisition duration and settle duration. It is time of each energy step and each science's data. That's why I think Ruth suggested that name.

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
tech3371: I am meeting with Ruth tomorrow. I will run by her.

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
greglucas: :+1: I think that is what I was thinking above and you said it much better :) I guess I am still in favor of just "acquisition_time" if you can do that.

bourque (approved):
This looks good to me!

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
tech3371: They want to keep in `spin_sector` in that first two name and then remove `spin_angle` from the main binned data and keep it as `spin_phase_density` and `flux`.

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
tech3371: They are ok with `acquisition_time`

PR #1345, opened 2025-02-06
CoDICE L1a lo-(n)sw-species and lo-(n)sw-priority validation

bourque:
This PR makes some changes to the `lo-(n)sw-species` and `lo-(n)sw-priority` data products to get them to a validated state.

For now, I am also switching to comparing data array shapes (and not values) in the data validation unit test. This is because Joey and I are still seeing some small discrepancies in the data array values between my CDFs and the validation CDFs (we think this may be due to a decompression lookup table issue, but still needs further investigation). I added a TODO for this as to not forget to properly compare the full data arrays in the future.

See #251


greglucas (approved):

greglucas:
The renaming looks good. Looks like some of the files need to be renamed though as well?

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -402,236 +402,221 @@ lo-sw-priority-p0_tcrs:
     <<: *counters
     CATDESC: Sunward Sector Triple Coincidence Pickup Ions Priority
     FIELDNAM: SW Sector Triple Coincidence PUI's
-    LABL_PTR_1: inst_az
-    LABL_PTR_2: spin_sector
-    LABL_PTR_3: esa_step
+    LABL_PTR_1: spin_sector
+    LABL_PTR_2: esa_step
```
greglucas, line 406: It looks like SWE has these reversed. I think there was preference for the mission to be consistent, do either CoDICE or SWE have a strong preference? @joeymukherjee @tech3371 

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/5f72c4bfea95f3641118f8e034b2d0afc86c45ab/imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml#L89-L90

It looks like CoDICE and SWE might be the only two with these specific values currently.

tech3371: I don't know if SWE and CoDICE need to be same. SWE wanted their data organized by voltage step and spin sector and CEM id. The 2D (or 3D) plot is represented by this, rows are energy steps and columns are angles and each cell contains 7 CEMs data. That's why, my depend(s) were reversed compared to CoDICE.

greglucas: As an end user I would find it easier to use data across instruments if we are consistent with (time, lat, lon) vs (time, lon, lat) vs (lon, lat, time). So I'd vote for one of the instruments to change :)

tech3371: I see. I see your point. Based on my observation, it seems like SWE has put lot of thought organizing their data that way. Also, given how far I am in with SWE, changing SWE will take some effort. I am little concern to change it, tbh.

bourque: I will add TODO here and discuss this with Joey to see if he has strong opinions/requirements. Thanks to #1320 it is quite easy for me to change the order of these things, so I don't think it would be a difficult thing to change from the CoDICE perspective.

Issue #1334, opened 2025-02-04
SPICE - Improve documentation


tmplummer:
### Overview

- Add examples. Reference this ticket for some examples: #897 
- Add time descriptions. Notes from Slack:
```
ET seconds - SPICE requires this for geometry queries - wrap all spice geometry calls with TDT to ET conversion
TDT nanoseconds - CDF epoch variable(s) - Store ET in epoch and do the conversion in write_cdf/load_cdf. This would invalidate the metadata as soon as conversion is done.
MET seconds - Raw data, spacecraft clock, sometimes useful for algorithms, pointing and repointing tables carry MET times - I don't think that we can totally get away from MET. We could convert ET or TDT internal to functions that query the repointing and spin tables.
datetime64 - currently used to convert TDT nanoseconds to date string but should be using spice for this. - I suspect we will always want to use either datetime64 or datetime to generate date strings in filenames.
```

Additionally, make clear that instrument pointing returns [lon, lat] when not returning cartesian coordinates.

PR #1346, opened 2025-02-07
1341 hi l1a de   update to cdf definition

### LINKED ISSUE #1341: Hi L1A DE - Update to CDF definition
### Algorithm Description:

I have decided to split the L1A file into two lists:
One list keeps the packet specific telemetry and the other keeps the events and associated event telemetry. This keeps the L1A closer to the packet structure and makes things a bit cleaner in higher levels.

tmplummer:
# Change Summary

## Overview
This PR got bigger than intended... sorry.
The main driver for these changes is that I identified that a variable needed in Hi L1C PSET processing was missing from the Hi L1B DE data product. The changes to the L1A DE product accomplish two things:
- Make the L1A CDF file closer to the L0 packets; in general, the L1A should be basically the raw CCSDS data.
- Makes it possible to pass the missing variable, CCSDS MET, into the L1B product.
The changes to the L1A product necessitated changes to L1B so those are included here as well.

Changes to the L1A CDF are summarized here:
- Epoch now stores CCSDS creation time
- All data that comes once per packet depends on the Epoch coordinate
- Added new coordinate `event_met`
- All data that comes in the list of DEs in each packet depend on the `event_met` coordinate
- Added `ccsds_index` to associate each DE with the packet it came from.

Changes to L1B CDF mirror the L1A changes:
- Epoch now stores CCSDS creation time (passed through from L1A)
- New coordinate `event_met` (passed through from L1A)

## Updated Files
   - Updates to L1A DE and L1B DE variable metadata
   - Added variables to L1A and L1B DE sections
   - Don't add 1/2 clock tick to DE MET time (I have a question out to Paul about this)
   - Add new CDF coordinate event_met
   - Some code reordering to achieve desired order of variables in the CDF file
- imap_processing/tests/hi/test_science_direct_event.py
   - Updates to coincide with changes to L1A DE CDF
   - Add 1/2 clock tick to event MET before calculating spin phase and HAE coordinates. (This was removed from L1A so it had to be added here)
   - Removed one artifact from before a refactor
   - Updates to match change in CDF variables
   - Renamed `_sct2e_wrapper` to `sct_to_et` to make it public.
   - Renamed `_sct2e_wrapper` to `sct_to_et`.
- imap_processing/tests/hi/test_data/l1/imap_hi_l1a_45sensor-de_20250415_v999.cdf
   - Updated test file to match changed L1A CDF definition.
   - No longer have duplicate Epoch values, so remove the code that was removing them before writing the CDF.

Closes: #1341


imap_processing/hi/l1b/hi_l1b.py
```diff
@@ -283,14 +294,17 @@ def de_nominal_bin_and_spin_phase(dataset: xr.Dataset) -> dict[str, xr.DataArray
...
     )
 
     # nominal_bin is the index number of the 90 4-degree bins that each DE would
     # be binned into in the histogram packet. The Hi histogram data is binned by
     # spacecraft spin-phase, not instrument spin-phase, so the same is done here.
-    met_seconds = dataset.event_met.values.astype(np.float64) / 1e9
+    # We have to add 1/2 clock tick to MET time before getting spin phase
+    met_seconds = (
+        dataset.event_met.values.astype(np.float64) + HALF_CLOCK_TICK_NS
+    ) / 1e9
```
greglucas: FYI, I think this is what I was curious about consolidating. Is it possible to move either to ns or to s, because you are doing math in both right now and it seems quite easy to forget to multiply/divide by 1e9 (maybe divide by the constant here too `/ SECONDS_TO_NS`
xref: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1334

Does your event_met need to be ns, or could that be float s too? I think you mentioned we might lose resolution at one point by doing that...

imap_processing/tests/hi/test_l1a.py
```diff
@@ -19,8 +19,6 @@ def test_sci_de_decom(hi_l0_test_data_path):
 
     # Write to CDF
     cdf_filename = "imap_hi_l1a_90sensor-de_20241105_v001.cdf"
-    # TODO: Dropping duplicates to ignore ISTP for now. Should be fixed by #1186
-    processed_data[0] = processed_data[0].sortby("epoch").groupby("epoch").first()
```
greglucas, line 23: 🎉

greglucas (approved):

imap_processing/hi/l1b/hi_l1b.py
tmplummer: I'm going to discuss this with Paul. The algorithm document specifically calls out storing Event MET in integer nanoseconds. The nice thing about using integer nanoseconds is that they have uniform precision where as floats have higher precision the closer the time is to the Epoch it is representing. 

My reading of 64-bit floats is that they should provide at least 15 to 17 significant digits of precision. A realistic MET value for April, 15, 2025 is `482372987` which has 9 sig figs. If I understand this correctly, the 64-bit float would be expected to provide 6-9 digits of precision in fractional seconds. That seems sufficient to me as long as no instrument clocks are higher than about 1μs resolution. Furthermore, all of the SPICE geometry queries use floating J2000 time as inputs, so even if the 64-bit integer is providing higher precision, that precision is lost before calculating geometry.
https://en.wikipedia.org/wiki/Double-precision_floating-point_format

PR #1351, opened 2025-02-10
HIT L1B - validate standard rates data

### LINKED ISSUE #960: Validate HIT L1B Standard Rates Data
Validate standard rates dataset against values in the validation CSV file

vmartinez-cu:
This PR adds a test to validate the L1B standard rates data. As part of this work, the helper test functions file `l1a_validation.py` was updated to apply to L1B validation as well.

Closes #960 

## Updated Files
- l1a_validation.py
   - Renamed file to l1_validation.py
   - Updated file to support both L1A and L1B validation since some of the validation data reformatting work is common for both.
   - Added constants to the top of the file for readability
   - Added a helper function to reformat/prepare standard rates validation data for comparison with dataset
   - Split out consolidation of sector rate fields in the validation data into it's own function
   - Added type hints
   - Cleaned up documentation
- test_hit_l1b.py
   - added a test to validate standard rates product and use helper functions to carry this out
- test_hit_l1a.py
   - minor updates to reflect changes to the helper functions file


vmartinez-cu:
The checks are failing because the validation file is barely too large to add to the repo. Waiting for an updated file from HIT

greglucas:
> The checks are failing because the validation file is barely too large to add to the repo. Waiting for an updated file from HIT

@vmartinez-cu, this might be a great use-case for the remote file updates @bourque did in https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1336

vmartinez-cu:
> > The checks are failing because the validation file is barely too large to add to the repo. Waiting for an updated file from HIT
> 
> @vmartinez-cu, this might be a great use-case for the remote file updates @bourque did in #1336

Yes, I was thinking that after Matthew shared his work during the tag up today. Though I think the problem is that all the numbers in the file have way more decimal places than defined for this product which results in a large file. When I reduce them to just one decimal place, the file size goes down by half. I'm following up with HIT on this

nkerman:
> > > The checks are failing because the validation file is barely too large to add to the repo. Waiting for an updated file from HIT
> > 
> > 
> > @vmartinez-cu, this might be a great use-case for the remote file updates @bourque did in #1336
> 
> Yes, I was thinking that after Matthew shared his work during the tag up today. Though I think the problem is that all the numbers in the file have way more decimal places than defined for this product which results in a large file. When I reduce them to just one decimal place, the file size goes down by half. I'm following up with HIT on this

Looks like the file is currently 500 kb, right? I was also thinking that's right on the cusp of 'too big'. It's nice if we can keep the test data within the repo, but if it's likely to change much/at all, then that's a big file to keep git history on. If it's unlikely to change, then I think it's probably okay here for now, but I expect there will be lots of large validation files across all the instruments, so we might want to settle on a standard place to keep them. 

On CLARREO, we had a big `processing_properties` file which contained lots of S3 file paths for testing and validation, all held in the same place, but I think it'd also be fine to let each validation test file keep its own path.

vmartinez-cu:
> > > > The checks are failing because the validation file is barely too large to add to the repo. Waiting for an updated file from HIT
> > > 
> > > 
> > > @vmartinez-cu, this might be a great use-case for the remote file updates @bourque did in #1336
> > 
> > 
> > Yes, I was thinking that after Matthew shared his work during the tag up today. Though I think the problem is that all the numbers in the file have way more decimal places than defined for this product which results in a large file. When I reduce them to just one decimal place, the file size goes down by half. I'm following up with HIT on this
> 
> Looks like the file is currently 500 kb, right? I was also thinking that's right on the cusp of 'too big'. It's nice if we can keep the test data within the repo, but if it's likely to change much/at all, then that's a big file to keep git history on. If it's unlikely to change, then I think it's probably okay here for now, but I expect there will be lots of large validation files across all the instruments, so we might want to settle on a standard place to keep them.
> 
> On CLARREO, we had a big `processing_properties` file which contained lots of S3 file paths for testing and validation, all held in the same place, but I think it'd also be fine to let each validation test file keep its own path.

This validation file shouldn't change. I agree though that we should probably look at a standard approach for storing and accessing validation files

greglucas (approved):

sdhoyt (approved):
Great work handling these validation files!

PR #1353, opened 2025-02-10
Temporary SPICE fix 

maxinelasp:
# Change Summary
Hotfix to get processing running again. Moving key SPICE files out of the test/ directory since that's no longer included in release. 

## Overview

Moved SPICE files and updated the path. This worked to fix the release locally.


greglucas:
@maxinelasp it looks like the tests are failing here, so likely a SPICE path reference needs to be updated somewhere in the conftest setup is my guess.

What is the underlying issue here though? I worry about moving these into the main package section because then people may always be pointing at these during processing rather than the actual SPICE kernels later. Is another option to add a SPICE kernel loader/downloader to the processing runner step somewhere?

tmplummer:
> @maxinelasp it looks like the tests are failing here, so likely a SPICE path reference needs to be updated somewhere in the conftest setup is my guess.
> 
> What is the underlying issue here though? I worry about moving these into the main package section because then people may always be pointing at these during processing rather than the actual SPICE kernels later. Is another option to add a SPICE kernel loader/downloader to the processing runner step somewhere?

@greglucas 
Sorry for not posting anything here. @maxinelasp and I had a brief slack conversation about this. I will add details here.

There are several places that currently reference these time kernels:
- The `@ensure_spice` decorator that is updated in this PR.
- The `furnish_time_kernels` fixture. (Only used in `test_time.py` and could be totally removed an rely on the `@ensure_spice` decorator. I generally like the explicit call in the test b/c it isn't opaque like `@ensure_spice`)
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/conftest.py#L177
- In the template metakernel files.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/spice/test_data/imap_ena_sim_metakernel.template#L1
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/spice/test_data/imap_simple_metakernel.template#L1
- In many individual tests that use the `furnish_kernels` fixture.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/conftest.py#L198

So, yes there is probably an overly complex multitude of ways that kernels are getting furnished for tests. I agree with Greg that the best option that we will eventually want is to download the latest sclk and leapsecond kernels as part of the processing runner step. This may also apply to ALL SPICE dependencies once that part of the infrastructure gets worked out. Is the infrastructure in place to add that auto-download for a hard coded location of the two time kernels now?

tech3371:
I thought about this too but every workaround will take time. Simpler solution for the time being may be to include test folder back in release. Then once some of us finish algorithm work and has more time, we can remove tests folder again and add actual download capabilities of SPICE. (FYI, I am not working on infrastructure work that was being blocked by this issue. That's why I am suggesting that temporarily fix.)

greglucas:
> Is the infrastructure in place to add that auto-download for a hard coded location of the two time kernels now?

Yes, put those files into the s3 bucket and reference the explicit filename when downloading. xref: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1336
I think we'd need to figure out where these need to be downloaded/used from still though, which I'm unclear about during processing.

maxinelasp:
Sorry, I never came back to update this. I just wanted something super quick since processing is currently broken due to the missing SPICE files. So yes, long term, we do want to download from AWS, but I was just trying to get a few test files quickly and thought this would work. It did break all the tests, I talked with Tim, and decided that it wasn't worth trying to fix this way. So I'll close this PR. I agree with Tenzin that a better temp fix is to add this folder back in to the release.

PR #1354, opened 2025-02-10
Ultra l1b - Bugs fixed for cdf creation

laspsandoval:
# Change Summary

## Overview
Found some small bugs that I want to fix before I forget.

## Updated Files
- de.py
   - Realized that the non-count filter was happening later in the code than it should.
- ultra_l1_utils.py
   - Added some information for the component coordinate
   - Added some information for the component coordinate
   
## Testing
Generated a cdf and viewed data.


nkerman (approved):
LGTM!

PR #1357, opened 2025-02-13
SWE: Put data into spin angle bins and add remaining CDF attrs

### LINKED ISSUE #887: SWE: Organize data by energy and spin phase
Organize data by energy and spin phase

- SWE data does not line up perfectly with SC spins. Because of this, each quarter will not be exactly same angle every time. This is an issue.

   - Eg. let's say first quarter angle goes from 0-365, second quarter angle goes from 5-370 and so on. Notice that it's not perfectly 0-360 or some consistent angle range. This is the issue.

- Because of this, when we are populating the checker board, we are assuming that the start angle is always same. The checker board has angle index of 1-30. Real angle correspond to the index is not always same because of what we wrote in above example.
- We need to revisit this pre and post launch. But right now, we assume that SWE data lines up with SC spin aka every angle index of 1-30 has same angle associated to it.
- pre-launch plan for this spin phase
    - we will discuss this again after we finish other L2 development work.

### LINKED ISSUE #915: Create SWE L2 CDF attrs and file

tech3371:
# Change Summary
closes #887 

## Overview
This PR does last piece of L2 which is organize data by energy and spin angle. I also added new data variables and their CDF attrs per SWE's requet.

## Updated Files
- imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
   - add CDF attrs for new data variables
   - Fix time bug. Change milliseconds to be microseconds
   - Main code change of this PR
   - bin flux and phase space density data into its bin.
   - New data variables and their CDF attrs
   - Minor refactor to share energy index and value.

## Testing

### DIFF
--- a/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
@@ -39,12 +39,12 @@ inst_az:
   CATDESC: Spin angle bins in instrument frame. Angle resolution is 12 degree nominally
   FIELDNAM: spin angle
   FILLVAL: -9223370000000000000
-  FORMAT: I3
+  FORMAT: E5.4
   LABLAXIS: Angle
   SCALE_TYP: linear
   UNITS: "Degree"
-  VALIDMAX: 360
-  VALIDMIN: 0
+  VALIDMAX: 359.999
+  VALIDMIN: 0.00
   VAR_TYPE: support_data
   VAR_NOTES: >
     Bin CENTERS  - Assuming 0.5 sec measurements and a spin starting
@@ -68,7 +68,7 @@ inst_el:
   CATDESC: Angle of each CEM detectors
   FIELDNAM: CEM Angle
   FILLVAL: -9223370000000000000
-  FORMAT: E14.7
+  FORMAT: I2
   LABLAXIS: Angle
   SCALE_TYP: linear
   UNITS: Degree
@@ -127,19 +127,22 @@ inst_el_label:
   VAR_TYPE: metadata
 
 # <=== Data Variables ===>
-# Default Attrs for all data variables unless overridden
 
-default_attrs: &default
+inst_az_spin_sector:
+  CATDESC: Spin angle organized by voltage step and spin sector
   DEPEND_0: epoch
-  DISPLAY_TYPE: 'no_plot'
-  LABLAXIS: ' '
-  FILLVAL: -9223372036854775808
-  FORMAT: I19
-  UNITS: ' '
-  VALIDMIN: 0
-  VALIDMAX: 9223372036854769664
-  VAR_TYPE: support_data
-  SCALETYP: linear
+  DEPEND_1: esa_step
+  DEPEND_2: spin_sector
+  LABL_PTR_1: esa_step_label
+  LABL_PTR_2: spin_sector_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Spin Angle
+  FILLVAL: -1.0000000E+31
+  FORMAT: E4.5
+  UNITS: Degree
+  VALIDMAX: 359.99
+  VALIDMIN: 0.0
+  VAR_TYPE: data
 
 phase_space_density_spin_sector:
   CATDESC: Phase space density organized by voltage step and spin sector and CEM
@@ -154,6 +157,7 @@ phase_space_density_spin_sector:
   FIELDNAM: Phase Space Density
   FILLVAL: -1.0000000E+31
   UNITS: s^3/ (cm^6 * ster)
+  FORMAT: E14.7
   VALIDMAX: 0.000015514
   VALIDMIN: 0
   VAR_TYPE: data
@@ -171,6 +175,7 @@ phase_space_density:
   FIELDNAM: Phase Space Density
   FILLVAL: -1.0000000E+31
   UNITS: s^3/ (cm^6 * ster)
+  FORMAT: E14.7
   VALIDMAX: 0.000015514
   VALIDMIN: 0
   VAR_TYPE: data
@@ -188,6 +193,7 @@ flux_spin_sector:
   FIELDNAM: Flux
   FILLVAL: -1.0000000E+31
   UNITS: 1 / (2 * eV * cm^2 * s * ster)
+  FORMAT: E14.7
   VALIDMAX: 0.000015514
   VALIDMIN: 0
   VAR_TYPE: data
@@ -205,6 +211,7 @@ flux:
   FIELDNAM: Flux
   FILLVAL: -1.0000000E+31
   UNITS: 1 / (2 * eV * cm^2 * s * ster)
+  FORMAT: E14.7
   VALIDMAX: 0.000015514
   VALIDMIN: 0
   VAR_TYPE: data
@@ -219,6 +226,9 @@ acquisition_time:
   DISPLAY_TYPE: spectrogram
   FIELDNAM: Acquisition time by volt step and spin sector
   FILLVAL: -1.0000000E+31
+  FORMAT: I20
+  VALIDMAX: 9223372036854775807
+  VALIDMIN: 0
   UNITS: sec
   VAR_TYPE: support_data
   VAR_NOTES: >
--- a/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
@@ -36,7 +36,7 @@ spin_sector:
   VAR_TYPE: support_data
 
 inst_az:
-  CATDESC: Spin angle bins. Angle resolution is 12 degree nominally
+  CATDESC: Spin angle bins in instrument frame. Angle resolution is 12 degree nominally
   FIELDNAM: spin angle
   FILLVAL: -9223370000000000000
   FORMAT: I3
--- a/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
@@ -108,6 +108,24 @@ cem_id_label:
   FORMAT: A1
   VAR_TYPE: metadata
 
+energy_label:
+  CATDESC: Energy of the particle in eV
+  FIELDNAM: Energy
+  FORMAT: A5
+  VAR_TYPE: metadata
+
+inst_az_label:
+  CATDESC: Spin angle bins in instrument frame. Angle resolution is 12 degree nominally
+  FIELDNAM: Spin Angle
+  FORMAT: A3
+  VAR_TYPE: metadata
+
+inst_el_label:
+  CATDESC: Angle of each CEM detectors
+  FIELDNAM: CEM Angle
+  FORMAT: A3
+  VAR_TYPE: metadata
+
 # <=== Data Variables ===>
 # Default Attrs for all data variables unless overridden
 
@@ -135,7 +153,24 @@ phase_space_density_spin_sector:
   DISPLAY_TYPE: spectrogram
   FIELDNAM: Phase Space Density
   FILLVAL: -1.0000000E+31
-  UNITS: " "
+  UNITS: s^3/ (cm^6 * ster)
+  VALIDMAX: 0.000015514
+  VALIDMIN: 0
+  VAR_TYPE: data
+
+phase_space_density:
+  CATDESC: Phase space density organized by energy(eV) and spin angle and CEM angle
+  DEPEND_0: epoch
+  DEPEND_1: energy
+  DEPEND_2: inst_az
+  DEPEND_3: inst_el
+  LABL_PTR_1: energy_label
+  LABL_PTR_2: inst_az_label
+  LABL_PTR_3: inst_el_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Phase Space Density
+  FILLVAL: -1.0000000E+31
+  UNITS: s^3/ (cm^6 * ster)
   VALIDMAX: 0.000015514
   VALIDMIN: 0
   VAR_TYPE: data
@@ -152,7 +187,24 @@ flux_spin_sector:
   DISPLAY_TYPE: spectrogram
   FIELDNAM: Flux
   FILLVAL: -1.0000000E+31
-  UNITS: " "
+  UNITS: 1 / (2 * eV * cm^2 * s * ster)
+  VALIDMAX: 0.000015514
+  VALIDMIN: 0
+  VAR_TYPE: data
+
+flux:
+  CATDESC: Flux organized by energy(eV) and spin angle and CEM angle
+  DEPEND_0: epoch
+  DEPEND_1: energy
+  DEPEND_2: inst_az
+  DEPEND_3: inst_el
+  LABL_PTR_1: energy_label
+  LABL_PTR_2: inst_az_label
+  LABL_PTR_3: inst_el_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Flux
+  FILLVAL: -1.0000000E+31
+  UNITS: 1 / (2 * eV * cm^2 * s * ster)
   VALIDMAX: 0.000015514
   VALIDMIN: 0
   VAR_TYPE: data
--- a/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
@@ -11,17 +11,45 @@ esa_step:
   VALIDMIN: 0
   VAR_TYPE: support_data
 
-spin_angle:
-  CATDESC: Spin angle. Angle resolution is 12 degree nominally
-  FIELDNAM: spin angle
+energy:
+  CATDESC: Energy of the particle in eV
+  FIELDNAM: Energy (eV)
+  FILLVAL: -9223370000000000000
+  FORMAT: E5.4
+  LABLAXIS: Energy
+  SCALE_TYP: linear
+  UNITS: eV
+  VALIDMAX: 2000.0
+  VALIDMIN: 0.0
+  VAR_TYPE: support_data
+
+spin_sector:
+  CATDESC: Spin sector. 30 measurements nominally
+  FIELDNAM: Spin Sector
   FILLVAL: -9223370000000000000
   FORMAT: I2
-  LABLAXIS: Angle
+  LABLAXIS: Spin Sector
   SCALE_TYP: linear
-  UNITS: "Degree"
+  UNITS: " "
   VALIDMAX: 29
   VALIDMIN: 0
   VAR_TYPE: support_data
+
+inst_az:
+  CATDESC: Spin angle bins. Angle resolution is 12 degree nominally
+  FIELDNAM: spin angle
+  FILLVAL: -9223370000000000000
+  FORMAT: I3
+  LABLAXIS: Angle
+  SCALE_TYP: linear
+  UNITS: "Degree"
+  VALIDMAX: 360
+  VALIDMIN: 0
+  VAR_TYPE: support_data
+  VAR_NOTES: >
+    Bin CENTERS  - Assuming 0.5 sec measurements and a spin starting
+    at Angle=0, these are the angles at the middle of each measurement
+    bin.
 
 cem_id:
   CATDESC: Data of each CEM detectors
@@ -36,16 +64,16 @@ cem_id:
   VAR_NOTES: " "
   VAR_TYPE: support_data
 
-spin_sector:
-  CATDESC: Spin sector. 30 measurements nominally
-  FIELDNAM: Spin Sector
+inst_el:
+  CATDESC: Angle of each CEM detectors
+  FIELDNAM: CEM Angle
   FILLVAL: -9223370000000000000
-  FORMAT: I2
-  LABLAXIS: Spin Sector
+  FORMAT: E14.7
+  LABLAXIS: Angle
   SCALE_TYP: linear
-  UNITS: " "
-  VALIDMAX: 29
-  VALIDMIN: 0
+  UNITS: Degree
+  VALIDMAX: 63
+  VALIDMIN: -63
   VAR_TYPE: support_data
 
 cycle:
--- a/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
@@ -52,7 +52,7 @@ inst_az:
     bin.
 
 cem_id:
-  CATDESC: Data of each CEM detectors
+  CATDESC: Data of each CEM detector
   FIELDNAM: CEM data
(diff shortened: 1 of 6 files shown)


imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
```diff
@@ -36,16 +64,16 @@ cem_id:
   VAR_NOTES: " "
   VAR_TYPE: support_data
 
-spin_sector:
-  CATDESC: Spin sector. 30 measurements nominally
-  FIELDNAM: Spin Sector
+inst_el:
+  CATDESC: Angle of each CEM detectors
```
greglucas:
```suggestion
  CATDESC: Angle of each CEM detector
```

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
```diff
@@ -107,7 +156,26 @@ phase_space_density_spin_sector:
...
   FILLVAL: -1.0000000E+31
-  UNITS: " "
+  UNITS: s^3/ (cm^6 * ster)
+  FORMAT: E14.7
+  VALIDMAX: 0.000015514
+  VALIDMIN: 0
+  VAR_TYPE: data
+
+phase_space_density:
+  CATDESC: Phase space density organized by energy(eV) and spin angle and CEM angle
```
greglucas: Same for flux below.
```suggestion
  CATDESC: Phase space density organized by energy (eV), spin angle, and CEM angle
```

imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
```diff
@@ -107,7 +156,26 @@ phase_space_density_spin_sector:
...
+  DEPEND_1: energy
+  DEPEND_2: inst_az
+  DEPEND_3: inst_el
+  LABL_PTR_1: energy_label
+  LABL_PTR_2: inst_az_label
+  LABL_PTR_3: inst_el_label
+  DISPLAY_TYPE: spectrogram
+  FIELDNAM: Phase Space Density
+  FILLVAL: -1.0000000E+31
+  UNITS: s^3/ (cm^6 * ster)
```
greglucas:
```suggestion
  UNITS: s^3 / (cm^6 * ster)
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -356,10 +326,15 @@ def populate_full_cycle_data(
                 full_cycle_data[esa_voltage_row_index][column_index] = corrected_counts[
                     step
                 ]
-                # Put acquisition time in acquisition_times array
+                # Acquisition time of each count data point will be calculated
```
```suggestion
                # Acquisition time (in seconds) of each count data point will be calculated
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -33,6 +37,8 @@
 # See doc string of calculate_flux() for more details.
 FLUX_CONVERSION_FACTOR = 6.187e30
 
+CEM_DETECTORS_ANGLE = np.array([-63, -42, -21, 0, 21, 42, 63])
```
greglucas, line 40: Do you want these in the `swe_utils` file too like your other global constants?

Should we be getting these from the SPICE Frame definitions rather than hard-coding here?

imap_processing/swe/l2/swe_l2.py
```diff
@@ -195,6 +201,59 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    -------
+    numpy.ndarray
+        Data in bins. Shape: (full_cycle_data, 24, 30, 7).
+    """
+    # Initialize with zeros instead of NaN because np.add.at() does not
+    # work with nan values. It results in nan + value = nan
+    binned_data = np.zeros((data.shape[0], 24, 30, 7), dtype=np.float64)
+
+    time_indices = np.arange(data.shape[0])[:, None, None]
+    energy_indices = np.arange(data.shape[1])[None, :, None]
```
greglucas: For the future, I wonder if we should make these global constants as well? You use the 24, 30, 7 values a lot throughout the code. `N_ESA_STEPS`, `N_ANGLES`, `N_CEMS`
```suggestion
    energy_indices = np.arange(24)[None, :, None]
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -195,6 +201,59 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    angle_indices = angle_bin_indices
```
greglucas: Call the argument to the function `angle_indices` instead of assigning another variable here?

imap_processing/swe/l2/swe_l2.py
```diff
@@ -195,6 +201,59 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    # Use np.add.at() to accumulate values into bins
+    np.add.at(binned_data, (time_indices, energy_indices, angle_indices), data)
+
+    # Count occurrences in each bin to compute the mean.
+    # Ensure float dtype for division
+    bin_counts = np.zeros_like(binned_data, dtype=float)
+    np.add.at(bin_counts, (time_indices, energy_indices, angle_indices), 1)
+
+    # Compute the mean. Replace zero counts with NaN to indicate no data in the bin
+    # because zero counts could be valid data.
```
greglucas: This confused me a bit, because you have number of elements in a bin "counts", and you have data "counts". Maybe add the word data or physical or something else here to distinguish the two counts.
```suggestion
    # because zero physical counts could be valid data.
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -216,33 +275,85 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
+    inst_el_label = xr.DataArray(
+        CEM_DETECTORS_ANGLE.astype(str),
+        name="inst_el_label",
+        dims=["inst_el"],
+        attrs=cdf_attributes.get_variable_attributes("inst_el_label"),
+    )
+
+    # Spin Angle bins storing bin center values.
+    inst_az_xr = xr.DataArray(
+        np.arange(6, 360, 12),
```
greglucas: Maybe say something about each bin being 12 degrees wide?

Minor comment, so only if this helps you, but I was a bit confused by the 6 at first because I thought it might have been related to NCEMS and off by one, but then realized a second later it was the offset centers.
Perhaps: `np.arange(0, 360, 12) + 6`

imap_processing/swe/l2/swe_l2.py
```diff
@@ -216,33 +275,85 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
+        name="inst_az",
+        dims=["inst_az"],
+        attrs=cdf_attributes.get_variable_attributes("inst_az"),
+    )
+    inst_az_label = xr.DataArray(
+        np.arange(6, 360, 12).astype(str),
```
greglucas: Can you do `inst_az_xr.astype(str)` here so you keep these in sync?

imap_processing/swe/l2/swe_l2.py
```diff
@@ -251,13 +362,96 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
-    data_acq_time = l1b_dataset["acquisition_time"].data.flatten()
+    # Calculate spin phase using SWE acquisition_time from the
+    # L1B dataset. The L1B dataset stores acquisition_time with
+    # dimensions (epoch, esa_step, spin_sector). Use center time
+    # to calculate spin phase. This center time calculation is
+    # necessary to accurately determine the center angle of the data.
+    #
+    # To determine the center acquisition time, we adjust the
+    # recorded acquisition_time as follows:
+    #   acquisition_time + (acq_duration / 1000000) / 2
```
greglucas, line 431: I think this is going to be a discussion during the upcoming meeting and whether we are recording our times as the "acquisition_start" or "center of accumulation" period. If you always need the center, can you convert to that earlier in the process and use that as the timings for everything?

greglucas, line 431: Is this equation correct? I think you want to remove settle_duration too.

```
center_of_acquisition = acquisition_start + settle_time + acquisition_time / 2
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -251,13 +362,96 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
+    #
+    # Here, acq_duration is given in microseconds and is stored
+    # in the L1B dataset with dimensions (epoch, cycle). Since acq_duration
+    # remains the same for all quarter cycles within a full sweep,
+    # we use the first acq_duration value for each full sweep to perform
+    # this adjustment.
+
+    acq_duration = l1b_dataset["acq_duration"].data[:, 0] / 2000000
```
greglucas, line 439: If acq_duration is the same for all cycles, can you remove it from the acq_duration dataset?

I think you could drop the `.data` calls, and potentially it will be auto-broadcast for you by just writing these with the appropriate datasets?

```python
data_acq_time = l1b_dataset["acquisition_time"] + l1b_data["acq_duration"] / 2e6
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -251,13 +362,96 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
+    #   - phi_begin <= center - 6
+    #   - phi_center = 6
+    #   - phi_end < center + 6
+    #
+    # For example, if the input angle is 8.4, it falls within the bin centered at 6.
+    #
+    # To make binning easier, we define bin edges as:
+    #   [0, 12, 24, 36, 48, 60, 72, 84, 96, 108, 120, 132, 144, 156, 168,
+    #    180, 192, 204, 216, 228, 240, 252, 264, 276, 288, 300, 312, 324,
+    #    336, 348]
```
greglucas: I think you need a 360 here too to get the high edge?

greglucas (commented):
I think you were testing the `put_data_into_angle_bins` manually before? It would be good to add some explicit tests for that function since it is non-trivial. I also have a feeling we'll have more teams wanting to do something similar, so may need to re-use or refactor that out to use it with other datasets.

imap_processing/swe/l2/swe_l2.py
tech3371: I thought about this too and asked Ruth. Sounds like there is some hesitation.

tech3371: I asked Ruth about this formula I was using but will double check on this.

imap_processing/swe/l2/swe_l2.py
tech3371: Ruth asked about this too. If we have 360, she wants that in bin 0. Given that our spin phase is `[0,1)`, I don't think we will have value that will equal 360. Is my understanding right?

imap_processing/swe/l2/swe_l2.py
tech3371: I can do that in a future PR. Also, Ruth is thinking about equatorial angle or something else. She told me to start with this for now.

imap_processing/swe/l2/swe_l2.py
tech3371: sounds good. I can do that in future PR

imap_processing/swe/l2/swe_l2.py
tech3371: Created this issue to ask Ruth about these questions and also captures things I want to do in future PR. https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1360

tech3371:
>I think you were testing the put_data_into_angle_bins manually before? It would be good to add some explicit tests for that function since it is non-trivial. I also have a feeling we'll have more teams wanting to do something similar, so may need to re-use or refactor that out to use it with other datasets.

@greglucas Did you want me to do this refactor for this PR or in the future?

imap_processing/swe/l2/swe_l2.py
greglucas: It'd be worth checking what bins you want the various values to go into. For example, a quick test like this. I think the final one is what you're after?

```python
>>> x = np.arange(0, 360, 12)
>>> vals = [0, 345, 348, 355, 360]
>>> x2 = np.arange(0, 361, 12)
>>> np.searchsorted(x, vals, side="left")
array([ 0, 29, 29, 30, 30])
>>> np.searchsorted(x, vals, side="right")
array([ 1, 29, 30, 30, 30])
>>> np.searchsorted(x2, vals, side="left")
array([ 0, 29, 29, 30, 30])
>>> np.searchsorted(x2, vals, side="right")
array([ 1, 29, 30, 30, 31])
```

Then set the 31 value (>= 360) to the first bin manually.
```python
indices[indices == 31] = 1
```

greglucas:
> @greglucas Did you want me to do this refactor for this PR or in the future?

I think some tests would really help (I thought you had something to test the for-loop version vs this vectorized version, so just dropping that in here to prove to yourself this is doing what you think and expanding dimensions properly etc...).

The refactor for more general use, no. We can tackle that when the need arises, just trying to mention the reason I think the tests are important in case we do move this later. Sorry for not being clear!

tech3371:
> > @greglucas Did you want me to do this refactor for this PR or in the future?
> 
> I think some tests would really help (I thought you had something to test the for-loop version vs this vectorized version, so just dropping that in here to prove to yourself this is doing what you think and expanding dimensions properly etc...).
> 
> The refactor for more general use, no. We can tackle that when the need arises, just trying to mention the reason I think the tests are important in case we do move this later. Sorry for not being clear!

Ah ok. Cool. I will write some tests for cases you mentioned and what I tested manually.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -251,13 +362,96 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
+    # this adjustment.
+
+    acq_duration = l1b_dataset["acq_duration"].data[:, 0] / 2000000
+    data_acq_time = (
+        l1b_dataset["acquisition_time"].data + acq_duration[:, np.newaxis, np.newaxis]
+    )
+
+    # Calculate spin phase
+    inst_spin_phase = get_instrument_spin_phase(
+        query_met_times=data_acq_time.flatten(),
```
tmplummer: I would suggest `.ravel()` instead of `.flatten()` here. `ravel()` creates a view into the data when possible while `flatten()` always copies the data.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -251,13 +362,96 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
+    # Example test cases:
+    #   - `np.searchsorted(x, [6], side="right") -> [1]` (Bin center test)
+    #   - `np.searchsorted(x, [8.4], side="right") -> [1]` (Edge case near center)
+    #   - `np.searchsorted(x, [12], side="right") -> [2]` (Bin end test)
+    #   - `np.searchsorted(x, [0], side="right") -> [1]` (Bin start test)
+    #
+    # Using `i-1` ensures that all input angles are assigned to the correct bin of
+    # centered angle bins.
+
+    spin_angle_bins_range = np.arange(0, 360, 12)
```
tmplummer: I suggest renaming this to be consistent with your above comment as well as histogram nomenclature.
```suggestion
    spin_angle_bin_edges = np.arange(0, 360, 12)
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -195,6 +201,59 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    """
+    # Initialize with zeros instead of NaN because np.add.at() does not
+    # work with nan values. It results in nan + value = nan
+    binned_data = np.zeros((data.shape[0], 24, 30, 7), dtype=np.float64)
+
+    time_indices = np.arange(data.shape[0])[:, None, None]
+    energy_indices = np.arange(data.shape[1])[None, :, None]
+    angle_indices = angle_bin_indices
+
+    # Use np.add.at() to accumulate values into bins
```
tmplummer, line 240: Cool, I learned something new. For the L2 code, we are currently using `np.bincount` for doing this accumulation but it requires a 1D array. If you want to go down the rabbit hole, I found this interesting SO post comparing the two functions: https://stackoverflow.com/questions/67449766/comparing-bincount-and-add-at

tmplummer (commented):
Overall, looks great. I left a couple of nits and thanks for teaching me about a new numpy function.

imap_processing/swe/l2/swe_l2.py
tech3371: cool. I didn't know about `np.bincount`. That's good to know. For my case, it seems like it won't work unless there is workaround. I have to keep energy steps information still when binning and bin their angle into it's angle bins across each energy steps. Because of this, I couldn't flatten my array and had to keep same shape.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -195,6 +201,119 @@ def calculate_flux(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    # because zero physical counts could be valid data.
+    bin_counts[bin_counts == 0] = np.nan
+    binned_data /= bin_counts
+
+    return binned_data
+
+
+def find_angle_bin_indices(
+    inst_spin_angle: np.ndarray,
+    spin_angle_bin_edges: np.ndarray,
```
greglucas: I think these are fixed based on your comment below. Can we remove this argument so you are just looking up in your fixed lookup table essentially? If we do need to change it later, perhaps add the argument then.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -216,33 +335,85 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
     )
 
-    # Phase space density in the spin sector. This is carrying over for L3 purposes.
-    # TODO: later, we will calculate and organize phase space density in the
-    # spin angle bins.
+    ############################################################
+    # Calculate phase space density and flux. Store data in shape
+    # (epoch, esa_step, spin_sector, cem_id). This is for L3 purposes.
+    ############################################################
+    phase_space_density = calculate_phase_space_density(l1b_dataset)[
```
greglucas, line 402: Since I see this as a changed line now, I think this would be better returned directly from the function as `return ds["phase_space_density"]` so you don't have to index into it later. (based on the name of the function you are returning a single dataarray, not the entire dataset.

This can be a follow-up PR.

imap_processing/swe/l2/swe_l2.py
greglucas: Thanks for opening that ticket!

imap_processing/swe/l2/swe_l2.py
```diff
@@ -251,13 +422,66 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
+    inst_spin_phase = get_instrument_spin_phase(
+        query_met_times=data_acq_time.ravel(),
+        instrument=SpiceFrame.IMAP_SWE,
+    )
+
+    # Convert spin phase to spin angle in degrees.
+    inst_spin_angle = get_spin_angle(inst_spin_phase, degrees=True).reshape(-1, 24, 30)
+
+    # Save spin angle in dataset per SWE request.
+    dataset["inst_az_spin_sector"] = xr.DataArray(
```
greglucas, line 454: I thought spin "sector" was a binned value (an index of values between two angles), but I think this is just any generic "angle" instead? Maybe I have that backwards though.

Below you are calculating the spin_angle_bins_indices, which I think is the "sector" value instead of the actual angle.

imap_processing/tests/swe/test_swe_l2.py
```diff
@@ -109,6 +111,201 @@ def test_calculate_flux():
     assert type(flux) == np.ndarray
 
 
+def test_find_angle_bin_indices():
+    """Test find_angle_bin_indices function."""
+    spin_angle_bin_center = np.arange(6, 360, 12)
+    spin_angle_bin_edges = np.arange(0, 360, 12)
+
+    start_angles = [0, 12, 24, 36, 48, 60, 72, 84, 96, 108]
```
greglucas: This is definitely above and beyond what I was expecting! :) I'd say you really only need to test a few intermediate values and the extremes, and I'd generally put them all into a single array at once. `np.array([0, 1, 6, 9, 12, 355])`

imap_processing/tests/swe/test_swe_l2.py
```diff
@@ -109,6 +111,201 @@ def test_calculate_flux():
...
+    # Test that the binned data has correct values in correct bins by
+    # checking that odd number columns are filled with nan
+    expected_binned_data = np.full(
+        (num_cycles, num_esa_step, num_angle_bins, num_cems), np.nan
+    )
+    np.testing.assert_array_equal(
+        binned_data[0, 0, 1::2, 0], expected_binned_data[0, 0, 1::2, 0]
+    )
+
+    # Now check that mean calculation is correct
```
greglucas, line 208: Where did these values come from? It is definitely non-obvious looking at this test data.

One thing that might help in the future is to simplify the tests as well if that is possible (in addition to a nice filled out array like this!). I can imagine initializing a `np.zeros(num_cycles, num_esa_step, num_angle_bins, num_cems)`, and then just filling one or two explicit entries that could be analyzed directly. `binned_data[0, 0, 0, 1:2] = 5`, along with angle_bin_indices being all zeros except for one or two values.

greglucas (approved):

imap_processing/swe/l2/swe_l2.py
tech3371: I changed based on feedback from previous PR feedback. That functions returns two things also. That was why I think we decided to return as intermediate dataset. Something I can revisit in follow-up PR.

imap_processing/swe/l2/swe_l2.py
tech3371: This variable store spin angle data organized by energy steps, spin sector. That's why I appended `spin_sector` like the other ones. I can change that name though.

greglucas: I'd say that is up to you and Ruth. But to me, the organization is in the dimension names already. The value the data represents should be in the variable name. But that is just my opinion and does not matter if you are both OK with this :)

imap_processing/tests/swe/test_swe_l2.py
tech3371: I manually calculated those values based on above bin values from actual data. Will think about it.

greglucas: Sorry if I wasn't clear on any of this too. This is great for now, and my approval still stands!

tech3371: That helps! I was about to look into it too.

PR #1358, opened 2025-02-13
CoDICE L1a lo-counters-aggregated and lo-counters-singles validation

bourque:
This PR gets the CoDICE L1a pipeline in a state to validate the `lo-counters-aggregated` and `lo-counters-singles` data products.

One thing to note here is that the data variables that make up the `lo-counters-aggregated` product may change (infrequently) over time, and thus they need to be able to be switched on/off. These code changes allow for that flexibility.

Pertains to #251


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -56,6 +56,20 @@ spin_sector:
...
 
+# TODO: These values will likely change, check with Joey
+spin_sector_pairs:
+    CATDESC: Spin sector Pairs
+    FIELDNAM: Spin Sector Pairs
+    FILLVAL: -1
+    FORMAT: I1
+    LABLAXIS: " "
+    SCALETYP: linear
+    SI_CONVERSION: " > "
```
greglucas: I couldn't figure this out from the SPDF site, but SI_CONVERSION seems weird as a greater than symbol. I don't see this on any other instruments metadata, so can it just be removed?

imap_processing/codice/codice_l1a.py
```diff
@@ -156,6 +156,13 @@ def define_coordinates(self) -> None:
                 values = self.calculate_epoch_values()
             elif name in ["esa_step", "inst_az", "spin_sector"]:
                 values = np.arange(self.config["dims"][name])
+            elif name == "spin_sector_pairs":
+                # TODO: Need to ask Joey about this. Somehow he managed to use
+                #       strings as coordinates in his validation data, even
+                #       though that is not allowed.
+                #       Ideally these should read "0-30 deg", "30-60 deg", etc.
+                #       For now, just use bogus integers
+                values = np.array([0, 3, 6, 9, 1, 1])
```
greglucas: Do these need to be unique? I'd suggest `np.arange(5)` to make it easier to know what is where.

This would be something good to ping our slack channel about quickly. His CDF file may not be fully ISTP compliant. I think others have dealt with similar issues and had to create another variable and point he coordinate at the label or something similar.

imap_processing/codice/constants.py
```diff
@@ -78,6 +77,46 @@
 HI_OMNI_SPECIES_VARIABLE_NAMES = ["h", "he3", "he4", "c", "o", "ne_mg_si", "fe", "uh"]
 HI_SECT_SPECIES_VARIABLE_NAMES = ["h", "he3he4", "cno", "fe"]
 
+# lo-counters-aggregated data product variables are dynamically determined
+# based on the number of active counters
+# TODO: Try to convince Joey to move to lower case variable names with
+#       underscores?
```
greglucas, line 83: :) @joeymukherjee is a reviewer here, so we shall see!

When you say they are dynamically determined, how will that be done? Are we going to need to push new code changes to get these updates or is there a configuration file we could use in our dependency tree?

imap_processing/tests/codice/data/validation/imap_codice_l1a_lo-counters-aggregated_20241110193700_v0.0.0.cdf
greglucas, line 1: Are we using these for later processing steps, or do they need to be explicitly saved here now? It seems like these have been added/changed because of the change in dimensions, so I'm slightly concerned about the churn of these files too. (I'm pretty sure we'll be moving towards energy as the dimension before angle)

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: It probably can be removed? I only include it because I see it all over the place in the attributes of @joeymukherjee's validation data. It seems like a placeholder / not filled out all the way. From the ISTP guidelines:

```
The conversion factor to SI units. This is the factor that the variable must be multiplied by in order to turn it to generic SI units. It will copntain two text fields separated by the delimiter >. The first is the conversion and the second is the standard unit that it converts to. For example the magnetic field for FGM will be in nT, and to convert to Tesla the value of SI_conv will be '1.0e-9>Tesla'. The use of text allows this attribute to be parsed and the value must be extracted in software.
```

In any case, I will add a TODO here and tackle this in a future PR since it affects a whole bunch of data products.

imap_processing/codice/constants.py
bourque: It sounds like these active/inactive variables will change very infrequently, if ever, on orbit. My plan was to just update the `True`/`False` values in the `LO_COUNTERS_AGGREGATED_ACTIVE_VARIABLES` dictionary when/if that happens and push the new code change. I suppose we could stick this all in some sort of config file and read it in, but that might be overkill?

imap_processing/tests/codice/data/validation/imap_codice_l1a_lo-counters-aggregated_20241110193700_v0.0.0.cdf
bourque: I don't quite understand what you are asking here. I just added these validation CDFs because they were not in the repo with the other validation CDFs yet. I suppose I will have to update them if Joey updates the validation CDFs, which may happen once or twice more before launch, but I wouldn't expect that to happen very often.

imap_processing/codice/codice_l1a.py
bourque: That is a good suggestion and a good way to get around this issue. I would like to ask @joeymukherjee how he would like us to handle this first.

imap_processing/tests/codice/data/validation/imap_codice_l1a_lo-counters-aggregated_20241110193700_v0.0.0.cdf
greglucas: OK, I was thinking that these were _produced_ by your code. You are getting these externally and comparing your produced files against these, that makes sense to me now. Sorry for the noise, sounds good!

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
greglucas: When you are validating data products, are you comparing metadata too? I'd say don't get hung up on the metadata exact comparisons.

greglucas (approved):

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: Yeah, there are too many subtle differences between the CDF metadata I am producing (or really, the conventions that the entire team have settled on) and Joey's validation CDFs (where Joey has not put as much thought into how things are named or some of the values). So I have been ignoring metadata in automated testing, but I've had some good discussions with Joey about the metadata and I think it is in a good place.

PR #1359, opened 2025-02-13
Ultra l1b data products

laspsandoval:
# Change Summary

## Overview
Created cullingmask and badtimes data products.

## New Files
- cullingmask.py
   - Filters data based on quality flags resulting in "good" data
- badtimes.py
   - Filters data based on quality flags resulting in "bad" data

## Updated Files
- ultra_l1b_culling.py
   - Corrected small bug
- ultra_l1b.py
   - Change what is passed into calculate_badtimes

## Testing
- test_badtimes.py
- test_cullingmask.py
- tset_ultra_l1b_culling.py

### DIFF
--- a/imap_processing/tests/ultra/unit/test_cullingmask.py
+++ b/imap_processing/tests/ultra/unit/test_cullingmask.py
@@ -1,0 +1,82 @@
+import numpy as np
+import xarray as xr
+
+from imap_processing.quality_flags import ImapAttitudeUltraFlags, ImapRatesUltraFlags
+from imap_processing.ultra.l1b.cullingmask import calculate_cullingmask
+
+
+def test_calculate_cullingmask_attitude():
+    """Test calculate_cullingmask for attitude culling."""
+
+    spin_numbers = np.array([0, 1])
+    energy_bins = np.array([10, 20, 30, 40])
+
+    quality_attitude = np.full(
+        spin_numbers.shape, ImapAttitudeUltraFlags.NONE.value, dtype=np.uint16
+    )
+    quality_attitude[1] |= ImapAttitudeUltraFlags.SPINRATE.value
+
+    quality_ena_rates = np.full(
+        (len(energy_bins), len(spin_numbers)),
+        ImapRatesUltraFlags.NONE.value,
+        dtype=np.uint16,
+    )
+
+    ds = xr.Dataset(
+        {
+            "quality_attitude": (("spin_number",), quality_attitude),
+            "quality_ena_rates": (
+                ("energy_bin_geometric_mean", "spin_number"),
+                quality_ena_rates,
+            ),
+        },
+        coords={
+            "spin_number": spin_numbers,
+            "energy_bin_geometric_mean": energy_bins,
+        },
+    )
+
+    result_ds = calculate_cullingmask(ds, name="test", data_version="v1")
+
+    np.testing.assert_array_equal(result_ds["spin_number"].values, np.array([0]))
+
+
+def test_calculate_cullingmask_rates():
+    """Test calculate_cullingmask for rates culling."""
+    spin_numbers = np.array([0, 1, 2, 3])
+    energy_bins = np.array([10, 20, 30, 40])
+
+    quality_attitude = np.full(
+        spin_numbers.shape, ImapAttitudeUltraFlags.NONE.value, dtype=np.uint16
+    )
+
+    quality_ena_rates = np.full(
+        (len(energy_bins), len(spin_numbers)),
+        ImapRatesUltraFlags.NONE.value,
+        dtype=np.uint16,
+    )
+
+    quality_ena_rates[:, 0] |= ImapRatesUltraFlags.ZEROCOUNTS.value
+    quality_ena_rates[0, 1] |= ImapRatesUltraFlags.ZEROCOUNTS.value
+    quality_ena_rates[0, 3] |= ImapRatesUltraFlags.ZEROCOUNTS.value
+    quality_ena_rates[0, 2] |= ImapRatesUltraFlags.HIGHRATES.value
+    quality_ena_rates[0, 3] |= ImapRatesUltraFlags.HIGHRATES.value
+
+    ds = xr.Dataset(
+        {
+            "quality_attitude": (("spin_number",), quality_attitude),
+            "quality_ena_rates": (
+                ("energy_bin_geometric_mean", "spin_number"),
+                quality_ena_rates,
+            ),
+        },
+        coords={
+            "spin_number": spin_numbers,
+            "energy_bin_geometric_mean": energy_bins,
+        },
+    )
+
+    result_ds = calculate_cullingmask(ds, name="test", data_version="v1")
+
+    expected_spins = np.array([1])
+    np.testing.assert_array_equal(result_ds["spin_number"].values, expected_spins)
--- a/imap_processing/tests/ultra/unit/test_cullingmask.py
+++ b/imap_processing/tests/ultra/unit/test_cullingmask.py
@@ -36,7 +36,9 @@ def test_calculate_cullingmask_attitude():
         },
     )
 
-    result_ds = calculate_cullingmask(ds, name="test", data_version="v1")
+    result_ds = calculate_cullingmask(
+        ds, name="imap_ultra_l1b_45sensor-cullingmask", data_version="v1"
+    )
 
     np.testing.assert_array_equal(result_ds["spin_number"].values, np.array([0]))
 
@@ -58,9 +60,10 @@ def test_calculate_cullingmask_rates():
 
     quality_ena_rates[:, 0] |= ImapRatesUltraFlags.ZEROCOUNTS.value
     quality_ena_rates[0, 1] |= ImapRatesUltraFlags.ZEROCOUNTS.value
-    quality_ena_rates[0, 3] |= ImapRatesUltraFlags.ZEROCOUNTS.value
     quality_ena_rates[0, 2] |= ImapRatesUltraFlags.HIGHRATES.value
-    quality_ena_rates[0, 3] |= ImapRatesUltraFlags.HIGHRATES.value
+    quality_ena_rates[0, 3] |= (
+        ImapRatesUltraFlags.ZEROCOUNTS.value | ImapRatesUltraFlags.HIGHRATES.value
+    )
 
     ds = xr.Dataset(
         {
@@ -76,7 +79,9 @@ def test_calculate_cullingmask_rates():
         },
     )
 
-    result_ds = calculate_cullingmask(ds, name="test", data_version="v1")
+    result_ds = calculate_cullingmask(
+        ds, name="imap_ultra_l1b_45sensor-cullingmask", data_version="v1"
+    )
 
     expected_spins = np.array([1])
     np.testing.assert_array_equal(result_ds["spin_number"].values, expected_spins)
--- a/imap_processing/tests/ultra/unit/test_cullingmask.py
+++ b/imap_processing/tests/ultra/unit/test_cullingmask.py
@@ -83,5 +83,5 @@ def test_calculate_cullingmask_rates():
         ds, name="imap_ultra_l1b_45sensor-cullingmask", data_version="v1"
     )
 
-    expected_spins = np.array([1])
+    expected_spins = np.array([0, 1])
     np.testing.assert_array_equal(result_ds["spin_number"].values, expected_spins)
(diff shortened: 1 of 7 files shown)


imap_processing/tests/ultra/unit/test_cullingmask.py
```diff
@@ -0,0 +1,82 @@
...
+        (len(energy_bins), len(spin_numbers)),
+        ImapRatesUltraFlags.NONE.value,
+        dtype=np.uint16,
+    )
+
+    quality_ena_rates[:, 0] |= ImapRatesUltraFlags.ZEROCOUNTS.value
+    quality_ena_rates[0, 1] |= ImapRatesUltraFlags.ZEROCOUNTS.value
+    quality_ena_rates[0, 3] |= ImapRatesUltraFlags.ZEROCOUNTS.value
+    quality_ena_rates[0, 2] |= ImapRatesUltraFlags.HIGHRATES.value
+    quality_ena_rates[0, 3] |= ImapRatesUltraFlags.HIGHRATES.value
```
greglucas: I like that you're testing these cases. I think it might be more obvious that you're combining the values if you put the two options on the same line. Also in the other test files where similar things were added.
```suggestion
    quality_ena_rates[0, 1] |= ImapRatesUltraFlags.ZEROCOUNTS.value
    quality_ena_rates[0, 2] |= ImapRatesUltraFlags.HIGHRATES.value
    quality_ena_rates[0, 3] |= ImapRatesUltraFlags.ZEROCOUNTS.value | ImapRatesUltraFlags.HIGHRATES.value
```

imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
```diff
@@ -40,20 +40,16 @@ def test_data(use_fake_spin_data_for_time):
...
+def test_get_spin(use_fake_spin_data_for_time):
     """Tests get_spin function."""
 
-    de_dataset = l1b_datasets[0]
-    use_fake_spin_data_for_time(
-        de_dataset["event_times"][0], de_dataset["event_times"][-1]
-    )
-    spin_number = get_spin(de_dataset["event_times"])
+    start = 4.45015658e08
+    stop = 4.45015873e08
```
greglucas: It might be more obvious to add the offset here rather than two large numbers. Additionally, perhaps explicitly set your nspins and spin_period rather than calculating nspins based upon those below. Looking at the test, I'm unclear whether the linspace=20 would be the same number of spins as what you've calculated, so making sure they are different will tell you about the unique being a different length than the npoints.
```suggestion
    nspins = 5
    spin_period = 15
    start = 4.45015658e08
    stop = start + (nspins + 1) * spin_period
```

imap_processing/ultra/l1b/cullingmask.py
```diff
@@ -25,14 +25,32 @@ def calculate_cullingmask(
...
-        "energy_bin_geometric_mean"
-    ]
-
-    # TODO: add more data to cullingmask_dict.
-
-    cullingmask_dataset = create_dataset(cullingmask_dict, name, "l1b", data_version)
+    # If the spin rate was too high or low then the spin should be thrown out.
+    rate_mask = (
+        extendedspin_dataset["quality_attitude"] & ImapAttitudeUltraFlags.SPINRATE.value
+    ) == 0
```
greglucas: I usually think of mask as "bad" and you're trying to hide those values. Maybe `valid_data` or something like that, or invert this to a not equal condition.

imap_processing/ultra/l1b/cullingmask.py
```diff
@@ -25,14 +25,32 @@ def calculate_cullingmask(
...
+    good_attitude = extendedspin_dataset["spin_number"].values[rate_mask]
+    good_attitude_dataset = extendedspin_dataset.sel(spin_number=good_attitude)
```
greglucas: I think you might be able to index into this directly without needing the intermediate. The `good_attitude` is confusing because it is actually `good_spins`.

```suggestion
    good_attitude_dataset = extendedspin_dataset.where(rate_mask, drop=True)
```

imap_processing/ultra/l1b/cullingmask.py
```diff
@@ -25,14 +25,32 @@ def calculate_cullingmask(
...
+
+    # Counts in every energy bin should be equal to zero to throw out the spin.
```
greglucas: I'm not sure I follow this comment. "If all counts are equal to zero we want to throw out the spin"? Is that actually what we want to do, just because we have zero counts does that mean it is a "bad" spin, or just one that we got no counts during, which seems like it is useful information still and would lead to different exposure times.

imap_processing/ultra/l1b/cullingmask.py
```diff
@@ -25,14 +25,32 @@ def calculate_cullingmask(
...
+        (
+            good_attitude_dataset["quality_ena_rates"]
+            & ImapRatesUltraFlags.ZEROCOUNTS.value
+        )
+        != 0
+    ).all(dim="energy_bin_geometric_mean")
+    zero_counts_mask = ~zero_counts
+    non_zero_dataset = good_attitude_dataset.sel(
+        spin_number=good_attitude_dataset["spin_number"][zero_counts_mask]
+    )
```
greglucas: Can these masks be combined for a large boolean "where" condition? It might also help to not compare to `== 0` and `!= 0` separately and let the "mask" nots work themselves out. It is tough to follow the logic between the cases for me currently because the opposite negations.

```
condition = ((ds.quality_ena_rates & ImapRatesUltraFlags.ZEROCOUNTS.value == 0) | (ds.quality_attitude & ImapAttitudeUltraFlags.SPINRATE.value))
```

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -103,8 +103,8 @@ def flag_attitude(eventtimes_met: NDArray) -> tuple[NDArray, NDArray, NDArray, N
     spin_period = spin_df.loc[spin_df.spin_number.isin(spins), "spin_period_sec"]
     spin_starttime = spin_df.loc[spin_df.spin_number.isin(spins), "spin_start_time"]
     spin_rates = 60 / spin_period  # 60 seconds in a minute
-    indices = (spin_rates > UltraConstants.CULLING_RPM_MIN) & (
-        spin_rates < UltraConstants.CULLING_RPM_MAX
+    indices = (spin_rates < UltraConstants.CULLING_RPM_MIN) & (
+        spin_rates > UltraConstants.CULLING_RPM_MAX
```
greglucas: Won't this always produce an empty result? Do you want to logical or them together instead? Previously your indices were the "good" ones between the MIN and MAX, and now I think they are the AND of less than min and more than max which won't be hit?

laspsandoval:
I realized I need to go back and fix attributes. Please hold off on reviewing until I finish that part.

greglucas:
> Please hold off on reviewing until I finish that part.

:+1: OK, I moved this to draft status now. I also don't think we should be getting too hung up on attributes, those are such a rabbit hole I think they can be separate follow-up PRs to clean things up as necessary.

imap_processing/ultra/l1b/cullingmask.py
laspsandoval: Yes. You are right. They shouldn't be thrown out. I was back and forth on this one, but you make a good point about exposure times.

imap_processing/ultra/l1b/badtimes.py
```diff
@@ -25,14 +32,15 @@ def calculate_badtimes(
     badtimes_dataset : xarray.Dataset
         Dataset containing the data.
     """
-    badtimes_dict = {}
-    badtimes_dict["spin_number"] = extendedspin_dataset["spin_number"]
-    badtimes_dict["energy_bin_geometric_mean"] = extendedspin_dataset[
-        "energy_bin_geometric_mean"
-    ]
+    spins = np.setdiff1d(extendedspin_dataset["spin_number"].values, cullingmask_spins)
```
nkerman: spins -> culled_spins?

imap_processing/ultra/l1b/badtimes.py
```diff
@@ -25,14 +32,15 @@ def calculate_badtimes(
     badtimes_dataset : xarray.Dataset
         Dataset containing the data.
```
nkerman: Can you add some more description to this?

imap_processing/ultra/l1b/cullingmask.py
```diff
@@ -25,14 +26,31 @@ def calculate_cullingmask(
...
-    cullingmask_dict["spin_number"] = extendedspin_dataset["spin_number"]
-    cullingmask_dict["energy_bin_geometric_mean"] = extendedspin_dataset[
-        "energy_bin_geometric_mean"
-    ]
-
-    # TODO: add more data to cullingmask_dict.
-
-    cullingmask_dataset = create_dataset(cullingmask_dict, name, "l1b", data_version)
+    # If the spin rate was too high or low then the spin should be thrown out.
+    # If the rates at any energy level are too high then throw out the entire spin.
```
nkerman, line 30: > `# If the rates at any energy level are too high then throw out the entire spin.`

throw out the entire spin at that energy level? Or all levels?

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -103,8 +103,8 @@ def flag_attitude(eventtimes_met: NDArray) -> tuple[NDArray, NDArray, NDArray, N
     spin_period = spin_df.loc[spin_df.spin_number.isin(spins), "spin_period_sec"]
     spin_starttime = spin_df.loc[spin_df.spin_number.isin(spins), "spin_start_time"]
     spin_rates = 60 / spin_period  # 60 seconds in a minute
-    indices = (spin_rates > UltraConstants.CULLING_RPM_MIN) & (
-        spin_rates < UltraConstants.CULLING_RPM_MAX
+    indices = (spin_rates < UltraConstants.CULLING_RPM_MIN) | (
```
nkerman: indices -> bad_spin_rate_indices ?

imap_processing/ultra/l1b/cullingmask.py
```diff
@@ -25,14 +26,31 @@ def calculate_cullingmask(
     cullingmask_dataset : xarray.Dataset
         Dataset containing the data.
```
nkerman: Again, I think it would be good to add more explanation here to the Dataset description.

imap_processing/ultra/l1b/badtimes.py
```diff
@@ -25,14 +32,15 @@ def calculate_badtimes(
...
+    spins = np.setdiff1d(extendedspin_dataset["spin_number"].values, cullingmask_spins)
 
-    # TODO: add more data to badtimes_dict.
+    filtered_dataset = extendedspin_dataset.sel(spin_number=spins)
 
-    badtimes_dataset = create_dataset(badtimes_dict, name, "l1b", data_version)
+    dataset_dict = {
+        **{var: filtered_dataset[var].values for var in filtered_dataset.data_vars},
+        **{coord: filtered_dataset[coord].values for coord in filtered_dataset.coords},
+    }
```
greglucas: Two other possibilities here that might simplify things.
1. Can you pass in `filtered_dataset` directly below? `create_dataset(filtered_dataset)`, datasets are mappings so should be usable in a dict-like sense.
2. You might be able to do this all in one go using `variables`: `{var: filtered_dataset[var] for var in filtered_dataset.variables}` https://docs.xarray.dev/en/stable/generated/xarray.Dataset.variables.html

imap_processing/ultra/l1b/cullingmask.py
```diff
@@ -25,14 +26,31 @@ def calculate_cullingmask(
...
-    cullingmask_dict["energy_bin_geometric_mean"] = extendedspin_dataset[
-        "energy_bin_geometric_mean"
-    ]
-
-    # TODO: add more data to cullingmask_dict.
-
-    cullingmask_dataset = create_dataset(cullingmask_dict, name, "l1b", data_version)
+    # If the spin rate was too high or low then the spin should be thrown out.
+    # If the rates at any energy level are too high then throw out the entire spin.
+    mask = (
```
greglucas: Suggest putting "good" or something similar to indicate this is the good values you are selecting (not the bad values which I think of as masked).
```suggestion
    good_mask = (
```

greglucas (approved):
Just some minor nits (in addition to Nat's). This looks good to me and is simpler to follow now too IMO. Thanks for updating!

imap_processing/ultra/l1b/cullingmask.py
laspsandoval: At all levels. We only used a single flag which was "HIGHRATES". 

In this dataset, we cannot filter on a specific spin/energy combination—filtering (or culling) is only done at the spin level.

If we wanted to make decisions on a per-energy-bin basis, you would need flags for each energy level independently.

PR #1361, opened 2025-02-13
Updates/improvements to CoDICE L0 and L1a hskp testing/validation

### LINKED ISSUE #1154: CoDICE: Update how L1a housekeeping is validated
Currently we validate housekeeping data by comparing the input values to a `.csv` file with engineering values. However, we are now receiving validation data in the form of CDFs (which is consistent with how validation data are retrieve for other data products). The L1a unit tests that make these comparisons should be modified to handle the new CDFs.

bourque:
This PR makes some changes to L0 testing, as well as L1a housekeeping testing. 

The housekeeping tests were originally set up before validation data were acquired, and validated the data by comparing values to CSV files. But now that L1a housekeeping CDFs have been acquired, we can now validate the housekeeping data similar to how the other L1a data products are being validated.

The L0 tests have also been updated to test that all expected APIDs are decommed successfully and contain the expected number of packets.

Closes #1154


imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -237,6 +241,53 @@ def test_l1a_validate_data_arrays(test_l1a_data: xr.Dataset, index):
...
+        "spare_2",
+        "spare_3",
+        "spare_4",
+        "spare_5",
+        "spare_6",
+        "spare_62",
+        "spare_68",
+    ]
+
+    # These variables are named differently between the decommed test_data and
```
greglucas: Would it make sense to do a mapping for now instead of ignoring these?

greglucas (approved):

imap_processing/tests/codice/test_codice_l1a.py
bourque: Actually I was mistaken here. These variables are simply not present in the validation data. I updated the comment here to reflect this and will ask Joey about it.

PR #1364, opened 2025-02-13
Add fill_value keyword to dataarray creation utilities

### LINKED ISSUE #1362: Hi Utils - Add optional fill_value keyword to dataarray creation functions
### Algorithm Description:

I found the need for being able to set the full value to something other than the FILLVAL from the CDF attribute manager.

tmplummer:
# Change Summary

## Overview
This is a short one. I found a need to add a fill_value argument to the utility function I use for creating xr.DataArrays.

## Updated Files
- imap_processing/hi/utils.py
   - Add fill_value keyword argument to two utility funcitons
   - Test that filled DataArrays have expected values.

Closes: #1362

### DIFF
--- a/imap_processing/hi/utils.py
+++ b/imap_processing/hi/utils.py
@@ -4,6 +4,7 @@ import re
 from collections.abc import Sequence
 from dataclasses import dataclass
 from enum import IntEnum
+from numbers import Number
 from typing import Optional, Union
 
 import numpy as np
@@ -99,6 +100,7 @@ def full_dataarray(
     attrs: dict,
     coords: Optional[dict[str, xr.DataArray]] = None,
     shape: Optional[Union[int, Sequence[int]]] = None,
+    fill_value: Optional[Number] = None,
 ) -> xr.DataArray:
     """
     Generate an empty xarray.DataArray with appropriate attributes.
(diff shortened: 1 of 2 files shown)


greglucas (approved):
I've seen we use a lot of different fill_values across the project and I'm wondering if we could somehow align all of these with something similar to what you're doing here. Like a default `get_fill_value(dtype)` that we try to keep consistent across the project. Also, I asked on slack, but do you know if "nan" is a valid fill value for floats?

tmplummer:
@greglucas 
> I've seen we use a lot of different fill_values across the project and I'm wondering if we could somehow align all of these with something similar to what you're doing here. Like a default `get_fill_value(dtype)` that we try to keep consistent across the project. Also, I asked on slack, but do you know if "nan" is a valid fill value for floats?

Yeah, I saw that slack thread (can't find it now). I found the following in the cdflib code for getting the ISTP standard fill values: https://github.com/MAVENSDC/cdflib/blob/91e8ee651bab1ffd4bc2ce6aff86905f5164337d/cdflib/xarray/xarray_to_cdf.py#L54
Maybe we should be using that to standardize?
I can't definitively find if NAN is ISTP compliant. I sure hope so event though it is not a "standard" fill value. I will also note that I found a function in cdflib that will convert FILLVAL to NaN on reading CDFs but not the other way around.
https://github.com/MAVENSDC/cdflib/blob/91e8ee651bab1ffd4bc2ce6aff86905f5164337d/cdflib/xarray/cdf_to_xarray.py#L286

PR #1367, opened 2025-02-14
MAG new tests and a bugfix for range

maxinelasp:
# Change Summary
I added some additional validation tests for L1A, and found a bug. This PR adds the tests and fixes the bug. 

## Overview

- Includes some time fixes to go through and properly use SPICE
- bug fix mostly involves moving lines of code around, and also accounts for an optional 8-bit padding at the end of the vectors. 
- Includes a bunch of new validation data from MAG 
## Testing

Skipping all the L1B tests because those are still failing for an unknown reason.


imap_processing/tests/mag/test_mag_l1b.py
```diff
@@ -60,7 +63,12 @@ def test_mag_processing(mag_l1a_dataset):
     np.testing.assert_allclose(
         mag_l1b["vectors"][0].values, [2.2972, 2.2415, 2.2381, 0], atol=1e-4
     )
-    np.testing.assert_allclose(mag_l1b["vectors"][1].values, [0, 0, 0, 0])
+    np.testing.assert_allclose(
```
nkerman, line 66: Maybe have the test code calculate these explicitly, or just add a comment for how they're calculated?

imap_processing/tests/mag/test_mag_l1b.py
```diff
@@ -163,6 +185,19 @@ def test_calibrate_vector():
...
+    assert np.allclose(cal_vector, expected_vector, atol=1e-9)
+
+    cal_vector = calibrate_vector(np.array([1.1, -2.0, 3.0, 1]), calibration_matrix)
+    expected_vector = np.array([(0.081202, -0.144636, 0.217628, 1)])
+    assert np.allclose(cal_vector, expected_vector, atol=1e-9)
+
+    cal_vector = calibrate_vector(
+        rescale_vector(np.array([7982, 48671, -68090, 0]), (1, 18)), calibration_matrix
+    )
+    print(f"HERE IS MY VECTOR: {cal_vector}")
```
nkerman: Did this print sneak through?

imap_processing/tests/mag/test_mag_validation.py
```diff
@@ -0,0 +1,151 @@
...
+    mag_l1a_out = mag_l1a(input_file, "v000")
+    expected_output = pd.read_csv(expected_output_file)
+
+    raw = mag_l1a_out[0]
+    primary = mag_l1a_out[1]
+    secondary = mag_l1a_out[2]
+
+    assert len(mag_l1a_out) == 3
+    print(primary["vectors"].data)
+    print(expected_output)
```
nkerman: More prints here - are we okay with print statements in the tests? I usually assume it should be logger statements, but I don't really know if one or another is right.

imap_processing/mag/l1b/mag_l1b.py
```diff
@@ -137,6 +143,10 @@ def update_vector(
     tuple[numpy.ndarray, numpy.ndarray]
         Updated vector and the same compression flags.
     """
+    vector = rescale_vector(input_vector, input_compression)
+    cal_vector = calibrate_vector(vector, calibration_matrix)
+    return cal_vector, input_compression
+
     vector = calibrate_vector(input_vector, calibration_matrix)
     return rescale_vector(vector, input_compression), input_compression
```
nkerman: I'm not sure what the deal is with the 2nd return statement here - is this an outdated return?

maxinelasp: Yes good catch! Will remove!

imap_processing/tests/mag/test_mag_validation.py
maxinelasp: No, we aren't ok with print statements, I just failed to push the PR removing them. Will remove

maxinelasp:
@nkerman I'd appreciate another review when you get a chance! I addressed your issues :)

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -809,20 +826,34 @@ def _process_vector_section(  # noqa: PLR0913
...
+        (numpy.ndarray, int)
+            A tuple consisting of: an array of processed vectors, and the index of the
+            end of the last vector. In a fully compressed case, this will be the same
+            as last_index.
         """
+        compressed_count = math.ceil(len(split_bits) / AXIS_COUNT) + 1
+        uncompressed_count = vector_count - compressed_count
+
+        # will have either 0 or 8 extra bits
+        # what if there is padding at the end of the vectors? keep that in mind
```
nkerman: Is this something that's still up in the air upstream/at the packet level? I worry a bit about it getting lost without a ticket or TODO: comment.

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -757,21 +767,28 @@ def process_compressed_vectors(
...
-                    + (primary_count + secondary_count - 2) * 2
+                range_bits[
+                    (primary_count - 1) * 2 : (primary_count + secondary_count - 2) * 2
                 ],
                 secondary_vectors,
             )
+
         return primary_vectors, secondary_vectors
 
     @staticmethod
```
nkerman, line 794: In recent PRs, I've been advised to reduce the number of inputs to functions. It's not something I feel strongly about, but perhaps consider whether all of these are necessary, or if any can be combined. 

In this case, it might not make sense, because of the compression going on.

imap_processing/tests/mag/test_mag.py
nkerman, line 1: I'm not sure why this file would be separate from the L1a, L1b, etc. test files? Is it about the interface between levels? It seems like it might belong in the L1b test file. 

I might be missing something, in which case, could you maybe add some clarity to the docstring?

nkerman (commented):
Looks good, Maxine! I added a couple requests for docs clarity, and a suggestion to look into reducing the number of arguments to one function. Otherwise, I think it's in great shape.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: Ah no this is resolved at this point, I have tests for it and it works. I'll remove that comment.

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: This function is already pretty terse and specific, so I'd prefer to leave it as all the information is needed. It doesn't make sense to combine any of these variables together. I do keep in mind the number of inputs, but there's a tradeoff between being clean and being clear and here I think being clear is worth it.

imap_processing/tests/mag/test_mag.py
maxinelasp: Good point, I had a vision for this file that didn't end up panning out. I moved this test and deleted the file.

nkerman (approved):
LGTM! I'd like to chat sometime about the general path of the MAG pipeline, and better understand how the instrument works. But as far as I can tell, this code all makes sense.

PR #1368, opened 2025-02-14
frame_transform checks for same frame

### LINKED ISSUE #1370: Geometry Frame Transform: Check if input frame == output frame before applying transform
Check if the input frame and output frame are the same to avoid unnecessary Spice calls. Doing it in `frame_transform` should catch it in subsequent functions- e.g. `frame_transform_az_el`

nkerman:
# Change Summary

## Overview
Small change suggested by @greglucas: Check that the frames are not the same before applying frame_transform(). If they are the same, return vector without Spice call 

Closes #1370

## Updated Files

- geometry.py
   - frame_transform() checks if input/output are the same frame before using spice to generate rotation matrix. 

## Testing

Add tests for multiple frames, multiple input shapes. Test that frame_transform_az_el() still works as well.

### DIFF
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -189,6 +189,10 @@ def frame_transform(
     result : np.ndarray
         3d Cartesian position vector(s) in reference frame `to_frame`.
     """
+    # If from_frame and to_frame are the same, no rotation needed
+    if from_frame == to_frame:
+        return position
+
     if position.ndim == 1:
         if not len(position) == 3:
             raise ValueError(
@@ -208,10 +212,6 @@ def frame_transform(
                     f"Position has {len(position)} elements and et has "
                     f"{np.asarray(et).size} elements."
                 )
-
-    # If from_frame and to_frame are the same, no rotation needed
-    if from_frame == to_frame:
-        return position
 
     # rotate will have shape = (3, 3) or (n, 3, 3)
     # position will have shape = (3,) or (n, 3)
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -209,6 +209,10 @@ def frame_transform(
                     f"{np.asarray(et).size} elements."
                 )
 
+    # If from_frame and to_frame are the same, no rotation needed
+    if from_frame == to_frame:
+        return position
+
     # rotate will have shape = (3, 3) or (n, 3, 3)
     # position will have shape = (3,) or (n, 3)
     rotate = get_rotation_matrix(et, from_frame, to_frame)
--- a/imap_processing/tests/spice/test_geometry.py
+++ b/imap_processing/tests/spice/test_geometry.py
@@ -169,9 +169,8 @@ def test_frame_transform(et_strings, position, from_frame, to_frame, furnish_ker
 )
 def test_frame_transform_same_frame(position, spice_frame):
     """Test that frame_transform returns position when input/output frames are same."""
-    position = np.array([1, 0, 0])
     result = frame_transform(0, position, spice_frame, spice_frame)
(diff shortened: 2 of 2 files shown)


imap_processing/spice/geometry.py
```diff
@@ -209,6 +209,10 @@ def frame_transform(
                     f"{np.asarray(et).size} elements."
                 )
 
+    # If from_frame and to_frame are the same, no rotation needed
```
greglucas: Do we even need to do all of the checks above in this caes? (Move this as the very first check when entering the function)

imap_processing/tests/spice/test_geometry.py
```diff
@@ -151,6 +151,29 @@ def test_frame_transform(et_strings, position, from_frame, to_frame, furnish_ker
...
+    "position",
+    [
+        np.array([1, 0, 0]),
+        np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]]),
+        np.random.rand(10, 3),
+    ],
+)
+def test_frame_transform_same_frame(position, spice_frame):
+    """Test that frame_transform returns position when input/output frames are same."""
+    position = np.array([1, 0, 0])
```
greglucas: You are parametrizing over this above.
```suggestion
```

imap_processing/tests/spice/test_geometry.py
```diff
@@ -151,6 +151,29 @@ def test_frame_transform(et_strings, position, from_frame, to_frame, furnish_ker
...
+    result = frame_transform(0, position, spice_frame, spice_frame)
+    np.testing.assert_array_equal(result, position)
```
greglucas: If you want this to be a true no-op, this is where an `is` check would be good to make sure they are the same object.
```suggestion
    assert result is position
```

imap_processing/tests/spice/test_geometry.py
```diff
@@ -222,6 +245,45 @@ def test_frame_transform_az_el(mock_get_rotation_matrix, az_range, el_range):
...
+@pytest.mark.parametrize(
+    "spice_frame",
+    [
+        SpiceFrame.IMAP_DPS,
+        SpiceFrame.IMAP_SPACECRAFT,
+        SpiceFrame.ECLIPJ2000,
+    ],
+)
+@pytest.mark.parametrize("degrees_bool", [True, False])
+def test_frame_transform_az_el_same_frame(spice_frame, degrees_bool):
```
greglucas, line 212: Since this calls out to the other method, I think this is testing the same thing as your other one so isn't needed? Your other test is great!

(I think I didn't quite realize what `degrees_bool` did before in my earlier comment and it doesn't actually apply a deg2rad transform between input/output, it just assumes both input/output are one or the other. i.e. I thought it would accept degrees input always and then transform to radians on the output.)

greglucas (approved):
Thanks for breaking this up into a small self-contained PR. This is very nice to review :)

Looks good to me, some minor suggestions that don't affect the content.

imap_processing/spice/geometry.py
nkerman: Good point - done

imap_processing/tests/spice/test_geometry.py
nkerman: Thanks!

imap_processing/tests/spice/test_geometry.py
nkerman: Ah, that makes sense. Thanks!

imap_processing/tests/spice/test_geometry.py
nkerman: I removed this test. Thanks!

PR #1376, opened 2025-02-18
MAG L1B

maxinelasp:
# Change Summary
Additional validation tests and bugfixes.

## Overview

Testing updates are described below.
Main changes are:
- There was a bug when reading Logical_source and determining if the file was a mago or magi file, where previously the Logical_source was processed into a list and I think perhaps a cdflib update made it so it was processed into a string. This fixes the bug and adds some extra guardrails so it won't happen again. I also added a test to check that either option works. 
- I also had a bug where I was applying a scale factor to the range (4th value in the vector) which should not happen. I explicitly defined the types more, fixed the bug, and added another check to throw an error if the range is not an integer. 

## Testing

- New version of validation tests
- Added some timestamp tests
- Created a new generator to create xarray datasets
- Moved to using specific test data for L1B explicitly so only L1B is being tested, rather than running through L1A first.

Reviewer note: I did pair-program a lot of these changes (>90%) with Alastair, so they have already been vetted by him.


imap_processing/tests/mag/test_mag_l1b.py
```diff
@@ -11,53 +11,15 @@
...
-    output_dataset["compression_label"] = compression_label
-    output_dataset.attrs["Logical_source"] = ["imap_mag_l1a_norm-mago"]
-
-    return output_dataset
-
-
-def test_mag_processing(mag_l1a_dataset):
-    mag_l1a_dataset.attrs["Logical_source"] = ["imap_mag_l1a_norm-mago"]
+    mag_l1a_dataset["vectors"].data[0, :] = np.array([1, 1, 1, 0])
+    mag_l1a_dataset["vectors"].data[1, :] = np.array([7982, 48671, -68090, 0])
```
tmplummer, line 25: These look like pretty specific values. You might want to comment about where they came from for future reference.

tmplummer (approved):
This all looks good to me.

PR #1379, opened 2025-02-19
SWE: L2 update from SWE and I-ALiRT Discussion and added SPASE metadata

### LINKED ISSUE #915: Create SWE L2 CDF attrs and file

### LINKED ISSUE #1360: SWE: Minor L2 improvement
- [x] Move constant into `swe_utils` which will then become instrument status summary or something.
- [x] We can calculate center time in L1B but keep L1A raw. In L1B, we calculate center time of acquisition time.
- [x] Center time for epoch should be start time of third quarter cycle and combine acquisition time coarse and fine. If project wants in second then we can use shcoarse but if we want in microseconds or finer, then we should use acquisition time coarse and fine.
- [x] Can instrument polar angle be from SPICE? No. Ruth wants to give this from ground calibration. Ruth wants to do this because the angle will depend on shape of the peak. Ground calibration data is used to calculate this angle.
- [x] For the future, I wonder if we should make these global constants as well? You use the 24, 30, 7 values a lot throughout the code. N_ESA_STEPS, N_ANGLES, N_CEMS
- [x] Is this equation correct? I think you want to remove settle_duration too.

      `center_of_acquisition = acquisition_start + settle_time + acquisition_time / 2`
Answer: No. This commit has the correct formula but need to correct millisecond to microseconds. https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1357/commits/d0d4166cf990af9a924c8284ccaf38190da58a7f
- [ ]

tech3371:
# Change Summary
closes #1360 

## Overview
This PR refactors and moves constant into utils code. Then refactor some time calculation for I-ALiRT reusability. Ruth and I added SPASE metadata for L2 product.

## Updated Files
- imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
   - Ruth and I walked through L2 CDF attrs and updated per her guidance
- imap_processing/swe/l1a/swe_science.py
   - update to use constant variables
   - updates to use constant variables
   - refactor of time calculations
   - update to calculate center time for both epoch and each point data's time
   - removed time calculation and update to use constant variables
   - refactored time functions

### DIFF
--- a/imap_processing/swe/l2/swe_l2.py
+++ b/imap_processing/swe/l2/swe_l2.py
@@ -17,7 +17,6 @@ from imap_processing.swe.utils.swe_utils import (
     ESA_VOLTAGE_ROW_INDEX_DICT,
     FLUX_CONVERSION_FACTOR,
     GEOMETRIC_FACTORS,
-    MICROSECONDS_IN_SECOND,
     N_ANGLE_BINS,
     N_CEMS,
     N_ESA_STEPS,
@@ -413,28 +412,13 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
 
     # Calculate spin phase using SWE acquisition_time from the
     # L1B dataset. The L1B dataset stores acquisition_time with
-    # dimensions (epoch, esa_step, spin_sector). Use center time
-    # to calculate spin phase. This center time calculation is
-    # necessary to accurately determine the center angle of the data.
-    #
-    # To determine the center acquisition time, we adjust the
-    # recorded acquisition_time as follows:
-    #   acquisition_time + (acq_duration / 1000000) / 2
-    #
-    # Here, acq_duration is given in microseconds and is stored
-    # in the L1B dataset with dimensions (epoch, cycle). Since acq_duration
-    # remains the same for all quarter cycles within a full sweep,
-    # we use the first acq_duration value for each full sweep to perform
-    # this adjustment.
-
-    acq_duration = l1b_dataset["acq_duration"].data[:, 0] / (2 * MICROSECONDS_IN_SECOND)
-    data_acq_time = (
-        l1b_dataset["acquisition_time"].data + acq_duration[:, np.newaxis, np.newaxis]
-    )
+    # dimensions (epoch, esa_step, spin_sector). acquisition_time is
+    # center time of acquisition time of each science measurement which
+    # is necessary to accurately determine the center angle of the data.
 
     # Calculate spin phase
     inst_spin_phase = get_instrument_spin_phase(
-        query_met_times=data_acq_time.ravel(),
+        query_met_times=l1b_dataset["acquisition_time"].data.ravel(),
         instrument=SpiceFrame.IMAP_SWE,
     )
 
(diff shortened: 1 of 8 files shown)


imap_processing/swe/utils/swe_utils.py
```diff
@@ -2,10 +2,44 @@
...
 from enum import IntEnum
 
+import numpy as np
+import numpy.typing as npt
 import pandas as pd
 
 from imap_processing import imap_module_directory
 
+N_ESA_STEPS = 24
+N_MEASUREMENTS = 30
```
greglucas: Is there another word for this? Is this `N_ANGLES`, or the same as `N_ANGLE_BINS` below? I think every value in the array is a measurement, so perhaps something else for the dimension would be more descriptive.

imap_processing/swe/utils/swe_utils.py
```diff
@@ -2,10 +2,44 @@
...
+        432e-6,
+    ]
+)
+
+ELECTRON_MASS = 9.10938356e-31  # kg
+
+# See doc string of calculate_phase_space_density() for more details.
+VELOCITY_CONVERSION_FACTOR = 1.237e31
+# See doc string of calculate_flux() for more details.
+FLUX_CONVERSION_FACTOR = 6.187e30
```
greglucas: Can you add what units these are going from / to?

imap_processing/swe/l2/swe_l2.py
```diff
@@ -12,33 +12,19 @@
 from imap_processing.spice.geometry import SpiceFrame
 from imap_processing.spice.spin import get_instrument_spin_phase, get_spin_angle
 from imap_processing.swe.utils.swe_utils import (
+    CEM_DETECTORS_ANGLE,
```
greglucas: There are a lot of singular imports here. A few minor suggestions would be to import the module `swe_utils` and then use `swe_utils.ENERGY_CONVERSION_FACTOR` and possibly make a separate `swe_constants` file for the constants to clearly call them out. Just thoughts/suggestions, no need to implement these.

greglucas (approved):
I really like the global constants approach to the repetitive dimensions, it makes it easier to reason about the dimension ordering and what you're expecting where.

imap_processing/swe/utils/swe_utils.py
tech3371: Would `N_SPIN_SECTOR` work?

greglucas: Sure (with a plural s to match the others), but is it different than the N_ANGLE_BINS below which is also 30? I think we were trying to move away from "sector", but that was only for the higher levels.

tech3371: Yep. At lower level, it's angle sectors and at L2 it becomes angle bins. They both are 30 length array. `N_ANGLE_SECTORS` might be better.

imap_processing/swe/utils/swe_utils.py
tech3371: I plan to have another PR in the future to map these to algorithm document. I can make sure this happens in there.

imap_processing/swe/l2/swe_l2.py
tech3371: done!

PR #1382, opened 2025-02-24
IDEX l2a processing

### LINKED ISSUE #1377: Initial l2a code and tests

lacoak21:
Process IDEX l1b data to create l2a data products. 

## Overview
This will use the time of flight signals to estimate ion composition and the target and ion grid signals to estimate the dust mass. It will also calculate a time of flight to mass scale to interpret the time of flight results. 

closes #1377
## New Dependencies
Scipy

## New Files
- imap_processing/idex/idex_l2a.py
   - L2a processing code
- imap_processing/idex/idex_l2a.py
   - L2a processing code tests
- imap_processing/idex/atomic_masses.csv
   - Masses for time to mass function

## Updated Files
- docs/source/conf.py
   - Ignore int enum docs
- docs/source/code-documentation/idex.rst
   - Add l2a code to IDEX docs
- imap_processing/cdf/config/imap_idex_global_cdf_attrs.yaml
   - Add l2a global attrs
   - Move signal conversion factor enum to Idex constants
   - Add other constants used in fits and noise filtering 
   - Fix L1a time calculation 
    - Remove conversion factors from l1b to idex constants 
         
## Testing
Added 14 new unit tests for l2a processing


imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+    """
+    # Extract EMG fit parameters: mu, sigma, lam
+    mu, sigma, lam = param
+    # Compute integral
+    area, _ = quad(emg, time_slice[0], time_slice[-1], args=(mu, sigma, lam))
+
+    return float(area)
+
+
+def estimate_dust_mass(
```
lacoak21, line 524: Subject to change. I think the overall structure will stay the same. I struggled to vectorize this because the fits need to happen for each signal independent of the event. If anyone has any suggestions let me know.

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+
+    # Calculate chi square and reduced chi square
+    chisqr = float(np.sum((signal - impact_fit) ** 2))
+    # To get reduced chi square divide by dof (number of points - number of params)
+    redchi = chisqr / (len(signal) - len(p0))
+
+    return param, float(sig_amp), chisqr, redchi, impact_fit
+
+
+def fit_impact(
```
lacoak21, line 627: Subject to change.

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+    time_of_impact = 0.0  # Time of dust hit
+    constant_offset = 0.0  # Initial baseline
+    amplitude: float = np.max(signal)  # Signal height
+    rise_time = 0.371  # How fast the signal rises (s)
+    discharge_time = 0.371  # How fast signal decays (s)
+
+    p0 = [time_of_impact, constant_offset, amplitude, rise_time, discharge_time]
+
+    try:
+        param, _ = curve_fit(
```
lacoak21: The IDEX team decided recently to use lmfit.minimize() instead of scipy.curve_fit here ( and for the emg fit in analyze_peaks) because it calculates the chi-square value and returns it from lmfit.minimize(). I kept curve_fit here because lmift uses scipy.curve_fit under the hood anyway but changes some of the default parameters.  I manually calculated the chi-square and chi-square reduced values.

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,184 @@
...
+        f"Running IDEX L2A processing on dataset: {l1b_dataset.attrs['Logical_source']}"
+    )
+
+    tof_high = l1b_dataset["TOF_High"]
+    hs_time = l1b_dataset["time_high_sr"]
+
+    # Load an array of known masses of ions
+    atomic_masses_path = f"{imap_module_directory}/idex/atomic_masses.csv"
+    atomic_masses = pd.read_csv(atomic_masses_path)
+    masses = np.round(atomic_masses["Mass"], 1)
```
greglucas: Why put multiple extra decimal places in the file and then round it off here?

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,184 @@
...
+        Estimated mass for each time per event (after the time has been aligned using
+        the best t_offset and stretch_factor).
+    """
+    # Create an array of random stretches
+    # eventually, the stretch_factor used to create the highest correlation is used to
+    # align the time
+    min_stretch = 1400
+    random_stretches = np.linspace(min_stretch, min_stretch + 100, 10)
+
+    time = high_sampling_time - high_sampling_time[:, 0:1]
```
greglucas, line 225: A quick comment explaining what this is doing (non-obvious to me)

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,184 @@
...
+
+    # Start with a time offset of 0
+    t_offset = 0
+    shift = np.zeros((len(random_stretches), len(tof_high)))
+    corr = np.zeros_like(shift)
+    for i in range(len(random_stretches)):
+        # Step 1
+        t_i = np.zeros(len(tof_high[0]))
+        # Step 2
+        t_calc = t_offset + random_stretches[i] * np.sqrt(masses)
```
greglucas: I think these can be brought outside the loops and then indexed into within the loops.

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,184 @@
...
+        # Loop through the elements of t_calc and set corresponding elements in t_i to 1
+        for idx in np.round(t_calc).astype(int):
+            if 0 <= idx < len(t_i):
+                t_i[idx] = 1
```
greglucas: Could this be vectorized to avoid the for loop? Something like comparing against the t_calc values to fill the indices.

t_i[np.arange(len(t_calc)) < t_calc] = 1

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,184 @@
...
+            if np.all(cross_correlation == 0):
+                logger.warning(
+                    "There are no correlations found between the TOF array "
+                    "and the expected mass times array. The resulting mass scale "
+                    "may be inaccurate."
+                )
+            # Find the lag corresponding to the maximum correlation
+            # Represents the time lag from where the arrays are most correlated
+            # Zero lag position
+            middle = len(t_i) - 1
```
greglucas: I'm not sure what "middle" means here, do you need to divide by 2?

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,184 @@
...
+        idex_constants.FM_SAMPLING_RATE
+        * shift[np.argmax(corr, axis=0), np.arange(len(shift[0]))]
+    )
+    # Get the best stretch in seconds
+    best_stretch = (
+        idex_constants.NS_TO_S_CONV_FACTOR * random_stretches[np.argmax(corr, axis=0)]
+    )
+
+    mass_scale = (
+        (time * idex_constants.MS_TO_S_CONV_FACTOR - best_shift[:, np.newaxis])
```
greglucas: Above you are calculating all of your shifts/stretches etc based on the unconverted time. Do we need to convert the time earlier or do the multiplication here after the shift?

imap_processing/tests/idex/test_idex_l2a.py
```diff
@@ -0,0 +1,113 @@
...
+    # Mass 4 expected tof
+    tof[-1, expected_stretch * 2 + expected_lag] = 1
+    # Mass 9 expected tof
+    tof[-1, expected_stretch * 3 + expected_lag] = 1
+
+    time = np.tile(np.arange(len(tof[0])), (15, 1))
+    stretch, shift, mass_scale = time_to_mass(tof, time, masses)
+
+    # Test with carbon mass
+    carbon_time = (stretch[0] * np.sqrt(carbon_mass)) / 1e-6  # Convert ms to s
```
greglucas, line 136: ms to s is 1e3?

imap_processing/idex/idex_l2a.py
```diff
@@ -182,3 +188,30 @@ def time_to_mass(
     ) ** 2
 
     return best_stretch, best_shift, mass_scale
+
+
+def calculate_kappa(mass_scales: np.ndarray, peaks_2d: list) -> NDArray:
+    """
+    Calculate the kappa value for each peak.
```
greglucas: What is the "kappa" value? Might help to reference/link to somewhere or indicate best-fit parameter or something more descriptive.

pyproject.toml
```diff
@@ -33,7 +33,7 @@ classifiers = [
 [tool.poetry.dependencies]
 cdflib = "^1.3.1"
 imap-data-access = ">=0.10.1"
-python = ">=3.9,<4"
+python = ">=3.10,<4"
```
greglucas, line 35: Do we need to bump Python version? We have had requests for Python 3.7 support even 😂

imap_processing/idex/idex_l2a.py
```diff
@@ -215,3 +234,42 @@ def calculate_kappa(mass_scales: np.ndarray, peaks_2d: list) -> NDArray:
...
+    numpy.ndarray
+        Signal-to-noise ratio at each event.
+    """
+    # Find indices where Time (High Sampling) is between -7 and -5 ns (no signal yet)
+    # To determine the baseline noise
+    baseline_noise = np.where(
+        np.logical_and(
+            hs_time >= BaselineNoiseTime.START, hs_time <= BaselineNoiseTime.STOP
+        ),
+        tof_high.data,
```
greglucas: Do you need the `.data` in all of these calls? I thought xarray might be able to be coerced to numpy so you don't have to explicitly access the data attribute.

imap_processing/idex/idex_l2a.py
```diff
@@ -215,3 +234,42 @@ def calculate_kappa(mass_scales: np.ndarray, peaks_2d: list) -> NDArray:
...
+        np.nan,
+    )
+    if np.all(np.isnan(baseline_noise)):
+        logger.warning(
+            "Unable to find baseline noise. "
+            f"There is no signal from {BaselineNoiseTime.START} to "
+            f"{BaselineNoiseTime.STOP} ns. Returning np.nan SNR values"
```
greglucas, line 348: You are returning zeros right below this, do you mean to return nan instead?

imap_processing/idex/idex_l2a.py
```diff
@@ -215,3 +234,42 @@ def calculate_kappa(mass_scales: np.ndarray, peaks_2d: list) -> NDArray:
...
+        The high sampling time array.
+
+    Returns
+    -------
+    numpy.ndarray
+        Signal-to-noise ratio at each event.
+    """
+    # Find indices where Time (High Sampling) is between -7 and -5 ns (no signal yet)
+    # To determine the baseline noise
+    baseline_noise = np.where(
```
greglucas, line 337: Could you get rid of the nan handling below by subsetting this array directly instead of setting values to nan's

```python
baseline_noise = tof_high[(hs_time >= BaselineNoiseTime.START) & (hs_time <= BaselineNoiseTime.STOP)]
```

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+    1) Make a vector with all zeros and a length of 8189, same as the TOF length: t_i
+    2) Calculate the times when each input mass should appear in the TOF data: t_calc
+        for each mass, calculate a time using this formula:
+
+            t_calc = t_offset + stretch_factor*sqrt(mass)
+
+            t_offset is the time offset (ns)
+            stretch factor (ns)
+
+        Then and set the value at the index of t_i that is closest to each of the
```
tech3371, line 190: ?
```suggestion
        Then set the value at the index of t_i that is closest to each of the
```

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+        Best shift value per event (shifts scale left or right).
+    numpy.ndarray
+        Estimated mass for each time per event (after the time has been aligned using
+        the best t_offset and stretch_factor).
+    """
+    # Create an array of random stretches
+    # eventually, the stretch_factor used to create the highest correlation is used to
+    # align the time
+    min_stretch = 1400
+    random_stretches = np.linspace(min_stretch, min_stretch + 100, 10)
```
tech3371, line 221: same here. Some comment in this function would be helpful.

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+
+    time = high_sampling_time - high_sampling_time[:, 0:1]
+
+    # Start with a time offset of 0
+    t_offset = 0
+    shift = np.zeros((len(random_stretches), len(tof_high)))
+    corr = np.zeros_like(shift)
```
tech3371: What does this `corr` mean, correlation or correction. May be safer to spell this out.

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+    4) Choose the stretch_factor that has the highest correlation
+
+    Parameters
+    ----------
+    tof_high : numpy.ndarray
+        The time of flight array for one dust event.
+    high_sampling_time : numpy.ndarray
+        The high sampling time array for one dust event.
+    masses : np.ndarray
+        Array of known masses of ions.
```
tech3371: What are shape of these input arrays? It helps me to document that and know what each dimension means to understand later lines of code. I am wondering if you can do same here.

imap_processing/idex/idex_l2a.py
```diff
@@ -273,3 +299,176 @@ def calculate_snr(tof_high: xr.DataArray, hs_time: xr.DataArray) -> NDArray:
...
+    # array.
+    fit_params = np.zeros((500, 3))
+    area_under_emg = np.zeros(500)
+    for peak in peaks_2d[event_num]:
+        # Take a slice of 5 samples on either side of the peak
+        start = max(0, peak - 5)
+        end = min(len(tof_high), peak + 6)
+
+        time_slice = np.asarray(high_sampling_time[start:end].data)
+        tof_slice = np.asarray(tof_high[start:end].data)
```
greglucas: If these are DataArrays, you can slice into them directly via indexing or with `.iloc[start:end]` as well, and they should already be arrays so not need the `.asarray()` calls.

imap_processing/idex/idex_l2a.py
```diff
@@ -273,3 +299,176 @@ def calculate_snr(tof_high: xr.DataArray, hs_time: xr.DataArray) -> NDArray:
...
+
+        param = fit_emg(time_slice, tof_slice, event_num)
+        if param is not None:
```
greglucas: If you invert this, you can remove a layer of indentation.

```python
if param is None:
    continue
...
```

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+        ],
+        # TODO: Determine dimension name
+        output_core_dims=[
+            ["time_of_flight", "peak_fit_parameters"],
+            ["time_of_flight"],
+        ],
+        vectorize=True,
+    )
+
+    l2a_dataset = l1b_dataset.copy()
```
tech3371, line 128: Does IDEX wants to carry all L1B data to l2? because this copy copies all data variable from l1B. Since you are adding more variables below and I didn't see you drop any variables from l1b, I am wondering if that's what you wanted.

imap_processing/idex/idex_l2a.py
```diff
@@ -273,3 +299,176 @@ def calculate_snr(tof_high: xr.DataArray, hs_time: xr.DataArray) -> NDArray:
...
+            # close to the same flight time.
+            if emg_mu < 0:
+                logger.warning(
+                    f"Warning: The EMG fit resulted in a negative mu "
+                    f"(center of gaussian) value: {emg_mu}."
+                )
+
+            idx = max(0, round(emg_mu))
+            while np.all(fit_params[idx:] != 0) and idx < 500:
+                idx += 1
```
greglucas: I think you can find the next nonzero value with numpy rather than the loop. Something like this, (it is probably not exactly right)
`idx = np.nonzero(np.all(fit_params != 0, axis=-1))`

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,754 @@
...
+        dims=("epoch", "time_high_sr_dim"),
+    )
+    snr = calculate_snr(tof_high, hs_time)
+    # Find peaks for each event. The peaks represent a TOF of an ion.
+    # Peaks_2d is a list of variable-length arrays
+    peaks_2d = [find_peaks(tof, prominence=0.01)[0] for tof in tof_high]
+    kappa = calculate_kappa(mass_scales, peaks_2d)
+
+    # Analyze peaks for estimating dust composition
+    peak_fits, area_under_fits = xr.apply_ufunc(
```
tech3371, line 107: I don't fully understand `xr.apply_ufunc()` but looks like a power tool.

imap_processing/tests/idex/test_idex_l2a.py
```diff
@@ -165,3 +168,77 @@ def test_calculate_snr_warning(caplog):
         "Unable to find baseline noise" in message
         for message in caplog.text.splitlines()
     )
+
+
+@pytest.mark.filterwarnings("ignore:invalid")
+@pytest.mark.filterwarnings("ignore:overflow")
```
greglucas: In our normal processing are we going to get these invalid/overflows potentially? Do we want to ignore them internal to the functions (rather than just in the tests) using something like: 

```python
with np.errstate(invalid="ignore", overflow="ignore"):
    my_potential_overflow_calculation()
```

tech3371 (commented):
I don't understand it a lot from algorithm/science perspective but I was able to follow on a high level. Nice work!

imap_processing/tests/idex/test_idex_l2a.py
```diff
@@ -165,3 +168,77 @@ def test_calculate_snr_warning(caplog):
...
+    assert np.all(area_under_curve == 0)
+
+
+def test_emg():
+    """Tests that emg() calculates an expected exponentially modified gaussian"""
+    mu = 4
+    sigma = 2.0
+    lam = 1.0
+    time = xr.DataArray(np.arange(100))
+    # The scipy.stats.exponnorm function's location and scale parameters match directly
```
greglucas: Should we be using this in place of the one you wrote then? Just inverting the input parameters as you've shown here?

imap_processing/idex/idex_constants.py
```diff
@@ -25,3 +26,27 @@ class IdexConstants:
...
     SAMPLE_RATE_MAX: int = 130
+
+
+# FM sampling rate (quartz oscillator)
+# Seconds per sample.
+FM_SAMPLING_RATE = 0.0038466235767167234e-6
+# Nanoseconds to seconds conversion
+NS_TO_S_CONV_FACTOR = 1e-9
+# Microseconds to seconds conversion
+MS_TO_S_CONV_FACTOR = 1e-6
```
greglucas: Others are leaving off the `CONV_FACTOR` and just calling it `NS_TO_S`. I also wonder if we should put some of these constants into a global constants namespace rather than everyone redefining the same factors... Maybe for later.

imap_processing/idex/idex_constants.py
```diff
@@ -25,3 +26,27 @@ class IdexConstants:
...
+
+TARGET_HIGH_FREQUENCY_CUTOFF = 100  # / 1e6  # Hz
```
greglucas: Remove the / 1e6 if it is unused, otherwise it is a bit confusing.

imap_processing/idex/idex_constants.py
```diff
@@ -25,3 +26,27 @@ class IdexConstants:
...
+
+TARGET_NOISE_FREQUENCY = 7000  # / 1e6  # MHz
+
+
+class ConversionFactors(float, Enum):
+    """Enum class for conversion factor values."""
```
greglucas: Converting what to what? It would help to list units or expected from/to conversions.

imap_processing/idex/idex_l1a.py
greglucas, line 1: Was this wrong before? I'm surprised the l1 tests didn't fail with this update, maybe we need some more tests on l1... This seems like something that could be a few line standalone PR to get in quick too as an FYI, rather than buried in this longer PR.

imap_processing/idex/idex_l2a.py
```diff
@@ -123,6 +125,29 @@ def idex_l2a(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
+            output_core_dims=[["fit_parameters"], [], [], [], ["time_low_sr_dim"]],
+            vectorize=True,
+            output_dtypes=[np.float64] * 6,
+        )
+        waveform_name = waveform.lower()
+        # Add variables
+        l2a_dataset[f"{waveform_name}_fit_parameters"] = fit_results[0]
+        l2a_dataset[f"{waveform_name}_fit_imapct_charge"] = fit_results[1]
+        # TODO: convert charge to mass
+        l2a_dataset[f"{waveform_name}_fit_imapct_mass_estimate"] = fit_results[1]
```
greglucas, line 157: re-using the same index?

imap_processing/idex/idex_l2a.py
```diff
@@ -472,3 +497,134 @@ def calculate_area_under_emg(time_slice: np.ndarray, param: np.ndarray) -> float
...
+def fit_impact(
+    time: np.ndarray,
+    time_of_impact: float,
+    constant_offset: float,
+    amplitude: float,
+    rise_time: float,
+    discharge_time: float,
+) -> NDArray:
+    """
+    Fit function for the Ion Grid and two target signals given by (Horanyi, 2014).
```
greglucas: Nice reference! We should include the actual paper citation here too.

https://numpydoc.readthedocs.io/en/latest/format.html#references

imap_processing/tests/idex/test_idex_l2a.py
```diff
@@ -242,3 +244,40 @@ def test_analyze_peaks_perfect_fits():
...
         # Test that there is a value greater than zero at this index
         assert area_under_curve[idx] > 0
+
+
+def test_estimate_dust_mass_no_noise_removal():
+    """
+    Test that estimate_dust_mass() is fitting the signal properly when there is no
+    noise removal.
+    """
+    pass
```
greglucas: Turn this into a `pytest.xfail()` mark for now so it doesn't count as a "good" test.

imap_processing/idex/idex_l2a.py
```diff
@@ -537,8 +537,19 @@ def estimate_dust_mass(
     #         information soon.
     signal = np.array(target_signal.data)
     time = np.array(low_sampling_time.data)
+    mask = np.logical_and(
```
greglucas: I generally think of mask as "bad" values. I'd suggest calling this `good_mask`, `good_vals`, something like that or invert the logic to make the mask the bad values for consistency across the project.

imap_processing/idex/idex_l2a.py
```diff
@@ -628,3 +639,114 @@ def fit_impact(
...
+
+    Returns
+    -------
+    numpy.ndarray
+        Signal with linear, sine wave, and high frequency background noise filtered out.
+    """
+    # Remove linear noise
+    signal = detrend(signal, type="linear")
+    # Remove sine wave Background
+    baseline_delined = signal[mask]
```
greglucas: delined -> detrended

imap_processing/idex/idex_l2a.py
lacoak21: That is a great point 😄. I thought we might need them eventually but we dont.

pyproject.toml
lacoak21: Unfortunately scipy requires: Python >=3.10. Maybe we can get a previous version of scipy similar to what Laura suggested when @nkerman mentioned that healpy also requires Python >=3.10.

lacoak21: Note: scipy==1.13.0 supports python  >= 3.9

imap_processing/idex/idex_l2a.py
lacoak21: Would something like this work? 

        t_cal_int = np.round(t_calc).astype(int)
        t_i[t_cal_int[t_cal_int < len(t_i[0])]] = 1

Im having trouble understanding t_i[np.arange(len(t_calc)) < t_calc] = 1
.

greglucas: That might work, I don't really know. I am not entirely clear what structure you're after here either, so I just commented something that looked similar trying to set all values less than some index equal to 1. That is where the arange() idea came from. If the code you put works, you might want to add some more explicit comments/descriptions about what elements are getting set to 1 here.

imap_processing/idex/idex_l2a.py
lacoak21: The best shift and stretch values are converted to seconds right before this calculation so I think it should be ok. I will double check with Ethan.

imap_processing/tests/idex/test_idex_l2a.py
lacoak21: Oops that was supposed to be "us" not "ms".

imap_processing/idex/idex_l2a.py
lacoak21: Since "tof_high" and "hs_time" are 2d I was having difficulties with boolean indexing. indexing with a 2d mask will flatten the results to be 1d. I am setting values to nans to avoid having to do this in a loop.

imap_processing/idex/idex_l2a.py
lacoak21: Good idea.

imap_processing/idex/idex_l2a.py
lacoak21: In the example CDFS they do carry all the l1b data to l2a. I will double check with them that this is what they want.

imap_processing/idex/idex_l2a.py
lacoak21: Yes! It allows you to apply a function along an axis. It is similar to np.apply_along_axis

imap_processing/idex/idex_l1a.py
lacoak21: Yes this was wrong before. I noticed it was wrong in the IDEX code as well and Ethan and I are working to make sure this is correct. I can add more tests in a separate PR.

imap_processing/idex/idex_l2a.py
lacoak21: There will eventually be a conversion factor applied to convert the charge to mass so eventually these will be different. I checked with the IDEX team that they do in fact want both of these.

imap_processing/idex/idex_l2a.py
lacoak21: Yes seems like that is working without .data

pyproject.toml
nkerman: Today @greglucas @tech3371 and I talked, and we decided to drop `3.10` in #1423 . I think that'll be applicable here too.

lacoak21: Great thanks for following back up. Did you mean you decided to drop 3.9?

nkerman: Yes, sorry I'm incapable of typing 😁 **Drop 3.9**

imap_processing/idex/idex_l2a.py
```diff
@@ -0,0 +1,789 @@
...
+class BaselineNoiseTime(IntEnum):
+    """
+    Time range in microseconds that mark the baseline noise before a Dust impact.
+
+    Attributes
+    ----------
+    STOP: int
+         Beginning of the baseline noise window.
+    START: int
+        End of the baseline noise window.
```
greglucas: This is backwards of what I would think of STOP and START being...

pyproject.toml
```diff
@@ -53,12 +52,13 @@ sphinx = {version="*", optional=true}
 sphinxcontrib-openapi = {version="^0.8.3", optional=true}
 mypy = {version="1.10.1", optional=true}
 requests = {version = "^2.32.3", optional = true}
+scipy = "^1.15.2"
```
greglucas: If this is required can you move it up into the section above this.

Also, this pins us to 1.15 versions only. Can we change this to `^1.13` to get a few allowable versions. Or at least `^1.15` to avoid the `.2` portion limiting our updates in only the patch version specifier.

greglucas (commented):
Two minor comments

imap_processing/idex/idex_l2a.py
lacoak21: OH yeah.. I dont understand why I did that.

greglucas (approved):

.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
```
tech3371: Looks like Maxine has some issue with version `3.13`, you may want to remove this for now.

tech3371 (approved):
Looks good still. Just had one comment.

PR #1390, opened 2025-02-25
Ultra l1a - Add unique id

### LINKED ISSUE #1271: Ultra l1a - Time Update
### Description of the issue

This is how it is done now:
    for time in np.unique(decom_events["SHCOARSE"]):
        # Get the nearest spin start and duration prior to the event
        spin_start = aux_spin_starts[aux_spin_starts <= time][-1]

In the header of the event packet is 'Spin number at start of integration', in the Aux packet (page 187), there is a 'Spin Number' along with the time and period of that spin. I believe that this is the same data that is in the MOC file - though I would recommend that when you build the L1B that you use the spin data in the Aux packet (since this is how the instrument does timing) - whereas for the extended spin table, that you use the spin data from the MOC (since all the other teams are using this).
The fragment of code above doesn't look right to me - you want to match the 'Spin number at start of integration' with the 'Spin number' in the aux packet.  My suggestion would be something like building a dictionary keyed on 'spin number' from the Aux packet containing the Spin time/period/phase, then when you process the events you can just look up time from spin number.
I'm going to assume that since the event packet item is called 'Spin number at start of integration' that the spin phase can roll over within the packet - so you want to be sure to have the time NOT roll over. Example: say that the events in a packet with spin phases 718, 719, 0, 1, then in the equation for time you'll want to use the spin phases 718, 719, 720, 721. 

* Also look into using tdb instead of met within the data product

From Bob's comment in a PR:
More generally for the SDC - I strongly suggest that you adopt a single internal time representation across the SDC, or at least in the libraries that relate to instrument processing. This need not be the same as whatever is represented in the CDF files, rather a single unambiguous time representation that is recognized across the codebase that can be easily converted to whatever you put in the CDF files. A suggestion would be to define something like an sdc_time class that has constructors that use various types of time (met, TDB, segmented UT, etc.) and methods that can return a time in those various representations. Mixing time systems is an easy path to error and/or confusing conversions inside units of code that have to identify and convert times every time they see it. Further, this is something that should be defined and managed by the developers, this is not something that should be determined by consensus of instrument teams - they should review it to be sure it is sufficient for their application, but it is really and implementation decision best made by those using it.

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

### LINKED ISSUE #1272: Ultra - Unique Identifier
### Summary

On the subject of building a unique identifier, I would recommend the following - use a U64 integer, the first 32 bits, use the U32 in the secondary header (SCLK - I think this is SHCOURSE), the next 8 bits would be the number of the event in the packet (the first event in the packet is 0, the second event is 1, etc. up through the number of events in the packet), take the next 4 bits for 'flight unit' (something like 0 = FM90, 1= FM45, 2-15 for any of the EMs - I think you only need to worry about 0 and 1 in the SDC). The rest is reserver/padding.

Yes. I think we came to that conclusion yesterday. Because the data isn’t in a database having a continuing unique number is more complicated (think reprocessing etc).
But in the cdf we could do:
coordinate_0: SHCOARSE (aka met, aka SCLK w/ a constant conversion)
coordinate_1: incrementing number within the packet. (edited) 
[8:15](https://lasp.slack.com/archives/C05HF56S58T/p1737040538307769?thread_ts=1737037579.307209&cid=C05HF56S58T)
So agreed. I think that is a good plan

Slack messages Feb 14th:
OK - I just found something potentially nasty while looking at the validation data set. TLDR - we are going to drop events that occur that come in at nearly the same time (within 0.02 seconds). Another way to say this is that we get 2 events during a spin with the same phase.
The event time that is calculated is basically the time of the spin start + the spin phase represented in 1/2 degree increments (i.e., represented as a integer between 0 and 719).   So if 2 events come in with the same phase (i.e., nearly the same time) they will be assigned the same epoch. HOWEVER, it looks like the L1A CDF file is indexed on this time (the variable is "epoch"), so repeat values are not allowed, and clearly some are being ignored.
The clue for this is that there are 3313 events in the validation data set and 2461 events in the L1A.  I started looking through the data to see where the differences are - the first place was at line 18 and 19 of the validation CSV - where the spin phases are the same with the same packet header.
Basically, I think this means that the CDF cannot be indexed on the "epoch" field. We should probably figure out how to build a sequence number.

My proposal for an event ID:  a 64 bit integer - the upper 32 bits are the "day of acquisition" (defined below) and the lower 32 bits are the event sequence number for the day. The "day of acquisition" would be floor(SHCOARSE/86400) (SHCOARSE is the MET measured in seconds and is available in the telemetry).  The real question is how to generate a good sequence number. Complications arise because we don't know if the SDC is going to get all the data for a single "day of acquisition", and we want the generation of the sequence numbers to be reproducible. The SDC folks are going to know best on this. Since this is a can of worms, we may want to continue in thread.

**I may not be getting something essential here, but it seems like you could construct a unique and monotonic integer with the MET/SHCOARSE/packetID and the event number. The MET/SHCOARSE/packetID is a 32 bit number holding the time in seconds, right? Not sure how many events we can put in a packet, but I'm pretty sure that it will be less than 4 billion, so that can be stored in 32 bits as well. So just put the 32 bits of the MET/SHCOARSE/packetID in the upper 32 bits of a 64 bit integer. Put the sequential number of each event in the packet into a 64 bit integer.  Then bitwise OR these to get the final U64.
In python, Say for a particular packet (like the first packet of the validation set):
MET = 445015651 = 0x1a866663
Shift these over 32 bits
packet_base = MET<< 32 = 1911327667253149696 = 0x1a86666300000000
Say there are 4 events in the packet, those events would have the event ID
event_id = [0x1a86666300000000,0x1a86666300000001,0x1a86666300000002,0x1a86666300000003}
The next packet (assuming one shows up a second later) would have
MET = 445015652 = 0x1a866664
Lets say there are 5 events in that packet, they would have numbers
[0x1a86666400000000,0x1a86666400000001,0x1a86666400000002,0x1a86666400000003,0x1a8666640000004]
and so on. these 64 bit integers are monotonic, so they sort nice, and since they are just integers, I can't see why they wouldn't be able to be a coordinate in a CDF file.**

### LINKED ISSUE #1352: Ultra l1b extendedspin bug
### Description of the issue

We need to have all of the information for the extendedspin table. However, in the l1b de data product we dropped all of the values with an invalid starttime. In other words, all of the spins without events were removed. We need to bring those back so that the extended spin table will have the required information for spins that did not include events.

In de.py

    de_dataset = de_dataset.where(
        de_dataset["START_TYPE"] != np.iinfo(np.int64).min, drop=True
    )

This is a simple change, but might propagate into other code so therefore this separate ticket has been created.

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

laspsandoval:
# Change Summary

## Overview
Adding a unique id and calculating the eventtimes in l1b using the universal spin table rather than in l1a using the aux packet. Also, removed the part of the code that deleted the non-event packets for testing. 

In future PRs I will:
1. Add the unique id as a coordinate
2. Take out the aux packet dependency for l1a and add the aux packet dependency for l1b (so it can be compared to the universal spin table in the form of a quality flag)

## Updated Files
- de.py
   - Now includes starttimes with nan values
- ultra_l1a.py
   - Calculate unique event IDs as 64-bit integers.
- ultra_l1b_extended.py
   - Eventtimes now calculated in ultra_l1b_extended.py
   - Added new attributes

## Testing
- conftest.py
   - Took out part that removes start_type with fill values
- test_ultra_l1a.py
   - Removed the drop duplicate section
- test_ultra_l1b_extended.py
   - Test for eventtimes calculation
- test_de.py
   - Test for de.py


imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -122,57 +122,36 @@ def initiate_data_arrays(decom_ultra: dict, apid: int) -> xr.Dataset:
 
 def get_event_time(decom_ultra_dict: dict) -> dict:
```
demajistre: Since the new routine is creating an event ID and not an event time, you should probably rename it

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -101,11 +101,32 @@ x_coin:
...
+  FIELDNAM: spin_starts
+  LABLAXIS: spin starts
+  UNITS: seconds
+
+phase_angle:
+  <<: *default_float32
+  CATDESC: Phase angle from the direct event packet.
+  FIELDNAM: phase_angle
+  LABLAXIS: phase angle
+  UNITS: seconds
```
greglucas: Degrees?

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -101,11 +101,32 @@ x_coin:
...
+
+spin:
+  <<: *default_uint8
```
greglucas: We are going to have more than 256 spins / day. This likely needs to be larger than a uint8 if we are going to store incrementing spins.

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -122,57 +122,36 @@ def initiate_data_arrays(decom_ultra: dict, apid: int) -> xr.Dataset:
...
+        else:
+            packet_counters[met] += 1
 
-            # If there were no events, the time is set to 'SHCOARSE'
-            if decom_events["COUNT"][event_index] == 0:
-                event_times.append(decom_events["SHCOARSE"][event_index])
-            else:
-                event_times.append(spin_start + (duration / 1000) * (phase_angle / 720))
+        # Create the 64-bit event ID using np.int64 instead of np.uint64.
+        packet_base = np.int64(met) << np.int64(32)
```
greglucas: We are OK with a negative ID then? uint seems like it would be easier for handling IDs, so what do we get by including negatives?

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -122,57 +122,36 @@ def initiate_data_arrays(decom_ultra: dict, apid: int) -> xr.Dataset:
...
+        event_id = packet_base | np.int64(packet_counters[met])
+        event_ids.append(event_id)
```
greglucas: I'm not exactly clear what we're trying to do here. Now we are bitwise oring with the number of other events that have occurred here. Would it be easier to just get some random hashes instead and start assigning them? Or even keep track of a global counter and increment that?

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
demajistre: This is a good point - in L1A, the spin angle is an integer and represents 1/2 degree increments (i.e., it's range is 0-719).  Here is should probably be converted to (floating point) degrees . BUT the algorithm for eventtime (in our algorithm document) works on the UNCONVERTED value. just so things get implemented correctly.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
laspsandoval: Great catch.

imap_processing/ultra/l1a/ultra_l1a.py
laspsandoval: The ultra team wants to have a single unique id (not a combination of SHCOARSE + SID) for each event that can be reproduced. The random hash won't accomplish that. A global counter gets complicated if we were to reprocess. So this is what Bob came up with and I couldn't think of a better idea. But if you have a better idea, let me know.

imap_processing/ultra/l1a/ultra_l1a.py
laspsandoval: It isn't ideal. When I used uint64 I get the following error thrown from cdf_to_xarray. I didn't dive in deep to understand the reason for the error in cdf_to_xarray, though. Maybe Bryan Harter would be able to comment on it. What do you think? 

```
        else:
            _warn_or_except(f"NOT SUPPORTED: Data in variable {var.name} has data type of {var.dtype}.", terminate_on_warning)
    
>       return STRINGS_TO_DATATYPES[cdf_data_type], element_size
E       KeyError: 'CDF_UINT8'

../../../../venv/lib/python3.11/site-packages/cdflib/xarray/xarray_to_cdf.py:194: KeyError
```

greglucas: Oh yeah, CDF does not have a datatype for uint64. I think Tim was running into this a while ago.
https://github.com/MAVENSDC/cdflib/issues/275

imap_processing/ultra/l1a/ultra_l1a.py
greglucas: I hesitate saying that a `packet_counters[met]` is less reproducible than a global count... That still relies on there being some notion of ordering coming in to determine what other number we are appending here.

What makes an event "unique", are you able to use other fields to distinguish between the events? I think Tim also had a recent PR that was possibly doing something similar so it might be worth seeing what Hi was doing to disentangle these duplicate times too.

laspsandoval: This is what Tim has said:

```
Hi L1A/B now have two independent time coordinate variables. One, the main "epoch" variable is derived from CCSDS creation time (SHCOARSE). The other is event MET times which similar to Ultra can have duplicate values.

All of the packet level telemetry have DEPEND_0=epoch , all event level telemetry have DEPEND_0=event_met.

Duplicates aren't allowed in the main epoch variable. So, using CCSDS creation time for epoch satisfies ISTP. 
[11:58](https://lasp.slack.com/archives/C06V39BMLR5/p1739991483069409?thread_ts=1739990979.760969&cid=C06V39BMLR5)
I need to work with Andriy to get some feedback on this approach, but L1 doesn't NEED to be ISTP compliant, so I can fall back on that.
```

laspsandoval: This is what Sean is doing Screen Shot 2025-02-19 at 12 05 22 PMd:

laspsandoval: For Ultra there is no "direct event" time. We only have SHCOARSE, which is the packet creation time. There are a lot of different ways to do this, but I don't know what the "right" way is for every event to have a unique id.

laspsandoval: In the past I had calculated "eventtimes" and used that as the unique id, but it turned out that even that had duplicates.

demajistre: I'm going to try to summarize what we're trying to do here.  Our individual events come down in variable length packets - the variability is the number of events that the packet contains.  The events themselves are NOT directly assigned a time by the instrument, although a moderate resolution time can be derived from the information in the event data.  In particular, each event has a Spin ID and a "spin phase" - the "spin phase" is an integer between 0 and 719 that is basically a 1/2 degree measure of the the spin phase angle where the event was acquired (this value is calculated by the instrument based on the time of the event and the spin packet that the instrument receives from the spacecraft).  In short, the instrument knows when an event comes in, but it doesn't report that time directly, it encodes the time in the spin angle.  The immediate problem is that the resolution of the spin angle (0-719) is pretty low, corresponding to about 0.02 seconds - and there is a reasonable possibility that there could be more than one event that comes in within this time/angle.  So it is very possible that we well have two events that come in at the same encoded time.   When Laura did this and associated this time with the CDF epoch variable, we had duplicate values of epoch, which caused us to drop events.  

We're proposing a solution here that gives each event a unique ID, eliminates the duplicated epoch field and defers the calculation of an event time until L1B. This last is attractive because the event time (based on the spin angle) is a derived field, and requires information outside of the packet (the spin packet) that is being decommed. 

I did not know that CDF didn't have a good 64 bit unsigned integer.  This could be trivially handled by shifting the packet time (SHCOARSE) by 31 bits instead of 32 bits. It would no longer be byte aligned, but it would always sort correctly (which is nice).  the number of bits required for the event number is much less than 30, so there is plenty of spare space in this bitfield.

Coming up with a hash is another valid approach.  But it would not sort and would require information outside of the packet itself - so I regard it as less attractive.

greglucas: OK, I am starting to see what is going on here. Thank you both for the explanations.

I think it would help to add some comments here on exactly what we are doing and why. Here is my suggestion:

```python
# Left shift SHCOARSE (u32) by 31 bits, to make room for our event counters
# (31 rather than 32 to keep it positive in the int64 representation)
# Append the current number of events in this packet to the right-most bits
# This will make each event a unique value including both the MET and event number in the packet
# NOTE: CDF does not allow for uint64 values, so we use int64 as the representation here
event_id = (np.int64(met) << np.int64(31)) | np.int64(packet_counters[met])
```

sdhoyt (approved):

PR #1396, opened 2025-02-25
Implement "Pull" Index matching method and value binning/projection; Convert TT -> TDB

### LINKED ISSUE #1394: Implement "Pull" Index matching and value projection method for ENA Maps
### Algorithm Description:
Must implement the "pull" method of index matching (Please see [description](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/247989696/Creating+L2+ENA+Sky+Maps+with+PointingSet+and+SkyMap+Objects#CreatingL2ENASkyMapswithPointingSetandSkyMapObjects-NoteonIndexMatchingIdx_match)).

Essentially should look like push index matching in reverse. The subsequent step of binning to project values to a `SkyMap` will be much simpler, as there is a guaranteed, unique matching of one SkyMap pixel to a PointingSet pixel. That means we can just say:

```
SkyMap_values = PointingSet_values[pull_matched_PSET_indices]
```

### LINKED ISSUE #1395: Convert ENA Map time from Epoch (TT J2000 ns) --> TDB ET, used by frame_transform
Very small change: When the ENA Map code uses `geometry.frame_transform_az_el`, the time is expected to be in Spice-compatible TDB ET. However the `epoch` of the PointingSets, where the code gets its time for the conversion, are stored in TT_J2000ns. These can be converted with an existing tool. 

1. Apply the conversion in the ENA Map code
2. In the Tests, change epoch to contain TT_J2000ns (not really necessary, but more realistic)

nkerman:
# Change Summary
Implement Pull index matching and projection, and fixes spice call using incorrect time format. 

## Overview
Implements the Pull method of index matching and projection (see [description](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/247989696/Creating+L2+ENA+Sky+Maps+with+PointingSet+and+SkyMap+Objects#CreatingL2ENASkyMapswithPointingSetandSkyMapObjects-NoteonIndexMatchingIdx_match) for explanation of 'pull' vs 'push' matching.) 

It turns out the binning part of projection is very simple for pull matching projection, because there is a guaranteed, unique matching of one SkyMap pixel to a PointingSet pixel. 
Closes #1394

Also, the spice frame_transform had been using the wrong format of time, and the time specified in the PointingSet's epoch has to be converted from TT J2000ns --> TDB.
Closes #1395

## Updated Files

- `imap_processing/ena_maps/ena_maps.py`
   - Allow for pull index matching
   - If using the pull index matching, apply simple binning
   - Apply time conversion
- `imap_processing/tests/ultra/test_data/mock_data.py`
   - Change the stored epoch in the mocked L1C data to use TTJ2000ns

## Testing

- Tests for pull method.

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -471,9 +471,9 @@ class RectangularSkyMap(AbstractSkyMap):
         """
         if value_keys is None:
             value_keys = list(pointing_set.data.data_vars.keys())
-        for pset_key in value_keys:
-            if pset_key not in pointing_set.data.data_vars:
-                raise ValueError(f"Value key {pset_key} not found in pointing set.")
+        for value_key in value_keys:
+            if value_key not in pointing_set.data.data_vars:
+                raise ValueError(f"Value key {value_key} not found in pointing set.")
 
         if index_match_method is IndexMatchMethod.PUSH:
             # Determine the indices of the sky map grid that correspond to
@@ -522,7 +522,7 @@ class RectangularSkyMap(AbstractSkyMap):
             elif index_match_method is IndexMatchMethod.PULL:
                 # We know that there will only be one value per sky map pixel,
                 # so we can use the matched indices directly
-                pointing_projected_values = raveled_pset_data[matched_indices_pull]
+                pointing_projected_values = raveled_pset_data[..., matched_indices_pull]
             else:
                 raise NotImplementedError(
                     "Only PUSH and PULL index matching methods are supported."
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -14,6 +14,7 @@ from numpy.typing import NDArray
 from imap_processing.cdf.utils import load_cdf
 from imap_processing.ena_maps.utils import map_utils, spatial_utils
 from imap_processing.spice import geometry
+from imap_processing.spice.time import ttj2000ns_to_et
 
 logger = logging.getLogger(__name__)
 
@@ -96,6 +97,7 @@ def match_coords_to_indices(
         Event time at which to transform the input spatial object to the output frame.
         This can be manually specified, e.g., for converting between Maps which do not
         contain an epoch value.
+        If specified, must be in SPICE compatible ET.
         The default value is None, in which case the event time of the PointingSet
         object is used.
 
@@ -121,12 +123,14 @@ def match_coords_to_indices(
     if isinstance(input_object, PointingSet) and isinstance(output_object, PointingSet):
         raise ValueError("Cannot match indices between two PointingSet objects.")
 
-    # If event_time is not specified, use event_time of the PointingSet, if present.
+    # If event_time is not specified, use epoch of the PointingSet, if present.
+    # The epoch will be in units of terrestrial time (TT) J2000 nanoseconds,
+    # which must be converted to ephemeris time (ET) for SPICE.
     if event_time is None:
         if isinstance(input_object, PointingSet):
-            event_time = input_object.data["epoch"].values
+            event_time = ttj2000ns_to_et(input_object.data["epoch"].values)
         elif isinstance(output_object, PointingSet):
-            event_time = output_object.data["epoch"].values
+            event_time = ttj2000ns_to_et(output_object.data["epoch"].values)
         else:
             raise ValueError(
                 "Event time must be specified if both objects are SkyMaps."
@@ -492,6 +496,8 @@ class RectangularSkyMap(AbstractSkyMap):
                     input_object=pointing_set,
                     output_object=self,
                 )
+                # Bin the values at the matched indices. There may be multiple
+                # pointing set pixels that correspond to the same sky map pixel.
                 pointing_projected_values = map_utils.bin_single_array_at_indices(
                     value_array=raveled_pset_data,
                     projection_grid_shape=(
@@ -501,69 +507,20 @@ class RectangularSkyMap(AbstractSkyMap):
                     projection_indices=matched_indices_push,
                 )
             elif index_match_method is IndexMatchMethod.PULL:
+                # Determine the indices of the pointing set grid that correspond to
+                # each pixel in the sky map.
                 matched_indices_pull = match_coords_to_indices(
                     input_object=self,
                     output_object=pointing_set,
                 )
+                # We know that there will only be one value per sky map pixel,
+                # so we can use the matched indices directly
                 pointing_projected_values = raveled_pset_data[matched_indices_pull]
             else:
                 raise NotImplementedError(
                     "Only PUSH and PULL index matching methods are supported."
                 )
             self.data_dict[value_key] += pointing_projected_values
-
-    def to_dataset(
-        self,
-        output_value_keys_dims: dict[str, list[str]],
-        global_attrs: dict[str, str] | None = None,
-    ) -> xr.Dataset:
-        """
-        Convert the map data to an xarray Dataset.
-
-        Parameters
-        ----------
-        output_value_keys_dims : dict[str, list[str]]
-            Dictionary of dimensions for each data variable in the map.
-            Keys are the data variable names, and values are lists of dimension names.
-        global_attrs : dict[str, str], optional
-            Dictionary of attributes to add to the dataset.
-            Default is None.
-
-        Returns
-        -------
-        xr.Dataset
-            The xarray Dataset containing the map data.
-        """
-        if global_attrs is None:
-            global_attrs = {}
-
-        dataset = xr.Dataset(attrs=global_attrs)
-
-        # Add azimuth, elevation grid as coordinates
-        dataset["azimuth_bin_center"] = xr.DataArray(
-            np.rad2deg(self.sky_grid.az_bin_midpoints),
-            dims=["azimuth_bin_center"],
-            attrs={"units": "degrees"},
-        )
-        dataset["elevation_bin_center"] = xr.DataArray(
-            np.rad2deg(self.sky_grid.el_bin_midpoints),
-            dims=["elevation_bin_center"],
-            attrs={"units": "degrees"},
-        )
-
-        for key in output_value_keys_dims.keys():
-            wrapped_data = spatial_utils.rewrap_even_spaced_az_el_grid(
-                self.data_dict[key]
-            )
-            # Flatten any length-1 dimensions
-            # TODO: this may need to be tweaked to handle epoch coordinate?
-            wrapped_data = np.squeeze(wrapped_data)
-            dataset[key] = xr.DataArray(
-                wrapped_data,
-                dims=output_value_keys_dims[key],
-            )
-
-        return dataset
 
     def __repr__(self) -> str:
         """
@@ -579,10 +536,3 @@ class RectangularSkyMap(AbstractSkyMap):
             f"{self.spice_reference_frame.name} ({self.spice_reference_frame.value}), "
             f"spacing_deg={self.spacing_deg}, num_points={self.num_points})"
         )
-
-
-# TODO:
-# Add pulling index matching in match_pset_coords_to_indices
-
-# TODO:
-# Check units of time which will be read in. Do we need to add j2000ns_to_j2000s?
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -475,6 +475,25 @@ class RectangularSkyMap(AbstractSkyMap):
             if pset_key not in pointing_set.data.data_vars:
                 raise ValueError(f"Value key {pset_key} not found in pointing set.")
 
+        if index_match_method is IndexMatchMethod.PUSH:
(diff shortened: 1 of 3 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -454,34 +463,46 @@ def project_pset_values_to_map(
...
+        for pset_key in pset_value_keys:
+            if pset_key not in pointing_set.data.data_vars:
+                raise ValueError(f"Value key {pset_key} not found in pointing set.")
+
+        for pset_key, map_key in zip(pset_value_keys, skymap_value_keys):
             # If multiple spatial axes present
             # (i.e (az, el) for rectangular coordinate PSET),
             # flatten them in the values array to match the raveled indices
-            raveled_pset_data = pointing_set.data[value_key].data.reshape(
+            raveled_pset_data = pointing_set.data[pset_key].data.reshape(
```
tmplummer: I didn't catch this earlier, but this imposes the requirement that the PSET variable has the spatial dimension as either the first or last dimension... I can't quite figure out which. But this is a gotcha that may not be documented?

imap_processing/tests/ultra/test_data/mock_data.py
```diff
@@ -129,6 +129,12 @@ def get_binomial_counts(distance_scaling, lon_bin, central_lon_bin):
 
     sensitivity = np.ones(grid_shape)
 
+    # Determine the epoch, which is TT time in nanoseconds since J2000 epoch
+    tdb_et = ensure_spice(spice.str2et, time_kernels_only=True)(timestr)
```
tmplummer: There is a `str_to_et` function in `imap_processing.spice.time`.

imap_processing/tests/ultra/test_data/mock_data.py
```diff
@@ -129,6 +129,12 @@ def get_binomial_counts(distance_scaling, lon_bin, central_lon_bin):
...
+    tt_j2000ns = (
```
tmplummer, line 115: This display of jumping through hoops to get the correct times reinforces the idea that perhaps we should get rid of ET altogether by making all of the geometry functions take tt_j2000ns as the time argument.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -454,34 +463,46 @@ def project_pset_values_to_map(
...
+            if map_key not in self.data_dict:
                 # Initialize the map data array if it doesn't exist (values start at 0)
                 output_shape = (self.num_points, *raveled_pset_data.shape[1:])
-                self.data_dict[value_key] = np.zeros(output_shape)
+                self.data_dict[map_key] = np.zeros(output_shape)
 
             if index_match_method is IndexMatchMethod.PUSH:
+                # Determine the indices of the sky map grid that correspond to
+                # each pixel in the pointing set.
+                matched_indices_push = match_coords_to_indices(
```
tmplummer: This looks like it is matching the coordinates for every variable that gets binned. I though that this matching was going to happen just once per PSET.

tmplummer (commented):
I have a few comments and am hitting the limit for being able to focus for now. This is what I have so far.

imap_processing/ena_maps/ena_maps.py
nkerman: Yeah, it requires the spatial coordinate/s (either `(lon, lat)` or `(healpix_index,)`) to come first. I think that it's okay to expect it as such, as a user can transpose as necessary. However, it definitely should be documented. I'll add documentation to the `PointingSet` class. 

I didn't think it would be a potential problem/I thought that the spatial coordinates would always come first until Matina said in the meeting that we might want to do (energy x lon x lat) instead of (lon x lat x energy). It sounded like she moved away from that, but I agree it should be documented.

imap_processing/tests/ultra/test_data/mock_data.py
nkerman: Done!

imap_processing/tests/ultra/test_data/mock_data.py
nkerman: It's definitely preferable to pick one or the other. I guess I don't have a strong understanding of why we would prefer TT vs ET. Since the epoch will be in TT, there will always need to be _somewhere_ the conversion happens.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -425,7 +429,8 @@ def __init__(
     def project_pset_values_to_map(
         self,
         pointing_set: PointingSet,
-        value_keys: list[str] | None = None,
+        pset_value_keys: list[str] | None = None,
+        skymap_value_keys: list[str] | None = None,
```
greglucas: Do we need to allow for the renaming here too? I feel like this is another extra argument that could lead to confusion and someone can later just take their map and rename variables instead. It also requires things to be in the same order to be mapped properly leading to another potential gotcha.

imap_processing/ena_maps/ena_maps.py
greglucas: I think the discussion ended with us having the spatial dimension(s) last. So it would be (time, energy, lon, lat). i.e. you would choose a time and energy to make a map of and then call plot on the remaining dimensions `plot(var[time, energy, ...])` (not intermixed)

imap_processing/ena_maps/ena_maps.py
nkerman: Good catch. I think this happened because I was trying to avoid having 2 different sets of conditionals:
```
if Push:
    ###
if Pull:
    ###
```
But I think it's fine to have those, as one set needs to be in the loop.

imap_processing/ena_maps/ena_maps.py
nkerman: I added `skymap_value_keys` for optional renaming because we may need to push _and_ pull some quantities. E.g., we may need "counts_pushed" and "counts_pulled". I could have a suffix instead, but I lean towards the user making a more explicit choice of names.

imap_processing/ena_maps/ena_maps.py
nkerman: Ah, darn. Do you know if that's just true for L2 and beyond? I think that the L1C definitions for Ultra, at least, are `(epoch, lon, lat, energy)`. I'll loop in @laspsandoval

imap_processing/ena_maps/ena_maps.py
greglucas: > we may need

I would leave it until we need it then. I would say someone should make two separate maps then and combine/work on them later.

imap_processing/tests/ultra/test_data/mock_data.py
```diff
@@ -129,6 +130,12 @@ def get_binomial_counts(distance_scaling, lon_bin, central_lon_bin):
 
     sensitivity = np.ones(grid_shape)
 
+    # Determine the epoch, which is TT time in nanoseconds since J2000 epoch
```
laspsandoval, line 113: Not in this PR, but I found it anyways. On line 120:
counts / 2 is unused

imap_processing/tests/ultra/test_data/mock_data.py
```diff
@@ -143,7 +150,7 @@ def get_binomial_counts(distance_scaling, lon_bin, central_lon_bin):
                 ["azimuth_bin_center", "elevation_bin_center", "energy_bin_center"],
                 sensitivity,
             ),
-            "epoch": ensure_spice(spice.str2et, time_kernels_only=True)(timestr),
+            "epoch": tt_j2000ns,  # nanoseconds since J2000 epoch in TT
```
laspsandoval: On line 155:

Dimensions (per Greg's Slack message during the meeting)
Time, energy, spin/azimuthal angle (0-360), polar/elevation angle (0-180)

imap_processing/tests/ultra/test_data/mock_data.py
laspsandoval, line 113: Line 89:
Why is the first energy bin being dropped?

imap_processing/ena_maps/ena_maps.py
laspsandoval: Thanks for the heads up. The L1C definitions were created a while back. We can change the l1c to align with that sequence.

imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -16,7 +16,7 @@
 @pytest.fixture()
 def l1c_pset_products():
     """Make fake L1C Ultra PSET products for testing"""
-    l1c_spatial_bin_spacing_deg = 10
+    l1c_spatial_bin_spacing_deg = 2
```
laspsandoval, line 19: Curious why you put 2 degrees here. It should be 0.5 x 0.5 degree binning right?

laspsandoval (commented):
Nice work Nat.

imap_processing/tests/ena_maps/test_ena_maps.py
nkerman: I've tested it with 0.5, and even smaller, but it's just pretty slow with so many pixels (`(360*180 / (0.5 ^2)`) in each PSET. All that should matter is testing with a variety of relative sizes of Map vs. PSET pixels.

imap_processing/ena_maps/ena_maps.py
nkerman: Makes sense - I think I often add this sort of early over-flexibility, which I can see affects legibility and potentially maintainability. I'm working to reduce it!

imap_processing/tests/ultra/test_data/mock_data.py
nkerman: Hmmm - I can't remember, but perhaps `build_energy_bins()` had previously been defined in such a way that the midpoints were just the edges + half the distance? Or maybe I just thought they were.  Either way, good catch, and I can get rid of this.

nkerman: Removed this!

imap_processing/ena_maps/ena_maps.py
nkerman: Dropped this parameter.

imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -192,23 +192,76 @@ def test_project_pset_values_to_map_push_method(self, mock_frame_transform_az_el
...
+                ultra_pset,
                 value_keys=["counts", "exposure_time"],
                 index_match_method=index_matching_method,
             )
+            expected_value_every_pixel += pset_num
+
+            total_pset_counts += ultra_pset.data["counts"].values
+
+        # Check that the map has been updated
+        assert rectangular_map.data_dict != {}
```
greglucas: assert some key is in it instead? This just says it isn't empty but not what you expect to be in it.

greglucas (approved):
Thanks for breaking this into a smaller PR and separating it from the other index re-arranging logic. I think this looks good.

imap_processing/tests/ena_maps/test_ena_maps.py
nkerman: Switching to:
```assert "counts" in rectangular_map.data_dict```

File: imap_processing/ena_maps/ena_maps.py [hunk 10]
```diff
@@ -121,12 +123,14 @@ def match_coords_to_indices(
...
         raise ValueError("Cannot match indices between two PointingSet objects.")
 
-    # If event_time is not specified, use event_time of the PointingSet, if present.
+    # If event_time is not specified, use epoch of the PointingSet, if present.
+    # The epoch will be in units of terrestrial time (TT) J2000 nanoseconds,
+    # which must be converted to ephemeris time (ET) for SPICE.
     if event_time is None:
         if isinstance(input_object, PointingSet):
-            event_time = input_object.data["epoch"].values
+            event_time = ttj2000ns_to_et(input_object.data["epoch"].values)
```
tmplummer, line 131: Can you rename this variable to `event_et`? Since we have several different combinations of epochs and time scales, we need to be diligent about naming time variables with what they contain.

imap_processing/tests/ultra/test_data/mock_data.py
tmplummer: Sorry if there was confusion from my comment. It is not relevant to this PR. Greg and I had a discussion about trying to reduce the number of different time variables that we all use. There should be an upcoming PR that will convert all of the geometry functions to using TT nanoseconds as input and do the conversion to ET internally.

tmplummer (approved):
LGTM.

nkerman:
>Can you rename this variable to event_et? Since we have several different combinations of epochs and time scales, we need to be diligent about naming time variables with what they contain.

>Sorry if there was confusion from my comment. It is not relevant to this PR. Greg and I had a discussion about trying to reduce the number of different time variables that we all use. There should be an upcoming PR that will convert all of the geometry functions to using TT nanoseconds as input and do the conversion to ET internally.

Thanks @subagonsouth - I'll give this sort of more specific name in my follow-up commits, but given your second comment, I think I'll merge this in as passed (so as to avoid having to run all the CI checks again)

PR #1399, opened 2025-02-26
Exclude certain variables in CoDICE housekeeping products

bourque:
There are some variables that exist in the raw packet data that do not need to be carried through to L1a (or further). This PR makes changes to exclude these variables.


imap_processing/codice/codice_l1a.py
```diff
@@ -547,12 +547,33 @@ def create_hskp_dataset(
...
+        "seq_flgs",
+        "src_seq_ctr",
+        "pkt_len",
+        "chksum",
+    ]
+
     for variable in packet:
-        attrs = cdf_attrs.get_variable_attributes(variable)
+        if variable not in exclude_variables:
+            attrs = cdf_attrs.get_variable_attributes(variable)
```
tech3371: you could this this and undo below line changes.
```suggestion
        if variable in exclude_variables:
             continue

        attrs = cdf_attrs.get_variable_attributes(variable)
```

tech3371 (approved):
one minor suggestions, otherwise LGTM

imap_processing/codice/codice_l1a.py
```diff
@@ -547,7 +547,30 @@ def create_hskp_dataset(
         attrs=cdf_attrs.get_global_attributes("imap_codice_l1a_hskp"),
     )
 
+    # These variables don't need to carry over from L0 to L1a
+    exclude_variables = [
```
tmplummer, line 565: Just a note that the general practice that I am used to across missions is that L1A tries to stay as close as you can get to the data in the raw packet. I have not seen this guidance stated for IMAP.

In particular, `src_seq_ctr` can be very useful for identifying missing packets. Maybe there are other tools such as WebTCAD or WebPODA that can look at this packet telemetry of science data though?

bourque: Good point @subagonsouth! It is easy enough for me to carry these over to L1a and ignore them in the corresponding validation test, so I will do that.

PR #1400, opened 2025-02-26
Fix CoDICE-lo L1a validation discrepencies

bourque:
This PR fixes the issues I was finding during CoDICE-lo L1a validation.

The changes include:
- Introducing new configurations for defining separate `input_dims` and `output_dims`, which configure how data arrays are reshaped; `input_dims` defines how the data are stored within the raw packet data, while `output_dims` defines how data should be ordered in the L1a CDF file.
- Updated `reshape_data()` algorithm to account for the differences in `input_dims` and `output_dims`; data are first reshaped based on their order in the raw packet data, then transposed to the order they are written to the CDF
- Updated `LOSSY_A` and `LOSSY_B` decompression lookup tables and how they are used to match validation data
- Removed a whole bunch of leftover L1a CDF files that are no longer needed
- Updated test to reflect new decompression values
- Updated tests to compare entire data arrays (shapes and values) instead of just shapes.

(I am still seeing issues with `lo-counters-aggregated`; I intend to address those in a future PR)


imap_processing/codice/constants.py
```diff
@@ -257,9 +282,10 @@
     },
     CODICEAPID.COD_LO_NSW_PRIORITY_COUNTS: {
         "dataset_name": "imap_codice_l1a_lo-nsw-priority",
-        "dims": {"spin_sector": 12, "esa_step": 128},
+        "input_dims": {"esa_step": 128, "spin_sector": 12},
         "instrument": "lo",
         "num_counters": 2,
+        "output_dims": {"spin_sector": 12, "esa_step": 128},
```
greglucas, line 288: In the meeting we discussed that the "esa_step" should be the first dimension here I thought, so similar to what the input_dims is for this case.

I'm wondering if most of these output_dims are only needed for the test comparisons with the other CDFs and not necessarily for the actual production products?

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -228,17 +228,11 @@ def test_l1a_validate_data_arrays(test_l1a_data: xr.Dataset, index):
         validation_dataset = load_cdf(VALIDATION_DATA[index])
 
         for counter in counters:
-            # Ensure the data array shapes are equal
-            assert (
-                processed_dataset[counter].data.shape
-                == validation_dataset[counter].data.shape
+            # Ensure the data arrays are equal
+            np.testing.assert_equal(
+                processed_dataset[counter].data, validation_dataset[counter].data
```
greglucas, line 233: Give the xarray version a try, and I think you can drop the `.data` access as well.
```suggestion
            xr.testing.assert_equal(
                processed_dataset[counter], validation_dataset[counter]
```

greglucas (commented):
Really nice tracking this down. This looks like a nice simplification and great to see the values lining up now.

imap_processing/codice/constants.py
bourque: Yes, exactly. These will eventually change to match the conventions we settled on during the STM. Joey is working on getting me updated validation data to reflect this, but for now I will keep it like this to match the current validation data.

imap_processing/tests/codice/test_codice_l1a.py
bourque: Cool, I wasn't aware of `xr.testing`!

Unfortunately we don't have quite the same metadata values (for example, Joey uses camelcase e.g. "EnergyTable" where i use lowercase + underscores e.g. "energy_table" (which I think is the SPDF preferred convention?)) so this will throw errors.

When I try to compare just the data arrays with `xr.testing`, I get an error `TypeError: <class 'numpy.ndarray'> not supported by assertion comparison`

I can get around this by converting the numpy arrays to `xr.DataArray` first:

```python
            xr.testing.assert_equal(
                xr.DataArray(processed_dataset[counter].data), xr.DataArray(validation_dataset[counter].data)
            )
```

Not sure its worth doing it this way though? It seems like its just doing the same thing as `np.testing` at this point.

greglucas: Agreed, `np.testing` is the way to go here then!

imap_processing/codice/constants.py
greglucas: My comment was more whether you want to change the validation data to meet your data product, vs change your data product to meet the validation data.

This is fine, but it will require another update later once you get new data because I view yours as the more right one.

greglucas (approved):

PR #1405, opened 2025-02-27
Ultra l1b - phi theta corrections

### LINKED ISSUE #1372: Ultra l1b extended Bug - theta and phi calculations
### Description of the issue

[@laura.sandoval](https://lasp.slack.com/team/U45RK4249)
 - There is an error in the code fragment that you pasted above - I didn't catch it because I didn't understand the geometry of the calibration that we were using.  Line 124 should be
theta = np.arccos(uz)
NOT
theta = np.arccos(uz / vmag)
When this is done, we get the correct theta for the calibration (close to 28 degrees).  Note that this is a naming convention nightmare:
[@George Clark](https://lasp.slack.com/team/U04U3UK547J)
 this is the opposite of George's (i.e., the flight software) definitions.  Here Theta is the Polar Angle and Phi is the Azimuthal Angle.  The flight software convention is that Theta is close to the Azimuthal angle and Phi is close to the Polar Angle.  I'd note that this is a common ambiguity with spherical coordinate systems.  I've most commonly seen Theta->Polar, but there are numerous examples of doing it the other way
The name of the routine is event_az_el, and the is VERY confusing and very likely to be misused (I think I mentioned this before but did not follow up).  Theta is a Polar Angle, NOT an elevation.  In degrees, the elevation is 90-Polar Angle (in radians it's PI/2-Polar Angle).  ALSO the last line of the code is return theta, phi , so even if the Theta were an elevation, you are returning El,AZ and NOT AZ,EL.
All of this is a great subject for the standards document under discussion.

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

laspsandoval:
# Change Summary

## Overview
This PR addresses the update to the phi and theta equations.

## Updated Files
- ultra_l1b_extended.py
   - phi and theta calculation change
- de.py
   - phi and theta calculation change

## Testing
   - Update to cdf attributes 
- test_ultra_l1b_extended.py
   - Test phi and theta calculation change
- test_de.py
   - Test phi and theta calculation change
- ultra45_raw_sc_ultrarawimg_withFSWcalcs_FM45_40P_Phi28p5_BeamCal_LinearScan_phi2850_theta-000_20240207T102740.csv
   - Revert to previous test data columns


imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -215,6 +215,20 @@ elevation:
   LABLAXIS: elevation
   UNITS: radians
 
+phi:
+  <<: *default_float32
+  CATDESC: Ultra instrument frame event phi.
```
greglucas: It might help to add what coordinate this refers to, or put a min/max to indicate what you expect for phi. Phi is just a generic angle otherwise.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -751,31 +750,34 @@ def determine_species(tof: np.ndarray, path_length: np.ndarray, type: str) -> ND
     return species_bin
 
 
-def get_de_az_el(v: NDArray) -> tuple[NDArray, NDArray]:
+def get_theta_phi(
```
greglucas: This is really confusing. It should be `get_phi_theta` because you are returning (phi, theta). Or the return should be reversed.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -751,31 +750,34 @@ def determine_species(tof: np.ndarray, path_length: np.ndarray, type: str) -> ND
...
+    front_position: tuple, back_position: tuple, d: np.ndarray
+) -> tuple[NDArray, NDArray]:
     """
-    Compute azimuth (phi) angles and elevation (theta).
+    Compute phi angles and theta.
```
greglucas: I like the extra detail you had before. Is the theta here a polar angle or elevation angle? Is the range from -90 -> 90 or 0 -> 180?
```suggestion
    Compute the instrument angles phi and theta (azimuth, polar angle).
```

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -236,6 +236,20 @@ elevation:
   LABLAXIS: elevation
   UNITS: radians
 
+phi:
+  <<: *default_float32
+  CATDESC: Ultra instrument frame event phi (azimuth with range -90 -> 90).
```
greglucas, line 241: -180 -> 180 ?

imap_processing/ultra/l1b/de.py
```diff
@@ -129,6 +131,11 @@ def calculate_de(de_dataset: xr.Dataset, name: str, data_version: str) -> xr.Dat
         (xb[ph_indices], yb[ph_indices]),
         d[ph_indices],
     )
+    phi[ph_indices], theta[ph_indices] = get_phi_theta(
```
greglucas, line 134: Just a note that this is all somewhat complicated to follow/track when indices are always being subset from variables. Don't need to change here, but it might be nice to think about whether this would help simplify things in the future and whether a later PR could help consolidate some of this logic.

Working with datasets we can subset by indices across all variables at once, so everything is consistent and we don't drop a set of indices somewhere along the way.
```python
ds_ph_subset = ds.isel(ph=ph_indices)
# Then pass that ds into the functions and have the function get the proper variables
get_phi_theta(ds_ph_subset)
# Or pass in those variables from the dataset
get_phi_theta(ds_ph_subset["xf"], ...)
```

imap_processing/ultra/l1b/de.py
```diff
@@ -129,6 +131,11 @@ def calculate_de(de_dataset: xr.Dataset, name: str, data_version: str) -> xr.Dat
...
+        (xf[ph_indices], yf[ph_indices]),
```
greglucas, line 135: Another option here would be to call `get_phi_theta` on all values, and then say `phi[~ph_indices] = np.nan` so you mask things out after the calculations.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -752,34 +751,37 @@ def determine_species(tof: np.ndarray, path_length: np.ndarray, type: str) -> ND
...
+    phi : np.array
+        Ultra instrument frame event azimuth.
+    theta : np.array
+        Ultra instrument frame event elevation.
     """
-    # Compute azimuth (phi) angles and elevation (theta)
-    spherical_coords = cartesian_to_spherical(v, degrees=False)
+    path_length = get_path_length(front_position, back_position, d)
+
+    phi = np.arctan((front_position[1] - back_position[1]) / d)
```
greglucas, line 781: Do we need to account for different quadrants and use `arctan2(front_position[1] - back_position[1], d)` to get -180 -> 180 here?

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
nkerman: For azimuth, I also thought we settled on [0, 360) rather than [-180, 180)?
Perhaps this was just a copy/paste issue from Elevation below?

Secondly, is it okay to have this in degrees, and az/el above it in radians?

imap_processing/ultra/l1b/ultra_l1b_extended.py
nkerman: I think `d` will always be positive here, right? Or another way of thinking about it is that the instrument frame azimuth is restricted to 180 degrees, since the ENA's enter the slit in front of the detector. This might also explain the previous discussion about angular ranges in the attrs.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
nkerman: Ah, I think I now understand that this is fundementally separate from the other type of azimuth, and **only covers the half circle** in the direction of the slit, from the point of view of the detector. Is that why this would be -90 to 90 degrees?

imap_processing/ultra/l1b/de.py
laspsandoval: Good idea! I created a ticket:
https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1418

imap_processing/ultra/l1b/de.py
laspsandoval: True. I like your other suggestion better though.

imap_processing/ultra/l1b/ultra_l1b_extended.py
laspsandoval: d is always positive (distance from slit to foil), and the Ultra instrument only detects ENAs entering from the front slit. This constrains the azimuth to a ±90° field of view.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
laspsandoval: Confirmed with Matina that this is ok:
For these particular ranges of phi and theta angels as  they are defined in the FSW (not in spherical coordinates, which caused the confusion last week) we should keep the ranges as Nick suggests.

nkerman (approved):
Thanks for clarifying on the angle range - LGTM!

PR #1406, opened 2025-02-27
I-ALiRT -  decom mag

### LINKED ISSUE #1327: I-ALiRT - Decom MAG packets
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Added code to decom I-ALiRT

## New Files
- ialirt_mag.xml
   - XTCE for parsing
- mag_l0_ialirt_data.py
   - Parser for different packets
- parse_mag.py
   - Parse science, time, and status
- grouping.py
   - Groups by sequence and time
- time.py
   - Calculates time from coarse and fine time
   
## Updated Files
- ialirt.xml
   - Updated w/ new packet definitions

## Testing
- test_parse_mag.py
   - Parsing test
- test_time.py
   - Time test
 - test_grouping.py
   - Grouping test
   
Test data:
- sample_decoded_i-alirt_data.csv
- 461971383-404.bin
- 461971384-405.bin
- 461971385-406.bin
- 461971386-407.bin
- 461971387-408.bin
- 461971388-409.bin
- 461971389-410.bin
- 461971390-411.bin
- 461971391-412.bin


imap_processing/ialirt/l0/mag_l0_ialirt_data.py
```diff
@@ -0,0 +1,219 @@
+"""Dataclasses for Level 0 MAG I-ALiRT data."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+
+@dataclass
+class Packet0:
+    """Dataclass for packet 0."""
```
maxinelasp: Can you add docstrings for each attribute in the class with a short description of what each variable is?

imap_processing/ialirt/l0/mag_l0_ialirt_data.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    hkp8v5_warn = (status >> 20) & 0x01
+    hkp8v5_danger = (status >> 19) & 0x01
+    hkp8v5c_warn = (status >> 18) & 0x01
+    hkp8v5c_danger = (status >> 17) & 0x01
+    hkn8v5 = ((status >> 9) & 0xFF) << 4
+    hkn8v5_current = ((status >> 0) & 0x1FF) << 3
+    return Packet2(
+        hkp8v5_warn, hkp8v5_danger, hkp8v5c_warn, hkp8v5c_danger, hkn8v5, hkn8v5_current
+    )
```
maxinelasp: instead of defining these functions, what if you override the __init__ function within each dataclass? 
```@dataclass
class Packet2:
    """Dataclass for packet 2."""

    hkp8v5_warn: int
    hkp8v5_danger: int
    hkp8v5c_warn: int
    hkp8v5c_danger: int
    hkn8v5: int
    hkn8v5_current: int


    def __init__(status: int) -> None:
    """
    Decode packet 2.

    Parameters
    ----------
    status : int
        24-bit integer value.

    Returns
    -------
    list[int]
        List of extracted bytes.

    Notes
    -----
    Bits 23-22 → Packet Number (2-bit value)
    Bits 21-17 → Various warning/danger flags (5-bit value)
    Bits 16-9 → hkn8v5 (8-bit value)
    Bits 8-0 → hkn8v5_current (9-bit value)
    """
        self.hkp8v5_warn = (status >> 20) & 0x01
        self.hkp8v5_danger = (status >> 19) & 0x01
        self.hkp8v5c_warn = (status >> 18) & 0x01
        self.hkp8v5c_danger = (status >> 17) & 0x01
        self.hkn8v5 = ((status >> 9) & 0xFF) << 4
        self.hkn8v5_current = ((status >> 0) & 0x1FF) << 3
```
this keeps everything nicely tied together.

imap_processing/ialirt/l0/mag_l0_ialirt_data.py
```diff
@@ -0,0 +1,219 @@
+"""Dataclasses for Level 0 MAG I-ALiRT data."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
```
maxinelasp, line 8: Could you add a little context for the four packet design at the top of the file? Then, in each packet class, could you define what is _different_ about each packet?

imap_processing/ialirt/l0/mag_l0_ialirt_data.py
```diff
@@ -0,0 +1,219 @@
+"""Dataclasses for Level 0 MAG I-ALiRT data."""
```
maxinelasp: I overall like this design a lot and I think it will keep things very organized, which is great since these packets can get confusing

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,254 @@
...
+        Status data.
+    pkt_counters : xr.DataArray
+        Packet counters.
+
+    Returns
+    -------
+    combined_packets : dict
+        Decoded packets.
+    """
+    decoders = {
```
maxinelasp, line 57: nice, I like this

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,254 @@
...
+
+    for pkt_num, decoder in decoders.items():
+        status_subset = status_values[pkt_counters == pkt_num]
+        decoded_packet = decoder(int(status_subset))
+        combined_packets.update(vars(decoded_packet))
+
+    return combined_packets
+
+
+def find_groups(data: xr.Dataset) -> xr.Dataset:
```
maxinelasp: Could we get a more descriptive data name?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    Group data based on `mag_acq_tm_coarse` values.
+
+    Parameters
+    ----------
+    data : xr.Dataset
+        Packets dataset.
```
maxinelasp: All four packets? Can you elaborate?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,254 @@
...
+
+    Returns
+    -------
+    grouped_data : xr.Dataset
+        Grouped data with an additional "group" coordinate.
```
maxinelasp: It doesn't look like the data is grouped, rather that there is another variable which defines groups. So I would slightly clarify here.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def parse_packet(xarray_data: xr.Dataset) -> list[dict]:
+    """
+    Parse the MAG packets.
+
+    Parameters
+    ----------
+    xarray_data : xr.Dataset
+        Packet data.
```
maxinelasp: Again, where does this data actually come from?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,254 @@
...
+        # Get science values for each group.
+        science_values = grouped_data["mag_data"][
+            (grouped_data["group"] == group).values
+        ]
+        science_data = extract_magnetic_vectors(science_values)
+
+        # Get time values for each group.
+        time_data = get_time(grouped_data, group, pkt_counter)
+
+        mag_data.append({**status_data, **science_data, **time_data})
```
maxinelasp, line 244: Nice use of **. Does it make sense to split by primary and secondary here? Or is it ok to leave them all combined? Also, would it make sense to put all these dictionaries into a larger dataclass?

imap_processing/tests/ialirt/unit/test_parse_mag.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def test_find_groups(xarray_data):
+    """Tests the find_groups function."""
+    grouped_data = find_groups(xarray_data)
+
+    assert len(np.unique(grouped_data["mag_acq_tm_coarse"])) == len(
+        np.unique(grouped_data["group"])
+    )
+
```
maxinelasp: Any tests on the actual content of the group variable, instead of just checking length?

imap_processing/tests/ialirt/unit/test_parse_mag.py
```diff
@@ -0,0 +1,145 @@
...
+    status_values = xarray_data["mag_status"].values
+    pkt_counter = get_pkt_counter(status_values)
+    assert np.array_equal(pkt_counter, np.array([0, 1, 2, 3, 0, 1, 2, 3, 0]))
+
+
+def test_find_groups(xarray_data):
+    """Tests the find_groups function."""
+    grouped_data = find_groups(xarray_data)
+
+    assert len(np.unique(grouped_data["mag_acq_tm_coarse"])) == len(
```
maxinelasp: Maybe instead of using `grouped_data` it would be better to use `xarray_data` (for example, what if `grouped_data` was empty here?)

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: I'll leave better comments, but just for your understanding now, we get a packets file every minute. But MAG has the ability to process every 4 seconds (the data is split between 4 packets, each arriving every second). So when we process we have to break up the packets file so that each set of 4 packets that are meant to be together are "grouped". So that's what I'm doing here based on mag_acq_tm_coarse.

imap_processing/ialirt/l0/mag_l0_ialirt_data.py
laspsandoval: Will do. Basically the combination of the 4 packets = the complete dataset. So there are different things in each packet, but combined they form what we need to process.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: I'll put a better comment. The data comes from the 1-min accumulated packets file that is converted into xarray (but needs to be broken up into chunks).

imap_processing/ialirt/l0/mag_l0_ialirt_data.py
laspsandoval: That means a lot coming from you (since you are the expert on classes) :)

laspsandoval:
I got some feedback from Alastair, so hold until I make this recommended changes.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: Thanks for the suggestion! For now, I’d like to keep the structure flat as it’s easier to work with directly, and we don’t really need to split primary and secondary data. If we find that future processing would benefit from more structure, I’m happy to revisit this, but for now the flat structure works well with our downstream processing.

imap_processing/ialirt/l0/mag_l0_ialirt_data.py
laspsandoval: In progress. I reached out to Alastair.

laspsandoval: done

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,246 @@
...
+        Status data.
+
+    Returns
+    -------
+    pkt_counters : xr.DataArray
+        Packet counters.
+    """
+    # mag_status is a 24 bit unsigned field
+    # The leading 2 bits of STATUS are a 2 bit 0-3 counter
+    pkt_counter = (status_values >> 22) & 0x03
```
greglucas, line 36: Is there a reason this isn't a variable / extracted in the classes you made?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    # Convert each 24-bit value to its three constituent bytes
+    science0 = get_bytes(int(science_values[0]))
+    science1 = get_bytes(int(science_values[1]))
+    science2 = get_bytes(int(science_values[2]))
+    science3 = get_bytes(int(science_values[3]))
+
+    # Primary sensor:
+    pri_x = (science0[0] << 8) | science0[1]
+    pri_y = (science0[2] << 8) | science1[0]
```
greglucas: I find the `get_bytes()` above to be more distracting / confusing to follow than direct bit manipulation inline here. Could this be something like the following (probably wrong, but hopefully it gets the idea across)?

```
pri_x = science_values[0] >> 8
pri_y = ((science_values[0] << 16) & 0xFF00) | (science_values[1] >> 8)
```

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,246 @@
...
+    ----------
+    accumulated_data : xr.Dataset
+        Packets dataset accumulated over 1 min.
+
+    Returns
+    -------
+    mag_data : list[dict]
+        Dictionaries of the parsed data product.
+    """
+    logger.info("Parsing MAG.")
```
greglucas: Might be useful to put some additional info here like: `for time period: accumulated_data["mag_acq_tm_coarse"]`

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -0,0 +1,246 @@
...
+        accumulated_data["mag_acq_tm_coarse"],
+        accumulated_data["mag_acq_tm_fine"],
+        65535,
+    )
+
+    # Add required parameters.
+    accumulated_data["time_seconds"] = time_seconds
+    sorted_data = accumulated_data.sortby("time_seconds", ascending=True)
+    status_values = sorted_data["mag_status"]
+    pkt_counter = get_pkt_counter(status_values)
```
greglucas: You are using `status_values` in a few different places, here and in the for-loop below. I might inline this to avoid the confusion.

```suggestion
    pkt_counter = get_pkt_counter(sorted_data["mag_status"])
```

imap_processing/ialirt/utils/grouping.py
```diff
@@ -0,0 +1,100 @@
...
+
+    for group in unique_groups:
+        src_seq_ctr = grouped_data["src_seq_ctr"][
+            (grouped_data["group"] == group).values
+        ]
+        src_seq_ctr_diff = np.diff(src_seq_ctr)
+
+        # Accept group only if all diffs are 1 or -16383
+        if np.all(np.isin(src_seq_ctr_diff, [1, -16383])):
+            valid_groups.append(group)
```
greglucas: I think you can modulo arithmetic this wrap-around. Do you want to add a log/message about groups that were skipped here too if it was invalid?
```suggestion
        src_seq_ctr_diff = np.diff(src_seq_ctr) % 16384

        if np.all(src_seq_ctr_diff == 1):
            valid_groups.append(group)
```

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: We need to know the pkt_counter to use the classes that I made, i.e., the pkt_counter determines if I should use class Packet0, class Packet1, etc.

imap_processing/ialirt/l0/mag_l0_ialirt_data.py
```diff
@@ -0,0 +1,155 @@
...
+        self.hk1v5c_danger = (status >> 17) & 0x01
+        # +1.8v voltage warning flag raised
+        self.hk1v8_warn = (status >> 16) & 0x01
+        # +1.8v voltage danger flag raised
+        self.hk1v8_danger = (status >> 15) & 0x01
+        # +1.8v current warning flag raised
+        self.hk1v8c_warn = (status >> 14) & 0x01
+        # +1.8v current danger flag raised
+        self.hk1v8c_danger = (status >> 13) & 0x01
+        # Outboard (MAGo) sensor is saturated (danger)
```
torimarbois, line 48: Should this be out _bound_, not out _board_? (This could be spacecraft/instrument lingo I just don't know yet)

imap_processing/ialirt/utils/time.py
```diff
@@ -0,0 +1,29 @@
+"""Common time functions for I-ALiRT instruments."""
+
+import xarray as xr
+
+
+def calculate_time(
+    coarse_time: xr.DataArray, fin_time: xr.DataArray, conversion: int
```
torimarbois: It seems a little funky to me to leave off one letter here - can we add the 'e' to the end of 'fin'?

imap_processing/ialirt/utils/grouping.py
```diff
@@ -0,0 +1,100 @@
+"""Common grouping functions for I-ALiRT instruments."""
+
+import numpy as np
+import xarray as xr
+
+
+def filter_valid_groups(grouped_data: xr.Dataset) -> xr.Dataset:
+    """
+    Filter out groups where `src_seq_ctr` diff are not 1 or -16383.
```
torimarbois: Will `src_seq_ctr` be the appropriate parameter to use for every instrument? Also, could you add a little more to this doc string or the one at the top of the file? Like, clarifying what this function is filtering out groups _of_? Groups of data?

torimarbois (commented):
Just some clarifying questions around variable names & docstrings.

imap_processing/ialirt/utils/grouping.py
laspsandoval: Great idea!

imap_processing/ialirt/l0/mag_l0_ialirt_data.py
maxinelasp: MAG has an "inboard" and "outboard" sensor, where one is closer to the spacecraft and one is further. So this is MAG specific lingo!

laspsandoval: It says outboard in the Algorithm Document. :)

imap_processing/ialirt/utils/time.py
maxinelasp: I overall agree, this probably comes from the packet time which is `FNTIME`, but I would still agree to add an "e"

imap_processing/ialirt/utils/grouping.py
laspsandoval: Yes. It is in the spacecraft packet and so available to every instrument. I added some comments.

torimarbois (approved):

PR #1408, opened 2025-02-28
Add auto generation of calibration product integer values column from…

### LINKED ISSUE #1407: Hi L1C PSET calibration product config - Convert strings to coincidence type values
### Algorithm Description:

In order to filter DEs into calibration products, I need to convert the detector strings in the calibration product configuration into coincidence type values.

tmplummer:
… detector strings

# Change Summary

## Overview
Adds functionality to the calibration product configuration DataFrame accessor for converting the coincidence_type_list of strings to a list of values that can be used to compare to the coincidence type values stored in the L1B DE product.

## Updated Files
   - Add `_add_coincidence_values_column` function to dataframe accessor class.
   - Add function that converts a single string of detectors to a integer coincidence type value.
   - Manually access the registered `cal_prod_config` namespace in the custom `from_csv` method to force `_init_` to run.
   - Add test coverage for the two new functions in the `hi_l1c.py` module.

Closes: #1407

### DIFF
--- a/imap_processing/hi/l1c/hi_l1c.py
+++ b/imap_processing/hi/l1c/hi_l1c.py
@@ -527,6 +527,13 @@ def coincidence_type_string_to_int(coincidence_type_str: str) -> int:
     """
     Convert a coincidence type string to a coincidence type integer value.
 
+    A coincidence string is a string containing all detectors that were hit
+    for a direct event. Possible detectors include: [A, B, C1, C2]. Converting
+    the coincidence type string to a coincidence type integer value involves
+    summing the coincidence bitmap value for each detector hit. e.g. "AC1C2"
+    results in 2**3 + 2**1 + 2**0 = 11. See `CoincidenceBitmap` for the mapping
+    from detector name to integer value.
+
     Parameters
     ----------
     coincidence_type_str : str
@@ -538,6 +545,9 @@ def coincidence_type_string_to_int(coincidence_type_str: str) -> int:
     coincidence_type : int
         The integer value of the coincidence type.
     """
+    # CoincidenceBitmap defines the detector names and their associated
+    # values. Use regex to find matches to the detector names.
     pattern = r"|".join(c.name for c in CoincidenceBitmap)
     matches = re.findall(pattern, coincidence_type_str)
+    # Sum the integer value assigned to the detector name for each match
     return sum(CoincidenceBitmap[m] for m in matches)
--- a/imap_processing/hi/l1c/hi_l1c.py
+++ b/imap_processing/hi/l1c/hi_l1c.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -428,3 +446,23 @@ def number_of_products(self) -> int:
...
+    coincidence_type_str : str
+        The coincidence type string containing the list of detectors hit.
+        e.g. "AC1C2".
+
+    Returns
+    -------
+    coincidence_type : int
+        The integer value of the coincidence type.
+    """
+    pattern = r"|".join(c.name for c in CoincidenceBitmap)
```
nkerman: Will this just be "A|B|C1|C2"?

imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -428,3 +446,23 @@ def number_of_products(self) -> int:
...
+
+def coincidence_type_string_to_int(coincidence_type_str: str) -> int:
+    """
+    Convert a coincidence type string to a coincidence type integer value.
+
+    Parameters
+    ----------
+    coincidence_type_str : str
+        The coincidence type string containing the list of detectors hit.
+        e.g. "AC1C2".
```
nkerman: It'd be helpful to also comment what this example should output. I think this would be  `2**3 + 2**1 + 2**0` = 11, right?

imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -409,12 +424,15 @@ def from_csv(cls, path: Path) -> pd.DataFrame:
...
         """
-        return pd.read_csv(
+        df = pd.read_csv(
             path,
             index_col=cls.index_columns,
             converters={"coincidence_type_list": lambda s: s.split("|")},
             comment="#",
         )
+        # Force the _init_ method to run by using the namespace
+        _ = df.cal_prod_config.number_of_products
```
nkerman, line 511: I'm confused whats happening here - the class's __init__ is calling this function itself - is this about initializing the df? Would that not happen if there's no rows in the dataFrame or something?

nkerman (commented):
Looks good! Couple questions/docs requests.

imap_processing/hi/l1c/hi_l1c.py
tmplummer: Yes. I just like to not hard code things in a bunch of different places. So it is derived from the `CoincidenceBitmap` enum.

imap_processing/hi/l1c/hi_l1c.py
tmplummer: Yeah, this is a bit confusing. The `CalibrationProductConfig` doesn't subclass `pd.Dataframe` and it is a bit opaque as to how it works. It basically adds the `.cal_prod_config` namespace to a dataframe which then allows one to call the custom methods that are defined in the class. But, the `_init_` method doesn't actually get called until you try to use the namespace.

It might be that this `from_csv` classmethod doesn't really belong as a classmethod. I'm far from a pandas expert.

Documentation here: https://pandas.pydata.org/docs/development/extending.html

imap_processing/hi/l1c/hi_l1c.py
greglucas: You could add this as a method on the Enum rather than a separate function here too if that'd be helpful. I think we do something similar for the BitFlags matching the bits that are set.

tmplummer: @greglucas Yes! That makes so much more sense. Don't know why that wasn't obvious to me when I implemented this.

greglucas (approved):

PR #1409, opened 2025-03-01
ENH: XTCE: Add support for segmented polynomials

greglucas:
# Change Summary

## Overview

The previous handling of segmented polynomials only considered the first row and didn't use any of the other factors. This iterates through the other rows creating additional calibrators and adds a ContextCalibrator for each segment that checks against the low and high value specified within the spreadsheet.

This follows the `ContextCalibrator` definition from space_packet_parser and creates a structure similar to that: https://github.com/lasp/space_packet_parser/blob/102987430f28cab48f787c2a40ccfebf8c99944b/tests/unit/test_xtce/test_encodings.py#L161-L190

@vmartinez-cu, I tried running this on the packets we have in the repository, but it didn't produce any differences from what I could tell. I think that is because all of the raw values are still 0, so it is still only using the first calibration value. Do you have different data you could test this with or have something else I should try to verify this is working as you expected?

## Updated Files

imap_xtce: Added support for the segmented polynomials to output new XML elements
hit_packet_definition.xml: Ran imap_xtce on the HIT spreadsheet to create the new XTCE definition

### DIFF
--- a/imap_processing/ccsds/excel_to_xtce.py
+++ b/imap_processing/ccsds/excel_to_xtce.py
@@ -352,21 +352,64 @@ class XTCEGenerator:
         conversion = analog_conversion.loc[
             (analog_conversion["mnemonic"] == row["mnemonic"])
             & (analog_conversion["packetName"] == row["packetName"])
-        ].iloc[0]
+        ]
+        name = f"{row['packetName']}.{row['mnemonic']}"
 
-        # Create the Conversion element
-        default_calibrator = Et.SubElement(encoding, "xtce:DefaultCalibrator")
-        polynomial_calibrator = Et.SubElement(
-            default_calibrator, "xtce:PolynomialCalibrator"
-        )
-        # FIXME: Use lowValue / highValue from the conversion sheet
-        # FIXME: Handle segmented polynomials (only using first segment now)
-        for i in range(8):
-            col = f"c{i}"
-            if conversion[col] != 0:
-                term = Et.SubElement(polynomial_calibrator, "xtce:Term")
-                term.attrib["coefficient"] = str(conversion[col])
-                term.attrib["exponent"] = str(i)
+        if len(conversion) == 1:
+            # Unsegmented Polynomial
+            segment = conversion.iloc[0]
+            # Create the Conversion element
+            default_calibrator = Et.SubElement(encoding, "xtce:DefaultCalibrator")
+            polynomial_calibrator = Et.SubElement(
+                default_calibrator, "xtce:PolynomialCalibrator"
+            )
+            # FIXME: Use lowValue / highValue from the conversion sheet
+            for i in range(8):
+                col = f"c{i}"
+                if segment[col] != 0:
+                    term = Et.SubElement(polynomial_calibrator, "xtce:Term")
+                    term.attrib["coefficient"] = str(segment[col])
+                    term.attrib["exponent"] = str(i)
+        else:
+            # Segmented Polynomial
+            context_calibrator_list = Et.SubElement(
+                encoding, "xtce:ContextCalibratorList"
+            )
+            # One context calibrator for each segment
+            for _, segment in conversion.iterrows():
+                context_calibrator = Et.SubElement(
+                    context_calibrator_list, "xtce:ContextCalibrator"
+                )
+
+                # The matching criteria for this calibrator
+                # This references the parameter name itself, so we use the uncalibrated
+                # value for comparison
+                # >= lowValue and <= highValue
+                context_match = Et.SubElement(context_calibrator, "xtce:ContextMatch")
+                comparison_list = Et.SubElement(context_match, "xtce:ComparisonList")
+                comparison = Et.SubElement(comparison_list, "xtce:Comparison")
+                comparison.attrib["comparisonOperator"] = ">="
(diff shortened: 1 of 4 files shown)


vmartinez-cu:
Thanks for working on this! I emailed you housekeeping test files. The L1A and L1B output should have nonzero values for temperature fields. You're getting zeros?

Temperature fields:
`temp0`
`temp1`
`temp2`
`temp3`
`analog_temp`
`hvps_temp`
`idpu_temp`
`lvps_temp`

greglucas:
> The L1A and L1B output should have nonzero values for temperature fields. You're getting zeros?

The housekeeping file in the repository https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/hit/PREFLIGHT_raw_record_2023_256_15_59_04_apid1251.pkts
has all 0s for the raw values.

But, it looks like there is another file now: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/hit/test_data/imap_hit_l0_raw_20100105_v001.pkts

In fact, it looks like the `PREFLIGHT` files are not being referenced anywhere. Should we remove those?

imap_processing/tests/ccsds/test_data/expected_output.xml
```diff
@@ -67,6 +67,41 @@
...
 			</xtce:EnumeratedParameterType>
+			<xtce:IntegerParameterType name="TEST_PACKET.VAR_SEGMENTED" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="8" encoding="unsigned">
+					<xtce:ContextCalibratorList>
+						<xtce:ContextCalibrator>
+							<xtce:ContextMatch>
+								<xtce:ComparisonList>
+									<xtce:Comparison comparisonOperator="&gt;=" value="0" parameterRef="TEST_PACKET.VAR_SEGMENTED" useCalibratedValue="false" />
+									<xtce:Comparison comparisonOperator="&lt;=" value="19" parameterRef="TEST_PACKET.VAR_SEGMENTED" useCalibratedValue="false" />
+								</xtce:ComparisonList>
```
tech3371, line 78: I see how it's done now. Nice feature!

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -352,21 +352,64 @@ def _add_analog_conversion(self, row: pd.Series, encoding: Et.Element) -> None:
...
-        )
-        # FIXME: Use lowValue / highValue from the conversion sheet
-        # FIXME: Handle segmented polynomials (only using first segment now)
-        for i in range(8):
-            col = f"c{i}"
-            if conversion[col] != 0:
-                term = Et.SubElement(polynomial_calibrator, "xtce:Term")
-                term.attrib["coefficient"] = str(conversion[col])
-                term.attrib["exponent"] = str(i)
+        if len(conversion) == 1:
```
tech3371: I had to look through little bit to understand what this if statement means. Can you add comment that says something like this where here it seems like we are checking if we have more than one value of some key in `mnemonics` columns of `AnalogConversions` tab which indicates if something is segmented or not.

imap_processing/ccsds/excel_to_xtce.py
```diff
@@ -352,21 +352,64 @@ def _add_analog_conversion(self, row: pd.Series, encoding: Et.Element) -> None:
...
+            # Unsegmented Polynomial
+            segment = conversion.iloc[0]
+            # Create the Conversion element
+            default_calibrator = Et.SubElement(encoding, "xtce:DefaultCalibrator")
+            polynomial_calibrator = Et.SubElement(
+                default_calibrator, "xtce:PolynomialCalibrator"
+            )
+            # FIXME: Use lowValue / highValue from the conversion sheet
```
tech3371: Is this a feature that's coming from `space_packet_parser`? I saw in XTCE of this PR that it has this
```
<xtce:Comparison comparisonOperator="&gt;=" value="20" parameterRef="TEST_PACKET.VAR_SEGMENTED" useCalibratedValue="false" />
```
Is this comment related to this?

tech3371: nvm, I saw you did this in else statement. Looks like there is some difference between HIT and SWAPI. Looks like HIT set low and high values for their unsegmented keys whereas SWAPI didn't. SWAPI set both values as 0 for unsegmented keys.

tech3371 (approved):
I had unblocking questions but looks great to me. Thank you for doing this and explaining it.

imap_processing/ccsds/excel_to_xtce.py
greglucas: I removed this FIXME now and added a NOTE comment instead here about why we ignore it for unsegmented polynomials.

imap_processing/ccsds/excel_to_xtce.py
greglucas: Yeah, this was confusing. I tried to add some additional comments throughout these if/else branches and factored out the polynomial creation into a separate inline function. Let me know if this could use more explanation still.

vmartinez-cu:
I don't think I've used the PREFLIGHT files, so these can be removed.

vmartinez-cu (approved):

PR #1412, opened 2025-03-03
Ultra l1b aux quality flags

### LINKED ISSUE #1381: Ultra l1b - extended spin
### Summary

Spin number, spin rate, spin start
Compare aux packet against universal spin table

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
One flag that the Ultra team thought is important is the comparison between their aux packet and the Universal Spin Table. They should contain the same information. This PR creates a quality flag for a mismatch between the aux packet and Universal Spin Table. I also recognized some minor changes that need to happen as I was working.

## Updated Files
- ultra_l1b_culling.py
    - Updated since we don't need to interpolate based on time to get the spin number. It is already in the de_dataset
    - Created function that compares the aux dataset to the Universal Spin Table
- quality_flags.py
   - added new flag
   - added an attribute
- extendedspin.py
   - Pass in the aux packet now. 
   - Updated since we don't need to interpolate based on time to get the spin number. It is already in the de_dataset
-  ultra_l1b.py
    - Pass in aux packet
-  de.py
    - Create spin data field.
    
## Testing
- conftest.py
   - added mocked aux dataset
- test_ultra_l1b_culling.py

### DIFF
--- a/imap_processing/tests/ultra/unit/conftest.py
+++ b/imap_processing/tests/ultra/unit/conftest.py
@@ -205,13 +205,15 @@ def aux_dataset(ccsds_path_theta_0, xtce_path):
 def test_aux_dataset():
     """Fixture to compute and return aux test data."""
 
-    epoch = np.arange(760591716184000000, 760591716184000000 + 7 * 15000000000, 15000000000)
-    spin_number = np.arange(127,142)
-    spin_start_time = np.arange(0, 105, 15)
-    spin_period_sec = np.full(7, 15)
+    num_spins = 15
+
+    epoch = np.arange(760591716184000000, 760591716184000000 + num_spins * 15000000000, 15000000000)
+    spin_number = np.arange(127, 142)
+    spin_start_time = np.arange(1905, 2115 + 15, 15)
+    spin_period_sec = np.full(num_spins, 15)
     spin_period_sec[-1] = 14
-    spin_start_sec = np.arange(0, 105, 15)
-    spin_start_subsec = np.zeros(7)
+    spin_start_sec = np.arange(1905, 2115 + 15, 15)
+    spin_start_subsec = np.zeros(num_spins)
 
     test_aux_dataset = xr.Dataset(
         data_vars={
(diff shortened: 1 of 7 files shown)


imap_processing/tests/ultra/unit/conftest.py
```diff
@@ -200,20 +201,55 @@ def aux_dataset(ccsds_path_theta_0, xtce_path):
     return l1a_aux_dataset
 
 
+@pytest.fixture()
+def test_aux_dataset():
```
nkerman: I think probably only tests should be functions which start with `test_<>`. Rename to `setup_test_aux_dataset` or similar?

imap_processing/tests/ultra/unit/conftest.py
```diff
@@ -200,20 +201,55 @@ def aux_dataset(ccsds_path_theta_0, xtce_path):
...
+    """Fixture to compute and return aux test data."""
+
+    num_spins = 15
+
+    epoch = np.arange(
+        760591716184000000, 760591716184000000 + num_spins * 15000000000, 15000000000
```
nkerman: Perhaps a comment on where this comes from?

imap_processing/tests/ultra/unit/conftest.py
```diff
@@ -200,20 +201,55 @@ def aux_dataset(ccsds_path_theta_0, xtce_path):
...
+    )
+    spin_number = np.arange(127, 142)
+    spin_start_time = np.arange(1905, 2115 + 15, 15)
```
nkerman: Extract 15 out to a variable? (`spin_duration_seconds`)

nkerman:
Looks great, Laura! Couple small requests.

imap_processing/tests/ultra/unit/conftest.py
laspsandoval: Yeah. That is really confusing. I think I was trying to simulate the real data, but really this only need to be the length of the data variables. It doesn't matter what it is. Good catch. I've changed it.

nkerman (approved):
Looks great - thanks Laura!

PR #1414, opened 2025-03-04
Reorder ultra map code dimensions -> (epoch, energy, az, el)

### LINKED ISSUE #1413: Reorder ultra map code dimensions -> `(epoch, energy, az, el)`
Spinoff of PR #1396 : The Ultra/Hi/Lo PSETs are now expected to have dimensions of the order: `(epoch, energy, az, el)`, rather than the previously assumed `(az, el, epoch, energy)`.

Reorder this in the data mocking tools, and fix the binning to allow this order.

nkerman:
# Change Summary

Reorder expected L1C PSET dimensions to put spatial dims last. 

## Overview

Spinoff of PR #1396: The Ultra/Hi/Lo PSETs are now expected to have dimensions of the order: `(epoch, energy, az, el)`, rather than the previously assumed `(az, el, epoch, energy)`.

Reorders the L1C data created in the data mocking tools, and fixes the binning to allow this order.

~~NOTE:~~
---
~~I've decided that the SkyMaps themselves should keep the spatial axis (always a 1D axis of unwrapped pixel on the sky) as axis 0. Given their state as a spatially tiled object, I think that this makes the most sense, and tasks like re-wrapping the unwrapped pixels onto a 2D grid is much simpler if we can assume that the only spatial axis is axis 0.~~

~~I spent quite a long time trying to make the above task work in an dimension-order-agnostic way, and I really couldn't without lots of conditionals, etc. The Ultra/Lo/Hi L2 code will need to decide how to order the `SkyMap.data_dict = dict[NDarray]` output into their own dataset anyways, so reordering will already be happening there.~~

I ended up making this work. Both the PSET and Maps now treat the last 1-2 axes as the spatial axes. 

Close #1413

## Updated Files

  - switch dim order of mocked L1c pset
  - Fix binning for spatial_axis as last axis of PSET arrays
- imap_processing/ena_maps/utils/map_utils.py
  - Allow for spatial_axis parameter
  - Clarify docs
## Testing

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -207,6 +207,9 @@ class PointingSet(ABC):
     """
     Abstract class to contain pointing set (PSET) data in the context of ENA sky maps.
 
+    Any spatial axes - (azimuth, elevation) for Rectangularly gridded tilings or
+    (pixel index) for Healpix - must be stored in the last axis/axes of each data array.
+
     Parameters
     ----------
     dataset : xr.Dataset
@@ -362,7 +365,14 @@ class UltraPointingSet(PointingSet):
 
 # Define the Map classes
 class AbstractSkyMap(ABC):
-    """Abstract base class to contain map data in the context of ENA sky maps."""
+    """
+    Abstract base class to contain map data in the context of ENA sky maps.
+
+    Data values are stored in a dictionary, where the 0th axis is the
+    only spatial dimension. If the map is rectangular,
+    this axis is the raveled 2D grid.
+    If the map is Healpix, this axis is the 1D array of Healpix pixel indices.
+    """
 
     @abstractmethod
     def __init__(self) -> None:
@@ -384,7 +394,8 @@ class RectangularSkyMap(AbstractSkyMap):
     """
     Map which tiles the sky with a 2D rectangular grid of azimuth/elevation pixels.
 
-    NOTE: Internally, the map is stored as a 1D array of pixels.
+    NOTE: Internally, the map is stored as a 1D array of pixels,
+    with the 0th axis as the spatial axis.
 
     Parameters
     ----------
@@ -470,15 +481,17 @@ class RectangularSkyMap(AbstractSkyMap):
             )
 
         for value_key in value_keys:
-            # If multiple spatial axes present
-            # (i.e (az, el) for rectangular coordinate PSET),
-            # flatten them in the values array to match the raveled indices
-            raveled_pset_data = pointing_set.data[value_key].data.reshape(
-                pointing_set.num_points, -1
-            )
+            pset_values = pointing_set.data[value_key]
+
+            # If there is an epoch dim with size 1, flatten it out
+            if "epoch" in pset_values.dims and pset_values["epoch"].size == 1:
+                pset_values = pset_values.squeeze("epoch")
+
+            raveled_pset_data = pset_values.data.reshape(-1, pointing_set.num_points)
+
             if value_key not in self.data_dict:
                 # Initialize the map data array if it doesn't exist (values start at 0)
-                output_shape = (self.num_points, *raveled_pset_data.shape[1:])
+                output_shape = (self.num_points, *raveled_pset_data.shape[:-1])
                 self.data_dict[value_key] = np.zeros(output_shape)
 
             if index_match_method is IndexMatchMethod.PUSH:
@@ -489,12 +502,16 @@ class RectangularSkyMap(AbstractSkyMap):
                         len(self.sky_grid.el_bin_midpoints),
                     ),
                     projection_indices=matched_indices_push,
+                    spatial_axis=-1,
                 )
             else:
                 raise NotImplementedError(
                     "The 'pull' method of index matching is not yet implemented."
                 )
             self.data_dict[value_key] += pointing_projected_values
+
+            # These are not used after the loop, and can be large - Delete them.
+            del pset_values, raveled_pset_data, pointing_projected_values
 
     def __repr__(self) -> str:
         """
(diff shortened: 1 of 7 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -489,13 +502,17 @@ def project_pset_values_to_map(
...
+                    spatial_axis=-1,
                 )
             else:
                 raise NotImplementedError(
                     "The 'pull' method of index matching is not yet implemented."
                 )
             self.data_dict[value_key] += pointing_projected_values
 
+            # These are not used after the loop, and can be large - Delete them.
+            del pset_values, raveled_pset_data, pointing_projected_values
```
greglucas: These should be garbage collected if there are no external references hanging onto them after the loop, so do they need to be explicitly deleted? This makes me think there is something else going on here.

imap_processing/ena_maps/utils/map_utils.py
```diff
@@ -99,6 +107,7 @@ def bin_values_at_indices(
     projection_grid_shape: tuple[int, int],
     projection_indices: NDArray,
     input_indices: NDArray | None = None,
+    spatial_axis: int = 0,
```
greglucas: I'd like to make this just be `-1` and not allow an option to change this if possible.

greglucas (commented):
I don't fully follow the question relating to 0th axis for the SkyMaps. I think that everything should be consistent and it will make our lives much easier if our coordinates are always indexed as `[..., -1]`, rather than in one class it is the last dimension and the other class it is the first dimension.

Can you comment on where the if/else clauses would need to be added to handle the -1 case rather than the 0 version? Right now it looks like we are already doing moveaxis/transpose, so I'm not sure it would be more complicated.

imap_processing/ena_maps/ena_maps.py
nkerman: I think I just need to learn more about Python's garbage collection. Nothing weird going on here, I just didn't know they would get deleted, and I noticed they could get pretty big.

nkerman: I removed this.

imap_processing/ena_maps/utils/map_utils.py
nkerman: Removed it as a parameter.

nkerman:
> I don't fully follow the question relating to 0th axis for the SkyMaps. I think that everything should be consistent and it will make our lives much easier if our coordinates are always indexed as `[..., -1]`, rather than in one class it is the last dimension and the other class it is the first dimension.
> 
> Can you comment on where the if/else clauses would need to be added to handle the -1 case rather than the 0 version? Right now it looks like we are already doing moveaxis/transpose, so I'm not sure it would be more complicated.

I made the switch so that everything expects the spatial axis to be -1, including the binning and re-wrapping. You're right, it wasn't quite as bad as I worried. I had been imagining we needed to handle the arbitrary ordering of dimensions (e.g. `(epoch, az, el, energy)`). 

As long as we can always be sure that it's the final 1 or 2 axes, things like `array[indices, ...]` can be switched over to `array[..., indices]` easily enough. The only tricky thing was fixing all the tests which were intentionally built with hardcoded values. And I'm excited to see what this will break in the other PRs!

imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -178,9 +178,16 @@ def test_project_pset_values_to_map_push_method(self, mock_frame_transform_az_el
         assert rectangular_map.data_dict != {}
 
         # Check that the map has the same values as the PSETs, summed
-        simple_summed_pset_counts = np.sum(
-            [pset["counts"].values for pset in self.l1c_pset_products], axis=0
-        ).reshape(rectangular_map.data_dict["counts"].shape)
+        simple_summed_pset_counts = np.zeros(rectangular_map.data_dict["counts"].shape)
```
greglucas: Do you want it to be the same datatype here too?
```suggestion
        simple_summed_pset_counts = np.zeros_like(rectangular_map.data_dict["counts"])
```

imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -178,9 +178,16 @@ def test_project_pset_values_to_map_push_method(self, mock_frame_transform_az_el
...
+        for pset in self.l1c_pset_products:
+            reshaped_pset_counts = pset["counts"].squeeze("epoch")
+            # transpose to put az, el first and then reshape to the map's counts shape
+            reshaped_pset_counts = reshaped_pset_counts.transpose(
+                "energy_bin_center",
+                "azimuth_bin_center",
+                "elevation_bin_center",
+            ).data.reshape(rectangular_map.data_dict["counts"].shape)
```
greglucas: You are putting az, el at the end now. Do you need to transpose here, or are you able to just directly add these now because you are stating that everything is coming in and going out with the spatial axes last?

greglucas (approved):
I think this looks good and makes things consistent. I appreciate the thrash here to update everything, so thanks for taking the time to do all of this! I think it will help us moving forward.

imap_processing/ena_maps/ena_maps.py
greglucas: :+1: Yeah, they may not be immediately but they will eventually if you don't have anything referencing that area of memory. But note that if you have multiple things pointing to the same underlying data you may have things being kept alive longer than you think from this one function.

nkerman: Ah, got it, thanks!
 I'm planning to read/watch a tutorial of Python garbage collection over lunch!

imap_processing/tests/ena_maps/test_ena_maps.py
nkerman: I had been using `allclose`, so it wasn't explicitly necessary, but I can do this, and use `array_equal`!

imap_processing/tests/ena_maps/test_ena_maps.py
nkerman: You're right, this is no longer necessary - removed!

PR #1415, opened 2025-03-04
Add repoint module, test coverage, and general conftest fixture for g…

### LINKED ISSUE #1384: SPICE - Implement pointing table query code
### Algorithm Description:
We need to implement code to load and query the repointing table. For now, this can just be to return a pd.DataFrame. In the future, some utility functions will need to be added to do things like get the mean time for a pointing.

tmplummer:
…enerating fake repoint data on the fly.

# Change Summary

## Overview
This adds some basic repoint utility functions as listed in the New Files section.

## New Files
   - Add `get_repoint_data`function that reads a CSV pointed to by the `REPOINT_DATA_FILEPATH` environment variable.
   - Add `interpolate_repoint_data` function that interpolates the repoint data to the input query times and adds a column that indicates if the query time is during a repoint maneuver or not.
- imap_processing/tests/spice/test_data/fake_repoint_data.csv
   - This is a fake repoint CSV used for unit tests
   - Add test coverage for the new functions in the `repoint` module

## Updated Files
   - Add general fixtures for generating fake repoint data on the fly based on input times.

Closes: #1384

### DIFF
--- a/imap_processing/spice/repoint.py
+++ b/imap_processing/spice/repoint.py
@@ -1,0 +1,118 @@
+"""Functions for retrieving repointing table data."""
+
+import logging
+import os
+from pathlib import Path
+from typing import Union
+
+import numpy as np
+import pandas as pd
+from numpy import typing as npt
+
+logger = logging.getLogger(__name__)
+
+
+def get_repoint_data() -> pd.DataFrame:
+    """
+    Read repointing file using environment variable and return as dataframe.
+
+    REPOINT_DATA_FILEPATH environment variable should point to a local
+    file where the repointing csv file is located.
+
+    Returns
+    -------
+    repoint_df : pd.DataFrame
+        The repointing csv loaded into a pandas dataframe. The dataframe will
+        contain the following columns:
+            - `repoint_start_time`: Starting MET time of each repoint maneuver.
+            - `repoint_end_time`: Ending MET time of each repoint maneuver.
+            - `repoint_id`: Unique ID number of each repoint maneuver.
+    """
+    repoint_data_filepath = os.getenv("REPOINT_DATA_FILEPATH")
+    if repoint_data_filepath is not None:
+        path_to_spin_file = Path(repoint_data_filepath)
+    else:
+        # Handle the case where the environment variable is not set
+        raise ValueError("REPOINT_DATA_FILEPATH environment variable is not set.")
+
+    logger.info(f"Reading repointing data from {path_to_spin_file}")
+    repoint_df = pd.read_csv(path_to_spin_file, comment="#")
+
+    return repoint_df
+
+
+def interpolate_repoint_data(
+    query_met_times: Union[float, npt.NDArray],
+) -> pd.DataFrame:
+    """
+    Interpolate repointing data to the queried MET times.
+
+    Query times that are between a pointing end time and the next pointing start
+    time will cause an exception to be raised.
+
+    Parameters
+    ----------
+    query_met_times : float or np.ndarray
+        Query times in Mission Elapsed Time (MET).
+
+    Returns
+    -------
+    repoint_df : pandas.DataFrame
+        Repoint table data with the interpolated such that there is one row
+        for each of the queried MET times. Output columns are:
+            - `repoint_start_time`
+            - `repoint_end_time`
+            - `repoint_id`
+            - `repoint_in_progress`
+
+    Raises
+    ------
+    ValueError : If any of the query_met_times are outside the set of
+    [repoint_start_time, repoint_end_time] time ranges defined in the repoint
+    table.
+    """
+    repoint_df = get_repoint_data()
+
+    # Ensure query_met_times is an array
+    query_met_times = np.asarray(query_met_times)
+    is_scalar = query_met_times.ndim == 0
+    if is_scalar:
+        # Force scalar to array because np.asarray() will not
+        # convert scalar to array
+        query_met_times = np.atleast_1d(query_met_times)
+
+    # Make sure no query times are before the first repoint in the dataframe.
+    repoint_df_start_time = repoint_df["repoint_start_time"].values[0]
+    if np.any(query_met_times < repoint_df_start_time):
+        bad_times = query_met_times[query_met_times < repoint_df_start_time]
+        raise ValueError(
+            f"{bad_times.size} query times are before the first repoint start "
+            f" time in the repoint table. {bad_times=}, {repoint_df_start_time=}"
+        )
+    # Make sure that no query times are after the valid range of the dataframe.
+    # We approximate the end time of the table by adding 24 hours to the last
+    # known repoint start time.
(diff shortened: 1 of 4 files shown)


imap_processing/spice/repoint.py
```diff
@@ -0,0 +1,118 @@
...
+from numpy import typing as npt
+
+logger = logging.getLogger(__name__)
+
+
+def get_repoint_data() -> pd.DataFrame:
+    """
+    Read repointing file using environment variable and return as dataframe.
+
+    REPOINT_DATA_FILEPATH environment variable should point to a local
```
greglucas, line 19: Can/should we make use of `IMAP_DATA_DIR`: https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/3e3189e2b7220c50ad597b47f0bcc44a7f53ea78/imap_data_access/__init__.py#L37
and place this relative to that?

I was hoping that we could sort of have one location to rule them all with subdirectories etc. under that without growing a new environment variable for every new file.

imap_processing/spice/repoint.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Parameters
+    ----------
+    query_met_times : float or np.ndarray
+        Query times in Mission Elapsed Time (MET).
+
+    Returns
+    -------
+    repoint_df : pandas.DataFrame
+        Repoint table data with the interpolated such that there is one row
```
greglucas:
```suggestion
        Repoint table data interpolated such that there is one row
```

imap_processing/spice/repoint.py
```diff
@@ -0,0 +1,118 @@
...
+    table.
+    """
+    repoint_df = get_repoint_data()
+
+    # Ensure query_met_times is an array
+    query_met_times = np.asarray(query_met_times)
+    is_scalar = query_met_times.ndim == 0
+    if is_scalar:
+        # Force scalar to array because np.asarray() will not
+        # convert scalar to array
```
greglucas: Can you use this directly above instead of np.asarray?
`query_met_times = np.atleast_1d(query_met_times)`

imap_processing/spice/repoint.py
```diff
@@ -0,0 +1,118 @@
...
+    repoint_df_start_time = repoint_df["repoint_start_time"].values[0]
+    if np.any(query_met_times < repoint_df_start_time):
+        bad_times = query_met_times[query_met_times < repoint_df_start_time]
+        raise ValueError(
+            f"{bad_times.size} query times are before the first repoint start "
+            f" time in the repoint table. {bad_times=}, {repoint_df_start_time=}"
+        )
+    # Make sure that no query times are after the valid range of the dataframe.
+    # We approximate the end time of the table by adding 24 hours to the last
+    # known repoint start time.
```
greglucas, line 96: Can we not use the `repoint_end_time` column here?

imap_processing/spice/repoint.py
```diff
@@ -0,0 +1,118 @@
...
+    row_indices = (
+        np.searchsorted(repoint_df["repoint_start_time"], query_met_times, side="right")
+        - 1
+    )
+    out_df = repoint_df.iloc[row_indices]
+
+    # Add a column indicating if the query time is during a repoint or not.
+    # The table already has the correct row for each query time, so we
+    # only need to check if the query time is less than the repoint end time to
+    # get the same result as `repoint_start_time <= query_met_times < repoint_end_time`.
```
greglucas, line 117: Really helpful comment. Thank you!

imap_processing/spice/repoint.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def interpolate_repoint_data(
+    query_met_times: Union[float, npt.NDArray],
+) -> pd.DataFrame:
+    """
+    Interpolate repointing data to the queried MET times.
+
+    Query times that are between a pointing end time and the next pointing start
+    time will cause an exception to be raised.
```
greglucas: After reading through the function, I think you are adding another flag "repointing_in_progress" instead of raising. Do you want to set these to nan's or somehow make it more obvious something is occuring in the in-between section?

greglucas (approved):
Looks good to me, I just have some minor docs suggestions.

imap_processing/tests/conftest.py
```diff
@@ -485,3 +485,83 @@ def make_data(start_met: float, end_met: Optional[float] = None) -> pd.DataFrame
...
+def use_test_repoint_data_csv(monkeypatch):
+    """Sets the REPOINT_DATA_FILEPATH environment variable to input path."""
+
+    def wrapped_set_repoint_data_filepath(path: Path):
+        monkeypatch.setenv("REPOINT_DATA_FILEPATH", str(path))
+
+    return wrapped_set_repoint_data_filepath
+
+
+def generate_repoint_data(
```
laspsandoval, line 500: Can you see any benefit to be able to start from a repointing number other than 1?

laspsandoval (approved):
Looks good. I can't wait to use it!

imap_processing/tests/conftest.py
laspsandoval: I ask only because that would have come in handy for the universal spin table since my de dataset has spin numbers starting at 121. But probably it's ok for the repoint table.

imap_processing/spice/repoint.py
tmplummer: I think that could be done. I was thinking that the repoint table would be a processing dependency which means that it will get downloaded to some sub-directory of `IMAP_DATA_DIR`. I figured that we could then have some code in the `pre_processing` method of the `ProcessInstrument` class that identifies the spin and repoint tables if they are in the dependency list, removes them and sets the environment variable so that their location doesn't have to get passed around.

imap_processing/spice/repoint.py
tmplummer: Oh, yeah... This comment needs to be updated. When we talked about this in the tag-up, I was misunderstanding the repoint table. The `repoint_start_time` and `repoint_end_time` define the start of the repoint maneuver and end (after nutation is damped below requirement). I implemented it so that an exception is raised if the query time is `> repoint_start_time[-1] + 24 * 60 * 60`.

imap_processing/spice/repoint.py
tmplummer: See my other comment about what `repoint_end_time` means: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1415/files#r1980411720

imap_processing/tests/conftest.py
tmplummer: Sure, I can make that a parameter.

imap_processing/spice/repoint.py
tmplummer: To further elaborate, I decided that the repoint table might be used to filter data that occurs during the repoint maneuver. That is why I added the boolean column indicating any query times that are during the repoint.

PR #1421, opened 2025-03-04
Add energy table data to CoDICE-Hi L1a products

### LINKED ISSUE #809: Add energy tables to CoDICE-hi L1a data products
The CoDICE-hi L1a CDFs need data variables in them that store the energy values in addition to the counters

bourque:
This PR adds support for CoDICE-Hi energy table data in the L1a CDF products. Currently this only implements the energy data for the `hi-omni` product, but can easily be extended for the other CoDICE-Hi products. I plan to add energy data to the other CoDICE-Hi products as I validate them (#251)

Closes #809


imap_processing/codice/constants.py
```diff
@@ -968,3 +972,287 @@
     254: 7864319,
     255: 4294967294,
 }
+
+OMNI_ENERGY_TABLE = {
```
bourque, line 980: I am putting this here to be consistent with where I am storing other lookup tables. These values are not expected to change much, if at all.

imap_processing/codice/codice_l1a.py
```diff
@@ -269,45 +271,84 @@ def define_support_variables(self, dataset: xr.Dataset) -> xr.Dataset:
...
+
+                # Data quality is named differently in packet data and needs to be
+                # treated slightly differently
+                elif variable_name == "data_quality":
+                    variable_data = self.dataset.suspect.data
+                    dims = ["epoch"]
+                    attrs = self.cdf_attrs.get_variable_attributes("data_quality")
+
+                # Spin period requires the application of a conversion factor
+                # See Table B.5 in the algorithm document
```
greglucas, line 340: Thanks for documenting where this is from! Do you want to store any of these factors with your other constants to multiply it by name instead?

imap_processing/codice/constants.py
greglucas: Are all of these contiguous regions with the min[1] == max[0]? If so, it might be easier to list all of the edges here and then calculate the centers/deltas from that one consistent list: `edges[1:] - edges[:-1]`

greglucas (approved):

imap_processing/codice/codice_l1a.py
bourque: Good call. It is a 'constant' ultimately so it should be with the other constants.

imap_processing/codice/constants.py
bourque: Good idea, that will help reduce the amount of repeated values

PR #1423, opened 2025-03-05
Healpix Sky Maps

### LINKED ISSUE #1397: Create Healpix Tessellated SkyMap objects
### Algorithm Description:

In the ENA Mapping code, build a new subclass of the `SkyMap` base class to handle maps whose pixels are Healpix tessellations of the entire sky. 

### Requirements:

Must adapt/subclass existing method/s:

1. `project_pset_values_to_map`

Should also be able to project to a `RectangularSkyMap`. The Ultra Instrument team is still deciding how they want this to be done; This may just be a placeholder for now, or a simplified version. 

#### Specify if any dependencies were added
Needs `Healpy` library.

### Related Issues/PRs:
Expects "pull" method of index matching to be available, so uses PR #1396 as a starting point.

nkerman:
# Change Summary

Add SkyMap subclass which tiles the sky with a Healpix tessellation of equal-area pixels. 

Closes #1397 

## Overview

Add `HealpixSkyMap(AbstractSkyMap)` object, which works in the same way as the existing `RectangularSkyMap`, but, rather than tiling the sky with a 2D grid of pixels at discrete, equal steps in azimuth and elevation, this map tiles the whole sky with a [Healpix](https://en.wikipedia.org/wiki/HEALPix) tessellation of equal-area pixels.

## New Dependencies

1. Healpy
2. Python required is now 3.10 or higher, removing support for 3.9. 
Please see [Slack thread](https://lasp.slack.com/archives/C04L7EUMJ8G/p1740607989943879) for previous discussion and decision to drop 3.9, though I expect we might have more discussion here. 

## Updated Files

   - Add Healpix Sky Map object
   - Move `project_pset_values_to_map()` function to the Base `SkyMap` class, and slightly adapt to work with both Rectangular and Healpix maps. 
   - Pull out determination of projection grid shape for PUSH binning into `get_binning_grid_shape()` function
     - This is a small change to keep the `project_pset_values_to_map()` function from having too many conditional branches
- updated poetry.lock, pyproject.toml
   - Update dependencies, Python versions

## Testing
Add tests for new function, Healpix Maps


imap_processing/ena_maps/ena_maps.py
```diff
@@ -178,34 +180,53 @@ def match_coords_to_indices(
...
     else:
         raise ValueError(
             "Tiling type of the output frame must be either RECTANGULAR or HEALPIX."
+            f"Received: {output_object.tiling_type}"
         )
 
     return flat_indices_input_grid_output_frame
 
 
+def get_binning_grid_shape(
```
nkerman: I primarily pulled this out into a function to keep the number of conditional branches down in `project_pset_values_to_map`.

pyproject.toml
```diff
@@ -33,7 +33,7 @@ classifiers = [
 [tool.poetry.dependencies]
 cdflib = "^1.3.1"
 imap-data-access = ">=0.10.1"
-python = ">=3.9,<4"
+python = ">=3.10,<4"
```
nkerman, line 36: @maxinelasp - I fear I'm a bit out of my depth when it comes to how to handle this change with Poetry and the CI actions. When you're back from sick time, I'd love some advice! (In the meantime, hope you feel better!)

pyproject.toml
```diff
@@ -54,12 +54,14 @@ sphinx = {version="*", optional=true}
 sphinxcontrib-openapi = {version="^0.8.3", optional=true}
 mypy = {version="1.10.1", optional=true}
 requests = {version = "^2.32.3", optional = true}
+healpy = "^1.18.0"
```
greglucas: It looks like healpy may not work with Windows. Can we switch over to one of the other implementations that does?

Is this meant to be optional? If it is required, we should move it up into that section. Also, can you alphabetize this?

In general, I don't think that we **require** the bump to Python 3.10 for this. Do we need the latest and greatest from healpy versions? Otherwise you should be able to relax this version, and then if it needs an earlier numpy version our package installation procedure will determine that and get the latest supported compatible versions of everything. So if someone is installing our package (imap_processing) under Python 3.9, they may get numpy 1.25 and healpy X. But, if someone is on Python 3.13, then they could get numpy 2.2 and healpy Y for example.

nkerman: @greglucas @tech3371 and I talked about this! Takeaways below:

1. Switch to `astropy-healpix` for the code itself, and make this a standard dependency. This fixes Windows compatibility. 
2. Add a set of extra dependencies for `"map_visualization"`, which consists of only `healpy` at present. This is because `astropy-healpix` does not have the same visualization features as does `healpy`. 
3. Drop support for Python 3.9. 

@greglucas, did you say you'll be able to change the CI to stop requiring 3.9 checks? I can look into that, but I may not have the Github permissions.

greglucas: Done! I changed the required checks to be all OS' 3.10 and 3.13.

Just a note on this
> Switch to astropy-healpix for the code itself, and make this a standard dependency

We may look into making certain packages optional in the future, but for now add it everywhere to keep moving forward. We have a few packaging related issues to look at later.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -380,62 +401,21 @@ class AbstractSkyMap(ABC):
...
     @abstractmethod
     def __init__(self) -> None:
-        pass
+        self.tiling_type: SkyTilingType
+        self.sky_grid: spatial_utils.AzElSkyGrid
+        self.num_points: int
+        self.data_dict: dict[str, NDArray]
 
-    def __repr__(self) -> str:
+    def copy(self) -> AbstractSkyMap:
```
greglucas: Are we using this anywhere / do we need it? A user could do the deepcopy themselves. Also, xarray has a `deep=True` keyword to choose whether to deepcopy or not, so we should maybe use that syntax if we are going to add it: https://docs.xarray.dev/en/latest/generated/xarray.DataArray.copy.html

But, my preference is generally to leave this up to end-users / ourselves in the instruments to do a deepcopy if necessary. (Might be worth leaving this for another PR with just this feature added if so)

imap_processing/ena_maps/ena_maps.py
```diff
@@ -527,8 +507,64 @@ def project_pset_values_to_map(
...
+    def __repr__(self) -> str:
+        """
+        Return a string representation of the map.
+
+        Returns
+        -------
+        str
+            String representation of the map.
+        """
+        return f"{self.__class__} Map)"
```
greglucas: Doesn't this return `RectangularSkyMap Map)`? (duplicate Map and extra parentheses) I'd suggest removing this and letting this return the default class repr.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -543,3 +579,38 @@ def __repr__(self) -> str:
...
+
+    def __init__(
+        self, nside: int, spice_frame: geometry.SpiceFrame, nested: bool = False
+    ):
+        # Define the core properties of the map:
+        self.tiling_type = SkyTilingType.HEALPIX
+        self.spice_reference_frame = spice_frame
+        self.nside = nside
+        self.num_points = hp.nside2npix(nside)
+        pixel_az, pixel_el = np.deg2rad(
```
greglucas, line 607: I thought we were using degrees in most places for az/el? It would be good to be consistent if we can, or indicate when one is `_deg` or `_rad` when it is different to clearly put what version that variable is in.

pyproject.toml
nkerman: Fantastic, thanks Greg! All checks pass now, too. 

Agreed WRT potentially optional dependency.

imap_processing/ena_maps/ena_maps.py
nkerman: We can get away without it. I had been planning on making copies in some tests, but if necessary, we can just apply deepcopy when necessary. 

The need for copying might come back at some point, if we, say want to make a map at a certain resolution and then downsample it while keeping the original. I'll cut it for now!

imap_processing/ena_maps/ena_maps.py
nkerman: Will do! I'm having the subclasses define reprs anyway, so that they can be a bit more specific (e.g. `"HealpixSkyMap with nside=32, nested=False"`) Just added the Healpix repr.

pyproject.toml
```diff
@@ -53,12 +53,14 @@ sphinx = {version="*", optional=true}
...
 mypy = {version="1.10.1", optional=true}
 requests = {version = "^2.32.3", optional = true}
+healpy = {version = "^1.18.0", optional = true}
 
 [tool.poetry.extras]
 dev = ["pre-commit", "ruff", "mypy"]
 doc = ["numpydoc", "pydata-sphinx-theme", "sphinx", "sphinxcontrib-openapi"]
 test = ["openpyxl", "pytest", "pytest-cov", "requests"]
 tools= ["openpyxl", "pandas"]
+map_visualization= ["healpy"]
```
tmplummer: Is `healpy` strictly for visualization of maps? or is it really needed for maps in general?

imap_processing/ena_maps/ena_maps.py
tmplummer: It seems to me that this should be a property of the `PointingSet` and `SkyMap` so that you can just do: `projection_object.binning_grid_shape`.

imap_processing/ena_maps/ena_maps.py
nkerman: I'm actually keeping all values internal to the Maps as radians, which I think are probably a better standard vs. degrees. @laspsandoval, please correct me if I'm wrong here, but I think that in L1, the azimuth and elevation values mapping out the whole sphere are stored as radians. It's phi and theta, which are totally different angles, representing the FOV of the detector, that Ultra has chosen to keep defined in degrees. These sometimes get referred to as azimuth/elevation, but are a separate quantity. 

The only place where degrees should come into the maps, is when interfacing externally:
1. `spacing_deg` is the init parameter for rectangular PointingSets and SkyMaps, because it's what the instrument teams will specify their values in. E.g. "Make us maps with 2 deg spacing from our PSETs on 0.5 deg spacing."
2. `healpy` and `astropy-healpix` automatically [output degrees](https://healpy.readthedocs.io/en/latest/generated/healpy.pixelfunc.pix2ang.html) if you ask for lon/lat style coordinates rather than co-latitude and longitude in radians. (Not my favorite default, since it changes the default units when you ask for the constant shift from lat -> co-lat). 🤷 
3. When the `UltraPointingSet` object interfaces with the L1C data, it makes sure that the 2D grid of the L1C data, which is expected to be saved as degrees, though that will also go away with Healpix L1C PSETs

tmplummer (commented):
Just a couple more comments that might want to be changed before merging.

pyproject.toml
maxinelasp: LGTM, I did end up updating some stuff in this PR to work with 3.13 btw #1435

pyproject.toml
nkerman: We put `astropy-healpix` as a general dependency, because it's Windows compatible, and this as an extra for visualization. I haven't run into any limits of what `astropy-healpix` can do as relates to building maps. 

Healpy is in there as an extra so we can easily plot healpix tessellations.

pyproject.toml
nkerman: Thanks Maxine - taking a look.

imap_processing/ena_maps/ena_maps.py
nkerman: I thought about that, but I figured that as a separate function it operates on any kind of spatial object (e.g. on a PointingSet as well - it only cares if they're Healpix or Rectangular tiling.)

imap_processing/ena_maps/ena_maps.py
laspsandoval: The imap_ultra_l1b_variable_attrs.yaml has azimuth and elevation parameters that are old and I need to delete them. For the cdf I think leadership recently decided on degrees. It doesn't matter to me what the code uses as long as we are consistent.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -539,7 +552,57 @@ def __repr__(self) -> str:
...
+    spice_frame : geometry.SpiceFrame
+        The reference Spice frame of the map.
+    nested : bool, optional
+        Whether the Healpix tessellation is nested. Default is False.
+    """
+
+    def __init__(
+        self, nside: int, spice_frame: geometry.SpiceFrame, nested: bool = False
+    ):
+        # Define the core properties of the map:
```
laspsandoval, line 591: Just so it is a bit easier to follow here, could you add a single line comment above each of these? 

For example:
```
# Tiles the sky using a HEALPIX grid
self.nside = nside
self.num_points = hp.nside2npix(nside)
```

```
# Stores Pixel Locations in Azimuth and Elevation Coordinates
pixel_az, pixel_el = np.deg2rad(
```

```
# Tracks Per-Pixel Solid Angle
self.solid_angle = hp.nside2pixarea(nside, degrees=False)
self.solid_angle_points = np.full(self.num_points, self.solid_angle)
```

```
# Supports Data Storage per Pixel
self.data_dict: dict[str, NDArray] = {}
```

```
# Sets Resolution and Pixel Count Based on nside
self.num_points = hp.nside2npix(nside)
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -539,7 +552,57 @@ def __repr__(self) -> str:
...
+        self.nside = nside
+        self.num_points = hp.nside2npix(nside)
+        pixel_az, pixel_el = np.deg2rad(
+            hp.pix2ang(
+                nside=nside, ipix=np.arange(self.num_points), nest=nested, lonlat=True
+            )
+        )
+        self.az_el_points = np.column_stack((pixel_az, pixel_el))
+        self.solid_angle = hp.nside2pixarea(nside, degrees=False)
+        self.solid_angle_points = np.full(self.num_points, self.solid_angle)
```
laspsandoval, line 620: If all pixels for HEALPIX have the same solid_angle (aka area), then is there a reason to track them?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -527,8 +495,53 @@ def project_pset_values_to_map(
...
+            spacing_deg=self.spacing_deg,
+        )
+
+        # Solid angles of each pixel in the map grid in units of steradians
+        self.solid_angle_grid = spatial_utils.build_solid_angle_map(
+            spacing_deg=self.spacing_deg,
+        )
+
+        # Unwrap the az, el, solid angle grids to series of points tiling the sky
+        az_points = self.sky_grid.az_grid.ravel()
```
laspsandoval, line 542: If you could add more comments that would help a bit for me to understand everything a bit better.

laspsandoval (commented):
Really great job. A few comments/suggestions.

imap_processing/ena_maps/ena_maps.py
nkerman: Just trying to keep it as consistent as possible with the RectangularSkyMap - that will allow functions (say ultra L2 which might have to scale by a solid angle ratio) work with both Healpix/Rectangular tilings.

imap_processing/ena_maps/ena_maps.py
nkerman: Added this, thanks for the suggestion!

imap_processing/ena_maps/ena_maps.py
nkerman: Sure! I've added more comments (and reordered a few lines) in the `RectangularSkyMap` __init__ as well as the Healpix one. 

Just as a note, this is not new code, despite the green highlight. Git seems to have gotten confused by me moving the `project_pset_values_to_map` method from the `RectangularSkyMap` to the `AbstractSkyMap` parent class.

imap_processing/ena_maps/ena_maps.py
nkerman: > The imap_ultra_l1b_variable_attrs.yaml has azimuth and elevation parameters that are old and I need to delete them.

I didn't realize those won't get used anymore. 

I don't really know if one is better, but I feel "safer" working in radians internally, since both numpy and the spice tools default to radians. But @greglucas and @laspsandoval, I can change it if you'd like. I'm just looking to pick one and use it consistently within the map code.

imap_processing/ena_maps/ena_maps.py
tmplummer: Oh, I see. This is beyond this PR, but this is a good opportunity to use pythons multiple inheritance. It seems like `RectangularSkyMap` should inherit `AzElSkyGrid` as should any `PointingSet` that has a rectangular sky grid (Lo does, Hi does although it is a 1x3600 rectangle). Then `HealPixSkyMap` as well as the `UltraPointingSet` can inherit some `HealPixPixels` object. Then, the `AzElSkyGrid` and `HealPixPixels` classes implement a `binning_grid_shape` method. I think that will work but I don't know if it will limit flexibility in some way.

Overall, it just seems weird to me to have this external function providing this functionality rather than embracing the OO design that we are building.

nkerman: Using multiple inheritance is a cool idea I never considered, and might be good for the rectangularly gridded objects! 

To me, it seems to be adding a fair amount of complexity in this case, especially if we need to implement and think about `HealPixPixel` objects or similar. Ultimately, **all** we need to do here is say:

```python
if tiling_type is Rectangular:
    binning_grid_shape = (360/spacing_deg, 180/spacing_deg)
elif tiling_type is Rectangular:
    binning_grid_shape = (num_points)
```

from within the `AbstractSkyMap.project_pset_values_to_map()` method. That conditional seems perfectly simple enough to live within that method, but `ruff/mypy` don't like having a bunch of simple conditionals within the same function. I only abstracted this function to avoid having to add a `#noqa <too many conditional branches>`, and I put it as a separate function to mirror our decision with the freestanding `match_coords_to_indices()` function. 

How about I get rid of this function and just define a `binning_grid_shape` attribute for the `SkyMap` subclasses, as you first suggested? There's no reason I can currently think of that we would need to get this quantity for a `PointingSet`, and as @greglucas would probably remind me, it's better not to build in flexibility until we know of a reason it is/might soon be necessary.

greglucas: Nice, Nat! I like your thinking about keeping it simpler at first and then expanding later :)

It looks like this is only called in one place currently, so adding this as a method on the class seems like it would make sense. You could also make it a classmethod to kind of namespace it onto an object without needing `self`.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -527,8 +495,89 @@ def project_pset_values_to_map(
...
+    |  |15, -75|45, -75|75, -75|105, -75|...|255, -75|285, -75|315, -75|345, -75|
+    |
+    ---------------------------------------------------------------> Azimuth (degrees)
+    Elevation (degrees)
+
+    Fig. 1: Example of a rectangular grid of pixels in azimuth and elevation coordinates
+    in degrees, with a spacing of 30 degrees. There will be 12 azimuth bins and 6
+    elevation bins in this example, resulting in 72 pixels in the map.
+
+    NOTE: Internally, the map is stored as a 1D array of pixels, and all data arrays
```
greglucas: There is a `Notes` section in numpydoc which might be good here? Or in rest I think you can add a `.. note::` to give a note highlight nearby this too.
https://numpydoc.readthedocs.io/en/latest/format.html#notes

imap_processing/ena_maps/ena_maps.py
```diff
@@ -527,8 +495,89 @@ def project_pset_values_to_map(
...
+    |  |3|9 |15|21|27|33|39|45|51|57|63|69|
+    |  |2|8 |14|20|26|32|38|44|50|56|62|68|
+    |  |1|7 |13|19|25|31|37|43|49|55|61|67|
+    |  |0|6 |12|18|24|30|36|42|48|54|60|66|
+    ---------------------------------------> Azimuth
+    Elevation
+
+    Fig. 2: The 1D indices of the pixels in Fig. 1.
+    Note that the indices are raveled from the 2D grid of (az, el) such that as one
+    increases in pixel index, elevation increments first, then azimuth.
```
greglucas, line 519: This is somewhat of an implementation detail and isn't apparent to the user though right? i.e. as a user I'm always referencing my 2d grid values, even though internally here you are going to the 1d flattened values. (similar to the degrees / radians conversation from earlier where it'd be nice to always interface with one as a user if possible)

imap_processing/ena_maps/ena_maps.py
greglucas: We discussed this on a call today and decided on doing this in a follow-up PR.

greglucas (commented):
@nkerman it looks like you need to rebase this to get the latest main changes. Feel free to leave out Python 3.13 for now if that is a pain and we can add it in a follow-up PR.

imap_processing/ena_maps/ena_maps.py
nkerman: Good point - I moved this into a Notes section in the numpydoc style.

imap_processing/ena_maps/ena_maps.py
nkerman: >This is somewhat of an implementation detail and isn't apparent to the user though right?

Yes, and no. The values of the maps pixels are stored internally as this 1D pixel grid. 
Consider a 1˚ spacing map, where you're projecting counts, which have an additional energy axis (in the PSET, they're defined as an array with dims `(energy, longitude, latitude)`) the map stores these as `data_dict['counts'] = NDArray with shape (<num_energy_bins>, <360 * 180 =64,800> spatial pixels)`. 

In my backburner Ultra L2 code (the next PR, after the angular changes), I'm experimenting with a "`to_dataset()`" method of the maps, which automatically wraps the unwrapped grids. But I think that will be left to the instrument-specific L2 codes, and a user who is actually interfacing with the SkyMaps will be working with these unwrapped grids. I'm in favor of keeping, and clarifying this note. 

If I manage to get `to_dataset` working so cleanly that a user need never interface with the unwrapped `data_dict`, then we can get rid of this!

greglucas: Sounds good. I'm also just fine leaving this note here, I was just making sure I was following the interfaces. This all makes sense to me, and I agree that would be great if we can abstract it to never interface with the 1D flattened coords, but keep that as a separate PR :+1:

imap_processing/ena_maps/ena_maps.py
nkerman: I think it's even easier than that - This can't change for a map, so it should just be a ~~property~~ **attribute** of the map itself. It's a `tuple[int, ...]` of length 2 for rectangular grids and 1 for healpix tessellations, and it can be defined permanently in instantiation. I've just made this change!

imap_processing/ena_maps/ena_maps.py
nkerman: Thanks Greg! It makes more sense in the proper Notes section, so thanks for the suggestion.

greglucas (approved):

PR #1424, opened 2025-03-05
1151 hi l1c   exposure times

### LINKED ISSUE #1151: Hi L1C - Exposure times
### Algorithm Description:

Calculate exposure times for pset bins.

tmplummer:
# Change Summary

## Overview
This PR adds the function that computes the Hi PSET exposure times.

## New Files

- new file 1

## Deleted Files
- imap_processing/tests/hi/data/l1/imap_hi_l1b_45sensor-de_20250415_v999.cdf
   - Although this file is small and removal doesn't remove the history stored in the repo, it does mean that any future changes don't get tracked in the repo.

## Updated Files
   - Replace "deg" units in hi_pset_spin_angle_bin variable attributes with empty string as suggested by SPDF documentation for unitless variables.
   - Remove `calibration_prod` dimension from `pset_exposure_times`. Exposure times are consistent across all calibration products.
   - Use global variables to define spin-bins
   - Add function that computes PSET exposure times
   - Add placeholder function `good_time_and_phase_mask` that will eventually filter DE clock ticks as well as DEs that are not in "goodtimes"
- imap_processing/hi/utils.py
   - Make `variable_shape` optional to be consistent with underlying `full_dataarray` function.
   - Add imap_hi_l1b_45sensor-de_20250415_v999.cdf test file to external data files
   - Add test coverage for new `pset_exposure_times` function

Closes: #1151


imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -8,22 +8,35 @@
...
 )
+from imap_processing.spice.spin import (
+    get_instrument_spin_phase,
+    get_spin_data,
+)
 from imap_processing.spice.time import ttj2000ns_to_et
 
+SPIN_PHASE_BIN_CENTERS = np.arange(0.05, 360, 0.1) / 360
+N_SPIN_BINS = SPIN_PHASE_BIN_CENTERS.size
+SPIN_PHASE_BIN_EDGES = np.linspace(0, 360, N_SPIN_BINS + 1) / 360
```
greglucas: A few minor comments:

These could be subtly off using arange / linspace, see the warning note here: https://numpy.org/doc/2.1/reference/generated/numpy.arange.html#numpy.arange

I'd suggest calculating one from the other to make sure it is consistent.

```suggestion
N_SPIN_BINS = 3600
SPIN_PHASE_BIN_EDGES = np.linspace(0, 360, N_SPIN_BINS + 1) / 360
SPIN_PHASE_BIN_CENTERS = (SPIN_PHASE_BIN_EDGES[:-1] + SPIN_PHASE_BIN_EDGES[1:) / 2
```

imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -300,6 +314,78 @@ def pset_geometry(pset_et: float, sensor_str: str) -> dict[str, xr.DataArray]:
...
+    # Generate exposure time variable filled with zeros
+    exposure_var = create_dataset_variables(
+        ["exposure_times"],
+        coords=pset_coords,
+        att_manager_lookup_str="hi_pset_{0}",
+        fill_value=0,
+    )
+
+    # find_second_de_packet_data only uses variables that have epoch
+    # as a coordinate, so the others are dropped from what gets passed in.
```
greglucas: Could this be handled in the `find_second_de_packet_data()` function so you can pass the full dataset in and then let it do whatever sorting/filtering it needs?

imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -300,6 +314,78 @@ def pset_geometry(pset_et: float, sensor_str: str) -> dict[str, xr.DataArray]:
...
+        good_mask = good_time_and_phase_mask(clock_tick_mets, spin_phases)
+        spin_phases = spin_phases[good_mask]
+        clock_tick_weights = clock_tick_weights[good_mask]
+
+        # TODO: Account for flyback time. See alg doc section 2.3.5
+
+        # Bin exposure times into spin-phase bins
+        new_exposure_times, _ = np.histogram(
+            spin_phases, bins=SPIN_PHASE_BIN_EDGES, weights=clock_tick_weights
+        )
```
greglucas, line 382: Looking at your below, I'm curious if this could be done in-place somehow rather than creating these histograms each time? Something like switching the inplace add to your coordinate lookup and increment.

`np.add.at(exposure_var["exposure_times"].values[:, i_esa], ...)`

This may be faster to create these new objects rather than indices though, so this is just purely curiosity for me and no need to update/change it.

imap_processing/hi/l1c/hi_l1c.py
tmplummer: Cool, that seems obvious in hindsight.

imap_processing/hi/l1c/hi_l1c.py
tmplummer: Are you suggesting getting the spin-bin indices for each exposure time and then doing the `np.add.at()`? Sort of deconstructing histogram code to achieve the in-place binning?

When I was first implementing this, I was going to generate all of exposure clock ticks for a full day and do a 2D histogram on spin-phase and esa_step, but then I estimated the memory requirement for an array containing times for a full day worth of 2ms clock ticks and decided that the for loop is justified here. I know that isn't quite what you are suggesting but is related to performance.

greglucas: > Are you suggesting getting the spin-bin indices for each exposure time and then doing the np.add.at()? Sort of deconstructing histogram code to achieve the in-place binning?

I think so. Basically you need to figure out which indices to add the values "at", and could use the arrays you already have for where to put those values. It may be that histogram is actually faster in this situation though.

> When I was first implementing this, I was going to generate all of exposure clock ticks for a full day and do a 2D histogram on spin-phase and esa_step, but then I estimated the memory requirement for an array containing times for a full day worth of 2ms clock ticks and decided that the for loop is justified here. I know that isn't quite what you are suggesting but is related to performance.

That makes sense. That is great to hear you're thinking about these neat things too :) That comment might be worth putting in somewhere so people realize you've put some thought into this later on.

imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -8,22 +8,35 @@
...
 )
+from imap_processing.spice.spin import (
+    get_instrument_spin_phase,
+    get_spin_data,
+)
 from imap_processing.spice.time import ttj2000ns_to_et
 
+N_SPIN_BINS = 3600
+SPIN_PHASE_BIN_EDGES = np.linspace(0, 360, N_SPIN_BINS + 1) / 360
+SPIN_PHASE_BIN_CENTERS = SPIN_PHASE_BIN_EDGES[:-1] + SPIN_PHASE_BIN_EDGES[1] / 2
```
greglucas: I think this may be wrong, or at least confusing to me at first glance... Sorry if I typed it out wrong and you used the suggested change feature!

```suggestion
SPIN_PHASE_BIN_CENTERS = (SPIN_PHASE_BIN_EDGES[:-1] + SPIN_PHASE_BIN_EDGES[1:]) / 2
```

imap_processing/hi/l1c/hi_l1c.py
```diff
@@ -276,8 +290,8 @@ def pset_geometry(pset_et: float, sensor_str: str) -> dict[str, xr.DataArray]:
     el = 0 if "90" in sensor_str else -45
     dps_az_el = np.array(
         [
-            np.arange(0.05, 360, 0.1),
-            np.full(3600, el),
+            SPIN_PHASE_BIN_CENTERS * 360,
```
greglucas, line 293: Why do we need to multiply by 360 here? Above your SPIN_PHASE_BIN_CENTERS are already in degrees I believe.
```suggestion
            SPIN_PHASE_BIN_CENTERS,
```

imap_processing/hi/l1c/hi_l1c.py
tmplummer: Oh, I see. You want to take the mean of the two edges for each bin. That is a more robust way. I was counting on the fact that `SPIN_PHASE_BIN_EDGES[1]` defines the width of a single bin since `SPIN_PHASE_BIN_EDGES[0] == 0`.

imap_processing/hi/l1c/hi_l1c.py
tmplummer: No, from above: `SPIN_PHASE_BIN_EDGES = np.linspace(0, 360, N_SPIN_BINS + 1) / 360`. Maybe it just confuses things to not just have `SPIN_PHASE_BIN_EDGES = np.linspace(0, 1, N_SPIN_BINS + 1)`?

This is something I will add to the agenda for my next Hi tag-up. Paul uses the term `spin-phase` to mean fraction of a spin. So, `spin_phase = spin_angle_degrees / 360`. That is what he has asked for throughout the algorithm document for L1. I don't know if Hi is the only instrument using spin-phase or not. Seems like yet another area to get consistency across instruments.

greglucas: OK, this makes sense to me, and I just didn't realize which one was in 0-1 and which one was 0-360. Not reading close enough on my part!

> Seems like yet another area to get consistency across instruments.

Yes, great idea!

greglucas (approved):

PR #1428, opened 2025-03-05
SWE: L1 and L2 CDF attrs fixes identified by SKTEditor

### LINKED ISSUE #1427: SWE: Fix CDF attrs error
SWE L1 and L2 CDF files were failing to load in SKTEditor. This issue fixes those issue.

tech3371:
# Change Summary
closes #1427 

## Overview
Fixing issues identified by SKTEditor. I sent this file to Andriy for review and will make new PR with his feedback.

## Updated Files
- l1b and l2 CDF attr yaml file
   - updates attrs
- l1 and l2 code 
   - update to skip schema check to avoid adding additional attributed needed

### DIFF
--- a/imap_processing/swe/l1b/swe_l1b_science.py
+++ b/imap_processing/swe/l1b/swe_l1b_science.py
@@ -568,14 +568,14 @@ def swe_l1b_science(l1a_data: xr.Dataset, data_version: str) -> xr.Dataset:
         met_to_ttj2000ns(center_time),
         name="epoch",
         dims=["epoch"],
-        attrs=cdf_attrs.get_variable_attributes("epoch"),
+        attrs=cdf_attrs.get_variable_attributes("epoch", check_schema=False),
     )
 
     esa_step = xr.DataArray(
         np.arange(swe_constants.N_ESA_STEPS),
         name="esa_step",
         dims=["esa_step"],
-        attrs=cdf_attrs.get_variable_attributes("esa_step"),
+        attrs=cdf_attrs.get_variable_attributes("esa_step", check_schema=False),
     )
 
     # NOTE: LABL_PTR_1 should be CDF_CHAR.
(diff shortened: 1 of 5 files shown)


imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -568,59 +568,68 @@ def swe_l1b_science(l1a_data: xr.Dataset, data_version: str) -> xr.Dataset:
         met_to_ttj2000ns(center_time),
         name="epoch",
         dims=["epoch"],
-        attrs=cdf_attrs.get_variable_attributes("epoch"),
+        attrs=cdf_attrs.get_variable_attributes("epoch", check_schema=False),
```
greglucas, line 571: Why do we need to remove this check now? Do we need to update something upstream in sammi-cdf to have check_schema work as expected there with skt-editor?

imap_processing/swe/l2/swe_l2.py
```diff
@@ -330,42 +330,48 @@ def swe_l2(l1b_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
...
-        attrs=cdf_attributes.get_variable_attributes("inst_el_label"),
+        attrs=cdf_attributes.get_variable_attributes(
+            "inst_el_label", check_schema=False
+        ),
     )
 
     # Spin Angle bins storing bin center values.
     inst_az_xr = xr.DataArray(
-        np.arange(0, 360, 12) + 6,
+        np.arange(0.0, 359.99, 12) + 6,
```
greglucas: Is this to try and get the right dtype? If so, maybe putting `np.arange(0, 360, 12, dtype=float) + 6` would be more explicit.

greglucas (approved):
Looks good. I wonder if there is a way to flag/catch these earlier? Were you running these through the GUI or with the skteditor cli, because I'm curious if we should set up an automated skteditor check on some of our files as a CI step...

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: We still need it. To your both comments, we still need to validate some of these through SKTEditor because both `cdflib` and `sammi-cdf` does most of it but it needs little more work catch those specifics. For example, dimensions attrs doesn't require `epoch` dimension but everything else does. We also need to add checks for format, min, value or fillval matches and so on.

tech3371: `sammi-cdf` has implemented CDF validator using API that SPDF provided that is supposed to act like SKTEditor. I have not looked close. I compare that to SKTEditor and right now, SKTEditor seems to give more specifics error.

PR #1429, opened 2025-03-05
add segmented polynomial handling to convert_raw_to_eu()

### LINKED ISSUE #1426: Add segmented polynomial handling
Update function convert_raw_to_eu() to handle segmented polynomials. Update IDEX conversion csv to contain the polynomial segments for each raw DN range.

lacoak21:
# Change Summary

closes #1426

## Overview
Update function convert_raw_to_eu() to handle segmented polynomials. Update IDEX conversion csv to contain the polynomial segments for each raw DN range.

## Updated Files
   - Make sure to only unpack each instrument setting once to avoid needless repetition. 
- imap_processing/idex/idex_variable_unpacking_and_eu_conversion.csv
   - Add the polynomial coeffs for each dn range. 
   - Fix test to use uppercase UNITS instead of units.
- imap_processing/tests/test_utils.py
   - Added coverage for segmented polynomials
- imap_processing/utils.py
   - Add segmented polynomial handling. 

## Testing
Added coverage for segmented polynomials


imap_processing/utils.py
```diff
@@ -103,33 +117,70 @@ def convert_raw_to_eu(
...
                 # Convert the raw value to engineering unit
                 dataset[mnemonics].data = np.polyval(
                     coeff_values, dataset[mnemonics].data
                 )
                 # Modify units attribute
                 if "unit" in row:
-                    dataset[mnemonics].attrs.update({"units": row["unit"]})
+                    dataset[mnemonics].attrs.update({"UNITS": row["unit"]})
+            except KeyError:
+                # TODO: Don't catch this error once packet definitions stabilize
```
greglucas: Our packet definitions should be quite stable at this point, are we running into issues still?

greglucas (commented):
I was a bit confused following the logic here, and just wanted to bring up an alternative inversion of the loop logic.

Right now, I believe it is going through each row in the analog spreadsheet, seeing if there are any dataset variables matching that and applying the calibration. What about alternatively iterating through the dataset variables, then subset the pandas dataframe based on that match because you could filter based on packetName and mnemonic right away without having to do that while loop.

Overall, I think this looks great. It is unfortunate we can't use https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1409 logic here and we sort of have to recreate it.

imap_processing/utils.py
lacoak21: OK great I can remove that comment. I havent noticed any issues but I saw that someone left this #TODO above as well and thought there might still be a reason for it.

lacoak21:
That was a great idea, Greg, with the inverted loop logic and hopefully it makes the code more readable.

greglucas (approved):
Wonderful, I find this logic easier to follow, thank you for switching that up :)

PR #1434, opened 2025-03-06
XTCE - Strip spaces from strings

vmartinez-cu:
This PR makes a minor change to excel_to_xtce.py to remove leading and trailing spaces from state conversion string values

## Updated Files
imap_processing/ccsds/excel_to_xtce.py


greglucas (approved):
Could you add a space to one of the ENUM values in our test: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/5db8b740cc495e43a4c3fd22ab87a126726829f5/imap_processing/tests/ccsds/test_excel_to_xtce.py#L206

That way we can verify this is getting stripped if someone updates it later too.

greglucas (approved):

PR #1435, opened 2025-03-06
Updating tests to require Python 3.13

maxinelasp:
# Change Summary
All the checks are failing because they are required to test for Python 3.13, but the test workflow file doesn't run tests for 3.13. So, this adds the tests for 3.13.


greglucas (approved):
Ahh, Thank you @maxinelasp! I updated this for @nkerman's PR where it was failing because we were requiring 3.9 and didn't realize we weren't already testing 3.13 everywhere and it was only on his new PR.

https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1423#discussion_r1982617092

maxinelasp:
@greglucas I assumed that someone updated the check for healpix but I didn't know the tests got updated too! This is blocking my merge, so is this PR still ok? Also, should I remove the tests for 3.9 since the checks aren't testing it anymore?

greglucas:
> This is blocking my merge, so is this PR still ok?

Yes for sure! I love these small PRs fixing things rather than tying it into more places.

> Also, should I remove the tests for 3.9 since the checks aren't testing it anymore?

If you want to you can, but just adding another test should be fine. But, the 3.9 tests are still running it is just that they aren't "required to pass" anymore. So there is a required-to-pass check in GitHub where some tests can fail and you can merge over it. Let me just change it back to 3.12 for the easiest solution of all.

pyproject.toml
```diff
@@ -43,7 +43,7 @@ sammi-cdf = "^1.0"
 # Optional dependencies
 numpydoc = {version="^1.5.0", optional=true}
 openpyxl = {version=">=3.0.7", optional=true}
-pandas = {version=">=2.0.0", optional=true}
+pandas = {version=">=2.2.3", optional=true}
```
greglucas, line 46: This is a pretty strict pin, do you know if you can just remake the poetry.lock file without bumping this version?

Also curious, does bumping this "optional" pandas dependency also influence our xarray dependence on pandas to be >= to this, or can it resolve without the optional if we don't include pandas in our extras?

maxinelasp: Well, if we are supporting 3.13, then it is a hard requirement to use the version of pandas compatible with 3.13. I believe if pandas is not included as an extra, then it has no influence on the dependency resolution.

greglucas: Yes, but if I want to run on Python 3.9, I can choose to install pandas 2.0 for example. Python 3.13 people can use later if they want, but we still support older versions too.

maxinelasp:
Windows 3.13 is failing with a segfault. I did some googling but couldn't find anything. I can already tell this is going to be a horrible time-sink so I'm just going to close this ticket. Nat's version of these changes seems to update some minor version somewhere and isn't seeing this issue, so we'll let his PR update this.

PR #1439, opened 2025-03-06
Ultra l1b - remove aux dependency

laspsandoval:
# Change Summary

## Overview
Removes aux dependency for l1 de data product.

## Updated Files
- ultra_l1a.py
   - Removes aux dependency required to process Ultra de l1a
- ultra_l1b.py
   - Changed dependencies for producing l1b data products
- conftest.py
   - Created l1b extended spin dataset separately from de dataset
- ultra_l1_utils.py
   - Added dimension "epoch" because I had to

## Tests
- test_de.py
- test_ultra_l1a.py
- test_ultra_l1b.py

### DIFF
--- a/imap_processing/ultra/l1b/extendedspin.py
+++ b/imap_processing/ultra/l1b/extendedspin.py
@@ -10,37 +10,34 @@ from imap_processing.ultra.l1b.ultra_l1b_culling import (
 from imap_processing.ultra.utils.ultra_l1_utils import create_dataset
 
 
-def calculate_extendedspin(  # noqa: PLR0913
-    aux_dataset: xr.Dataset,
-    hk_dataset: xr.Dataset,
-    rates_dataset: xr.Dataset,
-    de_dataset: xr.Dataset,
+def calculate_extendedspin(
+    dict_datasets: dict[str, xr.Dataset],
     name: str,
     data_version: str,
+    instrument_id: int,
 ) -> xr.Dataset:
     """
     Create dataset with defined datatypes for Extended Spin Data.
 
     Parameters
     ----------
-    aux_dataset : xarray.Dataset
-        Dataset containing l1a aux data.
-    hk_dataset : xarray.Dataset
-        Dataset containing l1a hk data.
-    rates_dataset : xarray.Dataset
-        Dataset containing l1a rates data.
-    de_dataset : xarray.Dataset
-        Dataset containing l1b de data.
+    dict_datasets : dict
+        Dictionary containing all the datasets.
     name : str
         Name of the dataset.
     data_version : str
         Version of the data.
+    instrument_id : int
+        Instrument ID.
 
     Returns
     -------
     extendedspin_dataset : xarray.Dataset
         Dataset containing the data.
     """
+    aux_dataset = dict_datasets[f"imap_ultra_l1a_{instrument_id}sensor-aux"]
+    de_dataset = dict_datasets[f"imap_ultra_l1b_{instrument_id}sensor-de"]
+
     extendedspin_dict = {}
     rates_qf, spin, energy_midpoints, n_sigma_per_energy = flag_spin(
         de_dataset["spin"].values,
--- a/imap_processing/ultra/l1b/ultra_l1b.py
+++ b/imap_processing/ultra/l1b/ultra_l1b.py
@@ -23,6 +23,13 @@ def ultra_l1b(data_dict: dict, data_version: str) -> list[xr.Dataset]:
     -------
     output_datasets : list[xarray.Dataset]
         List of xarray.Dataset.
+
+    Notes
+    -----
+    General flow:
+    1. l1a data products are created (upstream to this code)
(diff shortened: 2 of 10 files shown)


imap_processing/ultra/l1b/extendedspin.py
```diff
@@ -11,7 +11,8 @@
 from imap_processing.ultra.utils.ultra_l1_utils import create_dataset
 
 
-def calculate_extendedspin(
+def calculate_extendedspin(  # noqa: PLR0913
+    aux_dataset: xr.Dataset,
     hk_dataset: xr.Dataset,
     rates_dataset: xr.Dataset,
     de_dataset: xr.Dataset,
```
nkerman: If you wanted to get around the flag, you could pass in a dict of 
`input_datasets = {"aux": , "hk":, ...}`

imap_processing/ultra/l1b/ultra_l1b.py
```diff
@@ -27,20 +27,24 @@ def ultra_l1b(data_dict: dict, data_version: str) -> list[xr.Dataset]:
     output_datasets = []
     instrument_id = 45 if any("45" in key for key in data_dict.keys()) else 90
```
nkerman, line 36: Maybe add some comments on the control flow?

1. If the direct events are present, _always_ calculate the L1B Direct events dataset
2. If the direct events, housekeeping, rates, and aux data are present,...

---
Also, I think this `elif` presents an issue: even if all 4 L1A datasets are present, won't this satisfy the first if, and never run the elif?

Analogy:

```python
a=100
if a:
    print("if runs")
elif a>1:
    print("elif runs")
else:
    print("else runs")

>>>if runs
```

imap_processing/ultra/utils/ultra_l1_utils.py
```diff
@@ -36,6 +36,7 @@ def create_dataset(
         coords = {
             "spin_number": data_dict["spin_number"],
             "energy_bin_geometric_mean": data_dict["energy_bin_geometric_mean"],
+            "epoch": data_dict["spin_start_time"],
```
nkerman, line 40: Does this need to be the mean time, rather than the start?

imap_processing/ultra/l1b/ultra_l1b.py
laspsandoval: Thanks for your comment. It is always good to double check these things. And I will add better docstrings. Here is what I expect to happen:

1. All l1a data products are created (upstream to this code)
2. l1b de is created and dropped in s3 kicking off processing again
3. l1b extended (+ culling, +badtimes) created

So I think the part you might have overlooked from your comment "4 L1A datasets are present"  is that extendedspin data product is dependent on l1b de NOT l1a de:
        f"imap_ultra_l1a_{instrument_id}sensor-hk" in data_dict
        **and f"imap_ultra_l1b_{instrument_id}sensor-de" in data_dict**
        and f"imap_ultra_l1a_{instrument_id}sensor-rates" in data_dict
        and f"imap_ultra_l1a_{instrument_id}sensor-aux" in data_dict

laspsandoval: Screenshot 2025-03-10 at 8 56 51 AM

laspsandoval: So this is what we have in AWS for dependencies (you will add l2 dependencies here too). We have a dependency checker so the batch job will kick off if all upstream dependencies are available.

imap_processing/ultra/utils/ultra_l1_utils.py
laspsandoval: Since the universal spin table uses spin start I think we should keep it as it is to be consistent or else it might be even more confusing than it already is. My preference would really be to eliminate the epoch coordinate completely since it isn't needed, but this was a hack so it complies with our write_cdf function without errors or having to make a bunch of changes to it. So, really not ideal, I admit.

imap_processing/ultra/l1b/ultra_l1b.py
nkerman: Thanks Laura for the detailed response! That's exactly what happened - I hadn't noticed that it was an l1**b** in the case of `f"imap_ultra_l1b_{instrument_id}sensor-de"`.

Thanks for adding more documentation - some form of the big picture steps you wrote out would be perfect as the high level explanation. 

>All l1a data products are created (upstream to this code)
l1b de is created and dropped in s3 kicking off processing again
l1b extended (+ culling, +badtimes) created

imap_processing/ultra/utils/ultra_l1_utils.py
nkerman: Perfect, well explained. Maybe add a comment noting why it must be start time to align with the universal spin table.

nkerman (approved):
Looks good! Just a couple docs requests remaining. Approving now.

PR #1443, opened 2025-03-07
MNT: Removing ISTP check for l1 data files

### LINKED ISSUE #1442: MNT: Remove ISTP check for L1 files
Based on discussion on Slack, remove ISTP checks for L1 data files.

tech3371:
# Change Summary
closes #1442

## Overview
Removing ISTP checks for L1 files

## Updated Files
   - code to remove ISTP checks for L1 files

## Testing
-  imap_processing/tests/idex/test_idex_l1a.py
    - remove bad CDF test for IDEX l1a

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -135,12 +135,22 @@ def write_cdf(
         ]
 
     # Convert the xarray object to a CDF
-    xarray_to_cdf(
-        dataset,
-        str(file_path),
-        terminate_on_warning=True,
-        **extra_cdf_kwargs,
-    )  # Terminate if not ISTP compliant
+    if "l2" in data_level:
+        # Terminate if not ISTP compliant
+        xarray_to_cdf(
+            dataset,
+            str(file_path),
+            terminate_on_warning=True,
+            **extra_cdf_kwargs,
+        )
+    else:
+        # Skip ISTP check for L1 files
+        xarray_to_cdf(
+            dataset,
+            str(file_path),
+            istp=False,
+            **extra_cdf_kwargs,
+        )
 
     return file_path
 
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -3,7 +3,7 @@
 import logging
 import re
 from pathlib import Path
-from typing import Optional
+from typing import Any, Optional
 
 import imap_data_access
 import numpy as np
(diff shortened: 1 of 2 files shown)


vmartinez-cu (approved):
This is great, thanks for doing this!

vmartinez-cu:
It looks like some checks failed

lacoak21:
Looks good to me once the checks pass!

imap_processing/cdf/utils.py
```diff
@@ -135,12 +135,22 @@ def write_cdf(
         ]
 
     # Convert the xarray object to a CDF
-    xarray_to_cdf(
-        dataset,
-        str(file_path),
-        terminate_on_warning=True,
-        **extra_cdf_kwargs,
-    )  # Terminate if not ISTP compliant
+    if "l2" in data_level:
```
greglucas: I'd invert this and only do something different for l1. We also want to leave l3 alone too.

Additionally, rather than two separate calls can you update the dictionaries being passed to xarray_to_cdf. This will allow you to keep the tests we have and add the keyword `istp=True` to the write_cdf() call. I hesitate to remove tests.

```python
if "l1" in data_level:
    if "istp" not in extra_cdf_kwargs:
        extra_cdf_kwargs["istp"] = False
else:
    if "terminate_on_warning" not in extra_cdf_kwargs:
        extra_cdf_kwargs["terminate_on_warning"] = True
    if "istp" not in extra_cdf_kwargs:
        extra_cdf_kwargs["istp"] = True

xarray_to_cdf(dataset, str(file_path), **extra_cdf_kwargs)
```

tech3371: Yeah, that sounds good if they decides to use this for L3. I remember that they were not using this for L3 but would be nice to give them this option. Cool. I will make this change.

imap_processing/cdf/utils.py
```diff
@@ -135,12 +137,16 @@ def write_cdf(
...
     # Convert the xarray object to a CDF
-    xarray_to_cdf(
-        dataset,
-        str(file_path),
-        terminate_on_warning=True,
-        **extra_cdf_kwargs,
-    )  # Terminate if not ISTP compliant
+    if "l1" in data_level:
+        if "istp" not in extra_cdf_kwargs:
+            extra_cdf_kwargs["istp"] = False  # type: ignore
```
tech3371, line 142: Added ignore this error
```
map_processing/cdf/utils.py:140: error: Incompatible types in assignment (expression has type "bool", target has type "dict[Any, Any]")  [assignment]
```

greglucas: Weird. why is bool not a part of Any 🤷 

Did you have to add the `dict[Any, Any]` or could you just leave it alone above on entry into the function?

greglucas (approved):

imap_processing/cdf/utils.py
tech3371: Yeah, mypy <sigh>

PR #1446, opened 2025-03-08
HIT L1B: Fix key error

vmartinez-cu:
This PR fixes a key error when the housekeeping APID isn't found. A test to check for this case was also added.

### DIFF
--- a/imap_processing/hit/l1b/hit_l1b.py
+++ b/imap_processing/hit/l1b/hit_l1b.py
@@ -54,7 +54,7 @@ def hit_l1b(dependencies: dict, data_version: str) -> list[xr.Dataset]:
         datasets_by_apid = get_datasets_by_apid(packet_file, derived=True)
         # TODO: update to raise error after all APIDs are included in the same
         #  raw files. currently science and housekeeping are in separate files.
-        if datasets_by_apid[HitAPID.HIT_HSKP]:
+        if HitAPID.HIT_HSKP in datasets_by_apid:
             # Process housekeeping to L1B.
             l1b_datasets.append(
                 process_housekeeping_data(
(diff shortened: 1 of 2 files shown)


imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -54,7 +54,7 @@ def hit_l1b(dependencies: dict, data_version: str) -> list[xr.Dataset]:
         datasets_by_apid = get_datasets_by_apid(packet_file, derived=True)
         # TODO: update to raise error after all APIDs are included in the same
         #  raw files. currently science and housekeeping are in separate files.
-        if datasets_by_apid[HitAPID.HIT_HSKP]:
+        if HitAPID.HIT_HSKP in datasets_by_apid:
```
greglucas, line 57: Do you want to warn/raise or log anything if the housekeeping isn't present? I'm slightly confused by the TODO above this because I think that is what you're saying is happening (we got science apids but not housekeeping).

greglucas (approved):

sdhoyt (approved):

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: This is a temporary workaround since housekeeping and science APIDs are in separate L0 files with the same naming convention (`imap_hit_l0_raw`). If HIT uploads a science L0 file to the S3 bucket to run their tests, I don't want L1B processing to completely fail because it doesn't find the housekeeping APID. L1B processing should continue to the next `if` statement to run science processing on the L1A counts CDF dependency file. Eventually, when we get L0 files with both APIDs included, if the housekeeping APID is missing, then that would be a real error to raise, which the TODO is a reminder to fix. I could raise a warning instead, but it's not technically an error at the moment, and I would need to update the warning to an error down the road anyway

Issue #1449, opened 2025-03-10
Spice Geometry: Change default angular unit to `degrees=True`


nkerman:
Some of the `spice.geometry` functions which input/output angles can work in either degrees or radians. Their behavior is determined by the `degrees=bool` flag. Some but not all of these functions currently have this flag default to True. Change all to default to True, and fix other processing code already calling these functions.

nkerman:
Closed in #1450

PR #1450, opened 2025-03-10
Use degrees internally for ENA Maps

### LINKED ISSUE #1448: ENA Maps: Replace internal angles in radians with degrees
Big Picture: ENA maps had been internally working in radians - switch to degrees. 
---

Notes from [SDC Tagup - 2025/03/10](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/248013628/SDC+Tagup+-+2025+03+10) below:

### Angular units
Degrees was decided as our common unit to use with datasets
All output datasets in degrees, loaded into internal xarray datasets as degrees as well (no magic conversion to radians here)
When geometry functions need radians, the conversion should be internal to those and not leak back out
np.cos(np.deg2rad(x))
returns should all be in degrees as well, so there may be 2 conversions, on input and output
quantities with units of streradians or divided by some angular radian should be kept in those units and conversion happens when needed.
The calculation of sterradians would be require np.deg2rad(x), the input would be degrees, but the output is a sterradian because that is a useful unit in and of itself

nkerman:
# Change Summary

1. Changes internal working angular unit to degrees in ena_maps.py and spatial_utils.py
2. Changes default angular unit to degrees in spice.geometry

Closes #1448 #1449

## Overview

At the [SDC Tagup - 2025/03/10](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/248013628/SDC+Tagup+-+2025+03+10), we decided to use internal angular units of degrees for the ENA maps, to better mirror our input/output products. 

## Updated Files

   - Changed default angular units in angular functions. 
   - Internally store angles as degrees in `build_spatial_bins()` function and `AzElSkyGrid` class. 
- imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
   - calls `build_spatial_bins()`, so remove deprecated angular conversion to degrees
-  imap_processing/ena_maps/ena_maps.py
   - Switch to internally representing angles as degrees. 

## Testing

- Fix broken tests for all the above files.

### DIFF
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -268,7 +268,7 @@ def frame_transform_az_el(
     spherical_coords_in = np.array(
         [np.ones_like(az_el[..., 0]), az_el[..., 0], az_el[..., 1]]
     ).T
-    from_frame_cartesian = spherical_to_cartesian(spherical_coords_in, degrees=degrees)
+    from_frame_cartesian = spherical_to_cartesian(spherical_coords_in)
     # Transform to to_frame
     to_frame_cartesian = frame_transform(et, from_frame_cartesian, from_frame, to_frame)
     # Convert to spherical and extract azimuth/elevation
@@ -425,14 +425,14 @@ def cartesian_to_spherical(
         - r : Distance of the point from the origin.
         - azimuth : angle in the xy-plane
           In degrees if degrees parameter is True (by default):
-          output range=[0, 360],
+          output range=[0, 360) degrees,
           otherwise in radians if degrees parameter is False:
-          output range=[0, 2*pi].
+          output range=[0, 2*pi) radians.
         - elevation : angle from the xy-plane
           In degrees if degrees parameter is True (by default):
-          output range=[0, 180],
+          output range=[-90, 90) degrees,
           otherwise in radians if degrees parameter is False:
-          output range=[-pi/2, pi/2].
+          output range=[-pi/2, pi/2) radians.
     """
     # Magnitude of the velocity vector
     magnitude_v = np.linalg.norm(v, axis=-1, keepdims=True)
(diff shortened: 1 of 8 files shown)


laspsandoval (approved):
Looks good to me. Thanks for working on it.

imap_processing/spice/geometry.py
```diff
@@ -472,7 +472,7 @@ def spherical_to_cartesian(spherical_coords: NDArray, degrees: bool = False) ->
         - elevation : angle from the xy-plane in radians [-pi/2, pi/2].
```
laspsandoval: Change this docstring to degrees.

imap_processing/spice/geometry.py
```diff
@@ -457,7 +457,7 @@ def cartesian_to_spherical(
     return spherical_coords
```
laspsandoval, line 457: Change the spherical_to_cartesian function:

One thing to check. I think the output range should be -90, 90:

 - elevation : angle from the xy-plane
          In degrees if degrees parameter is True (by default):
          output range=[0, 180],

laspsandoval: Do we need the radians option in the spherical_to_cartesian function?

laspsandoval (commented):
I found a few things after I approved.

imap_processing/spice/geometry.py
nkerman: Done, thanks for noticing that.

imap_processing/spice/geometry.py
nkerman: >I think the output range should be -90, 90:

Correct - I must have missed that in a previous PR. Fixed!

nkerman: >Do we need the radians option in the spherical_to_cartesian function?

I suppose not anymore - I removed it!

tmplummer (commented):
I'm a bit late to the party, but I thought that the idea was to get rid of all of these keywords that allow for radians to be output from various functions. This will simplify function signatures and force conversion to radians to be explicit.

nkerman:
> I'm a bit late to the party, but I thought that the idea was to get rid of all of these keywords that allow for radians to be output from various functions. This will simplify function signatures and force conversion to radians to be explicit.

Ah sorry @subagonsouth, my understanding was that everything should default to degrees but not necessarily to remove the keyword. Would you like to see these dropped in a follow up PR?

PR #1451, opened 2025-03-11
I-ALiRT - SWE Algorithm Part 1

### LINKED ISSUE #1329: I-ALiRT - SWE Algorithm
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
The beginning of the SWE Algorithm (Algorithm Doc 24-26 midway). Right before calculation of counter-streaming.

## New Files
- process_swe.py
   - Applies the SWE algorithm up to page 26.5 

## Testing
- test_process_swe.py


imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    logger.info("Processing SWE.")
+
+    # Calculate time in seconds
+    time_seconds = combine_acquisition_time(
+        accumulated_data["swe_acq_sec"], accumulated_data["swe_acq_sub"]
+    )
+    accumulated_data["time_seconds"] = time_seconds
+
+    grouped_data = find_groups(accumulated_data, (0, 59), "swe_seq", "time_seconds")
```
tech3371, line 202: Can you add comment about what (0, 59) mean?

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+    unique_groups = np.unique(grouped_data["group"])
```
tech3371, line 203: Can you add comment here also for readability? If I understand correctly, it seems like you are getting total full cycle data available for processing.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+                f"Group {group} does not contain all values from 0 to "
+                f"59 without duplicates."
+            )
+            continue
+        # Prepare raw counts array just for this group
+        # (8 energy steps, 7 CEMs, 30 phi bins)
+        group_mask = grouped_data["group"] == group
+        grouped = grouped_data.sel(epoch=group_mask)
+
+        # Split into Q1 & Q2 (swe_seq 0-29) and Q3 & Q4 (swe_seq 30-59)
```
tech3371, line 222: These comments are helpful!

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+
+def prepare_raw_counts(grouped: xr.Dataset, cem_number: int = 7) -> NDArray:
+    """
+    Reformat raw counts into a 3D array binned by phi.
+
+    Parameters
+    ----------
+    grouped : xr.Dataset
+        Dataset containing grouped i-ALiRT packet data for 1 minute.
+    cem_number : int
```
tech3371, line 74: can you import this from constants too? It's `N_CEMS` is what you want from there.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+        - 7 corresponds to the 7 CEM detectors.
+        - 30 corresponds to the 30 phi bins.
+    """
+    raw_counts = np.zeros((8, cem_number, 30), dtype=np.uint8)
+
+    # Compute phi values and their corresponding bins
+    phi_values = np.array(
+        [
+            (12 + 24 * grouped["swe_seq"].values) % 360,  # Energy steps 0 and 1
+            (24 + 24 * grouped["swe_seq"].values) % 360,  # Energy steps 2 and 3
```
tech3371, line 98: can you add little comment? This is the I-ALiRT specifics and I didn't get chance to read that section yet.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+        ]
+    )
+    phi_bins = phi_to_bin(phi_values).astype(int)  # Get phi bin indices
```
tech3371, line 101: I don't know if you need this function since it's doing one operation. Do you plan to use this in other places in the future? Right now, it's this line that uses. I would remove the function and move that line here.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # Energy bin lookup table (indexed by quarter cycle)
+    energy_bins = np.array(
+        [
+            [1, 5, 3, 7],  # 0-14 (first quarter cycle)
+            [2, 6, 0, 4],  # 15-29 (second quarter cycle)
+            [3, 7, 1, 5],  # 30-44 (third quarter cycle)
+            [0, 4, 2, 6],  # 45-59 (fourth quarter cycle)
```
tech3371: I believe it's in this order.
```suggestion
            [1, 5, 7, 3],  # 0-14 (first quarter cycle)
            [2, 6, 4, 0],  # 15-29 (second quarter cycle)
            [3, 7, 5, 1],  # 30-44 (third quarter cycle)
            [0, 4, 6, 2],  # 45-59 (fourth quarter cycle)
```

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def prepare_raw_counts(grouped: xr.Dataset, cem_number: int = 7) -> NDArray:
+    """
+    Reformat raw counts into a 3D array binned by phi.
+
+    Parameters
+    ----------
+    grouped : xr.Dataset
+        Dataset containing grouped i-ALiRT packet data for 1 minute.
```
tech3371: Looks like you are passing only 30 seconds data.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+        [
+            [1, 5, 3, 7],  # 0-14 (first quarter cycle)
+            [2, 6, 0, 4],  # 15-29 (second quarter cycle)
+            [3, 7, 1, 5],  # 30-44 (third quarter cycle)
+            [0, 4, 2, 6],  # 45-59 (fourth quarter cycle)
+        ]
+    )
+
+    # The first 15 seconds is the first quarter cycle, etc.
+    quarter_cycles = np.floor(grouped["swe_seq"] / 15).astype(int)
```
tech3371, line 114: What does diving by 15 gives back? As I noted in above comment, looks like you are passing first two or last two quarters in this function. I don't know what this line is doing.

tech3371: I imagine that first two quarters seq values would be [0,29] and [30, 59] for last two quarters. Diving that by 15 still confuse me. Lot of this confusion could be me being not great with numpy tools. @greglucas  is great with these. It might be good to get second check on this besides me.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+        e2 = grouped[f"swe_cem{cem}_e2"].values
+        e3 = grouped[f"swe_cem{cem}_e3"].values
+        e4 = grouped[f"swe_cem{cem}_e4"].values
+
+        # Phi bins 0, 2...(12, 36, ...)
+        raw_counts[e_bins[:, 0], cem - 1, phi_bins[0]] = e1
+        raw_counts[e_bins[:, 1], cem - 1, phi_bins[0]] = e2
+        # Phi bins 1, 3...(24, 48, ...)
+        raw_counts[e_bins[:, 2], cem - 1, phi_bins[1]] = e3
+        raw_counts[e_bins[:, 3], cem - 1, phi_bins[1]] = e4
```
tech3371, line 129: I will need to look at these closely once I understand what you are doing above.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+    ----------
+    grouped : xr.Dataset
+        Dataset containing grouped i-ALiRT packet data for 1 minute.
+    cem_number : int
+        Number of CEMs (default 7).
+
+    Returns
+    -------
+    raw_counts : NDArray
+        Raw counts with shape (8, 7, 30).
```
tech3371, line 80: can you change these to be what they are? FYI, `decompress_counts()` function thinks last one is `n_energy_steps`. Can you double check that?
```suggestion
        Raw counts with shape (n_time, n_cem, n_phi).
```

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -0,0 +1,246 @@
...
+        counts_second_half = decompress_counts(raw_counts_second_half)
+
+        # Apply the deadtime correction
+        # acq_duration = 80 milliseconds
+        corrected_first_half = deadtime_correction(counts_first_half, 80 * 10**3)
+        corrected_second_half = deadtime_correction(counts_second_half, 80 * 10**3)
+
+        # Grab the latest calibration factor
+        in_flight_cal_df = read_in_flight_cal_data()
+        latest_cal = in_flight_cal_df.sort_values("met_time").iloc[-1][1::]
```
tech3371, line 241: I would double check this with Ruth. I am wondering if you or she would want to use times if there exists time in in-flight cal file that covers `met_time`, otherwise use last row time like you have to extrapolate. If so, you can use `np.searchsorted()` function and `np.clip()` to do that. I had to use those two together a lot to do stuff like that. That's why I know those two now.

tech3371 (commented):
I was able to follow most of your code. Nice job! There were some I-ALiRT specifics that I got stuck on and added comments for those.

imap_processing/ialirt/l0/process_swe.py
laspsandoval: Good catch.

imap_processing/ialirt/l0/process_swe.py
laspsandoval: Example:
swe_seq_number = np.array(range(60))
quarter_cycles = np.floor(swe_seq_number / 15).astype(int)

So in this case quarter_cycles =  array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1,
       1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
       2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3])
       
Where the first 15 values are always in the first quarter cycle, second 15 in the second quarter cycle, etc.

imap_processing/ialirt/l0/process_swe.py
laspsandoval: Example:

cem = 0 (first sensor)
phi = 12 degrees (first 0.5 second - 2 measurements taken)
energy bins = 1, 5 because this is the first quarter cycle and the first two measurements (first 0.5 second)

So the energy bins are different depending on which quarter cycle and which 0.5 second.

laspsandoval: But it took me about 1 hour to just sit and think about what was going on and how to organize it. So we can talk more about it in-person if that helps.

imap_processing/ialirt/l0/process_swe.py
laspsandoval: Sounds good. Ruth had said grab the latest, but I'll double check and see if we want to do something more sophisticated.

imap_processing/ialirt/l0/process_swe.py
laspsandoval: I found a typo in decompress_counts(). The raw_counts have dimensions (n_energy_step, n_cem, n_phi).

The shape shouldn't matter here, though, I think, for decompression?

decompression_table = np.array([decompressed_counts(i) for i in range(256)])
counts = decompression_table[raw_counts]

imap_processing/ialirt/l0/process_swe.py
laspsandoval: I here you. I do kind of like it to be separate because it helps with testing. Could I leave it for now? Is that ok?

imap_processing/ialirt/l0/process_swe.py
laspsandoval: Fixed.

tech3371 (approved):
LGTM code wise. I don't know too well with I-ALiRT stuff. Sorry if I missed anything.

PR #1455, opened 2025-03-12
Mag L1C first pass

maxinelasp:
# Change Summary
For MAG L1C, we take in two input files (one norm and one burst) and fill any gaps in the norm data by interpolating the burst data. This PR completes most of MAG L1C.

## Overview

Added in MAG L1C processing and initial unit tests. As part of this, I have included a configuration file and some options for different methods of interpolating data (which are not fully completed yet.) MAG has requested this primarily for pre-launch investigation, with the expectation that this will rarely (if ever) change post-commissioning. Therefore, this configuration file will be kept in the repo rather than as an ancillary file.

## Testing

This has my unit tests, with some additional validation data expected next week from MAG. That validation will be included in a separate PR.


imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,443 @@ def mag_l1c(
...
+        output_logical_source = input_logical_source_1.replace("l1b", "l1c")
+    elif "norm" in input_logical_source_2 and "burst" in input_logical_source_1:
+        normal_mode_dataset = second_input_dataset
+        burst_mode_dataset = first_input_dataset
+        output_logical_source = input_logical_source_2.replace("l1b", "l1c")
+
     else:
-        raise RuntimeError("Neither input dataset is norm data")
+        raise RuntimeError(
+            "L1C requires one normal mode and one burst mode input " "file."
```
alastairtree, line 66: There will be times, like during commissioning, when MAG is in burst mode all day, and the entire day needs to be downsampled. And in that case you will still need to load the previous days (30m before midnight?) data so that you can maintain the regular timestamps/frequency of the generated BM as the data transitions over midnight, e.g. previous day last vector is at 100ms to midnight means next days vector is at 400s after, keeping them 500ms apart.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,443 @@ def mag_l1c(
...
+        end_gap. Start_gap and end_gap both correspond to points in timeline_data.
+
+    Returns
+    -------
+    full_timeline: numpy.ndarray
+        Completed timeline.
+    """
+    # Generated timestamps should always be 0.5 seconds apart
+    # TODO: is this in the configuration file?
+    difference_ns = 0.5 * 1e9
```
alastairtree, line 492: Can we have a chat about how hard it would be to look at the earlier NM in the day, and just carry on at thre same rate, if there was any, and then if you cannot work out the existing rate then just having this as a fallback?

maxinelasp: Yes, added to the agenda for our tag-up

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: Let's discuss this during our tag-up. The previous day loading piece should be fine, but it will be difficult to trigger processing if we sometimes expect a file and sometimes don't, so we need to discuss what that dependency should look like.

imap_processing/mag/imap_mag_sdc-configuration_v001.yaml
```diff
@@ -0,0 +1,6 @@
+# This file describes configuration changes that are not expected to change often,
+# but may change.
+
+# interpolation method used in L1C processing
+L1C_interpolation_method: "linear"
```
alastairtree, line 5:
```suggestion
L1C_interpolation_method: "linear_filtered"
```
This is our current prefered default

imap_processing/mag/l1c/interpolation_methods.py
```diff
@@ -0,0 +1,211 @@
...
+    output_timestamps : numpy.ndarray
+        Output timestamps of shape (m,) to generate interpolated vectors for.
+
+    Returns
+    -------
+    numpy.ndarray
+        Interpolated vectors of shape (m, 3) where m is equal to the number of output
+        timestamps. Contains x, y, z components of the vector.
+    """
+    pass
```
alastairtree, line 122: Oh I see - still waiting on an implementation!

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,443 @@ def mag_l1c(
...
+    try:
+        global_attributes["is_mago"] = normal_mode_dataset.attrs["is_mago"]
+        global_attributes["is_active"] = normal_mode_dataset.attrs["is_active"]
+        global_attributes["missing_sequences"] = normal_mode_dataset.attrs[
+            "missing_sequences"
+        ]
+    except KeyError as e:
+        logger.info(
+            f"Key error when assigning global attributes, attribute not found in "
+            f"L1A file: {e}"
```
alastairtree: Think you mean not found in L1B file here? Might also be handy to add the filename.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,443 @@ def mag_l1c(
...
+    interpolation_function : InterpolationFunction
+        The interpolation function to use to fill in the gaps.
+
+    Returns
+    -------
+    np.ndarray
+        An (n, 8) shaped array containing the completed timeline.
+    """
+    norm_epoch = normal_mode_dataset["epoch"].data
+    vecsec_attr = normal_mode_dataset.attrs["vectors_per_second"]
```
alastairtree, line 225: `vecsec_attr ` seems to be a string? The instrument could be in different rate normals modes in a given day, and different rates per sensor so not sure how this data is going to flow here?

alastairtree (commented):
Have only reviewed about half of this but looks like you have made great progress! I will try and do ythe rest tomorrow, but should be plenty with this and the validation data to kepp you busy (sorry!). Thank you!

imap_processing/mag/l1c/interpolation_methods.py
maxinelasp: Yes - this will be added in the next PR, but we prefer to keep them as small as possible so I left it out here!

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: The vecsec_attr is setup to look like `timestamp_1:vecsec,timestamp_2:vecsec` so it can track different rates within a day.

imap_processing/mag/l1c/mag_l1c.py
alastairtree: Yup lets discuss in person but my initial thought is that the dependency has to be on both, when either has changed. 
- If the L1B burst mode file changed/arrived then that should trigger the generation of L1C because BM means there is a gap in NM. If the BM arrives before the NM, then you can still make NM from the BM that you have. This could also be the case where say the instrument was only on for 1h in BM, and off the rest of the day.
- If the L1B NM has changed then the L1C NM needs to be generated, as either a copy of the L1B, or with any available BM filling in the gaps.

imap_processing/tests/mag/test_mag_l1c.py
```diff
@@ -1,40 +1,292 @@
...
+    output_dataset["vectors"].data = vectors
+
+    return output_dataset
+
+
+@pytest.fixture()
+def norm_dataset():
+    dataset = mag_l1a_dataset_generator(10)
+    epoch_vals = np.array([0, 0.5, 1, 1.5, 2, 4, 4.25, 5.5, 5.75, 6]) * 1e9
+    vectors_per_second_attr = "0:2,4000000000:4"
```
alastairtree, line 42: I would be tempted to have a method like `generate_science_epochs(how_many_seconds, vectors_per_second)` and a constant say `FOUR_VECTORS_PER_SECOND = 4` to make very explicit what you are using to test and reduce the duplication accross tests (from experience you end up using it a lot) and then you can generate any length of data at any cadence.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,443 @@ def mag_l1c(
...
+            "direction_label", check_schema=False
+        ),
+    )
+
+    compression_label = xr.DataArray(
+        compression.values.astype(str),
+        name="compression_label",
+        dims=["compression_label"],
+        attrs=attribute_manager.get_variable_attributes(
+            "compression_label", check_schema=False
```
vmartinez-cu, line 122: Is `check_schema=False` temporary or is this recommended for L1 products?

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: To follow up: We have decided that both normal and burst mode will be soft dependencies, and if there is a gap at the beginning of the day in normal mode data, processing will attempt to get the previous days L1C file to create a continuous timeline. Both will be finished in a future PR.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: This will be a change in a future PR to carry through the known vectors per second rate in the generated timeline.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: Fixed and added logical source, which should be enough to determine the L1B file used in processing here. (The filename is not passed through at this point.)

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: Since this is a metadata attribute, it is permanent until SAMMI gets a feature fix to accommodate the lower requirements for metadata attributes. (i.e. this is a label and therefore doesn't have DEPEND_0, for example.)

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,444 @@ def mag_l1c(
...
+    # TODO: fill with FILLVAL?
+    filled_timeline: np.ndarray = np.zeros((len(new_timeline), 8))
+    filled_timeline[:, 0] = new_timeline
+    # Flags, will also indicate any missed timestamps
+    filled_timeline[:, 5] = -1
+
+    for index, timestamp in enumerate(normal_dataset["epoch"].data):
+        timeline_index = np.searchsorted(new_timeline, timestamp)
+        filled_timeline[timeline_index, 1:5] = normal_dataset["vectors"].data[index]
+        filled_timeline[timeline_index, 5] = 0
```
alastairtree:
```suggestion
        filled_timeline[timeline_index, 5] = L1C_DATA_ORIGIN.FROM_NORMAL_MODE
```
Be tempted to put some of the magic strings and integers into constants/enums to make the intention a bit clearer to the reader

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,444 @@ def mag_l1c(
...
+        An (n, 8) shaped array containing the timeline filled with normal mode data.
+        Gaps are marked as -1 in the generated flag column at index 5.
+        Indices: 0 - epoch, 1-4 - vector x, y, z, and range, 5 - generated flag,
+        6-7 - compression flags.
+    """
+    # TODO: fill with FILLVAL?
+    filled_timeline: np.ndarray = np.zeros((len(new_timeline), 8))
+    filled_timeline[:, 0] = new_timeline
+    # Flags, will also indicate any missed timestamps
+    filled_timeline[:, 5] = -1
```
alastairtree:
```suggestion
    filled_timeline[:, 5] = L1C_DATA_ORIGIN.NO_DATA
```

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,444 @@ def mag_l1c(
...
+        An array of shape (n, 8) containing the fully filled timeline.
+        Indices: 0 - epoch, 1-4 - vector x, y, z, and range, 5 - generated flag,
+        6-7 - compression flags.
+    """
+    burst_epochs = burst_dataset["epoch"].data
+    # Exclude range values
+    burst_vectors = burst_dataset["vectors"].data
+
+    for gap in gaps:
+        # TODO: we might need a few inputs before or after start/end
```
alastairtree, line 318: Yeah linear filtered needs data before and after the current point

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,444 @@ def mag_l1c(
...
+    """
+    gaps: np.ndarray = np.zeros((0, 2))
+    if vectors_per_second_attr is not None and vectors_per_second_attr != "":
+        vecsec_segments = vectors_per_second_attr.split(",")
+        end_index = epoch_data.shape[0]
+        for vecsec_segment in reversed(vecsec_segments):
+            start_time, vecsec = vecsec_segment.split(":")
+            start_index = np.where(int(start_time) == epoch_data)[0][0]
+            gaps = np.concatenate(
+                (find_gaps(epoch_data[start_index : end_index + 1], int(vecsec)), gaps)
```
alastairtree, line 427: I think this logic would miss a gap at the start of the day, say the instument was in BM 00:00 -> 01:00 and NM 01:00 -> 02:00, it should fill the first hour with NM from BM. Similarlyly for a BM window at the end of the day.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,444 @@ def mag_l1c(
...
+            )
+            end_index = start_index
+    else:
+        # TODO: How to handle this case
+        gaps = find_gaps(epoch_data, 2)  # Assume half second gaps
```
alastairtree, line 432:
```suggestion
        gaps = find_gaps(epoch_data, MAG_RATE.TWO_VECS_PER_SEC)
```
If you have no normal mode data from which you can start the timeline/gap finding you can assume 2 vec/sec on both sensors starting at midnight and then try and fill that with any available BM

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,444 @@ def mag_l1c(
...
+    numpy.ndarray
+        Array of timestamps of shape (n, 2) containing n gaps with start_gap and
+        end_gap. Start_gap and end_gap both correspond to points in timeline_data.
+    """
+    # Expected difference between timestamps in nanoseconds.
+    expected_gap = 1 / vectors_per_second * 1e9
+
+    diffs = abs(timeline_data[:-1] - np.roll(timeline_data, -1)[:-1])
+    gap_index = np.where(diffs != expected_gap)[0]
+    output: np.ndarray = np.zeros((len(gap_index), 2))
```
alastairtree, line 463: Would this need a tolerance on it so that a gap if not flagged based on clock drift? Say the gaps went 500ms, 500ms 502ms, 498ms, 500ms - that is not a gap, it is just clocks drifting. Similarly 500.1ms, 500.1ms, 500.1ms, 500.1ms is not 4 gaps. In the past I have used 7.5% of the cadence, so in this case 35ms, as a rough tolerance on how long before a gap is a gap. Maybe it doesn't matter because you then look at BM? But I don't think you want to fill a 500.1 gap in the overlapping data sections and end up with 2 data points at 500 and 500.1ms

alastairtree (commented):
Thanks Maxine - this is looking great. I had another quick look through and added a few more little bits. I know you will come back to these but just flagging a few thing so we dont forget them

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: Yes, I wasn't sure how to handle those cases so I left them out in this iteration. After our discussion yesterday I think I know how to tackle it.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: This is a good point, I'll address this in the next PR and add some tests to double-check the behavior.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +51,444 @@ def mag_l1c(
...
+) -> np.ndarray:
+    """
+    Create MAG L1C data from L1B datasets.
+
+    This function starts from the normal mode dataset and completes the following steps:
+    1. find all the gaps in the dataset
+    2. generate a new timeline with the gaps filled
+    3. fill the timeline with normal mode data (so, all the non-gap timestamps)
+    4. interpolate the gaps using the burst mode data and the method specified in
+        interpolation_function.
```
vmartinez-cu, line 202: This is a helpful docstring

imap_processing/mag/l1c/interpolation_methods.py
```diff
@@ -0,0 +1,211 @@
+# mypy: ignore-errors
+"""Module containing interpolation methods for MAG L1C."""
+
+from enum import Enum
+
+import numpy as np
+from scipy.interpolate import make_interp_spline
+
+
+def linear(
```
lacoak21, line 10: Is there a way to combine cubic, linear, and quadratic and instead pass in k in the enum?

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +52,444 @@ def mag_l1c(
     if isinstance(second_input_dataset.attrs["Logical_source"], list):
```
lacoak21, line 52: I might be misunderstanding this part but I thought that l1c processing can still occur if there is only one of the two. In other words l1c processing preferably takes in both norm and burst but if one is missing it can still run.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +52,444 @@ def mag_l1c(
...
+    5 - generated flag (0 for normal data, 1 for interpolated data, -1 for missing data)
+    6-7 - compression flags (is_compressed, compression_width)
+
+    Parameters
+    ----------
+    normal_mode_dataset : xarray.Dataset
+        The normal mode dataset, which acts as a base for the output.
+    burst_mode_dataset : xarray.Dataset
+        The burst mode dataset, which is used to fill in the gaps in the normal mode.
+    interpolation_function : InterpolationFunction
```
lacoak21, line 216: What exactly is burst mode?

vmartinez-cu (approved):
This was a lot of work. Great job on this!

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +52,444 @@ def mag_l1c(
...
+    Returns
+    -------
+    numpy.ndarray
+        Array of timestamps of shape (n, 2) containing n gaps with start_gap and
+        end_gap. Start_gap and end_gap both correspond to points in timeline_data.
+    """
+    # Expected difference between timestamps in nanoseconds.
+    expected_gap = 1 / vectors_per_second * 1e9
+
+    diffs = abs(timeline_data[:-1] - np.roll(timeline_data, -1)[:-1])
```
lacoak21, line 461: Really cool use of np.roll

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -39,19 +52,444 @@ def mag_l1c(
...
+    for gap in gaps:
+        # TODO: we might need a few inputs before or after start/end
+        burst_start = (np.abs(burst_epochs - gap[0])).argmin()
+        burst_end = (np.abs(burst_epochs - gap[1])).argmin()
+        gap_timeline = filled_norm_timeline[
+            np.nonzero(
+                (filled_norm_timeline > gap[0]) & (filled_norm_timeline < gap[1])
+            )
+        ]
+        # do not include range
```
lacoak21, line 326: What does this comment mean?

lacoak21 (approved):
Great Job Maxine! - Just had a few questions.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: The vector in this codebase and in the CDF files generally means the x, y, z components of the magnetic field and a range for the instrument, which determines the scale of measurements. When we're interpolating around the magnetic field vectors, we don't want to also mess with the scale as that should remain the same no matter what.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: You should see L1A! Actually, maybe don't see that. It's really complicated.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: The instrument measures at a high cadence and then is downsampled into normal mode (which is generally something like 4 vectors a second) and burst mode (more frequent samples, something like 32 vectors a second). This saves space but can only run one at a time, so L1C takes the burst mode data and downsamples it to the normal mode cadence so we can get a continuous normal mode reading off the instrument.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: Yes, you're exactly right. That will be in the next PR since I didn't learn that until after this one was completed.

imap_processing/mag/l1c/interpolation_methods.py
maxinelasp: Yes, certainly, but in this case I prefer to be super explicit. This way I can line up cubic, linear, and quadratic to the configuration file directly (which specifies in cubic, linear or quadratic, not by k value) and ensure that I'm not running some invalid k-value by accident.

imap_processing/mag/l1c/mag_l1c.py
lacoak21: Oh I see - I just saw the comment below 👍

PR #1456, opened 2025-03-12
CoDICE L1a hi-sectored Validation

bourque:
This PR enables CoDICE l1a validation for the `hi-sectored` data product. During this process, I also tweaked how dimensions are defined for CoDICE-Hi products, which now use specific energy bins as the energy dimension instead of `esa_step`.

Pertains to #251


imap_processing/codice/constants.py
```diff
@@ -1117,5 +1124,11 @@
...
 }
 
-# TODO: Add energy tables for hi-sectored, hi-counters-aggregated, and
-#       hi-counters-singles
+SECTORED_ENERGY_TABLE = {
+    "h": [0.05, 0.1, 0.2, 0.4, 0.8, 1.6, 3.2, 6.4, 12.8],
+    "he3he4": [0.025, 0.05, 0.1, 0.2, 0.4, 0.8, 1.6, 3.2, 6.4],
+    "cno": [0.025, 0.05, 0.1, 0.2, 0.4, 0.8, 1.6, 3.2, 6.4],
+    "fe": [0.0125, 0.025, 0.05, 0.1, 0.2, 0.4, 0.8, 1.6, 3.2],
+}
```
vmartinez-cu, line 1135: What do these values represent? Are they min, max, or mean energies?

vmartinez-cu (approved):
LGTM! Also, I like how you set up your constants file. It's a nice way to organize variable/dimension information for your products

imap_processing/codice/constants.py
bourque: These values represent the edges of the energy bins, and are used to calculate the centers and deltas of the bins which then get stored in the L1a CDF data product.  I can add an inline comment here to make it more clear!

lacoak21 (approved):
Looks good to me too!

lacoak21:
> LGTM! Also, I like how you set up your constants file. It's a nice way to organize variable/dimension information for your products

Agreed! Very nice organization.

PR #1473, opened 2025-03-14
MNT: Updating cli.py code with latest dependency code.

tech3371:
# Change Summary

## Overview
Initial take on cli update using latest change. I appreciate feedback on this. I got IDEX to work. I will update others once the design and approach looks good to many. I will leave some comment of where I appreciate input.

## New Dependencies
New release of `imap-data-access`. Right now, I update my local library code to reflect changes I needed to test this.

## Updated Files
   - updated 'Parent' attrs which is input files used to produce current file.
- imap_processing/cli.py
   - Right now, I commented out other instrument besides IDEX until we are happy with design. Once we land on good approach, I will go back and fix other instruments.
- pyproject.toml
   - updated `imap-data-access` version. Will need to update this again.

Related to this [PR](https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/131) and [Release](https://github.com/IMAP-Science-Operations-Center/imap-data-access/releases/tag/v0.14.0) from `imap-data-access`


imap_processing/cli.py
```diff
@@ -274,45 +254,33 @@ def __init__(
...
+
+            # walk through all file list and download
+            for filepath in dependency.file_path_list:
+                try:
+                    print(f"Downloading {filepath}")
+                    imap_data_access.download(filepath)
+                except HTTPError as e:
+                    raise ValueError(f"Unable to download {filepath} file") from e
+
+        return input_collection
```
tech3371: I appreciate some feedback here. Is everyone ok with what we return here? The reason we chose to return was documented on this [PR](https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/131). This is what was written on that PR
> This is to simplify download dependency in general. Greg, Luisa and I discussed some approach and we thought this change may help on imap_cli update work. Right now in cli.py, we have to go through filename_list and then create construct_path() and then download each dependency files. Then we loose that information after download. With this change, we create download path list and store in the class and store in its file_path_list parameter. Then use that to download the file in cli.py and then return InputCollection() to all our instrument processing code. Instrument processing code can use that information to look up their downloaded file from file_path_list and also allow them to have other additional informations from InputCollection() in processing, such as input_type or descriptor to decide what to with certain science or ancillary or spice input type.

imap_processing/cli.py
```diff
@@ -574,56 +552,59 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
...
+        #     datasets = hit_l2(l1b_dataset, self.version)
 
         return datasets
 
 
 class Idex(ProcessInstrument):
     """Process IDEX."""
 
-    def do_processing(self, dependencies: list) -> list[xr.Dataset]:
+    def do_processing(
```
tech3371, line 673: This IDEX code shows example of how return value from `self.download_dependencies()` will be used in each instrument processing code. If this approach looks ok, then I will update other instruments to do similar.

greglucas:
One thing we have to keep in mind now is that we have dependent packages on these tools too (L3 code has been developed and is running). Is there a way to provide a transition here? If not, we need to document how someone else should transition to this new paradigm.

imap_processing/cdf/utils.py
```diff
@@ -78,8 +79,9 @@ def write_cdf(
     ----------
     dataset : xarray.Dataset
         The dataset object to convert to a CDF.
-    parent_files : list of Path, optional
-        List of parent files that were used to make this file. These get added to
+    parent_files : ProcessingInputCollection, optional
```
maxinelasp: Interesting... I think I would actually prefer to leave this as a list of filenames, but that's not a strong preference.

imap_processing/cli.py
```diff
@@ -263,7 +243,7 @@ def __init__(
         self.descriptor = data_descriptor
 
         # Convert string into a dictionary
-        self.dependencies = loads(dependency_str.replace("'", '"'))
+        self.dependencies = dependency_str
```
maxinelasp: Will need to update the comment here

imap_processing/cli.py
maxinelasp: Yes I like returning this!

imap_processing/cli.py
```diff
@@ -377,7 +347,7 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
 
         Parameters
         ----------
-        dependencies : list
+        dependencies : ProcessingInputCollection
             List of dependencies to process.
```
maxinelasp: Update comment?

imap_processing/cli.py
```diff
@@ -632,38 +613,48 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
         """
         print(f"Processing IDEX {self.data_level}")
         datasets: list[xr.Dataset] = []
-
         if self.data_level == "l1a":
```
maxinelasp, line 693: I wonder if it would make sense to add a method in ProcessingInputCollection to check for file counts. Something like 
`def validate_file_inputs(science_count, ancillary_count, spice_count) -> bool: `
to remove some of these commonly duplicated if-statements.

maxinelasp:
To follow up on @greglucas's comment - I will be writing a dependency overview document and probably some kind of transition guide for Menlo which we can send out prior to deploying breaking changes.

imap_processing/cli.py
tech3371: Some tools to remove these if else state would be helpful. In all instruments, we are checking if we have expected number of dependencies. Before, we just assume that we get one file per dependent data product. Now that the `dependencies` is an `ProcessingInputCollection` class type, we can use `dependencies.processing_input` length to determine if we have expected number of dependents. For example, this what `dependencies.serialize()` output would look like. This is saying that there is two dependent product, one science and one ancillary. 
```
[
    {
        "type": "ancillary",
        "files": [
            "imap_mag_l1b-cal_20250101_v001.cdf",
            "imap_mag_l1b-cal_20250103-20250104_v002.cdf"
        ]
    },
    {
        "type": "science",
        "files": [
            "imap_mag_l1a_norm-magi_20240312_v000.cdf",
            "imap_mag_l1a_norm-magi_20240312_v001.cdf"
        ]
    }
]
```

Now let's look at another case. Here it means, we have two ancillary dependents and one science dependent. I am writing this to give us some background information of why we had to use `dependencies.processing_input` parameter. Now, can we think of ways to remove this if statement? Can you think of some idea?
```
[
    {
        "type": "ancillary",
        "files": [
            "imap_mag_l1b-cal_20250101_v001.cdf",
            "imap_mag_l1b-cal_20250103-20250104_v002.cdf"
        ]
    },
    {
        "type": "ancillary",
        "files": [
            "imap_mag_l1b-lut_20250101_v001.cdf",
        ]
    },
    {
        "type": "science",
        "files": [
            "imap_mag_l1a_norm-magi_20240312_v000.cdf",
            "imap_mag_l1a_norm-magi_20240312_v001.cdf"
        ]
    }
]
```

tech3371: Another case is this. In this case, let's say it's idex l2 that we are processing and other three are dependent datas. 
```
[
    {
        "type": "ancillary",
        "files": [
            "imap_mag_l1b-cal_20250101_v001.cdf",
            "imap_mag_l1b-cal_20250103-20250104_v002.cdf"
        ]
    },
    {
        "type": "ancillary",
        "files": [
            "imap_mag_l1b-lut_20250101_v001.cdf",
        ]
    },
    {
        "type": "science",
        "files": [
            "imap_mag_l1a_norm-magi_20240312_v000.cdf",
            "imap_mag_l1a_norm-magi_20240312_v001.cdf"
        ]
    },
    {
        "type": "science",
        "files": [
            "imap_idex_l2_sci_20240312_v000.cdf",
            "imap_idex_l2_sci_20240312_v001.cdf"
        ]
    }
]
```
Because of these cases, it was helpful to return `ProcessingInputCollection` to user so that they can use `dependencies` to get their own data and format dependent data how they need it. They can get their data by looking at `input_type`, `source` and so on. We leave that up to user.

imap_processing/cli.py
```diff
@@ -274,45 +254,33 @@ def __init__(
...
+        file_list : ProcessingInputCollection
+            Object containing the dependencies for the instrument.
+        """
+        input_collection = ProcessingInputCollection()
+        input_collection.deserialize(self.dependencies)
+        # Go through science, ancillary or SPICE dependencies list and
+        # download all files
+        for dependency in input_collection.processing_input:
+            if dependency.input_type == ProcessingInputType.SPICE_FILE:
+                print("SPICE is not implemented yet")
```
tmplummer:
```suggestion
                logger.warning("SPICE is not implemented yet")
```

imap_processing/cli.py
```diff
@@ -274,45 +254,33 @@ def __init__(
...
+                continue
+
+            # walk through all file list and download
+            for filepath in dependency.file_path_list:
+                try:
+                    print(f"Downloading {filepath}")
```
tmplummer:
```suggestion
                    logger.info(f"Downloading {filepath}")
```

tmplummer (commented):
This looks like an overall improvement to me! No concerns with these changes.

imap_processing/cli.py
lacoak21: The IDEX example look good to me!

imap_processing/cdf/utils.py
```diff
@@ -39,6 +39,9 @@ def load_cdf(
     dataset : xarray.Dataset
         The ``xarray`` dataset for the CDF file.
     """
+    if isinstance(file_path, imap_data_access.ImapFilePath):
```
maxinelasp, line 42: You'll need to update the typehint in the function definition too

`file_path: Path | ImapFilePath`

imap_processing/cli.py
```diff
@@ -232,13 +236,36 @@ class ProcessInstrument(ABC):
...
+                ]
+            },
+            {
+                "type": "science",
+                "files": [
+                    "imap_idex_l2_sci_20240312_v000.cdf",
+                    "imap_idex_l2_sci_20240312_v001.cdf"
+                ]
+            }
+        ]'.
```
maxinelasp: Could you add a note to the comment that this is the output of `ProcessingInputCollection.serialize`?

imap_processing/cli.py
```diff
@@ -574,39 +622,44 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
         print(f"Processing HIT {self.data_level}")
         datasets: list[xr.Dataset] = []
 
+        dependency_list = dependencies.processing_input
```
maxinelasp, line 625: Instead of `.processing_input`, we could consider using `get_science_files()` to make it clear they are science dependencies

imap_processing/cli.py
```diff
@@ -574,39 +622,44 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
...
-            if len(dependencies) > 1:
+            if len(dependency_list) > 1:
                 raise ValueError(
                     f"Unexpected dependencies found for HIT L1B:"
-                    f"{dependencies}. Expected only one dependency."
+                    f"{dependency_list}. Expected only one dependency."
                 )
             data_dict = {}
-            if self.dependencies[0]["data_level"] == "l0":
+            if dependency_list[0]["descriptor"] == "hk":
```
maxinelasp:
```suggestion
            if dependency_list[0].descriptor == "hk":
```
dependency_list is a list of ProcessingInput objects, no longer a dictionary.

maxinelasp: Also, should this be .contains rather than ==?

imap_processing/cli.py
```diff
@@ -574,39 +622,44 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
...
-            if len(dependencies) > 1:
+            if len(dependency_list) > 1:
                 raise ValueError(
                     f"Unexpected dependencies found for HIT L2:"
-                    f"{dependencies}. Expected only one dependency."
+                    f"{dependency_list}. Expected only one dependency."
                 )
             # Add L1B dataset to process science data
-            l1b_dataset = load_cdf(dependencies[0])
+            l1b_file_obj = dependency_list[0].imap_file_paths[0]
```
maxinelasp: Maybe we should add a method to ProcessingInput that's like `retrieve_one_file` which can double-check that `imap_file_paths` is a length of one and returns it. That can be additional verification that we only have one file here. Future work, though.

maxinelasp (approved):
Looks good! A few minor notes, but overall looks like a good first pass. There's definitely still room for individual instruments to do additional validation work, but I think it's clear how that would work at this point and it's definitely not your responsibility. Thanks for doing some of the tedious work here!

tech3371:
Hi All,
This PR is ready for review. I will fix the pytest test tomorrow. I appreciate if you can look at your instrument code or SWE for example. I tried to test processing from l0 to l1a for all instrument through `imap_cli` locally. Some failed while some passed but the cli updates worked. I added `TODO` and notes where I was not sure and appreciate if you can double check in future PR. Testing higher level is little bit more than I can handle. I appreciate if you can test it during your development and verification. Thank you so much!!!

tech3371:
List of commands I ran to test locally. I hope this helps for future tests.
```
imap_cli --instrument codice --data-level l1a --start-date 20250223 --version 001 --descriptor raw --dependency '[{"type": "science", "files": ["imap_codice_l0_raw_20250223_v001.cdf"]}]'                                      
imap_cli --instrument glows --data-level l1a --start-date 20250206 --version 000 --descriptor raw --dependency '[{"type": "science", "files": ["imap_codice_l0_raw_20250206_v000.pkts"]}]' 
imap_cli --instrument hi --data-level l1a --start-date 20241212 --version 000 --descriptor raw --dependency '[{"type": "science", "files": ["imap_hi_l0_raw_20241212_v000.pkts"]}]' 
imap_cli --instrument hit --data-level l1a --start-date 20250102 --version 001 --descriptor sci --dependency '[{"type": "science", "files": ["imap_hit_l0_raw_20250102_v001.pkts"]}]'                                                
imap_cli --instrument lo --data-level l1a --start-date 20241212 --version 000 --descriptor sci --dependency '[{"type": "science", "files": ["imap_lo_l0_raw_20241212_v000.pkts"]}]'                                                   
imap_cli --instrument mag --data-level l1a --start-date 20250102 --version 000 --descriptor sci --dependency '[{"type": "science", "files": ["imap_mag_l0_raw_20250102_v000.pkts"]}]'           
imap_cli --instrument mag --data-level l1c --start-date 20251017 --version 115 --descriptor norm-magi --dependency '[{"type": "ancillary", "files": ["imap_mag_calibration_20240229_v002.cdf"]}, {"type": "science", "files": ["imap_mag_l1b_norm-magi_20251017_v115.cdf"]}]'                                        
imap_cli --instrument ultra --data-level l1a --start-date 20241212 --version 000 --descriptor sci --dependency '[{"type": "science", "files": ["imap_ultra_l0_raw_20241212_v000.pkts"]}]'                                          
imap_cli --instrument swe --data-level l1a --start-date 20241212 --version 000 --descriptor sci --dependency '[{"type": "science", "files": ["imap_swe_l0_raw_20241212_v000.pkts"]}]'              
imap_cli --instrument swe --data-level l1b --start-date 20250129 --version 001 --descriptor sci --dependency '[{"type": "science", "files": ["imap_swe_l1a_sci_20250129_v001.cdf"]},{"type": "ancillary","files": ["imap_swe_l1b-in-flight-cal_20240510_20260716_v001.csv"]}]'                 
```

imap_processing/cli.py
```diff
@@ -364,21 +380,27 @@ def pre_processing(self) -> list[Path]:
         list[Path]
             List of dependencies downloaded from the IMAP SDC.
         """
-        self._dependency_list = self.download_dependencies()
-        return self._dependency_list
+        input_collection = ProcessingInputCollection()
+        input_collection.deserialize(self.dependencies)
+        input_collection.download_all_files()
+        self._dependency_collection = input_collection
```
maxinelasp: Nice and clean 😁

imap_processing/cli.py
```diff
@@ -526,27 +567,34 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
...
                 )
-            datasets = hi_l1a.hi_l1a(dependencies[0], self.version)
+            science_files = dependencies.get_file_paths(source="hi")
+            datasets = hi_l1a.hi_l1a(science_files[0], self.version)
         elif self.data_level == "l1b":
-            dependencies = [load_cdf(dependency) for dependency in dependencies]
-            datasets = [hi_l1b.hi_l1b(dependencies[0], self.version)]
+            # TODO: check this and update with new features
+            science_files = dependencies.get_file_paths(source="hi")
+            hi_dependencies = [load_cdf(file) for file in science_files]
```
maxinelasp, line 583: Not relevant for now, but maybe we could have load_cdf also accept a ProcessingInputCollection at some point...

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -140,8 +146,8 @@ def read_in_flight_cal_data() -> pd.DataFrame:
     columns = ["met_time", "cem1", "cem2", "cem3", "cem4", "cem5", "cem6", "cem7"]
 
     # Create an empty DataFrame with the specified columns
-    empty_df = pd.DataFrame(columns=columns)
-    return empty_df
+    in_flight_cal_df = pd.read_csv(in_flight_cal_path, names=columns)
```
maxinelasp: In general, I think it's preferable to read all the input files in imap_processing.cli and pass them into processing as dataframes or datasets  (so basically call read_in_flight_cal_data in cli.py and then pass in the dataframe instead of the path to swe_l1b) but that's not a very strong feeling. It just would be nice to have all the inputs and outputs managed in one file as much as possible so any failures to find files come from there instead of deep in instrument processing.

maxinelasp (approved):
Overall looks good, I did a reasonably thorough review of all instruments. I think individual instruments will definitely want to go back and add more specific validation, but this is really helpful to set that up!!

imap_processing/cli.py
```diff
@@ -406,10 +428,16 @@ def post_processing(self, datasets: list[xr.Dataset]) -> None:
             return
 
         logger.info("Writing products to local storage")
-        logger.info("Parent files: %s", self._dependency_list)
+
+        list_of_files = [
+            dep_ojb.filename_list
+            for dep_ojb in self._dependency_collection.processing_input
```
lacoak21: dep_obj.filename_list
for dep_obj in self._dependency_collection.processing_input
```suggestion
            dep_obj.filename_list
            for dep_obj in self._dependency_collection.processing_input
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -140,8 +146,8 @@ def read_in_flight_cal_data() -> pd.DataFrame:
     columns = ["met_time", "cem1", "cem2", "cem3", "cem4", "cem5", "cem6", "cem7"]
 
     # Create an empty DataFrame with the specified columns
```
lacoak21: Remove TODOs and update this comment?

lacoak21 (commented):
Looks great Tenzin! Just had a few minor comments. Its cool to see swe l1b processing using the ancillary file now 👍 .

lacoak21 (approved):

imap_processing/cli.py
tech3371: Yeah, something to think about

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: Totally. I have a TODO to rethink that and do that in future PR.

imap_processing/cli.py
tech3371: good catch! Sometimes my brain thinks faster than my finger can type and this happens...lol

imap_processing/tests/hi/test_hi_l1c.py
```diff
@@ -40,6 +40,7 @@ def test_hi_l1c_not_implemented():
         hi_l1c.hi_l1c([None, None], "0")
 
 
+@pytest.mark.xfail(reason="Revisit this. The test is failing due to a bug in the code.")
```
tech3371: @subagonsouth Can you look at these two tests in future PR? It was giving error like these. I looked at the test code and I can't figure what changes in this PR would cause this. Sorry for this.
```
E               KeyError: "No variable named 'tof_ab'. Variables on the dataset include ['ccsds_met', 'esa_step', 'trigger_id', 'ccsds_index', 'esa_energy_step', ..., 'hae_latitude', 'hae_longitude', 'quality_flag', 'epoch', 'event_met']"
```

imap_processing/cdf/utils.py
```diff
@@ -26,8 +26,8 @@ def load_cdf(
 
     Parameters
     ----------
-    file_path : Path
-        The path to the CDF file.
+    file_path : Path | ImapFilePath
```
greglucas: This should be in the `load_cdf()` declaration, and numpy doc style uses the word "or" I believe.
```suggestion
    file_path : Path or ImapFilePath
```

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -22,6 +25,8 @@ def swe_l1b(l1a_dataset: xr.Dataset, data_version: str) -> xr.Dataset:
         The l1a data input.
     data_version : str
         Version of the data product being created.
+    in_fligh_cal_path : pathlib.Path
```
greglucas: Minor typo here.
```suggestion
    in_flight_cal_path : pathlib.Path
```

imap_processing/swe/l1b/swe_l1b_science.py
```diff
@@ -125,23 +126,21 @@ def read_in_flight_cal_data() -> pd.DataFrame:
...
     """
-    # TODO: Read in in-flight calibration file.
-
     # Define the column headers
     columns = ["met_time", "cem1", "cem2", "cem3", "cem4", "cem5", "cem6", "cem7"]
 
-    # Create an empty DataFrame with the specified columns
-    empty_df = pd.DataFrame(columns=columns)
-    return empty_df
+    in_flight_cal_df = pd.read_csv(in_flight_cal_path, names=columns)
```
greglucas: I think it would be helpful to keep this in a separate PR in the future. Only update the CLI here, then update SWE specific things in a separate one.

imap_processing/tests/test_cli.py
```diff
@@ -37,11 +45,21 @@ def test_main(mock_instrument):
         "mag",
         "--dependency",
         (
-            '[{"instrument": "mag", '
-            '"data_level": "l0", '
-            '"descriptor": "sci", '
-            '"version": "v001", '
-            '"start_date": "20240430"}]'
+            "["
+            "{"
```
greglucas, line 49: combine some of these characters together instead of on separate lines? Or does ruff do this autoformatting for you?

imap_processing/cli.py
```diff
@@ -263,8 +319,8 @@ def __init__(
         self.descriptor = data_descriptor
 
         # Convert string into a dictionary
-        self.dependencies = loads(dependency_str.replace("'", '"'))
-        self._dependency_list: list = []
+        self.dependencies = dependency_str
+        self._dependency_collection: ProcessingInputCollection = None
```
greglucas: Can we turn dependencies into the InputCollection? Why do we need to keep both of them around, we were previously loading it so it seems like we should load it into `self.dependencies` again.

imap_processing/cli.py
greglucas: See my other comment, but I think this could be:
Move these two lines up into the other block.
```python
        self.dependencies = ProcessingInputCollection().deserialize(input_str)
```

Then preprocessing is just downloading all files currently.
```suggestion
        self.dependencies.download_all_files()
```

imap_processing/cli.py
```diff
@@ -526,27 +567,34 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
...
+            hi_dependencies.append(
                 Path(__file__).parent
                 / "tests/hi/test_data/l1"
                 / "imap_his_pset-calibration-prod-config_20240101_v001.csv"
             )
-            dependencies[0] = load_cdf(dependencies[0])
-            datasets = [hi_l1c.hi_l1c(dependencies, self.version)]
+            # TODO: check this and update with new features
+            science_files = dependencies.get_file_paths(source="hi")
+            hi_dependencies[0] = load_cdf(science_files[0])
```
greglucas: I don't think this is the same as it was before? You are overwriting the hi_dependencies path of the input configuration file.

Previously it was overwriting itself, but you've made new different variables now.

I'd suggest putting the `hi_dependencies = dependencies.get_file_paths(source="hi")` at the entry instead of an empty list.

imap_processing/cli.py
```diff
@@ -574,39 +624,44 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
...
+            if len(dependency_list) > 1:
                 raise ValueError(
                     f"Unexpected dependencies found for HIT L1B:"
-                    f"{dependencies}. Expected only one dependency."
+                    f"{dependency_list}. Expected only one dependency."
                 )
             data_dict = {}
-            if self.dependencies[0]["data_level"] == "l0":
+            l0_files = dependencies.get_file_paths(source="hit", descriptor="raw")
+            l1a_files = dependencies.get_file_paths(source="hit")
```
greglucas, line 645: The l1a_files will contain **both** the raw descriptor and the l1a specific descriptors. Is that what you want, or is that supposed to be subset as well?

imap_processing/cli.py
```diff
@@ -672,27 +735,36 @@ def do_processing(self, dependencies: list) -> list[xr.Dataset]:
...
                 )
-            datasets = lo_l1a.lo_l1a(dependencies[0], self.version)
+            science_files = dependencies.get_file_paths(source="lo")
+            datasets = lo_l1a.lo_l1a(science_files[0], self.version)
 
         elif self.data_level == "l1b":
             data_dict = {}
-            for dependency in dependencies:
-                dataset = load_cdf(dependency)
+            # TODO: Check this and update with new features as needed.
```
greglucas, line 751: I think this would be `for dep in dependencies.get_file_paths(): dataset = load_cdf(dep)` to be the same as before.

ping @sdhoyt

imap_processing/tests/hi/test_hi_l1c.py
greglucas: @tech3371 can you remove these xfails? I don't get a failure locally when I pull your code down. I get xpass, so I think this works as expected.

greglucas (dismissed):
I think the hi cli isn't doing what it was previously.

imap_processing/cli.py
tech3371: I remember you saying that. I tried before too. For some reason, I can't get code working when I do all that class in one line. I really don't know why.

imap_processing/cli.py
tech3371: I think here and other instrument is gonna need more update after this PR. I do appreciate if they can make those changes you are requesting below in a future PR since they know their code better than I do. I am trying to get the basic thing setup with new changes.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: I do have more changes that I plan to do in future PR for SWE specific. I just brought this one to show how more than one dependency can look.

imap_processing/tests/hi/test_hi_l1c.py
tech3371: I will look at this next week again. somehow it fails on my end.

imap_processing/cli.py
greglucas: OK, two lines would work too. But, move it up to the other section remove the `_dependency_collection` was my main point.

imap_processing/cli.py
greglucas: I'm pretty sure this is wrong though, so I think it should be fixed. I'll try to inline suggest the equivalent to the previous in a separate comment.

greglucas: I couldn't do it inline so I pushed a new commit with my suggestion. Let me know if you agree with that or feel free to push it off your commit stack.

imap_processing/cli.py
sdhoyt: I'm not sure I understand the reason for getting the input_type to specify the descriptor. The dependencies should only contain what's listed as a dependency for the job, correct?

imap_processing/cli.py
tech3371: Thank you for making that change

imap_processing/cli.py
tech3371: did my last commit before break fixed this or did you want more change there?

imap_processing/cli.py
tech3371: To answer Greg:
I had to change Lo to get files with instrument and descriptor specified because Lo has more than one upstream dependencies in L1B. This is what we have on `sds-data-manager` for Lo.
```
lo, l1a, de, lo, l1b, de, HARD, DOWNSTREAM
lo, l1a, spin, lo, l1b, de, HARD, DOWNSTREAM
lo, l1a, TODO, lo, l1b, badtimes, HARD, DOWNSTREAM
lo, l1a, histogram, lo, l1b, monitorrates, HARD, DOWNSTREAM
lo, l1a, spin, lo, l1b, monitorrates, HARD, DOWNSTREAM
lo, l1a, histogram, lo, l1b, histogramrates, HARD, DOWNSTREAM
lo, l1a, spin, lo, l1b, histogramrates, HARD, DOWNSTREAM
lo, l1a, histogram, lo, l1b, derates, HARD, DOWNSTREAM
lo, l1a, spin, lo, l1b, derates, HARD, DOWNSTREAM
lo, l1a, star, lo, l1b, prostar, HARD, DOWNSTREAM
lo, l1a, hk, lo, l1b, hk, HARD, DOWNSTREAM
```
So if we do `for dep in dependencies.get_file_paths(): dataset = load_cdf(dep)`, it will mix up datasets of both `l1a` `spin` data and say `de` data which is not what Sean wants there, I think. I think he is loading `spin` data and assigning that to the `data_dict`, then load `de` data and add to `data_dict` and so on. I could have misunderstand. I am replying to Sean as well and based on what you both think, I can change accordingly or do that in future PR.

To answer Sean:
Sean, you dependencies inputs can look like with new changes
```
[
            {
                "type": "science",
                "files": [
                    "imap_lo_l1a_spin_20250101_v001.cdf",
                ]
            },
            {
                "type": "science",
                "files": [
                    "imap_lo_l1a_de_20240312_v001.cdf"
                ]
            }
        ]
```
Then this gets loaded into `ProcessingInputCollection()` class (done by `pre_processing()`) which then can be used to get back specific data you want. In this example, you have two dependencies. So if you want to get back `spin` data, then you can call `get_file_paths(lo, spin)` to get back download paths of `spin` files listed in above dependencies input. Then use that to load the data. If we do, `get_file_paths()` without specifying anything, it will return download paths of both `spin` and `de`. Is that what you are expecting here? If so, I can make change Greg suggested.

imap_processing/cli.py
```diff
@@ -344,41 +360,38 @@ def process(self) -> None:
         logger.info(f"IMAP Processing Version: {imap_processing._version.__version__}")
         logger.info(f"Processing {self.__class__.__name__} level {self.data_level}")
         logger.info("Beginning preprocessing (download dependencies)")
-        dependencies = self.pre_processing()
+        self.pre_processing()
         logger.info("Beginning actual processing")
-        products = self.do_processing(dependencies)
+        products = self.do_processing(self.dependencies)
```
greglucas: Do we want to pass in `self.dependencies` or should these be referenceable from the task/class itself?

I think this is a case of we are storing some things on the class, but others we are passing around in more of a functional style. It is just a bit confusing of when to reference the value that was passed in versus the value that is stored on the class.

```python
self.pre_processing()  # Internally modifies self.dependencies potentially (resolving the download path)
products = self.do_processing(self.dependencies)
```

What if we got rid of `self.dependencies` altogether and just passed around `dependencies` here.

```python
dependencies = self.pre_processing()
products = self.do_processing(dependencies)
final_products = self.post_processing(products)
```

imap_processing/cli.py
greglucas: Yes! Thank you

imap_processing/cli.py
tech3371: What ever approach we take here, we need to have `self.dependencies` or `self.dependencies_str` to store input from `--dependencies`. I will push some changes with this change.

greglucas: OK, I see after your latest commit because of the entry-point and initialization from the cli versus when we call these methods.

Is the latest commit the direction you prefer? Or my other suggestion was to go the opposite direction and reference `self` things everywhere.
```
self.pre_processing()  # Internally modifies self.dependencies potentially (resolving the download path)
self.products = self.do_processing()
self.post_processing()
```

@subagonsouth I think you had the initial idea for these various stages of processing. Do you have any preference for what is class vs function argument storage/data flow?

To be clear, I'm fine with any of these, I would just prefer it to be consistent if we can and noted the slight inconsistency when looking at it.

imap_processing/swe/l1b/swe_l1b_science.py
tech3371: I am undoing SWE changes since recent I-ALiRT code merge needs more changes. I will do these changes in upcoming PR.

imap_processing/cli.py
tech3371: Some clear direction would be helpful here. I think this PR will block many people and I like to merge and get this basic changes working with infrastructure as soon as possible. Then we can work through these updates later if possible. Appreciate some guidance here.

imap_processing/cli.py
sdhoyt: Thanks for the clarification! I think what you have will work. I will make any needed adjustments after this is merged

sdhoyt (approved):
Thanks for all your work on this!

imap_processing/cli.py
tmplummer: Sorry for the slow response. There has been lots to catch back up with. I am happy with the current implementation in this PR where the `dependency` arg is getting passed around.

tmplummer (approved):
These changes look good to me. Thanks for all of this work!

PR #1477, opened 2025-03-17
Ultra l1c pset - Structure for Ultra l1c Sensitivity

laspsandoval:
# Change Summary

## Overview
This is a small PR to simply create the structure of the sensitivity in HEALPIX with TODOs for using the Aux S3 API and the Universal Spin Table and Pointing Table.

## Updated Files
- ultra_l1c_pset_bins.py
   - Define the structure of the sensitivity

## Testing
test_ultra_l1c_pset_bins.py
conftest.py

### DIFF
--- a/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
+++ b/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
@@ -1,10 +1,8 @@
 """Module to create pointing sets."""
 
-from pathlib import Path
-
 import astropy_healpix.healpy as hp
-import cdflib
 import numpy as np
+import pandas
 from numpy.typing import NDArray
 
 from imap_processing.ena_maps.utils.spatial_utils import build_spatial_bins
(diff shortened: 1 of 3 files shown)


imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -228,27 +224,37 @@ def get_helio_exposure_times(
     return exposure_3d
 
 
-def get_pointing_frame_sensitivity(
-    constant_sensitivity: Path, n_spins: int, sensor: str
-) -> NDArray:
+def get_spacecraft_sensitivity(
```
nkerman, line 227: Now that they're not in the repo, how will the code actually get these static `dataFrames`?

nkerman (commented):
Thanks @subagonsouth for catching that we shouldn't put the large files into version control. Now I'm just wondering how exactly we will access this data, which I'm pretty sure will be static/unchanging for all data sets.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
laspsandoval: There will be a aux api that I think will be similar. It's being developed by Maxine and Luisa currently.

laspsandoval: Also there is a very good chance these will all change. So having a s3 bucket w/ version control etc. will keep them organized.

nkerman (approved):
The changes look good to me still! Thanks for removing the big files from the history.

tmplummer (approved):
LGTM. Thanks for making the new PR!

PR #1484, opened 2025-03-17
ENA Maps: Change UltraPointingSet to Healpix

### LINKED ISSUE #1480: ENA Maps: Change UltraPointingSet to Healpix
Previously had defined UltraPointingSet as a rectangularly binned grid. Switch to a Healpix Tessellation with nominal nside of 128

### LINKED ISSUE #1483: Mock Ultra L1C data on healpix grid
Make horizontal stripe, rather than vertical stripe in the counts. Only needs to be vaguely realistic for testing/visual verification with Instrument team.

nkerman:
# Change Summary

Previously, we had defined UltraPointingSet as a rectangularly-binned grid. Switch to a Healpix Tessellation with nominal `nside` of 128.

Closes #1480
Closes #1483 

## Overview

1. Rename the existing implementation of `UltraPointingSet` to `RectangularPointingSet`. Keep it for reference or future use by Lo/Hi, and to continue testing the projection and index matching from rectangular PSETs to both types of `SkyMap`. 
2. Implement a new `UltraPointingSet` with Healpix tiling.
3. Extract the hardcoded coordinate names for L1C, L2 into an enum to prevent issues with typos, etc. This was spurred by the change from "elevation/azimuth_bin_center" -> "latitude /elevation_bin_center" at the project level, and remaining uncertainty about coordinate names. 
4. Mock the healpix tessellated L1C data, with a horizontal stripe of counts and exptime at constant latitude. The Ultra team asked me to use a horizontal stripe rather than a vertical stripe, so I also changed the Rectangular PSET mocking. 

## New Files

-  imap_processing/tests/ena_maps/conftest.py
   - fixtures for data mocking. 
-  imap_processing/ena_maps/utils/coordinates.py
   - Enum of coordinate names for making ENA maps. 

## Updated Files

   - Define Healpix tessellated `UltraPointingSet`
   - Rename rectangularly gridded PSET
   - Now mocks both healpix and rectangular Pointing Set datasets with horizontal stripe in counts

## Testing

Adapt existing tests to use `RectangularPointingSet`, and add new tests with Healpix gridded `UltraPointingSet`.

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -372,6 +372,7 @@ class AbstractSkyMap(ABC):
         self.tiling_type: SkyTilingType
         self.sky_grid: spatial_utils.AzElSkyGrid
         self.num_points: int
+        self.spatial_coords: dict[str, xr.DataArray | NDArray]
         self.binning_grid_shape: tuple[int, ...]
         self.data_dict: dict[str, NDArray]
 
@@ -468,6 +469,108 @@ class AbstractSkyMap(ABC):
 
             self.data_dict[value_key] += pointing_projected_values
 
+    def data_dict_value_to_dataarray(
+        self,
+        data_dict_key: str,
+        dims: list[str] | None = None,
+        **kwargs: dict,
+    ) -> xr.DataArray:
+        """
+        Convert an individual array in the data_dict to an xarray DataArray.
+
+        In the case of a rectangular grid, the data array is rewrapped to the original
+        2D grid shape before conversion to an xarray DataArray.
+
+        Parameters
+        ----------
+        data_dict_key : str
+            The key of the data array in the data_dict.
+        dims : list[str] | None, optional
+            The ordered dimensions of the data array. Default is None.
+        **kwargs : dict
+            Additional keyword arguments to pass to the xarray DataArray constructor.
+
+        Returns
+        -------
+        xr.DataArray
+            The data array converted to an xarray DataArray.
+        """
+        data = self.data_dict[data_dict_key]
+
+        # Rewrap the data to the original 2D grid shape if rectangular
+        if self.tiling_type is SkyTilingType.RECTANGULAR:
+            data = spatial_utils.rewrap_even_spaced_az_el_grid(
+                data, self.binning_grid_shape
+            )
+
+        # Add an extra dim at the start for epoch:
+        if dims is not None and dims[0] == "epoch":
+            data = np.expand_dims(data, axis=0)
+
+        return xr.DataArray(
+            data,
+            dims=dims,
+            **kwargs,
+        )
+
+    def to_xarray(
+        self,
+        non_spatial_coords: dict[str, xr.DataArray | NDArray],
+        data_variables_and_dims: dict[str, list[str]],
+    ) -> xr.Dataset:
+        """
+        Convert the data_dict to an xarray Dataset, including spatial coords.
+
+        In the case of a rectangular grid, all data arrays are rewrapped to the
+        original 2D grid shape before conversion to xarray DataArrays.
+
+        Parameters
+        ----------
+        non_spatial_coords : dict[str, xr.DataArray | NDArray],
+            Dictionary of the non-spatial coordinates to include in the dataset.
+            The keys are the names of the coordinates, and the values are either
+            numpy arrays or xarray DataArrays, the latter of which can include
+            coordinate attributes.
+
+            Example using DataArrays to give coord attributes:
+            ```
+            {
+                "epoch": xr.DataArray(
+                    [1234.5], dims=["epoch"], attrs={"units": "J2000ns"}),
+                "energy_bin_center": xr.DataArray(
+                    [1.0, 2.0, 3.0], dims=["energy_bin_center"],
+                    attrs={"units": "keV"}),
+            }
+            ```
+            Example using NDArrays:
+            ```
+            {
+                "epoch": np.array([1234.5]),
+                "energy_bin_center": np.array([1.0, 2.0, 3.0]),
+            }
+            ```.
+        data_variables_and_dims : dict[str, list[str]]
+            The dictionary of data variables (key) and their dimensions (value).
+
+        Returns
+        -------
+        xr.Dataset
+            An xarray Dataset containing the data variables and coordinates.
+        """
+        # Get full dict of coordinates
+        coords = {**non_spatial_coords, **self.spatial_coords}
+
+        return xr.Dataset(
+            data_vars={
+                key: self.data_dict_value_to_dataarray(
+                    key,
+                    dims=dims,
+                )
+                for key, dims in data_variables_and_dims.items()
+            },
+            coords=coords,
+        )
+
 
 class RectangularSkyMap(AbstractSkyMap):
     """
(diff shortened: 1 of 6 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -340,6 +339,97 @@ def __init__(
...
+        If the longitude/az or latitude/el bin centers don't match the constructed grid.
+        Or if the longitude or latitude bin spacing is not uniform.
+    ValueError
+        If multiple epochs are found in the dataset.
+    """
+
+    def __init__(
+        self,
+        l1c_dataset: xr.Dataset | pathlib.Path | str,
+        spice_reference_frame: geometry.SpiceFrame = geometry.SpiceFrame.IMAP_DPS,
```
greglucas, line 372: Is this already in the l1c_dataset, or is that a future ticket? Up on the pset parent class I think we could define a property.

```python
@property
def spice_reference_frame(self):
    return geometry.SpiceFrame(self.data.attrs["SPICE_FRAME"])
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -340,6 +339,97 @@ def __init__(
...
+        self.tiling_type = SkyTilingType.HEALPIX
+        self.num_points = self.data[CoordNames.HEALPIX_INDEX.value].size
+        self.nside = hp.npix_to_nside(self.num_points)
+
+        # Determine if the HEALPix tessellation is nested, default is False
+        self.nested = bool(
+            self.data[CoordNames.HEALPIX_INDEX.value].attrs.get("nested", False)
+        )
+
+        # Get the azimuth and elevation coordinates of the healpix pixel centers (deg)
```
greglucas, line 398: In the datasets we use longitude and latitude centers it looks like. Is that what we settled on at the meeting last week? I remember you asked the question about it, but I forgot what the decision was.

Then you use az/el in the classes, but then this `pix2ang` has a `lonlat=True`. So there is a lot of longitude/latitude and azimuth/elevation mixing going on here, including with the CoordNames which you mention below. Would it be feasible to change everything over to longitude/latitude to align with the dataset values so we are consistent everywhere?

imap_processing/ena_maps/ena_maps.py
nkerman: I don't think it's in the L1C Pointing Set definition yet\*, so I hadn't thought to add it. As long as we have a single global attribute, this is great and I can add it now. 

\*Tim was talking about this - I _think_ it was at yesterday's tag up - and he wasn't entirely sure if it can be a global attribute or if it needs to be on individual variables. For a Hi L1C file, some angles might be in the HAE frame already while others are DPS. @subagonsouth am I getting this right?

imap_processing/ena_maps/ena_maps.py
nkerman: Yeah, good point, Greg. 

I've been quite conflicted about this too, and spent a while converting almost everything to lon/lat before deciding to reverse it. Part of it was definitely that it caused a bunch of tiny merge conflicts 😁, another part was I didn't think that the L1C definition was going to change from azimuth -> longitude, and the geometry function is defined as `frame_transform_az_el`. I decided at the time to keep it in az/el within our code and just define the coordinate names in an enum, allowing them to differ for L1C/L2. 

All that said none of those are particularly strong reasons, and it's doable if you think we should switch to lon/lat.

greglucas: I am fine leaving that as a follow up PR, but I do really appreciate consistency as you know ;) so yes I do think we should do it if it makes sense.

nkerman: Made a ticket for this in a future PR Issue #1491

imap_processing/ena_maps/ena_maps.py
tmplummer: Correct. For Hi PSET, I think of the main product as being in the `IMAP_DPS` frame but, we provide `hae_latitude` and `hae_longitude` coordinates for each pixel. I suppose that the different frame is obvious by the variable name though.

nkerman: Thanks Tim - I'd rather have the map-maker (e.g., Ultra L2) explicitly pass in the `spice_reference_frame` when they make the PSET then. Each instrument's pointing set object can pick whichever default makes sense to them. For Ultra thats the DPS frame.

imap_processing/ena_maps/utils/coordinates.py
```diff
@@ -0,0 +1,17 @@
...
+    """Enumeration of the names of the coordinates in the L1C and L2 ENA datasets."""
+
+    TIME = "epoch"
+    ENERGY = "energy_bin_center"
+    HEALPIX_INDEX = "healpix_pixel_index"
+
+    # The names of the az/el angular coordinates may differ between L1C and L2 data
+    AZIMUTH_L1C = "longitude_bin_center"
+    ELEVATION_L1C = "latitude_bin_center"
+    AZIMUTH_L2 = "longitude_bin_center"
```
tmplummer, line 16: I think that the "_bin_center" will be dropped in the L2/3 products. In general I would recommend that change in all products. Bin width will likely need to be specified using the `DELTA_PLUS` and `DELTA_MINUS` attributes to point to a "longitude_delta" or similarly named variable. Alternatively, for HEALPix, one of the description attributes can specify that the value is the center of the pixel.

tmplummer (commented):
This all seems reasonable to me. To be honest, I feel a bit lost in the woods with some of the details. I've lost that big picture and need to start using some of this code to have more meaningful reviews.

imap_processing/ena_maps/utils/coordinates.py
nkerman: >I think that the "_bin_center" will be dropped in the L2/3 products. 

This makes me think we should keep this split of L1C and L2 az/el values in the Enum - which Greg and I were discussing on the other PR. I'll switch it to be just "longitude"/"latitude" at L2. 

>In general I would recommend that change in all products.

@laspsandoval is there any reason not to drop "bin_center" at L1? If so, the enum should be easy enough to change.

nkerman:
> This all seems reasonable to me. To be honest, I feel a bit lost in the woods with some of the details. I've lost that big picture and need to start using some of this code to have more meaningful reviews.

That's fair, @subagonsouth, me too! I'm really trying to wrap up changes to the underlying ENA map code and just write the Ultra L2 around it. These darn ENA Map PRs are hydras! Every time I put one up, it spurs 2 more 😁 🐍

imap_processing/ena_maps/utils/coordinates.py
laspsandoval: Let's drop "bin_center". Sounds good.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -340,6 +339,97 @@ def __init__(
...
+        self.spice_reference_frame = spice_reference_frame
+
+        # Read in the data and store the xarray dataset as data attr
+        if isinstance(l1c_dataset, (str, pathlib.Path)):
+            self.data = load_cdf(pathlib.Path(l1c_dataset))
+        elif isinstance(l1c_dataset, xr.Dataset):
+            self.data = l1c_dataset
+
+        # A PSET must have a single epoch
+        self.epoch = self.data["epoch"].values
```
laspsandoval, line 384: You probably don't have to store this and instead do:

len(np.unique(self.data["epoch"].values)) > 1

laspsandoval (approved):
I had a single comment. Besides that I think it looks good.

imap_processing/ena_maps/utils/coordinates.py
nkerman: Done!

imap_processing/ena_maps/ena_maps.py
nkerman: The epoch actually needs to be stored so that it can be used in the projection later.

nkerman:
I mistakenly merged this before dropping the "_bin_center" from the coord names. Expect a _tiny_ PR to follow this that captures the addressing of @laspsandoval's comments

PR #1485, opened 2025-03-17
Conversions for ena map dictionaries to `xarray DataArray/Datasets`

### LINKED ISSUE #1481: Produce xarray Datasets from ENA Maps
The ENA SkyMaps - both Healpix and Rectangularly tiled - keep their pixels' values (e.g., "counts") in dictionaries, with multiple spatial dims unwrapped in the case of Rectangular maps. 

```python
sky_map.data_dict = {
    "counts": NDArray<1 epoch, M energy bins, N unwrapped spatial pixels >
}
```

Define a function which produces an xarray dataset from this dictionary, and handles all re-wrapping of spatial dimensions as necessary.

nkerman:
# Change Summary

The ENA SkyMaps - both Healpix and Rectangularly tiled - keep their pixels' values (e.g., "counts") in dictionaries, with multiple spatial dims unwrapped in the case of Rectangular maps. This PR defines functions to produce xarray DataArray's of individual dict key/val pairs, and a dataset of all the contained DataArrays. 

Closes #1481

## Overview

The SkyMap holds projected data in a dict:
```python
sky_map.data_dict = {
    "counts": NDArray<M energy bins, N unwrapped spatial pixels >
    "exposure_time": NDArray<N unwrapped spatial pixels >
}
```
where `N` is the number of healpix pixels or the number of rectangular grid pixels, the latter of which is equal to `num_az_bins * num_el_bins`

This PR changes the core way that `SkyMap` objects store data from a dictionary to a Dataset. The core dataset (`sky_map.data_1d`) remains an unwrapped tiling of the sky, where the last dimension is always a generic "pixel", regardless of whether it's a Healpix or Rectangular tiling. However, there is now also a `sky_map.data` property, which rewraps this data, fixes the resulting coordinates, etc.

NOTE: Previously, this PR had created functions to convert from the `data_dict` (see strikethrough text below)

~~This PR defines a function which produces an xarray dataset from this dictionary, and handles all re-wrapping of spatial dimensions as necessary.~~

~~This PR also enables calling the `sky_map.to_xarray` method to produce a formatted, xarray dataset containing the data_dict's values. The user must specify the coordinates of each output DataArray (e.g., `{'counts': ['epoch', 'energy_bin_center', 'longitude_bin_center', 'latitude_bin_center'], 'exposure_time': ['epoch', 'longitude_bin_center', 'latitude_bin_center']}`). These spatial coordinates are determined by the SkyMap object itself, while the user inputs the other coordinates.~~

## Updated Files

-  imap_processing/ena_maps/ena_maps.py
   - Change core data storage in SkyMap object to an xr.Dataset
   - ~~Add `to_xarray` and `data_dict_value_to_dataarray` methods to the `AbstractSkyMap`~~
   - ~~Add `spatial_coords` attributes to both types of `SkyMap` objects~~
- Also added the CoordNames Enum here - see #1484 for more details. 

## Testing

Fix tests to work with Dataset method of storage.

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -503,8 +503,9 @@ class AbstractSkyMap(ABC):
     If the map is rectangular, this axis is the raveled 2D grid.
     If the map is Healpix, this axis is the 1D array of Healpix pixel indices.
 
-    The data can be accessed via the .data property, which rewraps the data to the
-    original 2D grid shape if the map is rectangular.
+    The data can be also accessed via the to_dataset method, which rewraps the data to
+    a 2D grid shape if the map is rectangular and formats the data as an xarray
+    Dataset with the correct dims and coords.
     """
 
     @abstractmethod
@@ -517,8 +518,7 @@ class AbstractSkyMap(ABC):
         self.binning_grid_shape: tuple[int, ...]
         self.data_1d: xr.Dataset
 
-    @property
-    def data(self) -> xr.Dataset:
+    def to_dataset(self) -> xr.Dataset:
         """
         Get the SkyMap data as a formatted xarray Dataset.
 
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -526,11 +526,15 @@ class AbstractSkyMap(ABC):
         -------
         xr.Dataset
             The SkyMap data as a formatted xarray Dataset with dims and coords.
+            If the SkyMap is empty, an empty xarray Dataset is returned.
             If the SkyMap is Rectangular, the data is rewrapped to a 2D grid of
             lon/lat (AKA az/el) coordinates.
             If the SkyMap is Healpix, the data is unchanged from the data_1d, but
             the pixel coordinate is renamed to CoordNames.HEALPIX_INDEX.value.
         """
+        if len(self.data_1d.data_vars) == 0:
+            return xr.Dataset()
+
         if self.tiling_type is SkyTilingType.HEALPIX:
             # return the data_1d as is, but with the pixel coordinate
             # renamed to CoordNames.HEALPIX_INDEX.value
@@ -631,7 +635,15 @@ class AbstractSkyMap(ABC):
             # If multiple spatial axes present
             # (i.e (az, el) for rectangular coordinate PSET),
             # flatten them in the values array to match the raveled indices
-            raveled_pset_data = pset_values.data.reshape(1, -1, pointing_set.num_points)
+            non_spatial_axes_shape = tuple(
+                size
+                for key, size in pset_values.sizes.items()
+                if key not in pointing_set.spatial_coords
+            )
+            raveled_pset_data = pset_values.data.reshape(
+                *non_spatial_axes_shape,
+                pointing_set.num_points,
+            )
 
             if value_key not in self.data_1d.data_vars:
                 # Initialize the map data array if it doesn't exist (values start at 0)
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -258,7 +258,7 @@ class PointingSet(ABC):
         -------
         non_spatial_coords : dict[str, xr.DataArray]
             Dictionary of coordinate names and their data arrays.
-            E.g.: {"epoch": [12345,], "energy_bin_center": [100, 200, 300]} .
+            E.g.: {"epoch": [12345,], "energy": [100, 200, 300]} .
         """
         non_spatial_coords = {}
         for coord_name in self.data.coords:
@@ -292,8 +292,8 @@ class RectangularPointingSet(PointingSet):
         Currently, the dataset is expected to be tiled in a rectangular grid,
         with data_vars indexed along the coordinates:
             - 'epoch' : time value (1 value per PSET)
-            - 'longitude_bin_center' : (number of longitude/az bins in L1C)
-            - 'latitude_bin_center' : (number of latitude/el bins in L1C)
+            - 'longitude' : (number of longitude/az bins in L1C)
+            - 'latitude' : (number of latitude/el bins in L1C)
     spice_reference_frame : geometry.SpiceFrame
         The reference Spice frame of the pointing set. Default is IMAP_DPS.
 
@@ -398,9 +398,9 @@ class UltraPointingSet(PointingSet):
         Currently, the dataset is expected to be tiled in a HEALPix tessellation,
         with data_vars indexed along the coordinates:
             - 'epoch' : time value (1 value per PSET, from the mean of the PSET)
-            - 'energy_bin_center' : (number of energy bins in L1C)
-            - 'healpix_pixel_index' : HEALPix pixel index
-        Only the 'healpix_pixel_index' coordinate is used in this class for projection.
+            - 'energy' : (number of energy bins in L1C)
+            - 'healpix_index' : HEALPix pixel index
+        Only the 'healpix_index' coordinate is used in this class for projection.
     spice_reference_frame : geometry.SpiceFrame
         The reference Spice frame of the pointing set. Default is IMAP_DPS.
 
@@ -534,7 +534,12 @@ class AbstractSkyMap(ABC):
             the pixel coordinate is renamed to CoordNames.HEALPIX_INDEX.value.
         """
         if len(self.data_1d.data_vars) == 0:
-            return xr.Dataset()
+            # If the map is empty, return an empty xarray Dataset,
+            # with the unaltered spatial coords of the map
+            return xr.Dataset(
+                {},
+                coords={**self.spatial_coords},
+            )
 
         if self.tiling_type is SkyTilingType.HEALPIX:
             # return the data_1d as is, but with the pixel coordinate
@@ -834,9 +839,9 @@ class HealpixSkyMap(AbstractSkyMap):
         # Define binning_grid_shape for consistency with RectangularSkyMap
         self.binning_grid_shape = (self.num_points,)
         self.spatial_coords = {
-            "healpix_pixel_number": xr.DataArray(
+            CoordNames.HEALPIX_INDEX.value: xr.DataArray(
                 np.arange(self.num_points),
-                dims=["healpix_pixel_number"],
+                dims=[CoordNames.HEALPIX_INDEX.value],
             )
         }
 
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -222,6 +222,49 @@ class PointingSet(ABC):
         self.num_points = 0
         self.az_el_points = np.zeros((self.num_points, 2))
         self.data = xr.Dataset()
+        self.spatial_coords: tuple[str, ...] = ()
+
+    @property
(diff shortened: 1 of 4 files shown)


imap_processing/ena_maps/utils/coordinates.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class CoordNames(Enum):
+    """Enumeration of the names of the coordinates in the L1C and L2 ENA datasets."""
+
+    TIME = "epoch"
+    ENERGY = "energy_bin_center"
+    HEALPIX_INDEX = "healpix_pixel_index"
+
+    # The names of the az/el angular coordinates may differ between L1C and L2 data
```
greglucas: Why would they differ? Here they are identical, so this seems like it would be confusing.

greglucas (commented):
Is there a reason we need `data_dict`, or could we make that a `dataset` right away instead? The xarray dataset is basically a nice dictionary holding things. I think it'd be nice if we could start storing all the quantities in the xarray datasets right away rather than our own intermediate storage object.

nkerman:
> Is there a reason we need `data_dict`, or could we make that a `dataset` right away instead? The xarray dataset is basically a nice dictionary holding things. I think it'd be nice if we could start storing all the quantities in the xarray datasets right away rather than our own intermediate storage object.

This only really applies to the `RectangularSkyMap`, but the values in the dictionary must be spatially unwrapped (`n_energy` x `n_pixels`) for efficient binning, while the Dataset should be re-wrapped to (`epoch` x `n_energy` x `n_az` x `n_el`).

imap_processing/ena_maps/utils/coordinates.py
nkerman: I don't know if they should differ conceptually, but given that the ENA instruments are still determining all their L2 data structures, names, shapes, etc. - I thought allowing them to be defined independently but all in this enum was the safest approach.

greglucas:
What about making the dataset have those dimensions (energy, pixel) in these routines and upon creation then and transform it at the end for output as desired? I think there some nice things for sticking with datasets/arrays when possible here rather than our own separate dictionaries.

nkerman:
> What about making the dataset have those dimensions (energy, pixel) in these routines and upon creation then and transform it at the end for output as desired? I think there some nice things for sticking with datasets/arrays when possible here rather than our own separate dictionaries.

That's a good point - Let me try that and see if it gets messy. So non-spatial dimensions and coordinates would get carried through the projection as spatially-unwrapped DataArrays, and this `to_xarray` method or `.data` property would perform any rewrapping at the end as necessary. 

Zooming out, I worry a bit about my timing and prioritization here, ahead of our mid-April L2 timeline - I think I need to just pick a way that works soon and write the L2 Ultra code with that framework. Unfortunately, I don't think I'm a fast enough programmer to get that done, while changing the structure of the map code, which requires fixing merge conflicts and rebuilding broken tests.

nkerman:
Okay, @greglucas - I finally got this all working using xarray datasets as the core data storage system for the SkyMaps! It requires the PointingSets to have some additional properties, and I'm somewhat worried to see how well this will merge with the other PR, but I think it will all work. I'll update the description of this PR to reflect the changes.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -435,19 +543,35 @@ def project_pset_values_to_map(
...
+            )
 
-            if value_key not in self.data_dict:
+            if value_key not in self.data_1d.data_vars:
                 # Initialize the map data array if it doesn't exist (values start at 0)
                 output_shape = (*raveled_pset_data.shape[:-1], self.num_points)
-                self.data_dict[value_key] = np.zeros(output_shape)
+                self.data_1d[value_key] = xr.DataArray(
+                    np.zeros(output_shape),
+                    dims=pointing_set.unwrapped_dims_dict[value_key],
```
tmplummer, line 659: Is this working correctly? These should be the dimension names getting passed in. For example `dims=["epoch", "energy", "spatial"]`.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -361,19 +409,79 @@ class AbstractSkyMap(ABC):
...
         self.sky_grid: spatial_utils.AzElSkyGrid
         self.num_points: int
+        self.non_spatial_coords: dict[str, xr.DataArray | NDArray]
+        self.spatial_coords: dict[str, xr.DataArray | NDArray]
         self.binning_grid_shape: tuple[int, ...]
-        self.data_dict: dict[str, NDArray]
+        self.data_1d: xr.Dataset
+
+    @property
+    def data(self) -> xr.Dataset:
```
tmplummer: IMO, as implemented, this is not a property and should be named something like `to_dataset()`. I am wondering if the Dataset could instead be an attribute of the SkyMap that gets generated early on and the unwrapped views of the underlying ndarrays can be provided to the projection function so that no re-wrapping needs to be done.
For this PR:
- I suggest not wrapping this with the property decorator and renaming.
- Write a PR about the unwrapped arrays just being a view into the correctly shaped numpy arrays that are stored in the Dataset.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -435,19 +543,35 @@ def project_pset_values_to_map(
...
             # If multiple spatial axes present
             # (i.e (az, el) for rectangular coordinate PSET),
             # flatten them in the values array to match the raveled indices
-            raveled_pset_data = pset_values.data.reshape(-1, pointing_set.num_points)
+            non_spatial_axes_shape = tuple(
+                size
+                for key, size in pset_values.sizes.items()
+                if key not in pointing_set.spatial_coords
+            )
+            raveled_pset_data = pset_values.data.reshape(
```
tmplummer, line 649: I *think* that this implementation enforces that the spatial dimensions of the PSET data be the last dimension(s)? I think that using that fact could possible simplify how the `unwrapped_dims_dict()` function is implemented. No action necessary for this PR, but something to think about.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -218,6 +219,49 @@ def __init__(self, dataset: xr.Dataset, spice_reference_frame: geometry.SpiceFra
...
+        variable_dims = {}
+        for var_name in self.data.data_vars:
+            pset_dims = self.data[var_name].dims
+            non_spatial_dims = tuple(
+                dim for dim in pset_dims if dim not in self.spatial_coords
+            )
+
+            variable_dims[var_name] = (
+                *non_spatial_dims,
+                CoordNames.GENERIC_PIXEL.value,
```
tmplummer, line 248: I'm not sure that i grok what this `CoordNames` part of the tuple is for.

imap_processing/ena_maps/utils/coordinates.py
```diff
@@ -0,0 +1,19 @@
...
+    GENERIC_PIXEL = "pixel"
+
+    TIME = "epoch"
+    ENERGY = "energy_bin_center"
+    HEALPIX_INDEX = "healpix_pixel_index"
+
+    # The names of the az/el angular coordinates may differ between L1C and L2 data
+    AZIMUTH_L1C = "longitude_bin_center"
+    ELEVATION_L1C = "latitude_bin_center"
+    AZIMUTH_L2 = "longitude_bin_center"
```
tmplummer: I think that we should be moving away from using azimuth and elevation. Map coordinates are definitively named latitude, longitude. I also think we should be removing `bin_center`. The CDF variable attributes will document that `latitude` is the pixel center and another variable will define the DELTA_PLUS and DELTA_MINUS from the center.

tmplummer (commented):
A few comments to consider. I understand that some of my comments may not play well with the merge conflicts and whatnot, so feel free to turn them into tickets if you don't think they fit here.

imap_processing/ena_maps/ena_maps.py
nkerman: These are all working in my tests - probably the best example is that, without passing in any dims explicitly, this code correctly handles dims for both healpix and rectangular maps:

```python
rect_map = ena_maps.RectangularSkyMap(
    spacing_deg=2,
    spice_frame=geometry.SpiceFrame.ECLIPJ2000,
)
hp_map = ena_maps.HealpixSkyMap(
    nside=32,
    spice_frame=geometry.SpiceFrame.ECLIPJ2000,
)
for ultra_pset in ultra_psets:
    rect_map.project_pset_values_to_map(
        ultra_pset,
        value_keys=["counts"],
    )
    hp_map.project_pset_values_to_map(
        ultra_pset,
        value_keys=["counts"],
    )
```

Here's the core data_1d structures:
Screenshot 2025-03-19 at 11 01 21 AM


And the data properties:
Screenshot 2025-03-19 at 10 59 42 AM

imap_processing/ena_maps/ena_maps.py
nkerman: Dims in a `PointingSet`
1. Rectangular: `rect_pset.data['counts'].dims = ('epoch', 'energy', 'lon', 'lat')`. 
2. Healpix: `hp_pset.data['counts'].dims = ('epoch', 'energy', 'healpix_idx')`. 

For the unwrapped, 1D map data structure, we need to drop the spatial dimensions, and replace them with a single generic pixel dimension. Dims in `SkyMap's data_1d` (either type)
1. `any_map.data_1d['counts'].dims = ('epoch', 'energy', 'pixel')`. 

The above step is what you're looking at here - we get rid of the spatial dims of the PSET, and add a generic "pixel" dim. 

Then, for the rewrapped `data` structure, this should get put back to:
1. Rectangular: `rect_map.data['counts'].dims = ('epoch', 'energy', 'lon', 'lat')`. 
3. Healpix: `hp_map.data['counts'].dims = ('epoch', 'energy', 'healpix_idx')`.

imap_processing/ena_maps/utils/coordinates.py
nkerman: I have a ticket #1491 up for this! I'd prefer to do it separately and after the two open PRs, as I think they're at risk of getting seriously confusing. 

I'll add in your notes about bin_center as well to the ticket!

imap_processing/ena_maps/ena_maps.py
nkerman: You mean it could be something like:
```python
pset_dims = pset.data['key'].dims
if pset is rectangular:
  return (pset_dims[:-2], "pixel")
elif pset is healpix:
  return (pset_dims[:-1], "pixel")
```
right?

I thought about that, but it seemed safer/more guaranteed to be correct to just have the `PointingSet` know which dim names are spatial.

imap_processing/ena_maps/ena_maps.py
nkerman: Will do!

imap_processing/ena_maps/ena_maps.py
tmplummer: Ah, yes. I see now that `CoordNames.GENERIC_PIXEL.value` is just `"pixel"`. I am not used to Enums allowing string values.

imap_processing/ena_maps/ena_maps.py
tmplummer: Yes, that is what I had in mind. Now that I understand it, your implementation is good.

tmplummer (approved):
LGTM

PR #1486, opened 2025-03-17
Ultra l1c - Basic spacecraft pset data

### LINKED ISSUE #1331: Ultra l1c pset spacecraft
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Generates l1c spacecraft pset dataset

## New Files
- spacecraft_pset.py
   - Creates spacecraft psets

## Deleted Files
- pset.py
   - Renamed to spacecraft_pset.py since we will have psets in two frames (spacecraft and helio)

## Updated Files
- ultra_l1c.py
   - Now structured to include spacecraft_pset.py and heliosphere_pset.py
- de.py
   - Added energies for spacecraft and helio frames that I seemed to have missed the first time around.
- ultra_l1_utils.py
   - Updated to be able to create l1c pset datasets
- ultra_l1c_pset_bins.py
   - Added placeholder for background rates
   - Added ability to calculate center points for lat/lon corresponding to healpix indices
   - Calculate energy_bin_geometric_means
   - Added attributes 
- imap_ultra_l1c_variable_attrs.yaml
   - Added attributes 
   - Added attributes 
   
## Testing
test_ultra_l1c_pset_bins.py
test_spacecraft_pset.py
conftest.py


laspsandoval:
pre-commit.ci autofix

imap_processing/tests/conftest.py
```diff
@@ -142,6 +142,15 @@ def test_data_paths():
             imap_module_directory
             / "tests/hi/data/l1/imap_hi_l1b_45sensor-de_20250415_v999.cdf",
         ),
+        (
+            "https://api.dev.imap-mission.com/download/test_data/ultra-90_raw_event_data_shortened.csv",
+            imap_module_directory
+            / "tests"
+            / "ultra"
+            / "test_data"
```
tmplummer: I noticed after syncing my repo with `dev` and running tests that the ultra files that get downloaded are not in the `.gitignore`. I modified my test_data directory for Hi to simply be `data` which is already in the `.gitignore`. It would be nice if you do that as well or at least add another entry so that the ultra files get ignored.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -92,6 +103,10 @@ def get_spacecraft_histogram(
     # Compute number of HEALPix pixels that cover the sphere
     n_pix = hp.nside2npix(nside)
 
+    # Calculate the corresponding longitude (az) latitude (el)
+    # center coordinates
+    latitude, longitude = hp.pix2ang(nside, np.arange(n_pix), lonlat=True)
```
tmplummer: Just want to check that `pix2ang` returns `(latitude, longitude)` not `(longitude, latitude)`. From the docs and the `lonlat=True` keyword, I would expect the latter.

imap_processing/ultra/utils/ultra_l1_utils.py
```diff
@@ -98,6 +104,12 @@ def create_dataset(
                 dims=["energy_bin_geometric_mean", "spin_number"],
                 attrs=cdf_manager.get_variable_attributes(key),
             )
+        elif key == "counts":
+            dataset[key] = xr.DataArray(
+                data_dict[key].T,
```
tmplummer: This seems very opaque to put in here that it transposes any array passed in with the `counts` key. Is there a reason that the counts ndarray isn't instantiated with the dimensions that match the desired output?

imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
```diff
@@ -35,29 +36,38 @@ def test_data():
...
 def test_get_spacecraft_histogram(test_data):
     """Tests get_histogram function."""
     v, energy = test_data
 
-    energy_bin_edges, _ = build_energy_bins()
+    energy_bin_edges, _, _ = build_energy_bins()
     subset_energy_bin_edges = energy_bin_edges[:3]
 
-    hist = get_spacecraft_histogram(v, energy, subset_energy_bin_edges, nside=1)
+    hist, _, _, _ = get_spacecraft_histogram(
```
tmplummer: Is there a reason that other outputs from `get_spacecraft_histogram` are not tested?

tmplummer (commented):
A few comments to check up on.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
nkerman: It returns it in `lon, lat` order - but the wording is admittedly pretty vague in the healpy docs. 

>if True, return angles will be longitude and latitude in degree, otherwise, angles will be co-latitude and longitude in radians (default)

imap_processing/tests/conftest.py
laspsandoval: Nice idea!

imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
```diff
@@ -7,6 +7,29 @@ default_attrs: &default
...
+  CATDESC: Latitude corresponding to healpix index with range [-90, 90].
+  FIELDNAM: latitude_bin_center
+  LABLAXIS: latitude bin center
+  UNITS: degrees
+
+longitude_bin_center:
+  <<: *default_float32
+  CATDESC: Longitude corresponding to healpix index with range [0, 360].
+  FIELDNAM: longitude_bin_center
+  LABLAXIS: longitude bin center
```
nkerman: drop the "_bin_center"?

imap_processing/ultra/l1c/spacecraft_pset.py
```diff
@@ -0,0 +1,86 @@
...
+    # Calculate exposure
+    constant_exposure = TEST_PATH / "ultra_90_dps_exposure.csv"
+    df_exposure = pd.read_csv(constant_exposure)
+    exposure_pointing = get_spacecraft_exposure_times(df_exposure)
+
+    # For ISTP, epoch should be the center of the time bin.
+    pset_dict["epoch"] = np.mean(de_dataset.epoch.data[[0, -1]]).astype(np.int64)
+    pset_dict["counts"] = counts
+    pset_dict["latitude_bin_center"] = latitude
+    pset_dict["longitude_bin_center"] = longitude
```
nkerman: Drop "bin_center" here as well?

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
```diff
@@ -193,7 +193,21 @@ energy:
 
 tof_energy:
   <<: *default_float32
-  CATDESC: Energy measured using kinetic energy equation.
+  CATDESC: Energy measured using kinetic energy equation in the instrument frame.
+  FIELDNAM: energy
+  LABLAXIS: energy
+  UNITS: keV
+
+energy_spacecraft:
```
nkerman, line 201: Confirming "energy_spacecraft" and "energy_heliosphere" will always be in separate files - given that they have the same field name.

imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
```diff
@@ -40,11 +55,11 @@ species_bin:
...
 
-exposure_times:
-  <<: *default
-  CATDESC: Exposure time for a spin.
-  FIELDNAM: exposure_times
-  LABLAXIS: exposure times
+exposure:
+  <<: *default_float32
+  CATDESC: Exposure time for a pointing.
+  FIELDNAM: exposure
```
nkerman: In his draft map CDF definition, Tim has this as "exposure_factor". I don't think it needs to be consistent at L1, but I was wondering what caused you to change this from "exposure_time"?

imap_processing/ultra/l1c/spacecraft_pset.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # TODO: calculate sensitivity and interpolate based on energy.
+
+    # Calculate exposure
+    constant_exposure = TEST_PATH / "ultra_90_dps_exposure.csv"
+    df_exposure = pd.read_csv(constant_exposure)
+    exposure_pointing = get_spacecraft_exposure_times(df_exposure)
+
+    # For ISTP, epoch should be the center of the time bin.
+    pset_dict["epoch"] = np.mean(de_dataset.epoch.data[[0, -1]]).astype(np.int64)
```
nkerman: This is no longer super clear to me after this [slack thread](https://lasp.slack.com/archives/C04L7EUMJ8G/p1742853926974909?thread_ts=1742853754.129869&cid=C04L7EUMJ8G). I definitely this this way makes the most sense for a pointing set though, so I'd suggest sticking with the midpoint.

imap_processing/ultra/l1c/spacecraft_pset.py
```diff
@@ -0,0 +1,86 @@
...
+    pset_dict["counts"] = counts
+    pset_dict["latitude_bin_center"] = latitude
+    pset_dict["longitude_bin_center"] = longitude
+    pset_dict["energy_bin_geometric_mean"] = energy_bin_geometric_means
+    pset_dict["background_rates"] = background_rates
+    pset_dict["exposure"] = exposure_pointing
+    pset_dict["healpix"] = healpix
```
nkerman, line 82: Does this end up getting treated as a dimension/coordinate by the DataArray? I think it should.

imap_processing/ultra/utils/ultra_l1_utils.py
```diff
@@ -40,7 +40,13 @@ def create_dataset(
             "epoch": data_dict["spin_start_time"],
         }
         default_dimension = "spin_number"
-
+    elif "healpix" in data_dict:
+        coords = {
+            "healpix": data_dict["healpix"],
+            "energy_bin_geometric_mean": data_dict["energy_bin_geometric_mean"],
+            "epoch": data_dict["epoch"],
```
nkerman, line 49: Nevermind, I see you handle the coords here!

imap_processing/ultra/utils/ultra_l1_utils.py
```diff
@@ -40,7 +40,13 @@ def create_dataset(
             "epoch": data_dict["spin_start_time"],
         }
         default_dimension = "spin_number"
-
+    elif "healpix" in data_dict:
```
nkerman, line 45: For these if/elif/else, can you comment with the 3 different scenarios these dict keys indicate? E.g. if the spin number is present, we're building the extended spin table.

imap_processing/ultra/utils/ultra_l1_utils.py
```diff
@@ -78,7 +84,7 @@ def create_dataset(
     }
 
     for key in data_dict.keys():
-        if key in ["epoch", "spin_number", "energy_bin_geometric_mean"]:
+        if key in ["epoch", "spin_number", "energy_bin_geometric_mean", "healpix"]:
```
nkerman, line 91: Add a comment like "skip this for coordinates"?

nkerman (commented):
Nice, Laura! This is a big jump in functionality PR. 

One general comment in addition to the specific ones:
Given the flux on coordinate/dim names, what do you think about using a centralized Enum like the `CoordNames` enum? I get a little worried with all the hardcoded strings ('epoch', 'healpix', 'energy_geometric_mean', etc.) in all these files.

imap_processing/cdf/config/imap_ultra_l1b_variable_attrs.yaml
laspsandoval: Yes. We will have a file for l1c pset spacecraft and l1c pset heliosphere. So you are right.

imap_processing/ultra/l1c/spacecraft_pset.py
laspsandoval: I'll go with what Greg said and if we need to it can be changed later.

imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
laspsandoval: I had remembered a conversation where Ultra had said they wanted exposure. No real reason for the change, though. I can change it to exposure_factor.

laspsandoval:
> Nice, Laura! This is a big jump in functionality PR.
> 
> One general comment in addition to the specific ones: Given the flux on coordinate/dim names, what do you think about using a centralized Enum like the `CoordNames` enum? I get a little worried with all the hardcoded strings ('epoch', 'healpix', 'energy_geometric_mean', etc.) in all these files.

I think that is a good idea! If ok Iput it into a ticket: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1543.

As you have probably noticed this PR is getting huge. Apologies for that.

imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
nkerman: Doesn't really matter to me, I just need to know what it's called for the input to L2.

nkerman (approved):
LGTM - thanks for addressing the comments, Laura

PR #1488, opened 2025-03-18
Codice l1a support variable validation

### LINKED ISSUE #1461: Add unit test(s) to validate CoDICE support variables
We currently have a unit test that validates the data arrays for all of the "counter" data variables in CoDICE L1a data products (e.g. `h`, `he`, `cno`, `tcrs`, etc.), but we currently don't have any testing set up to validate some of the "support" variables (e.g. `energy_table`, `energy_h`, `NSFOHalfSpin`, etc.). I need to write unit tests for these.

bourque:
This PR adds a new unit test to the suite of CoDICE L1a tests to validate the values of the "support" variables in the data products. These are data variables that get stored in the final L1a CDF files that serve as "support" for analysis and/or future calculations in level 2 products (e.g. energy tables, data quality flags, etc.).

I also added a simplification to how acquisition times are gathered, which now uses a simple lookup table instead of making a calculation from a more complicated lookup table.

Closes #1461


imap_processing/codice/constants.py
```diff
@@ -1023,6 +1014,529 @@
     255: 4294967294,
 }
```
greglucas, line 1018: Can you add a link for where this data/table comes from and how it is organized? Your comment on the variable below this is a great example!

imap_processing/tests/codice/data/validation/imap_codice_l1a_hi-ialirt_20241110193700_v0.0.0.cdf
greglucas, line 1: You added a bunch of these validation files, but I don't see where you are using them. Do you need to remove some xfails or point to these paths somehow? (or maybe I missed where these get picked up)

bourque: These are being used in the new `test_codice_l1a.test_l1a_validate_support_variables` function. Unlike the other unit tests which are currently broken for _some_ data products, this unit test can be run over all products.

greglucas: OK, so you already had all of the filenames added to the `conftest`, the files themselves just hadn't been added yet?

greglucas (approved):
How are you deciding between adding the files in git vs remote because I think you have a mixture right?

imap_processing/tests/codice/data/validation/imap_codice_l1a_hi-ialirt_20241110193700_v0.0.0.cdf
bourque: Correct

bourque:
> How are you deciding between adding the files in git vs remote because I think you have a mixture right?

Currently the only file I am pulling down from the cloud is the L0 test file because it is very large. The other (validation) files are all small enough to be committed in the repo. Though, it would be simple enough to just pull all of them from the cloud, I just haven't gotten around to doing that.

Edit: I opened an issue for this! #1537

PR #1490, opened 2025-03-18
HIT L2 - add algorithm for standard intensity

### LINKED ISSUE #1072: HIT L2 - write standard fluxes algorithm code
Write Standard Fluxes Algorithm Code

L1B divided by 3 constants from table 31 (delta E, Geometry factor, efficiency). That table is for all ranges?

vmartinez-cu:
This PR implements the algorithm for the L2 standard intensity data product. Validation will be handled in a future PR.

Closes #1072 

## New Files
Required ancillary data files were added. Stored locally for now
- imap_hit_l1b-to-l2-standard-dt0-factors_20250219_v002.csv
- imap_hit_l1b-to-l2-standard-dt1-factors-20250219_v002.csv
- iimap_hit_l1b-to-l2-standard-dt2-factors-20250219_v002.csv
- imap_hit_l1b-to-l2-standard-dt3-factors-20250219_v002.csv
- constants.py
   - Contains a dictionary needed for summing particle data

## Updated Files
- hit_utils.py
   - added common functions for handling particle data
- hit_l2.py
   - added functions to create the standard intensity data product
   - replaced "flux" with "intensity" for all L2 products
   - added systematic uncertainties (new variables requested by HIT)

## Testing
- imap_processing/tests/hit/test_hit_l2.py
   - added tests for standard intensity data
   - replaced "flux" with "intensity" for consistency


imap_processing/hit/l2/hit_l2.py
```diff
@@ -129,35 +140,262 @@ def process_summed_flux_data(l1b_summed_rates_dataset: xr.Dataset) -> xr.Dataset
...
+
+    Parameters
+    ----------
+    l2_dataset : xr.Dataset
+        The L2 dataset to calculate the intensity for.
+    ancillary_data_frames : dict
+        Dictionary containing ancillary data for each dynamic threshold state.
+    """
+    for var in l2_dataset.data_vars:
+        if (
```
sdhoyt: Do all the variables that meet this condition have a species in the name? Would it be possible to have a list of all the names the variable could contain rather than what it should not? That would make it a little easier to see what's going on in the code.

imap_processing/hit/l2/hit_l2.py
```diff
@@ -129,35 +140,262 @@ def process_summed_flux_data(l1b_summed_rates_dataset: xr.Dataset) -> xr.Dataset
...
+        energy_max = np.zeros(len(energy_ranges), dtype=np.float32)
+
+        # Sum particle rates for each energy range and add them to the dataset
+        for i, energy_range in enumerate(energy_ranges):
+            summed_rates, summed_rates_delta_minus, summed_rates_delta_plus = (
+                sum_particle_data(l1b_standard_rates_dataset, energy_range)
+            )
+
+            # Create namedtuple to store summed counts and uncertainties
+            summed_rates = SummedRates(
```
sdhoyt: Would it make sense to set this tuple inside `sum_particle_data()` and return the named tuple rather than the unnamed tuple from `sum_particle_data()`?

sdhoyt (approved):
Great work and great docstrings!

imap_processing/hit/l2/hit_l2.py
vmartinez-cu: I plan to update L1B processing for the summed rates product to also use the sum_particle_data() function in hit_utils. For that process, counts are summed rather than rates. Keeping the tuple outside of the common function allows me to name the tuple and it's elements accordingly - SummedCounts for L1B and SummedRates for L2. I'll think about a way to name the tuple more generically that works for both.

sdhoyt: Ah ok that makes sense. I think it's good as then

Issue #1492, opened 2025-03-18
HIT L0 - Add sectorate angles


vmartinez-cu:
The sectorates data variable has declination and azimuth dimensions that are currently index values. Replace these index values with angle values provided by HIT.

PR #1493, opened 2025-03-18
HIT L0 - Add sectorate angles

vmartinez-cu:
This PR replaces index values with angle values for the declination and azimuth coordinates that are dimensions for sectorates.

Closes issue #1492 

## Updated Files
- hit/l0/constants.py
   - Add angle constants
   - Update dimension dtypes to match CDF definitions
- hit/l0/decom_hit.py
   - Use angles for sectorate dimensions
- tests/hit/test_decom_hit.py
   - Add assertions for new angle values

### DIFF
--- a/imap_processing/hit/l0/decom_hit.py
+++ b/imap_processing/hit/l0/decom_hit.py
@@ -4,7 +4,9 @@ import numpy as np
 import xarray as xr
 
 from imap_processing.hit.l0.constants import (
+    AZIMUTH_ANGLES,
     COUNTS_DATA_STRUCTURE,
+    DECLINATION_ANGLES,
     EXPONENT_BITS,
     FLAG_PATTERN,
     FRAME_SIZE,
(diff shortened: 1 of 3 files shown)


imap_processing/hit/l0/decom_hit.py
```diff
@@ -93,6 +95,17 @@ def parse_count_rates(sci_dataset: xr.Dataset) -> None:
                 # Reshape data to 8x15 for declination and azimuth look directions
                 parsed_data = np.array(parsed_data).reshape((-1, *field_meta.shape))
                 dims = ["epoch", "declination", "azimuth"]
+                # Add angle values to coordinates
+                sci_dataset.coords["declination"] = xr.DataArray(
```
tmplummer, line 99: Are any attributes going to be added for these? I know they are for L1 products, but it might be worth adding details about being the center of the bins if in fact that is what the values are.

tmplummer (approved):
Except for the comment/question about CDF attributes for the new coordinates, this looks good to me.

bourque (approved):
LGTM!

sdhoyt (approved):

sdhoyt (approved):

imap_processing/hit/l0/decom_hit.py
vmartinez-cu: Attributes will be added eventually. I'm not sure if they are the center of the bins, so I'll clarify that with HIT. Good suggestion

PR #1495, opened 2025-03-18
Unpack I-ALiRT SWAPI data, update utils

torimarbois:
# Change Summary

## Overview
This is the pre-work necessary before the actual processing involved in producing the SWAPI I-ALiRT data produce. The `process_swapi_ialirt()` function unpacks the packet file according to the SWAPI I-ALiRT packet definition, groups the data into valid full sweeps and creates the final epoch time array that will be passed along in the final result.

## New Files

- ialirt/l0/process_swapi.py

## Updated Files

- packet_definitions/ialirt_swapi.xml
   - update some parameter names to be identical to the SWAPI science packet parameter names, in order to be able to use the swapi processing code that Tenzin has already written
- ialirt/utils/grouping.py
   - add a clarification to the function inputs description
- swapi/swapi_utils.py
   - add the SWAPI I-ALiRT `APID`
- ialirt/unit/test_process_swapi.py
   - condense existing test helper functions
   - create placeholder test for new `process_swapi_alirt()` function
- utils.py
   - fix typo
- pyproject.toml
   - add helpful exception to pre-commit hook

## Testing

New unit test for `process_swapi_ialirt()` function.

### DIFF
--- a/imap_processing/ialirt/l0/process_swapi.py
+++ b/imap_processing/ialirt/l0/process_swapi.py
@@ -2,6 +2,7 @@
 
 import logging
 
+import numpy as np
 import xarray as xr
 from xarray import DataArray
 
@@ -32,6 +33,20 @@ def process_swapi_ialirt(unpacked_data: xr.Dataset) -> dict[str, DataArray]:
     sci_dataset = unpacked_data.sortby("epoch", ascending=True)
 
     grouped_dataset = find_groups(sci_dataset, (0, 11), "swapi_seq_number", "swapi_acq")
+
+    for group in np.unique(grouped_dataset["group"]):
+        # Sequence values for the group should be 0-11 with no duplicates.
+        seq_values = grouped_dataset["swapi_seq_number"][
+            (grouped_dataset["group"] == group)
+        ]
+
+        # Ensure no duplicates and all values from 0 to 11 are present
+        if not np.array_equal(seq_values.astype(int), np.arange(12)):
+            logger.info(
+                f"SWAPI group {group} does not contain all sequence values from 0 to "
+                f"11 without duplicates."
+            )
+            continue
 
     total_packets = len(grouped_dataset["swapi_seq_number"].data)
 
--- a/imap_processing/ialirt/l0/process_swapi.py
+++ b/imap_processing/ialirt/l0/process_swapi.py
@@ -1,0 +1,71 @@
+"""Functions to support I-ALiRT SWAPI processing."""
+
+import logging
+
+import xarray as xr
+from xarray import DataArray
+
+from imap_processing import imap_module_directory
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
+from imap_processing.ialirt.utils.grouping import find_groups
+
+# from imap_processing.swapi.l1.swapi_l1 import process_sweep_data
+from imap_processing.swapi.swapi_utils import SWAPIAPID
+from imap_processing.utils import packet_file_to_datasets
+
+logger = logging.getLogger(__name__)
+
+
+TIME_PER_BIN = 0.167  # seconds
+
+
+def process_swapi_ialirt(file_path: str) -> dict[str, DataArray]:
+    """
+    Process L0 data, extract I-ALiRT variables and calculate coincidence count rate.
+
+    Parameters
+    ----------
+    file_path : str
+        L0 SWAPI file.
+
+    Returns
+    -------
+    swapi_data : dict
+        Dictionary containing all data variables for SWAPI I-ALiRT product.
+    """
+    cdf_manager = ImapCdfAttributes()
+
+    # Packet definition
+    xtce_definition = (
+        f"{imap_module_directory}/ialirt/packet_definitions/ialirt_swapi.xml"
+    )
+    unpacked_data = packet_file_to_datasets(file_path, xtce_definition)
+    sci_dataset = unpacked_data[SWAPIAPID.SWP_IALIRT]
+
+    grouped_dataset = find_groups(sci_dataset, (0, 11), "seq_number", "swapi_acq")
+
+    total_packets = len(grouped_dataset["seq_number"].data)
+
+    # It takes 12 sequence data to make one full sweep
+    total_sequence = 12
+    total_full_sweeps = total_packets // total_sequence
+
+    # Separate out the epoch time associated with each sweep, taken from the 1st packet
+    epoch_values = grouped_dataset["epoch"].data.reshape(total_full_sweeps, 12)[:, 0]
+
+    filtered_epoch = xr.DataArray(
+        epoch_values,
+        name="epoch",
+        dims=["epoch"],
+        attrs=cdf_manager.get_variable_attributes("epoch", check_schema=False),
+    )
+
+    # raw_coin_count = process_sweep_data(grouped_dataset, "coin_cnt")
+    # raw_coin_rate = raw_coin_count / TIME_PER_BIN
+
+    swapi_data = {
+        "epoch_time": filtered_epoch,
+        # more variables to go here
+    }
+
+    return swapi_data
--- a/imap_processing/ialirt/l0/process_swapi.py
+++ b/imap_processing/ialirt/l0/process_swapi.py
@@ -5,66 +5,49 @@ import logging
 import xarray as xr
 from xarray import DataArray
 
-from imap_processing import imap_module_directory
-from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.ialirt.utils.grouping import find_groups
 
 # from imap_processing.swapi.l1.swapi_l1 import process_sweep_data
-from imap_processing.swapi.swapi_utils import SWAPIAPID
-from imap_processing.utils import packet_file_to_datasets
+# from imap_processing.swapi.l2.swapi_l2 import TIME_PER_BIN
 
 logger = logging.getLogger(__name__)
 
 
-TIME_PER_BIN = 0.167  # seconds
-
-
-def process_swapi_ialirt(file_path: str) -> dict[str, DataArray]:
+def process_swapi_ialirt(unpacked_data: xr.Dataset) -> dict[str, DataArray]:
     """
-    Process L0 data, extract I-ALiRT variables and calculate coincidence count rate.
+    Extract I-ALiRT variables and calculate coincidence count rate.
 
     Parameters
     ----------
-    file_path : str
-        L0 SWAPI file.
+    unpacked_data : xr.Dataset
+        SWAPI I-ALiRT data that has been parsed from the spacecraft packet.
 
     Returns
     -------
     swapi_data : dict
         Dictionary containing all data variables for SWAPI I-ALiRT product.
     """
-    cdf_manager = ImapCdfAttributes()
+    logger.info("Processing SWAPI.")
 
-    # Packet definition
-    xtce_definition = (
-        f"{imap_module_directory}/ialirt/packet_definitions/ialirt_swapi.xml"
-    )
-    unpacked_data = packet_file_to_datasets(file_path, xtce_definition)
-    sci_dataset = unpacked_data[SWAPIAPID.SWP_IALIRT]
+    sci_dataset = unpacked_data.sortby("epoch", ascending=True)
 
-    grouped_dataset = find_groups(sci_dataset, (0, 11), "seq_number", "swapi_acq")
+    grouped_dataset = find_groups(sci_dataset, (0, 11), "swapi_seq_number", "swapi_acq")
 
-    total_packets = len(grouped_dataset["seq_number"].data)
+    total_packets = len(grouped_dataset["swapi_seq_number"].data)
 
-    # It takes 12 sequence data to make one full sweep
+    # It takes 12 sequence data to make one full SWAPI sweep
     total_sequence = 12
     total_full_sweeps = total_packets // total_sequence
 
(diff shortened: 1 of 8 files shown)


imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -0,0 +1,71 @@
...
+from imap_processing.ialirt.utils.grouping import find_groups
+
+# from imap_processing.swapi.l1.swapi_l1 import process_sweep_data
+from imap_processing.swapi.swapi_utils import SWAPIAPID
+from imap_processing.utils import packet_file_to_datasets
+
+logger = logging.getLogger(__name__)
+
+
+TIME_PER_BIN = 0.167  # seconds
```
tech3371: can you import this from `swapi_l2.py` file? We can move that to a `constants.py` like other instrument in the future. This way, we both are using the same value and if we update or move it, it won't lose track.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -0,0 +1,71 @@
...
+    ----------
+    file_path : str
+        L0 SWAPI file.
+
+    Returns
+    -------
+    swapi_data : dict
+        Dictionary containing all data variables for SWAPI I-ALiRT product.
+    """
+    cdf_manager = ImapCdfAttributes()
```
tech3371: I don't think you will need this since this I-ALiRT data are formatted in a way to be able to write to DB. You may need it in the future based on what direction you and @laspsandoval go for archival purposes. I would remove this for the time being since Laura doesn't set this in hers.

pyproject.toml
```diff
@@ -127,6 +127,9 @@ checks = ["all", #report on all checks, except the following
     "ES01", # Ignore No extended summary found
     "RT02" ] # Ignore The first line of the Returns section
 exclude = ['__init__' ] # don't report on objects that match any of these regex
+override_SS05 = [  # override SS05 to allow docstrings starting with these words
```
tech3371, line 131: Good find. Many of us were wondering about this too.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -0,0 +1,71 @@
...
+        attrs=cdf_manager.get_variable_attributes("epoch", check_schema=False),
+    )
+
+    # raw_coin_count = process_sweep_data(grouped_dataset, "coin_cnt")
+    # raw_coin_rate = raw_coin_count / TIME_PER_BIN
+
+    swapi_data = {
+        "epoch_time": filtered_epoch,
+        # more variables to go here
+    }
```
tech3371:
```suggestion
    # raw_coin_count = process_sweep_data(grouped_dataset, "coin_cnt")
    # raw_coin_rate = raw_coin_count / TIME_PER_BIN

    swapi_data = {
        "epoch_time": epoch_values,
        # more variables to go here
    }
```

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -0,0 +1,71 @@
...
+        name="epoch",
+        dims=["epoch"],
+        attrs=cdf_manager.get_variable_attributes("epoch", check_schema=False),
+    )
+
+    # raw_coin_count = process_sweep_data(grouped_dataset, "coin_cnt")
+    # raw_coin_rate = raw_coin_count / TIME_PER_BIN
+
+    swapi_data = {
+        "epoch_time": filtered_epoch,
```
tech3371: Looks like Laura is storing hers as `met` I would check with Laura on this.

imap_processing/ialirt/l0/process_swapi.py
tech3371: I made suggested change below based on Laura's approach on other instrument.

tech3371 (commented):
SWAPI code part looks good to me. I think there are some difference in approach that I-ALiRT is taking. That part, I would ask Laura. For example, she is taking in unpacked `xr.Dataset` (return value of `packet_file_to_datasets()`) as input and then does processing. Also the final data dictionary is using `met` instead of `epoch_time`. These are some of things I noticed that I don't have answers to.

imap_processing/ialirt/l0/process_swapi.py
laspsandoval: Agreed. This is all going into a database so anything cdf-related needs to be removed.

imap_processing/ialirt/l0/process_swapi.py
laspsandoval: Yes. We are using met. You can see the structure of the database in:
/Users/lasa6858/sds-data-manager/tests/conftest.py

        algorithm_table = dynamodb.create_table(
            TableName=os.environ["ALGORITHM_TABLE"],
            KeySchema=[
                # Partition key
                {"AttributeName": "apid", "KeyType": "HASH"},
                # Sort key
                {"AttributeName": "met", "KeyType": "RANGE"},
            ],
            AttributeDefinitions=[
                {"AttributeName": "apid", "AttributeType": "N"},
                {"AttributeName": "met", "AttributeType": "N"},
                {"AttributeName": "insert_time", "AttributeType": "S"},
                {"AttributeName": "product_name", "AttributeType": "S"},
            ],

imap_processing/ialirt/packet_definitions/ialirt_swapi.xml
```diff
@@ -83,7 +83,7 @@
 				<xtce:LongDescription>CCSDS Packet Length (number of bytes after Packet length minus 1)</xtce:LongDescription>
 			</xtce:Parameter>
 			<!-- SWAPI -->
-			<xtce:Parameter name="SWAPI_SHCOARSE" parameterTypeRef="uint32">
```
laspsandoval: Take a look at ialirt.xml. Right now these instrument packet definitions (ialirt_swapi.xml, ialirt_hit.xml, etc.) are being used only for testing. But when we actually process the data we will only use the final ialirt.xml. The prefixes are used for keeping track of which parameter belongs to which instrument.  So I think you probably need to revert back to the original ialirt_swapi.xml here unless you find something that is really off. And then also change ialirt.xml so that it matches if a change is needed.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -0,0 +1,71 @@
...
+from imap_processing.swapi.swapi_utils import SWAPIAPID
+from imap_processing.utils import packet_file_to_datasets
+
+logger = logging.getLogger(__name__)
+
+
+TIME_PER_BIN = 0.167  # seconds
+
+
+def process_swapi_ialirt(file_path: str) -> dict[str, DataArray]:
```
laspsandoval: At this point we will already have the xarray data. 

The swapi data is part of the spacecraft packet. So you will need to start with what you expect to use that is already in the xarray structure.

In other words start with this data as an input:

```
@pytest.fixture()
def xarray_data(binary_packet_path, xtce_swapi_path):
    """Create xarray data"""
    apid = 1187

    xarray_data = packet_file_to_datasets(
        binary_packet_path, xtce_swapi_path, use_derived_value=True
    )[apid]
    return xarray_data
```

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    cdf_manager = ImapCdfAttributes()
+
+    # Packet definition
+    xtce_definition = (
+        f"{imap_module_directory}/ialirt/packet_definitions/ialirt_swapi.xml"
+    )
+    unpacked_data = packet_file_to_datasets(file_path, xtce_definition)
+    sci_dataset = unpacked_data[SWAPIAPID.SWP_IALIRT]
+
```
laspsandoval, line 34: Remove everything above here. Sort just to make certain the data is sorted. Like this:
sorted_data = sci_dataset.sortby("time_seconds", ascending=True)

imap_processing/swapi/swapi_utils.py
```diff
@@ -14,6 +14,7 @@ class SWAPIAPID(IntEnum):
     SWP_HK = 1184
     SWP_SCI = 1188
     SWP_AUT = 1192
+    SWP_IALIRT = 1187
```
laspsandoval: This can be deleted since you are starting with a spacecraft packet. We won't use the SWAPI I-ALiRT APID other than just for testing.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -0,0 +1,71 @@
...
+    grouped_dataset = find_groups(sci_dataset, (0, 11), "seq_number", "swapi_acq")
+
+    total_packets = len(grouped_dataset["seq_number"].data)
+
+    # It takes 12 sequence data to make one full sweep
+    total_sequence = 12
+    total_full_sweeps = total_packets // total_sequence
+
+    # Separate out the epoch time associated with each sweep, taken from the 1st packet
```
laspsandoval: We don't need the epoch_values or the data to be put into an xarray. I think follow what I did in my most recent PR. 

https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1451/files#diff-80ba5acb97aaa5702596e6cb7445ab1a8aa241de27289ac8be4419d1d6baffae
def process_swe(accumulated_data: xr.Dataset) -> list[dict]:

laspsandoval (commented):
Good first pass. I think the only issue is that I should have explained that the I-ALiRT parsing and data structure will be very different from SWAPI regular processing. We will pass in a xarray and we have no use for cdfs (at least not in any of our processing code). The algorithm work should be similar, though, so I think this is the most confusing part.

pyproject.toml
laspsandoval: Nicely done! That will make life easier for us all.

imap_processing/ialirt/l0/process_swapi.py
torimarbois: Definitely. Good idea!

imap_processing/ialirt/l0/process_swapi.py
torimarbois: @laspsandoval I'm remembering how the folks from SWPC would prefer epoch/timestamps to receive with the data instead of MET - should we also include that in our schema & instrument dictionaries that are returned from processing?

imap_processing/swapi/swapi_utils.py
torimarbois: @laspsandoval I see that the APID ends up in the database - is that the instrument apid, or the ialirt-instrument apid?

laspsandoval:
Re-run your tests to get rid of this error. It should be fixed by now.
ERROR imap_processing/tests/idex/test_idex_l1a.py::test_validate_l1a_idex_data_variables - requests.exceptions.ConnectionError: HTTPSConnectionPool(host='api.dev.imap-mission.com', port=443): Max retries exceeded with url: /download/test_data/imap_codice_l0_raw_20241110_v001.pkts (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x14be1c950>: Failed to resolve 'api.dev.imap-mission.com' ([Errno 8] nodename nor servname provided, or not known)"))

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    logger.info("Processing SWAPI.")
+
+    sci_dataset = unpacked_data.sortby("epoch", ascending=True)
+
+    grouped_dataset = find_groups(sci_dataset, (0, 11), "swapi_seq_number", "swapi_acq")
+
+    total_packets = len(grouped_dataset["swapi_seq_number"].data)
+
+    # It takes 12 sequence data to make one full SWAPI sweep
```
laspsandoval, line 53: You are going to want to do something like this before you calculate the met because it is possible that there might be data missing from the groups or duplicate data.

    for group in unique_groups:
        # Sequence values for the group should be 0-11 with no duplicates.
        seq_values = grouped_data["swe_seq"][(grouped_data["group"] == group).values]

        # Ensure no duplicates and all values from 0 to 11 are present
        if not np.array_equal(seq_values, np.arange(11)):
            logger.warning(
                f"Group {group} does not contain all values from 0 to "
                f"11 without duplicates."
            )
            continue

laspsandoval (approved):
One small change (delete met calculation for now) and then I think you are good! On to the next one.

imap_processing/ialirt/l0/process_swapi.py
torimarbois: @laspsandoval doesn't the `find_groups()` util function filter out groups with missing packets/non-sequential data? Should we add this duplication filter to that function? I also don't quite understand the point of just throwing a warning when duplicates are found rather than removing the entire group from the set of groups.

laspsandoval: find_groups() groups the data, but it doesn't do anything with it. So we need to do a check afterwards to make certain the sequence values are there. It would be unlikely to have duplicate packets within the group. So you just need to make certain that the seq counters are all accounted for.

torimarbois: And if the seq counters are not all accounted for, we should only throw an error, not skip or throw out the group?

laspsandoval: Just log it and move on. This would be very rare for this to happen because the src_seq_ctr would be correct (as determined by find_groups) so this really should always be correct. It's just nice to do the last check.

laspsandoval: So change this to logger.info. I'll need to go back and change it else where too:

logger.info(
            f"Group {group} does not contain all values from 0 to "
            f"11 without duplicates."
        )

PR #1499, opened 2025-03-19
Ultra l1a - updates to parsing

laspsandoval:
# Change Summary

## Overview
We noticed that the files dropped in s3 by the POC were not processing, which was concerning. So I looked into it and discovered that the binary string is not always long enough to fill all of the parameters. I reached out to Larry Brown who showed me how he handles it in MIDL (appending zeros for the block). So that is what I did as well and my data now matches the test data.

## Updated Files
- decom_tools.py
   - Appends zeros to block when binary blob is insufficient in length.

## Testing
- test_decom_apid_881.py
   - Added new test with test data that I know had a problem.
- ultra45_raw_sc_ultraimgrates_20220530_00.csv
- FM45_UltraFM45_Functional_2024-01-22T0105_20240122T010548.CCSDS
- conftest.py

I ran all of the files in s3 through the code and they decom'd. I also ran all Ultra data that they have provided through the code and they also decom'd. So I feel confident now that the data is being properly parsed.


imap_processing/tests/ultra/unit/conftest.py
```diff
@@ -83,13 +83,21 @@ def xtce_path():
 
 
 @pytest.fixture()
-def rates_test_path():
+def rates_test_paths():
```
greglucas: I don't think you are using these both within the same test, so could you add another fixture with a new name that you would use for your new test?

I also don't think a fixture is necessarily needed for all of these paths if they aren't being re-used. Just hard-code it into the read_csv and it will be much more obvious where the test data is coming from and what is going on.

greglucas (approved):

PR #1500, opened 2025-03-19
Update pointing table to include subsecond fields

### LINKED ISSUE #1441: SPICE repoint - Update to include subsecond start and end fields
### Algorithm Description:
The repoint table will have subsecond fields for repoint start and end times. This needs to be added to the test csv. CSV columns will be:
- repoint_start_sec
- repoint_start_subsec
- repoint_end_sec
- repoint_end_subsec
- repoint_id

The start/end fields should be combined to provide the additional columns:
- repoint_start_time
- repoint_end_time

tmplummer:
# Change Summary

## Overview
This adds the sec and subsec columns to the pointing csv as well as the needed changes to the reader to compute the floating point start and end met columns.

## Updated Files
   - Update documentation to reflect the new columns
   - Compute floating point met columns by combining sec and subsec columns
   - Rename columns: `repoint_start_time` -> `repoint_start_met`, `repoint_end_time` -> `repoint_end_met`
   - Update fixture that generates a fake repoint table to include the correct columns.
- imap_processing/tests/spice/test_data/fake_repoint_data.csv
   - Update fake pointing test csv to have correct columns
   - Check for new and renamed columns
   - Correct expected values to be floating point

Closes: #1441

### DIFF
--- a/imap_processing/spice/repoint.py
+++ b/imap_processing/spice/repoint.py
@@ -24,8 +24,12 @@ def get_repoint_data() -> pd.DataFrame:
     repoint_df : pd.DataFrame
         The repointing csv loaded into a pandas dataframe. The dataframe will
         contain the following columns:
-            - `repoint_start_time`: Starting MET time of each repoint maneuver.
-            - `repoint_end_time`: Ending MET time of each repoint maneuver.
+            - `repoint_start_sec`: Starting MET seconds of repoint maneuver.
+            - `repoint_start_subsec`: Starting MET milliseconds of repoint maneuver.
+            - `repoint_start_met`: Starting MET time of repoint maneuver.
+            - `repoint_end_sec`: Ending MET seconds of repoint maneuver.
+            - `repoint_end_subsec`: Ending MET milliseconds of repoint maneuver.
+            - `repoint_end_met`: Ending MET time of repoint maneuver.
             - `repoint_id`: Unique ID number of each repoint maneuver.
     """
     repoint_data_filepath = os.getenv("REPOINT_DATA_FILEPATH")
@@ -37,6 +41,14 @@ def get_repoint_data() -> pd.DataFrame:
 
     logger.info(f"Reading repointing data from {path_to_spin_file}")
     repoint_df = pd.read_csv(path_to_spin_file, comment="#")
+
+    # Compute times by combining seconds and subseconds fields
+    repoint_df["repoint_start_met"] = (
+        repoint_df["repoint_start_sec"] + repoint_df["repoint_start_subsec"] / 1e3
+    )
+    repoint_df["repoint_end_met"] = (
+        repoint_df["repoint_end_sec"] + repoint_df["repoint_end_subsec"] / 1e3
+    )
 
     return repoint_df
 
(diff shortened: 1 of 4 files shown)


imap_processing/spice/repoint.py
```diff
@@ -24,8 +24,12 @@ def get_repoint_data() -> pd.DataFrame:
     repoint_df : pd.DataFrame
         The repointing csv loaded into a pandas dataframe. The dataframe will
         contain the following columns:
-            - `repoint_start_time`: Starting MET time of each repoint maneuver.
-            - `repoint_end_time`: Ending MET time of each repoint maneuver.
+            - `repoint_start_sec`: Starting MET seconds of repoint maneuver.
+            - `repoint_start_subsec`: Starting MET milliseconds of repoint maneuver.
+            - `repoint_start_met`: Starting MET time of repoint maneuver.
```
maxinelasp: Could you add that `repoint_start_met` and `repoint_end_met` are floats calculated from the start_sec and start_subsec fields? I misunderstood this comment the first time I read it.

imap_processing/spice/repoint.py
```diff
@@ -84,29 +100,29 @@ def interpolate_repoint_data(
...
-            f" time in the repoint table. {bad_times=}, {repoint_df_start_time=}"
+            f" time in the repoint table. {bad_times=}, {repoint_df_start_met=}"
         )
     # Make sure that no query times are after the valid range of the dataframe.
     # We approximate the end time of the table by adding 24 hours to the last
     # known repoint start time.
-    repoint_df_end_time = repoint_df["repoint_start_time"].values[-1] + 24 * 60 * 60
-    if np.any(query_met_times >= repoint_df_end_time):
-        bad_times = query_met_times[query_met_times >= repoint_df_end_time]
+    repoint_df_end_met = repoint_df["repoint_start_met"].values[-1] + 24 * 60 * 60
```
maxinelasp, line 120: why not use `repoint_end_met` here?

imap_processing/spice/repoint.py
```diff
@@ -24,8 +24,12 @@ def get_repoint_data() -> pd.DataFrame:
     repoint_df : pd.DataFrame
         The repointing csv loaded into a pandas dataframe. The dataframe will
         contain the following columns:
-            - `repoint_start_time`: Starting MET time of each repoint maneuver.
-            - `repoint_end_time`: Ending MET time of each repoint maneuver.
+            - `repoint_start_sec`: Starting MET seconds of repoint maneuver.
+            - `repoint_start_subsec`: Starting MET milliseconds of repoint maneuver.
+            - `repoint_start_met`: Starting MET time of repoint maneuver.
+            - `repoint_end_sec`: Ending MET seconds of repoint maneuver.
+            - `repoint_end_subsec`: Ending MET milliseconds of repoint maneuver.
```
maxinelasp, line 37: `subsec` has a lot of meanings and is often not ms. Could you use something like `repoint_end_ms` instead?

maxinelasp (approved):
Looks good, just a few comments/questions

imap_processing/spice/repoint.py
tmplummer: I suppose that I could rename these, but I believe that the CSV which comes from the MOC has the column names in the first line. Not 100% sure as I don't see it specifically documented.

imap_processing/spice/repoint.py
tmplummer: I can try to be more explicit in the docstrings about this. The repoint table specifies the start/end time of a repoint maneuver, not the bounderies of a pointing. It will likely be common that we have science data from after the last end time in the repoint table but I don't think we will ever have science data from the next pointing without getting an updated repoint table.

imap_processing/spice/repoint.py
maxinelasp: If there's an explicit reason to name it this way then that is fine!

imap_processing/spice/repoint.py
maxinelasp: Ahh, I see, ok. It might be better to rename some variables away from "repoint" which here means both "a repointing maneuver" and "the day between repointing maneuvers" to something else (like, "observational day" is used by GLOWS)

PR #1511, opened 2025-03-20
Lo - Add SHCOARSE and MET to L1A Direct Events

### LINKED ISSUE #1506: Lo - Add SHCOARSE and MET fields to Lo L1A DE
An `shcoarse` and `met` field should be added to the Lo L1A DE CDF. The `shcoarse` field should be per packet to eventually use as a coordinate for the CCSDS header information and the `met` should be the shcoarse, but on a per packet group frequency to use for L1A validation and to determine which spin a direct event was measured in.

sdhoyt:
# Change Summary

## Overview
This PR adds `shcoarse` to the Lo L1A DE dataset and adds a `met` field. The `shcoarse` is retained per packet and the `met` is the `shcoarse`, but only one time per packet group in retained, similar to the `epoch` field. The `met` field is needed to officially validate the L1A DE and will be used in L1B to determine which spin a DE was measured in.

## Updated Files
- imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
   - changed shcoarse depends to shcoarse instead of epoch. Epoch will be per packet group and shocoarse will be per packet so their lengths will differ.
   - added a met field which is just the shcoarse, but on a per packet group interval. This is mostly being used for validation
- imap_processing/lo/l1a/lo_l1a.py
   - removed shcoarse as a variable to be dropped from the dataset

## Testing
- imap_processing/tests/lo/test_lo_science.py
   - added met to the unit test
- ran cli and examined resulting CDF

Closes #1506

### DIFF
--- a/imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
@@ -110,7 +110,7 @@ esa_step_coord:
 shcoarse:
   <<: *default_uint32
   CATDESC: Mission Elapsed Time
-  DEPEND_0: shcoarse
+  DEPEND_1: shcoarse
   FIELDNAM: Spacecraft Time
   Units: 'ns'
   VAR_TYPE: support_data
--- a/imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
@@ -110,7 +110,7 @@ esa_step_coord:
 shcoarse:
   <<: *default_uint32
   CATDESC: Mission Elapsed Time
-  DEPEND_0: epoch
+  DEPEND_0: shcoarse
   FIELDNAM: Spacecraft Time
   Units: 'ns'
   VAR_TYPE: support_data
--- a/imap_processing/lo/l0/lo_science.py
+++ b/imap_processing/lo/l0/lo_science.py
@@ -410,9 +410,9 @@ def combine_segmented_packets(dataset: xr.Dataset) -> xr.Dataset:
     dataset.coords["epoch"] = dataset["epoch"].values[seg_starts]
     # drop any group of segmented epochs that aren't sequential
     dataset.coords["epoch"] = dataset["epoch"].values[valid_groups]
-    # Set met to the first segment start times for the valid groups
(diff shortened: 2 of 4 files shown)


greglucas (approved):
Are there cases where met != epoch?

I'm a bit confused why/how met is dependent on shcoarse. In your testing you have shcoarse as a 10-element array, but your met is a 3-element array the same as epoch. How can a 3-element array be dependent on a 10-element array?

sdhoyt:
> Are there cases where met != epoch?
> 
> I'm a bit confused why/how met is dependent on shcoarse. In your testing you have shcoarse as a 10-element array, but your met is a 3-element array the same as epoch. How can a 3-element array be dependent on a 10-element array?

@greglucas Met is dependent on epoch, not shcoarse. met contains the shcoarse values, but filtered to be the length of the number of packet groups (same as epoch) rather than the number of individual packets like shcoarse is.

imap_processing/cdf/config/imap_lo_l1a_variable_attrs.yaml
```diff
@@ -110,7 +110,7 @@ esa_step_coord:
 shcoarse:
   <<: *default_uint32
   CATDESC: Mission Elapsed Time
-  DEPEND_0: epoch
+  DEPEND_1: shcoarse
```
greglucas, line 113: @sdhoyt, I was looking at this and confused. I thought this was the new MET because the CATDESC is Mission Elapsed Time.

How can shcoarse be dependent on itself?

sdhoyt: My plan was to make shcoarse a coordinate that the CCSDS header info would depend on, but I want to check with Lo to see if they want the headers and I was going to circle back later to finish that anyway. From what I've seen it looks like coordinates are supposed to depend on themselves. It's possible I'm doing that wrong though.

sdhoyt: And yeah, I need to update the descriptions to make things more clear

greglucas: > From what I've seen it looks like coordinates are supposed to depend on themselves. It's possible I'm doing that wrong though.

👍 It just looked odd to me, if that is what it is supposed to be then no problem at all!

PR #1524, opened 2025-03-22
Lo L1B - Average Spin Duration Calculation

### LINKED ISSUE #1508: Lo - Get Average Spin Durations for Each ASC
Get the Average spin duration for each Aggregated Science Cycle (ASC), which corresponds to  each epoch of the L1A spin data

sdhoyt:
# Change Summary

## Overview
This PR calculates the average spin duration for each Aggregated Science Cycle (ASC) which corresponds to each epoch in the L1A spin data.

## Updated Files
   - function was added to combine the acq start/end time seconds and sub seconds fields from L1A into a single time in seconds
   - function was added to calculate the average spin duration for each ASC

## Testing
New unit tests added for functions

Closes #1508


vmartinez-cu (approved):
Nice job!

PR #1527, opened 2025-03-24
MNT: Update SWAPI DataLevel metadata

greglucas:
I noticed that the L2 file had the wrong data level in the description field when I was copying information over. Then I also noticed that other instruments don't put the quotes around the data level value, so I removed those too.


tech3371 (approved):
Thank you!

PR #1528, opened 2025-03-24
MNT: Update glows and mag DataLevel metadata for CDF properties

greglucas:
I noticed these were the only instruments that had "L" as a prefix in the DataLevel attribute. This updates these instruments to remove the "L" to be consistent with the other instruments. I'm not sure which way is official, so if we need to update the other instruments to add an "L", we can do that too, but I took the low-hanging fruit for now.

ping @tech3371 who may know better the requirements on naming these.


tech3371 (approved):
I think everyone has done without `L`, so it makes sense to remove that to be consistent. Thank you again for catching these details.

maxinelasp (approved):

greglucas:
Superseded by #1535

PR #1532, opened 2025-03-25
FIX: L0 is not a valid processing step, we only process l1a+

greglucas:
Noticed this when going through the cli options. It is nicer if we don't print this as an option if we don't do anything with it on any of the instruments.

### DIFF
--- a/imap_processing/__init__.py
+++ b/imap_processing/__init__.py
@@ -19,7 +19,6 @@ from imap_processing._version import __version__, __version_tuple__  # noqa: F40
 # Eg. imap_module_directory = /usr/local/lib/python3.11/site-packages/imap_processing
 imap_module_directory = Path(__file__).parent
 
-# TODO: should we move this to imap-data-access?
 PROCESSING_LEVELS = {
     "codice": ["l1a", "l1b", "l2"],
     "glows": ["l1a", "l1b", "l2"],
--- a/imap_processing/__init__.py
+++ b/imap_processing/__init__.py
@@ -27,7 +27,7 @@ PROCESSING_LEVELS = {
     "hit": ["l1a", "l1b", "l2"],
     "idex": ["l1a", "l1b", "l2a", "l2b"],
     "lo": ["l1a", "l1b", "l1c", "l2"],
-    "mag": ["l1a", "l1b", "l1c", "l2pre", "l2"],
+    "mag": ["l1a", "l1b", "l1c", "l1d", "l2"],
     "swapi": ["l1", "l2", "l3a", "l3b"],
     "swe": ["l1a", "l1b", "l2"],
     "ultra": ["l1a", "l1b", "l1c", "l2"],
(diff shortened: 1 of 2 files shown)


maxinelasp (commented):
Per the comment, we should actually probably get this from imap-data-access

imap_processing/__init__.py
```diff
@@ -21,14 +21,14 @@
 
 # TODO: should we move this to imap-data-access?
 PROCESSING_LEVELS = {
-    "codice": ["l0", "l1a", "l1b", "l2"],
-    "glows": ["l0", "l1a", "l1b", "l2"],
-    "hi": ["l0", "l1a", "l1b", "l1c", "l2"],
-    "hit": ["l0", "l1a", "l1b", "l2"],
-    "idex": ["l0", "l1a", "l1b", "l2a", "l2b"],
-    "lo": ["l0", "l1a", "l1b", "l1c", "l2"],
-    "mag": ["l0", "l1a", "l1b", "l1c", "l2pre", "l2"],
```
maxinelasp, line 30: If you're updating this anyway, could we update L2pre here?

greglucas:
> Per the comment, we should actually probably get this from imap-data-access

I'm not so sure. This is only used by the CLI for nice messages if you try to process a level that isn't valid for your instrument. I'd almost prefer to just get rid of this entirely and rely on the valid data levels in imap-data-access: https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/440d53019167719fec7782c6b2d9ab2f6ab962c2/imap_data_access/__init__.py#L79-L96

But, this is a bit more accurate.

imap_processing/__init__.py
```diff
@@ -21,14 +21,14 @@
...
-    "swapi": ["l0", "l1", "l2", "l3a", "l3b"],
-    "swe": ["l0", "l1a", "l1b", "l2"],
-    "ultra": ["l0", "l1a", "l1b", "l1c", "l2"],
+    "codice": ["l1a", "l1b", "l2"],
+    "glows": ["l1a", "l1b", "l2"],
+    "hi": ["l1a", "l1b", "l1c", "l2"],
+    "hit": ["l1a", "l1b", "l2"],
+    "idex": ["l1a", "l1b", "l2a", "l2b"],
+    "lo": ["l1a", "l1b", "l1c", "l2"],
+    "mag": ["l1a", "l1b", "l1c", "l2pre", "l2"],
```
greglucas:
```suggestion
    "mag": ["l1a", "l1b", "l1c", "l1d", "l2"],
```

imap_processing/__init__.py
greglucas: Updated to l1d on the line below. It looks like you can only make suggestions on additions and not deletions I guess.

maxinelasp (approved):
Sounds good, I didn't realize this was only used in CLI messages. I'd remove the TODO in that case since decided not to use it.

imap_processing/__init__.py
```diff
@@ -21,14 +21,14 @@
 
 # TODO: should we move this to imap-data-access?
```
greglucas:
```suggestion
```

PR #1535, opened 2025-03-25
MNT: Remove Data_level as a field from the CDF variables

greglucas:
This is unused and is not a required data attribute. We were inconsistent in the naming, so remove it to avoid confusion.

supersedes #1528


vmartinez-cu (approved):
Good catch! I didn't realize this wasn't a required ISTP metadata field

maxinelasp (approved):
Wow, really is unused. Thanks for the pass!

PR #1538, opened 2025-03-25
Lo L1B - Create Lo Spin Bin Direct Event Field

### LINKED ISSUE #1509: Lo - Set the Spin Bin for L1B DE
Set the spin bin for the L1B DE product. This involves calculating the spin phase for each Direct Event and binning the DE into a spin bin (0 - 60)

sdhoyt:
# Change Summary

## Overview
This PR calculates the spin phase and spin bin for each direct event. There are 60 spin bins that the spin can be binned into for each direct event.

## Updated Files
   - Add spin phase calculation
   - Add binning for spin bin

## Testing
- Added new unit tests for both spin phase and spin bin functions

Closes #1509

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -50,7 +50,6 @@ def lo_l1b(dependencies: dict, data_version: str) -> list[Path]:
         # Get the average spin durations for each epoch
         avg_spin_durations = get_avg_spin_durations(acq_start, acq_end)  # noqa: F841
         # get spin phase for each DE
-        # get spin phase for each DE
         spin_phase = get_spin_phase(l1a_de)
         # calculate and set the spin bin based on the spin phase
         # spin bins are 0 - 60 bins
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -49,7 +49,7 @@ def lo_l1b(dependencies: dict, data_version: str) -> list[Path]:
         acq_start, acq_end = convert_start_end_acq_times(spin_data)
         # Get the average spin durations for each epoch
         avg_spin_durations = get_avg_spin_durations(acq_start, acq_end)  # noqa: F841
-        # get spin phase for each DE
+        # get spin phase (0 - 360 degrees) for each DE
         spin_phase = get_spin_phase(l1a_de)
         # calculate and set the spin bin based on the spin phase
         # spin bins are 0 - 60 bins
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -184,7 +184,7 @@ def get_spin_phase(l1a_de_data: xr.Dataset) -> Union[np.ndarray[np.float64], Any
     for asc_de_times in de_time_asc_groups:
         # DE Time is 12 bit DN. The max possible value is 4096
         spin_phase.extend(asc_de_times / 4096 * 360)
-    return np.array(spin_phase).astype(np.float64)
+    return np.array(spin_phase, dtype=np.float64)
 
 
 def set_spin_bin(l1b_de: xr.Dataset, spin_phase: np.ndarray) -> xr.Dataset:
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -189,7 +189,7 @@ def get_spin_phase(l1a_de_data: xr.Dataset) -> Union[np.ndarray[np.float64], Any
 
 def set_spin_bin(l1b_de: xr.Dataset, spin_phase: np.ndarray) -> xr.Dataset:
     """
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -49,6 +50,11 @@ def lo_l1b(dependencies: dict, data_version: str) -> list[Path]:
         # Get the average spin durations for each epoch
         avg_spin_durations = get_avg_spin_durations(acq_start, acq_end)  # noqa: F841
         # get spin phase for each DE
+        # get spin phase for each DE
```
vmartinez-cu: Looks like there's a duplicate comment here

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -158,6 +164,54 @@ def get_avg_spin_durations(
...
+    return spin_phase
+
+
+def set_spin_bin(l1b_de: xr.Dataset, spin_phase: np.ndarray) -> xr.Dataset:
+    """
+    Set the spin bin (0 - 60 bins) for each Direct Event where each bin is 6 degrees.
+
+    Parameters
+    ----------
+    l1b_de : xarray.Dataset
```
vmartinez-cu: Minor suggestion - consider using consistent parameter names for the input dataset. In `get_spin_phase` you have `l1a_de_data` and here you have `l1b_de`. Both are descriptive and work, but you could use one format for both.

vmartinez-cu (approved):
LGTM, nice job!

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -158,6 +163,54 @@ def get_avg_spin_durations(
     return avg_spin_durations
 
 
+def get_spin_phase(l1a_de: xr.Dataset) -> Union[np.ndarray[np.float64], Any]:
```
greglucas: I think we've been using "phase" to be the `[0, 1)` value, and "angle" to be the `[0, 360)` version. So do we want to call this function `get_spin_angle()` instead?

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -158,6 +163,54 @@ def get_avg_spin_durations(
...
+        The L1A DE dataset.
+
+    Returns
+    -------
+    spin_angle : np.ndarray
+        The spin angle for each DE.
+    """
+    de_times = l1a_de["de_time"].values
+    # DE Time is 12 bit DN. The max possible value is 4096
+    spin_angle = np.mod(de_times / 4096 * 360, 360).astype(np.float64)
```
greglucas: This seems dangerous to use modulo here? Why would we be getting values outside of 0-4096 and if so is there a separate issue?
```suggestion
    # DE Time is 12 bit DN. The max possible value is 4096
    spin_angle = l1a_de["de_time"].astype(float) / 4096 * 360
```

sdhoyt: I made a bad assumption here. I'm using 4096 (which is correct) because the DE time is a 12 bit number. I'm realizing now that the max a 12 bit number can be is 4095, not 4096, so that fraction will never result in 1, so there's no need to use the mod to turn an angle of 360 into an angle of 0. I'll switch this back to what I had before.

Thanks for catching that!

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -158,6 +163,54 @@ def get_avg_spin_durations(
...
+
+    Returns
+    -------
+    spin_angle : np.ndarray
+        The spin angle for each DE.
+    """
+    de_times = l1a_de["de_time"].values
+    # DE Time is 12 bit DN. The max possible value is 4096
+    spin_angle = np.array(de_times / 4096 * 360, dtype=np.float64)
+    return spin_angle
```
greglucas, line 183: As an FYI, I think we use `.values` too much everywhere in our codebase, we should try to stick with xarray as much as possible. The return value shouldn't be `Any` here either, it is always an array currently? I think it would then be `xr.DataArray` if you take the below suggestion.
```suggestion
    # DE Time is 12 bit DN. The max possible value is 4096
    spin_angle = l1a_de["de_time"].astype(float) / 4096 * 360
    return spin_angle
```

greglucas (approved):

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: I did that so I wasn't carrying over the coordinate and dimension information that's stored in the de_times data array because I don't want that as part of the spin_angle array. Is there a preferred way of sticking with xarray and not carrying that over into the spin_angle variable?

greglucas: Oh, I see... I'm not sure. This seems to indicate the opposite would happen and they were getting dropped when they wanted them preserved 🤷 
https://stackoverflow.com/questions/70289775/dataarray-deletes-attributes-in-simple-computation

Fine to keep this as-is, my approval still stands.

PR #1542, opened 2025-03-26
Mag l1c updates 

maxinelasp:
# Change Summary
Some changes as requested by the team, to complete a few more pieces, and in anticipation of the validation tests. This does not include any validation test work but does set it up.

## Overview

- Changed the code to accept either norm or burst or both files 
- added some enums to clarify pieces
- Added addtional filter options for interpolation
- additional work around tracking vector rates

## Testing

New tests and new validation


imap_processing/mag/constants.py
```diff
@@ -59,6 +59,23 @@ class PrimarySensor(Enum):
     MAGI = 1
 
 
+class VecSec(Enum):
+    """Enum for all valid vector rates (Vectors per second)."""
+
+    ONE_VEC_PER_S = 1
```
greglucas, line 65: The enum is already called VecSec, so I think you could shorten things and go `VecSec.ONE` everywhere (maybe make the enum `VecPerSec`?
```suggestion
    ONE = 1
```

imap_processing/mag/l1c/interpolation_methods.py
```diff
@@ -1,16 +1,25 @@
...
+
+logger = logging.getLogger(__name__)
 
 
 def linear(
     input_vectors: np.ndarray,
     input_timestamps: np.ndarray,
     output_timestamps: np.ndarray,
+    input_rate: Optional[VecSec] = None,
+    output_rate: Optional[VecSec] = None,
```
greglucas, line 22: These aren't used in any of the below methods?

Edit: I think I see below you use them for other methods so this is for consistency in the call signatures.

imap_processing/mag/l1c/interpolation_methods.py
```diff
@@ -94,10 +119,101 @@ def cubic(
...
+
+    Returns
+    -------
+    VecSec
+        Estimated rate of the timestamps.
+    """
+    samples_per_second = timestamps.shape[0] / (timestamps[-1] - timestamps[0]) * 1e9
+    per_second = VecSec(
+        POSSIBLE_RATES[(np.abs(POSSIBLE_RATES - samples_per_second)).argmin()]
+    )
```
greglucas, line 142: This is the average rate of timestamps and then finding out which sampling rate was most common during this time period. There can be several issues with this and it feels a bit arbitrary to choose the "average" as what you want to use.

Here is an issue I can think of: half the time we're in 1 sample / s, half the time we're in 4 samples / s. The estimate here would be closest to 2 samples / s, of which your instrument was never actually in that mode. Does that matter?

Is there a reason not to take the first two timestamps subtract them and call that your rate? That seems like it is also arbitrary but at least easier to figure out what was chosen and why.

imap_processing/mag/l1c/interpolation_methods.py
```diff
@@ -94,10 +119,101 @@ def cubic(
...
+            f"the normal mode output rate {output_rate}. "
+            f"Both rates are required"
+        )
+
+    decimation_factor = int(input_rate.value / output_rate.value)
+    cic1 = np.ones(decimation_factor)
+    cic1 = cic1 / decimation_factor
+    cic2 = np.convolve(cic1, cic1)
+    delay = (len(cic2) - 1) // 2
+    input_filtered = input_timestamps
```
greglucas, line 206: Might want to call this `timestamps_filtered`? There are two inputs, timestamps and vectors.

greglucas (commented):
Looks good overall!

Sorry I haven't been able to keep up with the MAG reviews. One general comment would be that you are often passing around raw numpy arrays as times and vectors together into functions and that is where xarray is really nice to keep these things aligned and pass in via the same object. There are also some nice subsetting, interpolation, and resampling/decimation handling on the dataset/dataarray directly too that might simplify some of the structure here. No need to change now, just mentioning it in case it inspires any future ideas to go to the datasets sooner.

imap_processing/mag/l1c/interpolation_methods.py
maxinelasp: Yes, unfortunately due to Enum weirdness they need to all have the same signature.

imap_processing/mag/l1c/interpolation_methods.py
maxinelasp: I agree there are some issues with this method. I have discussed with the team and will be making some minor changes (in particular, there are some rules about when rates can change that I can exploit more heavily here). However, we really shouldn't ever use this method in normal processing. Unless the file is using an older version of the processing code (which is partly why I added this, I don't want it to immediately blow up in that case) we will be getting the rate from the `vectors_per_second` attribute on the CDF file.

imap_processing/mag/l1c/interpolation_methods.py
maxinelasp: Yeah good call!

imap_processing/mag/constants.py
maxinelasp: I will contemplate this one. I think you're right. I might not change it in this PR because I actually have a lot of changes in the next iteration of this code that would make merging very annoying, but perhaps in the next PR I will rework it.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -405,32 +495,37 @@ def find_all_gaps(
...
+    gaps: np.ndarray = np.zeros((0, 3))
+    if vecsec_dict is None:
+        # TODO: when we go back to the previous file, also retrieve expected
+        #  vectors per second
+        # If no vecsec is provided, assume 2 vectors per second
+        vecsec_dict = {0: VecSec.TWO_VECS_PER_S.value}
+
+    end_index = epoch_data.shape[0]
+    for start_time in reversed(sorted(vecsec_dict.keys())):
+        start_index = np.where(start_time == epoch_data)[0][0]
```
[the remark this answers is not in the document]
maxinelasp: This is an assumption we want to make (and fail if it isn't true), but it also was buggy and I fixed it in #1553

maxinelasp:
@bourque , @subagonsouth , @vmartinez-cu and @lacoak21 just bumping this to request a review, since I have a lot of changes to these files I'm trying to merge

imap_processing/mag/l1c/interpolation_methods.py
```diff
@@ -94,10 +119,101 @@ def cubic(
...
+        Output timestamps of shape (m,) to generate interpolated vectors for.
+    input_rate : VecSec, optional
+        Expected rate of input timestamps.
+    output_rate : VecSec, optional
+        Expected rate of output timestamps.
+
+    Returns
+    -------
+    numpy.ndarray
+        Filtered something something.
```
vmartinez-cu: The return in this docstring can be updated to describe the input_filtered and vectors_filtered return values

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -405,32 +495,37 @@ def find_all_gaps(
...
+        gaps = np.concatenate(
+            (
+                find_gaps(
+                    epoch_data[start_index : end_index + 1], vecsec_dict[start_time]
+                ),
+                gaps,
```
vmartinez-cu, line 525: Not required for this PR - I think you could simplify the find gaps function a little by using 
`diffs = abs(np.diff(timeline_data))` 
instead of 
`diffs = abs(timeline_data[:-1] - np.roll(timeline_data, -1)[:-1])` 
I had to do something similar for HIT to calculate the difference of consecutive elements

vmartinez-cu (approved):
LGTM!

imap_processing/mag/l1c/mag_l1c.py
tmplummer, line 519: Just a note that numpy documentation says to use `np.nonzero()` for cases like this where only the `condition` argument is supplied to `np.where()`.

tmplummer (approved):
Not much to add. This all looks good to me!

imap_processing/mag/l1c/interpolation_methods.py
maxinelasp: Haha, good catch.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: Good suggestion, thank you!

Issue #1545, opened 2025-03-26
Lo L1B - retain or retrieve spin number from previous pointing


sdhoyt:
For the L1B direct event product, then spin cycle equation needs the spin start number for the first spin in each ASC. 

The way the code is currently written, the spin numbers will reset for each pointing, however spin numbers will not reset for each pointing and will continue incrementing across pointings. 

The last spin number from the previous pointing needs to be used to get the spin_start for the first ASC spin in the current pointing (`last spin of previous pointing + 1`). This spin number needs to be retained or retrieved and figuring out how to do this is part of this ticket.

The original code for spin cycle was introduced in PR #

PR #1546, opened 2025-03-26
Lo L1B - Add spin cycle to direct events

### LINKED ISSUE #1510: Lo - Set the spin cycle for L1B DE
Set the spin cycle for the Lo L1B DE data product. The ESA steps need to be grouped by ASC (epoch) and used in the equation:

spin_start + 7 + (esa - 1) * 2

where esa is all the ESAs for a ASC and the spin_start is the spin number for the first spin in an ASC which should be 1 + the last spin from the previous ASC. 

The spin number does not reset for each pointing.

sdhoyt:
# Change Summary

## Overview
This PR adds spin cycle to the L1B direct events. Spin cycle is calculated as follows:

`spin_cycle = spin_start + 7 + (esa_step - 1) * 2`

where `spin_start` is the spin number for the first spin in an Aggregated Science Cycle (first spin number of an epoch in the L1A spin dataset) and `esa_step` is the ESA step for a direct event.

The way the code is currently written, the spin numbers will reset for each pointing, however this will not be the case. More work needs be done to figure out how to retain or retrieve the last spin number of the previous pointing. This will be handled in issue #1545 

## Updated Files
   - added spin cycle function

## Testing
Added unit test for spin cycle function

Closes #1510


sdhoyt:
@greglucas Since you brought this up in my previous PR: I wasn't able to get this working with the numpy functions without using `.values` on the xarray data arrays when splitting up the esa steps into aggregated science cycles:

```python
de_asc_groups = np.split(l1a_de["esa_step"].values, np.cumsum(counts)[:-1])
```

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -211,6 +213,52 @@ def set_spin_bin(l1b_de: xr.Dataset, spin_angle: np.ndarray) -> xr.Dataset:
...
+    spin_cycle = []
+    for i, esa_asc_group in enumerate(de_asc_groups):
+        # TODO: Spin Number does not reset for each pointing. Need to figure out
+        #  how to retain this information across days
+        # increment the spin_start by 28 after each aggregated science cycle
+        spin_start = i * 28
+        # calculate the spin cycle for each DE in the ASC group
+        # TODO: Add equation number in algorithm document when new version is
+        # available. Add to docstring as well
+        spin_cycle.extend(spin_start + 7 + (esa_asc_group - 1) * 2)
```
greglucas: This was a bit tricky to follow what was going on, but looks good.

One alternative suggestion instead of the split/cumsum combination would be to create an empty spin_cycle variable initially in your current dataset, and then do something like iterate through the subgroups by downselecting each spin and then updating that subsetted array in place.

```python
for group_start, group_end in ...:
    group_ds = ds.isel(epoch=slice(group_start, group_end))
    group_ds["spin_cycle"] = spin_start + 7 + (group_ds["esa_step"] - 1) * 2)
```

greglucas (approved):

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -211,6 +213,52 @@ def set_spin_bin(l1b_de: xr.Dataset, spin_angle: np.ndarray) -> xr.Dataset:
...
+    l1b_de : xarray.Dataset
+        The L1B DE dataset.
+
+    Returns
+    -------
+    l1b_de : xarray.Dataset
+        The L1B DE dataset with the spin cycle added for each direct event.
+    """
+    counts = l1a_de["de_count"].values
+    # split the esa_steps into ASC groups
```
bourque, line 240: As a reader that knows very little about this instrument I wasn't sure what "ASC" stands for. Maybe spell it out somewhere?

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -216,3 +217,27 @@ def test_spin_bin():
...
+    # in an Aggregated Science Cycle (first spin number of an epoch)
+    # and esa_step is the esa_step for a direct event
+    spin_cycle_expected = np.array([7, 9, 39, 41, 43])
+    spin_cycle_data = xr.Dataset()
+
+    # Act
+    spin_cycle_data = set_spin_cycle(de, spin_cycle_data)
+
+    # Assert
+    np.testing.assert_array_equal(spin_cycle_data["spin_cycle"], spin_cycle_expected)
```
bourque, line 243: I had never thought about tests in this way before (i.e Arrange, Act, Assert). I like it.

bourque (approved):
LGTM!

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: I agree this is better. I'm having trouble getting this to work and want to move on to the next PR, so I wrote an issue to make this update in the future: #1556

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: I have it spelled out in the docstring, but I didn't put the acronym in there. I'll add the acronym in parentheses to the docstring

imap_processing/tests/lo/test_lo_l1b.py
sdhoyt: Nice! This is how I learned to write unit tests. I think it's a helpful way to think about and organize the tests. I think it also helps with readability.

PR #1549, opened 2025-03-27
Updating conftest

maxinelasp:
# Change Summary
As part of MAG validation, I wanted to only download the MAG test files from AWS without requiring everyone's test files (they're very small). In doing this, I reworked some of how the download step works and fixed the file paths. 

This shouldn't affect anyone as it's mostly a slight rearrangement. The MAG files will be added in a later PR. (why doesn't github let me stack reviews....)


greglucas (approved):
I agree it'd be nice to only download the required data for your tests and not everyone's data. Another idea for later is to turn this into a test decorator that would receive the filename and location to save it, so you could use it like this, where that specific marked test needs `mag_filename`:

```python
@pytest.mark.download_data("mag_filename", "path/to/local/file")
def test_mag():
```

PR #1550, opened 2025-03-27
Ultra l1a cleanup

### LINKED ISSUE #1313: Ultra l1a - Attribute changes needed
### Description of the issue

**Kicking off the CDF questions - L1A DE**
It looks like there is a record in this table for each event. How does this work with the data at the beginning of the packet that isn't event by event (SHCOURSE, spin, pkt_len, etc.). Are they just repeated for each event in the packet? Would it be better to break this in to two files, or one file with two 'coordinates' - the event time and the packet time?
There is an 'epoch' and an 'eventtimes' - one is a U64 the other is a float64 - are these the same quantity in different representations?
The phase angle is the only float value here, and it looks like it's the only thing that has been 'calibrated'.  In the packet data it is an integer. Why do this here?

**L1A PHxTOF**
This is a complicated data format, and it's not immediately clear how the packet->CDF processing works.  ISTM that the only way to be sure would be to produce images at L1B and verify them against MIDL. Since this is a diagnostic product and not on the path to Level 2, we should defer this.

**L1A AUX**
Why are the spin bit flags (spinperiodvalid, spinphasevalid, etc.) unpacked into strings? ISTM fine to just leave the 8 bits (including the spares) as a single U8. Same comment about the bitflags that start 'Hardware Mode'
The variables 'ints, intn1, intn2' should probably be renamed 'spin_s, integration_m1, integration_m2'

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

laspsandoval:
# Change Summary

## Overview
Cleanup for L1a. Changes to accommodate the use of packet_file_to_datasets tool and organization.

## Updated Files
- ultra_l1a.py
   - cleanup and organization
- decom_tools.py
   - accommodation for different input format
- decom_ultra.py
   - removed tools that were no longer necessary based on capabilities of packetfile_to_dataset
   - restructuring based on xarray format created by packetfile_to_dataset
- ultra_utils.py
   - removed tools that were no longer necessary based on capabilities of packetfile_to_dataset
   - capitalization change

- ultra_l1b_extended.py
   - cleanup and organization
- imap_ultra_l1a_variable_attrs.yaml, imap_ultra_global_cdf_attrs.yaml
   - added l1a attributes
- de.py
   - capitalization change
- conftest.py
   - update fixtures

## Testing
- test_decom_apid_880.py
- test_decom_apid_881.py
- test_decom_apid_883.py
- test_decom_apid_896.py
- test_ultra_l1a.py
- test_ultra_l1b_extended.py


imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
```diff
@@ -8,27 +8,79 @@ instrument_base: &instrument_base
...
   Instrument_type: "Particles (space)"
 
-# TODO: revisit this to be more specific
-imap_ultra_l1a_sci:
+imap_ultra_l1a_45sensor-aux:
     <<: *instrument_base
-    Data_type: L1A_SCI>Level-1A Science data
-    Logical_source: imap_ultra_l1a_sci
-    Logical_source_description: IMAP-Ultra Instrument Level-1A Science Data.
+    Data_level: 1A
```
greglucas: There are no Data_level's anymore. Remove them from all the other locations in this file too.
```suggestion
```

imap_processing/tests/ultra/unit/conftest.py
```diff
@@ -129,11 +134,21 @@ def decom_test_data(request, xtce_path):
...
+        ULTRA_RATES.apid[0]: process_ultra_rates,
+        ULTRA_TOF.apid[1]: process_ultra_tof,
+        ULTRA_EVENTS.apid[1]: process_ultra_events,
+        ULTRA_RATES.apid[1]: process_ultra_rates,
+    }
 
-    data_packet_list = process_ultra_apids(grouped_data[apid], apid)
-    return data_packet_list, packets
+    process_function = strategy_dict.get(apid, lambda *args: False)
+    data_packet_xarray = process_function(datasets_by_apid[apid])
```
greglucas, line 149: Do you need to iterate over apid?

```python
for apid, ds in datasets_by_apid.items():
```

Edit: I later saw that you are getting a specific apid from `request.param`. Ignore this.

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -1,269 +1,226 @@
...
+        "row": xr.DataArray(np.arange(54), dims=["row"], name="row"),
+        "column": xr.DataArray(np.arange(180), dims=["column"], name="column"),
     }
 
-    sorted_packets = sort_by_time(data, "SHCOARSE")
+    dataset = xr.Dataset(coords=coords)
+
+    # Add scalar keys (2D: epoch x sid)
+    for key in scalar_keys:
+        dataset[key.upper()] = xr.DataArray(
```
greglucas: It looked like you changed most of the other places to be lowercase variable names now, this one too?
```suggestion
        dataset[key] = xr.DataArray(
```

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -1,269 +1,226 @@
...
+    dataset : xarray.Dataset
+        Dataset containing the decoded and decompressed data.
     """
+    decom_data = defaultdict(list)
     if (
         isinstance(ULTRA_RATES.mantissa_bit_length, int)
         and isinstance(ULTRA_RATES.len_array, int)
         and isinstance(ULTRA_RATES.block, int)
         and isinstance(ULTRA_RATES.width, int)
     ):
```
greglucas: if this isn't true do we want to just skip this, or should we raise / log that happening here? It seems like these statements should always be true unless something bad is happening.

greglucas (commented):
Very minor comments. This PR is large but most of it boilerplate changing cases from upper to lower.

imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -264,8 +47,7 @@ def ultra_l1a(
         f"{imap_module_directory}/ultra/packet_definitions/" f"ULTRA_SCI_COMBINED.xml"
     )
 
-    packets = decom.decom_packets(packet_file, xtce)
-    grouped_data = group_by_apid(packets)
+    datasets_by_apid = packet_file_to_datasets(packet_file, xtce)
```
nkerman, line 50: This seems like a great + cleaner approach!

nkerman (approved):
Everything seems reasonable to me! I like the simplification of packet -> dataset conversion!

PR #1551, opened 2025-03-27
Mag L2 first pass

maxinelasp:
# Change Summary
Create an empty MAG L2 CDF file and set up structure for processing. 

## Overview

Added new mag_l2 and mag_l2_data files with some overall structure definitions


maxinelasp:
@bourque @nkerman @lacoak21 pinging for a review so I could hopefully get this merged before MSIM next week! This really just sets up a dataclass and a few empty methods and generates an L2 CDF, so not as complicated as it looks!

imap_processing/cli.py
```diff
@@ -752,6 +753,24 @@ def do_processing(self, dependencies: list[Path]) -> list[xr.Dataset]:
             # Input datasets can be in any order
             datasets = [mag_l1c(input_data[0], input_data[1], self.version)]
 
+        if self.data_level == "l2":
+            # TODO: Overwrite dependencies with versions from offsets file
```
lacoak21, line 757: Ah is see - So this is when you would open up the offset ancillary file and read in the dependencies. @tech3371 here is the mag example that only has an upstream ancillary.

lacoak21 (approved):
Looks good for just generating the L2 CDF

imap_processing/cli.py
maxinelasp: Yeah, so I could set the science files as dependencies if needed to make the system work, and then just not use them here. There will never be a case where we don't have science dependencies, because they're needed to generate the offsets file.

PR #1553, opened 2025-03-27
Correcting minor bug in L1B

maxinelasp:
# Change Summary
Change corrects a small bug in L1B around the vectors per second attribute.

### DIFF
--- a/imap_processing/mag/l1a/mag_l1a_data.py
+++ b/imap_processing/mag/l1a/mag_l1a_data.py
@@ -1107,13 +1107,14 @@ class MagL1a:
         """
         output_str = ""
         last_vectors_per_second = None
-        for start_time, packet in self.packet_definitions.items():
+        for _, packet in self.packet_definitions.items():
             vecsec = packet.vectors_per_second
+            time = packet.start_time.to_j2000ns()
             if vecsec != last_vectors_per_second:
                 if output_str == "":
-                    output_str = f"{start_time}:{vecsec}"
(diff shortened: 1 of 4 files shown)


imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -1107,13 +1107,14 @@ def vectors_per_second_attribute(self) -> str:
...
         output_str = ""
         last_vectors_per_second = None
-        for start_time, packet in self.packet_definitions.items():
+        for _, packet in self.packet_definitions.items():
             vecsec = packet.vectors_per_second
+            time: np.int64 = packet.start_time.to_j2000ns().astype(np.int64)
             if vecsec != last_vectors_per_second:
                 if output_str == "":
-                    output_str = f"{start_time}:{vecsec}"
+                    output_str = f"{time}:{vecsec}"
```
vmartinez-cu, line 1115: Just curious about the name change from `start_time` to `time`. Is "start" no longer a helpful descriptor to keep?

vmartinez-cu (approved):
LGTM!

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: More that, "start_time" is used everywhere and sometimes its coarse seconds and sometimes it's nanoseconds and that was the original source of the confusion in this function, so I got annoyed and changed it 😂

sdhoyt (approved):
Looks good!

PR #1554, opened 2025-03-27
HIT - Refactor processing to use functions in hit utils

### LINKED ISSUE #1552: HIT - remove duplicate code
Update L1B and L2 processing code to use supporting functions in hit utils that are common to both such as adding energy variables to the dataset and summing particle data.

vmartinez-cu:
This PR removes duplicate code by refactoring L1B and L2 processing to use supporting functions in hit utils that handle summing particle data and adding energy variables that use a geometric mean.  The geometric mean replaces the energy index that was initially used, but didn't conform with mission requirements.

Closes #1552 

## Updated Files
- hit_utils.py
   - Add function to sum particle data and add results to the dataset. Used for both L1B and L2 products
- constants.py
   - Rename dictionary for clarity
- hit_l2.py
   - Update processing to use a function in hit utils that handles summing particle data and adding it to the dataset
   - Remove duplicate code
- hit_l1b.py
   - Update processing to use functions in hit utils that handles summing particle data and adding it to the dataset
   - Remove duplicate code
   - Add function to calculate rates for L1B products

## Testing
Corresponding test files were updated to reflect above changes.

### DIFF
--- a/imap_processing/hit/l2/hit_l2.py
+++ b/imap_processing/hit/l2/hit_l2.py
@@ -186,7 +186,7 @@ def calculate_intensities(
 
 def calculate_intensities_for_a_species(
     species_variable: str, l2_dataset: xr.Dataset, ancillary_data_frames: dict
-) -> None:
+) -> xr.Dataset:
     """
     Calculate the intensity for a given species in the dataset.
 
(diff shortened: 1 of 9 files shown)


imap_processing/hit/l2/hit_l2.py
```diff
@@ -601,7 +502,7 @@ def process_standard_intensity_data(
             l2_standard_intensity_dataset, particle, len(energy_ranges)
         )
         # Add standard particle rates and statistical uncertainties to the dataset
-        add_standard_particle_rates_to_dataset(
+        add_summed_particle_data_to_dataset(
```
sdhoyt: I know the way this is written works, but updating the dataset without returning it might cause some confusion down the road (same for other functions in this file). This is just my opinion and I'm not sure what's considered best practice so take or leave this comment.

sdhoyt (approved):
Looks good! Nice work!

imap_processing/hit/l2/hit_l2.py
vmartinez-cu: This is a great suggestion. I was modifying datasets in-place to avoid extra memory usage, but I don't think that should be much of a concern. I've updated functions to return the updated datasets instead

PR #1555, opened 2025-03-27
ENA L0 - Gather pointings

laspsandoval:
# Change Summary

## Overview
Creates a simple function that can be used once the cli.py has been updated.

## Updated Files
- repoint.py
   - Combines xarrays
   - Filters out data outside of the timerange of the pointing defined by repoint_start, repoint_end

## Testing
- test_repoint.py

## Notes on how it will most likely be used:

**cli.py**
```
if self.data_level == “l1a”:
    datasets = ultra_l1a.ultra_l1a(dependencies, self.version, self.pointing_number) 
```

**instrument_l1a.py**
```
datasets = []

for packet_file in dependencies:
    apid_datasets = packet_file_to_datasets(packet_file, xtce)
    datasets.append(apid_datasets)

repoint_df = get_repoint_data()

row = repoint_df.loc[repoint_df.repoint_id == pointing_number]
repoint_start = row.repoint_start_time.values[0]
repoint_end = row.repoint_end_time.values[0]

dict_ds = combine_repointings(datasets, repoint_start, repoint_end)
```

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -870,14 +870,11 @@ class Ultra(ProcessInstrument):
         datasets: list[xr.Dataset] = []
 
         if self.data_level == "l1a":
-            # File path is expected output file path
-            if len(dependencies) > 1:
-                raise ValueError(
-                    f"Unexpected dependencies found for ULTRA L1A:"
-                    f"{dependencies}. Expected only one dependency."
-                )
-
-            datasets = ultra_l1a.ultra_l1a(dependencies[0], self.version)
+            data_dict = {}
+            for dependency in dependencies:
+                dataset = load_cdf(dependency)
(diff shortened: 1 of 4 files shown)


greglucas (commented):
I'm a bit confused by the naming, because I don't think we are combining pointing sets here, but rather combining raw packet files (datasets).

Two suggestions to consider as alternatives to try and narrow the purpose of the function a bit more.

* Combine all packet files together into a single larger file and read that in so you don't have to pass in multiple of these packet_file_to_dataset objects. Alternatively, write a separate `combine_packet_file_to_datasets()` function to just do that task.
* Do the filtering on a single dataset passed in by repointing number (get the start/stop based on that) and load the repointing file within this routine to determine the start/stop times.

laspsandoval:
Closing:

We will now split L0 by pointings.

PR #1567, opened 2025-04-02
ENH: Add a spacecraft CLI parser

### LINKED ISSUE #1565: Produce spacecraft data automatically
### Requirements:

We have the processing code for quaternions within `imap_processing` already, but now we need to be able to add a CLI wrapper to call this similar to the other instruments from our remote job kickoff scripts. This means adding a processing step for it and the processing levels to the lookup dictionary.

greglucas:
This adds the ability to call the spacecraft quaternion processing functions from the CLI.

I tested this locally with the following command and it worked as expected and produced both l1a and l1b files.
```
imap_cli --instrument spacecraft --data-level l1a --descriptor quaternions --start-date 20241202 --version v001 --dependency '[{"type": "science", "files": ["imap_spacecraft_l0_raw_20241202_v001.pkts"]}]'
```

closes #1565


imap_processing/tests/test_cli.py
```diff
@@ -165,6 +165,39 @@ def test_hi_l1(mock_instrument_dependencies, data_level, n_prods):
         assert mocks["mock_upload"].call_count == n_prods
 
 
+@mock.patch("imap_processing.cli.quaternions.process_quaternions")
```
tmplummer: I recommend adding `autospec=True` here. That will cause the mock function to have the same signature as the mocked function and thus the signature of the call will also be tested.
```suggestion
@mock.patch("imap_processing.cli.quaternions.process_quaternions", autospec=True)
```

tmplummer (approved):
LGTM other than the `autospec=True` comment.

imap_processing/tests/test_cli.py
greglucas: Interesting, I didn't realize that, thank you for that great suggestion!

imap_processing/cli.py
```diff
@@ -831,6 +832,46 @@ def do_processing(
...
+        """
+        print(f"Processing Spacecraft {self.data_level}")
+
+        if self.data_level != "l1a":
+            raise NotImplementedError(
+                f"Spacecraft processing not implemented for level {self.data_level}"
+            )
+
+        # File path is expected output file path
+        input_files = dependencies.get_file_paths(source="spacecraft")
```
tech3371, line 883: I think it's this that you want. Then move this line below before passing it to the code.
```suggestion
        input_files = dependencies.processing_input
```

tech3371 (approved):
one minor comment. otherwise lgtm

imap_processing/cli.py
greglucas: I'm not sure I follow what you mean. I want the file path directly and don't really care about processing input I don't think? i.e. I don't care what type of input it is for this processing, so I am just skipping the processing_input separation into science/ancillary/spice files because it is unnecessary for this instrument. I might be missing what processing_input is used for.

greglucas: Going to merge this so I can get it in before MSIM. Feel free to open an issue for follow-up on this and I'm happy to address it then.

PR #1570, opened 2025-04-03
BUG: Update write_cdf to handle parent_files

### LINKED ISSUE #1569: BUG: write_cdf fails with Parents files
This came during testing dependency updates. This issue is changing how we carry files used to produced a data file. Before, we were carry previous file's parent files but now we are only carry immediate files used to a data file.

tech3371:
# Change Summary
closes #1569 

## Overview
This came during testing dependency updates. This issue is changing how we carry files used to produced a data file. Before, we were carry previous file's parent files but now we are only carry immediate files used to a data file.

## Updated Files
   - update to resolved above

## Testing

### DIFF
--- a/imap_processing/tests/cdf/test_utils.py
+++ b/imap_processing/tests/cdf/test_utils.py
@@ -103,9 +103,20 @@ def test_parents_injection(test_dataset):
     test_dataset : xarray.Dataset
         An ``xarray`` dataset object to test with
     """
+    # Deep copy the dataset to ensure that the original is not modified
+    test_ds1 = test_dataset.copy(deep=True)
+    test_ds1.attrs["Data_version"] = "v001"
     parent_paths = ["test_parent1.cdf", "test_parent2.cdf"]
-    new_dataset = load_cdf(write_cdf(test_dataset, parent_files=parent_paths))
+    new_dataset = load_cdf(write_cdf(test_ds1, parent_files=parent_paths))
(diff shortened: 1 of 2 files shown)


imap_processing/tests/cdf/test_utils.py
```diff
@@ -103,10 +103,21 @@ def test_parents_injection(test_dataset):
     test_dataset : xarray.Dataset
         An ``xarray`` dataset object to test with
     """
+    # Deep copy the dataset to ensure that the original is not modified
```
greglucas, line 106: Do you need this deepcopy to pass the tests? I thought the fixture was per-test, so it would create a new xarray object each test we run.

I think you should be able to modify the in-memory dataset in your second version no problem as well since you've already saved/loaded the first one.

tech3371: Yeah, I need to add that to pass the tests. I tried different combinations but this one solved it.

greglucas: Hm... I just tested this locally and it worked without the deepcopy. Can you see if the commit I just pushed up works for you?

tech3371: thank you! you always have neat solutions!

greglucas (approved):

tmplummer (approved):
Great, thanks!

PR #1571, opened 2025-04-03
MNT: Add network marks to IDEX test files to avoid downloading

greglucas:
I noticed that these two tests were failing when the tests should be skipped on the 3.12 runners.
https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/14248070127/job/39933951656#step:6:1463

This adds in a network mark to avoid downloading the data and skip these tests on the other runners.


lacoak21 (approved):
Thank you for adding that Greg! I didn't think to add those.

PR #1577, opened 2025-04-04
HIT L1B - process sectored rates dataset

### LINKED ISSUE #1063: HIT L1B - write sectored rates algorithm code
Write Sectored Rates Algorithm Code

Apply equations

- livetime = livetime_counter / 270 
- count_rates = raw_counts / livetime
- Use [HIT_L1A-L1B_Table_JGM_4_1_24.xlsx](https://lasp.colorado.edu/galaxy/download/attachments/209101633/HIT_L1A-L1B_Table_JGM_4_1_24.xlsx?version=1&modificationDate=1722365110981&api=v2) and use column B to get which rates to use in the equation.

Note: sector rates consist of data from 10 science frames (10 minutes). Sum livetimes from 10 science frames and save the value to a variable (i.e. livetime_10) and use it for the sector rates for the next 10 frames. It will be staggered. The first 10 packets won't have sector rates and this is fine.

vmartinez-cu:
This PR implements the sectored rates algorithm

To calculate sectored rates, counts are divided by livetime. 

First, sectored counts are grouped into complete sets, which span 10 minutes. Sectored counts are transmitted 10 minutes after they are collected, so to calculate rates correctly, the sectored counts are divided by the sum of livetime values from the previous 10 minutes.

The algorithm handles grouping and subsetting the counts data, calculating the rates, and creating an l1b dataset with the rates data.

Closes #1063 

## Updated Files
- hit_l1b.py
   - add function to process sector rates dataset
   - add supporting functions
   - update functions that coordinate L1B science dataset creation
- hit/l1b/constants.py
   - Update livestim pulse constant to be all caps
   - Add fill values for different data types
- hit_utils.py
   - remove commented out code
- test_hit_l1b.py
   - add unit tests for new functions


imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -145,11 +146,39 @@ def process_science_data(
...
 
-    return l1b_science_datasets
+    return list(l1b_datasets.values())
+
+
+def initialize_l1b_dataset(l1a_counts_dataset: xr.Dataset, coords: list) -> xr.Dataset:
+    """
+    Initialize the L1B dataset.
+
+    Create a dataset and add coordinates and dynamic threshold state
```
lacoak21:
```suggestion
    Create a dataset and add coordinates and the dynamic threshold state data array
```

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -145,11 +146,39 @@ def process_science_data(
...
+        A list of coordinates to assign to the L1B dataset.
+
+    Returns
+    -------
+    l1b_dataset : xr.Dataset
+        An L1B dataset with coordinates and dynamic threshold state.
+    """
+    # Create a new dataset to store the L1B standard rates
+    l1b_dataset = xr.Dataset()
+    l1b_dataset = l1b_dataset.assign_coords(
```
lacoak21: I think you could also just do : 
l1b_dataset = xr.Dataset(
    coords={coord: l1a_counts_dataset.coords[coord] for coord in coords}
)

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -269,6 +287,29 @@ def calculate_rates(
...
+
+    Returns
+    -------
+    xr.DataArray
+        Livetime summed over 10-minute intervals. Values repeated for each epoch in the
+        10-minute intervals to match the original livetime array shape.
+        [5,5,5,5,5,5,5,5,5,5, 6,6,6,6,6,6,6,6,6,6, 7,7,7,7,7,7,7,7,7,7].
+    """
+    livetime_10min_sum = [
+        livetime[i : i + 10].sum().item() for i in range(0, len(livetime) - 9, 10)
```
lacoak21, line 305: Would the length of livetime ever not be a multiple of 10? If so, I think this would exclude some values from the livetime_10min_sum.

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -329,3 +359,164 @@ def process_summed_rates_data(
...
+    # Filter out start indices that are less than or equal to the bin size
+    # since the previous 10 minutes are needed
+    start_indices = start_indices[start_indices > bin_size]
+    data_slice = slice(start_indices[0], start_indices[-1] + bin_size)
+
+    # Subset data to include only complete sets of sectored counts
+    l1b_sectored_rates_dataset = l1a_counts_dataset.isel(epoch=data_slice)
+
+    # Subset livetime staggered from sectored counts by 10 minutes
+    livetime_slice = slice(start_indices[0] - bin_size, start_indices[-1])
```
lacoak21, line 405: Wouldn't you also want to subtract bin_size from start_indices[-1]?

lacoak21 (commented):
Great job Veronica! I just had a few comments/ questions.

imap_processing/hit/l1b/hit_l1b.py
nkerman: Good question which got me thinking on this/playing around with the ranges as I started a review!

**If** the desired behavior, supposing `livetime` to be `length=105`, is that you sum each group of 10, and then the last 5, I think you can just remove the `- 9`. 

You could then create `livetime_expanded` in the same way, but trim it to the length of livetime. 

---

But perhaps it's supposed to limit to multiples of 10, in which case what you have already is great!

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: Good question. This was confusing for me at first. Livetime needs to be offset from sectored counts by 10 minutes. The ending index of this livetime slice is the starting index of the final set of 10 minutes of sectored counts which results in the livetime being 10 minutes before the last set of counts

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -329,3 +359,164 @@ def process_summed_rates_data(
...
+    # TODO
+    #  -filter by epoch values in day being processed
+    #  -get middle epoch (or get mod 5 value for 6th frame)
+    #  -consider refactoring calculate_rates function to handle sectored rates
+
+    # Define particles and coordinates
+    particles = ["h", "he4", "cno", "nemgsi", "fe"]
+
+    # Extract relevant data variable names that start with a particle name
+    data_vars = [
```
nkerman, line 455: An example of this would be "he4_sectored_counts", right? Trying to keep the species-sepecific data_vars / coords separate in my head.

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -329,3 +359,164 @@ def process_summed_rates_data(
...
+        if "sectored_counts" in var:
+            counts = l1a_counts_dataset[var]
+
+            # Compute rates, skipping fill values
+            livetime_10min_aligned = livetime_10min.broadcast_like(counts)
+            rates = (
+                (counts / livetime_10min_aligned)
+                .where(counts != FILLVAL_INT64, FILLVAL_FLOAT32)
+                .astype(np.float32)
+            )
```
nkerman: This is a really nice, concise way of handling the missing values! I plan to use something like this going forward!

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -329,3 +359,164 @@ def process_summed_rates_data(
...
+
+            # Determine the new variable name
+            if "_sectored_counts_delta_" in var:
+                new_var = var.replace("sectored_counts", "stat_uncert")
```
nkerman, line 492: So this would rename "he4_sectored_counts_delta_plus" to "he4_stat_uncert_delta_plus", right? I had thought we wanted it to be something like "he4_sectored_counts_stat_uncert"?

nkerman: Ah, but I see below you're dropping the "sectored_counts" in the main variable as well. Seems fine then!

imap_processing/hit/l1b/hit_l1b.py
lacoak21: Ohh ok i understand now.

lacoak21 (approved):

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: Good idea, thanks!

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: Good question. The length of livetime will always be multiples of 10 due to how livetime is sliced before the summing operation

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: I plan to drop "delta" too based on recent nomenclature discussions. I just need to double check with HIT first since they defined the uncertainty variable names to include "delta" initially. I don't think they will mind dropping it. I will need to refactor L1A accordingly so this renaming part might change in the future

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: Correct, along with energy variables
`he4_sectored_counts, he4_energy_delta_minus, he4_energy_delta_plus`

These are the subset of data variables from the L1A dataset that are needed for the L1B dataset

nkerman (approved):

vmartinez-cu:
pre-commit.ci autofix

PR #1578, opened 2025-04-04
Ultra l2 healpix

### LINKED ISSUE #1582: Ultra L2 Main Code (Healpix output)
1. Use the ENA Map Code to take many L1C pointing sets (in a `data_dict: {}`) and project them to an HAE skymap, dealing with summed and averaged quantities correctly and accounting for changes in solid angle ($\Omega$) and exposure factor ($\tau_{exp}$).
2. Next, calculate the subsequent L2 quantities such as flux and flux_uncertainty. 
3. Apply output formatting and return a dataset. 

Note that additional PRs will be forthcoming to add necessary metadata attributes.

nkerman:
# Change Summary

The core of the Ultra L2 pipeline to produce Healpix tessellated map. 

Closes #1582 

## Overview

The core of Ultra L2 consists of these steps:

1. Use the ENA Map Code to take many L1C pointing sets (in a `data_dict: {}`) and project them to an HAE skymap, dealing with summed and averaged quantities correctly and accounting for changes in solid angle (Ω) and exposure factor ($\tau_{exp}$).
2. Next, calculate the subsequent L2 quantities such as flux and flux_uncertainty.
3. Apply output formatting and return a dataset of the Healpix tessellated map.

In order to standardize the way L2 code, L3 code, and the mapper tool will call the ENA Map code, this PR also begins to define a simple class `MapProperties` for defining properties of the output map such as the tiling type (`Rectangular/Healpix`) and resolution parameters (`nside/spacing_deg`).

Note that additional PRs will be forthcoming to add necessary metadata attributes.

## New Files

  - Ultra L2 code (with plenty of remaining TODOs)
- imap_processing/ena_maps/utils/map_properties.py
  - Define `MapProperties`` class to input/store map properties
- imap_processing/tests/ena_maps/data/test_ultra_l2_map_config.json
  - JSON input for testing `MapProperties` class

## Updated Files

- imap_processing/ena_maps/utils/coordinates.py
  - Change energy coord name
  
  - Change energy coord name
  - Add solid angle attr to `UltraPointingSet`
  - Added a TODO to note where change is likely to allow for mean-ing within PSET pixels rather than over all PSETs. 

## Testing

- Tests for Ultra L2 and MapProperties class


imap_processing/ena_maps/utils/map_properties.py
```diff
@@ -0,0 +1,108 @@
+"""Simple class to hold ENA Sky Map configuration properties."""
+
+import json
+from pathlib import Path
+
+from imap_processing.ena_maps import ena_maps
+from imap_processing.spice import geometry
+
+
+class MapProperties:
```
greglucas: I'm not sure I totally follow everywhere this is used, but my initial reading is that this is only used to load/store configurations of SkyMaps. Can this live on the SkyMap itself? This seems like it is essentially trying to serialize/deserialize a SkyMap's properties to me, so something like the following with opinionated properties might be useful there rather than making a separate object to hold these things that is kind of transient in a way as it is just created and then disposed of once we create a SkyMap from the configurations.

```
class SkyMap:
    def to_json(self):
        # Define your explicit output format
    @classmethod
    def from_json(cls, s):
        # Load from a configuration file
```

imap_processing/ena_maps/utils/map_properties.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Parameters
+    ----------
+    properties : dict | str | Path
+        A dictionary or a path to a JSON file containing the map properties.
+        The dictionary should contain the following keys:
+            - sky_tiling_type: str
+                Type of sky tiling (HEALPIX or RECTANGULAR).
+            - spice_reference_frame: int | str
+                Reference frame for the map (e.g., ECLIPJ2000).
```
greglucas: Do we need to support both? I think you can be opinionated here and dictate which one should be our standard.

imap_processing/ena_maps/utils/map_properties.py
```diff
@@ -0,0 +1,108 @@
...
+            - projection_method_and_values: dict
+                Dictionary with keys "PUSH" and/or "PULL" containing lists of values.
```
greglucas: Values is a bit vague here, maybe break this into two separate lists? `push_values`, `pull_values`. Or just have a single list indicating `pull_values` and everything else is pushed by default (or the reverse whichever is easier/better default), so a user doesn't have to enter every single value multiple times.

imap_processing/tests/ultra/data/mock_data.py
```diff
@@ -277,8 +282,15 @@ def mock_l1c_pset_product_healpix(  # noqa: PLR0913
...
-    exposure_time = peak_exposure * (prob_scaling_factor / prob_scaling_factor.max())
+    # Generate exposure times using gaussian distribution, but wider
+    prob_scaling_factor_exptime = counts_scaling_params[1] * np.exp(
+        -(lat_diff**2) / (2 * (3 * width_scale) ** 2)
+    )
+    exposure_time = peak_exposure * (
+        prob_scaling_factor_exptime / prob_scaling_factor_exptime.max()
+    )
+    # Add a tiny exposure time to avoid nans
+    exposure_time += 1e-10
```
greglucas: Won't we run into this where we didn't view any part of the sky so the exposure time is 0 in certain spots? Do we need to handle this some other way more fundamentally so we don't get into this situation in the first place? Something like only do the calculation where exposure_time != 0, or live with NaNs and do a `with np.errstate()` ignore on calculations you still want to propagate but ignore.

imap_processing/tests/ultra/unit/test_ultra_l2.py
```diff
@@ -0,0 +1,278 @@
...
+        )
+
+        np.testing.assert_allclose(
+            ultra_pset_from_dataset.data["counts"].values,
+            ultra_pset_from_path.data["counts"].values,
+            rtol=1e-6,
+        )
+
+        # delete cdf_filepath once we're done with it
+        Path(cdf_filepath).unlink()
```
greglucas: This should be cleaned up automatically since we're using tmpdirs for the writing I thought. Generally speaking we shouldn't be relying on unlinking in the tests because an error between the creation / deletion will mean this code-path doesn't get reached and we have an extra file laying around somewhere.

I think you can remove this and let the tmpdir (per test) clean up for you as it is managed externally.
```suggestion
```

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+)
+from imap_processing.spice import geometry
+
+logger = logging.getLogger(__name__)
+logger.info("Importing ultra_l2 module")
+
+# Set some default values for the map properties
+PSET_SPICE_FRAME = geometry.SpiceFrame.IMAP_DPS
+DEFAULT_L2_HEALPIX_NSIDE = 32
+DEFAULT_L2_HEALPIX_NESTED = False
```
greglucas, line 38: Aren't these defined within the constant below already?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+        If input_data is neither an xarray Dataset nor a path to a CDF file.
+    KeyError
+        If any of the required variables are missing from the input data.
+    """
+    # Allow for passing in EITHER xarray Datasets (preferable for testing)
+    if isinstance(input_data, xr.Dataset):
+        if not inplace:
+            input_data = input_data.copy(deep=True)
+        ultra_pointing_set = ena_maps.UltraPointingSet(
+            l1c_dataset=input_data, spice_reference_frame=PSET_SPICE_FRAME
```
greglucas: Can `UltraPointingSet()` default to PSET_SPICE_FRAME so you don't have to pass it in explicitly in all of these paths here?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+    ------
+    ValueError
+        If input_data is neither an xarray Dataset nor a path to a CDF file.
+    KeyError
+        If any of the required variables are missing from the input data.
+    """
+    # Allow for passing in EITHER xarray Datasets (preferable for testing)
+    if isinstance(input_data, xr.Dataset):
+        if not inplace:
+            input_data = input_data.copy(deep=True)
```
greglucas: There are a lot of options here again. I think if someone passes in a dataset they can assume it will be shared memory under the hood. Or you could always make a deepcopy if you prefer, but don't bother with giving an option unless it is absolutely necessary.

In general, this seems like a classmethod to me again. `UltraPointingSet.from_dataset()` and `UltraPointingSet.from_cdf()` or similar and they could be separate methods to make it explicit as well.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+        ultra_pointing_set = ena_maps.UltraPointingSet(
+            l1c_dataset=input_data, spice_reference_frame=PSET_SPICE_FRAME
+        )
+    # OR paths to CDF files (preferable for projecting many PointingSets)
+    elif isinstance(input_data, str | Path):
+        if isinstance(input_data, str):
+            input_data = Path(input_data)
+        ultra_pointing_set = ena_maps.UltraPointingSet(
+            l1c_dataset=load_cdf(input_data), spice_reference_frame=PSET_SPICE_FRAME
```
greglucas: Do you need to get the reference frame from the cdf file?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+
+def generate_ultra_healpix_skymap(
+    ultra_l1c_psets: list[str | xr.Dataset],
+    output_map_properties: MapProperties = DEFAULT_L2_MAP_PROPERTIES,
+) -> ena_maps.HealpixSkyMap:
+    """
+    Generate a Healpix skymap from ULTRA L1C pointing sets.
+
+    This function combines IMAP Ultra L1C pointing sets into a single L2 HealpixSkyMap.
+    It handles the projection of values from pointing sets to the map,applies necessary
```
greglucas:
```suggestion
    It handles the projection of values from pointing sets to the map, applies necessary
```

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+        [
+            "observation_time",
+            "pointing_set_exposure_times_solid_angle",
+            "num_pointing_set_pixel_members",
+        ]
+    )
+
+    for ultra_l1c_pset in ultra_l1c_psets:
+        pointing_set = read_into_pointing_set(ultra_l1c_pset)
+        logger.info(f"PSET epoch: {pointing_set.epoch}")
```
greglucas: Nice logging! Maybe add some more info / quantities about what you're projecting from this pset too?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+        # Initial processing for weighted quantities at PSET level
+        # Weight the values by exposure and solid angle
+        # (in that order to avoid double weighting by solid angle)
+        for (
+            quantity_to_weight
+        ) in VARIABLES_TO_WEIGHT_BY_POINTING_SET_EXPOSURE_TIMES_SOLID_ANGLE:
+            pointing_set.data[quantity_to_weight] = (
+                pointing_set.data[quantity_to_weight]
+                * pointing_set.data["pointing_set_exposure_times_solid_angle"]
+            )
```
greglucas: I think you could use `*=` here?

Or alternatively could you do this as a one-liner instead of a for-loop by passing in the list of variables?

`pointing_set[VARIABLES_TO_WEIGHT_BY_POINTING_SET_EXPOSURE_TIMES_SOLID_ANGLE] *= pointing_set["pointing_set_exposure_times_solid_angle"]`

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+    # Subsequent processing for weighted quantities at SkyMap level
+    for (
+        quantity_to_weight
+    ) in VARIABLES_TO_WEIGHT_BY_POINTING_SET_EXPOSURE_TIMES_SOLID_ANGLE:
+        skymap.data_1d[quantity_to_weight] = (
+            skymap.data_1d[quantity_to_weight]
+            / skymap.data_1d["pointing_set_exposure_times_solid_angle"]
+        )
+
+    # TODO: Ask Ultra team about this - I think this is a decent
```
greglucas, line 180: This will potentially be incorrect if one pointing lasted 3 days and another pointing only lasted a half-day or things like that where the view-time per pixel would be wildly different between the pointing sets I think. Do we need to take into account time here as well?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+        skymap.data_1d["exposure_factor"]
+        * skymap.data_1d["sensitivity"]
+        * skymap.solid_angle
+        * delta_energy
+    )
+
+    # Drop the variables that are no longer needed
+    skymap.data_1d = skymap.data_1d.drop_vars(
+        VARIABLES_TO_DROP_AFTER_FLUX_CALCULATION,
+        errors="ignore",
```
greglucas: What errors are being ignored? Do we need to log anything here so we aren't missing something, or add a comment indicating why this is used.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+            "Data_version": data_version,
+        }
+
+    # TODO: Implement conversion to Rectangular map
+    elif output_map_properties.sky_tiling_type is ena_maps.SkyTilingType.RECTANGULAR:
+        map_attrs = {
+            "Spacing_degrees": output_map_properties.spacing_deg,
+            "Data_version": data_version,
+        }
+        pass
```
greglucas: Raise so we don't think this is working?
```suggestion
        raise NotImplementedError
```

greglucas (commented):
Mostly looking at the overall architecture of this and not the implementation/math details.

My overall comment/question is around whether you need the intermediate classes or if you can add class methods to serialize/deserialize to your json map definition instead.

Looks like this is very close to working all the way through, very nice!

imap_processing/tests/ultra/unit/test_ultra_l2.py
```diff
@@ -0,0 +1,278 @@
...
+from imap_processing.ena_maps.utils.map_properties import (
+    MapProperties,
+)
+from imap_processing.tests.ultra.data.mock_data import mock_l1c_pset_product_healpix
+from imap_processing.ultra.l2 import ultra_l2
+
+
+class TestUltraL2:
+    @pytest.fixture()
+    def _setup_spice_kernels(self, spice_test_data_path):
```
tmplummer: Although it is not obvious where to find it, there is a common fixture for furnishing a list of kernels: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/conftest.py#L284

imap_processing/ena_maps/utils/coordinates.py
```diff
@@ -9,7 +9,7 @@ class CoordNames(Enum):
     GENERIC_PIXEL = "pixel"
 
     TIME = "epoch"
-    ENERGY = "energy"
+    ENERGY = "energy_bin_geometric_mean"
```
tmplummer: Is this an agreed upon coordinate name? I think that it is specific to Ultra. So far I have only heard Hi refer to central energy.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,334 @@
...
+            skymap.data_1d[quantity_to_weight]
+            / skymap.data_1d["pointing_set_exposure_times_solid_angle"]
+        )
+
+    # TODO: Ask Ultra team about this - I think this is a decent
+    # but imperfect approximation for meaning the exposure:
+    # (dividing by the 1/(number of PSETs)) to fix the exposure being mean-ed over
+    # all pixels in all PSETs which feed into a map superpixel,
+    # rather than being mean-ed over pixels in a PSET and summed over PSETs
+    skymap.data_1d["exposure_factor"] /= skymap.data_1d[
```
tmplummer, line 185: I'm surprised that this is being averaged at all. From the CATDESC, exposure_factor is `Exact or approximate exposure time over which counts in a pixel are accumulated`.

tmplummer (commented):
Just a few comments after Greg's very thorough review.

imap_processing/ena_maps/utils/coordinates.py
nkerman: Ah got it - I can either make this enum value `ENERGY_ULTRA` now, or split it when others come up. I'd lean towards the former - sound good?

imap_processing/ultra/l2/ultra_l2.py
nkerman: The idea here is that, if you're projecting multiple (say 4) adjacent pixels from the same PSET into a map superpixel, you don't want to sum their exposure times, since the map superpixel doesn't get 4x the exposure time. It just gets that exposure time over a larger solid angle/area. 

I think the _right_ way to do this that isn't yet implemented in the ENA map code would be to mean this value over PSET pixels but sum it over multiple PSETs. See "Concern related to projecting exposure time" [toward the bottom of this Diagram](https://lasp.colorado.edu/galaxy/x/9IhQDw). This current approach approximates that, **but** it doesn't correctly deal with cases where map superpixels get different numbers of PSET pixels - which is a small effect but does really happen often. 

Unfortunately the person I most need to run this by is out for the next week. Nick's initial response was we might just need to change our PSETs and maps to both be `nside=64`. I think we can fix this though, if we allow the ENA map projection to mean before summing on some variables. For now, I think this is the best solution

imap_processing/tests/ultra/unit/test_ultra_l2.py
nkerman: Thanks! I totally forgot about that. I made the switch, but out of curiosity, is it bad to define a fixture like this? For testing, I think it's a bit cumbersome to do the:

```
kernels = [a,b,c]
with furnish_kernels(kernels):
    # DO PROCESSING
```

as opposed to just having a single fixture to furnish them for the whole test.

imap_processing/ena_maps/utils/map_properties.py
nkerman: I like the idea of this being a `to_json`/`from_json` method on the `AbstractSkyMap`, but I'm not entirely sure at a glance that it will offer the same structure I was going for with this class. 

The reason I added this class was to give a consistent structure for me/ Menlo developers/ the mapping tool, etc. to write their JSON definitions to conform to. There's probably other ways to do this, but I figured a MapProperties class with a `1-to-1` correspondence with the JSON was the most legible.

imap_processing/ena_maps/utils/map_properties.py
nkerman: Will do - making it only `str`

imap_processing/ena_maps/utils/coordinates.py
tmplummer: I think that the subtext is that Ultra wants a coordinate named "energy_bin_geometric_mean" in their PSET products? I would think that just naming it "energy" and using the CATDESC to let users know that it is a geometric mean of the bin along with what the DELTA_PLUS and DELTA_MINUS would be more consistent. But this discussion is pretty much in the weeds as related to this PR. I think that your suggestion makes sense.

imap_processing/tests/ultra/unit/test_ultra_l2.py
tmplummer: Nothing wrong with wrapping the shared `furnish_kernels` fixture in your own `_setup_spice_kernels`. The main thing is that there is a little bit of code duplication by not using `furnish_kernels` and your implementation doesn't clean up after itself. That second point is mitigated by the fact that I anticipated this and implemented a fixture that automatically cleans up (clears) furnished kernels after each test.

imap_processing/ultra/l2/ultra_l2.py
tmplummer: Ah, ok. Thanks for explaining that.

imap_processing/tests/ultra/unit/test_ultra_l2.py
nkerman: This file doesn't get deleted on my local machine as is, but I think it will in production. 

---
I think if I add the `tmp_path` fixture, it will get deleted though.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Actually they're not guaranteed to be. Eventually (soon!), the `DEFAULT_ULTRA_L2_MAP_PROPERTIES` will be a Rectangular tiling, without these core Healpix properties. In that case, we still need to know the default L2 healpix properties, since it will make a healpix map first, **then** convert to rectangular.

imap_processing/ultra/l2/ultra_l2.py
nkerman: `UltraPointingSet` already defaults to IMAP_DPS, so I can drop this!

imap_processing/ultra/l2/ultra_l2.py
nkerman: Got it - I think it's better to always make a copy, since making a map should never change your original input PSETs. 

Interesting point on the class methods - `UltraPointingSet` is already instantiated from an xarray Dataset, so we only really need to define the `from_cdf()` one. I can definitely extract that out to a class method, but I think there needs to be some wrapper method like this one, that will read in either a path or a `Dataset` to an `UltraPointingSet`, so that we can allow the main `ultra_l2` methods to read in either type. I think these methods should be able to work in the exact same way either with paths when handling 100s of pointings or datasets when handling a few pointings in tests.

imap_processing/ultra/l2/ultra_l2.py
nkerman: It will always be `IMAP_DPS` - see [thread above](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1578#discussion_r2033940079). 

I don't think Laura has added this information at L1 yet, but eventually the `UltraPointingSet` `init` could check the attribute to get the frame. For now, `IMAP_DPS` is a safe assumption!

imap_processing/ultra/l2/ultra_l2.py
nkerman: Sometimes in manual testing I've have weird results with `*=` and `/=` on xarray objects, but I think you're right that it works fine here. I didn't realize you could use the latter syntax, but that's great! I'm switching to that.

imap_processing/ultra/l2/ultra_l2.py
nkerman: I think you're right. I'll note this as well and bring it up ASAP with Ultra, though I'm not sure progress will happen until Nick is back from vacation. I think that adding the "*mean-over-a-PSET-but-sum-over-different-PSETs*" functionality I've been thinking about to the ena_maps code will fix this as well but I need some more info from the Ultra team.

imap_processing/ultra/l2/ultra_l2.py
nkerman: My bad - this is not necessary anymore after a change I made!

imap_processing/ultra/l2/ultra_l2.py
nkerman: Yep!

imap_processing/ena_maps/utils/map_properties.py
nkerman: Done - I made it so that there's 2 keys: `values_to_push_project`, `values_to_pull_project`, which default to `[]`.

I think it's not a great idea to push/pull everything by default at this level. The codes which call the ena map code can be responsible for adding values which must always be pushed - like how the Ultra L2 code currently adds the `REQUIRED_L1C_VARIABLES` to the values to push project.

imap_processing/tests/ultra/data/mock_data.py
nkerman: Yes, I think it's probably most correct to live with NaNs (You don't actually know the flux at all if you had zero exposure time), but I can ask Ultra what they want. I think I will make the flux calculation something like this, which I saw in Veronica's code, and thought was a nice method 

```python
###### EXAMPLE from Veronica's code
            rates = (
                 (counts / livetime_10min_aligned)
                 .where(counts != FILLVAL_INT64, FILLVAL_FLOAT32)
                 .astype(np.float32)
             )

###### In this case:
            flux = (
                 (counts / exptime and other factors)
                 .where(exposure_factor != 0, np.nan)
                 .astype(np.float32)
             )

```

I had added this offset for testing and my visualizations, but I should drop it so that we can properly see what happens when there's nans.

greglucas: As an alternative that I often use when I'm ok with getting nan/inf values I just ignore it as I find it more readable:
```python
            with np.errstate(divide='ignore'):
                flux = counts.astype(np.float32) / exptime
```

imap_processing/ena_maps/utils/map_properties.py
greglucas: You would define what the to/from json is. i.e. you write your own protocol. Here you are just encapsulating that protocol in a class, but the same logic could live in the class methods.

imap_processing/ena_maps/utils/map_properties.py
greglucas: Sounds good!

imap_processing/tests/ultra/unit/test_ultra_l2.py
greglucas: This is somewhat concerning, where is the file located? It should be produced in the tmpdir with `write_cdf()` because we are already setting the tmpdir for every single test (to avoid situations like this with extra files hanging around).

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/185826652987fb8ad6b384430b2ac1b139423066/imap_processing/tests/conftest.py#L26

imap_processing/ultra/l2/ultra_l2.py
greglucas: :+1: `from_path_or_dataset()` classmethod then?

imap_processing/tests/ultra/unit/test_ultra_l2.py
nkerman: I was seeing files in `/private/var/folders/l2/210x3x1s0lxbryhpc1qrfykm00bs6r/T/pytest-of-nake7532/pytest-21/test_read_into_pointing_set0/imap/ultra/l1c/2025/05`. However, I just removed `tmp_path`, and it seems to still be deleting the file. I wonder if it managed to persist from a time I was debugging a test, and then stopped the debugger before it could clean up.

imap_processing/ultra/l2/ultra_l2.py
greglucas: Interesting, that is a nice document / diagram you linked to as well that explains the issue clearly!

You might need an area-weighted exposure time to project rather than just "seconds", so "seconds * solidangle" and then in the final projection you would divide by the solid angle of your super-pixel which is large as you say. I don't think you'd want to do a mean before projecting, but I haven't thought a lot about this either. I'm thinking about this like how you weight a quantity by integral(x * dA) / integral(dA), where here you'd be projecting the numerator "time * dA"

imap_processing/tests/ultra/unit/test_ultra_l2.py
greglucas: Ahh, as long as it isn't in the `imap_processing/tests` common area then I'm not concerned. That `/private/var` area is tmp storage on the system and may or may not be cleared out as the system needs it, so I don't think you're necessarily guaranteed for it to clean up, but likely will be on next reboot or something as well.

nkerman: Thanks!

imap_processing/ultra/l2/ultra_l2.py
nkerman: I spoke with the Ultra Team and explained the issue I encountered here, and we decided that this solution will be good for now. We'll have more discussions in the next 2 weeks, and I am pretty sure I'll be able to get the changes we decide on merged before end of May.

nkerman: I'll make a ticket for now.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Hi Greg - I agree that it is in principle a solid angle weighted quantity. 
Note that, because PSET pixels all have the same solid angle, this reduces to an unweighted mean. 
Note also, that, after projecting, one can either divide by the sum of the weights (sum of PSET pixel solid angles) **OR** divide by the solid angle of the map pixel. These must be equal, but only averaged across all map pixels. It is not equal across individual pixels (see the 2 plots with real data at the bottom of that diagram, showing that some map pixels get the mean number of 64 PSET pixels, while others get 63, 65, etc.)

In practice, I don't think (and I've sketched this out with some simple examples like those in the diagram) that we can just solid angle weight the exposure times before projection and divide by the solid angle of the map pixel after, mostly because we get uneven numbers of PSET pixels going into a map.

imap_processing/ena_maps/utils/coordinates.py
nkerman: > I think that the subtext is that Ultra wants a coordinate named "energy_bin_geometric_mean" in their PSET products? I would think that just naming it "energy" and using the CATDESC to let users know that it is a geometric mean of the bin along with what the DELTA_PLUS and DELTA_MINUS would be more consistent.

This makes sense to me, but I don't see anything wrong with a more specific dimension name either - I think that is what the Ultra team wants. 

> But this discussion is pretty much in the weeds as related to this PR. I think that your suggestion makes sense.

Sounds good - using `CoordNames.ENERGY_ULTRA` for now

imap_processing/tests/ultra/unit/test_ultra_l2.py
nkerman: Got it! Yeah I think I was trying to actively avoid cleaning up properly here, since I find a bunch of repeated `with` wrappings mildly distracting. 

But overall, I don't think this is much of an issue, and I'mhappy using `furnish_kernels` along with the `with` statements to clear old kernels.

imap_processing/ultra/l2/ultra_l2.py
nkerman:
```
        logger.info(
            f"Projecting a PointingSet with {pointing_set.num_points} pixels "
            f"at epoch:{pointing_set.epoch}\n"
            f"These values will be projected: {value_keys_to_push_project}"
        )
```

nkerman: Add the above, thanks!

imap_processing/ultra/l2/ultra_l2.py
nkerman: >from_path_or_dataset() classmethod then?

Done! Moved this code to be a method: `UltraPointingSet.from_path_or_dataset()`

imap_processing/tests/ultra/data/mock_data.py
nkerman: That's preferable, thanks!

imap_processing/ena_maps/utils/map_properties.py
nkerman: Thanks Greg, I just finished with this.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -685,8 +736,118 @@ def project_pset_values_to_map(
...
+
+    @classmethod
+    def from_dict(cls, properties: dict) -> RectangularSkyMap | HealpixSkyMap:
+        """
+        Create a SkyMap object from a dictionary of properties.
+
+        Parameters
+        ----------
+        properties : dict
+            Dictionary containing the map properties.
```
greglucas: Do you have the specific properties that are required / needed listed somewhere? This might be a good place to do that if not.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -685,8 +736,118 @@ def project_pset_values_to_map(
...
+        -------
+        RectangularSkyMap | HealpixSkyMap
+            An instance of a SkyMap object with the specified properties.
+
+        Raises
+        ------
+        ValueError
+            If the sky tiling type is not recognized.
+        """
+        sky_tiling_type = SkyTilingType[properties["sky_tiling_type"]]
```
greglucas: Do you want this to be the value instead of the key lookup?
```suggestion
        sky_tiling_type = SkyTilingType(properties["sky_tiling_type"])
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -685,8 +736,118 @@ def project_pset_values_to_map(
...
+        elif sky_tiling_type is SkyTilingType.RECTANGULAR:
+            skymap = RectangularSkyMap(
+                spacing_deg=properties["spacing_deg"],
+                spice_frame=spice_reference_frame,
+            )
+        else:
+            raise ValueError(f"Unknown sky tiling type: {sky_tiling_type}")
+
+        # Store requested variables to push/pull, which will be done by the instrument
+        # code which creates and uses the SkyMap object.
```
greglucas, line 844: Should these be added to the initializer instead of updating properties here?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -685,8 +736,118 @@ def project_pset_values_to_map(
...
+                "nested": self.nested,
+            }
+        elif isinstance(self, RectangularSkyMap):
+            map_properties_dict = {
+                "sky_tiling_type": "RECTANGULAR",
+                "spice_reference_frame": self.spice_reference_frame.name,
+                "spacing_deg": self.spacing_deg,
+            }
+        else:
+            raise ValueError("Unknown SkyMap type.")
```
greglucas: Add the potential members you know about to be more helpful?
```suggestion
            raise ValueError(f"Unknown SkyMap type, must be one of {SkyTilingType.members}")
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -685,8 +736,118 @@ def project_pset_values_to_map(
...
+
+        if self.values_to_push_project:
+            map_properties_dict["values_to_push_project"] = self.values_to_push_project
+        if self.values_to_pull_project:
+            map_properties_dict["values_to_pull_project"] = self.values_to_pull_project
```
greglucas: Do you want to add this to the json output regardless so that it is an empty list in the format and consistent no matter what is present?

imap_processing/tests/ena_maps/data/test_ultra_l2_map_config.json
```diff
@@ -0,0 +1,13 @@
+{
+    "sky_tiling_type": "HEALPIX",
+    "spice_reference_frame": "ECLIPJ2000",
+    "values_to_push_projection": [
```
greglucas: I think you use "project" everywhere above.
```suggestion
    "values_to_push_project": [
```

imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -836,3 +895,132 @@ def test_match_coords_to_indices_map_to_map_no_et_error(self):
...
+        # Expect there to be a AttributeError when accessing a non-existent key
+        with pytest.raises(AttributeError):
+            _ = skymap_from_json.spacing_deg
+
+        # Check that the json output is the same as the original input
+        with open(temp_file_path_input) as f:
+            original_json = json.load(f)
+        with open(temp_file_path_output) as f:
+            output_json = json.load(f)
+        assert original_json == output_json
```
greglucas: You can combine this into a single context manager line if you'd like as well, but what you have is fine too.
```suggestion
        with open(temp_file_path_input) as f_in, open(temp_file_path_output) as f_out:
            assert json.load(f_in) == json.load(f_out)
```

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,303 @@
...
+        # so this quantity is projected to map pixels for use in weighted averaging
+        pointing_set.data["pointing_set_exposure_times_solid_angle"] = (
+            pointing_set.data["exposure_factor"] * pointing_set.solid_angle
+        )
+
+        # Initial processing for weighted quantities at PSET level
+        # Weight the values by exposure and solid angle
+        pointing_set.data[
+            VARIABLES_TO_WEIGHT_BY_POINTING_SET_EXPOSURE_TIMES_SOLID_ANGLE
+        ] *= pointing_set.data["pointing_set_exposure_times_solid_angle"]
```
greglucas, line 167: Nice! I'm glad this worked out because I find this to be simple and readable.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -0,0 +1,303 @@
...
+            index_match_method=ena_maps.IndexMatchMethod.PUSH,
+        )
+
+    # Subsequent processing for weighted quantities at SkyMap level
+    for (
+        quantity_to_weight
+    ) in VARIABLES_TO_WEIGHT_BY_POINTING_SET_EXPOSURE_TIMES_SOLID_ANGLE:
+        skymap.data_1d[quantity_to_weight] = (
+            skymap.data_1d[quantity_to_weight]
+            / skymap.data_1d["pointing_set_exposure_times_solid_angle"]
```
greglucas: Does this one not work like the above?

```python
skymap.data_1d.data[VARIABLES_TO_WEIGHT_BY_POINTING_SET_EXPOSURE_TIMES_SOLID_ANGLE] /= skymap.data_1d.data["pointing_set_exposure_times_solid_angle"]
```

greglucas (approved):
Looks good to me. I just have some minor formatting suggestions to take into account, but I like the overall structure and the to/from methods now.

imap_processing/ena_maps/ena_maps.py
nkerman: Good idea! I added a bunch of documentation, including an example dict

imap_processing/ena_maps/ena_maps.py
nkerman: I think it makes more sense for it to be the key lookup here, but I'll add a .upper() to ensure that, eg "HEALpix" would work.

imap_processing/ena_maps/ena_maps.py
nkerman: Ah, I realize that they're not necessarily being initialized by the subclasses of AbstractSkyMap. I'll do so. 

I think they should be initialized as empty lists, and updated in the from_dict method.

imap_processing/ena_maps/ena_maps.py
nkerman: Done!

imap_processing/ena_maps/ena_maps.py
nkerman: I'm kind of ambivalent here - I'd rather these keys be fully optional in the dicts, since the SkyMaps aren't self-projecting. It's the instrument code which needs to interact with these keys and tell the map to project them from the PSETS. I'll add it!

imap_processing/tests/ena_maps/data/test_ultra_l2_map_config.json
nkerman: Ah - I think I missed the need to `git rm` this. It's no longer used!

imap_processing/tests/ena_maps/test_ena_maps.py
nkerman: This is cleaner as you have it, so I'll use this in the future! Turns out I need to do a few _more_ things with the dicts loaded from json, so I think it's neater this way. 

```
# Check that the json output is the same as the original input ONLY
        # after adding automatically added
        # "values_to_push_project"/"values_to_pull_project" key to the original dict
        with open(temp_file_path_input) as f:
            original_json = json.load(f)
        with open(temp_file_path_output) as f:
            output_json = json.load(f)
        # The output json will have added an empty list for values_to_pull_project
        assert original_json != output_json
        # add the values_to_pull_project key to the original json
        original_json["values_to_pull_project"] = []
        assert original_json == output_json

```

imap_processing/ultra/l2/ultra_l2.py
nkerman: Yeah, it's great! Thank you for suggesting it!

imap_processing/ultra/l2/ultra_l2.py
nkerman: Yes! I think this works here. Thanks for catching it.

PR #1581, opened 2025-04-07
Lo L1B - Fixed CLI dependency loop

sdhoyt:
# Change Summary

## Overview
Fixed the CLI dependency loop for Lo L1B. Needed to add `.processing_input` to fix this error:

`TypeError: 'ProcessingInputCollection' object is not iterable`

## Testing
Ran CLI and was able to produce an L1B DE CDF


greglucas (approved):

tech3371 (approved):
Thank you!

PR #1588, opened 2025-04-09
Lo L1B - Set Direct Event Epoch Field

### LINKED ISSUE #1512: Lo - Set the L1B DE Epoch Time
An absolute event time must be calculated by determining which spin each DE was measured in, then adding the start time for that spin to the direct event time by first converting the de_time from DN to EU using:

 de_time / 4096 * average_spin_duration

Once this absolute event time is calculated, this can be converted to an Epoch time.

sdhoyt:
# Change Summary

## Overview
The Lo L1B Direct Event Epoch field is set by doing the following:
   - find the start time of each direct event in EU (relative to the start of the spin)
   - find the spin that each direct event was measured in
   - take the start time for the spin and add that to the start time of the direct event 

## Updated Files
imap_processing/lo/l1b/lo_l1b.py
   - added function to get the spin start times
   - added function to get the met for each event
   - added function to convert met to epoch in the dataset

## Testing
Added three new unit tests and ran in the CLI

Closes #1512


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -260,6 +266,127 @@ def set_spin_cycle(l1a_de: xr.Dataset, l1b_de: xr.Dataset) -> xr.Dataset:
...
+    # Find the closest stop_acq for each shcoarse
+    closest_stop_acq_indices = np.abs(met[:, None] - acq_end.values).argmin(axis=1)
+    # There are 28 spins per epoch (1 aggregated science cycle)
+    # Set the spin_cycle_num to the spin number relative to the
+    # start of the ASC
+    spin_cycle_num = l1b_de["spin_cycle"] % 28
+    # Get the seconds portion of the start time for each spin
+    start_sec_spins = spin_data["start_sec_spin"].values[
+        closest_stop_acq_indices, spin_cycle_num
+    ]
```
greglucas, line 305: Can you stay within xarray and subset based on your dimensions something like this? I find this a bit more explicit what dimensions you're using and subsetting on.
```suggestion
    spin_subset = spin_data.isel(epoch=closest_stop_acq_indices, spin_number=spin_cycle_num)
    spin_start_time = spin_subset["start_sec_spin"] + spin_subset["start_subsec_spin"] * 1e-6
```

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -260,6 +266,127 @@ def set_spin_cycle(l1a_de: xr.Dataset, l1b_de: xr.Dataset) -> xr.Dataset:
...
+    l1a_de: xr.Dataset,
+    l1b_de: xr.Dataset,
+    spin_start_time: xr.DataArray,
+    avg_spin_durations: xr.DataArray,
+) -> xr.Dataset:
+    """
+    Get the event MET for each direct event.
+
+    Each direct event is converted from a data number to engineering unit in seconds.
+    de_eu_time de_dn_time / 4096 * avg_spin_duration
```
greglucas: Great docstrings, makes it very clear exactly what is going on.

```suggestion
    de_eu_time = de_dn_time / 4096 * avg_spin_duration
```

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -260,6 +266,127 @@ def set_spin_cycle(l1a_de: xr.Dataset, l1b_de: xr.Dataset) -> xr.Dataset:
...
+
+    Returns
+    -------
+    l1b_de : xr.Dataset
+        The L1B DE dataset with the event MET.
+    """
+    counts = l1a_de["de_count"].values
+    de_time_asc_groups = np.split(l1a_de["de_time"].values, np.cumsum(counts)[:-1])
+    de_times_eu = []
+    for i, de_time_asc in enumerate(de_time_asc_groups):
```
greglucas, line 356: I thought this looked familiar, because you are also doing this similar logic in `set_spin_cycle`. I'm wondering if it would make sense to try and combine this logic / consolidate in some way? It looks like there you are using `esa_step` as your ASC group instead of `de_time` here, so maybe it wouldn't be easy to do that though. It just struck me as similar bookkeeping.

A thought would be that rather than keeping these all as separate datasets, could you combine them into something like a "asc_datasets" that has the various dimensions and variables from the other products in it so you could align things across each one in that sense and pass in a single dataset here where you could iterate over ASC groups once and add multiple variables within that ASC loop.

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -241,3 +245,110 @@ def test_spin_cycle():
...
+            "start_subsec_spin": (
+                ["epoch", "spin"],
+                [[2000, 3000, 4000, 5000, 6000], [1000, 1500, 2000, 3000, 4000]],
+            ),
+        }
+    )
+
+    end_acq = xr.DataArray([0, 1], dims="epoch")
+    spin_start_times_expected = np.array([20.002, 50.0015, 55.002, 60.003, 65.004])
+    spin_start_times = get_spin_start_times(l1a_de, l1b_de, spin, end_acq)
```
greglucas, line 288: Really nice tests!

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -241,3 +245,110 @@ def test_spin_cycle():
...
+
+    np.testing.assert_allclose(
+        spin_start_times,
+        spin_start_times_expected,
+        atol=1e-4,
+        err_msg=f"Spin start times: {spin_start_times} vs\
+         expected spin start times {spin_start_times_expected}",
```
greglucas: This should automatically print these out for you with the numpy testing assertions and highlight the exact entries that were different so do you want to override that?

greglucas (approved):
Minor suggestions/thoughts, but overall this looks great again. I really like that you are asserting against individual implementations of these functions and not combining too many things together into an integration test at the end.

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: I might not be understanding how to use `isel`, but it looks like that returns a 2D array filtered using the epoch and spin indices used in the function. The indexing I'm using returns a 1D array selecting a single spin from the epoch for that DE.

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Thanks!

imap_processing/lo/l1b/lo_l1b.py
greglucas: I see, I missed that. And now I just read this excellent tutorial explaining the differences between numpy and xarray indexing defaults and how to do it with xarray if you do want that behavior from numpy by default too by creating a new dimension to index along: https://tutorial.xarray.dev/intermediate/indexing/advanced-indexing.html

What you have is good here 👍

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Yeah, I'd still like to implement what you suggested before with `isel` and have a ticket written to do that later. I'll add this line to that ticket and see if there's a way to combine this.

imap_processing/tests/lo/test_lo_l1b.py
sdhoyt: Thanks!

imap_processing/tests/lo/test_lo_l1b.py
sdhoyt: Yeah, no reason to override that. I'll remove it

PR #1592, opened 2025-04-09
MNT: Bring parent file handling into the post-processing step

greglucas:
# Change Summary

## Overview

These variables shouldn't just be injected upon writing the cdf, but rather anytime we create the datasets, so we should put it in post-processing instead. Granted, we do call `write_cdf()` after this, but it seems better to keep things contained within the processing steps rather than pass them through.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -19,7 +19,6 @@ from pathlib import Path
 from typing import final
 
 import imap_data_access
-import numpy as np
 import xarray as xr
 from imap_data_access.processing_input import (
     ProcessingInputCollection,
@@ -432,16 +431,16 @@ class ProcessInstrument(ABC):
 
         logger.info("Writing products to local storage")
 
-        list_of_files = [
-            dep_obj.filename_list for dep_obj in dependencies.processing_input
-        ]
-        list_of_files = np.array(list_of_files).flatten()
-        logger.info("Parent files: %s", list_of_files)
+        # Parent files used to create these datasets
+        # https://spdf.gsfc.nasa.gov/istp_guide/gattributes.html.
+        parent_files = [p.name for p in dependencies.get_file_paths()]
+        logger.info("Parent files: %s", parent_files)
 
-        products = [
-            write_cdf(dataset, parent_files=list_of_files.tolist())  # type: ignore[attr-defined]
-            for dataset in datasets
-        ]
+        products = []
+        for ds in datasets:
+            ds.attrs["Parents"] = parent_files
+            products.append(write_cdf(ds))
+
         self.upload_products(products)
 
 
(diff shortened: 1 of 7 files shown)


imap_processing/cli.py
```diff
@@ -432,16 +431,16 @@ def post_processing(
...
         logger.info("Writing products to local storage")
 
-        list_of_files = [
-            dep_obj.filename_list for dep_obj in dependencies.processing_input
-        ]
-        list_of_files = np.array(list_of_files).flatten()
-        logger.info("Parent files: %s", list_of_files)
+        # Parent files used to create these datasets
+        # https://spdf.gsfc.nasa.gov/istp_guide/gattributes.html.
+        parent_files = [p.name for p in dependencies.get_file_paths()]
```
greglucas, line 436: @tech3371 is this the same as before but simplified with your recent refactor to use `get_file_paths()`? I think this is doing what we want and will include all ancillary files etc... Is that what we want?

imap_processing/cli.py
```diff
@@ -596,7 +595,7 @@ def do_processing(
                 / "imap_his_pset-calibration-prod-config_20240101_v001.csv"
             )
             hi_dependencies[0] = load_cdf(hi_dependencies[0])
-            datasets = [hi_l1c.hi_l1c(hi_dependencies, self.version)]
+            datasets = hi_l1c.hi_l1c(hi_dependencies, self.version)
```
greglucas, line 598: @subagonsouth this was different than your other functions which all returned `list[xr.Dataset]` so mocking it didn't work. I changed this to be the same return type. Do you want this to be a solo Dataset instead, or is this OK? Feel free to push changes or suggest what you'd like.

pyproject.toml
```diff
@@ -138,6 +138,7 @@ disable_error_code = ['import-not-found', # Unable to locate module or package s
                 'no-untyped-call', # Function calls are only made to functions that are fully typed
                 'type-arg', # Requires type arguments to be specified in list[] and dict[]
                 'union-attr', # Item "float" of "float | Any" has no attribute "reshape"
+                'warn-return-any', # Returning Any from a function
```
greglucas: @subagonsouth this is to avoid that weird issue with typing in hi_l1c where I think you are returning the correct type and I can get it inferred locally, but mypy doesn't like it. This seems inconsequential so lets just skip over it IMO.

imap_processing/cli.py
tech3371: Ah yes. That does looks much cleaner. And yes, we do want to include science/ancillary and SPICE data used which is not implemented yet. For SPICE, I imagine that we may get a metekernel  but then we need to derived actual kernel inside the metakernel for this. Those are things that are not implemented yet but something I have thought about for this for future.

imap_processing/cli.py
```diff
@@ -432,16 +431,16 @@ def post_processing(
...
-        logger.info("Parent files: %s", list_of_files)
+        # Parent files used to create these datasets
+        # https://spdf.gsfc.nasa.gov/istp_guide/gattributes.html.
+        parent_files = [p.name for p in dependencies.get_file_paths()]
+        logger.info("Parent files: %s", parent_files)
+
+        products = []
+        for ds in datasets:
+            ds.attrs["Parents"] = parent_files
+            products.append(write_cdf(ds))
```
tech3371, line 442: Really good decision to bring this out here!

tech3371 (approved):
parent files change looks good to me!

imap_processing/cli.py
greglucas: :+1: That sounds reasonable to me to add a helper function that would extract those files as well and inject here later once that is set up.

PR #1593, opened 2025-04-10
Version consolidation

greglucas:
# Change Summary

## Overview

We are currently passing version around into every single instrument and requiring all instrument teams to inject the `Data_version` attribute on their own. This is handled in several different ways across instruments, and means that it is also a requirement to do instrument processing. An instrument processing step should only produce a data file, it doesn't need to know anything about which version it is (it may need to know other versions).

Our post-processing step should be injecting the necessary fields, so this pushes that logic into one common place in post-processing to do the global attribute injection.

This is mostly a find/remove job, with one addition being a UserWarning emitted when a default version of `v001` for a file is created (in testing this happens frequently, which I've suppressed as we don't care about it).

Sits on top of #1592 because that is needed to fix some cli mocks.

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -100,10 +100,10 @@ def write_cdf(
     version = dataset.attrs.get("Data_version", None)
     if version is None:
         warnings.warn(
-            "No Data_version attribute found in dataset. Using default v001.",
+            "No Data_version attribute found in dataset. Using default v999.",
             stacklevel=2,
         )
-        version = "v001"
+        version = "v999"
     elif not re.match(r"v\d{3}", version):
         raise ValueError(
             f"The Data_version attribute {version} does not match expected format vXXX."
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -2,6 +2,7 @@
 
 import logging
 import re
+import warnings
 from pathlib import Path
 
 import imap_data_access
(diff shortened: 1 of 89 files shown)


imap_processing/cdf/utils.py
```diff
@@ -102,15 +97,18 @@ def write_cdf(
...
-        version = f"v{int(dataset.attrs['Data_version']):03d}"  # vXXX
-    else:
-        version = dataset.attrs["Data_version"]
+    version = dataset.attrs.get("Data_version", None)
+    if version is None:
+        warnings.warn(
+            "No Data_version attribute found in dataset. Using default v001.",
+            stacklevel=2,
+        )
+        version = "v001"
```
tech3371: I don't know if we need to but should we raise value missing error instead of default to something?

imap_processing/cli.py
```diff
@@ -432,16 +431,18 @@ def post_processing(
...
-        logger.info("Parent files: %s", list_of_files)
+        logger.info("Dataset version: %s", self.version)
+        # Parent files used to create these datasets
+        # https://spdf.gsfc.nasa.gov/istp_guide/gattributes.html.
+        parent_files = [p.name for p in dependencies.get_file_paths()]
+        logger.info("Parent files: %s", parent_files)
+
+        products = []
+        for ds in datasets:
+            ds.attrs["Data_version"] = self.version
```
tech3371, line 442: nice!

imap_processing/cli.py
```diff
@@ -1049,21 +1045,21 @@ def do_processing(
...
-            datasets = ultra_l1b.ultra_l1b(data_dict, self.version)
+            datasets = ultra_l1b.ultra_l1b(data_dict)
 
         elif self.data_level == "l1c":
             data_dict = {}
             for dep in dependency_list:
                 dataset = load_cdf(dep.imap_file_paths[0])
                 data_dict[dataset.attrs["Logical_source"]] = dataset
-            datasets = ultra_l1c.ultra_l1c(data_dict, self.version)
+            datasets = ultra_l1c.ultra_l1c(data_dict)
```
tech3371, line 1062: cool. It reduced one parameter for all instruments!

tech3371 (approved):
Thank you for doing this!

imap_processing/cdf/utils.py
nkerman: I would suggest defaulting to something that's clearly not a standard version (maybe v999 or v000)

greglucas: Great suggestions, thank you!

I just pushed a new commit with v999. That seemed more anomalous to me than 000, although it might also mean that if we do push up files with that default then we will always see that as the latest. I can change to 000 if we'd prefer that as well.

greglucas: We do need this as a warning because our tests use `write_cdf()` pretty frequently without having a version within the dataset.

I was actually thinking that might have been a better way to go initially would be to create an "empty" dataset in preprocessing for every instrument that has everything they will need for attributes already on it. But, I think that would be yet another major refactor/lift and probably much harder than this one.

nkerman: Ah - good point about that showing up as the latest... I think that's still fine, but I wonder if `v-999` or similar is an allowable value?

If not, 999 works!

greglucas: unfortunately not :( Our version checking is too strict! Would have been nice to have some kind of preliminary flag or something for exactly this situation though I agree.

bourque (approved):
LGTM! Thanks for simplifying things for us!

maxinelasp (approved):
Looks great! I only reviewed GLOWS and MAG and the CLI but thank youuuu for doing this, this is way better

Issue #772, opened 2024-08-21
CoDICE: Add support for Event Data (lo-pha and hi-pha)


bourque:
Currently, the CoDICE L1 pipeline creates CDF files with bogus data arrays for the event data (`lo-pha` and `hi-pha`), in order to get things working for SIT-3. These data products should be properly implemented.

PR #1595, opened 2025-04-10
CoDICE L1a lo-pha (direct events) implementation

### LINKED ISSUE #1155: CoDICE: Add support for continuation packets
It was recently discovered that CoDICE data products may use continuation packets, several of which form the data for a single data product. There is a flag in the CCSDS data that indicate which packets are stand-alone, which are part of a segment, which are the first of the segment, and which are the last of the segment. I think I will need to implement some sort of logic to piece these segmented packets together before the data can be decompressed and arranged into a CDF.

bourque:
This PR implements CoDICE L1a `lo-pha` (soon to be renamed to `lo-de`) processing and validation. Pertains to #772 and #251 . Closes #1155 

This PR required a few new functions, namely to support continuation packets and group packets appropriately based on their `seq_flgs` values. The structuring/shaping of the event data is also a bit tricky since it requires digging into individual ranges of bits to pull out specific fields for specific priorities.

I also added some code to begin handling `hi-pha` products, but I will finish implementing/validation that product in a future PR.


imap_processing/codice/codice_l1a.py
```diff
@@ -566,55 +566,125 @@ def set_data_product_config(
...
+    # Gather the CDF attributes
     cdf_attrs = ImapCdfAttributes()
     cdf_attrs.add_instrument_global_attrs("codice")
     cdf_attrs.add_instrument_variable_attrs("codice", "l1a")
     cdf_attrs.add_global_attribute("Data_version", data_version)
 
     # Define coordinates
     epoch = xr.DataArray(
-        packet.epoch,
+        packets.epoch[::num_priorities],
```
greglucas, line 638: Can you add a comment about why you're doing this? Should num_priorities be another dimension of the dataset?

imap_processing/codice/codice_l1a.py
```diff
@@ -719,6 +789,75 @@ def get_params(dataset: xr.Dataset) -> tuple[int, int, int, int]:
     return table_id, plan_id, plan_step, view_id
 
 
+def group_data(packets: xr.Dataset) -> list[bytes]:
+    """
+    Organize continuation packets into appropriate groups.
```
greglucas, line 776: Bummer, I didn't realize you also had continuation packets to deal with otherwise I would have pushed harder to get a space_packet_parser release out the door with that support.
https://github.com/lasp/space_packet_parser/pull/115

@vmartinez-cu I think is doing a grouping for HIT, but I'm not sure you can re-use much code because your decommed fields you're indexing into are different. But worth hearing what she has to say.

imap_processing/codice/codice_l1a.py
```diff
@@ -719,6 +789,75 @@ def get_params(dataset: xr.Dataset) -> tuple[int, int, int, int]:
...
+    """
+    grouped_data = []  # Holds the properly grouped data to be decompressed
+    current_group = bytearray()  # Temporary storage for current group
+    group_byte_count = None  # Temporary storage for current group byte count
+
+    for packet_data, group_code, byte_count in zip(
+        packets.event_data.data, packets.seq_flgs.data, packets.byte_count.data
+    ):
+        # Convert from numpy array to byte object
+        compressed_packet_values = ast.literal_eval(str(packet_data))
```
greglucas: Why do you need to do this step with literal_eval? Are you trying to get to integers, do you need to call `.to_bytes()` on the result `int(packet_data, 2).to_bytes()?
If you have a numpy array you should be able to call `tobytes()` https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.tobytes.html But, if I remember right, I think you have a string dtype with 1s and 0s representing the binary? So can you append all those 1s and 0s together as a string and do this in one final step?

imap_processing/codice/codice_l1a.py
```diff
@@ -719,6 +789,75 @@ def get_params(dataset: xr.Dataset) -> tuple[int, int, int, int]:
...
+        # If the group code is 0, this means the data is part of the middle of
+        # the group
+        elif group_code == 0:
+            current_group += compressed_packet_values
+
+        # If the group code is 2, this means the data is the last data in the
+        # group
+        elif group_code == 2:
+            current_group += compressed_packet_values
+            values_to_decompress = current_group[:group_byte_count]
```
greglucas, line 829: Do you know if there are any that are longer than the group_byte_count that are getting trimmed by this? Do we need to warn/log that we got more data than expected in the packets?

imap_processing/codice/codice_l1a.py
```diff
@@ -719,6 +789,75 @@ def get_params(dataset: xr.Dataset) -> tuple[int, int, int, int]:
...
+            grouped_data.append(values_to_decompress)
+            current_group = bytearray()
+            group_byte_count = None
+
+        else:
+            raise ValueError(f"Unexpected group code: {group_code}")
```
greglucas: This is unreachable because the group_code is 2 bits long.

imap_processing/codice/codice_l1a.py
```diff
@@ -741,6 +880,115 @@ def log_dataset_info(datasets: dict[int, xr.Dataset]) -> None:
...
+    for i in range(num_priorities):
+        for field in constants.LO_DE_BIT_STRUCTURE:
+            if field not in ["Priority", "Spare"]:
+                data[f"P{i}_{field}"] = []
+        data[f"P{i}_NumEvents"] = []
+        data[f"P{i}_DataQuality"] = []
+
+    # decompressed_data is one large list of values of length
+    # (<number of epochs> * <8 priorities>)
+    # Chunk the data into each epoch/priority combination
```
greglucas: Rather than appending/updating lists here can you make the initial data arrays with 0s of the expected shape? (nepoch, 10000) and then as you're going through the loops just fill in the proper index value?

imap_processing/codice/codice_l1a.py
```diff
@@ -741,6 +880,115 @@ def log_dataset_info(datasets: dict[int, xr.Dataset]) -> None:
...
+                    constants.LO_DE_BIT_STRUCTURE.items()
+                ):
+                    data_per_epoch[f"P{priority_num}_{field_name}"].append(
+                        int(bit_string[index : index + bit_length], 2)
+                    )
+                    index += bit_length
+
+        # Append the epoch data to the final, restructured list of data
+        # For any data with less than 10000 events, pad the data array up to
+        # 10000
```
greglucas: I think this is because you have a variable number of events / epoch and CDF / arrays in general don't like spare matrices but rather fully filled structured arrays. Not ideal, but seems fine 🤷

imap_processing/codice/codice_l1a.py
```diff
@@ -741,6 +880,115 @@ def log_dataset_info(datasets: dict[int, xr.Dataset]) -> None:
...
+                if field not in ["Priority", "Spare"]:
+                    event_data = np.array(
+                        data_per_epoch[f"P{i}_{field}"], dtype=np.uint16
+                    )
+                    padding_size = 10000 - event_data.size
+                    padded_event_data = np.pad(
+                        event_data,
+                        (0, padding_size),
+                        mode="constant",
+                        constant_values=0,
```
greglucas: Is 0 the FILL_VAL here? Maybe we need to set it to whatever your FILL_VAL is in the attributes to notify folks this is an entry to be ignored.

bourque:
pre-commit.ci autofix

imap_processing/codice/codice_l1a.py
bourque: Thanks for catching this! I think this is some sort of leftover code from an earlier implementation, because it turns out this step is not needed at all. `packet_data` is already a `bytes` type, so I can just directly append other grouped data and/or cut out the padding on that variable.

imap_processing/codice/codice_l1a.py
bourque: Good catch -- I will use the FILL_VAL here (which is `255`) so that it is more apparent.

imap_processing/codice/codice_l1a.py
bourque: Thanks for the suggestion here, this does simplify things quite a bit!

bourque:
Note that codecov is failing because I have turned off testing to `hi-pha` data products as I am working to properly implement that data product. This should resolve itself once I finish that.

imap_processing/codice/codice_l1a.py
```diff
@@ -725,6 +856,99 @@ def log_dataset_info(datasets: dict[int, xr.Dataset]) -> None:
...
+    # There is one epoch per set of priorities
+    num_epochs = len(packets.epoch.data) // num_priorities
+
+    # Initialize data arrays for each priority and field to store the data
+    # We also need arrays to hold number of events and data quality
+    for priority_num in range(num_priorities):
+        for field in constants.LO_DE_BIT_STRUCTURE:
+            if field not in ["Priority", "Spare"]:
+                data[f"P{priority_num}_{field}"] = np.full(
+                    (num_epochs, 10000), 255, dtype=np.uint16
```
greglucas, line 895: 255 is an odd fill value for a uint16. Should these be uint8, or the maximum uint16 value 2**16 - 1?

imap_processing/codice/codice_l1a.py
```diff
@@ -725,6 +856,99 @@ def log_dataset_info(datasets: dict[int, xr.Dataset]) -> None:
...
+
+            # Number of events and data quality can be determined at this stage
+            num_events = len(priority_data) // num_priorities
+            data[f"P{priority_num}_NumEvents"][epoch_index] = num_events
+            data[f"P{priority_num}_DataQuality"][epoch_index] = data_quality[i]
+
+            # Iterate over each event
+            for event_index in range(num_events):
+                event_start = event_index * num_priorities
+                event_end = (event_index + 1) * num_priorities
```
greglucas: Earlier you did start + num_priorities, so for consistency it might be good to use that notation here too.
```suggestion
                event_end = event_start + num_priorities
```

imap_processing/codice/codice_l1a.py
```diff
@@ -725,6 +856,99 @@ def log_dataset_info(datasets: dict[int, xr.Dataset]) -> None:
...
+                event = priority_data[event_start:event_end]
+                # Separate out each individual field from the bit string
+                # The fields are packed into the bit string in reverse order, so
+                # we need to back them out in reverse order
+                bit_string = "".join(f"{byte:08b}" for byte in event)
```
greglucas: space_packet_parser recommends this for formatting the bitstring from a bytes-like object instead of needing a for-loop: `f"{int.from_bytes(data, byteorder='big'):0{len(data)*8}b}"`
https://github.com/lasp/space_packet_parser/blob/main/docs/source/changelog.md

This is fine too, just not sure if there is a big performance difference on it or not.

greglucas (approved):
Looks good, just some minor comments that don't impact the logic.

imap_processing/codice/codice_l1a.py
bourque: Good point. Ultimately I should be assigning a specific dtype and fill val for each of these bit fields (I have a TODO for this in the code) -- some of these are going to be uint8, while others are uint16. I'll tackle that in a future PR.

imap_processing/codice/codice_l1a.py
bourque: Good to know, thanks!

bourque:
Note that the failed tests here are due to some current issue with `scipy` and Windows (see Slack thread for discussion). I am going to merge this despite the failed tests so I can start working off of this code.

PR #1596, opened 2025-04-10
ENA Maps: Convert HealpixSkyMap to RectangularSkyMap

### LINKED ISSUE #1398: Implement Proper Conversion from HealpixSkyMap to RectangularSkyMap
### Algorithm Description:

Nick Dutton on the Ultra Instrument team is working on the algorithm to convert (via interpolation) from a Healpix map of the sky to a Rectangularly gridded map of the sky. This method "approximates a double integral over the healpix and square pixel overlapping regions". 

### Related Issues/PRs:

#1397

nkerman:
# Change Summary

Converts a HealpixSkyMap to RectangularSkyMap using recursive subdivision of each rectangular pixel.

Closes #1398 

## Overview

To get the value of a Rectangular Sky Map pixel, you can approximate it as the value of the Healpix Sky Map pixel into which the rectangular pixel's center falls. If the rectangular pixel overlaps multiple Healpix pixels, this will be less accurate, especially if the gradient between those Healpix pixel values is large. You can increase the accuracy by subdividing the rectangular pixel into 4 equal area subpixels, and taking the mean of their center values. Weight this mean by their solid angles, which change by latitude. You can do this recursively until you get a value which doesn't change significantly between subsequent subdivisions. 

This is a somewhat computationally expensive process (running locally, it can take ~5 seconds to interpolate each variable on medium gradients), but will be quite rare operation, only run at the end of some map-making. I've tried to parallelize it, but I've been so far unable to make it significantly faster, and it doesn't seem like a high priority. 

The best way I can explain this process is through [this diagram - see the current version on Galaxy confluence](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/256936180/Ultra+L2+Diagrams#UltraL2Diagrams-ConversionfromHealpix%E2%86%92RectangularConceptualDiagram). 

image

Figures:
1. Original Healpix Sky Map (nside=32) Flux
image
2. Interpolated Rectangular Sky Map (2˚) Flux image
3. The number of subdivisions necessary at each pixel. Note how this is very low except around high gradients. newplot
# New Dependencies

## Updated Files

   - Add methods to `HealpixSkyMap` to: 
     - calculate a rectangular pixel's mean value at a given number of subdivisions
     - calculate a rectangular pixel's mean value recursively using more and more subdivisions until the tolerance is met
     - Build a `RectangularSkyMap` by interpolating each pixel's value. 

## Testing

Tests for the 3 new methods

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -1203,7 +1203,7 @@ class HealpixSkyMap(AbstractSkyMap):
 
     # Allow for 5 arguments, and self to be passed.
     # ruff: noqa: PLR0913
-    def get_pixel_value_recursive_subdivs(
+    def get_rect_pixel_value_recursive_subdivs(
         self,
         rect_pix_center_lon_lat: np.typing.NDArray | tuple[float, float],
         rect_pix_spacing_deg: float,
@@ -1339,7 +1339,7 @@ class HealpixSkyMap(AbstractSkyMap):
             # for each value key.
             healpix_values_array = self.data_1d[value_key]
             best_value_and_recursion_depth_by_pixel = [
-                self.get_pixel_value_recursive_subdivs(
+                self.get_rect_pixel_value_recursive_subdivs(
                     rect_pix_center_lon_lat=lon_lat,
                     rect_pix_spacing_deg=rect_map.spacing_deg,
                     value_key=value_key,
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -1090,7 +1090,7 @@ class HealpixSkyMap(AbstractSkyMap):
         self,
         rect_pix_center_lon_lat: np.typing.NDArray | tuple[float, float],
         rect_pix_spacing_deg: float,
-        value_key: str,
+        value_array: xr.DataArray,
         num_subdivisions: int,
     ) -> np.typing.NDArray:
         """
@@ -1106,8 +1106,8 @@ class HealpixSkyMap(AbstractSkyMap):
             The center longitude and latitude of the rectangular pixel.
         rect_pix_spacing_deg : float
             The spacing of the rectangular pixel in degrees.
-        value_key : str
-            The name of the value to interpolate from the healpix map.
+        value_array : xr.DataArray
+            The data array containing the healpix map values.
         num_subdivisions : int
             The number of subdivisions to create for the rectangular pixel.
             The more subdivisions, the more accurate the interpolation, but also
@@ -1118,9 +1118,8 @@ class HealpixSkyMap(AbstractSkyMap):
         np.typing.NDArray
             The mean value of the healpix map at the subpixel centers.
 
-            If the array associated with the key value_key has a single value
-            at each pixel, the output will be a single value,
-            but if there are other dimensions,
+            If value_array has a single value at each pixel, the output
+            will be a single value, but if there are other dimensions,
             (e.g., if self.data_1d['flux'].sizes =
             {"epoch": 1, "energy": 24, "pixel": 16200}),
             the output will be an array with the same dims except the pixel dimension
@@ -1186,9 +1185,7 @@ class HealpixSkyMap(AbstractSkyMap):
             lonlat=True,
         )
         # Get the healpix values at the rectangular subpixel centers
-        hp_vals_at_rect_pix_ctrs = self.data_1d[value_key].values[
-            ..., hp_pix_at_rect_subpix_ctrs
-        ]
+        hp_vals_at_rect_pix_ctrs = value_array.values[..., hp_pix_at_rect_subpix_ctrs]
 
         # Weighted mean (weighted by solid angle) of these values over the pixel axis,
         # which is the last axis of this array
@@ -1196,8 +1193,7 @@ class HealpixSkyMap(AbstractSkyMap):
             hp_vals_at_rect_pix_ctrs * rect_subpix_solid_angle_by_lat
         )
         mean_pixel_value = (
-            weighted_hp_vals_at_rect_pix_ctrs.sum(axis=(-1))
-            / full_rect_pixel_solid_angle
+            weighted_hp_vals_at_rect_pix_ctrs.sum(axis=-1) / full_rect_pixel_solid_angle
         )
         return mean_pixel_value
 
@@ -1205,7 +1201,7 @@ class HealpixSkyMap(AbstractSkyMap):
         self,
         rect_pix_center_lon_lat: np.typing.NDArray | tuple[float, float],
         rect_pix_spacing_deg: float,
-        value_key: str,
+        value_array: xr.DataArray,
         *,
         rtol: float = 1e-3,
         atol: float = 1e-12,
@@ -1227,8 +1223,8 @@ class HealpixSkyMap(AbstractSkyMap):
             The center longitude and latitude of the rectangular pixel.
         rect_pix_spacing_deg : float
             The spacing of the rectangular pixel in degrees.
-        value_key : str
-            The name of the value to interpolate from the healpix map.
+        value_array : xr.DataArray
+            The data array containing the healpix map values to interpolate from.
         rtol : float, optional
             The relative tolerance for convergence, by default 1e-3.
         atol : float, optional
@@ -1255,7 +1251,7 @@ class HealpixSkyMap(AbstractSkyMap):
                 self.calculate_rect_pixel_value_from_healpix_map_n_subdivisions(
                     rect_pix_center_lon_lat=rect_pix_center_lon_lat,
                     rect_pix_spacing_deg=rect_pix_spacing_deg,
-                    value_key=value_key,
+                    value_array=value_array,
                     num_subdivisions=depth,
                 )
             )
@@ -1263,6 +1259,8 @@ class HealpixSkyMap(AbstractSkyMap):
             # Determine if tolerance is met
             # (skip on the 0th iteration, as there's no delta)
             if depth > 0:
+                # TODO: Ask Nick/Ultra Instrument team if we need to compare each value
+                # in the pixel's array, or just the mean value.
                 if np.isclose(
                     mean_pixel_value.mean(),
                     previous_mean_pixel_value.mean(),
@@ -1328,8 +1326,6 @@ class HealpixSkyMap(AbstractSkyMap):
         # Dict to hold the subdivision depth by pixel for each value key
         subdiv_depth_dict = {}
         for value_key in value_keys:
-            self.data_1d[value_key]
-
             # For each of the values, calculate each pixel's value with
             # recursive subdivision. Unfortunately, this must be done independently
             # for each value key.
@@ -1338,7 +1334,7 @@ class HealpixSkyMap(AbstractSkyMap):
                 self.get_rect_pixel_value_recursive_subdivs(
                     rect_pix_center_lon_lat=lon_lat,
                     rect_pix_spacing_deg=rect_map.spacing_deg,
-                    value_key=value_key,
+                    value_array=healpix_values_array,
                     max_subdivision_depth=max_subdivision_depth,
                 )
                 for lon_lat in rect_map.az_el_points
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -1085,6 +1085,7 @@ class HealpixSkyMap(AbstractSkyMap):
             }
         )
 
+    # Define methods for converting a Healpix map to a Rectangular map:
     def calculate_rect_pixel_value_from_healpix_map_n_subdivisions(
         self,
         rect_pix_center_lon_lat: np.typing.NDArray | tuple[float, float],
@@ -1093,10 +1094,11 @@ class HealpixSkyMap(AbstractSkyMap):
         num_subdivisions: int,
     ) -> np.typing.NDArray:
         """
-        Interpolate the value of a rectangular pixel from a healpix map w/ subdivisions.
+        Interpolate the value of 1 rectangular pixel from a healpix map w/ subdivisions.
 
-        This function splits a rectangular pixel into smaller subpixels
-        and calculates the mean value of the healpix map at those subpixel centers.
+        This function splits a single rectangular pixel into smaller subpixels
+        and calculates the solid angle weighted mean value of
+        the healpix map at all of the subpixel centers.
 
         Parameters
         ----------
@@ -1115,10 +1117,12 @@ class HealpixSkyMap(AbstractSkyMap):
         -------
         np.typing.NDArray
             The mean value of the healpix map at the subpixel centers.
+
             If the array associated with the key value_key has a single value
             at each pixel, the output will be a single value,
-            but if there are other dimensions, (e.g., if
-            self.data_1d['flux'].sizes = {"epoch": 1, "energy": 24, "pixel": 16200}),
+            but if there are other dimensions,
+            (e.g., if self.data_1d['flux'].sizes =
+            {"epoch": 1, "energy": 24, "pixel": 16200}),
             the output will be an array with the same dims except the pixel dimension
             (e.g., (1, 24)).
         """
@@ -1230,30 +1234,12 @@ class HealpixSkyMap(AbstractSkyMap):
         """
         relative_tolerance, absolute_tolerance = tolerances
 
-        # Calculate mean value at the 0th level of recursion (no subdivision of the pix)
-        # TODO: I think we can put this in the 0th level of recursion and
-        # skip calculating the delta and comparing it to the tolerance
-        depth = 0
-        rect_pix_center_lon_lat = np.reshape(rect_pix_center_lon_lat, (-1, 2))
-        hp_pix_at_rect_subpix_ctr = hp.ang2pix(
-            nside=self.nside,
-            nest=self.nested,
-            theta=rect_pix_center_lon_lat[:, 0],
-            phi=rect_pix_center_lon_lat[:, 1],
-            lonlat=True,
-        )
-        hp_vals_at_rect_pix_ctrs = self.data_1d[value_key].values[
-            ..., hp_pix_at_rect_subpix_ctr
-        ]
-        mean_pixel_value_at_level = [
-            hp_vals_at_rect_pix_ctrs.mean(axis=(-1)),
-        ]
-
         # Recursively subdivide a pixel and calculate its mean value until either the
         # difference between consecutive levels is within the specified tolerances
         # or the maximum recursion depth is reached
+        depth = 0
+        mean_pixel_value_at_level = []
         while depth < MAX_RECURSION_DEPTH:
-            depth += 1
             mean_pixel_value = (
                 self.calculate_rect_pixel_value_from_healpix_map_n_subdivisions(
                     rect_pix_center_lon_lat=rect_pix_center_lon_lat,
@@ -1265,21 +1251,22 @@ class HealpixSkyMap(AbstractSkyMap):
             mean_pixel_value_at_level.append(mean_pixel_value)
 
             # Determine if tolerance is met
-            abs_delta = np.abs(
-                mean_pixel_value_at_level[-1].mean()
-                - mean_pixel_value_at_level[-2].mean()
-            )
-            total_abs_tolerance = (
-                relative_tolerance * np.abs(mean_pixel_value_at_level[-1].mean())
-                + absolute_tolerance
-            )
-            if abs_delta < total_abs_tolerance:
-                break
+            if depth > 0:
+                abs_delta = np.abs(
+                    mean_pixel_value_at_level[-1].mean()
+                    - mean_pixel_value_at_level[-2].mean()
+                )
+                total_abs_tolerance = (
+                    relative_tolerance * np.abs(mean_pixel_value_at_level[-1].mean())
+                    + absolute_tolerance
+                )
+                if abs_delta < total_abs_tolerance:
+                    break
+            depth += 1
 
         # Only keep the last (best) mean pixel value
         return mean_pixel_value_at_level[-1], depth
 
-    # Define methods for converting a Healpix map to a Rectangular map
     def to_rectangular_skymap_with_recusive_subdivision(
         self,
         rect_spacing_deg: float,
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -26,7 +26,7 @@ from imap_processing.spice.time import ttj2000ns_to_et
 logger = logging.getLogger(__name__)
 
 # Set the maximum recursion depth for the conversion from Healpix to rectangular SkyMap.
-MAX_RECURSION_DEPTH = 8
+MAX_SUBDIV_RECURSION_DEPTH = 8
 
 
 class SkyTilingType(Enum):
@@ -1085,7 +1085,7 @@ class HealpixSkyMap(AbstractSkyMap):
             }
         )
 
-    # Define methods for converting a Healpix map to a Rectangular map:
+    # Define several methods for converting a Healpix map to a Rectangular map:
     def calculate_rect_pixel_value_from_healpix_map_n_subdivisions(
         self,
         rect_pix_center_lon_lat: np.typing.NDArray | tuple[float, float],
@@ -1094,7 +1094,7 @@ class HealpixSkyMap(AbstractSkyMap):
         num_subdivisions: int,
     ) -> np.typing.NDArray:
         """
-        Interpolate the value of 1 rectangular pixel from a healpix map w/ subdivisions.
+        Interpolate the value of a rectangular pixel from a healpix map w/ subdivisions.
 
         This function splits a single rectangular pixel into smaller subpixels
         and calculates the solid angle weighted mean value of
@@ -1127,7 +1127,8 @@ class HealpixSkyMap(AbstractSkyMap):
             (e.g., (1, 24)).
         """
         # Assumes that you already checked the pixel doesn't fall entirely in an HP pix
-        # TODO: Add this here? It shouldn't really be necessary, as the next function
+        # TODO: Ask Nick if we need to add this here to mimic his code.
+        # It shouldn't really be necessary, as the next function
         # get_pixel_value_recursive_subdivs will finish at 1 subdivision
 
         # Ensure input contains lon in the first column and lat in the second column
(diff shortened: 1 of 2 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,296 @@ def __init__(
...
+        )
+
+        # We must weight by solid angle, which is not exactly equal for all subpixels
+        # Calculate the solid angle of the full rectangular pixel (sterad)
+        full_rect_pixel_solid_angle = np.deg2rad(rect_pix_spacing_deg) * (
+            np.sin(np.deg2rad(bottom_edge_lat + rect_pix_spacing_deg))
+            - np.sin(np.deg2rad(bottom_edge_lat))
+        )
+
+        # Calculate solid angle of each subpix from the rect_subpix_lat_ctrs (sterad)
```
laspsandoval, line 1161: Instead of using:
Δsin(lat) × Δlon
Could you do:
Δlat × Δlon × cos(lat)
To make a more accurate calculation at the poles?

laspsandoval (commented):
I had a question/comment about possibly using another equation to calculate the solid angles at the poles a bit more accurately. I'm interested to hear what you think. Nice work! I'll wait until others have commented before approving.

imap_processing/ena_maps/ena_maps.py
nkerman: Hi @laspsandoval - thanks! Out of curiosity, where are you getting this simplification from? All the online references I've encountered point to using 
[1]

$$(\phi_1 - \phi_0) \times ((sin(\theta_1) - sin(\theta_0))$$

When I integrate it myself...
[2]

$$\int\int{{sin(\theta)}}d\theta d\phi$$

I get 

[3]

$$(C \theta_1 - \phi_1 cos(\theta_1)) - (C \theta_0 - \phi_0 **cos(\theta_0)$$

I can't seem to get this to reduce to what you have here, but it's been a long time since I've done a double integral!

nkerman: Okay, I'm probably way too caught up in this, but I wanted to look at this numerically too:

image

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+        ----------
+        rect_pix_center_lon_lat : np.typing.NDArray | tuple[float, float]
+            The center longitude and latitude of the rectangular pixel.
+        rect_pix_spacing_deg : float
+            The spacing of the rectangular pixel in degrees.
+        value_key : str
+            The name of the value to interpolate from the healpix map.
+        tolerances : tuple[float, float], optional
+            The relative and absolute tolerances for convergence,
+            by default (1e-3, 1e-12).
```
greglucas: Should be separate atol and etol parameters I think to match numpy's conventions.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+            ..., hp_pix_at_rect_subpix_ctrs
+        ]
+
+        # Weighted mean (weighted by solid angle) of these values over the pixel axis,
+        # which is the last axis of this array
+        weighted_hp_vals_at_rect_pix_ctrs = (
+            hp_vals_at_rect_pix_ctrs * rect_subpix_solid_angle_by_lat
+        )
+        mean_pixel_value = (
+            weighted_hp_vals_at_rect_pix_ctrs.sum(axis=(-1))
```
greglucas:
```suggestion
            weighted_hp_vals_at_rect_pix_ctrs.sum(axis=-1)
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+        )
+        return mean_pixel_value
+
+    # Allow for 5 arguments, and self to be passed.
+    # ruff: noqa: PLR0913
+    def get_rect_pixel_value_recursive_subdivs(
+        self,
+        rect_pix_center_lon_lat: np.typing.NDArray | tuple[float, float],
+        rect_pix_spacing_deg: float,
+        value_key: str,
```
greglucas: I think you can avoid the maximum arguments ruff complaint by making things keyword only after a certain amount. I think this would make sense here too, so you aren't implicitly relying on the ordering of arguments, but forcing people use keywords which is more descriptive.
```suggestion
        value_key: str,
        *,
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+        tuple[list[float], int]
+            The mean value at the final level of subdivision and the depth of recursion.
+        """
+        relative_tolerance, absolute_tolerance = tolerances
+
+        # Recursively subdivide a pixel and calculate its mean value until either the
+        # difference between consecutive levels is within the specified tolerances
+        # or the maximum recursion depth is reached
+        depth = 0
+        mean_pixel_value_at_level = []
```
greglucas: Can you keep track of the `prev_mean_pixel_value`, instead of keeping this list? It looks like you only want the current and previous values.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+                    rect_pix_spacing_deg=rect_pix_spacing_deg,
+                    value_key=value_key,
+                    num_subdivisions=depth,
+                )
+            )
+            mean_pixel_value_at_level.append(mean_pixel_value)
+
+            # Determine if tolerance is met
+            # (skip on the 0th iteration, as there's no delta)
+            if depth > 0:
```
greglucas, line 1266: To avoid an indentation of the math portion, I think you can swap this conditional.
```suggestion
            if depth > 0:
                continue
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+                abs_delta = np.abs(
+                    mean_pixel_value_at_level[-1].mean()
+                    - mean_pixel_value_at_level[-2].mean()
```
greglucas: I'm not entirely following what the `mean_pixel_value_at_level[-1]` is. Within the loops you are only comparing `.mean()` quantities, but then you return the entire array of values at the end. Can you switch over to just the single value?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+                )
+                total_abs_tolerance = (
+                    relative_tolerance * np.abs(mean_pixel_value_at_level[-1].mean())
+                    + absolute_tolerance
+                )
+                if abs_delta < total_abs_tolerance:
```
greglucas: Can you use numpy's `isclose()` here to use their relative / absolute tolerances?
https://numpy.org/doc/stable/reference/generated/numpy.isclose.html

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+                    break
+            depth += 1
+
+        # Only keep the last (best) mean pixel value
+        return mean_pixel_value_at_level[-1], depth
```
greglucas: Do you need the depth returned, or is it more for informational purposes?

You might want to add some logging debug statements inside the loop for tracking/informational purposes that might be a way to help debug what is going on later. This goes for all of these functions, where it might be helpful to log: current pixel coordinates, depth, value, ... as you're going through this chain. (Not info level please as that would pollute the logs I'm guessing)

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+        )
+
+        # Dict to hold the subdivision depth by pixel for each value key
+        subdiv_depth_dict = {}
+        for value_key in value_keys:
+            self.data_1d[value_key]
+
+            # For each of the values, calculate each pixel's value with
+            # recursive subdivision. Unfortunately, this must be done independently
+            # for each value key.
```
greglucas, line 1341: Depending on the number of values covering the map I assume? I am kind of curious if it is worthwhile just making the maximum recursion depth mapping: 256 x 256 x nside once (~8M) points, and then re-use that mapping for every single value as a kind of index lookup table. So just live with the fact that you'll do one expensive calculation of subdivisions, but use that knowledge elsewhere in your interpolations to speed those subsequent steps up.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+            healpix_values_array = self.data_1d[value_key]
+            best_value_and_recursion_depth_by_pixel = [
+                self.get_rect_pixel_value_recursive_subdivs(
+                    rect_pix_center_lon_lat=lon_lat,
+                    rect_pix_spacing_deg=rect_map.spacing_deg,
+                    value_key=value_key,
```
greglucas: Rather than value_key, can you pass the value array in? `self.data_1d[value_key]` I think. That way this could actually be a self-contained helper function and not rely on "self". I would rather project a value than a name, and this might also save many __getitem__ calls later too?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+                    max_subdivision_depth=max_subdivision_depth,
+                )
+                for lon_lat in rect_map.az_el_points
+            ]
+
+            interpolated_data_by_rect_pixel = np.moveaxis(
+                [r[0] for r in best_value_and_recursion_depth_by_pixel], 0, -1
+            )
```
greglucas: You are doing another for loop over all of the values, right after returning these values. Can you return the values in the order you want them in?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -867,6 +870,304 @@ def __init__(
...
+                    CoordNames.GENERIC_PIXEL.value,
+                    CoordNames.HEALPIX_INDEX.value,
+                ):
+                    rect_map.data_1d.coords[coord] = healpix_values_array.coords[coord]
+
+            # Add the subdivision depth by pixel of this value_key to the dictionary
+            # This may be necessary for uncertainty estimation
+            subdiv_depth_dict[value_key] = np.array(
+                [r[1] for r in best_value_and_recursion_depth_by_pixel]
+            )
```
greglucas: OK, one of my previous comments was wondering about this piping through of depth. When I read "may be necessary" I lean towards removing it for now and adding it in later when it is necessary. It is adding quite some complexity in my opinion of rearranging axes, and adding an extra tuple return for something that "might" happen.

imap_processing/ena_maps/ena_maps.py
nkerman: Yeah, I had originally done them with separate parameters but didn't want to have to ignore ruff's 5 parameter limit (of which I'm not a huge fan). 

I think I can use your kwarg suggestion here too though!

imap_processing/ena_maps/ena_maps.py
nkerman: Yeah that works! They should be small enough values that I doubt it's an efficiency issue, but this is a bit simpler!

imap_processing/ena_maps/ena_maps.py
nkerman: This doesn't play very nicely with the switch to storing a `previous_mean_pixel_value`, since it would mean we then need to set `previous_mean_pixel_value` in two different places, as we can't do it before the conditionals. 
I think it's okay to leave the indent as is, but definitely feel free to push back on that.

imap_processing/ena_maps/ena_maps.py
nkerman: SO the deal here is that a given pixel may have a bunch of different values: e.g., if there are 24 different energy bands, then the shape of a single pixel's data will actually be `(1 epoch x 24 energies)`. 

The only place we want to calculate the mean of all these values (`mean_pixel_value.mean()`) would be in the tolerance check (I should also run this by Nick, but I think we want to compare a single value for each pixel to check its tolerance, not 24). Everywhere else, we want to return the full (1, 24) array (`mean_pixel_value`). 

I can add some comments to this effect.

imap_processing/ena_maps/ena_maps.py
nkerman: Yep - not sure why I was using a manual version of this - I think I was trying to track some of Nick's sketches too closely.

imap_processing/ena_maps/ena_maps.py
nkerman: > Do you need the depth returned, or is it more for informational purposes?

I checked in about this today and it's looking like Ultra does want to be able to track this, at least as an option in debug mode or similar. 

I want to spend a bit of time thinking through a better way of passing out the pixel-level information, because I think the main problem/inelegant bit here is how we have to unwrap those outputs. 
I considered, (but ultimately lean against,) passing in previously-intialized arrays of (`calculated_mean_value_by_pixel` and `subdivision_depth_by_pixel`) into the `get_rect_pixel_value_recursive_subdivs` method. Ultimately though, I want these to be independent methods that only care about a pixel's location on the sky. Let me know if you have any other ideas for splitting up the per-pixel outputs more neatly. 

>You might want to add some logging debug statements inside the loop for tracking/informational purposes that might be a way to help debug what is going on later. This goes for all of these functions, where it might be helpful to log: current pixel coordinates, depth, value, ... as you're going through this chain. (Not info level please as that would pollute the logs I'm guessing)

That's a good idea. I can add some logger debugs, as long as we don't care if **lots** end up getting generated.

imap_processing/ena_maps/ena_maps.py
nkerman: Yep that should work. I had originally been thinking you only needed to do recursion once per dataset, and thus that we could pass in all the values at once, but then realized it must be done separately for each value_key.

nkerman: To confirm, the functions would still be methods of `HealpixSkyMap`, they just wouldn't require self, right?

imap_processing/ena_maps/ena_maps.py
nkerman: This relates back to [this thread](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1596/files#r2047626840). I don't love the separation of these values either and am looking for a neater way. Ultra does, at least somewhat want the ability to output an additional map-like data variable of `subdivision_depth_by_pixel`. 

We could push back, but it's not hard in my opinion, more just inelegant.

nkerman (commented):
Initial responses to Greg's PR comments - thanks Greg!

imap_processing/ena_maps/ena_maps.py
greglucas: > ruff's 5 parameter limit (of which I'm not a huge fan)

I am ambivalent and get annoyed sometimes too, but I think the idea here is that you don't want to allow people to pass in `method(0, 1, 2, 3, 4, 5, 6)` because you don't know what those numbers line up with, whereas if you add explicit keyword arguments it helps identify the mapping more explicitly. (As an aside, it is really helpful to make keyword-only for deprecation/moving around purposes on larger projects that others are dependent on so you can change the order without impacting people downstream)

imap_processing/ena_maps/ena_maps.py
greglucas: If you are in a hot-loop (recursively going down these paths for all points) you will find that appending to a list can slow things down I think.

imap_processing/ena_maps/ena_maps.py
greglucas: Nope, that is fine if it simplifies other things!

imap_processing/ena_maps/ena_maps.py
greglucas: > That's a good idea. I can add some logger debugs, as long as we don't care if lots end up getting generated.

Just make sure they are DEBUG-level and we should only be running in INFO-level nominally and not spamming the logs. So they are there for your debugging purposes.

greglucas: > I want to spend a bit of time thinking through a better way of passing out the pixel-level information, because I think the main problem/inelegant bit here is how we have to unwrap those outputs.
I considered, (but ultimately lean against,) passing in previously-intialized arrays of (calculated_mean_value_by_pixel and subdivision_depth_by_pixel) into the get_rect_pixel_value_recursive_subdivs method. Ultimately though, I want these to be independent methods that only care about a pixel's location on the sky. Let me know if you have any other ideas for splitting up the per-pixel outputs more neatly.

I agree with you wanting to do this on a per-pixel basis, I think that makes sense too at first glance unless you wanted to make the large grid all at once like in one of my other comments. If these are tied together, you might want to recursively pass a tuple `(value, depth)` as your input combined together so that is sort of "the" quantity you want to pass around.

imap_processing/ena_maps/ena_maps.py
greglucas: Yeah, I think that is fine, a staticmethod then, or just a general helper somewhere if you prefer. I don't have a preference either way.

imap_processing/ena_maps/ena_maps.py
greglucas: Yeah, it is fine to keep it around if it is desired 👍

greglucas:
> SO the deal here is that a given pixel may have a bunch of different values: e.g., if there are 24 different energy bands, then the shape of a single pixel's data will actually be (1 epoch x 24 energies).

> The only place we want to calculate the mean of all these values (mean_pixel_value.mean()) would be in the tolerance check (I should also run this by Nick, but I think we want to compare a single value for each pixel to check its tolerance, not 24). Everywhere else, we want to return the full (1, 24) array (mean_pixel_value).

OK, I wasn't following that reading throught he code initially, can you add in a comment about the shapes / dimensions being passed around somewhere so that it is easier to track?

I don't understand why you would want to average over energy though, so I think I'm still missing that part. If you get 5 counts in one energy level and 0 in another, I don't want to make that pixel be 2.5 for the average value. I'd want each energy to be averaged individually, I think.

imap_processing/ena_maps/ena_maps.py
nkerman: That's fair! ~~Reading about suggested usage of kwargs, it seems like Python/PEP suggests using them when passing arguments to functions called within the main function.~~

Oh wow - my mistake, I was confusing using kwargs with keyword_only arguments. I think I now understand how to adjust this more neatly.

nkerman: Argh! Never mind, apparently the keyword-only arguments do still bother ruff/pylint. There's about 20 different interrelated GitHub issue tickets and threads about this. 
As best I can tell, they had the original error `"too-many-arguments / R0913"` which counted all types of arguments. Then one group decided that this should **only** count positional arguments, and added `"too-many-positional-arguments / R0917"`, and presumably wanted to fix the original to not count keyword-only arguments. But then another group decided that it was broken because `"too-many-arguments / R0913"` only counted positional arguments and added that back. 

All told, I advocate that we **always** ignore `"too-many-arguments / R0913"`, but still use `"too-many-positional-arguments / R0917"`. For now, I'm going to add back in the `# noqa` for `"too-many-arguments / R0913"` to this line.

imap_processing/ena_maps/ena_maps.py
nkerman: Linking this to the [above thread](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1596#discussion_r2047594722).

imap_processing/ena_maps/ena_maps.py
nkerman: Good to know - I'm reading up on hot loops.

nkerman:
> > SO the deal here is that a given pixel may have a bunch of different values: e.g., if there are 24 different energy bands, then the shape of a single pixel's data will actually be (1 epoch x 24 energies).
> 
> > The only place we want to calculate the mean of all these values (mean_pixel_value.mean()) would be in the tolerance check (I should also run this by Nick, but I think we want to compare a single value for each pixel to check its tolerance, not 24). Everywhere else, we want to return the full (1, 24) array (mean_pixel_value).
> 
> OK, I wasn't following that reading throught he code initially, can you add in a comment about the shapes / dimensions being passed around somewhere so that it is easier to track?
> 
> I don't understand why you would want to average over energy though, so I think I'm still missing that part. If you get 5 counts in one energy level and 0 in another, I don't want to make that pixel be 2.5 for the average value. I'd want each energy to be averaged individually, I think.

Sorry, I think I've done a poor job of explaining what I "mean" by "mean" (average) here, since there are 2 different types of means happening here. You are not, **for the purposes of calculating your output value**, taking the mean across multiple energy bands of a pixel. 

You are splitting up a pixel into subpix,
```
-----------------------------------------
|                   |                   |
|                   |                   |
|        V1         |        V2         |
| = shape(1, 24)    |                   |
|                   |                   |
|                   |                   |
|                   |                   |
|                   |                   |
-----------------------------------------
|                   |                   |
|                   |                   |
|                   |                   |
|        V3         |         V4        |
|                   |                   |
|                   |                   |
|                   |                   |
|                   |                   |
-----------------------------------------
```
getting the values (V1,2,3,4) of those subpix. If the quantity is exposure time, each of these might be a single float at each subpix, **but** for flux it might be `V1.shape=(1 epoch, 24 energies)`. 

Then you get the `mean_pixel_value` by taking the mean of these values, along only the pixel axis (Energy bins are handled separately). This `mean_pixel_value.shape=(1, 24)` still for flux. Then, **only** when we want to compare the `mean_pixel_value` at 7 subdivisions with `previous_mean_pixel_value` at 6 subdivisions, we can take their mean to compare a single number. Does that make sense?

imap_processing/ena_maps/ena_maps.py
greglucas: Sounds reasonable!

greglucas:
I think I'm getting closer to understanding, but still struggling because you are calling `mean(axis=-1)` within the loop which I think corresponds to your pixel axis comment above (makes sense), but then the comparison is done with `mean_pixel_value.mean()` without an axis. I think this is odd because it is now averaging over all of the energies. Perhaps that should `np.all(np.isclose(mean_pixel_value, prev_mean_pixel_value))` so you still go down the path further for a single energy bin that may need more subdividing. Or does this add too much computation and more unnecessary subdividing?

imap_processing/ena_maps/ena_maps.py
nkerman: This is a super interesting idea - my instinct says it won't be more efficient, since it means a **ton** of total subpixels being averaged (1.1 Billion subpixels, each of which can have several dimensions). 

As an order-of-magnitude test, I timed this at >30 seconds before it crashed my kernel:
```
np.repeat(np.arange(24, dtype=float)[:, np.newaxis], 1.1E9, axis=1).mean(axis=1)
```

Currently, when running the recursion method with higher dimensionality and with multiple value keys (flux, flux_unc, etc.) it takes about 15 seconds maximum. If you look at the recursion map I put in the description, we can see that it really doesn't need to do too much recursion for most pixels. 
These are fairly rarely run as well, (at a meeting yesterday, Ultra seemed to think it would take more like 20 minutes based on their code.)

nkerman:
> I think I'm getting closer to understanding, but still struggling because you are calling `mean(axis=-1)` within the loop which I think corresponds to your pixel axis comment above (makes sense), but then the comparison is done with `mean_pixel_value.mean()` without an axis. I think this is odd because it is now averaging over all of the energies. Perhaps that should `np.all(np.isclose(mean_pixel_value, prev_mean_pixel_value))` so you still go down the path further for a single energy bin that may need more subdividing. Or does this add too much computation and more unnecessary subdividing?

Especially WRT:
> Perhaps that should `np.all(np.isclose(mean_pixel_value, prev_mean_pixel_value))`

This is a good question, and ultimately one for the instrument team. I didn't want the comparison to have to have all different values of each pixel converge to the tolerance, and I think that the mean converging is a good enough check, but I have a comment on another thread that I want Nick to confirm this way is what he has in mind. I'll add a `TODO: ask Ultra Instrument team` specifically in the code for this, so it's not just hanging out in a thread.

Unfortunately Nick is out sick at the moment. I can make a ticket to keep track of this so it's not a blocker. 


---

I did some tests with the `np.all_close()` approach, and it increases runtime by a factor of 2 with my test data, from 8s -> 16s (but this is very gradient dependent). 

Screenshot 2025-04-17 at 1 51 57 PM


Screenshot 2025-04-17 at 1 51 11 PM

I still plan to ask Nick which one the instrument team needs. 2x isn't a dealbreaker, but it is slower.

imap_processing/ena_maps/ena_maps.py
nkerman: Darn, actually, I remembered why these should actually receive the self key, because the lowest-level method: `calculate_rect_pixel_value_from_healpix_map_n_subdivisions` needs to get `self.nside, self.nested`. 

We could pass those in separately, but I think it's neater that they have access to  properties of the map which calls them. I'll replace value_key with value_array though still. Sound reasonable?

imap_processing/ena_maps/ena_maps.py
nkerman: I don't think we can actually, since we need to iterate over pixel, and each pixel might have 1 or more dimensions. I might be missing an obvious way to do this though.

imap_processing/ena_maps/ena_maps.py
nkerman: > Just make sure they are DEBUG-level and we should only be running in INFO-level nominally and not spamming the logs. So they are there for your debugging purposes.

I added debugs for the pixel-level stuff and some infos for the full map outputs

Something vaguely like:
"""
Summary of subdivision depth for flux:
Mean +/- std number of subdivisions ...
    2.3 +/- 0.54321
Min / Max number of subdivisions:
    1 / 8
The maximum allowed depth is 8.
"""

```
            logger.info(
                f"Summary of subdivision depth for {value_key}:\n"
                "Mean +/- std number of subdivisions for the "
                f"{rect_map.num_points} pixels of {value_key} is:\n"
                f"    {np.mean(subdiv_depth_of_value_by_pixel):.6f}."
                f" +/- {np.std(subdiv_depth_of_value_by_pixel):.6f}.\n"
                "Min / Max number of subdivisions: \n"
                f"    {np.min(subdiv_depth_of_value_by_pixel):.6f} / "
                f"{np.max(subdiv_depth_of_value_by_pixel):.6f}.\n"
                f"The maximum allowed depth is {max_subdivision_depth}."
            )

```

nkerman: > I agree with you wanting to do this on a per-pixel basis, I think that makes sense too at first glance unless you wanted to make the large grid all at once like in one of my other comments. If these are tied together, you might want to recursively pass a tuple (value, depth) as your input combined together so that is sort of "the" quantity you want to pass around. 

Sorry - I'm not sure how that's different from the current situation? They are passed as a tuple right up until the point we need to separate value and depth.

nkerman: Okay, I think I found a _slightly_ more elegant way to do this! I replaced the 2 list comprehensions with a single `zip`.

```
            interpolated_data_by_rect_pixel, subdiv_depth_of_value_by_pixel = zip(
                *best_value_and_recursion_depth_by_pixel
            )
```

imap_processing/ena_maps/ena_maps.py
nkerman: Closing as duplicate.

imap_processing/ena_maps/ena_maps.py
greglucas: Yes, I was forgetting that piece too!

imap_processing/ena_maps/ena_maps.py
greglucas: >> I agree with you wanting to do this on a per-pixel basis, I think that makes sense too at first glance unless you wanted to make the large grid all at once like in one of my other comments. If these are tied together, you might want to recursively pass a tuple (value, depth) as your input combined together so that is sort of "the" quantity you want to pass around.

> Sorry - I'm not sure how that's different from the current situation? They are passed as a tuple right up until the point we need to separate value and depth.

Yeah, it is very similar, but what I was trying to suggest is a single tuple quantity that you would pass around and not unpack during each iteration, to make clear that they are intimately tied together. Call it a `namedtuple("ValueDepth", ["value", "depth"])` for example.

But I agree, this is basically the same as what you have already, so lets not overcomplicate this.

greglucas (approved):
Thanks for the time in addressing all of the comments and the extra investigations! I think this is in a good state and there are comments/follow-ups that can take place later.

One last thing that I'll leave here because it can easily be a follow-up small PR. Is there a way to say "the value I am projecting is empty in all energies" and then avoid any subdivisions. Right now, I think you have a minimum of 2 subdivisions because you need to get the first "previous_value" to then compare against the next iteration. I'm wondering if you can somehow avoid that double call somehow with some extra logic addition.

PR #1600, opened 2025-04-11
TST: Add pytest-xdist to run tests in parallel

greglucas:
Our test suite is getting pretty long to run, so lets try and add some parallelization to see if we can speed things up.

Make it opt-in but run the tests in parallel on the CI systems at least and allow users to locally use more processors.

I run with `pytest -n auto` locally and it started spinning the CPU fans but finished in under a minute.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,9 +46,9 @@ jobs:
         run: |
           # Ignore the network marks from the remote test environment
           if [ "$USE_EXTERNAL_DATA" = "true" ]; then
-            poetry run pytest --color=yes --cov --cov-report=xml
+            poetry run pytest -n auto --color=yes --cov --cov-report=xml
           else
-            poetry run pytest --color=yes --cov --cov-report=xml -m "not external_kernel and not external_test_data"
+            poetry run pytest -n auto --color=yes --cov --cov-report=xml -m "not external_kernel and not external_test_data"
           fi
 
       - name: Upload coverage reports to Codecov
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,7 @@ jobs:
           if [ "$USE_EXTERNAL_DATA" = "true" ]; then
             # Running across multiple processes causes errors due to large downloads not
             # finishing before the tests get hit in test_codice_l1a
-            poetry run pytest -n auto --dist loadscope --color=yes --cov --cov-report=xml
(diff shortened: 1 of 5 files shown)


greglucas:
Unfortunately it looks like the external test marks at the module level might be trying to download things, and then other tests in that same module get kicked off before the downloads finish. The remote data download probably needs some more thought and whether we can figure out how to wait to run all tests in that module until a file is finished downloading.

link to logs: https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/14402909184/job/40397345854#step:6:4112

tmplummer:
> Unfortunately it looks like the external test marks at the module level might be trying to download things, and then other tests in that same module get kicked off before the downloads finish. The remote data download probably needs some more thought and whether we can figure out how to wait to run all tests in that module until a file is finished downloading.

Did you try this option `--dist loadscope`?
https://pytest-xdist.readthedocs.io/en/latest/distribution.html

greglucas:
> Did you try this option --dist loadscope?
https://pytest-xdist.readthedocs.io/en/latest/distribution.html

Nope, worth a shot here! Thanks for that suggestion.

.github/workflows/test.yml
```diff
@@ -46,9 +46,11 @@ jobs:
         run: |
           # Ignore the network marks from the remote test environment
           if [ "$USE_EXTERNAL_DATA" = "true" ]; then
-            poetry run pytest --color=yes --cov --cov-report=xml
+            # Running across multiple processes causes errors due to large downloads not
+            # finishing before the tests get hit in test_codice_l1a
```
greglucas:
```suggestion
```

greglucas:
OK, this is working now and producing >2x speedups on the non-download test and a few minutes improvement on the download runner. Most importantly though, it should help local development if people want to run the entire test suite.

greglucas:
>> Unfortunately it looks like the external test marks at the module level might be trying to download things, and then other tests in that same module get kicked off before the downloads finish. The remote data download probably needs some more thought and whether we can figure out how to wait to run all tests in that module until a file is finished downloading.

> Did you try this option --dist loadscope?
https://pytest-xdist.readthedocs.io/en/latest/distribution.html

This unfortunately didn't work either, so it must be something with how the test/fixtures are collected and then evaluated. I insted opted to directly change the codice fixture to call the external_data gathering directly rather than with the indirection through the module-level mark. The mark is still important for skipping the tests.

vmartinez-cu (approved):
Nice work on this!

PR #1607, opened 2025-04-11
Fix build_solid_angle_map for 6 degree spacing

### LINKED ISSUE #1606: BUG - build_solid_angle_map fails for 6 degree spacing
### Description of the issue

Attempting to call build_solid_angle_map fails for 6 degree spacing. 6 degrees is an expected spacing for IMAP HI L2 and L3 maps, so this will be needed.

```
from imap_processing.ena_maps.utils.spatial_utils import build_solid_angle_map
build_solid_angle_map(6)
```

A 60 x 30 array should be returned.

```
  File "<input>", line 1, in <module>
  File "[...]\imap_processing\imap_processing\ena_maps\utils\spatial_utils.py", line 83, in build_solid_angle_map
    raise ValueError("Spacing must divide evenly into pi radians.")
ValueError: Spacing must divide evenly into pi radians.
```

We will submit a pull request with a suggested fix soon.

pleasant-menlo:
# Change Summary

## Overview
When represented as a float, 6 degrees in radians is slightly greater than pi/30, making (pi / 6 degrees) % 1 = 0.999... This made the "divide evenly" check fail.

Change the behavior to compute an integer number of bins rather than relying on rounding details.

Fixes #1606.

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
N/A

## Updated Files
- spatial_utils.py
  - Calculate the integer number of bins

## Testing
- Add 6 and 6.666666666666667 degrees (corresponding to 30 and 27 latitude bins) as valid options for tests to check.

### DIFF
--- a/imap_processing/ena_maps/utils/spatial_utils.py
+++ b/imap_processing/ena_maps/utils/spatial_utils.py
@@ -85,7 +85,9 @@ def build_solid_angle_map(
     if not np.isclose(proposed_number_of_lat_bins, number_of_lat_bins):
         raise ValueError("Spacing must divide evenly into pi radians.")
 
-    latitudes = np.linspace(-np.pi / 2, np.pi / 2, num=number_of_lat_bins+1, endpoint=True)
+    latitudes = np.linspace(
+        -np.pi / 2, np.pi / 2, num=number_of_lat_bins + 1, endpoint=True
+    )
     sine_latitudes = np.sin(latitudes)
     delta_sine_latitudes = np.diff(sine_latitudes)
     solid_angle_by_latitude = np.abs(spacing * delta_sine_latitudes)
--- a/imap_processing/ena_maps/utils/spatial_utils.py
+++ b/imap_processing/ena_maps/utils/spatial_utils.py
@@ -79,17 +79,20 @@ def build_solid_angle_map(
     if spacing <= 0:
         raise ValueError("Spacing must be positive valued, non-zero.")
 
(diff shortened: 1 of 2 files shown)


imap_processing/ena_maps/utils/spatial_utils.py
```diff
@@ -79,17 +79,22 @@ def build_solid_angle_map(
...
 
-    if not np.isclose((np.pi / spacing) % 1, 0):
+    proposed_number_of_lat_bins = 180 / spacing_deg
+    number_of_lat_bins = round(180 / spacing_deg)
+    number_of_lon_bins = 2 * number_of_lat_bins
+    if not np.isclose(proposed_number_of_lat_bins, number_of_lat_bins):
         raise ValueError("Spacing must divide evenly into pi radians.")
 
-    latitudes = np.arange(-np.pi / 2, np.pi / 2 + spacing, step=spacing)
+    latitudes = np.linspace(
```
nkerman: This is a flaw in the original code, but can we call this `latitude_edges` or similar? Latitudes on its own to me would default to meaning the latitudes of the pixel centers.

nkerman (approved):
LGTM! Thanks for catching and fixing the bug. I put one rename suggestion in.

imap_processing/ena_maps/utils/spatial_utils.py
pleasant-menlo: Thanks for the suggestion Nat. We have renamed this variable to be latitude_edges.

PR #1609, opened 2025-04-14
Lo L1B - Calculate TOF1 and Set Coincidence Type for Direct Events

### LINKED ISSUE #1513: Lo - Calculate TOF1 for Golden Triples
TOF1 is not transmitted for Golden Triple events, but can be calculated from the other fields using the equation:

tof0 + tof3 - tof2 - cksm - left_cksm_bound

All values in the equation must be right shifted. The TOFs and Checksum were left shifted in L1A to decompress the right shift done on board, so these must be right shifted again for this equation. All values in the equation must all be DN, not EU, so this must happen prior to the EU conversion of the TOFs. The left checksum bound will eventually be in ancillary data, but for now will use the value `-21`. 

This will only happen for Golden tripes (case number = 0, mode = 0).

### LINKED ISSUE #1514: Lo - Set the Coincidence type for L1B DE
Set the coincidence type as a 6 digit string of 0s and 1s, based on the following table:

Image

The format of the string will be `"<TOF0><TOF1><TOF2><TOF3><CkSm><Mode>"`. If the case number has a 1 in the table then the string character will be `1`. If the table has a `-` then the string character will be `0`. The mode character will just be the value of the mode field which is either a `0` or `1`.

sdhoyt:
# Change Summary

## Overview
This PR calculates the TOF1 value for golden triples and sets the coincidence type for each direct event

## Updated Files
   - added function to calculate TOF1 for golden triples
   - added function to set the coincidence type for each direct event

## Testing
Added unit tests and ran CLI

Closes #1513
Closes #1514


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -59,6 +59,11 @@ def lo_l1b(dependencies: dict) -> list[Path]:
         l1b_de = set_event_met(l1a_de, l1b_de, spin_start_time, avg_spin_durations)
         # set the epoch for each event
         l1b_de = set_each_event_epoch(l1b_de)
+        # calculate the TOF1 for golden triples
+        # store in the l1a dataset to use in l1b calculations
+        l1a_de = calculate_tof1_for_golden_triples(l1a_de)
```
greglucas, line 64: Do you not want to / need to store the TOF1 value you calculate within this function? Right now, this is getting dropped because the l1a_de produced here doesn't get returned later. If you don't need it, would it make sense to push this call into the `set_coincidence_type()` where it is used?

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -385,6 +389,100 @@ def set_each_event_epoch(l1b_de: xr.Dataset) -> xr.Dataset:
...
+    attr_mgr_l1a : ImapCdfAttributes
+        Attribute manager used to get the fill values for the L1A DE dataset.
+
+    Returns
+    -------
+    l1b_de : xarray.Dataset
+        The L1B DE dataset with the coincidence type added.
+    """
+    tof0_fill = attr_mgr_l1a.get_variable_attributes("tof0")["FILLVAL"]
+    tof0_mask = l1a_de["tof0"].values != tof0_fill
```
greglucas, line 461: To make the the string formatting below a little nicer I think you could cast to int within these blocks instead.
```suggestion
    tof0_mask = (l1a_de["tof0"].values != tof0_fill).astype(int)
```

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -385,6 +389,100 @@ def set_each_event_epoch(l1b_de: xr.Dataset) -> xr.Dataset:
...
+    tof2_mask = l1a_de["tof2"].values != tof2_fill
+    tof3_fill = attr_mgr_l1a.get_variable_attributes("tof3")["FILLVAL"]
+    tof3_mask = l1a_de["tof3"].values != tof3_fill
+    cksm_fill = attr_mgr_l1a.get_variable_attributes("cksm")["FILLVAL"]
+    cksm_mask = l1a_de["cksm"].values != cksm_fill
+
+    coincidence_type = [
+        f"{int(tof0_mask[i])}{int(tof1_mask[i])}{int(tof2_mask[i])}{int(tof3_mask[i])}{int(cksm_mask[i])}{l1a_de['mode'].values[i]}"
+        for i in range(l1a_de["de_count"].values.sum())
+    ]
```
greglucas, line 474: Is this the same as `range(len(l1b_de["epoch"]))`? It would be nice to use a dimension if you have it.

greglucas (approved):
Looks good. (Thank you for breaking these into such manageable chunks for review!) A few minor suggestions that don't impact the overall results.

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: I'll be creating a function in the next PR that converts the TOF values in l1a_de to EU, so the l1a data will get used again there

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: That worked, thanks!

imap_processing/lo/l1b/lo_l1b.py
greglucas: :+1: Do you plan on saving these as part of the l1 dataset somewhere then as an export? Or are these purely just intermediary results that we don't care to have as output.

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: No, but it's the same as `range(len(l1b_de["direct_events"])` which is a coordinate, so I'll switch it to that.

greglucas: Perfect, that sounds like a great idea to iterate over ranges of coordinates! But then doesn't this mean that below you shouldn't be adding it to the "epoch" as the coordinate dim but rather the "direct_events" coordinate?

sdhoyt: The coordinate for the direct events changes from a direct event index in L1A to epoch in L1B because the epoch time for each direct event isn't calculated until L1B. So this is part of the switch over from using an index to using epoch for the direct event data.

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: This is just an intermediary result that will be used to calculate the TOF EU data to put into the L1B dataset. The L1A TOF data needs to be used to calculate the TOF1s for golden triples before the TOFs can be converted to EU, so I did that in the L1A dataset. I could copy the TOFs from L1A to L1B and then run the TOF1 calc on the L1B dataset instead if you think that would be better.

greglucas: I think this is up to you and whether you think that might help debug things later or if it helps start processing at l1b instead of l1a and things like that. What you have is fine, it just stuck out to me as being an intermediate that isn't saved currently.

PR #1610, opened 2025-04-14
MAG L1C updates for i-alirt

maxinelasp:
# Change Summary
Moving one piece of logic into its own function so I-ALiRT can reuse it.


imap_processing/mag/l1b/mag_l1b.py
```diff
@@ -209,6 +211,37 @@ def mag_l1b_processing(
     return output_dataset
 
 
+def retrieve_matrix_from_l1b_calibration(
```
maxinelasp, line 214: We expect I-ALiRT to use the same file or a very similar looking file for processing.

laspsandoval (approved):
Thanks for doing this!

PR #1612, opened 2025-04-14
Lo L1B - Convert TOF Data Numbers to Engineering Units

### LINKED ISSUE #1515: Lo - Convert TOF DNs to EU for L1B DE
Convert the TOFs from Data Numbers (DN) to Engineering Units (EU) in the L1B DE product. This conversion will use the equation:

TOF_EU = C0 + C1 * TOF_DN

Where C0 and C1 are the following:

Image

(The conversion sheet sent has a factor of 2 multiplied by C1, but this is a built-in left shift. Because the left shift has already been performed during the unpacking and decompression, this factor of 2 should not be used in the equation for L1B processing)

sdhoyt:
# Change Summary

## Overview
This PR adds code to convert the Lo L1A direct event DN TOFs to L1B EU TOF values

I also forgot to push changes in response to comments from my last PR for the coincidence type function so those minor changes are added in this PR as well

## New Files
- imap_processing/lo/l1b/tof_conversions.py
   - stores TOF conversion factors

## Updated Files
   - added function to calculate EU TOFs
 
## Testing
Added new unit test and ran CLI

 Closes #1515


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -483,6 +491,66 @@ def set_coincidence_type(
...
+        # Create a mask for the TOF
+        mask = l1a_de[tof] != fillval_1a
+        # convert the DE TOF to engineering units
+        tof_eu = np.where(
+            mask,
+            conv.C0 + conv.C1 * l1a_de[tof],
+            fillval_1b,
+        )
+        # Add the EU TOF to the dataset. If the TOF is not present, set it to the fill
+        # value for the L1B TOF data.
```
greglucas: The second sentence should be above the where statement and merged into your other comment up there I believe since you are doing that work within where, not the dataset creation.

imap_processing/lo/l1b/tof_conversions.py
```diff
@@ -0,0 +1,9 @@
+"""Lo TOF EU Conversions."""
+
+from collections import namedtuple
+
+tof_conv = namedtuple("tof_conv", ["C0", "C1"])
```
greglucas, line 5: Add a comment about where these values come from.

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -397,14 +398,18 @@ def test_set_coincidence_type(attr_mgr_l1a):
...
             "de_count": ("epoch", [3]),
-            "coincidence_type": ("direct_event", [0, 0, 4]),
-            "mode": ("direct_event", [1, 0, 1]),
-            "tof0": ("direct_event", [5, 2, 10]),
-            "tof1": ("direct_event", [10, 4, tof_fill]),
-            "tof2": ("direct_event", [15, 6, 20]),
-            "tof3": ("direct_event", [20, 8, 30]),
-            "cksm": ("direct_event", [25, ckm_fill, ckm_fill]),
-        }
+            "coincidence_type": ("direct_events", [0, 0, 4]),
```
greglucas, line 401: I was wondering what changed here when first looking at this lol, took me a bit to figure out it was the plural.

Do you want to have this pluralized "events" versus "event" before? I think the previous version might be more standard to align with "epoch" being singular, and "time" being singular.

greglucas (approved):
Looks great! Minor comments around naming again.

greglucas:
It looks like you are working off a common branch on your local system or something and these are getting to have 37+ commits for a small update like this. Could you rebase those old commits out to make the commit history smaller during review?

imap_processing/tests/lo/test_lo_l1b.py
sdhoyt: I agree. I changed it here to match what I have in the yaml and L1A code, but I'll create an issue to change L1A to use make it singular.

#1613

PR #1615, opened 2025-04-14
Lo L1B - Species Identification and Badtimes Initialization 

### LINKED ISSUE #1517: Lo - L1B DE Species Identification
Perform Hydrogen/Oxygen/Unknown species identification using the following equation and tables:

Image

Select which table to used based on the U_PAC value from housekeeping. Then use both the TOF2 and TOF0S equations to determine which species an event falls in. If it does not fall into a range for a single species in both tables, then it is labeled as Unknown.

sdhoyt:
# Change Summary

## Overview
This PR adds code to identify the species of each Lo L1B Direct Event as either Hydrogen, Oxygen, or Unknown. The badtimes field is also initialized.

## Updated Files
   - added function to identify the direct event species
   - added function to initialize badtimes (always not bad for now)

## Testing
Added two unit tests and ran CLI

Closes #1517

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -72,6 +72,11 @@ def lo_l1b(dependencies: dict) -> list[Path]:
         l1b_de = set_coincidence_type(l1a_de, l1b_de, attr_mgr_l1a)
         # convert the TOFs to engineering units
         l1b_de = convert_tofs_to_eu(l1a_de, l1b_de, attr_mgr_l1a, attr_mgr_l1b)
+        # set the species for each direct event
+        l1b_de = identify_species(l1b_de)
+        # set the badtimes
+        l1b_de = set_bad_times(l1b_de)
+        # TODO: set the direction for each direct event
 
     return [l1b_de]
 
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -550,6 +555,83 @@ def convert_tofs_to_eu(
...
+    ----------
+    l1b_de : xarray.Dataset
+        The L1B DE dataset.
+
+    Returns
+    -------
+    l1b_de : xarray.Dataset
+        The L1B DE dataset with the bad times added.
+    """
+    # Initialize all times as not bad for now
```
greglucas, line 623: Add a note/todo about coming back to this?

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -465,3 +467,39 @@ def test_convert_tofs_to_eu(attr_mgr_l1b, attr_mgr_l1a):
...
+def test_identify_species(attr_mgr_l1b):
+    # Arrange
+    fill_val = attr_mgr_l1b.get_variable_attributes("tof2")["FILLVAL"]
+    l1b_de = xr.Dataset(
+        {
+            "tof2": ("epoch", [14, 80, 500, fill_val]),
+        }
+    )
+
+    expected_species = np.array(["H", "O", "U", "U"])
```
greglucas: Add a value under and between too.
```suggestion
            "tof2": ("epoch", [1, 14, 50, 80, 500, fill_val]),
        }
    )

    expected_species = np.array(["U", "H", "U", "O", "U", "U"])
```

greglucas (approved):

PR #1616, opened 2025-04-14
Mag L2 Calibration and offsets

maxinelasp:
# Change Summary
Does the first draft of calibration and offsets work for MAG. Although there is more to do around input files, this is the core of L2. 

## Overview

Finished methods for applying calibration and offsets. This code could produce a valid output L2 file, as long as the test offset file matched the input L1B file. Future work is to fully integrate the offset file so we can run this in production. 

As I have gone, I have kept in mind that L1D is expected to use some of these same functions and look the same, and I-ALiRT will use a subset of the methods but not look the same. That's why some of the design decisions might seem odd at first glance. 

## Updated Files

Filled in some missing pieces in the MAG L2 outline. 

## Testing

Added some basic tests, more to come soon (including validation)


imap_processing/cli.py
```diff
@@ -851,10 +851,18 @@ def do_processing(
                 / "mag"
                 / "validation"
                 / "calibration"
-                / "imap_mag_l1b-calibration_20240229_v001.cdf"
+                / "imap_mag_l2-calibration-matrices_20251017_v004.cdf"
```
nkerman, line 855: If a file is passed in like this, will it get automatically stored and tracked in the `attrs`?

imap_processing/mag/imap_mag_sdc-configuration_v001.yaml
```diff
@@ -4,3 +4,7 @@
 # interpolation method used in L1C processing
 L1C_interpolation_method: "linear_filtered"
 
+# Setting for L2 output
+# if this setting is updated, the dependency settings for L2 will also need to change.
+always_output_mago: True
```
nkerman: I read this as mango for some reason! 🥭 😆

imap_processing/mag/l2/mag_l2.py
```diff
@@ -36,39 +44,73 @@ def mag_l2(
     """
     # TODO we may need to combine multiple calibration datasets into one timeline.
 
+    with open(
+        Path(__file__).parent.parent / "imap_mag_sdc-configuration_v001.yaml"
+    ) as f:
+        configuration = yaml.safe_load(f)
+
+    always_output_mago = bool(configuration["always_output_mago"])
```
nkerman: Could you instead use a `imap_mag_sdc-configuration_v001` `.py` which contains an importable dictionary of properties? 

I lean toward that approach here, because it doesn't require additional reading-in or type-switching you need when using JSON/YAML, etc.

imap_processing/mag/l2/mag_l2.py
```diff
@@ -36,39 +44,73 @@ def mag_l2(
...
+
+    # TODO Check that the input file matches the offsets file
```
nkerman, line 73: Is the todo to check more than just the timestamps? If so, and these datasets need to be exactly equal, could you add a comment on the difference between them?

imap_processing/mag/l2/mag_l2.py
```diff
@@ -36,39 +44,73 @@ def mag_l2(
...
+    if not np.array_equal(input_data["epoch"].data, offsets_dataset["epoch"].data):
+        raise ValueError("Input file and offsets file must have the same timestamps.")
+
+    calibration_matrix = retrieve_matrix_from_l2_calibration(
+        calibration_datasets, always_output_mago
+    )
+
+    vectors = np.apply_along_axis(
```
nkerman, line 81: nitpick: do you mind adding the parameters here? I find it a bit hard to read apply along axis in the best of times 😁 , and there's additional args being passed in here.

imap_processing/mag/l2/mag_l2.py
```diff
@@ -36,39 +44,73 @@ def mag_l2(
...
     Returns
     -------
     np.ndarray
-        Rotated and timeshifted vectors.
+        Calibration matrix in the shape (3, 3, 4) to rotate vectors.
     """
-    raise NotImplementedError
+    # TODO: allow for multiple inputs
+    if isinstance(calibration_datasets, list):
+        calibration_dataset = calibration_datasets[0]
```
nkerman, line 126: Maybe raise `NotImplementedError` for now?

imap_processing/mag/l2/mag_l2.py
```diff
@@ -36,39 +44,73 @@ def mag_l2(
     """
     # TODO we may need to combine multiple calibration datasets into one timeline.
 
+    with open(
```
laspsandoval: Please make a comment on all imports and what they are intended for. I think that will help me understand what is happening. For example, I have written from my  notes for l1d that we will remove the magnetic field of the spacecraft (in the sensor frame) using a calibration file. Is that what we are doing here? I might be way off. Just assume I know nothing. Which is basically the case.

imap_processing/mag/l2/mag_l2_data.py
```diff
@@ -84,6 +104,65 @@ def calculate_magnitude(
...
+        offset_vectors: np.ndarray = vectors[:, :3] + offsets
+
+        # TODO: CDF files don't have NaNs. Emailed MAG to ask what this will look like.
+        # Any values where offsets is nan must also be nan
+        offset_vectors[np.isnan(offsets).any(axis=1)] = np.nan
+
+        return offset_vectors
+
+    @staticmethod
+    def shift_timestamps(epoch: np.ndarray, timedelta: np.ndarray) -> np.ndarray:
```
laspsandoval, line 142: In l1b we shifted the timestamps using the calibration file. Why do we need to shift the time again?

imap_processing/mag/l2/mag_l2_data.py
```diff
@@ -84,6 +104,65 @@ def calculate_magnitude(
         """
         return np.zeros(vectors.shape[0])  # type: ignore
 
+    @staticmethod
+    def apply_offsets(vectors: np.ndarray, offsets: np.ndarray) -> np.ndarray:
+        """
+        Apply the offsets to the vectors by adding them together.
```
laspsandoval, line 110: Are these the offsets by the gradiometer? More details please :) Something like:
Apply offsets from gradiometer created by differences between 2 sensors in different locations on spacecraft inboard vs. outboard.

That might be slightly wrong, but you get the idea.

laspsandoval (commented):
I basically just have a lot of questions and requests for comments. Otherwise it looks good. Nice job.

imap_processing/mag/l2/mag_l2_data.py
maxinelasp: Basically, this is a hand-tuned vector by vector time shift, whereas the calibration file in L1B was more of a "general" timeshift. All these shifts are why we include a 30 minute buffer, so we should always have the right data in the file.

imap_processing/mag/l2/mag_l2.py
maxinelasp: Will do! I can try to clarify more on the difference between the offsets file, the calibration file, and the settings file, because I agree that's confusing.

imap_processing/mag/l2/mag_l2_data.py
maxinelasp: The problem is, the offsets could be from the hand-created file in L2, or from the gradiometry calculations for L1D or I-ALiRT. Ideally, this function will be used for both options. I can add more notes about it.

nkerman (commented):
Looks great! A couple small requests (mainly a bit of documentation, and considering a different format of the config file._

imap_processing/cli.py
maxinelasp: This is a temporary way of doing things, but I don't think this will be tracked in the attrs as it currently is. We expect the normal method of receiving files in prod though.

imap_processing/mag/l2/mag_l2.py
maxinelasp: Yeah, this is a good point. I will make that change.

imap_processing/mag/l2/mag_l2.py
maxinelasp: the expected input file should match an attribute in the offsets file. The datasets are not equal, but all the timestamps should be equal.

imap_processing/mag/l2/mag_l2.py
maxinelasp: Yeah that's fair 😅

imap_processing/mag/l2/mag_l2.py
maxinelasp: I can raise an error for a len > 1!

imap_processing/mag/imap_mag_sdc_configuration_v001.py
```diff
@@ -0,0 +1,8 @@
+"""File describing configuration changes that are not expected to change often."""
```
nkerman, line 1: Nice! I like this as a Python file.

imap_processing/mag/l2/mag_l2.py
```diff
@@ -4,26 +4,58 @@
...
-    calibration_dataset: xr.Dataset,
-    offset_dataset: xr.Dataset,
+    calibration_datasets: list[xr.Dataset],
+    offsets_dataset: xr.Dataset,
     input_data: xr.Dataset,
 ) -> list[xr.Dataset]:
     """
     Complete MAG L2 processing.
 
+    Processing uses 4 data input sources:
```
nkerman, line 21: Super helpful docstring! 

One small thing to fix: Below, you refer to the configuration yaml, where I think it should now be the python file.

nkerman (approved):
LGTM, thanks Maxine!

There's one typo in the docstring to fix.

PR #1619, opened 2025-04-15
HIT L2 - add algorithm for sectored intensity

### LINKED ISSUE #1074: HIT L2 - write sectored fluxes algorithm code
Write Sectored Fluxes Algorithm Code

L1B divided by 3 constants from table 31 (delta E, Geometry factor, efficiency). That table is for all ranges?

vmartinez-cu:
This PR implements the algorithm to convert l1b sectored rates to l2 sectored intensities (i.e. flux). A function was added to process sectored data and some existing functions were updated to handle this additional data product.

Closes #1074 

pre-commit.ci autofix

## Updated Files
   - minor updates to l1b logical sources
   - rename L2 logical source for sectored intensities to macropixel intensity since project leadership prefers not having 
     "sector" in L2 file names

- hit/l2/constants.py

   - add constants needed for L2 processing
- hit_l2.py
   - Add a function to process sectored data
   - Update `calculate_intensities_for_a_species()` to use vectorized operations
   - Add seconds to the `IntensityFactors` namedTuple since it could either be 60 or 600
   - Rename systematic uncertainty variables to <variable>_sys_err_minus and <variable>_sys_err_plus
   - Move list of valid data variables to process to the constants file
   - Updates to docstrings
   - Add handling for the sectored dataset in `hit_l2()`

## Testing
Add test for new function and update tests for modified functions


imap_processing/hit/l2/constants.py
```diff
@@ -8,6 +8,40 @@
 L2_SUMMED_ANCILLARY_PATH_PREFIX = (
     imap_module_directory / "hit/ancillary/imap_hit_l1b-to-l2-summed-dt"
 )
+L2_SECTORED_ANCILLARY_PATH_PREFIX = (
+    imap_module_directory / "hit/ancillary/imap_hit_l1b-to-l2-sector-dt"
+)
```
greglucas, line 13: As an FYI, these names are all slightly inconsistent. Your variable name is "sectored", the file name is "sector" and now the data product is "macro-pixel". Just making sure you are all aware of this and you don't anticipate needing to change these later.

imap_processing/hit/l2/hit_l2.py
```diff
@@ -102,10 +112,11 @@ def hit_l2(dependency: xr.Dataset) -> list[xr.Dataset]:
 class IntensityFactors(NamedTuple):
     """A namedtuple to store factors for the intensity equation."""
 
-    delta_e_factor: np.ndarray
+    delta_e: np.ndarray
     geometry_factor: np.ndarray
     efficiency: np.ndarray
     b: np.ndarray
+    seconds: int
```
greglucas: seconds is the unit, not the variable, so I'd just making this more descriptive with "time" and then adding a comment about it needing to be in units of seconds.
```suggestion
    integration_time: int
```

imap_processing/hit/l2/hit_l2.py
```diff
@@ -137,49 +148,63 @@ def get_intensity_factors(
...
+    Calculate the intensities for given arrays of rates and equation factors.
 
     Uses vectorization to calculate the intensities for an array of rates
-    at an epoch.
+    for all epochs.
 
-        This function uses equation 9 and 11 from the HIT algorithm document:
-        (Summed L1B Rates) / (60 * Delta E * Geometry Factor * Efficiency) - b
+        This function uses equation 9 and 12 from the HIT algorithm document:
+        ((Summed L1B Rates) / (Seconds * Delta E * Geometry Factor * Efficiency)) - b
```
greglucas:
```suggestion
        ((Summed L1B Rates) / (Delta time * Delta E * Geometry Factor * Efficiency)) - b
```

imap_processing/hit/l2/hit_l2.py
```diff
@@ -137,49 +148,63 @@ def get_intensity_factors(
...
-        The energy bin width factors for an epoch.
-    geometry_factor : np.ndarray
-        The geometry factors for an epoch.
-    efficiency : np.ndarray
-        The efficiency factors for an epoch.
-    b : np.ndarray
-        The b factors for an epoch.
+    rates : xr.DataArray
+        The L1B rates to be converted to intensities.
+    factors : IntensityFactors
```
greglucas: I don't love that this is now a named_tuple, is there a need to combine these factors together because you just unpack them again later. I feel like it'd be cleaner to just pass through each individual factor rather than having to maintain an intermediate "factors" tuple.

imap_processing/hit/l2/hit_l2.py
```diff
@@ -188,67 +213,111 @@ def calculate_intensities_for_a_species(
...
-        rates: xr.DataArray = updated_ds[species_variable][epoch]
-
-        updated_ds[species_variable][epoch] = calculate_intensities(
-            rates,
-            factors.delta_e_factor,
-            factors.geometry_factor,
-            factors.efficiency,
-            factors.b,
+    # Get the dynamic threshold state for all epochs (one per epoch)
+    dynamic_threshold_states = updated_ds["dynamic_threshold_state"].values.astype(int)
```
greglucas: Should this be saved/loaded as an integer datatype initially so you don't have to cast here?

imap_processing/hit/l2/hit_l2.py
```diff
@@ -188,67 +213,111 @@ def calculate_intensities_for_a_species(
...
+    }
+
+    # Retrieve intensity calculation factors from ancillary data for all epochs and
+    # energy bins. This will be a list of IntensityFactors named tuples for each epoch
+    factors_per_epoch = [
+        get_intensity_factors(
+            species_energy_bins, species_ancillary_data_by_state[state]
         )
+        for state in dynamic_threshold_states
+    ]
```
greglucas: I'm not entirely clear what is going on here because you are calling it factors_per_epoch, but you are looping over dynamic_threshold_states.

Then below, you are stacking all of these factors individually. Are you able to turn these factors into a dataset right away if they are meant to go along with each epoch in your dataset? Instead of the named_tuple that you are then unpacking here.

imap_processing/hit/l2/hit_l2.py
```diff
@@ -188,67 +213,111 @@ def calculate_intensities_for_a_species(
...
+        [factor.geometry_factor for factor in factors_per_epoch]
+    )
+    efficiencies = np.stack([factor.efficiency for factor in factors_per_epoch])
+    b = np.stack([factor.b for factor in factors_per_epoch])
+    seconds = SECONDS_PER_MIN
+
+    # Handle sectored rates which are multidimensional
+    # (epoch, energy, azimuth, declination)
+    if "declination" in updated_ds[species_variable].dims:
+        # The factors are 1D arrays containing values for each declination (8) and each
```
greglucas: Make a constant for `N_DECLINATION = 8`?

imap_processing/hit/l2/hit_l2.py
```diff
@@ -446,15 +503,14 @@ def process_summed_intensity_data(l1b_summed_rates_dataset: xr.Dataset) -> xr.Da
     # Add systematic uncertainties to the dataset. These will not
     # have the intensity calculation applied to them
     for var in l2_summed_intensity_dataset.data_vars:
-        if "_" not in var:
-            particle = str(var)
+        if var in VALID_SPECIES:
             l2_summed_intensity_dataset = add_systematic_uncertainties(
                 l2_summed_intensity_dataset,
-                particle,
+                str(var),
```
greglucas: Your VALID_SPECIES is only strings, so you are guaranteed a string here already.
```suggestion
                var,
```

imap_processing/hit/l2/hit_l2.py
```diff
@@ -531,8 +584,60 @@ def process_standard_intensity_data(
...
+        L2_SECTORED_ANCILLARY_PATH_PREFIX,
+    )
+
+    # Add systematic uncertainties to the dataset. These will not
+    # have the intensity calculation applied to them
+    for var in l2_sectored_intensity_dataset.data_vars:
+        if var in VALID_SECTORED_SPECIES:
+            l2_sectored_intensity_dataset = add_systematic_uncertainties(
+                l2_sectored_intensity_dataset,
+                str(var),
```
greglucas:
```suggestion
                var,
```

imap_processing/tests/hit/test_hit_l2.py
```diff
@@ -289,27 +298,63 @@ def test_calculate_intensities_for_a_species():
...
-    # Assertions
-    assert np.allclose(intensities.values, expected_intensities.values), (
-        "Intensities mismatch"
+    # Call the function for seconds = 60
+    intensities_60 = calculate_intensities(rate, factors_60)
+
+    # Assertions for seconds = 60
+    assert np.allclose(intensities_60.values, expected_intensities_60.values), (
+        "Intensities mismatch for seconds = 60"
+    )
```
greglucas: Use the numpy testing assertions since they add helpful output about which quantities did not compare equal. Here and below.
```suggestion
    np.testing.assert_allclose(intensities_60, expected_intensities_60)
```

greglucas (commented):
Is L1b also going to change name to "macro-pixel" or is that only at L2?

imap_processing/hit/l2/constants.py
vmartinez-cu: Yes, I've noticed this too. The algorithm document uses both "sector" and "sectored". I can check with HIT on their preference for the ancillary file. As for macro-pixel, this is only for L2. This is HIT's solution to leadership not wanting "sector" being used in L2 file names. However, sector/ed is how HIT talks about the data

imap_processing/hit/l2/hit_l2.py
vmartinez-cu: Good suggestion!

imap_processing/hit/l2/hit_l2.py
vmartinez-cu: I originally passed in each factor, but ended up having too many arguments for one of the PEP standards (6>5) and so it didn't pass our pre-commit checks. I don't love it either, but wasn't sure how best to get all the information in otherwise. I could multiply the denominator factors and pass in the result, but I prefer to have each component explicitly shown in the calculation to make it clear what's going on

sdhoyt (approved):
Looks great! Congrats on finishing your last product!

imap_processing/hit/l2/hit_l2.py
vmartinez-cu: I struggled with this part of the workflow. I was able to implement your suggestion and create a dataset of the ancillary data aligned with epoch. I think it's easier to follow now

vmartinez-cu:
pre-commit.ci autofix

PR #1620, opened 2025-04-16
SWE: lut update and refactor of L1B and L2

### LINKED ISSUE #1598: SWE: Refactor SWE L1B to work with new LUT

tech3371:
# Change Summary
closes #1598 

## Overview
Sorry in advance for big PR this time. It started out small but effected many places and had to do them together.

This PR refactors L1B and L2 code to work with new LUT received from SWE. Refactor was done to accommodate for future changes and new additions.

## Deleted Files
- imap_processing/swe/l1b/swe_esa_lookup_table.csv
- imap_processing/tests/swe/test_swe_l1b_science.py

## Updated Files
- imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml
   - Adding new variable in L1B for L2 purposes
- imap_processing/ialirt/l0/process_swe.py
   - Updated I-ALiRT code to read in-flight cal data with new changes
- imap_processing/cli.py
   - update L1B cli code to pass new parameters
- imap_processing/swe/l1a/swe_science.py
   - adding check schema for dimension variable
- imap_processing/swe/l1b/swe_l1b.py
   - **MAIN changes of this PR. Refactoring of code. This is what was swe_l1b_science.py. Renamed to simplify**
   - Changes to L2 based on L1B changes.
- imap_processing/swe/utils/swe_constants.py
   - Adding minor constant

## Testing
- Added some LUT for unit testings
- Update L1B and L2 tests


imap_processing/cli.py
```diff
@@ -1016,16 +1016,8 @@ def do_processing(
...
-            l1a_dataset = load_cdf(l1a_file_path)
-            # TODO: fix this to read multiple files as needed.
-            # TODO: pass pandas dataframe to swe_l1b
-            # in_flight_cal_path = dependencies.get_file_paths(
-            #     "swe", "l1b-in-flight-cal"
-            # )[0]
-            # TODO: read lookup table and in-flight calibration data here.
-            datasets = swe_l1b(l1a_dataset)
+
+            datasets = swe_l1b_science(dependencies.serialize())
```
greglucas: You might be able to ignore the documentation issues from imap-data-access.* within the nitpick ignore cross references of `conf.py`
```suggestion
            datasets = swe_l1b_science(dependencies)
```

imap_processing/tests/swe/lut/imap_swe_l1b-in-flight-cal_20240510_20260716_v000.csv
```diff
@@ -0,0 +1,3 @@
+met_time,cem1,cem2,cem3,cem4,cem5,cem6,cem7
+453050308,1,1,1,1,1,1,1
+1782864000,1,1,1,1,1,1,1
```
greglucas, line 3: Not needed now, but something to consider is making these two different values so that you can linearly interpolate and make sure you are doing that step correctly as well, including how they are applied to each CEM. Just something to consider for a future PR.
```suggestion
453050308,1,1,1,1,1,1,1
1782864000,0.4,0.5,0.6,0.7,0.8,0.9,1

```

imap_processing/tests/swe/test_swe_l1b_science.py
```diff
@@ -37,22 +45,8 @@ def test_get_full_cycle_data_indices():
...
-            "cem3": [1, 2],
-            "cem4": [1, 2],
-            "cem5": [1, 2],
-            "cem6": [1, 2],
-            "cem7": [1, 2],
-        }
-    ),
-)
-def test_in_flight_calibration_factor(mock_read_in_flight_cal_data, l1a_test_data):
+@pytest.mark.skip(reason="Need to fix it")
```
greglucas: Open an Issue for this? It looks like this hasn't changed from your included test file, so do we need to be concerned that you're skipping it here?

imap_processing/tests/swe/test_swe_l1b_science.py
```diff
@@ -100,4 +98,139 @@ def test_in_flight_calibration_factor(mock_read_in_flight_cal_data, l1a_test_dat
...
+    expected_checkerboard = pd.read_csv(
+        imap_module_directory / "tests/swe/lut/checker-board-indices.csv", header=None
+    ).values
+    esa_lut_file = (
+        imap_module_directory / "tests/swe/lut/imap_swe_esa-lut_20250301_v000.csv"
+    )
+    checkerboard_pattern = get_checker_board_pattern(esa_lut_file)
+    assert np.all(checkerboard_pattern == expected_checkerboard), (
+        "Checkerboard pattern is not same as the one in LUT file."
+    )
```
greglucas: np.testing.assert_all_equal()

imap_processing/tests/swe/test_swe_l2.py
```diff
@@ -298,33 +247,62 @@ def test_put_data_into_angle_bins():
...
+        {
+            "type": "ancillary",
+            "files": [
+                "imap_swe_eu-conversion_20240510_v000.csv",
+            ],
+        },
+    ]
+    l1b_dataset = swe_l1b_science(json.dumps(dependencies))[0]
+    l1b_dataset.attrs["Data_version"] = "v000"
+    print(l1b_dataset)
```
greglucas:
```suggestion
```

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,928 @@
...
+    corrected_count : numpy.ndarray
+        Corrected counts.
+    """
+    # deadtime is 360 ns
+    deadtime = 360e-9
+    if isinstance(acq_duration, int):
+        # Convert acq_duration to a numpy array for consistency
+        acq_duration = np.array([acq_duration])
+    correct = 1.0 - (deadtime * (counts / (acq_duration[..., np.newaxis] * 1e-6)))
+    correct = np.maximum(0.1, correct)
```
greglucas, line 99: Where does 0.1 come from?

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,928 @@
...
+    corrected_count = np.divide(counts, correct)
```
greglucas: I think you might want to bring the below astype() up into this to force counts to be float64 before the division? Same for the below function.
```suggestion
    corrected_count = counts.astype(np.float64) / correct
```

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,928 @@
...
+    in_flight_cal_df.columns = [
+        "met_time",
+        "cem1",
+        "cem2",
+        "cem3",
+        "cem4",
+        "cem5",
+        "cem6",
+        "cem7",
+    ]
```
greglucas: Get these column names from the header instead of hard coding them here?

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,928 @@
...
+        acquisition_times.min() < cal_times.min()
+        or acquisition_times.max() > cal_times.max()
+    ):
+        error_msg = (
+            f"Acquisition min/max times: {acquisition_times.min()} to "
+            f"{acquisition_times.max()}. "
+            f"Calibration min/max times: {cal_times.min()} to {cal_times.max()}. "
+            "Acquisition times should be within calibration time range."
+        )
+        raise ValueError(error_msg)
```
```suggestion
        raise ValueError(
            f"Acquisition min/max times: {acquisition_times.min()} to "
            f"{acquisition_times.max()}. "
            f"Calibration min/max times: {cal_times.min()} to {cal_times.max()}. "
            "Acquisition times should be within calibration time range."
        )
```

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,928 @@
...
+        # Starting ESA step value for this quarter cycle. Eg.
+        # 0, 180, 360, 540
+        start_esa_step = esa_step[i][0]
+        # Populate the first two columns of the checkerboard pattern
+        # using the start_esa_step value. Eg.
+        #  Even row indices: 0, 1, 2, 3, 4, 5
+        #  Odd row indices: 6, 7, 8, 9, 10, 11
+        first_two_columns[even_row_indices, 0] = np.arange(
+            start_esa_step, start_esa_step + 6
+        )
```
```suggestion
        first_two_columns[even_row_indices, 0] = np.arange(6) + start_esa_step
```

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,928 @@
...
+        first_two_columns[odd_row_indices, 1] = np.arange(
+            start_esa_step + 6, start_esa_step + 12
+        )
```
```suggestion
        first_two_columns[odd_row_indices, 1] = np.arange(6) + start_esa_step + 6
```

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,928 @@
...
+    Returns
+    -------
+    var_names : dict
+        Dictionary with subset data populated in the checkerboard pattern.
+    """
+    # Flatten with top-down and left-right order. This will be used as
+    # indices to take and put data in the checkerboard pattern.
+    checkerboard_pattern = checkerboard_pattern.flatten(order="F")
+
+    # Variables that need to be put in the checkerboard pattern
```
greglucas, line 516: Would it be beneficial to store these as an xarray dataset organized by these dimensions and return that dataset? An xarray dataset is kind of just a nicer dictionary storage container.

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,928 @@
...
+        attrs=cdf_attrs.get_variable_attributes("acq_duration"),
+    )
+
+    dataset["esa_energy"] = xr.DataArray(
+        esa_energies,
+        dims=["epoch", "esa_step", "spin_sector"],
+        attrs=cdf_attrs.get_variable_attributes("esa_energy"),
+    )
+
+    # create xarray dataset for each metadata field
```
greglucas: Can you change all of the below from `metadata_field` to `data_field`. The metadata is really the text associated with the data.
```suggestion
    # create xarray dataarray for each data field
```

imap_processing/swe/l1b/swe_l1b.py
tech3371: From algorithm document. I don't know why SWE chose that number.

imap_processing/swe/l1b/swe_l1b.py
tech3371: Since I am reading multiple files and I am concatenating the datas, I don't have info about the column names. That's why I defined the column names. Do you mean that I read one file and extract column names from it instead of hard coding myself? Just curious

imap_processing/tests/swe/lut/imap_swe_l1b-in-flight-cal_20240510_20260716_v000.csv
tech3371: sounds good!

imap_processing/swe/l1b/swe_l1b.py
tech3371: I will think about it.

imap_processing/cli.py
tech3371: Looks like I pushed some changes while you were reviewing. Sorry. This comment has been addressed

imap_processing/tests/swe/test_swe_l1b_science.py
tech3371: I fixed this with above changes

jtniehof:
Of course this happens while I'm working on #1639. I think this fixes the energy ordering by using the actual values from l1b? Can there be a test similar to the one in #1639 to double-check this and make sure it doesn't regress?

tech3371:
> Of course this happens while I'm working on #1639. I think this fixes the energy ordering by using the actual values from l1b? Can there be a test similar to the one in #1639 to double-check this and make sure it doesn't regress?

Hi @jtniehof ,
I didn't know you were working on that. Sorry for missing that. Yeah, lot of this refactoring is to make current implementation better and make future addition easier. I also added new energy variable to simplify L2 and get energy in the pattern of checker board pattern we need in L2. Hope that helps

jtniehof:
No problem, I didn't tell you I was working on it because it was one of those cases where fully defining the problem basically involved fixing it :) At least the test from that might be useful--trying to figure a good unit test took most of the time.

imap_processing/swe/l1b/swe_l1b.py
tech3371: Greg, I am thinking of keeping it dictionary since it's an intermediate data product.

imap_processing/ialirt/l0/process_swe.py
tech3371, line 1: @laspsandoval I had to update SWE I-ALiRT to work with new changes. Can you review this file?

laspsandoval (approved):
I-ALiRT looks good to me.

imap_processing/tests/swe/test_swe_l2.py
```diff
@@ -345,6 +317,13 @@ def test_swe_l2(mock_read_in_flight_cal_data, use_fake_spin_data_for_time):
         swe_constants.N_ANGLE_SECTORS,
     )
 
+    rate = l1b_dataset.science_data.to_numpy()
+    psd = l2_dataset.phase_space_density_spin_sector.to_numpy()
+    rate = rate[2, :, :, 3]  # nonzero counts at all energy & spin
+    psd = psd[2, :, :, 3]
+    cal_factor = psd / rate  # same CEM, should be constant at a given energy
+    assert np.allclose(cal_factor, cal_factor[:, 0:1], rtol=1e-9, atol=0)
```
greglucas, line 325: If the `atol=0` I don't think `rtol` should be set to anything because it won't have an impact? Are you able to use the numpy testing assertions here too? There might be an issue with broadcastable shapes, but array comparisons are nicer with `np.testing` because they tell you which element failed.

imap_processing/swe/l1b/swe_l1b.py
greglucas: Might be good to add a quick comment about that.

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,923 @@
...
+
+    # Ensure acq_duration_sec is broadcastable to data
+    if acq_duration_sec.ndim < data.ndim:
+        acq_duration_sec = acq_duration_sec[
+            ..., np.newaxis
+        ]  # Add a new axis for broadcasting
+
+    # Perform element-wise division
+    count_rate = data / acq_duration_sec
+    return count_rate.astype(np.float64)
```
greglucas: To preserve precision, you should upcast before the division.

```suggestion
    count_rate = data.astype(np.float64) / acq_duration_sec
    return count_rate
```

imap_processing/swe/l1b/swe_l1b.py
greglucas: Yes. You currently call `read_csv(..., skiprows=1, header=None)` which you could change to `read_csv(...)` and let it infer the column names for you. Alternatively, put these into the `read_csv(..., header=0, names=your_list)` call directly so you are concatenating across the same names.

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -1,48 +1,923 @@
...
+
+    # create xarray dataarray for each data field
+    for key, value in full_cycle_l1a_data.items():
+        if key in ["science_data", "acq_duration"]:
+            continue
+        metadata_field = key.lower()
+        dataset[metadata_field] = xr.DataArray(
+            value.data.reshape(-1, swe_constants.N_QUARTER_CYCLES),
+            dims=["epoch", "cycle"],
+            attrs=cdf_attrs.get_variable_attributes(metadata_field),
```
```suggestion
        varname = key.lower()
        dataset[varname] = xr.DataArray(
            value.data.reshape(-1, swe_constants.N_QUARTER_CYCLES),
            dims=["epoch", "cycle"],
            attrs=cdf_attrs.get_variable_attributes(varname),
```

greglucas (approved):
Some minor suggestions that don't impact the overall restructuring, so looks good to me!

imap_processing/tests/swe/test_swe_l2.py
jtniehof: This one's me (and thanks @tech3371 for incorporating the test). "[The relative difference and the absolute difference atol are added together to compare against the absolute difference](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html)". So setting `atol` to zero means it's a strict relative comparison. PSDs in these units are really small numbers (10^-40 ish). I did find some differences that seem to be at the level of rounding, which is why it's not a strict equals (plus I'm always paranoid about floating point comparisons).

I like the `np.testing` methods but local style seemed to be bare asserts.

greglucas: Ahh, this makes sense about the atol/rtol and when those get applied, thanks for clarifying that!

> I like the np.testing methods but local style seemed to be bare asserts.

I think there is a mixture here and I much prefer `np.testing` whenever we can. I think the issue specifically here is that the shape also gets checked in the testing versions, so it doesn't "broadcast" out like your version here.

This is fine with me as-is, so we can consider this resolved!

tech3371: I checked with Greg on what he was thinking for this and he mentioned that what you had was good, Jon.

tech3371:
I am going to merge this even though `ubuntu-latest, 3.10` tests fails. It was unrelated to this PR and it's being worked on by Laura.

PR #1622, opened 2025-04-16
Ultra l1b - add params as a dependency to the extendedspin

laspsandoval:
# Change Summary

## Overview
Added params as a dependency to extendedspin. This will inform what rates data to use (sector or normal)

## Updated Files
- ultra_l1b.py
   - Added a dependency to extendedspin.py


greglucas (approved):

greglucas:
You also need to update the tests

PR #1623, opened 2025-04-17
ENA - dps frame code updated to include universal pointing table

laspsandoval:
# Change Summary

## Overview
Previously we had just used the intervals in the ck kernel to determine the start_time and end_time of the dps kernel. This adds the universal pointing table arrays so that it can be used instead.

## Updated Files
- pointing_frame.py
   - This was actually in kernels.py, but I had to move it since I was getting a circular dependency when I imported libraries from time.py due to ensure_spice.
   - Adds the universal pointing table arrays so that it can be used to determine start and end time of dps kernel.

## Testing
test_pointing_frame.py


imap_processing/spice/pointing_frame.py
```diff
@@ -0,0 +1,282 @@
...
+
+    Assumptions:
+    - The MOC has removed timeframe in which nutation/procession are present.
+    TODO: We may come back and have a check for this.
+    - The pointing frame kernel is made based on the most recent ck kernel.
+    In other words 1:1 ratio.
+
+    References
+    ----------
+    https://numpydoc.readthedocs.io/en/latest/format.html#references
```
greglucas:
```suggestion
```

imap_processing/spice/pointing_frame.py
```diff
@@ -0,0 +1,282 @@
...
+    Parameters
+    ----------
+    pointing_frame_path : pathlib.Path
+        Location of pointing frame kernel.
+    ck_path : pathlib.Path
+        Location of the CK kernel.
+    repoint_start_met : numpy.ndarray
+        Start time of the repointing in MET.
+    repoint_end_met : numpy.ndarray
+        End time of the repointing in MET.
```
greglucas, line 64: Just to verify, we are creating one pointing "frame" that includes all historical pointings too and updating it every time we get a new pointing file? The alternative would be to create a separate frame file for every pointing period.

Do we modify the existing pointing frames that were already in this file, overwrite them, append to the file?

imap_processing/spice/pointing_frame.py
```diff
@@ -0,0 +1,282 @@
...
+                q_avg,
+                # 0.0 Angular rotation terms.
+                np.array([0.0, 0.0, 0.0]),
+                # Rates (seconds per tick) at which the quaternion and
+                # angular velocity change.
+                np.array([1.0]),
+            )
+
+
+def _get_et_times(et_start: float, et_end: float) -> NDArray[np.float64]:
```
greglucas: I'd inline this function into the one place it is used. The naming of this is slightly confusing as it isn't "getting" et_times, you are passing those in and then interpolating between them.

imap_processing/spice/pointing_frame.py
```diff
@@ -0,0 +1,282 @@
...
+    # TODO: Queried pointing start and stop times here.
+    # TODO removing the @ensure_spice decorator when using the repointing table.
+
+    # 1 spin/15 seconds; 10 quaternions / spin.
+    num_samples = (et_end - et_start) / 15 * 10
+    # There were rounding errors when using spiceypy.pxform so np.ceil and np.floor
+    # were used to ensure the start and end times were included in the array.
+    et_times = np.linspace(
+        np.ceil(et_start * 1e6) / 1e6, np.floor(et_end * 1e6) / 1e6, int(num_samples)
+    )
```
greglucas: I'm a bit confused at what the desired outcome is here, won't this actually mean they possibly are **not** included?

If my input is a lot of decimals, this pushes it above my input value and linspace doesn't include what I passed.
```
np.ceil(1.2345678 * 1e6) / 1e6
1.234568
```

So my thought is that this is needed because the values within pxform require all values to be within the valid range of the CK and before it could have been outside that range?

greglucas (commented):
I don't see where the repointing table inclusions are used? It is quite hard to follow what is new vs old here because things were copied to a new location.

I added some other comments/questions because I don't remember what exactly happens here it has been so long since I've looked at this!

imap_processing/spice/pointing_frame.py
laspsandoval: @subagonsouth and the group had discussed this and decided that the simplest thing would be to have a 1:1 relationship between the attitude kernel and the dps kernel. So each time a new attitude kernel is delivered we create a new dps kernel. No appending, overwriting, etc. We just create a new one.

laspsandoval:
> I don't see where the repointing table inclusions are used? It is quite hard to follow what is new vs old here because things were copied to a new location.
> 
> I added some other comments/questions because I don't remember what exactly happens here it has been so long since I've looked at this!

The repointing table will be included in what is passed. You can see a comment in the test:

    # This is how the repoint data is generated.
    # We will use fake data for now to match the coverage of the attitude kernel.
    # repoint_df = get_repoint_data()
    # repoint_start_met = repoint_df["repoint_start_met"].values
    # repoint_end_met = repoint_df["repoint_end_met"].values

imap_processing/spice/pointing_frame.py
laspsandoval: You are correct. I'll modify my comment.

imap_processing/spice/pointing_frame.py
greglucas: Thanks for clarifying, that makes sense!

imap_processing/tests/spice/test_pointing_frame.py
```diff
@@ -0,0 +1,185 @@
...
+def test_create_pointing_frame(
+    spice_test_data_path, pointing_frame_kernels, tmp_path, et_times, fake_repoint_data
+):
+    """Tests create_pointing_frame function."""
+
+    # This is how the repoint data is generated.
+    # We will use fake data for now to match the coverage of the attitude kernel.
+    # repoint_df = get_repoint_data()
+    # repoint_start_met = repoint_df["repoint_start_met"].values
+    # repoint_end_met = repoint_df["repoint_end_met"].values
```
greglucas: Just a note that I don't think this will be quite right, I think it will actually need to be inverted. Our repointing file gives us the start and stop of the repointing maneuver (so ~15 minutes period). I think we'll want something like going from the end of the first repointing to the start of the second repointing. But we'll also need to handle some other potential edge cases like if a repointing maneuver hasn't finished.
```suggestion
    # repoint_df = get_repoint_data()
    # repoint_start = repoint_df["repoint_end_met"].values[:-1]
    # repoint_end_met = repoint_df["repoint_start_met"].values[1:]
```

greglucas: I wonder if we should have a method on the table that returns an iterable of the valid repointings so this is more re-usable. (not here)

greglucas (approved):

greglucas:
> The repointing table will be included in what is passed. You can see a comment in the test:

OK, I missed that this wasn't being used yet and only in the comments, thanks for clarifying. This is in preparation for that work then 👍

imap_processing/spice/pointing_frame.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    Create the pointing frame.
+
+    Parameters
+    ----------
+    pointing_frame_path : pathlib.Path
+        Location of pointing frame kernel.
+    ck_path : pathlib.Path
+        Location of the CK kernel.
+    repoint_start_met : numpy.ndarray
```
tmplummer, line 61: I think that there is some terminology confusion here. We need to be careful and distinguish between a repointing and a pointing.

A repointing is the time during which the s/c is changing the spin-axis. This is the time when the nutation will be largest and ENAs will ignore this data (Hi does not take science during a repoint).

A pointing is the time between the end of one repoint and the start of the next repoint, when the spin-axis is fixed and ENAs are collecting data that goes into a PSET.

 The repoint-table that we get from the MOC defines the start and end times of a repoint, not a pointing.

So, just to be sure, are the inputs to this function repoint start/end METs or pointing start/end METs?

imap_processing/spice/pointing_frame.py
```diff
@@ -0,0 +1,257 @@
...
+    repoint_start_met: NDArray,
+    repoint_end_met: NDArray,
+) -> None:
+    """
+    Create the pointing frame.
+
+    Parameters
+    ----------
+    pointing_frame_path : pathlib.Path
+        Location of pointing frame kernel.
```
tmplummer, line 58: I just realized that a frame kernel is a different thing from an attitude kernel. Is this the path to write out the new pointing frame attitude kernel (CK)?

imap_processing/spice/pointing_frame.py
```diff
@@ -0,0 +1,257 @@
...
+            # so np.ceil and np.floor were used to ensure the start
+            # and end times were within the ck range.
+            et_times = np.linspace(
+                np.ceil(et_start_repoint[i] * 1e6) / 1e6,
+                np.floor(et_end_repoint[i] * 1e6) / 1e6,
+                int(num_samples),
+            )
+
+            # Create a rotation matrix
+            rotation_matrix = _create_rotation_matrix(et_times)
```
tmplummer, line 127: Will you please write a ticket for handling gaps in the S/C CK?

Currently, if there is a gap, I think that the creation of the pointing kernel will totally fail. An intermediate solution would be to wrap this call in a try/except block that catches the `spiceypy` exception and just continues to the next pointing. But perhaps it is better to fail more obviously at this point.

tmplummer (commented):
I was too slow. I see that Greg already pointed out the pointing/repointing discrepancy. Everything else looks good to me.

PR #1626, opened 2025-04-17
MNT: Add too many arguments to list of rules to ignore

greglucas:
# Change Summary

## Overview

This rule has been causing more headache than it is worth. We shouldn't be changing good-enough code to satisfy the linter.


nkerman (approved):
Thank you for this, Greg!!!

PR #1633, opened 2025-04-21
Lo L1B: Direct Event Pointing Bins

### LINKED ISSUE #1519: Lo - Set L1B DE Pointing Bins
Set the L1B pointing bin for each direct event. This will be a 3600 x 40 bin (360 degrees x 4 degrees (-2 to +2))

The 3600 binning is determined using the spin phase and the 40 binning is determined by using the event time to get the instrument pointing in the despun frame. This will provide a direction for the event which can then be binned into one of the 40 bins.

It doesn't look like we have SPICE kernels for the sample data that Lo sent. Need to see if there's a way to get data from the MOC for this time period so this can be run in the CLI and tested/validated.

sdhoyt:
# Change Summary

## Overview
This PR adds the pointing direction and pointing bin fields to the Lo L1B Direct Events product. These are the final two fields needed for this product before moving on to L1C Pointing Sets.

I also updated the SPICE code to workaround a Lo related bug and updated the docstring to the `instrument_pointing` function so it's clearer what order the coordinates are returned.

## Updated Files
   - added function to set the pointing direction
   - added function to set the pointing bin
   - Change IMAP_LO enums ot IMAP_LO_BASE as workaround for Lo bug in SPICE kernel
   - Updated docstring so it's more clear what order the coordinates are returned in `instrument_pointing`

## Testing
Added two unit tests

Closes #1519

### DIFF
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -63,7 +63,7 @@ class SpiceFrame(IntEnum):
 
 
 BORESIGHT_LOOKUP = {
-    SpiceFrame.IMAP_LO: np.array([0, -1, 0]),
+    SpiceFrame.IMAP_LO_BASE: np.array([0, -1, 0]),
     SpiceFrame.IMAP_HI_45: np.array([0, 1, 0]),
     SpiceFrame.IMAP_HI_90: np.array([0, 1, 0]),
     SpiceFrame.IMAP_ULTRA_45: np.array([0, 0, 1]),
@@ -136,7 +136,7 @@ def get_spacecraft_to_instrument_spin_phase_offset(instrument: SpiceFrame) -> fl
     """
     # TODO: Implement retrieval from SPICE?
     offset_lookup = {
-        SpiceFrame.IMAP_LO: 330 / 360,
+        SpiceFrame.IMAP_LO_BASE: 330 / 360,
         SpiceFrame.IMAP_HI_45: 255 / 360,
         SpiceFrame.IMAP_HI_90: 285 / 360,
         SpiceFrame.IMAP_ULTRA_45: 33 / 360,
@@ -325,7 +325,7 @@ def instrument_pointing(
     """
     Compute the instrument pointing at the specified times.
 
-    By default, the coordinates returned are Latitude/Longitude coordinates in
+    By default, the coordinates returned are (Longitude, Latitude) coordinates in
     the reference frame `to_frame`. Cartesian coordinates can be returned if
     desired by setting `cartesian=True`.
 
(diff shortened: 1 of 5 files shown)


imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -503,3 +512,67 @@ def test_set_bad_times():
...
+                7.9794907049e17,
+                7.9794907153e17,
+                7.9794907254e17,
+            ],
+        },
+    )
+    # Need to better verify these values. Latitudes are likely to
+    # be 0, but not sure what values to expect from Longitudes.
+    # The expected values are based on what I'm getting from the
+    # SPICE tools.
```
greglucas: We are spinning with a period of 15 seconds, or 24 degrees/s. So you could probably choose an epoch 1 second later and see if it is 24 degrees off? It looks like you are using 1.04 seconds currently.

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -503,3 +512,67 @@ def test_set_bad_times():
...
+        atol=1e-1,
+    )
+
+
+def test_pointing_bins():
+    # Arrange
+    l1b_de = xr.Dataset(
+        {
+            "direction_lat": ("epoch", [0, 0, 0]),
+            "direction_lon": ("epoch", [91.3, 116.3, 140.5]),
```
greglucas: Add some negatives in here too to test you handle -180, 180 and not 0, 360 on accident.

imap_processing/tests/lo/test_lo_l1b.py
sdhoyt: That's a great idea, I didn't think to use the spin period for this. Thanks!

greglucas (approved):

imap_processing/spice/geometry.py
```diff
@@ -325,7 +325,7 @@ def instrument_pointing(
     """
     Compute the instrument pointing at the specified times.
 
-    By default, the coordinates returned are Latitude/Longitude coordinates in
+    By default, the coordinates returned are (Longitude, Latitude) coordinates in
```
tmplummer, line 328: Thanks for catching and fixing this!

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -58,7 +61,10 @@ def attr_mgr_l1a():
     return attr_mgr
 
 
-def test_lo_l1b():
+@patch("imap_processing.lo.l1b.lo_l1b.instrument_pointing")
+@pytest.mark.external_kernel
+@pytest.mark.use_test_metakernel("imap_ena_sim_metakernel.template")
```
tmplummer: I think that you can remove these marks given that you are mocking out the `instrument_pointing` function, no?

tmplummer (approved):
LGTM. Just the one comment/question about the mark fixtures.

imap_processing/tests/lo/test_lo_l1b.py
sdhoyt: True, thanks!

PR #1634, opened 2025-04-21
Pass start_date and repointing into write_cdf from CLI

maxinelasp:
# Change Summary
This wires up the CLI, repointing, and write_cdf methods. This ensures the output file uses the input start_date value from the CLI rather than computing from the epoch. It also allows us to pass a repointing number into the CLI instead of a date, and have that passed through to create a filename. 

This is the first step to fixing the repointing issues we saw in MSIM, although it does require updates to batch_starter to properly pass through the repointing number (but isn't a breaking change for anything)

## Overview

- Added the ability to pass repointing 
- Added keyword arguments to write_cdf for start_date and repointing that replace the calculated values
- Deprecated end_date and removed it from the processing class
- Made start_date optional

This does NOT introduce any breaking changes to the CLI.

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -1,5 +1,7 @@
 """Various utility functions to support creation of CDF files."""
 
+from __future__ import annotations
+
 import logging
 import re
 import warnings
@@ -64,6 +66,9 @@ def load_cdf(
 
 def write_cdf(
     dataset: xr.Dataset,
+    *,
+    start_date: str | None = None,
+    repointing: str | None = None,
     **extra_cdf_kwargs: dict,
 ) -> Path:
     """
@@ -73,13 +78,25 @@ def write_cdf(
     fills in the final attributes, and converts the whole dataset to a CDF.
     The date in the file name is determined by the time of the first epoch in the
     xarray Dataset.  The first 3 file name fields (mission, instrument, level) are
-    determined by the "Logical_source" attribute.  The version is determiend from
+    determined by the "Logical_source" attribute.  The version is determined from
     "Data_version".
+
+    The start_date and repointing parameters are used to override the computed values.
+    If these are not included, start_date is generated from the first epoch in the
+    dataset and repointing is retrieved from the dataset attributes (or not included if
+    the attribute is not present).
 
     Parameters
     ----------
     dataset : xarray.Dataset
         The dataset object to convert to a CDF.
+    start_date : str
+        In the output filename, this is the start_date. If not provided or None,
+        the first epoch in the dataset is used.
+    repointing : str
+        The repointing number to use in the output filename. If not provided or None,
+        the repointing number is retrieved from the dataset attributes. If none are
+        provided, it is excluded from the filename.
     **extra_cdf_kwargs : dict
         Additional keyword arguments to pass to the ``xarray_to_cdf`` function.
 
@@ -94,8 +111,11 @@ def write_cdf(
     # Convert J2000 epoch referenced data to datetime64
     # TODO: This implementation of epoch to time string results in an error of
     #       5 seconds due to 5 leap-second occurrences since the J2000 epoch.
-    dt64 = TTJ2000_EPOCH + dataset["epoch"].values[0].astype("timedelta64[ns]")
-    start_time = np.datetime_as_string(dt64, unit="D").replace("-", "")
+    # TODO: Create a ttj2000_to_datetime function to handle this conversion
+    if start_date is None:
+        # If no start time is included, then use the first epoch in the dataset
+        dt64 = TTJ2000_EPOCH + dataset["epoch"].values[0].astype("timedelta64[ns]")
+        start_date = np.datetime_as_string(dt64, unit="D").replace("-", "")
 
     version = dataset.attrs.get("Data_version", None)
     if version is None:
@@ -109,14 +129,17 @@ def write_cdf(
             f"The Data_version attribute {version} does not match expected format vXXX."
         )
 
-    repointing = dataset.attrs.get("Repointing", None)
+    # TODO: Do we need to retrieve this from the dataset?
+    if repointing is None:
+        repointing = dataset.attrs.get("Repointing", None)
+    repointing_int = int(repointing[-5:]) if repointing else None
     science_file = imap_data_access.ScienceFilePath.generate_from_inputs(
         instrument=instrument,
         data_level=data_level,
         descriptor=descriptor,
-        start_time=start_time,
+        start_time=start_date,
         version=version,
-        repointing=repointing,
+        repointing=repointing_int,
     )
     file_path = Path(science_file.construct_path())
     if not file_path.parent.exists():
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -108,6 +108,7 @@ def _parse_args() -> argparse.Namespace:
         '--data-level "l1a" '
         '--descriptor "all" '
         ' --start-date "20231212" '
+        '--repointing "repoint12345" '
         '--version "v001" '
         '--dependency "['
         "    {"
@@ -195,16 +196,24 @@ def _parse_args() -> argparse.Namespace:
     parser.add_argument(
         "--start-date",
         type=str,
-        required=True,
-        help="Start time for the output data. Format: YYYYMMDD or repoint#####",
+        required=False,
+        help="Start time for the output data. Format: YYYYMMDD",
     )
 
     parser.add_argument(
         "--end-date",
         type=str,
         required=False,
-        help="End time for the output data. If not provided, start_time will be used "
+        help="DEPRECATED: Do not use this."
+        "End time for the output data. If not provided, start_time will be used "
         "for end_time. Format: YYYYMMDD",
+    )
+    parser.add_argument(
+        "--repointing",
+        type=str,
+        required=False,
+        help="Repointing time for output data. Replaces start_time if both are "
+        "provided. Format: repoint#####",
     )
 
     parser.add_argument(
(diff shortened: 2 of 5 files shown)


imap_processing/cdf/utils.py
```diff
@@ -73,13 +78,25 @@ def write_cdf(
...
     ----------
     dataset : xarray.Dataset
         The dataset object to convert to a CDF.
+    start_date : str
+        In the output filename, this is the start_date. If not provided or None,
+        the first epoch in the dataset is used.
+    repointing : str
+        The repointing number to use in the output filename. If not provided or None,
+        the repointing number is retrieved from the dataset attributes. If none are
+        provided, it is excluded from the filename.
```
tech3371, line 99: What input format do you expect for repointing? Eg. `repoint00001` or `00001` or `1`. Based on input format, you can update below code to handle it.

imap_processing/cli.py
```diff
@@ -196,7 +196,7 @@ def _parse_args() -> argparse.Namespace:
         "--start-date",
         type=str,
         required=True,
-        help="Start time for the output data. Format: YYYYMMDD",
+        help="Start time for the output data. Format: YYYYMMDD or repoint#####",
```
tech3371: can you add another help for `repoint` or more details? Right now, it seems like we are expecting one or another. In actual, we are required to have start-date but repoint is optional. If repoint input is given, it would be appended to start date like this `<start-date>-repoint<repoint number in five digits representation>`.

imap_processing/cli.py
```diff
@@ -419,6 +418,9 @@ def post_processing(
         Child classes can override this method to customize the
         post-processing actions.
 
+        If start_date is used to generate the output file name by default, and can
+        either be a date in the form YYYYMMDD or a repointing in the form repoint#####.
```
tech3371: same here. This statement seems wrong to me or did I miss recent change?

imap_processing/cli.py
```diff
@@ -444,11 +446,24 @@ def post_processing(
         if not isinstance(self.version, str) or r.match(self.version) is None:
             self.version = f"v{int(self.version):03d}"  # vXXX
 
+        # Start date is either the start date or the repointing.
+        # if it is the repointing, default to using the first epoch in the file as
+        # start_date.
+        # If it is start_date, skip repointing in the output filename.
```
tech3371, line 486: same here.

imap_processing/cli.py
```diff
@@ -444,11 +446,24 @@ def post_processing(
...
+
+        start_date = None
+        if imap_data_access.ScienceFilePath.is_valid_date(self.start_date):
+            start_date = self.start_date
+
+        repointing = None
+        if imap_data_access.ScienceFilePath.is_valid_repointing(self.start_date):
+            repointing = self.start_date
```
tech3371: Is this a place holder for future work? otherwise, looks like we are passing start_date to repoint validation function when it's expecting `repointingXXXXX`

imap_processing/tests/cdf/test_utils.py
```diff
@@ -95,6 +95,19 @@ def test_written_and_loaded_dataset(test_dataset):
...
+    output_file_path = write_cdf(test_dataset)
+    assert "imap_swe_l1a_sci_20100101_v001.cdf" in output_file_path.name
+    output_file_path = write_cdf(test_dataset, start_date="20001212")
+    assert "imap_swe_l1a_sci_20001212_v001.cdf" in output_file_path.name
+    output_file_path = write_cdf(test_dataset, repointing="repoint12345")
+    assert "imap_swe_l1a_sci_20100101-repoint12345_v001.cdf" in output_file_path.name
+    output_file_path = write_cdf(
+        test_dataset, start_date="20001212", repointing="12345"
+    )
+    assert "imap_swe_l1a_sci_20001212-repoint12345_v001.cdf" in output_file_path.name
```
tech3371, line 108:
```suggestion
    assert "imap_swe_l1a_sci_20100101_v001.cdf" == output_file_path.name
    output_file_path = write_cdf(test_dataset, start_date="20001212")
    assert "imap_swe_l1a_sci_20001212_v001.cdf" == output_file_path.name
    output_file_path = write_cdf(test_dataset, repointing="repoint12345")
    assert "imap_swe_l1a_sci_20100101-repoint12345_v001.cdf" == output_file_path.name
    output_file_path = write_cdf(
        test_dataset, start_date="20001212", repointing="12345"
    )
    assert "imap_swe_l1a_sci_20001212-repoint12345_v001.cdf" == output_file_path.name
```

tech3371: Nice tests! That helps

imap_processing/cdf/utils.py
```diff
@@ -109,14 +129,17 @@ def write_cdf(
             f"The Data_version attribute {version} does not match expected format vXXX."
         )
 
-    repointing = dataset.attrs.get("Repointing", None)
+    # TODO: Do we need to retrieve this from the dataset?
```
greglucas, line 132: This is a good question! I think so and and that we should standardize this being in the metadata global attributes and then only read that as you do below and not allow a passing in here to reduce the potential multiple ways of setting this.

imap_processing/cli.py
greglucas: I don't love this multiple uses of `start_date`. Can we add another optional input parameter that is `--repoint` or similar? Do we want the batch starter to pass in both start date and repoint number? Another option if it is an exclusive would be to say "if `--repoint` is provided, the start_date is ignored and calculated based on the repointing table or something similar to that. But then it is explicit that it is a separate parameter and not mixed-use.

imap_processing/cdf/utils.py
maxinelasp: either `repoint00000` or `00000` will work for now. I'll update the docstring.

imap_processing/cli.py
maxinelasp: For the CLI, if repoint is provided, we should just take the start date out of the epoch values inside the dataset (which was the previous behavior). I think we should only define one time format in the CLI (start date or repoint) and then if we're using repoint, we can generate the other number from the data. I think that will be easier to deal with.

imap_processing/cli.py
maxinelasp: This was changed in the other code. If repoint is passed in on the CLI, the start_date is generated from epoch (which was the behavior before.) If start_data is passed in on the CLI, then it is used in the output file name. 

Example: We have data that spans from 01-01-2025 with a repointing number of 00001. 
we would use the repointing number to identify the time range, `imap_cli --start_date repoint00001`. In this case, the repointing number is the key piece of information. Then, the output file retrieves the start date from the epoch and outputs a file `imap_..._20250101-repoint00001_v001`

imap_processing/cli.py
maxinelasp: No, this works as intended, if the repointing is passed into start_date then we want to assign repointing to that number. I did it this way because I didn't want to add another CLI element and duplicate the time reporting for processing runs that use repointing. I can add a --repoint CLI piece instead if that makes more sense.

imap_processing/cli.py
maxinelasp: Sure, that seems fine. I think passing in both repoint and start_date is not ideal, but your idea works

imap_processing/cdf/utils.py
maxinelasp: If that's the case, then we shouldn't pass in the start_date or the repointing to the CLI and allow the output datasets to generate the entire filename.

maxinelasp: Right now, the start_date and end_date values are not used anywhere by anyone. The repointing number is kind of difficult to get in some cases, because it's in the filename of the L0 file, but it is possible to retrieve it and pass it in (although that could be unreliable). But, I could deprecate half the CLI inputs and make it so that the repoint number is only retrieved from the dataset attrs, and then get the repointing number from the input filename or something.

maxinelasp:
@greglucas @tech3371 I think I addressed both of your concerns by pulling repointing out into a new optional parameter. I also formally removed end_date (left it as an arg input since batch starter uses it occasionally, but removed it from the actual processing)

imap_processing/cli.py
maxinelasp: Reworked this to be more clear.

imap_processing/cdf/utils.py
```diff
@@ -109,14 +129,17 @@ def write_cdf(
...
+    if repointing is None:
+        repointing = dataset.attrs.get("Repointing", None)
+    repointing_int = int(repointing[-5:]) if repointing else None
```
greglucas, line 135: Seems like we are moving between "repoint12345" and "12345" in various places here. Can we settle on only one version? It looks like the int would be the one desired with the `generate_from_inputs(repointing=12345)` and (personally) preferring `ds.attrs["Repointing"] = 12345` rather than `ds.attrs["Repointing"] = "repoint12345"`

greglucas (commented):
Looks good. Just one question on whether we can change over to a common "repointing" style to be passed around everywhere. That can be done as a follow-up PR too.

greglucas (approved):

imap_processing/cdf/utils.py
maxinelasp: "repoint12345" is already used pretty much everywhere except here. I generally agree that the pure int would be preferable, but I think "repoint" is more commonly used. I was planning on smoothing out the standard everywhere in my future PRs (including in imap_data_access). 

These specific lines work for "repoint12345" or for "12345", but everywhere else is expecting "repoint12345". Do you want me to change these lines to enforce the standard here? I'm trying to avoid breaking changes where possible which is why I left it like this.

maxinelasp: But, this attribute is only used by GLOWS and I was going to rip it out in the follow-on PR to this anyway, so it is fine to break it.

greglucas: > But, this attribute is only used by GLOWS and I was going to rip it out in the follow-on PR to this anyway, so it is fine to break it.

Ahh, you were going to remove it in favor of it being in the filename only?

My approval stands and feel free to tackle this in follow-up PRs.

maxinelasp: Ok, I'll handle this in my GLOWS update PR and I think it will be easier to discuss there.

PR #1635, opened 2025-04-22
Removing crashing mag test

maxinelasp:
# Change Summary
Temporary fix to unblock test runs.


vmartinez-cu (approved):
Thanks for the quick fix!

PR #1645, opened 2025-04-22
Ultra l1c - added fixture for downloading data

laspsandoval:
# Change Summary

## Overview
Added fixture similar to what was done here: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1600/files

Fixture was trying to download data and being used in multiple places but the download hadn’t necessarily happened in the same subprocess that the other test was running in. The fix was to explicitly call downloading the files within the fixture and not relying on the mark.

## Updated Files
- test_ultra_l1c_pset_bins.py
   - Added fixture

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
@@ -100,6 +100,7 @@ def test_get_background_rates():
     assert background_rates.shape == hp.nside2npix(128)
 
 
+@pytest.mark.external_test_data
 def test_get_spacecraft_exposure_times():
     """Test get_spacecraft_exposure_times function."""
     constant_exposure = TEST_PATH / "ultra_90_dps_exposure.csv"
@@ -163,6 +164,7 @@ def test_get_helio_exposure_times():
     assert np.array_equal(np.squeeze(exposures[2]), exposure_3d[:, :, 23])
 
 
+@pytest.mark.external_test_data
 def test_get_spacecraft_sensitivity():
     """Tests get_spacecraft_sensitivity function."""
     # TODO: remove below here with lookup table aux api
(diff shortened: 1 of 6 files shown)


imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
```diff
@@ -99,8 +110,7 @@ def test_get_background_rates():
     assert background_rates.shape == hp.nside2npix(128)
 
 
-@pytest.mark.external_test_data
```
greglucas: This is still needed on all of these tests to tell it that it requires external data. This will make it only download in the one runner still.

imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
```diff
@@ -35,6 +36,16 @@ def test_data():
     return v, energy
 
 
+@pytest.fixture(scope="session")
+def test_l1c_data():
+    """Ensure external test data is downloaded before tests run."""
+
+    # Make sure we have the data available here. This test collection gets
+    # skipped at the module level if the mark isn't present. We can't decorate
```
greglucas: This file doesn't have the module level skip so this comment can be removed and the other marks can be utilized still.

tech3371 (approved):

imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
```diff
@@ -197,20 +197,6 @@ def test_get_spacecraft_sensitivity():
         df_sensitivity_test.to_numpy(), expected_sensitivity.to_numpy(), atol=1e-6
     )
 
-
-@pytest.mark.external_test_data
-def test_grid_sensitivity():
```
greglucas, line 202: Do you not need this test anymore?

Was it only this one test causing all the failures? :( Really weird, hopefully it wasn't just luck for this single run passing!

greglucas (approved):

PR #1646, opened 2025-04-22
Ultra L2: Enable Rectangular Output and Attribute Handling

### LINKED ISSUE #1643: Enable Rectangular Tiled Output for Ultra L2
Now that #1596 is merged, we can enable the Ultra L2 code to output RectangularSkyMap Datasets if requested.

### LINKED ISSUE #1644: Add global and variable-level Attributes to Ultra L2
Ensure compliance with ISTP.

nkerman:
# Change Summary

## Overview

Accomplishes two closely-related tasks:
1. Now that the #1596 has been merged, and we can convert from Healpix -> Rectangular sky maps, we can easily enable Ultra L2 to produce Rectangular tiled output. Previously, it raised a `NotImplementedError`.
2. Add ISTP-compliant metadata attributes, both global, and variable specific. These will be different for the rectangular and healpix tiled datasets. 

While the metadata is currently ISTP compliant, I don't expect it will be the final version, and I expect additional metadata PRs to come, for the following reasons:
1. The exact desired metadata format for the maps is still being worked out. 
2. I have yet to fully understand SPASE metadata, and what other values must be added. 

Closes #1644
Closes #1643 

## New Files

-  imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
-  imap_processing/cdf/config/imap_ultra_l2-rectangular_variable_attrs.yaml
   - YAML containing Variable level metadata for the healpix and rectangular L2 maps, respectively.

## Updated Files

-  imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
   - Add global metadata for Ultra L2
-  imap_processing/tests/ultra/data/mock_data.py
   - Small: add epoch dim to the `exposure_factor` variable in the Healpix tiled l1C dataset. Better reflects actual L1C output
   - Bug fix: coords had been dropped on `rectangular_map.to_dataset()` call. Only became evident with use of `to_rectangular_map` in this PR 
   - Enable Rectangular Output
   - Add metadata attributes 

## Testing

Added test for Rectangular output

Note: I may add some more testing - please don't hesitate to suggest regions that could be better tested!


imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
```diff
@@ -259,3 +259,28 @@ imap_ultra_l1a_90sensor-imgparams:
     Data_type: L1A_90Sensor-ImgParams>Level-1A Image Parameters for Ultra90
     Logical_source: imap_ultra_l1a_90sensor-imgparams
     Logical_source_description: IMAP-Ultra Instrument Level-1A Image Parameters Data for Ultra90.
+
+# Ultra L2 Global attributes for different sensors and sky tiling types
+imap_ultra_l2_45sensor-enamaphealpix:
```
greglucas: Should we also include the "frame" in the descriptor too? Is the frame in the variable names so you'd produce multiple in a single file?

If the rect is derived from healpix, would it make sense to put those both together in a single file that has the descriptor "enamaps"?

No need to update here, but things to ask how folks want these broken up. My impression is this decision may have already happened.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,147 @@
...
+
+energy_bin_geometric_mean_label:
+  VAR_TYPE: metadata
+  CATDESC: Geometric mean of energy bin.
+  FIELDNAM: energy_bin_geometric_mean
+  FORMAT: A10
+  DEPEND_1: energy_bin_geometric_mean
+
+# Define Coordinate-like variables
+# (longitude, latitude are not technically coordinates for healpix tiling)
```
greglucas: Interesting that they aren't coordinates. Do you mean "dimension" here? (no need to update this, just writing down details in case it helps clarify anything)

xarray has both dimension coordinates and non-dimension coordinates, but maybe CDF doesn't support that.

I often have data that is an array of "site locations", each "site" has an associated lat/lon, but the dimension of the data is still per-site. So my Dataset structure looks like this:
```python
dims: (time, site)
coords:
  - time(time)
  - site(site)
  - latitude(site)
  - longitude(site)
Data Variables:
  - var1(time, site)
  - var2(time, site)
```
where lat/lon aren't in the data variables section because they aren't the measured quantities, they are just a different way of labeling the sites.

imap_processing/cdf/config/imap_ultra_l2-rectangular_variable_attrs.yaml
```diff
@@ -0,0 +1,150 @@
...
+  VAR_TYPE: data
+  LABL_PTR_1: energy_bin_geometric_mean_label
+  LABL_PTR_2: longitude_label
+  LABL_PTR_3: latitude_label
+
+flux_uncertainty:
+  <<: *default_float32
+  CATDESC: Statistical uncertainty on flux.
+  FIELDNAM: flux uncertainty
+  UNITS: counts/(s * cm^3 * Sr * KeV)
```
greglucas: Should this be per unit area for a flux?
```suggestion
  UNITS: counts/(s * cm^2 * Sr * KeV)
```

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -296,4 +345,41 @@ def ultra_l2(
...
+            continue
+
+        # The longitude and latitude variables will be present only in Healpix tiled
+        # map, and, as support_data, should not have schema validation
+        map_dataset[variable].attrs.update(
+            cdf_attrs.get_variable_attributes(
+                variable_name=variable,
+                check_schema=(
+                    True if variable not in ["longitude", "latitude"] else False
+                ),
```
greglucas: You can let the bool fall out of your computation
```suggestion
                check_schema=variable not in ["longitude", "latitude"],
```

nkerman: Good point!

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
nkerman: >I often have data that is an array of "site locations", each "site" has an associated lat/lon, but the dimension of the data is still per-site....

Yes, I think this is directly analogous to what's happening here - The **coordinate dimension** is "healpix_index", analogous to "site", and at each of those indices, we have a central lon/lat value. I take that to mean that longitude/latitude are data_vars, since they're values 'measured'/'calculated' at spots along the dimension of healpix index. 

I currently think of coordinates as something which should not depend on anything aside from themselves (e.g. the only dimension of a coordinate is itself). However, maybe it's better to think of these as non-dimension coordinates, which do indeed depend on another coordinate: healpix_index?

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
nkerman: > Should we also include the "frame" in the descriptor too? Is the frame in the variable names so you'd produce multiple in a single file?

I'm still getting a handle on how these global attributes/the logical source work (I think that's the correct term for `imap_ultra_l2_45sensor-enamaphealpix`?), how they work with SDS-Data-Manager, etc. If we're going to need to make $N_{maps} = M \times O \times P \times Q$ different maps of combinations of M tilings, O reference frames, P spacings, and Q time ranges `(e.g. a M=Rectangularly tiled, O=HAE frame, P=6 degree spacing, Q = 6 months)` do we need to have $N_{maps}$ different Logical Sources defined here? I think the answer is yes, but I think it's probably worth doing a few simple ones that are not time range dependent as a test case.

> If the rect is derived from healpix, would it make sense to put those both together in a single file that has the descriptor "enamaps"?

I think Dan et al want them in separate files.

These are somewhat computationally expensive, and so it _could_ make sense to only create the healpix map once, but it's really not too bad (especially the healpix ones). Given how infrequently the files will be created, I kind of like the "one run --> one map" approach. 

> No need to update here, but things to ask how folks want these broken up. My impression is this decision may have already happened.

I bet that @subagonsouth will have a lot more thoughts and ideas on this. Tim, I need to plug back into the conversation with Dan, as it's super related to this. Thank you for leading that on the SDC side!

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
greglucas: That would be my opinion, yes, but I don't know that CDF allows for that representation or if that is only netcdf that makes those distinctions. I think and aligns with weather stations having a latitude/longitude basically as metadata, it isn't a "measurement value" / data_var. So those lat/lon values would go with the coordinates and depend on "station". The dims do have a 1-1 mapping of depending upon themselves. There is some good terminology descriptions of the options here: https://docs.xarray.dev/en/latest/user-guide/terminology.html

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
tmplummer, line 1: I haven't though all of the way through this yet, but I think that we ultimately want to put all of the common/required Map global attributes into a single common yaml file.

One thing I did for Hi was to define common global attribute strings that are formatted to be used with the `str.format()` function so that the sensor string can be injected. For example: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/cdf/config/imap_hi_global_cdf_attrs.yaml#L19

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
tmplummer: Yes, the current proposal for descriptor has the following frames:
**frame: spacecraft (sf), heliospheric (hf), intermediate heliospheric (cg)**

>  If the rect is derived from healpix, would it make sense to put those both together in a single file that has the descriptor "enamaps"?
>> I think Dan et al want them in separate files.

I am still under the impression that we are not producing any healpix maps. The Map product CDF definition allows for HealPix but none are listed in the document that contains all of the L2/3 map products.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
tmplummer, line 1: Again, it is preferable to put all of the variable attributes in a common Map yaml. I don't yet know how we will override certain attributes if, for example, each instrument needs a different CATDESC.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,147 @@
...
+  FILLVAL: .NAN
+  FORMAT: F12.6
+  VALIDMIN: -3.4028235e+38
+  VALIDMAX: 3.4028235e+38
+  dtype: float32
+
+# Define Healpix Tiled Map values
+
+# Define Coordinates
+healpix_index:
```
tmplummer: We should start with these names matching the Map CDF definition spreadsheet in order to reduce the number of changes that will need to be made in the future. For example, the current name for this coordiante is `pixel_index`.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
tmplummer: Cool. I didn't know about this distinction between coordinates and dimensions in xarray. The real question is what does cdflib support. CDF has no concept of coordinates, everything is just a variable.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,147 @@
...
+  <<: *default_float32
+  CATDESC: Exposure time calculated from multiple pointing sets.
+  FIELDNAM: exposure_times
+  DEPEND_0: epoch
+  DEPEND_1: healpix_index
+  UNITS: seconds
+  VAR_TYPE: data
+  LABL_PTR_1: healpix_index_label
+
+flux:
```
tmplummer: It has been agreed that we are not reporting flux. It is `ena_intensity`.

tmplummer (commented):
Some comments to consider.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
nkerman: The description of Non-dimension coordinate[#](https://docs.xarray.dev/en/latest/user-guide/terminology.html#term-Non-dimension-coordinate) there is helpful. I bet that CDF will end up treating it the same as, I think Tim's message implies, but I'll give a try to switching lon/lat to be coordinates.

imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
nkerman: Thanks for the input, Tim!

>Yes, the current proposal for descriptor has the following frames:
frame: spacecraft (sf), heliospheric (hf), intermediate heliospheric (cg)

Got it - I'll add **hf** to these for now. 

>I am still under the impression that we are not producing any healpix maps. The Map product CDF definition allows for HealPix but none are listed in the document that contains all of the L2/3 map products.

Okay. I was under the impression that we would be, given the Map product definition is  defined with them, but that is a good point that it's not present in the spreadsheet listing output maps. I think we should still have definitions for the Healpix Datasets probably.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
nkerman: Ah, yeah thank you. I was referring to the Algorithm Doc, but that is getting really out of date.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
nkerman: I'll come back tomorrow to thinking about the common Map yaml, which I'm a bit concerned about; however, for now, I've made all the updates I could find to align with the excel spreadsheet.

imap_processing/cdf/config/imap_ultra_l2-healpix_variable_attrs.yaml
nkerman: I tried this, and even if I add it as a coord, `CDFLib` treats it as a regular data_var, rather than a coord, since it is not dependent on itself. 

If it makes sense to you, I'll leave this as it was (that is: healpix longitude/latitude created as data_vars, rather than coords.) 

However, I kept all these changes in a stash if you'd prefer to instantiate them as coords, even if they'll always be read back in as data_vars.

greglucas: No, please carry on here with what you've got! I would much prefer to be working in the NetCDF world where this is valid, but alas that is not the world we are living in.

nkerman: Sounds good. CDFs are **terrible** compared to NetCDFs!

nkerman:
Thanks @subagonsouth for the suggestions! I've made most of the switches we spoke about here. 

At present, I haven't made sure that this is fully compatible with the evolving IMWG standards (I have another PR for starting to keep track of the naming conventions), but I lean towards prototyping a couple of maps to make sure they work, and then we can change the names, attrs, etc.  Let me know what you think!

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,167 @@
...
+  VALIDMAX: 9223372036854775807
+  VAR_TYPE: data
+  UNITS: " "
+
+default_float32_attrs: &default_float32
+  <<: *default
+  FILLVAL: .NAN
+  FORMAT: F12.6
+  VALIDMIN: -3.4028235e+38
+  VALIDMAX: 3.4028235e+38
```
tech3371: Not for this PR but once this PR is merged, you might want to run L2 CDF file through SKTEditor. It will point of lot of smaller issues that `cdflib` doesn't catch but something SPDF seems to care. Since this is L2, I wanted to mention that. Eg. `.NAN` or scientific notation, it was causing some issue for me in the past when I viewed it through SKTEditor.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,167 @@
+default_coords_int: &default_coords_int
+  # Assumed values for all int coordinate attrs unless overwritten
+  FILLVAL: -9223372036854775808
+  FORMAT: I19
+  VALIDMIN: -9223372036854775808
+  VALIDMAX: 9223372036854769664
```
tech3371: Since this is only used by `pixel_index`, you may want to pick valid min/max that you know about instead of the defaults.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,167 @@
...
+  dtype: float32
+
+# Define Healpix Tiled Map values
+
+# Define Coordinates
+pixel_index:
+  <<: *default_coords_int
+  CATDESC: Healpix index corresponding to the bin center.
+  FIELDNAM: pixel_index
+  LABLAXIS: HEALPix \#
```
tech3371: Do you want to spell out `#`? If label axis is getting too long, you can abbreviate. Something like `idx`.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,167 @@
...
+  DEPEND_1: pixel_index
+  CATDESC: HEALPix pixel center longitude in range [0, 360].
+  FIELDNAM: Longitude
+  LABLAXIS: Longitude
+  UNITS: degrees
+  VAR_TYPE: support_data
+  LABL_PTR_1: pixel_index_label
+  DISPLAY_TYPE: no_plot
+  VALIDMIN: 0
+  VALIDMAX: 360
```
tech3371: Nice find with this! It was nice to know that we can store data that doesn't have `epoch` has dimension. Can you reword the comment little bit? I added suggestion with what I think you are trying to say there. Don't hesitate to correct it :)
```suggestion
# Define Non-epoch Data Variables
# (longitude, latitude are data variable that doesn't use epoch as dimension. Instead, it uses for healpix index for healpix tiling)
longitude:
  <<: *default_coords_float32
  DEPEND_1: pixel_index
  CATDESC: HEALPix pixel center longitude in range [0, 360].
  FIELDNAM: Longitude
  LABLAXIS: Longitude
  UNITS: degrees
  VAR_TYPE: support_data
  LABL_PTR_1: pixel_index_label
  DISPLAY_TYPE: no_plot
  VALIDMIN: 0
  VALIDMAX: 360
```

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,167 @@
...
+  VAR_TYPE: support_data
+  LABL_PTR_1: pixel_index_label
+  DISPLAY_TYPE: no_plot
+  VALIDMIN: -90
+  VALIDMAX: 90
+
+# Define Data variables
+exposure_factor:
+  <<: *default_float32
+  CATDESC: Exact or approximate exposure time over which counts in a pixel are accumulated.  Used as a weighting factor for combining data quantities sensibly.
```
tech3371: I think there is unwritten requirement for description to be under char 80. If it's long, it will need to go into `VAR_NOTES`.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,167 @@
...
+ena_intensity_stat_unc:
+  <<: *default_float32
+  CATDESC: ENA intensity statistical uncertainty.
+  FIELDNAM: Intensity stat unc
+  UNITS: counts/(s * cm^2 * Sr * KeV)
+  DEPEND_0: epoch
+  DEPEND_1: energy
+  DEPEND_2: pixel_index
+  VAR_TYPE: data
+  LABLAXIS: Statistical Unc.
```
tech3371:
```suggestion
  LABLAXIS: Statistical Unc
```

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,167 @@
...
+  <<: *default
+  CATDESC: Exposure time weighted mean collection date of data in a pixel.
+  FIELDNAM: J2000 Nanoseconds
+  UNITS: ns
+  DEPEND_0: epoch
+  DEPEND_1: pixel_index
+  VAR_TYPE: data
+  LABLAXIS: epoch
+  LABL_PTR_1: pixel_index_label
+  DISPLAY_TYPE: image
```
tech3371: I have not seen this display type. interesting.

imap_processing/cdf/config/imap_enamaps_l2-rectangular_variable_attrs.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+longitude_delta:
+  VAR_TYPE: metadata
+  dtype: float32
+  CATDESC: Half-width of longitude pixel
+  FORMAT: I3
+  UNITS: degrees
+  FIELDNAM: longitude delta
+
+latitude:
```
tech3371: I came back to this PR after sometime and got confused with `latitude` coordinate from above file. Should you rename either one to be something else or do they mean same?

tech3371 (commented):
I looked at CDF attrs. I would highly recommend running your CDF file into SKTEditor. It will let you know smaller details.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: Thanks Tenzin! This is super helpful for catching small CDF errors. I got plenty of feedback from `SKTEditor`:

RECTANGULAR
```
###############################
Compliance Check for /Users/nake7532/Projects/IMAP/imap_processing/imap_processing/ena_maps/data/imap/ultra/l2/2025/05/imap_ultra_l2_90sensor-enamapRectangular-999mo_20250501_v999.cdf
CDF File Version: 3.9.0
File Last Leap Second: 2017-01-01
Majority: Row
/Users/nake7532/Projects/IMAP/imap_processing/imap_processing/ena_maps/data/imap/ultra/l2/2025/05/imap_ultra_l2_90sensor-enamapRectangular-999mo_20250501_v999.cdf is not ISTP-Compliant.
Global errors:
	Warning: CDF is set for row major array variables and column major is recommended.
	Attribute 'units' should be named 'UNITS'.
	Global attribute Spacing_degrees is of type CDF_INT8.
	    Datatypes other than CDF_CHAR may be problematic.
	Global attribute Spice_reference_frame is of type CDF_INT8.
	    Datatypes other than CDF_CHAR may be problematic.
	Logical_file_id should be 'imap_ultra_l2_90sensor-enamaprectangular_20250501_v999'.  It is 'imap_ultra_l2_90sensor-enamapRectangular-999mo_20250501_v999'.
The following variables are not ISTP-compliant:
	exposure_factor
		FILLVAL value of '0.0' is non-standard.
		    The recommended value is '-1.0E31'.
		Width of F12 FORMAT may be insufficient for values.  Adjust FORMAT or VALIDMIN/VALIDMAX value to resolve.
	obs_date
		FILLVAL value of '-9.223372036854776E18' is non-standard.
		    The recommended value is '-1.0E31'.
		FORMAT 'I12' is invalid for a CDF_DOUBLE type variable
	sensitivity
		FILLVAL value of '0.0' is non-standard.
		    The recommended value is '-1.0E31'.
		Width of F12 FORMAT may be insufficient for values.  Adjust FORMAT or VALIDMIN/VALIDMAX value to resolve.
	ena_intensity
		FILLVAL value of '0.0' is non-standard.
		    The recommended value is '-1.0E31'.
		Width of F12 FORMAT may be insufficient for values.  Adjust FORMAT or VALIDMIN/VALIDMAX value to resolve.
	ena_intensity_stat_unc
		FILLVAL value of '0.0' is non-standard.
		    The recommended value is '-1.0E31'.
		Width of F12 FORMAT may be insufficient for values.  Adjust FORMAT or VALIDMIN/VALIDMAX value to resolve.
	longitude_delta
		FORMAT 'I3' is invalid for a CDF_DOUBLE type variable
	latitude_delta
		FORMAT 'I3' is invalid for a CDF_DOUBLE type variable
	longitude
		Attributes named UNITS and UNITS have names that differ only in case.
		Attributes named UNITS and UNITS have names that differ only in case.
	latitude
		Attributes named UNITS and UNITS have names that differ only in case.
		Attributes named UNITS and UNITS have names that differ only in case.
	energy
		FORMAT 'I2' is invalid for a CDF_DOUBLE type variable
```

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: Of course! 🤦 
This can also be an unsigned int. Thanks

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: I'm trying to follow the [conventions being set by the mini-IMWG](https://o365coloradoedu-my.sharepoint.com/:x:/g/personal/plummert_colorado_edu/EX8IRmgRKw5Hpik9wPAXZQABBgOemDqOv-uLMlRxZLJaTw?e=aeHEKG) (primarily Dan R and Tim's work). Is it okay to use char's like "#"? I could probably put it in quotes instead in the yaml:  

```
LABLAXIS: "HEALPix #"
```

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: While these aren't epoch dependent, I don't think that's their defining characteristic (none of the other coordinates are either). [Greg's previous comment](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1646#discussion_r2056267450) on this PR got me to understand the difference:
> ...
xarray has both dimension coordinates and non-dimension coordinates
...

nkerman: I can add some more context to the comment though!

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: Good to know. I'm checking with Tim and Dan, but suggesting we do "Exact or approximate exposure time over which pixel counts are accumulated."

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: I think we want the period there to indicate the shortening of Uncertainty -> Unc.

Are periods not allowed in LABLAXIS?

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: Image is listed here: https://spdf.gsfc.nasa.gov/istp_guide/vattributes.html#DISPLAY_TYPE

imap_processing/cdf/config/imap_enamaps_l2-rectangular_variable_attrs.yaml
nkerman: Latitude and Longitude always refer to the pixel center's latitude/longitude. For Rectangular Maps, this is a dimension coordinate (I think of this as a "real" coordinate), but for Healpix Maps, it's just a separate value for every Healpix pixel: a non-dimension coordinate. 

I think they should have the same name.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
tmplummer: This wording was copied from some Hi document. Feel free to modify as desired for now. We will likely have to update these once the CDF definition is finalized but no need to wait for that.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
tmplummer, line 1: Could the common variables be put into some sort of `imap_enamaps_common_variables.yaml`? I think that when additional attribute files get loaded into the `AttributeManager`, it is the same as doing a `dict.update()` so it is possible to load multiple attributes files where the later ones overwrite any collisions with previously loaded files.

I think that ultimately, we will want a bunch of variables in a base map variables file that is independent of map type and product level.

tmplummer (commented):
Thanks for the move to common files. I think that we need to explore going even further toward shared variable definitions that are common to all maps at all data levels. That can be for another ticket though.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: I totally agree, Tim. The challenge is that anything which is pixel dependent (every data_var,: e.g., ena_intensity, exposure_factor, ...) all need different `DEPEND`s and `LABL_PTR`s for healpix and rectangular. Worse still, there are different numbers of coords between the two. As such, some additional logic would be required to add to OR fill the final depends and label_pointers with either one or two values.

Of the following variables, only `energy` and `energy_label` are completely the same between the two tilings. 

- Coords: 
  - ~~pixel_index~~ (only in healpix)
  - ~~pixel_index_label~~ (only in healpix)
  - energy
  - energy_label
  - ~~longitude~~  (different usages and depends in healpix/rectangular)
  - ~~latitude~~   (different usages and depends in healpix/rectangular)
- Data_vars: 
  - ~~exposure_factor~~  (different dimensions/depends in healpix/rectangular)
  - ~~ena_intensity~~  (different dimensions/depends in healpix/rectangular)
  - ~~ena_intensity_stat_unc~~   (different dimensions/depends in healpix/rectangular)
  - ~~sensitivity~~   (different dimensions/depends in healpix/rectangular)
  - ~~obs_date~~   (different dimensions/depends in healpix/rectangular)

---
There's probably a decent way to do this; I think it will require logic in the `AttributeManager` though, as the yamls don't really support logic.

nkerman: I realize after writing this that (and this is probably what you meant!) that we could still have the main information on the attributes (CATDESC, VALIDMIN/VALIDMAX/FILLVAL, ...) in such a `imap_enamaps_common_variables.yaml` file, and not include DEPENDS/LABL_PTRs there at all. Then those would _always_ be loaded in from a separate HEALPix-specific or Rectangular-specific yaml. 

This could get a little bit weird with coordinates, but let me give it a try.

tmplummer: Don't bang your head against the wall too hard on trying to get this to work. If the correct place to do this is in the `AttributeManager`, lets put in a feature request for SAMMI to add inheritance across yaml files.

nkerman: Turns out it's not bad at all! I just hadn't pushed it yet because I got caught diagnosing the sudden test failures. (Still working on that!)

It works fine to load in the common attrs and overwrite them with additional information in the rect/hp specific yaml files, and I just pushed the changes.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
tech3371: Yep. You can ignore this warning. Otherwise, everything is something SPDF like us to fix.
```
Warning: CDF is set for row major array variables and column major is recommended.
```

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
tech3371: SPDF has been particular about these things. That's why I mention that.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -268,32 +279,133 @@ def ultra_l2(
...
+    # Rename any variables as necessary for L2 Map schema compliance
+    # Energy at L1C is named "energy_bin_geometric_mean", but at L2 it is standardized
+    # to "energy" for all instruments.
+    map_dataset = map_dataset.rename({"energy_bin_geometric_mean": "energy"})
+
     # Add the defined attributes to the map's global attrs
     map_dataset.attrs.update(map_attrs)
+
+    # Add the "label" coordinates to the map dataset
+    for coord_var, coord_data in map_dataset.coords.items():
```
tmplummer, line 377: Perhaps not for this PR, but I imagine that lots of this code handling global attributes, coordinate label variables, and variable attributes will be common and should be moved to a common location. Perhaps it makes sense to make it all part of the `SkyMap` classes. It could be baked into the `to_dataset` function, or maybe a `cdf_ify_dataset` type function.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,91 @@
...
+  VAR_TYPE: metadata
+  CATDESC: Healpix index corresponding to the bin center.
+  FIELDNAM: pixel_index
+  FORMAT: I10
+  DEPEND_1: pixel_index
+
+# Define Non-dimension Coordinates
+# (longitude, latitude are not dimension coordinates for healpix tiling, but rather
+# describe the lon/lat coordinate of the Healpix pixel center)
+longitude:
```
tmplummer, line 41: It looks like there is more room for commonality. It looks like most of the lat/lon attributes are the same. I would think that even the CATDESC of `Pixel center longitude in range [0, 360]` could be made common for `longitude`. Only differences: `DEPEND_1`, `LABL_PTR_1`, and `DELTA_PLUS/MINUS`?

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
```diff
@@ -0,0 +1,91 @@
...
+  LABLAXIS: Latitude
+  UNITS: degrees
+  VAR_TYPE: support_data
+  LABL_PTR_1: pixel_index_label
+  DISPLAY_TYPE: no_plot
+  VALIDMIN: -90
+  VALIDMAX: 90
+
+# Define Data variables
+exposure_factor:
```
tmplummer, line 50: It is probably worth adding at least 1 comment indicating that the below variable attributes supplement/override values defined in the common yaml.

tmplummer (approved):
A few more comments to look at before merging. I'll give approval because I think you are really close... assuming you can figure out the testing issue.

tmplummer:
I see that you got tests working. Did you find a root cause of the issue? Can you summarize it, if so?

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: Finally got rid of all the warnings aside from the row major one!

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: Done! I consolidated most of the data here into the common file.

imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
nkerman: Done!

imap_processing/ultra/l2/ultra_l2.py
nkerman: That makes sense - I think this might be much easier to extract (or rather know what/how much to extract as common between instruments) once we have more than one instrument to look at. At this point, I'm not sure what exactly is common for all the map instruments.

nkerman:
> I see that you got tests working. Did you find a root cause of the issue? Can you summarize it, if so?

I think it was an under-the-hood memory failure. I cut the size of the PSETs (increased the `spacing_deg`/decreased `nside`) and everything started to consistently pass. I wish it were a less opaque error that said **anything** about the cause.

PR #1652, opened 2025-04-23
HIT L2 - add total uncertainties

### LINKED ISSUE #1651: Add total uncertainties
Add total uncertainty variables that combine statistical and systematic uncertainties using equation 13 from the algorithm document. This work is listed in the change request log for Q1.

vmartinez-cu:
This PR adds total uncertainty variables to L2 datasets. It also fixes the shape of systematic uncertainties to match the data variables they reference.

Closes #1651 

## Updated Files
- hit_l2.py
   - Add new function to add total uncertainties to datasets
   - Fix shape of systematic uncertainties added to datasets
- test_hit_l2.py
   - Add test for new function
   - Update test for systematic uncertainties to reflect changes

### DIFF
--- a/imap_processing/hit/l2/hit_l2.py
+++ b/imap_processing/hit/l2/hit_l2.py
@@ -29,7 +29,6 @@ logger = logging.getLogger(__name__)
 # TODO:
 #  - review logging levels to use (debug vs. info)
 #  - determine where to pull ancillary data. Storing it locally for now
-#  - add function to calculate combined uncertainty and add this to L2 datasets
 
 
 def hit_l2(dependency: xr.Dataset) -> list[xr.Dataset]:
@@ -576,16 +575,20 @@ def process_summed_intensity_data(l1b_summed_rates_dataset: xr.Dataset) -> xr.Da
         L2_SUMMED_ANCILLARY_PATH_PREFIX,
     )
 
-    # Add systematic uncertainties to the dataset. These will not
-    # have the intensity calculation applied to them
+    # Calculate the intensity for each species
+    l2_summed_intensity_dataset = calculate_intensities_for_all_species(
+        l2_summed_intensity_dataset, ancillary_data_frames, VALID_SPECIES
+    )
+
+    # Add total and systematic uncertainties to the dataset
     for var in l2_summed_intensity_dataset.data_vars:
         if var in VALID_SPECIES:
             l2_summed_intensity_dataset = add_systematic_uncertainties(
                 l2_summed_intensity_dataset, var
             )
-    l2_summed_intensity_dataset = calculate_intensities_for_all_species(
-        l2_summed_intensity_dataset, ancillary_data_frames, VALID_SPECIES
-    )
+            l2_summed_intensity_dataset = add_total_uncertainties(
+                l2_summed_intensity_dataset, var
+            )
 
     return l2_summed_intensity_dataset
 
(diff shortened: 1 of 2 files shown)


imap_processing/hit/l2/hit_l2.py
```diff
@@ -29,7 +29,6 @@
 # TODO:
 #  - review logging levels to use (debug vs. info)
 #  - determine where to pull ancillary data. Storing it locally for now
-#  - add function to calculate combined uncertainty and add this to L2 datasets
```
greglucas, line 32: Whittling down this list, nice!

imap_processing/hit/l2/hit_l2.py
```diff
@@ -413,19 +409,67 @@ def add_systematic_uncertainties(
...
 
+def add_total_uncertainties(dataset: xr.Dataset, particle: str) -> xr.Dataset:
+    """
+    Add total uncertainties to the dataset.
+
+    This function calculates the total uncertainties for a given particle
+    by combining the statistical uncertainties and systematic uncertainties.
+
+    The total uncertainties are calculated as the square root of the sum
+    of the squares of the statistical and systematic uncertainties.
```
greglucas, line 433: Just verifying that this is what was given to you? Sometimes you want to just linearly add uncertainties so it would be just `stat + sys`, but this all depends on what the underlying values are.

imap_processing/tests/hit/test_hit_l2.py
```diff
@@ -29,6 +30,8 @@
     reshape_for_sectored,
 )
 
+np.random.seed(42)  # Set a random seed for reproducibility
```
greglucas: You should set this within your tests that uses random numbers explicitly so that this doesn't impact other tests and they can explicitly set the values they need later without impacting the others. The order the tests are run could be random and this assumes they have to be run in a specific order to pop the values off in that specific order for your tests.

imap_processing/tests/hit/test_hit_l2.py
```diff
@@ -459,25 +462,86 @@ def test_calculate_intensities():
...
+        )
+        assert (
+            updated_dataset[f"{particle}_sys_err_plus"].shape == dataset[particle].shape
+        )
+        np.testing.assert_array_equal(
+            updated_dataset[f"{particle}_sys_err_minus"].values, 0
+        )
+        np.testing.assert_array_equal(
+            updated_dataset[f"{particle}_sys_err_plus"].values, 0
+        )
```
greglucas, line 501: Shouldn't these be a function of the random values you passed in?

vmartinez-cu: The systematic uncertainties will be zeros unless changed later in the mission, but that isn't expected to happen. I wasn't sure if I should use random values or the zeros that we expect for now

imap_processing/tests/hit/test_hit_l2.py
vmartinez-cu: Good to know! I didn't realize it could affect other tests

imap_processing/hit/l2/hit_l2.py
vmartinez-cu: The calculation comes from the algorithm doc, but I'll double check with HIT

vmartinez-cu: Just confirmed with HIT that the addition needs to be done in a quadrature

sdhoyt (approved):
Looks good!

PR #1654, opened 2025-04-23
1590 hi l2   pointing set class

### LINKED ISSUE #1590: Hi L2 - Pointing Set class
### Algorithm Description:
Create the HiPointingSet that inherits from imap_processing.ena_maps.ena_maps.PointingSet

tmplummer:
# Change Summary

## Overview
This PR adds the basic HiPointingSet class that will allow projection (binning) of PSET pixels to SkyMap pixels.

## New Files
- imap_processing/hi/l2/__init__.py
   - module __init__ file
   - Adds the HiPointingSet class with `_init_` and `from_cdf` methods
   - Adds test coverage for the HiPointingSet class

## Updated Files
- updated file 1
   - Add another external Hi test file (l1c PSET example file)

Closes: #1590


imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,54 @@
...
+        self.az_el_points = np.column_stack(
+            (
+                np.squeeze(self.data["hae_longitude"]),
+                np.squeeze(self.data["hae_latitude"]),
+            )
+        )
+        self.spatial_coords = "spin_angle_bin"
+
+    @classmethod
+    def from_cdf(cls, cdf_path: str | Path) -> HiPointingSet:
```
greglucas: Would this be beneficial to put on the `PointingSet` superclass for other instruments to use too?

imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    PointingSet object specific to Hi L1C PSet data.
+
+    Parameters
+    ----------
+    dataset : xarray.Dataset
+        Hi L1C pointing set data loaded in an xarray.DataArray.
+    """
+
+    def __init__(self, dataset: xr.Dataset):
```
greglucas: Does your method signature need to be the same as the ABC you inherit from? i.e. do you need to allow for an optional SpiceFrame here?

Would it make sense to call `super().__init__()` to set some of these? It is a bit confusing what is required at what point. I realize I missed this on @nkerman's PRs! So, apologies for only asking this question for you ;)

This is in hi_l2, but Nat put `UltraPointingSet` in ena_maps. Should we try to put all of these over there, or break out the UltraPointingSet into the ultra/ location for consistency?

imap_processing/tests/hi/test_hi_l2.py
```diff
@@ -0,0 +1,42 @@
...
+    """Test suite for HiPointingSet class."""
+
+    def test_init(self, pset_path):
+        """Test coverage for __init__ method."""
+        pset_ds = load_cdf(pset_path)
+        hi_pset = HiPointingSet(pset_ds)
+        assert isinstance(hi_pset, HiPointingSet)
+        assert hi_pset.spice_reference_frame == SpiceFrame.ECLIPJ2000
+        assert hi_pset.num_points == 3600
+        np.testing.assert_array_equal(hi_pset.az_el_points.shape, (3600, 2))
```
laspsandoval: assert hi_pset.az_el_points.shape == (3600, 2)

laspsandoval (approved):
Looks good to me.

imap_processing/hi/l2/hi_l2.py
tmplummer: > Does your method signature need to be the same as the ABC you inherit from? i.e. do you need to allow for an optional SpiceFrame here?

Haha, yeah... good catch. I am also not clear on what attributes are required to be set. I think I have the minimum set. I don't want to call `super().__init__()` as is because I find it to be useless to essentially set everything to 0/empty. @nkerman maybe the `PointingSet.__init__()` method should just be a `pass`?

> This is in hi_l2, but Nat put UltraPointingSet in ena_maps. Should we try to put all of these over there, or break out the UltraPointingSet into the ultra/ location for consistency?

Consistency is good. My personal opinion is that this is Hi specific code, so it should go in a Hi specific module. I now realize that it is not L2 specific code though so maybe just bundling it into the `ena_maps` module makes sense.

nkerman: I think you'll need the values you've specified here, and you should probably specify the tiling_type as well, although it's not strictly necessary for PSETs if you don't plan to use PULL projection. 

I had always imagined that each instrument's PointingSet class would have its own, independent init - there's just too much variation between how the different PSETs are defined (healpix/rect, what the L1C files look like, dimensionality, etc.) to do a generic init. Would it be helpful to have some better documentation about what variables a PointingSet must have?

> I don't want to call super().__init__() as is because I find it to be useless to essentially set everything to 0/empty. @nkerman maybe the PointingSet.__init__() method should just be a pass?

IIRC, all the stuff in the generic `PointingSet`'s `__init__` was added to make mypy happy. If there's a better way to keep mypy from complaining about unset types, I'm happy to drop this/make it a `pass`, but I couldn't figure it out at the time. 

> I now realize that it is not L2 specific code though so maybe just bundling it into the ena_maps module makes sense.

That these would be useful for both L2/L3, and expecting that the other instruments would find commonality/extract shared functionality, is why I put those in the ENA_map repo. But if we want to push everything out into the instrument code dirs, that's fine with me.

imap_processing/hi/l2/hi_l2.py
nkerman: This could potentially combine/replace `UltraPointingSet.from_path_or_dataset` if the reading-in is similar enough between the instruments.

imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,54 @@
...
+import numpy as np
+import xarray as xr
+
+from imap_processing.cdf.utils import load_cdf
+from imap_processing.ena_maps.ena_maps import PointingSet
+from imap_processing.hi.l1c.hi_l1c import N_SPIN_BINS
+from imap_processing.spice.geometry import SpiceFrame
+
+
+class HiPointingSet(PointingSet):
```
nkerman: Is there any world in which it would make sense to make this a subclass of `RectangularPointingSet`?

imap_processing/hi/l2/hi_l2.py
greglucas: I think the way to do this would be to use abstract properties. That way we could ensure that each subclass is setting the given properties and they can do that however they want. That way everyone knows what they need to set though.

imap_processing/hi/l2/hi_l2.py
tmplummer: There is such a world... 😆 
I decided that it didn't make sense in this world since Hi PSET is a 1D array of pixels with a single spatial dimension. I really want to start out with the simplest implementation and add things as needed. Comparing with the `RectangularPointingSet` I see so many differences with very little common code that I decided not to inherit from that class.

imap_processing/hi/l2/hi_l2.py
tmplummer: > IIRC, all the stuff in the generic PointingSet's __init__ was added to make mypy happy.

Thanks, that makes sense.

> I think the way to do this would be to use abstract properties.

I'll make a new branch and see if using abstract properties makes more sense. I feel like the notion of abstract attributes is missing from the ABC. Would you suggest making the attributes private and then just implement property setters and getters for each of those?

imap_processing/hi/l2/hi_l2.py
nkerman: SGTM!

tmplummer:
@greglucas @nkerman, I have rebased this PR onto the `PointingSet` refactor. There is additional refactoring as I realized that some additional attributes could be contributed to properties. I also extracted a new abstract `HealpixPointingSet` class. Let me know what you think.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -253,6 +259,18 @@ def __init__(
         if len(np.unique(self.data["epoch"].values)) > 1:
             raise ValueError("Multiple epochs found in the dataset.")
 
+    @property
+    def num_points(self) -> int:
```
nkerman, line 263: Nice, I like this as a property!

imap_processing/ena_maps/ena_maps.py
```diff
@@ -412,10 +430,40 @@ def __init__(
                 self.sky_grid.el_grid.ravel(),
             )
         )
-        self.num_points = self.az_el_points.shape[0]
 
 
-class UltraPointingSet(PointingSet):
+class HealPixPointingSet(PointingSet, ABC):
+    """
+    Abstract base class for HealPix pointing sets.
```
nkerman: Nitpick: You usually use HEALPix, rather than HealPix here. I have a hard time with mixed case, so I think it's arguably better to use Healpix, with just the H capitalized, in documentation/longer comments. I think lowercase healpix is also fine for short comments. 

Note: I'm sure I've strayed from this rule myself! It's just easier to see in unfamiliar comments.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -412,10 +430,40 @@ def __init__(
...
+
+    Defines additional properties and absract properties that are required
+    for a PointingSet instance to be used with the match_coords_to_indices
+    function.
+    """
+
+    tiling_type: SkyTilingType = SkyTilingType.HEALPIX
+
+    @property
```
nkerman, line 446: WRT all the properties methods:

Can you think of a situation where the number of pixels would change after initialization? Why wouldn't we want to just set them there?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -503,6 +537,32 @@ def __init__(
...
+
+        Returns
+        -------
+        num_points: int
+            The number of HEALPix pixels in the pointing set.
+        """
+        return self.data[CoordNames.HEALPIX_INDEX.value].size
+
+    @property
+    def nested(self) -> bool:
```
nkerman, line 553: I like this implementation a lot - thanks!

imap_processing/hi/l2/__init__.py
nkerman, line 1: Do we need __init__.py in each of our subdirectories? I thought this wasn't necessary in Python 3.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -519,6 +579,27 @@ def __repr__(self) -> str:
         )
 
 
+class HiPointingSet(PointingSet):
```
nkerman, line 582: It's fantastic how small this class definition can be!

nkerman (commented):
I really like most of this reorganization! There are a couple small change requests, and I'm interested whether some of the `@properties` should stay in the inits. Otherwise looks good by me.

imap_processing/hi/l2/__init__.py
tmplummer: I will remove this since there is no longer any `hi.l2` code associated with this PR.

I didn't know about that change in 3.3+. It looks like most the other directories in `imap_processing` still have them and it is still recommended to have them:
https://stackoverflow.com/questions/37139786/is-init-py-not-required-for-packages-in-python-3-3

imap_processing/ena_maps/ena_maps.py
tmplummer: No, this change wasn't implemented to allow the number of pixels to change. I just find it better practice to make a value that is only dependent on another property/attribute a property. It makes it a bit clearer to the user where it comes from and also makes it so that a user can not set this value thinking that it would somehow redefine the class instance.

There can be an argument made that then this value is recomputed each time it is accessed. I figured that with such a light computation, this still made sense.

greglucas: As a note, if there is an expensive computation you only want to compute once there is a `cached_property` decorator: https://docs.python.org/3/library/functools.html#functools.cached_property

greglucas (approved):

PR #1655, opened 2025-04-24
TST/CI: Add pre-download utility and cache mechanism for data

greglucas:
# Change Summary

## Overview

We are running into race conditions for downloading the tests while using pytest-xdist across multiple processes. This adds a command line option to pre-fetch the data in a single process before running the tests, which will then see the files already exist and not download them.

Additionally, add a cache so that the GitHub Actions can just restore the data files directly rather than needing to go and download them from our servers all the time.


laspsandoval (approved):
This looks good. I think it should work.

PR #1656, opened 2025-04-24
IDEX l1b - modify copy of test dataset

lacoak21:
# Change Summary

## Overview
IDEX l1b validation test uses an l1b dataset fixture that has a "module" scope. I _think_ recent changes in the way we run our github tests have caused this fixture to be rerun after the l1a dataset was modified for another test. That other test modifies the dataset in a way that is supposed to cause an error. This then causes an error in the set up of the l1b validation test. 

This update should fix this by decoupling the l1a and l1b datasets. 

## Updated Files
   - Use a copy of the l1a dataset in the l1b fixture

### DIFF
--- a/imap_processing/tests/idex/conftest.py
+++ b/imap_processing/tests/idex/conftest.py
@@ -46,6 +46,7 @@ def decom_test_data_sci() -> xr.Dataset:
     return PacketParser(TEST_L0_FILE_SCI).data[0]
 
 
+@pytest.fixture
 def decom_test_data_catlst() -> xr.Dataset:
     """Return a ``xarray`` dataset containing the catalog list summary data.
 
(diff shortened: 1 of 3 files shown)


greglucas (commented):
Is there a need for this to be stored at the module-level, or should it just be remade for each test?

I'm a bit confused where the deepcopy needs to occur here and which test is interlinked. Ideally there wouldn't be any coupling and the mocks would be separate if we can.

lacoak21:
> Ideally there wouldn't be any coupling and the mocks would be separate if we can.

OK that makes sense. I will decouple all IDEX tests.

greglucas (approved):

greglucas:
Can you rebase this onto the latest main branch and push the changes to see if the remote tests pass now?

lacoak21:
How did you get the failing test to pass @greglucas ?

greglucas:
> How did you get the failing test to pass @greglucas ?

🤔 I'm not sure what you mean?

PR #1658, opened 2025-04-24
Remove obs day calculation from glows l1a

maxinelasp:
# Change Summary
Remove all the estimation work of observational day/repointing number from glows since we can now get it from the L0 file. 

Generally deleting, one small change in the repointing CDF piece per conversation with @greglucas .

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -129,9 +129,12 @@ def write_cdf(
             f"The Data_version attribute {version} does not match expected format vXXX."
         )
 
-    # TODO: Do we need to retrieve this from the dataset?
-    if repointing is None:
-        repointing = dataset.attrs.get("Repointing", None)
+    if "repointing" in dataset.attrs:
+        # If provided, assume that we want to override the passed-in value
+        repointing = dataset.attrs["repointing"]
+    elif repointing is not None:
+        dataset.attrs["repointing"] = repointing
+
     repointing_int = int(repointing[-5:]) if repointing else None
     science_file = imap_data_access.ScienceFilePath.generate_from_inputs(
         instrument=instrument,
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -66,9 +66,6 @@ def load_cdf(
 
 def write_cdf(
     dataset: xr.Dataset,
-    *,
-    start_date: str | None = None,
-    repointing: str | None = None,
     **extra_cdf_kwargs: dict,
 ) -> Path:
     """
@@ -81,22 +78,16 @@ def write_cdf(
     determined by the "Logical_source" attribute.  The version is determined from
     "Data_version".
 
-    The start_date and repointing parameters are used to override the computed values.
+    The start_date and repointing attributes in the dataset are used to override the
+    computed values.
+
     If these are not included, start_date is generated from the first epoch in the
-    dataset and repointing is retrieved from the dataset attributes (or not included if
-    the attribute is not present).
+    dataset and repointing is not included if the attribute is not present or None.
 
     Parameters
     ----------
     dataset : xarray.Dataset
         The dataset object to convert to a CDF.
-    start_date : str
-        In the output filename, this is the start_date. If not provided or None,
-        the first epoch in the dataset is used.
-    repointing : str
-        The repointing number to use in the output filename. If not provided or None,
-        the repointing number is retrieved from the dataset attributes. If none are
-        provided, it is excluded from the filename.
     **extra_cdf_kwargs : dict
         Additional keyword arguments to pass to the ``xarray_to_cdf`` function.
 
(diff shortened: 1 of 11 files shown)


imap_processing/cdf/utils.py
```diff
@@ -129,9 +129,12 @@ def write_cdf(
...
         )
 
-    # TODO: Do we need to retrieve this from the dataset?
-    if repointing is None:
-        repointing = dataset.attrs.get("Repointing", None)
+    if "repointing" in dataset.attrs:
+        # If provided, assume that we want to override the passed-in value
+        repointing = dataset.attrs["repointing"]
+    elif repointing is not None:
+        dataset.attrs["repointing"] = repointing
```
greglucas: I feel like this is reverse of what usually happens?

This new method
1. dataset.attrs
2. keyword argument
3. None

I think you usually want the opposite, if you've specified it as a keyword argument you want that to take precedence and override anything else that was there.

Previous fallback order:
1. keyword argument
2. dataset.attrs
3. None

One other new thing is that you are setting the `dataset.attrs["repointing"]` global attribute. I think this is probably a good thing to make sure it is all consistent. Do we want lowercase now while it was "Repointing" before?

maxinelasp: Happy to switch it, I think it is easier for people to override this in specific weird cases but it's also easy enough to override the keyword. 

I didn't think of repointing vs Repointing, I didn't notice that I lowercased it. I think you're right that we want "Repointing" to match even though I hate that the global attrs are all capitalized

maxinelasp: I figure most people will just let the attribute get set here and not use it in processing at all

greglucas: > I think you're right that we want "Repointing" to match even though I hate that the global attrs are all capitalized

I completely agree!

> I figure most people will just let the attribute get set here and not use it in processing at all

Care to just make that a rule here then? I think it will make this all easier if we can start simplifying these implementations and be able to explicitly track where something was set.

maxinelasp: I could imagine a world where someone would want to explicitly set that value for some reason, but in general I think it makes sense to ask people to let it get set here and override the write_cdf call if needed.

maxinelasp: Are you suggesting a code change or is what I have good and you're talking about a rule of thumb for the team?

greglucas: > I could imagine a world where someone would want to explicitly set that value for some reason, but in general I think it makes sense to ask people to let it get set here and override the write_cdf call if needed.

> Are you suggesting a code change or is what I have good and you're talking about a rule of thumb for the team?

Yes, both :) I have also been saying this to others too where if there isn't a current need for multiple ways of doing something lets just implement the one you think is best for now. Then later we can add the other option if/when it is needed.

My initial question/suggestion would be whether this could be made similar to the version handling which I think we had get set as an attribute in the preprocessing step, but then read in here without a keyword argument.

maxinelasp: Oh, I see what you're saying. I didn't make the connection between the version and this change. I like that idea better, I'll put up a revision to set the attributes in post_processing and then retrieve them both out of the attributes instead.

imap_processing/cli.py
```diff
@@ -488,10 +488,10 @@ def post_processing(
         products = []
         for ds in datasets:
             ds.attrs["Data_version"] = self.version
+            ds.attrs["Repointing"] = self.repointing
```
greglucas, line 491: Do we only want to set this if it is not None?

imap_processing/tests/glows/test_glows_l1a_data.py
```diff
@@ -43,17 +43,15 @@ def test_histogram_list(histogram_test_data, decom_test_data):
 def test_histogram_obs_day(packet_path):
     l1a = glows_l1a(packet_path)
 
-    assert len(l1a) == 3
+    assert len(l1a) == 2
```
greglucas, line 46: Is this an expected reduction in number of files? Below it looks like you are only comparing against the first item, do you want to say anything about the second item?

greglucas (approved):
Nice simplification, I like this style of injecting the attributes in the post-processing steps as needed.

imap_processing/tests/glows/test_glows_l1a_data.py
maxinelasp: Yes as we are only producing one histogram file from the test input now. Before, two were being produced to test the observation day splitting, but now that is handled pre-processing so we only get one hist file back. The DE file (second return) is tested elsewhere.

imap_processing/cli.py
maxinelasp: I thought we would want this regardless, but I could see it either way. I think we probably want the same global attributes, but it might not make sense to include this one for instruments which don't have repointings. Happy to add an "if not None" if you vote the other way, I'm only slightly pro including it in all cases.

imap_processing/tests/glows/test_glows_l1a_data.py
greglucas: Can you bring this into slack to ask the team? I dont have a good feel for it myself so could go either way too. I lean slightly against including it because in-situ instruments have no notion of repointing so itd be an odd attribute for someone to see I think.

PR #1661, opened 2025-04-25
I-ALiRT -  MAG l1b equivalent

laspsandoval:
# Change Summary

## Overview
Performs the following steps:
1. Fetch and select the required Engineering (ENG) calibration
2. Transform to URF
3. Apply time shift

## Updated Files
- parse_mag.py
   - Fetch and select the required Engineering (ENG) calibration
   - Transform to URF
   - Apply time shift
- mag_l1b.py
   - docstring change

## Testing
- test_parse_mag.py

### DIFF
--- a/imap_processing/ialirt/l0/parse_mag.py
+++ b/imap_processing/ialirt/l0/parse_mag.py
@@ -2,6 +2,7 @@
 
 import logging
 from pathlib import Path
+from typing import cast
 
 import numpy as np
 import xarray as xr
@@ -192,14 +193,14 @@ def get_time(
     ][pkt_counter == 2]
 
     time_data = {
-        "pri_coarsetm": int(pri_coarsetm.item()),
-        "pri_fintm": int(pri_fintm.item()),
-        "sec_coarsetm": int(sec_coarsetm.item()),
-        "sec_fintm": int(sec_fintm.item()),
+        "pri_coarsetm": cast(int, pri_coarsetm.item()),
+        "pri_fintm": cast(int, pri_fintm.item()),
+        "sec_coarsetm": cast(int, sec_coarsetm.item()),
+        "sec_fintm": cast(int, sec_fintm.item()),
     }
 
-    primary_time = TimeTuple(int(pri_coarsetm), int(pri_fintm))
-    secondary_time = TimeTuple(int(sec_coarsetm), int(sec_fintm))
+    primary_time = TimeTuple(time_data["pri_coarsetm"], time_data["pri_fintm"])
+    secondary_time = TimeTuple(time_data["sec_coarsetm"], time_data["sec_fintm"])
     time_data["pri_met"] = primary_time.to_seconds()
     time_data["primary_ttj2000ns"] = met_to_ttj2000ns(time_data["pri_met"])
     time_data["primary_epoch"] = shift_time(
--- a/imap_processing/ialirt/l0/parse_mag.py
+++ b/imap_processing/ialirt/l0/parse_mag.py
@@ -191,7 +191,6 @@ def get_time(
     secondary_time = TimeTuple(int(sec_coarsetm), int(sec_fintm))
     time_data["pri_met"] = primary_time.to_seconds()
     time_data["primary_ttj2000ns"] = met_to_ttj2000ns(time_data["pri_met"])
-    # TODO: is this ok?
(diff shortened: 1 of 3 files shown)


imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -166,17 +192,105 @@ def get_time(grouped_data: xr.Dataset, group: int, pkt_counter: xr.DataArray) ->
...
+        "pri_coarsetm": int(pri_coarsetm.item()),
+        "pri_fintm": int(pri_fintm.item()),
+        "sec_coarsetm": int(sec_coarsetm.item()),
+        "sec_fintm": int(sec_fintm.item()),
     }
 
+    primary_time = TimeTuple(int(pri_coarsetm.item()), int(pri_fintm.item()))
+    secondary_time = TimeTuple(int(sec_coarsetm.item()), int(sec_fintm.item()))
+    time_data["pri_met"] = primary_time.to_seconds()
+    time_data["primary_ttj2000ns"] = met_to_ttj2000ns(time_data["pri_met"])
```
maxinelasp: Do you need the epoch and the values for `pri_met` and `primary_ttj2000ns` to match? Right now you're shifting `primary_epoch` only.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -166,17 +192,105 @@ def get_time(grouped_data: xr.Dataset, group: int, pkt_counter: xr.DataArray) ->
...
+    -------
+    updated_vector_mago : numpy.ndarray
+        Calibrated mago vector.
+    updated_vector_magi : numpy.ndarray
+        Calibrated magi vector.
+    time_data : dict
+        Time data.
+    """
+    # Get calibration data
+    calibration_dataset = load_cdf(
```
maxinelasp: Do you expect this to come from AWS at any point? It might make sense to have it passed into this function so you can easily handle that.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -166,17 +192,105 @@ def get_time(grouped_data: xr.Dataset, group: int, pkt_counter: xr.DataArray) ->
...
+    calibration_matrix_magi, time_shift_magi = retrieve_matrix_from_l1b_calibration(
+        calibration_dataset, is_mago=False
+    )
+
+    # Get time values for each group.
+    time_data = get_time(
+        grouped_data, group, pkt_counter, time_shift_mago, time_shift_magi
+    )
+
+    input_vector_mago = np.array(
```
maxinelasp, line 263: Are we expecting primary to always equal mago, and secondary to always equal magi?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -166,17 +192,105 @@ def get_time(grouped_data: xr.Dataset, group: int, pkt_counter: xr.DataArray) ->
...
+    primary_time = TimeTuple(int(pri_coarsetm.item()), int(pri_fintm.item()))
+    secondary_time = TimeTuple(int(sec_coarsetm.item()), int(sec_fintm.item()))
+    time_data["pri_met"] = primary_time.to_seconds()
+    time_data["primary_ttj2000ns"] = met_to_ttj2000ns(time_data["pri_met"])
+    time_data["primary_epoch"] = shift_time(
+        time_data["primary_ttj2000ns"], time_shift_mago
+    )
+    time_data["sec_met"] = secondary_time.to_seconds()
+    time_data["secondary_ttj2000ns"] = met_to_ttj2000ns(time_data["sec_met"])
+    time_data["secondary_epoch"] = shift_time(
```
maxinelasp, line 209: Do you need all these time_data values? I would be inclined to just return the shifted TTJ2000 timestamps and if needed, convert the time formats later. But I'm not sure how these are going to be used.

maxinelasp (approved):

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: Yes. Confirmed with Alastair.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: I was storing every step, just in case, but I think you are right. I really only need to keep epoch.

PR #1662, opened 2025-04-25
1657 mapping   pointingset class should use abstractproperty

### LINKED ISSUE #1657: Mapping - PointingSet class should use abstractproperty
### Algorithm Description:

Try using `@abstractproperty` for the required attributes in the PointingSet abstract base class.

tmplummer:
# Change Summary

## Overview
I experimented with making some abstractmethod properties in the PointingSet base class, but what we really want are required instance variables. I settled on defining what I think is the minimum set of attributes into class body with type-hints only. I also identified some common code amongst the base class and child classes and moved that into the base class `__init__` method. 

## Updated Files
   - Move required attributes into Class body
   - Move common code into ABC class `__init__` method and call `super().__init__()` in subclasses.
   - Unify `__init__()` method signatures across classes.
   - Remove some unused/duplicate class variables.
   - Create common `from_cdf()` method in the base class.
   - Remove `from_path_or_dataset()` method from UltraPointingSet
   - Remove portion of test from `test_from_cdf` that checked that dataset mutation did not occur.
   - Update instantiation calls to match common `__init__` signature.
   - Update ultra_l2 code to call different method when instantiating UltraPointingSet from str or xr.Dataset

Closes: #1657

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import json
 import logging
-import pathlib
 from abc import ABC, abstractmethod
 from enum import Enum
 from pathlib import Path
@@ -220,14 +219,40 @@ class PointingSet(ABC):
         The reference Spice frame of the pointing set.
     """
 
+    # Attributes that are set in the ABC __init__ method
+    data: xr.Dataset
+    spice_reference_frame: geometry.SpiceFrame
+    # Attributes required to be set in a subclass
+    az_el_points: np.ndarray
+    num_points: int
+    spatial_coords: tuple[str, ...]
+
     @abstractmethod
-    def __init__(self, dataset: xr.Dataset, spice_reference_frame: geometry.SpiceFrame):
+    def __init__(
+        self,
+        dataset: xr.Dataset,
+        spice_reference_frame: geometry.SpiceFrame = geometry.SpiceFrame.IMAP_DPS,
+    ):
         """Abstract method to initialize the pointing set object."""
         self.spice_reference_frame = spice_reference_frame
-        self.num_points = 0
-        self.az_el_points = np.zeros((self.num_points, 2))
-        self.data = xr.Dataset()
-        self.spatial_coords: tuple[str, ...] = ()
+        self.data = dataset
+
+    @classmethod
+    def from_cdf(cls, cdf_path: Path) -> PointingSet:
+        """
+        Generate a PointingSet object from a CDF file.
+
+        Parameters
+        ----------
+        cdf_path : str | Path
+            Location of Pointing Set CDF file.
+
+        Returns
+        -------
+        pointing_set : PointingSet
+            Input CDF file data loaded into PointingSet.
+        """
+        return cls(load_cdf(cdf_path))
 
     @property
     def unwrapped_dims_dict(self) -> dict[str, tuple[str, ...]]:
@@ -292,8 +317,8 @@ class RectangularPointingSet(PointingSet):
 
     Parameters
     ----------
-    l1c_dataset : xr.Dataset | pathlib.Path | str
-        L1c xarray dataset containing the pointing set data or the path to the dataset.
+    dataset : xr.Dataset
+        L1c xarray dataset containing the pointing set data.
         Currently, the dataset is expected to be tiled in a rectangular grid,
         with data_vars indexed along the coordinates:
             - 'epoch' : time value (1 value per PSET)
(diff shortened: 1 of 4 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -220,14 +219,43 @@ class PointingSet(ABC):
         The reference Spice frame of the pointing set.
     """
 
+    # The minimum set of class attributes for any PointingSet to function with
```
nkerman, line 228: This makes defining a new PSET class way more clear, thanks Tim

imap_processing/ena_maps/ena_maps.py
```diff
@@ -220,14 +219,43 @@ class PointingSet(ABC):
...
+    # a SkyMap using only the PUSH method of projecting are defined here.
```
nkerman, line 229: I want to say that it will also need to define an epoch to work with PUSH-ing, and I'd rather we always defined tiling_type for PSETs as well. I think it's one of their key properties, at least conceptually.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -418,27 +438,26 @@ class UltraPointingSet(PointingSet):
...
     ):
-        # Store the reference frame of the pointing set
-        self.spice_reference_frame = spice_reference_frame
-
-        # Read in the data and store the xarray dataset as data attr
-        if isinstance(l1c_dataset, (str, pathlib.Path)):
-            self.data = load_cdf(pathlib.Path(l1c_dataset))
-        elif isinstance(l1c_dataset, xr.Dataset):
-            self.data = l1c_dataset
+        super().__init__(dataset, spice_reference_frame)
```
nkerman, line 456: Definitely more neat from an OO perspective!

nkerman (commented):
Thanks Tim, all of this seems totally reasonable - I hope `mypy` will learn to accept the changes. It's my experience that `mypy` must be counseled through these trying times 🤣

imap_processing/ena_maps/ena_maps.py
tmplummer: Maybe an oversight, but `epoch` isn't strictly necessary as implemented. `match_coords_to_indices` uses `obj.data.epoch.values`: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/ena_maps/ena_maps.py#L141

For `tiling_type`, I am not sure what to do for Hi. The Hi PSET really doesn't have a pixel width and I want it to fail if someone tried to use the PULL method. I'll think about this and see if I can get some feedback from Paul about how to think about Hi Pset pixels.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -220,14 +225,43 @@ class PointingSet(ABC):
...
+        ----------
+        cdf_path : str | Path
+            Location of Pointing Set CDF file.
+
+        Returns
+        -------
+        pointing_set : PointingSet
+            Input CDF file data loaded into PointingSet.
+        """
+        return cls(load_cdf(cdf_path))
```
greglucas: Now that you've simplified everything else, should we go even further here and let this be handled in the super initializer and not worry about a separate classmethod? I'm indifferent, just figured I would suggest it in case you care one way or another.

```
if isinstance(dataset, (str, Path)):
    dataset = load_cdf(dataset)
self.dataset = dataset
```

greglucas (approved):

imap_processing/ena_maps/ena_maps.py
greglucas: Could the width be the amount of nutation around the DPS "mean" axis? I think this is what Lo is trying to figure out by making the 40 wide pointing set. I wonder if rather than making those 40 individual bins if you could somehow set the "smear" as your width of a pixel.

tmplummer: It isn't quite that simple. The 0.25 degree nutation is much less than the PSF of the collimator which has a FWHM of ~4 degrees. The Hi PSET is binning events into 0.1 degree bins and simply ignoring the PSF of the collimator. I think that Paul's approach just relies on shear statistics and the coarser binning of the Maps to overcome this challenge. From the alg. document discussing the angular response calibration file:

> The single important thing to know about this file is that the SDC will never have to use it for anything;
indeed, ground processing at the SDC can utterly ignore this file’s presence. Code up through Level 2 will
never attempt to deconvolve the instrument point spread function (PSF), or indeed to do anything with
the instrument PSF at all, because it’s really hard when Poisson statistics are so bad as to make Fourier
transforms nonviable.

My understanding is that Hi will NEVER want/need to use the PULL method. Furthermore, the simplification of just computing the Pset pixel coordinates in L1C, allows for the `HiPointingSet` to be a very stripped down class where we don't need to care about pixel spacing, pixel edges, or tiling type.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -214,20 +219,41 @@ class PointingSet(ABC):
...
-        self.spatial_coords: tuple[str, ...] = ()
+
+        if isinstance(dataset, (str, Path)):
+            dataset = load_cdf(dataset)
+        self.data = dataset
+
+        # A PSET must have a single epoch
+        if len(np.unique(self.data["epoch"].values)) > 1:
+            raise ValueError("Multiple epochs found in the dataset.")
+        self.epoch = self.data["epoch"].values[0]
```
greglucas: This one seems like it could be a property getter, maybe keeping the dataset validation here. The typehint above is also `np.ndarray`, but I think it is a scalar datetime64 since you're plucking out a single value?

```python
@property
def epoch(self):
    return self.data["epoch"].values[0]
```

PR #1666, opened 2025-04-28
HIT L2 - Add CDF attributes to yaml

### LINKED ISSUE #1075: HIT L2 - create CDF yaml
Define attributes needed for L2 products
Create yaml file with standard, summed, and sectored flux CDF attributes

vmartinez-cu:
This PR adds CDF attributes for all three L2 products to a yaml file. There are many data variables in these products so I created default attributes where possible to avoid repetition while balancing readability.

Closes #1075


sdhoyt (approved):
😵‍💫 Looks good!

PR #1665, opened 2025-04-25
I-ALiRT - SC packet check

laspsandoval:
# Change Summary

## Overview
I hadn't looked at the SC packet in a while. This is just an update based on finalizing the instrument packet definitions, any reordering that happened since I've last looked, and the incorporation of a spacecraft packet with MAG non-zero values.

## Updated Files
- ialirt.xml
   - Order change
- ialirt_codicehi.xml
   - Fixed small typo for the description
- ialirt_codicelo.xml
   - Fixed small typo for the description

## Testing
- conftest.py
   - Added data to download
- test_data/l0 directory renamed
- test_decom_ialirt.py
   - Update to test based on new data

Everything else in tests/unit is just a path change.

PR #1667, opened 2025-04-28
I-ALiRT - Script for binary data

laspsandoval:
# Change Summary

## Overview
Create the script for generating the binary blob from each packet. This will be imported into the lambda in sds-data-manager.

## New Files
- generate_binary.py
   - Generate the binary blob data.

## Testing
- test_generate_binary.py
   - Test generate_binary.py

Note: PR tests will fail until I-ALiRT - SC packet check #1665 is merged in.

### DIFF
--- a/imap_processing/ialirt/l0/generate_binary.py
+++ b/imap_processing/ialirt/l0/generate_binary.py
@@ -24,10 +24,10 @@ def generate_binary(
     -------
     binary_blob_data : list
         Binary blob data for each packet.
-    time_data : list
+    met_data : list
         SCLK time in seconds.
     """
-    time_data = []
+    met_data = []
     binary_blob_data = []
 
     # Set up the parser from the input packet definition
@@ -40,11 +40,11 @@ def generate_binary(
             binary_blob = packet.raw_data
             # Subsecond time conversion specified in 7516-9054 GSW-FSW ICD.
             # Value of SCLK subseconds, unsigned, (LSB = 1/256 sec)
-            time = (
+            met = (
                 packet.user_data["SC_SCLK_SEC"]
                 + packet.user_data["SC_SCLK_SUB_SEC"] * 256
             )
             binary_blob_data.append(binary_blob)
-            time_data.append(time)
+            met_data.append(met)
 
-    return binary_blob_data, time_data
+    return binary_blob_data, met_data
--- a/imap_processing/ialirt/l0/generate_binary.py
+++ b/imap_processing/ialirt/l0/generate_binary.py
@@ -38,10 +38,7 @@ def generate_binary(
         for packet in packet_generator:
             # Subsecond time conversion specified in 7516-9054 GSW-FSW ICD.
             # Value of SCLK subseconds, unsigned, (LSB = 1/256 sec)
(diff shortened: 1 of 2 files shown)


imap_processing/ialirt/l0/generate_binary.py
```diff
@@ -0,0 +1,54 @@
...
+    with packet_file.open("rb") as binary_data:
+        packet_generator = packet_definition.packet_generator(binary_data)
+
+        # Iterate over the packets and access the raw binary data
+        for packet in packet_generator:
+            # Subsecond time conversion specified in 7516-9054 GSW-FSW ICD.
+            # Value of SCLK subseconds, unsigned, (LSB = 1/256 sec)
+            met = (
+                packet.user_data["SC_SCLK_SEC"]
+                + packet.user_data["SC_SCLK_SUB_SEC"] * 256
```
greglucas: * Avoid using `.user_data` or `.header` from space_packet_parser if possible and just access the underlying dictionary.
* The `SC_SCLK_SUB_SEC` should be divided by 256 because it is a fraction of a second.

```suggestion
                packet["SC_SCLK_SEC"]
                + packet["SC_SCLK_SUB_SEC"] / 256
```

imap_processing/ialirt/l0/generate_binary.py
```diff
@@ -0,0 +1,54 @@
...
+            )
+            ingest_data.append(
+                {
+                    "apid": 478,
```
greglucas, line 44: Do you have a definition for the safemode packet / are you storing anything other than apid 478? I'm wondering if this should be `packet["APID"]` or whatever the field is in your definition.

laspsandoval: Thanks for your suggestions! It will only be apid 478. I'm not working with the safemode packets.

greglucas: :+1: Then do you even need the apid field? I guess it is there so you can get to a sort-key after this.

imap_processing/ialirt/l0/generate_binary.py
```diff
@@ -0,0 +1,51 @@
+"""I-ALiRT data to populate binary blob database."""
```
torimarbois, line 1: Nit - should this say "I-ALiRT _**code**_ to populate.."?

torimarbois (commented):
Just wanting to understand - why is this addition necessary? Do we not have code yet that unpacks the I-ALiRT packet in order to pass it into the different instrument processing steps? I thought we already had 'unpacking' code.

laspsandoval:
> Just wanting to understand - why is this addition necessary? Do we not have code yet that unpacks the I-ALiRT packet in order to pass it into the different instrument processing steps? I thought we already had 'unpacking' code.

Good question! The scientists wanted a way to look at intermediate data. So we are creating a database with "binary blob" data that they can query and then do their own processing on.  But now that you mention it, maybe they should just use an API in s3.

Nice catch!

laspsandoval:
Closing this ticket since we decided to eliminate the "ingest" DB.

PR #1674, opened 2025-05-01
IDEX housekeeping - Recalculate epoch time with respect to fine grained time

lacoak21:
# Change Summary

## Overview
IDEX epochs are calculated with shcoarse and shfine time values. Fix housekeeping to calculate epoch by using both. 

## Updated Files
- imap_processing/idex/idex_l0.py
   - Use derived values for housekeeping. 
   - move epoch calculation out of DustEventClass as it now needed for housekeeping as well. 

## Testing
Add checks for specific variables needed in housekeeping

### DIFF
--- a/imap_processing/idex/idex_l0.py
+++ b/imap_processing/idex/idex_l0.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 
 def decom_packets(
     packet_file: Union[str, Path],
-) -> tuple[list[Any], dict[int, Dataset]]:
+) -> tuple[list[Any], dict[int, Dataset], dict[int, Dataset]]:
     """
     Decom IDEX data packets using IDEX packet definition.
 
@@ -27,7 +27,7 @@ def decom_packets(
     -------
     Tuple[list, dict]
         Returns a list of all unpacked science data and a dictionary of datasets
-        indexed by their APIDs.
+        indexed by their APIDs, one for raw and derived values.
 
     Notes
     -----
@@ -40,8 +40,15 @@ def decom_packets(
     hk_xtce_file = f"{xtce_base_path}/idex_housekeeping_packet_definition.xml"
 
     science_decom_packet_list = decom.decom_packets(packet_file, science_xtce_file)
-    datasets_by_apid = packet_file_to_datasets(
+    raw_datasets_by_apid = packet_file_to_datasets(
+        packet_file, hk_xtce_file, use_derived_value=False
+    )
+    derived_datasets_by_apid = packet_file_to_datasets(
         packet_file, hk_xtce_file, use_derived_value=True
     )
 
-    return list(science_decom_packet_list), datasets_by_apid
+    return (
+        list(science_decom_packet_list),
+        raw_datasets_by_apid,
+        derived_datasets_by_apid,
+    )
--- a/imap_processing/idex/idex_l0.py
+++ b/imap_processing/idex/idex_l0.py
@@ -40,6 +40,8 @@ def decom_packets(
     hk_xtce_file = f"{xtce_base_path}/idex_housekeeping_packet_definition.xml"
 
     science_decom_packet_list = decom.decom_packets(packet_file, science_xtce_file)
-    datasets_by_apid = packet_file_to_datasets(packet_file, hk_xtce_file)
+    datasets_by_apid = packet_file_to_datasets(
+        packet_file, hk_xtce_file, use_derived_value=True
+    )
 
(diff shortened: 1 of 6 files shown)


imap_processing/idex/idex_l0.py
```diff
@@ -40,6 +40,8 @@ def decom_packets(
     hk_xtce_file = f"{xtce_base_path}/idex_housekeeping_packet_definition.xml"
 
     science_decom_packet_list = decom.decom_packets(packet_file, science_xtce_file)
-    datasets_by_apid = packet_file_to_datasets(packet_file, hk_xtce_file)
+    datasets_by_apid = packet_file_to_datasets(
+        packet_file, hk_xtce_file, use_derived_value=True
```
greglucas, line 47: The `use_derived_value` is typically used for `l1b` housekeeping data, so I think some folks are calling this routine twice in one processing step to produce both `l1a` and `l1b` files. Does IDEX want both versions, the raw and derived or just derived?

imap_processing/idex/idex_l1a.py
```diff
@@ -89,13 +89,15 @@ def __init__(self, packet_file: Union[str, Path]) -> None:
             logger.info("Processing IDEX L1A Event Message data.")
             data = datset_by_apid[IDEXAPID.IDEX_EVT]
             data.attrs = self.idex_attrs.get_global_attributes("imap_idex_l1a_evt")
+            data["epoch"] = calculate_idex_epoch_time(data["shcoarse"], data["shfine"])
```
greglucas: What about passing in just `data` and letting the function calculate the value based upon that? May be complicated because of the different lowercase/uppercase conventions between call sites though so probably not worth it.

imap_processing/idex/idex_l1a.py
```diff
@@ -228,6 +230,39 @@ def _read_waveform_bits(waveform_raw: str, high_sample: bool = True) -> list[int
...
+    Returns
+    -------
+    numpy.ndarray[numpy.int64]
+        The mission elapsed time converted to nanoseconds since the J2000 epoch
+        in the terrestrial time (TT) timescale.
+    """
+    # Number of microseconds since the last second
+    microseconds_since_last_second = 20 * shfine_time
+    # Get met time in seconds including shfine
+    met = shcoarse_time + microseconds_since_last_second * 1e-6
```
greglucas: Definitely a nit here, but I find the split out into another quantity slightly more confusing than just inline multiplying by the 20 microseconds. Just suggesting it here in case that reads better to you too but it is fine as-is as well.
```suggestion
    # Get met time in seconds including shfine which is the number of 20 microseconds ticks
    met = shcoarse_time + shfine_time * 20e-6
```

imap_processing/tests/idex/test_idex_l1a.py
```diff
@@ -292,27 +292,35 @@ def test_decode_sub_frame_psel_3():
...
+def test_catlst_dataset(decom_test_data_catlst: xr.Dataset):
+    """Verify that the dataset contains what we expect and can be written to a cdf.
 
     Parameters
     ----------
     decom_test_data_catlst : xarray.Dataset
         The dataset to test with
     """
+    assert "shcoarse" in decom_test_data_catlst
+    assert "shfine" in decom_test_data_catlst
```
greglucas: Can you also assert on an expected data value? The new function you split out isn't tested individually from what I can see.

imap_processing/idex/idex_l0.py
lacoak21: They want just the derived value. Might be worth processing twice and returning both.

imap_processing/idex/idex_l1a.py
lacoak21: Yeah I thought it might just be easier to pass them in explicitly.

imap_processing/idex/idex_l1a.py
lacoak21: I think I agree with you on this one.

imap_processing/idex/idex_l0.py
lacoak21: Actually for the catalog list housekeeping packet I think they probably want raw but for the event message packet they want the derived value. 

I could return the raw version for the catalog list and event message packet and also return the l1b derived event message packet.

greglucas: That sounds confusing to me. I'd suggest just doing l1a and l1b for all of these and then letting folks downstream do whatever they want with it.

greglucas (approved):

PR #1676, opened 2025-05-01
PULL exposure_factor rather than PUSH

### LINKED ISSUE #1675: Ultra L2: PULL project Exposure factor, rather than PUSH
To get a good approximation of the `sum_over_different_PSETs(mean_over_pixels_in_PSET(exposure_factor))` approach, we can simply use PULL projection to get the summed exposure_factor from multiple PSETs into Map pixels.

nkerman:
# Change Summary

Small PR to PULL `exposure_factor` data from PSETs, rather than pushing it. 

Closes #1675

## Overview

Ultra L2: 
1. Add category of variables to PULL project, which must always include "exposure_factor"
2. Remove temporary, imperfect approximation for meaning the exposure. This is no longer necessary with PULL approach

- ultra_l2.py
  - Switch exposure to pull projection and remove old approximation for meaning exposure.

### DIFF
--- a/imap_processing/ultra/l2/ultra_l2.py
+++ b/imap_processing/ultra/l2/ultra_l2.py
@@ -46,7 +46,7 @@ REQUIRED_L1C_VARIABLES_PUSH = [
     "counts",
     "sensitivity",
     "background_rates",
-    "observation_time",
+    "obs_date",
 ]
 REQUIRED_L1C_VARIABLES_PULL = [
     "exposure_factor",
--- a/imap_processing/ultra/l2/ultra_l2.py
+++ b/imap_processing/ultra/l2/ultra_l2.py
@@ -22,9 +22,11 @@ DEFAULT_ULTRA_L2_MAP_STRUCTURE: ena_maps.RectangularSkyMap | ena_maps.HealpixSky
             "spice_reference_frame": "ECLIPJ2000",
             "values_to_push_project": [
                 "counts",
-                "exposure_factor",
                 "sensitivity",
                 "background_rates",
+            ],
+            "values_to_pull_project": [
+                "exposure_factor",
             ],
             "nside": 32,
             "nested": False,
@@ -40,11 +42,14 @@ DEFAULT_L2_HEALPIX_NESTED = False
 
 
 # These variables must always be present in each L1C dataset
-REQUIRED_L1C_VARIABLES = [
+REQUIRED_L1C_VARIABLES_PUSH = [
     "counts",
-    "exposure_factor",
     "sensitivity",
     "background_rates",
+    "observation_time",
+]
+REQUIRED_L1C_VARIABLES_PULL = [
+    "exposure_factor",
 ]
 
 # These variables are projected to the map as the mean of pointing set pixels value,
@@ -153,7 +158,7 @@ def generate_ultra_healpix_skymap(
     2. Iterate over the input pointing sets and read them into UltraPointingSet objects.
     3. For each pointing set, weight certain variables by exposure and solid angle of
     the pointing set pixels.
-    4. Project the pointing set values to the map using the push method.
+    4. Project the pointing set values to the map using the push/pull methods.
     5. Perform subsequent processing for weighted quantities at the SkyMap level
     (e.g., divide weighted quantities by their summed weights to
     get their weighted mean)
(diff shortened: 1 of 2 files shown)


imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -135,15 +140,28 @@ def generate_ultra_healpix_skymap(
     # Get full list of variables to push to the map: all requested variables plus
     # any which are required for L2 processing
     value_keys_to_push_project = list(
-        set(output_map_structure.values_to_push_project + REQUIRED_L1C_VARIABLES)
+        set(output_map_structure.values_to_push_project + REQUIRED_L1C_VARIABLES_PUSH)
```
tmplummer, line 201: Not related to changes in this PR, but I had to go look at the function signature to figure out what `output_map_structure` was. It might be easier to follow this code if it was renamed to `output_skymap`. Just a thought.

Ohhh, upon further inspection, I am confused about inputting `output_map_structure` into this function, especially allowing a `RectangularSkyMap` to be input to a `generate_healpix_skymap` function. Maybe this function is doing too many things? I think that it should either allow the input of THE skymap that gets projected into, or it should allow for a complete skymap configuration dictionary to instantiate the skymap from.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -135,15 +140,28 @@ def generate_ultra_healpix_skymap(
...
     )
+    value_keys_to_pull_project = list(
+        set(output_map_structure.values_to_pull_project + REQUIRED_L1C_VARIABLES_PULL)
+    )
+    # If there are overlapping variable names, issue a warning
+    if set(value_keys_to_push_project).intersection(set(value_keys_to_pull_project)):
```
tmplummer: Unless there is a reason to allow this, I think this should raise a `ValueError`.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -166,26 +184,25 @@ def generate_ultra_healpix_skymap(
...
 
+        # Project values such as counts via the PUSH method
         skymap.project_pset_values_to_map(
             pointing_set=pointing_set,
             value_keys=value_keys_to_push_project,
             index_match_method=ena_maps.IndexMatchMethod.PUSH,
         )
 
+        # Project values such as exposure_factor via the PULL method
+        skymap.project_pset_values_to_map(
```
tmplummer, line 261: It seems somewhat unintuitive to me that one needs to call `skymap.project_pset_values_to_map()` twice in order to get all of the PSET values projected. It seems like the `SkyMap` should be looking at it's lists of keys to push and pull project and doing all of the projection in one function call.

tmplummer (commented):
I got a bit off-track in my review. My overall comments are not really geared toward the purpose of this PR. If you want to move forward with this PR and create a new ticket to address those, that works for me.

imap_processing/ultra/l2/ultra_l2.py
nkerman: > It seems like the SkyMap should be looking at it's lists of keys to push and pull project and doing all of the projection in one function call.

That's how I had originally written it, something like:

```python
skymap.project_pset_values_to_map(
    value_keys_and_methods = {"counts":PUSH, "exposure_factor":PULL, ...}
)
```
At the time, I thought it likely we might project the same values with both PUSH and PULL, renaming them as necessary in the map (e.g. `map["counts_pushed"], map["counts_pulled"]`. 
But in the discussion of #1396, Greg had convinced me it made sense to build multiple maps and combine them in this case. 

Since the maps now have the ability to arbitrarily project whatever they want, as many times as they want, I'm a convert to that way of thinking. To me it's more intuitive to just:

```python
map.project(vars_to_push, PUSH)
map.project(vars_to_pull, PULL)
```

rather than use some sort of dict mapping of variable to push/pull method:

```python
map.project({"counts":PUSH, "exposure_factor":PULL, ...})
```

imap_processing/ultra/l2/ultra_l2.py
nkerman: I don't think it's likely to be used, so I can raise an error. If we need it, we can always switch it back.

imap_processing/ultra/l2/ultra_l2.py
nkerman: I had originally defined a separate small class to store the parameters of an output map, but the feedback I received was to use an empty map to store the configuration parameters. 

The reason we pass in either an empty Rectangular OR Healpix SkyMap here is that we **always** start by making a healpix map - but, if the user asks for a Healpix map to be output, we can directly generate it in this function with whatever parameters the user wants (eg: 128, non-nested). On the other hand, if the user asks for a Rectangular map, we should use this function to make the default parameter Healpix map, then later we can transform/interpolate it to a rectangular map. 

Since all of the map parameters are already fully captured by an empty map, it seems like the perfect way to pass them into this function. The alternatives I can think of would be passing in a dict of parameters, or going back to another class which just stores the map parameters.

imap_processing/ultra/l2/ultra_l2.py
tmplummer: I guess that I was thinking that skymap should use `self.values_to_pull_project` and `self.values_to_push_project` in the `project_pset_values_to_map()` function. I'm not sure I understand what those class attributes are for otherwise. It looks like they are not actually used other than in the `to_dict()` function.

nkerman: Gotcha - yeah I think they're imperfect attributes, but when in #1578, I was asked to just make the `MapProperties` class a part of the `SkyMap`, I still needed the user to be able to pass in any extra quantities to push/pull via `JSON/dict` input. 

At the same time, I feel that calling `map.project()` should **not** always project all values at the same time. Some operations may need to be performed after projecting counts but before projecting "exposure_factor", or similar. So I don't want the `map.project()` to default to pulling or pushing particular values without being explicitly told by the L2 code. That is: I think `map.project()` should **always** be called with an explicit list of `values_to_project` and the method. As such, I tried to heavily document in the code, but I realize that this can be transparent to the user:

`ena_maps.py` line 850 in `AbstractSkyMap`
```python
            - "values_to_push_project" : list[str], optional
                The names of the variables to project to the map with the PUSH method.
                NOTE: The projection is done by the instrument code, so this value can
                only be used to inform that code. No values are projected automatically.
```

One way to neaten this up a bit would be to have the l2 code explicitly project using `output_map_structure.values_to_push_project` directly, rather than the value it currently uses, which looks like: `set(output_map_structure.values_to_push_project, other_defaults)`. It could modify the attribute directly, and still project using this attribute? Let me know what your think here - I might be missing an obvious neater solution.

tmplummer: > Some operations may need to be performed after projecting counts but before projecting "exposure_factor", or similar.

I don't think that this statement agrees with Jon's mapping workflow of `operations on psets() -> project to map() -> operations on maps()`. It sounds like you are anticipating the need for feedback from projected values back to the PSET which he explicitly stated was not needed.

It sounds like you have reasons to have implemented things as they are, so no need to make changes. I'm just trying identify ways to further streamline the map making process.

tmplummer (approved):
Good to go.

imap_processing/ultra/l2/ultra_l2.py
nkerman: I'll admit this is another case of me trying to keep things flexible - we _could_ lock in the projected quantities and make sure the same values are always projected, but I think a bit of flexibility is worth it here. All the same, I'll make the neatening up change I mentioned above, as I totally agree the similar variables are confusing. 

I didn't push back on Jon using the phrase "one shot project" because I think he meant that as in you're projecting all the PSETs at the same time. Which is still what you're essentially doing, even if you project exposure, do some checks or multiplication and then project counts.

PR #1678, opened 2025-05-01
Glows cdf updates

maxinelasp:
# Change Summary
Adding new CDF attributes to GLOWS, some minor rearranging and tidying up as I investigated. 

This renames a few variables but otherwise does not change functionality. 

.yaml files come directly from the GLOWS team, with some minor updates from me. 

This is not completed work, but is an incremental improvement. There is future work around these CDF files ensuring all the variables make sense and match, and that the output types match.


imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -129,9 +129,7 @@ def generate_de_dataset(
     # TODO: Block header per second, or global attribute?
 
     # Store timestamps for each DirectEventL1a object.
-    time_data = np.zeros(len(de_l1a_list), dtype="datetime64[ns]")
-    # TODO: Should each timestamp point to a list of direct events, each with a
-    #  timestamp? Or should the list be split out to make the timestamps?
+    time_data = np.zeros(len(de_l1a_list), dtype=np.int64)
```
tmplummer, line 132: Should this initialize to the FILLVAL instead of 0?

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -293,7 +293,7 @@ def generate_histogram_dataset(
         Dataset containing the GLOWS L1A histogram CDF output.
     """
     # Store timestamps for each HistogramL1A object.
-    time_data = np.zeros(len(hist_l1a_list), dtype="int64")
+    time_data = np.zeros(len(hist_l1a_list), dtype=np.int64)
```
tmplummer, line 296: Same question about using FILLVAL.

tmplummer (commented):
TBH, the variable attributes are too much for me to review in great detail. I had a couple of comments about the initialization of time arrays.

maxinelasp:
@subagonsouth thanks for the review - don't worry about the CDF attributes as I have gone over them closely with GLOWS. 

For the FILLVAL comments, the time_data is just initialized to save time by filling the array instead of appending - it's set to the length of the input list, and then immediately filled with the list. So there is no way any of the indices could be skipped. Should I still update to use FILLVAL instead? I think it's safe enough to leave it, any bugs would result in an error.

imap_processing/cdf/config/imap_glows_l1b_variable_attrs.yaml
```diff
@@ -10,15 +11,18 @@ max_epoch: &max_epoch 3155630469184000000
 # We need to define this if we have DEPEND_1 or more.
 # TODO: I am not sure what the FIELDNAM should be.
```
lacoak21, line 12: Does this change address the TODO?

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
```diff
@@ -58,301 +59,421 @@ bins_dim:
...
   VALIDMAX: 99999
 
 start_time:
   <<: *time_data_defaults
-  CATDESC: Start time
-  FIELDNAM: Start time of the observation day
+  CATDESC: Start time (UTC) of observational day
+  FIELDNAM: Observation start time UTC
+  DISPLAY_TYPE: no_plot
+  LABLAXIS: N/A
```
lacoak21, line 126: Can you remove LABAXIS if the display type is no_plot? 

After searching through previous slack threads and the ISTP metadata guidelines im still unsure so probably worth keeping 😄

lacoak21 (approved):
I only checked the attributes for any obvious errors and had a few questions. Great job with the attention to detail!

imap_processing/cdf/config/imap_glows_l1b_variable_attrs.yaml
maxinelasp: No, that's a GLOWS team TODO

PR #1682, opened 2025-05-02
IDEX Initial L2c processing

lacoak21:
# Change Summary

## Overview
IDEX L2c processing. IDEX wants a pointing set of raw counts. 

Main question is do I even need IDEXPointingSet class. Probably not at this point. 

TODO: 
rectangular psets

## New Files

- imap_processing/cdf/config/imap_idex_l2c_variable_attrs.yaml
   - attrs

-imap_processing/ena_maps/ena_maps.py
   - Idex pointing set
   - rename vars
- imap_processing/idex/idex_l2c.py
  - l2c processing

## Testing
l2c testing.


imap_processing/cdf/config/imap_idex_l2c_variable_attrs.yaml
```diff
@@ -0,0 +1,28 @@
+int_fillval: &int_fillval -9223372036854775808
+
+healpix_index:
```
nkerman: The ENA teams decided to call this "pixel_index" if you want to maintain consistency.

nkerman: imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml

nkerman: Ah, but also this might have only just been changed in dev. It should get updated on a rebase.

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,138 @@
...
+        dims=CoordNames.HEALPIX_INDEX.value,
+    )
+
+    # Create a histogram of the raw dust event counts for each pixel
+    counts = np.histogram(hpix_idx, bins=n_pix, range=(0, n_pix))[0]
+    counds_da = xr.DataArray(
+        counts,
+        name="counts",
+        dims=CoordNames.HEALPIX_INDEX.value,
+    )
```
nkerman: You may be able to use the existing methods in `imap_processing/ena_maps/utils/map_utils.py`, but it's not a ton of overlap, so no pressure.

imap_processing/cdf/config/imap_idex_l2c_variable_attrs.yaml
lacoak21: Great thanks! Will do that.

imap_processing/idex/idex_l2c.py
lacoak21: OK I might just keep it as is for simplicity.

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -386,18 +386,18 @@ ephemeris_velocity_z:
...
 
-right_ascension:
+longitude:
   <<: *spice_base
   VALIDMIN: 0
-  FIELDNAM: Right Ascension
-  CATDESC: Right ascension of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
-  LABLAXIS: Right Ascension
+  FIELDNAM: Longitude
+  CATDESC: Longitude of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
```
greglucas: Are these the spacecraft position, or the dust event location in the sky?

imap_processing/idex/idex_constants.py
```diff
@@ -58,3 +60,15 @@ class ConversionFactors(float, Enum):
...
+# Default IDEX Healpix parameters
+# Used in IDEX l2c processing
+IDEX_HEALPIX_NSIDE = 8
+IDEX_HEALPIX_NESTED = False
+# Default IDEX Rectangular parameters
+# Used in IDEX l2c processing
+IDEX_SPACING_DEG = 10
+
+# Define the pointing reference frame for IDEX
+IDEX_POINTING_REFERENCE_FRAME = SpiceFrame.ECLIPJ2000
```
greglucas: I think we should think about renaming this to avoid the confusion with "pointing set(s)". Maybe `IDEX_EVENT_REFERENCE_FRAME`? i.e. the reference frame the event data is currently in. or `IDEX_MAP_REFERENCE_FRAME` if we are skipping the accumulation in an intermediate projection.

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,236 @@
...
+    Returns
+    -------
+    l2b_dataset : list[xarray.Dataset]
+        The``xarray`` dataset containing the science data and supporting metadata.
+    """
+    logger.info(
+        f"Running IDEX L2C processing on datasets: "
+        f"{l2b_dataset.attrs['Logical_source']}"
+    )
+    # For ISTP, epoch should be the center of the time bin.
```
greglucas: For long collection periods this should actually be the start of the collection period. This is what the other ENA instruments are doing. I think you'll need to update the metadata to indicate it is the left-edge of the reference time, but @nkerman would know better I believe.

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,236 @@
...
+    nested : bool
+        Healpix nested parameter.
+
+    Returns
+    -------
+    pset : xarray.Dataset
+        IDEX pointing set dataset.
+    """
+    longitude = l1b_dataset["longitude"].copy()
+    latitude = l1b_dataset["latitude"].copy()
```
greglucas: Do you need to make a copy here?

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,236 @@
...
+        f"{l2b_dataset.attrs['Logical_source']}"
+    )
+    # For ISTP, epoch should be the center of the time bin.
+    epoch = xr.DataArray(
+        [np.mean(l2b_dataset["epoch"].data[[0, -1]]).astype(np.int64)],
+        name="epoch",
+        dims=["epoch"],
+    )
+    l2c_healpix_dataset = idex_healpix_pset(l2b_dataset, epoch)
+    l2c_rectangular_dataset = idex_rectangular_pset(l2b_dataset, epoch)
```
greglucas: Change these method names to `idex_healpix_map`? i.e. I think you are skipping the intermediate pset creation now and binning directly in the map frame.

greglucas (commented):
Could you rebase this with the latest changes?

I think this all makes sense to me. I'm not sure about our terminology here and whether we want to call these "maps" / "sky grids" etc. or which one this most closely corresponds with a map or a pset. @nkerman what is your take on calling this current iteration a pset vs a map?

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: Oh this is actually supposed to be IDEX position.

lacoak21: Good point. Those CATDESCs need to be updated.

imap_processing/idex/idex_constants.py
lacoak21: Thats a good idea. I like IDEX_EVENT_REFERENCE_FRAME

imap_processing/idex/idex_l2c.py
lacoak21: No - I will remove those.

imap_processing/idex/idex_l2c.py
nkerman: Thanks for bringing this to my attention @greglucas.

In ena_maps.py, the non-spatial coords of the output map (including epoch) default to those of the 0th PSET. If the PSETs are not passed in chronological order, this must be overridden. I had not yet been overriding the epoch metadata at all in the ENA map metadata, as I thought it was always meant to be consistent/constant. 
Looking at `imap_processing/cdf/config/imap_constant_attrs.yaml`: there's nothing to indicate where the epoch falls within the time range. 

We could either override the CATDESC or variable comments, or another way to track this is to use an `epoch_delta_plus` variable linked to epoch in the ISTP compliant way. 

For the ENA maps, the per-pixel obs dates will be more useful anyway, so I admit I haven't given too much thought to this yet.

imap_processing/cdf/config/imap_idex_global_cdf_attrs.yaml
```diff
@@ -55,5 +55,21 @@ imap_idex_l2b_sci:
...
+  <<: *instrument_base
+  Data_level: 2C
+  Data_type: L2C_PSET-HEALPIX>Level-2C Science Data
+  Logical_source: imap_idex_l2c_healpix-pset-1week
+  Logical_source_description: IMAP Mission IDEX Instrument Level-2C Data Healpix
+
+imap_idex_l2c_sci-rectangular:
+  <<: *instrument_base
+  Data_level: 2C
+  Data_type: L2C_PSET-RECTANGULAR>Level-2C Science Data
```
lacoak21: So this should be changed to L2C_MAP-RECTANGULAR>Level-2C Science Data

nkerman:
> I think this all makes sense to me. I'm not sure about our terminology here and whether we want to call these "maps" / "sky grids" etc. or which one this most closely corresponds with a map or a pset. @nkerman what is your take on calling this current iteration a pset vs a map?

The Pointing Set vs Map dichotomy sort of breaks down for data which is not **first** gathered into pixelated sets over a/many pointing/s, and **then** those pointings are combined into a broader Map over the whole sky. Either one is a way of handling pixels on the sky, so the language we tailored for the ENA's doesn't really hold up here. 

Agreed that what you're building here is slightly more analogous to a `SkyMap` than a `PointingSet` because it doesn't have anything to do with a single IMAP pointing - though honestly, I'm not sure why the IDEX team wants to build maps of a ~week's worth of very sparse data. The projected number of counts is on the order of 10, right? That makes me think that they will eventually want to aggregate multiple of these L2C's into a single map, similar to the ENA PointingSets. 

My understanding of IDEX is pretty weak, so my language for discussing it here is falling apart, but given that (I think) one of these L2C files corresponds 1:1 with the weekly collections, and thus with a single L0 packet, you could call this a `"CollectionSet"` - this both draws the parallel to PSET/`"PointingSet"` without the misleading word "Pointing", and also reserves the term Map/`"SkyMap"` for some future product.  

There's no reason this could not eventually be wrapped by a PointingSet class and multiple L2C's could be combined, so I wouldn't worry about that if they don't think they want it yet.

imap_processing/cdf/config/imap_idex_l2c_variable_attrs.yaml
```diff
@@ -0,0 +1,82 @@
...
+  LABLAXIS: Latitude Pixel
+  VAR_TYPE: support_data
+  DISPLAY_TYPE: no_plot
+  UNITS: "degrees"
+  VALIDMIN: 0
+  VALIDMAX: 180
+  FILLVAL: *int_fillval
+  DICT_KEY: SPASE>Support>SupportQuantity:Positional
+
+healpix_counts:
```
nkerman: Consider making these variables' metadata not specific to the tiling, and instead having them both inherit from "counts" which has the CATDESC, VALIDMIN, etc..., and override with the LABL_PTR and DEPEND's.

imap_processing/cdf/config/imap_idex_l2c_variable_attrs.yaml
```diff
@@ -0,0 +1,82 @@
...
+  CATDESC: Number of raw dust events for each healpix pixel
+  FIELDNAM: Dust Counts
+  FORMAT: I4
+  LABLAXIS: Counts
+  VAR_TYPE: data
+  DISPLAY_TYPE: time_series
+  DEPEND_0: pixel_index
```
nkerman: I think DEPEND_0 must be epoch, right? This should get pushed to DEPEND_1

imap_processing/cdf/config/imap_idex_l2c_variable_attrs.yaml
```diff
@@ -0,0 +1,82 @@
...
+  FILLVAL: *int_fillval
+  DICT_KEY: SPASE>Particle>ParticleType:Dust,ParticleQuantity:Counts,Qualifier:Array
+
+rectangular_counts:
+  CATDESC: Number of raw dust events for each rectangular pixel
+  FIELDNAM: Dust Counts
+  FORMAT: I4
+  VAR_TYPE: data
+  DISPLAY_TYPE: time_series
+  DEPEND_0: rectangular_lon_pixel
```
nkerman: ditto above WRT epoch.

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,236 @@
+"""
+Perform IDEX L2c Processing.
+
+Examples
+--------
+.. code-block:: python
+    from imap_processing.idex.idex_l1a import PacketParser
+    from imap_processing.idex.idex_l1b import idex_l1b
+    from imap_processing.idex.idex_l1b import idex_l2a
+    from imap_processing.idex.idex_l1b import idex_l2b
```
nkerman:
```suggestion
    from imap_processing.idex.idex_l1b import idex_l2b
    from imap_processing.cdf.utils import write_cdf
```

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,236 @@
...
+    ].attrs = idex_attrs.get_variable_attributes("rectangular_lat_pixel")
+    l2c_rectangular_dataset[
+        "rectangular_lon_pixel"
+    ].attrs = idex_attrs.get_variable_attributes("rectangular_lon_pixel")
+    l2c_rectangular_dataset[
+        "rectangular_lat_pixel_label"
+    ].attrs = idex_attrs.get_variable_attributes("rectangular_lat_pixel_label")
+    l2c_rectangular_dataset[
+        "rectangular_lon_pixel_label"
+    ].attrs = idex_attrs.get_variable_attributes("rectangular_lon_pixel_label")
```
nkerman: For all these variables, do we want the output file to have the names be "rectangular_counts" or just "counts"? Given the Healpix and Rectangular data are held in separate Datasets, for the ENA files we've gone with just "counts"

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,236 @@
...
+        counts,
+        name="rectangular_counts",
+        dims=("rectangular_lon_pixel", "rectangular_lat_pixel"),
+    )
+    rec_lon_pixels = xr.DataArray(
+        np.arange(grid.grid_shape[0]), dims="rectangular_lon_pixel"
+    )
+    rec_lat_pixels = xr.DataArray(
+        np.arange(grid.grid_shape[1]), dims="rectangular_lat_pixel"
+    )
```
nkerman: I think these should be `grid.az_bin_midpoints, grid.el_bin_midpoints` - what you have now, for a spacing of 4 deg would give `rec_lon_pixels = [0,1,2,...90]` rather than `[2, 6, 10, ..., 358]`. I think you want the latter.

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,236 @@
...
+            "latitude": latitude,
+            "rectangular_lon_pixel_label": rec_lon_pixels.astype(str),
+            "rectangular_lat_pixel_label": rec_lat_pixels.astype(str),
+        },
+    )
+    pset_attrs = {
+        "sky_tiling_type": SkyTilingType.RECTANGULAR.value,
+        "spacing_degree": spacing_deg,
+        "spice_reference_frame": IDEX_EVENT_REFERENCE_FRAME,
+        "num_points": counts.size,
```
nkerman:
```suggestion
        "Sky_tiling_type": SkyTilingType.RECTANGULAR.value,
        "Spacing_degrees": spacing_deg,
        "Spice_reference_frame": IDEX_EVENT_REFERENCE_FRAME,
        "num_points": counts.size,
```

nkerman: tiny suggestion for consistency.

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,236 @@
...
+            "latitude": latitude,
+        },
+    )
+    pset_attrs = {
+        "sky_tiling_type": SkyTilingType.HEALPIX.value,
+        "HEALPix_nside": nside,
+        "HEALPix_nest": nested,
+        "spice_reference_frame": IDEX_EVENT_REFERENCE_FRAME,
+        "num_points": n_pix,
+    }
```
nkerman:
```suggestion
    pset_attrs = {
        "Sky_tiling_type": SkyTilingType.HEALPIX.value,
        "HEALPix_nside": nside,
        "HEALPix_nest": nested,
        "Spice_reference_frame": IDEX_EVENT_REFERENCE_FRAME,
        "num_points": n_pix,
    }
```

nkerman: For consistency.

nkerman (commented):
Lots of great stuff here! I'm still a bit confused about the product that IDEX is looking for, so mostly focusing on a few technical implementation bits here.

imap_processing/idex/idex_l2c.py
lacoak21: OK I will follow ENA and have then just be named counts.

imap_processing/idex/idex_l2c.py
lacoak21: yes! Thank you good catch.

lacoak21:
> Lots of great stuff here! I'm still a bit confused about the product that IDEX is looking for, so mostly focusing on a few technical implementation bits here.

Thanks for your review Nat!

lacoak21:
> > I think this all makes sense to me. I'm not sure about our terminology here and whether we want to call these "maps" / "sky grids" etc. or which one this most closely corresponds with a map or a pset. @nkerman what is your take on calling this current iteration a pset vs a map?
> 
> The Pointing Set vs Map dichotomy sort of breaks down for data which is not **first** gathered into pixelated sets over a/many pointing/s, and **then** those pointings are combined into a broader Map over the whole sky. Either one is a way of handling pixels on the sky, so the language we tailored for the ENA's doesn't really hold up here.
> 
> Agreed that what you're building here is slightly more analogous to a `SkyMap` than a `PointingSet` because it doesn't have anything to do with a single IMAP pointing - though honestly, I'm not sure why the IDEX team wants to build maps of a ~week's worth of very sparse data. The projected number of counts is on the order of 10, right? That makes me think that they will eventually want to aggregate multiple of these L2C's into a single map, similar to the ENA PointingSets.
> 
> My understanding of IDEX is pretty weak, so my language for discussing it here is falling apart, but given that (I think) one of these L2C files corresponds 1:1 with the weekly collections, and thus with a single L0 packet, you could call this a `"CollectionSet"` - this both draws the parallel to PSET/`"PointingSet"` without the misleading word "Pointing", and also reserves the term Map/`"SkyMap"` for some future product.
> 
> There's no reason this could not eventually be wrapped by a PointingSet class and multiple L2C's could be combined, so I wouldn't worry about that if they don't think they want it yet.

I like the term collectionSet although it seems like Matina is happy with this product as a "map". Does that sounds reasonable to continue with "map" labels in this case?

lacoak21:
pre-commit.ci autofix

lacoak21:
pre-commit.ci autofix

imap_processing/idex/idex_l2c.py
lacoak21: OK I overrode the metadata of the epoch. Right now the epoch is the first value but technically it is not the beginning of the collection period. That would be the science acquisition start time which comes from the idex l1b housekeeping file. I am going to put up a PR for that soon and then can update that here.

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
```diff
@@ -385,18 +385,18 @@ ephemeris_velocity_z:
...
 
-right_ascension:
+longitude:
   <<: *spice_base
   VALIDMIN: 0
-  FIELDNAM: Right Ascension
-  CATDESC: Right ascension of the IMAP spacecraft position as observed from Earth in the ECLIPJ2000 reference frame.
-  LABLAXIS: Right Ascension
+  FIELDNAM: Longitude
+  CATDESC: Longitude (spherical coordinate) of the dust event (IDEX pointing) as observed from Earth in the ECLIPJ2000 reference frame.
```
greglucas: Is this under 80 characters? I would remove parentheses if necessary.
https://spdf.gsfc.nasa.gov/istp_guide/vattributes.html#CATDESC

imap_processing/cdf/config/imap_idex_l2c_variable_attrs.yaml
```diff
@@ -0,0 +1,83 @@
...
+  FILLVAL: *int_fillval
+  DICT_KEY: SPASE>Support>SupportQuantity:Positional
+
+rectangular_lon_pixel:
+  CATDESC: Longitude pixel index for IDEX SkyMap
+  FIELDNAM: Rectangular Longitude Index
+  FORMAT: F12.6
+  LABLAXIS: Longitude Pixel
+  VAR_TYPE: support_data
+  UNITS: "degrees"
```
greglucas: I've seen this in other peoples PRs too. Do you know why only the UNITS variables are in quotes?

imap_processing/idex/idex_l1b.py
```diff
@@ -350,15 +353,15 @@ def get_spice_data(
...
+    # Get Idex pointing in the defined frame
     idex_pointing = instrument_pointing(
-        et, SpiceFrame.IMAP_IDEX, SpiceFrame.J2000, cartesian=True
+        et, SpiceFrame.IMAP_IDEX, IDEX_EVENT_REFERENCE_FRAME, cartesian=True
     )
     solar_lon = solar_longitude(et, degrees=True)
-
-    range_ra_and_dec = cartesian_to_spherical(idex_pointing)
+    # longitude and latitude
+    lon_and_lat = cartesian_to_spherical(idex_pointing)
```
greglucas: This is a bit misleading because it looks like it also has "r" as the first index. Maybe `spherical_coords = ...` or `func()[:, 1:]` to drop the "r" component.

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,255 @@
...
+    from imap_processing.idex.idex_l1a import PacketParser
+    from imap_processing.idex.idex_l1b import idex_l1b
+    from imap_processing.idex.idex_l2a import idex_l2a
+    from imap_processing.idex.idex_l2b import idex_l2b
+    from imap_processing.cdf.utils import write_cdf
+
+    l0_file = "imap_processing/tests/idex/imap_idex_l0_sci_20231214_v001.pkts"
+    l1a_data = PacketParser(l0_file)
+    l1b_data = idex_l1b(l1a_data)
+    l1a_data = idex_l2a(l1b_data)
```
greglucas:
```suggestion
    l2a_data = idex_l2a(l1b_data)
```

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,255 @@
...
+    idex_attrs = get_idex_attrs("l2c")
+    # Epoch should be the start of the collection period.
+    # TODO should epoch be start of sci acquisition?
+    epoch = xr.DataArray(
+        l2b_dataset["epoch"].data[0:1].astype(np.int64),
+        name="epoch",
+        dims=["epoch"],
+        attrs=idex_attrs.get_variable_attributes("epoch", check_schema=False),
+    )
+    # Update metadata to indicate that epoch is left-edge of the reference time.
```
greglucas: Can you define this separate "epoch" variable in the l2c as "epoch_pointing" or something similar and say `idex_attrs.get_variable_attributes("epoch_pointing")` above so you are putting this text into the configuration files instead of code.

Also, this looks like it might be longer than 80 characters.

imap_processing/idex/idex_l2c.py
```diff
@@ -0,0 +1,255 @@
...
+    # Make sure longitude values are in the range [0, 360)
+    longitude_wrapped = np.mod(l1b_dataset["longitude"], 360)
+    latitude = l1b_dataset["latitude"]
+    # Create a 2d histogram of the raw dust event counts for each pixel using the grid
+    # bin edges
+    counts, _, _ = np.histogram2d(
+        longitude_wrapped, latitude, bins=[grid.az_bin_edges, grid.el_bin_edges]
+    )
+    counts_da = xr.DataArray(
+        counts[np.newaxis, :, :].astype(int),
```
greglucas: Counts should generally be uint as they can't be negative and you can probably reduce the space by making it a smaller bit range too. Maybe even uint8?
```suggestion
        counts[np.newaxis, :, :].astype(np.uint16),
```

greglucas (approved):
Looks good to me, a few minor nits but the overall structure looks great!

imap_processing/cdf/config/imap_idex_l1b_variable_attrs.yaml
lacoak21: Oh good catch. I didnt realize the limit was 80 characters.

imap_processing/cdf/config/imap_idex_l2c_variable_attrs.yaml
lacoak21: I am not sure. I saw others with this and copied it.

lacoak21: I will remove them.

imap_processing/idex/idex_l2c.py
lacoak21: Yes good idea.

PR #1692, opened 2025-05-05
Optionally energy-dependent `exposure_factor`

### LINKED ISSUE #1691: Ultra L2: Handle both energy-dependent/independent "exposure_factor"
In the case of the heliosphere frame PSETs, the `"exposure_factor"` variable will have `dims=("epoch", "energy_...","pixel_index")` while the spacecraft frame is not energy dependent, and will have `dims=("epoch","pixel_index")`. This needs to be handled properly in Ultra L2, especially by handling the metadata differently between the two cases. 

Note that "obs_date" will be weighted-meaned by exposure time, and thus must be handled similarly.

nkerman:
# Change Summary

Adapt Ultra L2 metadata and observation date calculation to handle exposure_factor with and without an energy dimension. 

Closes #1691

## Overview

In the case of the heliosphere frame PSETs, the "exposure_factor" variable will have dims=("epoch", **"energy_..."**,"pixel_index") while the spacecraft frame is not energy dependent, and will have dims=("epoch","pixel_index"). This needs to be handled properly in Ultra L2, especially by handling the metadata differently between the two cases.

## Updated Files

-  imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
-  imap_processing/cdf/config/imap_enamaps_l2-rectangular_variable_attrs.yaml
   - Add overwrites for `exposure_factor_energy_dependent` and `obs_date_energy_dependent`'s LABL_PTRs and DEPENDs
   - Add wrapper function to get metadata on vars with optional energy dependence. 
     - These variables must take in the energy independent metadata, and overwrite their DEPENDs and LABL_PTRs
-  imap_processing/tests/ultra/mock_data.py
   - Mock the exposure_factor either with/out energy dependence. 

## Testing

Test new function.

### DIFF
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -103,16 +103,6 @@ latitude:
 # Define Data variables, but not their pixel-level DEPENDs
 # (e.g. pixel_index for HEALPix maps, longitude/latitude for rectangular maps).
 # Those are defined in the tiling specific YAML files.
-exposure_factor:
-  <<: *default_float32
-  CATDESC: Exact or approximate exposure time over which counts in a pixel are accumulated.  Used as a weighting factor for combining data quantities sensibly.
-  FIELDNAM: Exposure Times
-  DEPEND_0: epoch
-  UNITS: s
-  VAR_TYPE: data
-  LABLAXIS: Exposure
-  DISPLAY_TYPE: no_plot
-
 ena_intensity:
   <<: *default_float32
   CATDESC: Mono-energetic ENA intensity.
@@ -145,7 +135,17 @@ sensitivity:
   LABLAXIS: sensitivity
   DISPLAY_TYPE: image
 
-obs_date:
+exposure_factor: &exposure_factor_energy_independent
+  <<: *default_float32
+  CATDESC: Exact or approximate exposure time over which counts in a pixel are accumulated.  Used as a weighting factor for combining data quantities sensibly.
+  FIELDNAM: Exposure Times
+  DEPEND_0: epoch
+  UNITS: s
+  VAR_TYPE: data
+  LABLAXIS: Exposure
+  DISPLAY_TYPE: no_plot
+
+obs_date: &obs_date_energy_independent
   <<: *default_int64
   datatype: int64
   CATDESC: Exposure time weighted mean collection date of data in a pixel.
@@ -155,3 +155,12 @@ obs_date:
   VAR_TYPE: data
   LABLAXIS: epoch
   DISPLAY_TYPE: image
+
+# These copied metadata vars will allow for variables
+# to be either energy-dependent or independent.
+# The tiling-specific YAML files will override the DEPENDs and LABL_PTRs.
+exposure_factor_energy_dependent:
+  <<: *exposure_factor_energy_independent
+
+obs_date_energy_dependent:
+  <<: *obs_date_energy_independent
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -137,7 +137,7 @@ sensitivity:
 
 exposure_factor: &exposure_factor_energy_independent
   <<: *default_float32
-  CATDESC: Exact or approximate exposure time over which counts in a pixel are accumulated.  Used as a weighting factor for combining data quantities sensibly.
+  CATDESC: Exact or approximate time over which counts are accumulated.
   FIELDNAM: Exposure Times
   DEPEND_0: epoch
   UNITS: s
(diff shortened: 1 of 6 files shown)


imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -66,6 +66,62 @@
...
+        variable_name=variable_name,
+        check_schema=check_schema,
+    )
+    if (variable_name in INCONSISTENTLY_ENERGY_DEPENDENT_VARIABLES) and (
+        "energy" in variable_dims
+    ):
+        # These are overrides which only affect the latter DEPEND_N and LABL_PTR_N
+        # They do not cover all the metadata values so we will not
+        # use the schema validation.
+        metadata.update(
```
maxinelasp: I think if your original `metadata` variable includes _extra_ DEPEND_N variables this might cause weird behavior. 

 if for example, your initial metadata variable has DEPEND_0, DEPEND_1, DEPEND_2 and your override variable has DEPEND_0, and DEPEND_1, the extra DEPEND_2 will still be hanging around. 

Why not have an `exposure_factor` and an `exposure_factor_energy_default` and then select between them here, rather than overwriting? (so with an if/else to select between exposure_factor and exposure_factor_energy).

nkerman: >if for example, your initial metadata variable has DEPEND_0, DEPEND_1, DEPEND_2 and your override variable has DEPEND_0, and DEPEND_1, the extra DEPEND_2 will still be hanging around.

Yeah - I've done it in the reverse here, which I don't think has that issue. 

>Why not have an exposure_factor and an exposure_factor_energy_default and then select between them here, rather than overwriting? (so with an if/else to select between exposure_factor and exposure_factor_energy). 

With the ENA maps, many of the CDF variables (both `data_vars` and `coords`) are broadly the same between different maps, and so we can define most of their metadata in `imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml`. But, then a few metadata values need to change depending on the tiling (Rectangular vs. Healpix) so there are overwriting metadata values in `imap_processing/cdf/config/imap_enamaps_l2-<rectangular/healpix>_variable_attrs.yaml`. 

Regardless of whether the energy dimension is present, "exposure_factor" needs the same, common metadata values in the `common` yaml file under the variable name "exposure_factor". These are things like CATDESC and FORMAT. That means, I'll still need something like the logic in this function to query first the common name ("exposure_factor") and then the energy-related one ("exposure_factor_energy_dependent"). I could duplicate the common metadata under the energy dependent name in the common yaml, but that seems a bit messy. I suppose that I could use the yaml inheritance to keep it cleaner though.

maxinelasp: I see, in that case I would rely on the SAMMI yaml merging capability and have one attribute manager which has common + rectangular and one attribute manager which has common + healpix. SAMMI will merge duplicate variables together in a more robust but very similar way (check out the[ _merge() function](https://github.com/swxsoc/sammi/blob/main/sammi/cdf_attribute_manager.py#L239))

So you would have: 
healpix_attribute_manager.load_variable_attributes(common_file.yaml)
healpix_attribute_manager.load_variable_attributes(healpix_file.yaml)
rect_attribute_manager.load_variable_attributes(common_file.yaml)
rect_attribute_manager.load_variable_attributes(healpix_file.yaml)

(you would just have one "exposure_factor" in each file and retrieve it as "exposure_factor".)

and then here you can pick between the two attribute managers and just get your variable attributes from there. I think that's more reliable (as SAMMI handles more complex merges) and more obvious what is happening. Make sense? It might be a little overkill for what you're doing here, but if you anticipate more variable changes (which is what I would expect) it will be a lot easier to handle.

nkerman: Maxine and I chatted about this offline and found a good solution!

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -145,7 +135,17 @@ sensitivity:
   LABLAXIS: sensitivity
   DISPLAY_TYPE: image
 
-obs_date:
+exposure_factor: &exposure_factor_energy_independent
+  <<: *default_float32
+  CATDESC: Exact or approximate exposure time over which counts in a pixel are accumulated.  Used as a weighting factor for combining data quantities sensibly.
```
maxinelasp: Just for the record, CATDESC is meant to top out at 60 chars, so this one is too long.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -66,6 +66,56 @@
...
+    variable_name = variable_array.name
+    variable_dims = variable_array.dims
+
+    # These variables must get metadata with a different key if they are energy
+    # dependent.
+    if (variable_name in INCONSISTENTLY_ENERGY_DEPENDENT_VARIABLES) and (
+        (CoordNames.ENERGY_L2.value in variable_dims)
+        or (CoordNames.ENERGY_ULTRA_L1C.value in variable_dims)
+    ):
+        variable_name = f"{variable_name}_energy_dependent"
```
maxinelasp, line 111: Nice!

maxinelasp (approved):
Looks good!

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
nkerman: Oops, yeah an old version of this CATDESC snuck in - I put a shorter version in!

PR #1693, opened 2025-05-06
IDEX L1a Error handling for missing event data

### LINKED ISSUE #1688: Incomplete event error handling
Currently the IDEX l1a code completely errors out if there is a dropped packet and the full data for an event is missing. No CDFs are produced in this case. 

The IDEX team has requested that a warning be logged if this is the case but still produce a l1a CDF with events that are complete.

lacoak21:
# Change Summary

closes #1688

## Overview
The IDEX team requested that a warning be logged for incomplete events (dropped packets) in the data, while still allowing the CDF to be created with he remainder of the complete events.

## Updated Files
- imap_processing/cdf/config/imap_idex_l1a_variable_attrs.yaml
   - small metadata fix.
   - Catch missing data error and log a warning.

### DIFF
--- a/imap_processing/idex/idex_l1a.py
+++ b/imap_processing/idex/idex_l1a.py
@@ -668,7 +668,7 @@ class RawDustEvent:
                 attrs=idex_attrs.get_variable_attributes("ion_grid_attrs"),
             ),
         }
-        # Determine the 3 coordinate variables
+        # Determine coordinate variables
         coords = {
             "epoch": xr.DataArray(
                 name="epoch",
@@ -676,6 +676,8 @@ class RawDustEvent:
                 dims=("epoch"),
                 attrs=idex_attrs.get_variable_attributes("epoch"),
             ),
+        }
+        sampling_rates = {
             "time_low_sample_rate": xr.DataArray(
                 name="time_low_sample_rate",
                 data=[
@@ -692,8 +694,7 @@ class RawDustEvent:
             ),
         }
         expected_shapes = {
-            "time_low_sample_rate_index": coords["time_low_sample_rate"].shape[1],
-            "time_high_sample_rate_index": coords["time_high_sample_rate"].shape[1],
+            f"{name}_index": array.shape[1] for name, array in sampling_rates.items()
         }
         if any(
             var.shape[1] != expected_shapes[var.dims[1]] for var in data_vars.values()
@@ -709,7 +710,7 @@ class RawDustEvent:
 
         # Combine to return a dataset object
         dataset = xr.Dataset(
-            data_vars=data_vars | trigger_vars,
+            data_vars=data_vars | trigger_vars | sampling_rates,
             coords=coords,
         )
         return dataset
--- a/imap_processing/idex/idex_l1a.py
+++ b/imap_processing/idex/idex_l1a.py
@@ -152,7 +152,7 @@ class PacketParser:
         processed_dust_impact_list = [
             dust_event.process() for dust_event in dust_events.values()
         ]
-
+        processed_dust_impact_list = [x for x in processed_dust_impact_list if x]
         data = xr.concat(processed_dust_impact_list, dim="epoch")
         data.attrs = self.idex_attrs.get_global_attributes("imap_idex_l1a_sci")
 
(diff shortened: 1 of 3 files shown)


imap_processing/idex/idex_l1a.py
```diff
@@ -152,7 +152,7 @@ def _create_science_dataset(self, science_decom_packet_list: list) -> xr.Dataset
         processed_dust_impact_list = [
             dust_event.process() for dust_event in dust_events.values()
         ]
-
+        processed_dust_impact_list = [x for x in processed_dust_impact_list if x]
```
greglucas:
```suggestion
        processed_dust_impact_list = [x for x in processed_dust_impact_list if x is not None]
```

imap_processing/idex/idex_l1a.py
```diff
@@ -599,7 +600,7 @@ def _populate_bit_strings(
         raw_science_bits = convert_to_binary_string(packet["IDX__SCI0RAW"])
         self._append_raw_data(scitype, raw_science_bits)
 
-    def process(self) -> xr.Dataset:
+    def process(self) -> Union[xr.Dataset, None]:
```
greglucas:
```suggestion
    def process(self) -> xr.Dataset | None:
```

imap_processing/idex/idex_l1a.py
```diff
@@ -684,23 +685,31 @@ def process(self) -> xr.Dataset:
...
-                "Target_Low": target_low,
-                "Ion_Grid": ion_grid,
-                "time_low_sample_rate": time_low_sample_rate,
-                "time_high_sample_rate": time_high_sample_rate,
-            }
-            | trigger_vars,
-            coords={"epoch": epoch},
-        )
-
+        try:
```
greglucas: Is there a way you can check for this outside of a ValueError exception? What is the ValueError arising from here...

imap_processing/idex/idex_l1a.py
```diff
@@ -684,23 +685,31 @@ def process(self) -> xr.Dataset:
...
+            )
+        except ValueError as e:
+            # The IDEX team requests that a warning be logged for incomplete events
+            # (dropped packets) in the data, while still allowing the CDF to be created
+            # with the remainder of the complete events.
+            logger.warning(
+                f"Missing packet for event number %s. Error: {e}Skipping event..",
+                self.event_number,
+            )
+            return None
```
greglucas, line 709:
```suggestion
            return
```

imap_processing/idex/idex_l1a.py
```diff
@@ -684,23 +685,31 @@ def process(self) -> xr.Dataset:
...
+                coords={"epoch": epoch},
+            )
+        except ValueError as e:
+            # The IDEX team requests that a warning be logged for incomplete events
+            # (dropped packets) in the data, while still allowing the CDF to be created
+            # with the remainder of the complete events.
+            logger.warning(
+                f"Missing packet for event number %s. Error: {e}Skipping event..",
+                self.event_number,
+            )
```
greglucas: 🤔 How will they know this case happened without looking at the logs for every single job that comes in? I'm wondering if it would make sense to put in FILL_VALs for the missing items so you know there was a missing event somewhere. Or another thing this gets me to thinking about is that perhaps we also need some query to check for log warnings/errors to catch notices like this during processing. None of this is for here, just writing down my thoughts.

imap_processing/tests/idex/test_idex_l1a.py
```diff
@@ -84,6 +85,30 @@ def test_bad_cdf_file_data(decom_test_data_sci: xr.Dataset):
     del decom_test_data_sci["Bad_data"]
 
 
+def test_incomplete_event(caplog):
```
greglucas, line 88: Great test, thank you for adding this!

greglucas (approved):
Very minor style suggestions. Looks great though!

imap_processing/idex/idex_l1a.py
lacoak21: Something like the latter would be ideal. I think it would be good to add the fill_vals for the missing event though.

imap_processing/idex/idex_l1a.py
lacoak21: Yes I can check all of the dimensions beforehand.

imap_processing/idex/idex_l1a.py
lacoak21: I am getting a mypy return-value expected error here without returning None

greglucas: Annoying :( An empty return does return None... Sorry for the noise!

PR #1694, opened 2025-05-06
I-ALiRT - Calculate current s/c attitude from raw spin phase and angular momentum

### LINKED ISSUE #1373: Calculate current s/c attitude from raw spin phase and angular momentum
### Summary

On I-ALiRT, we are getting packets back ahead of any historical/true attitude from SPICE. So we have to use either the predicted attitude kernels from SPICE or calculate it ourselves from the raw onboard data that comes down with the I-ALiRT packet. The onboard data should be more accurate in case the spacecraft drifts out of any nominal SPICE predictions, so we should use that if possible.

The onboard data that we have access to are:
* spin period estimate
* spin phase angle estimate (@ next 1 pps) NOTE: This appears like we might want to look at the historical previous packet to get the angle
* inertial RA of system angular momentum vector (0 -> 2pi, 16 bit unsigned int LSB of 9.5875e-5 rad/count)
* inertial dec of system angular momentum vector (-pi/2 -> pi/2, 16 bit unsigned int LSB of 4.7938e-5 rad/count)

Validity flags: 
* spin period validity (0: invalid, 1: valid)
* spin phase validity (0: invalid, 1: valid)
* attitude estimation source (0: star tracker, 1: sun sensor)
* catbed heater flag (0: off, 1: on)
* autonomy application validity flag (0: not operational, 1: operational)

The spin_phase is the angle rotating about the direction of the angular momentum vector. So the algorithm is something like: fix the RA/dec angular momentum vector in our reference coordinate system, then spin the spacecraft around with the proper spin phase angle after that. Use the geometry angle offset routines we already have for each instrument after that to calculate look direction of individual instruments.

### Requirements

  - [ ] Write up the math for calculating the look direction of each instrument from the raw packet data
  - [ ] Implement a lookup function similar to the geometry functions we already have for calculating instrument look direction
  - [ ] Verify that the predicted attitude kernels and this manual calculation match expectations
  - [ ] Extra credit:
    - [ ] Add checks for when to use raw data or not based on validity flags. Need to check with guidance and nav team about this. Maybe we just don't produce any I-ALiRT data when any of these flags are out of nominal...
    - [ ] Add additional warnings/checks when predicts don't match actual onboard calculations or things drift out of some threshold.

cc @laspsandoval

laspsandoval:
# Change Summary

## Overview
This is the beginning of a utility module that reconstructs the spacecraft’s attitude (orientation in space) using onboard telemetry instead of SPICE attitude kernels. Specifically, we use right ascension, declination, and spin phase values from I-ALiRT packets to build a transformation from instrument reference frames into the spacecraft’s spinning frame (URF).

Left to do:
- Verify that the predicted attitude kernels and this manual calculation match expectations
- Specific to MAG: Despin Vectors, transform and output the single vector to other reference frames

## New Files
- ialirt_spice.py
   - reconstructs the spacecraft’s attitude (orientation in space) using onboard telemetry instead of SPICE attitude kernels

## Testing
- test_ialirt_spice.py


imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,145 @@
...
+
+    sc_inertial_decline : np.ndarray
+        Declination in radians.
+
+    Returns
+    -------
+    z_axis : np.ndarray
+        Unit vectors of the spacecraft Z-axis (N, 3).
+    """
+    # Convert right ascension from counts to radians (0-2pi).
```
tmplummer:
```suggestion
    # Convert right ascension from radians to degrees.
```

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,145 @@
...
+    ra_deg = np.degrees(sc_inertial_right)
+    # Convert declination from counts to radians (-pi/2 to pi/2).
```
tmplummer:
```suggestion
    # Convert declination from radians degrees.
```

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def get_z_axis(sc_inertial_right: NDArray, sc_inertial_decline: NDArray) -> NDArray:
+    """
+    Compute the spacecraft Z-axis (angular momentum direction) in inertial coordinates.
+
+    Parameters
+    ----------
+    sc_inertial_right : np.ndarray
+        Right ascension in radians.
```
tmplummer:
```suggestion
        Right ascension of the spacecraft spin-axis in radians.
```

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,145 @@
...
+
+    sc_inertial_decline : np.ndarray
+        Declination in radians.
```
tmplummer:
```suggestion
        Declination of the spacecraft spin-axis in radians.
```

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def transform_instrument_vectors_to_urf(
+    instrument_vectors: NDArray,
+    spin_phase: NDArray,
+    sc_inertial_right: NDArray,
+    sc_inertial_decline: NDArray,
+) -> NDArray:
+    """
+    Transform instrument-frame vectors into the spacecraft URF frame.
```
tmplummer, line 80: URF frame is new to me. I don't see it in the Technical Memo about Standard Coordinate Frames. What is it?

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Returns
+    -------
+    x_axis : NDArray
+        Array of shape (N, 3) perpendicular to z_axis.
+    y_axis : NDArray
+        Array of shape (N, 3) perpendicular to z_axis.
+    """
+    # Pick a fixed reference vector.
+    v_ref = np.array([0, 1, 0])
```
tmplummer: I'm surprised that the x and y axes get defined somewhat arbitrarily by taking the cross of either the x or y axis (presumably in the equatorial frame?). I guess that since it is only being used to rotate around the z axis, the only thing that matters is that the x and y are orthogonal and right handed?

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
tmplummer, line 1: Cool, the tests look good.

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,145 @@
...
+    x_temp = np.cross(v_refs, z_axis)
+    x_axis = x_temp / np.linalg.norm(x_temp, axis=-1, keepdims=True)
+
+    # Take the cross product to get the Y-axis.
+    y_axis = np.cross(z_axis, x_axis)
+
+    return x_axis, y_axis
+
+
+def rotate_frame_about_spin_axis(z_axis: NDArray, spin_phase: NDArray) -> NDArray:
```
tmplummer: This function name and description are not accurate. Can you change it to `get_rotation_matrix` or similar?

tmplummer (commented):
Overall, looks great. Just a couple of nits.

imap_processing/ialirt/l0/ialirt_spice.py
laspsandoval: Yeah, I'm not quite sure if I will even use this function. I wrote it and then realized maybe it isn't necessary. I'll delete it, I think.

imap_processing/ialirt/l0/ialirt_spice.py
laspsandoval: URF = Unrotated Reference Frame
Terminology used by the MAG team.
it's a spacecraft-fixed frame that rotates with the spacecraft,

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,108 @@
...
+    This matrix acts just like SPICE's pxform(instrument_frame, "IMAP_SPACECRAFT", et).
+    A forward rotation that transforms vectors from the instrument's local frame
+    to the spacecraft’s rotating frame (URF)
+    """
+    # Rotation matrix to rotate about z_axis by spin_phase
+    rot_matrices = []
+    for z, phase in zip(z_axis, spin_phase):
+        rot = spice.axisar(z, float(phase))
+        rot_matrices.append(rot)
+    rot_matrices = np.array(rot_matrices)
```
tmplummer: I don't know how long of input vectors you will have, but you can avoid `append`ing to a list by using list comprehension which is faster. 
```suggestion
    rot_matrices = np.array(
        [
            spice.axisar(z, float(phase))
            for z, phase in zip(z_axis, spin_phase)
        ]
    )
```

imap_processing/ialirt/l0/ialirt_spice.py
tmplummer: This might be useful info to add to the docstring.

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -0,0 +1,108 @@
...
+    vectors_urf : np.ndarray
+        Vectors in the spacecraft URF frame. Shape: (N, 3).
+    """
+    z_axis = get_z_axis(sc_inertial_right, sc_inertial_decline)
+    rot_matrices = get_rotation_matrix(z_axis, spin_phase)
+
+    vectors_urf = []
+    for r, v in zip(rot_matrices, instrument_vectors):
+        vectors_urf.append(spice.mxv(r, v))
+    vectors_urf = np.array(vectors_urf)
```
tmplummer: Same comment about list comprehension...
```suggestion
    vectors_urf = np.array(
        [
            spice.mxv(r, v)
            for r, v in zip(rot_matrices, instrument_vectors)
        ]
    )
```

tmplummer (approved):
A couple more suggestions, but otherwise looks great.

PR #1701, opened 2025-05-06
Add Ultra L2 to the CLI

### LINKED ISSUE #1700: Add Ultra L2 to the CLI
Ultra L2 is ready to be added to the CLI code.

nkerman:
# Change Summary

Add Ultra's L2 processing to the CLI.

Closes #1700 

## Updated Files

- `imap_processing/cli.py`
   - Add ultra L2 processing to the CLI code
- Add some `__init__.py` files to subdirectories. 

## Testing

- Add test of Ultra l2 CLI with everything being mocked
- I'm looking for some more robust and 'real' ways to test this, but it has the sort of tests I've found in `test_cli.py`
  - I'm very open to suggestions!

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -66,6 +66,7 @@ from imap_processing.swe.l1b.swe_l1b import swe_l1b
 from imap_processing.ultra.l1a import ultra_l1a
 from imap_processing.ultra.l1b import ultra_l1b
 from imap_processing.ultra.l1c import ultra_l1c
+from imap_processing.ultra.l2 import ultra_l2
 
 logger = logging.getLogger(__name__)
 
@@ -1135,6 +1136,19 @@ class Ultra(ProcessInstrument):
                 dataset = load_cdf(dep.imap_file_paths[0])
                 data_dict[dataset.attrs["Logical_source"]] = dataset
             datasets = ultra_l1c.ultra_l1c(data_dict)
+
+        elif self.data_level == "l2":
+            all_pset_filepaths = dependencies.get_file_paths(
+                source="ultra", descriptor="pset"
+            )
+            # There can be many PSET files, so avoid reading them all in.
+            # The filename stem (logical_file_id) contains
+            # all the information needed in the key.
+            data_dict = {
+                pset_filepath.stem: pset_filepath
+                for pset_filepath in all_pset_filepaths
+            }
+            datasets = ultra_l2.ultra_l2(data_dict)
 
         return datasets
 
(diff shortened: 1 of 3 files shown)


imap_processing/cli.py
```diff
@@ -1136,6 +1137,19 @@ def do_processing(
                 data_dict[dataset.attrs["Logical_source"]] = dataset
             datasets = ultra_l1c.ultra_l1c(data_dict)
 
+        elif self.data_level == "l2":
+            all_pset_filepaths = dependencies.get_file_paths(
+                source="ultra", descriptor="pset"
+            )
+            # There can be many PSET files, so avoid reading them all in.
+            # The filename stem (logical_file_id) contains
```
tmplummer, line 1145: Just as an FYI, you should be able to get the sensor from the `Logical_source` of the CDF. It just seemed kind of odd to me to pass in a dictionary where every key is an attribute (`stem`) of the value (`Path`).

imap_processing/cli.py
```diff
@@ -1136,6 +1137,19 @@ def do_processing(
...
+            # all the information needed in the key.
+            data_dict = {
+                pset_filepath.stem: pset_filepath
+                for pset_filepath in all_pset_filepaths
+            }
+            datasets = ultra_l2.ultra_l2(data_dict)
```
tmplummer, line 1151: It looks like the type hints and docstring for `ultra_l2()` needs to be updated to accept `dict[str, Path]`.

tmplummer (approved):
Looks good. Just a couple of recommendations.

imap_processing/cli.py
nkerman: Ah, thanks! Nice catch.

imap_processing/cli.py
nkerman: That's a good point, and I've been thinking about whether I should switch the input `data_dict`, which I modeled on other inputs I saw in `cli.py` to some other type. What makes me hesitant though is that the simplest way to get the sensor number is just to look for "45sensor" in the filename. It's in the logical source too, but this way you don't have to read it in... But then it gets read in later anyways... 🙃 

I could probably just make this a list input and check the 0th item's logical source. I'll think on that and if there are any downsides.

nkerman: For this PR, I decided to leave it as a dict. I like that it's more consistent with the other Ultra levels, and the data dict's key will only be an attribute (`stem`) of the value (`Path`) when you're passing in file paths/`Paths`. If you're piping in the L1C Datasets directly, this is no longer the case.

PR #1702, opened 2025-05-06
Remove code to split MAG packets by day

maxinelasp:
# Change Summary
This piece of code is no longer needed now that we have finalized how packets work, and was breaking other things with the change to always use the input CLI start_date in the output files. 

## Overview

Removed some code which previously sorted packets by data for MAG L1A

Now, MAG L1A will always output 6 files.


lacoak21 (approved):
LGTM!

PR #1703, opened 2025-05-06
Add a query to check if the file exists pre-upload

maxinelasp:
# Change Summary
We were having infra errors all over the place with this bug. This makes it so if ANY of the files that are attempting upload already exist, we raise an error and don't upload ANY of them. This way we don't get partial file uploads with a failed processing job. 

## Overview

Add a check prior to upload to see if the file exists already.

## Testing

All the unit tests mocked an upload and tested it. Instead, I tested that they call "write_cdf". now, test_post_processing is the only place where we check if the file is uploaded, and it checks either case.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -22,6 +22,7 @@ from typing import final
 
 import imap_data_access
 import xarray as xr
+from imap_data_access import ScienceFilePath
 from imap_data_access.processing_input import (
     ProcessingInputCollection,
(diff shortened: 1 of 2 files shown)


vmartinez-cu (approved):
Adding a check for existing files is a helpful addition. Nice job!

lacoak21 (approved):
Looks great- Thanks for doing that Maxine!

PR #1704, opened 2025-05-07
I-ALiRT - MAG validation with SC data

laspsandoval:
# Change Summary

## Overview
Received a spacecraft packet with MAG data so I'm going through and finding changes that are needed.

## Deleted Files
- time.py
   - Not needed since MAG regular processing has time functions
- test_time.py

## Updated Files
- parse_mag
   - sclk and sclk sub proved to be more reliable for determining correct order of the data compared with coarse and fine time
   - When data was missing the coarse data was equal to zero so we skip processing for that case
   - We were decoding the values as unsigned 16bit ints (-32,768 to 32,767) instead of signed for pri_x, pri_y, pri_z, sec_x, sec_y, sec_z and it was leading to incorrect values. So this data is now converted to a signed 16 bit int.
   - Removed duplicate time sorting.

## Testing
- test_parse_mag


imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -115,18 +114,24 @@ def extract_magnetic_vectors(science_values: xr.DataArray) -> dict:
         Magnetic vectors.
     """
     # Primary sensor:
-    pri_x = (int(science_values[0]) >> 8) & 0xFFFF
-    pri_y = ((int(science_values[0]) << 8) & 0xFF00) | (
-        (int(science_values[1]) >> 16) & 0xFF
+    pri_x = int(np.uint16((int(science_values[0]) >> 8) & 0xFFFF).astype(np.int16))
```
maxinelasp: So this goes from int->np.uint16->np.int16->int? Is there any way to skip any of these conversions?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -306,20 +311,16 @@ def process_packet(
...
 
-    # Note that the fine time second is split into 65535.
-    time_seconds = calculate_time(
-        accumulated_data["mag_acq_tm_coarse"],
-        accumulated_data["mag_acq_tm_fine"],
-        65535,
-    )
+    # Subsecond time conversion specified in 7516-9054 GSW-FSW ICD.
+    # Value of SCLK subseconds, unsigned, (LSB = 1/256 sec)
+    met = accumulated_data["sc_sclk_sec"] + accumulated_data["sc_sclk_sub_sec"] / 256
```
maxinelasp: Maybe there could be a "SCTimeTuple" (to use my own naming scheme) and move this magic number/calculation that I assume you'll be using again into a shared method?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -332,14 +333,21 @@ def process_packet(
...
         if not np.array_equal(pkt_counter, np.arange(4)):
-            logger.warning(
+            logger.info(
                 f"Group {group} does not contain all values from 0 to "
                 f"3 without duplicates."
             )
             continue
 
+        if (coarse_time == 0).all():
+            logger.info(f"Group {group} contains zero timestamps.")
```
maxinelasp: nitpick - I read this as "there are no timestamps in group" so maybe rephrase? 
```suggestion
            logger.info(f"Group {group} contains timestamps equal to zero.")
```

imap_processing/ialirt/utils/time.py
```diff
@@ -1,29 +0,0 @@
-"""Common time functions for I-ALiRT instruments."""
-
-import xarray as xr
-
-
-def calculate_time(
```
maxinelasp, line 6: Oh look, this is the function I suggested. Why delete this?

imap_processing/tests/ialirt/unit/test_parse_mag.py
```diff
@@ -212,12 +248,68 @@ def test_calculate_l1b(grouped_data, xarray_data, calibration_dataset):
...
+            mag_sc_test_data["pri_fine"] == packet["pri_fintm"]
+        )
+        matching_rows = mag_sc_test_data[index]
+
+        if matching_rows.empty:
+            continue
+
+        row = matching_rows.iloc[0]
+
+        # Row that does not match
```
maxinelasp, line 294: Why no match? Can you elaborate in the comment?

maxinelasp (commented):
I have a few kind of nitpicky comments but overall looks good!

imap_processing/tests/ialirt/unit/test_parse_mag.py
laspsandoval: Everything matches except pri_x. And I don't know why. I'll ask at tomorrow's meeting.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: I know. But here is why:

1. 
science_values[0]
<xarray.DataArray 'mag_data' ()> Size: 8B
array(262619.)
Coordinates:
    epoch    int64 8B 798510969184000000
    group    int64 8B 1
    
So have I have to convert it to int here.

2. 
If I have a large value (which I do in some cases) in throws an error if I use np.int16 directly because signed 16 bit integers go from -32768 to 32767
So first I store only the 16 lower bits
np.uint16(val) → 56149
Then I interpret them as an unsigned number.
np.int16(np.uint16(val)) → -9387

I got rid of one int at least though.

imap_processing/ialirt/utils/time.py
laspsandoval: That was using coarse and fine time.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: Ok. I'll update the time.py and bring it back :)

imap_processing/ialirt/utils/time.py
maxinelasp: Ahh, I see!

imap_processing/tests/ialirt/unit/test_parse_mag.py
maxinelasp: Haha, I've been there. If it's a temporary thing, that's fine to leave.

imap_processing/ialirt/l0/parse_mag.py
maxinelasp: Mysterious. Thanks for explaining!

maxinelasp (approved):
Thanks for making those changes!

PR #1705, opened 2025-05-07
I-ALiRT: Create SWAPI count rate optimization function

torimarbois:
# Change Summary

## Overview
Add the main piece of SWAPI I-ALiRT processing: a beefy optimization function for coincidence count rates. You can compare the code to the science document located [here](https://lasp.colorado.edu/galaxy/pages/viewpage.action?pageId=146768943&spaceKey=IMAP&title=IMAP%2BAlgorithm%2BDocument%2Bfrom%2BInstrument%2BTeams&preview=/146768943/247997433/05899-Algorithms_AN%20-%2006%20-%20signed.pdf), if desired.

## New Files

- tests/ialirt/test_data/ialirt_test_data.csv
   - This includes energy passbands/steps and count rates that the SWAPI team used to verify their model, and theoretically these energy steps won't change, so I'm using it as a sort of look up table.

## Updated Files

- ialirt/l0/process_swapi.py
   - Adds the new optimization function and organizes its call & outputs within the overall processing function.
- tests/ialirt/unit/test_process_swapi.py
   - Test that the optimization function does not return `None`.

## Testing

I added unit tests to verify results aren't None, but verifying the output values against real data felt too tricky to try and implement (mainly because I don't have an exact set of data that produces an exact set of optimized parameters).

However, I did graph my results and compared the results of the optimized parameters my model found to the results with the parameters that the science model found when using the six highest count rate values (which is what the science team did):
image
The blue line is the model using the optimized parameters from the science document, the orange line is the model using the optimized parameters that this code found, and the blue points are the provided data points. As you can see, they are very close!


imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -3,17 +3,122 @@
...
+        Dictionary containing the optimized speed, density, and temperature values for
+        each sweep included in the input count_rates array.
+    """
+    # Read in energy passbands
+    energy_data = pd.read_csv(
+        f"{imap_module_directory}/tests/ialirt/test_data/ialirt_test_data.csv"
+    )
+    energy_passbands = energy_data["Energy [eV/q]"].to_numpy()
+
+    def count_rate(
```
laspsandoval: I personally prefer keeping functions separately. It's simpler for unit testing and reuse is less complicated.

imap_processing/tests/ialirt/unit/test_process_swapi.py
```diff
@@ -87,7 +90,20 @@ def test_decom_packets(xarray_data, swapi_test_data):
 
 
 def test_process_swapi_ialirt(xarray_data):
-    """Placeholder test for the process_swapi_ialirt function."""
+    """Test that the process_swapi_ialirt function returns expected keys."""
 
     swapi_result = process_swapi_ialirt(xarray_data)
```
laspsandoval, line 140: Based on the description of the test, add some more keys to test.

imap_processing/tests/ialirt/unit/test_process_swapi.py
```diff
@@ -87,7 +90,20 @@ def test_decom_packets(xarray_data, swapi_test_data):
...
+def test_optimize_parameters(xarray_data):
+    """Test the optimize_pseudo_parameters function."""
+
+    energy_data = pd.read_csv(
+        f"{imap_module_directory}/tests/ialirt/test_data/ialirt_test_data.csv"
+    )
+    count_rates = energy_data["Count Rates [Hz]"].to_numpy()
+
+    result = optimize_pseudo_parameters(count_rates)
+    assert result is not None
```
laspsandoval: Do you have a result to compare to? If not could you create some input data and then have expected output data?

imap_processing/tests/ialirt/unit/test_process_swapi.py
```diff
@@ -3,7 +3,10 @@
 import pytest
 
 from imap_processing import imap_module_directory
-from imap_processing.ialirt.l0.process_swapi import process_swapi_ialirt
+from imap_processing.ialirt.l0.process_swapi import (
+    optimize_pseudo_parameters,
+    process_swapi_ialirt,
+)
 from imap_processing.utils import packet_file_to_datasets
```
laspsandoval, line 14: Make certain to add a test for the count_rates function.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -3,17 +3,122 @@
...
+        temp : float
+            Temperature [K].
+
+        Returns
+        -------
+        count_rate : float | np.ndarray
+            Particle coincidence count rate.
+        """
+        # Scientific constants used in optimization model
+        boltz = 1.380649e-23  # Boltzmann constant, J/K
```
laspsandoval: This is enough constants that I think you could bring it out to a separate dataclass like:
/Users/lasa6858/imap_processing/imap_processing/ultra/constants.py

That way they are stored in a single place and we can see the algorithm more closely.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -3,17 +3,122 @@
...
+
+        # thermal velocity of solar wind ions
+        thermal_velocity = np.sqrt(2 * boltz * temp / prot_mass)
+        beta = 1 / (thermal_velocity**2)
+        # convert energy to Joules
+        center_speed = np.sqrt(2 * energy_pass * 1.60218e-19 / prot_mass)
+        speed = speed * 1000  # convert km/s to m/s
+        density = density * 1e6  # convert 1/cm**3 -to 1/m**3
+
+        return (
```
laspsandoval: In the notes section could you put where these equations are from?

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -3,17 +3,122 @@
...
+            * np.sqrt(np.pi / (beta * speed * center_speed))
+            * erf(np.sqrt(beta * speed * center_speed) * (az_fov / 2))
+            * (
+                center_speed**4
+                * speed_energy_width
+                * np.arcsin(thermal_velocity / center_speed)
+            )
+        )
+
+    initial_param_guess = np.array([550, 5.27, 1e5])
```
laspsandoval, line 109: Why is this the guess here?

imap_processing/tests/ialirt/unit/test_process_swapi.py
```diff
@@ -87,7 +90,20 @@ def test_decom_packets(xarray_data, swapi_test_data):
...
+    assert len(swapi_result["met"]) == len(swapi_result["pseudo_speed"])
+
+
+def test_optimize_parameters(xarray_data):
+    """Test the optimize_pseudo_parameters function."""
+
+    energy_data = pd.read_csv(
+        f"{imap_module_directory}/tests/ialirt/test_data/ialirt_test_data.csv"
+    )
+    count_rates = energy_data["Count Rates [Hz]"].to_numpy()
```
laspsandoval: So will count_rates be a lookup table that is used in operations? Or is it just test data? If it's a lookup table we should put it in the ialirt (not test) directory.

laspsandoval (commented):
Most of my comments are about documentation since it is difficult for me to follow where the equations are coming from; not having a SWAPI background. Nice job. Thanks for doing this.

laspsandoval:
Another question: how close should the orange and blue lines be from each other? Did the instrument team give any idea of criteria for success. Could you plot the difference of the two as well?

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -3,17 +3,122 @@
...
+    solution_dict = {  # type: ignore
+        "pseudo_speed": [],
+        "pseudo_density": [],
+        "pseudo_temperature": [],
+    }
+
+    if count_rates.ndim > 1:
```
greglucas: Are we expecting both cases of array and float inputs? Can you just do one or the other and fix what the input is here.

imap_processing/ialirt/l0/process_swapi.py
torimarbois: I thought it should go before the function call. Do you think it should be elsewhere?

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -3,17 +3,122 @@
...
+                count_rate,
+                energy_passbands,
+                current_sweep_count_rates,
+                initial_param_guess,
+            )
+            solution_dict["pseudo_speed"].append(sol[0][0])
+            solution_dict["pseudo_density"].append(sol[0][1])
+            solution_dict["pseudo_temperature"].append(sol[0][2])
+    else:
+        sol = curve_fit(count_rate, energy_passbands, count_rates, initial_param_guess)
```
bishwassth: I suggest doing a weighted fit (inverse variance weighted). The "Count Rates Error" in the CSV file (3rd column) can be used as weights. Also, suggest checking the reduced chi-square value for the goodness of the fit.

In addition, are you using only 6 energy bins around the maximum count rates (1 where max count rate is + 3 on the left + 2 on the right) for this fitting?

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -3,17 +3,122 @@
...
+        Returns
+        -------
+        count_rate : float | np.ndarray
+            Particle coincidence count rate.
+        """
+        # Scientific constants used in optimization model
+        boltz = 1.380649e-23  # Boltzmann constant, J/K
+        at_mass = 1.6605390666e-27  # atomic mass, kg
+        prot_mass = 1.007276466621 * at_mass  # mass of proton, kg
+        eff_area = 3.3e-5 * 1e-4  # effective area, meters squared
```
bishwassth: Can we edit this code directly later to update the value of the effective area?

bishwassth (commented):
My main comment is on the implementation of the non-linear fitting.

imap_processing/ialirt/l0/process_swapi.py
torimarbois: Laura suggested adding a comment explaining where this guess was pulled from in the algorithm document.

bishwassth: The initial guess for the pseudo-speed can be obtained from the energy corresponding to the maximum/peak count rate (energy_peak_rate), i.e., 
**speed_guess = sqrt(2 * energy_peak_rate * 1.60218e-19 / proton_mass)/1000** km/s.
It is not straightforward to come up with a good initial guess for the pseudo-density and temperature.  Some nominal values, like the following, should be okay.
 **dens_guess = 5 cm^-3**, and
 **T_guess = 1e5 K**.

imap_processing/ialirt/l0/process_swapi.py
torimarbois: Yes.

imap_processing/ialirt/l0/process_swapi.py
torimarbois: I don't have a formula to calculate the count rates error in real-time processing, so unless that gets created and I can implement it in the code, it doesn't make sense to me to add them as weights here when I only have one instance of them in the test data you sent me. Unless the count rates error remains the same for every passband every time?

I can work on getting the chi-square value from curve_fit, but it's surprisingly not a straightforward calculation.

Yes, the fit you see in the plots shown above are including only those 6 energy steps.

imap_processing/tests/ialirt/unit/test_process_swapi.py
torimarbois: @bishwassth are the energy steps in the spreadsheet you sent me (in `ialirt_test_data.csv`) going to be the ones used for real-time processing? Should I use them as a lookup table for this parameter optimization?

torimarbois:
> Another question: how close should the orange and blue lines be from each other? Did the instrument team give any idea of criteria for success. Could you plot the difference of the two as well?

@laspsandoval This is the percent difference between the expected count rates & the actual count rates calculated with the optimized parameters for the 6 points surrounding the maximum count rate. @bishwassth is this satisfactory?

image

imap_processing/ialirt/l0/process_swapi.py
bishwassth: I was under the impression that the SWAPI science processing code up to the level 2 was used for the SWAPI I-ALiRT, just for coarse scan, though (not the fine scans), and it creates arrays of energy (which is always fixed for coarse scans), count rates, and count rate errors for each 12 s sweep. If it is only using the part of the code that creates arrays of _energy_ and _count rates_, I can provide a conversion factor from count rates to _count rates error_. Of course, the count rate error doesn't remain the same over time.

imap_processing/tests/ialirt/unit/test_process_swapi.py
bishwassth: The first 63 energy steps are very close to what will be used for real-time processing, but may not be exactly similar. The energies for real-time processing will be from the ESA unit conversion ADP and will be in the form of a lookup table (e.g., Table 3 in the algorithms document). BTW, could you confirm if the I-ALiRT code is using "ESA unit conversion ADP" or not? If not, we have to supply them in a different lookup table.

imap_processing/ialirt/l0/process_swapi.py
torimarbois: I was mistaken, there is code available for me to calculate the count rate error. Right now, the SWAPI science processing code calculates that as the `sqrt()` of the count rate. Is that the same conversion factor I should use for this?

imap_processing/tests/ialirt/unit/test_process_swapi.py
torimarbois: I see the file called `imap_swapi_esa-unit-conversion_20250211_v000.csv` in our lookup table folder for SWAPI. Is this what you mean? And are the energy passbands listed in this file? None of the columns are named simiilar to that.

bishwassth: Yes, that is the correct file to use. You have to read the columns for "Energy" in this file from ESA step # 0-62. Note that the energy values for the first 34 ESA step # (0-33) are fixed to "1,163", which will be updated to realistic values in space.

imap_processing/ialirt/l0/process_swapi.py
bishwassth: The error in the count rate is calculated as: _sqrt(counts)/TIME_PER_BIN_, where _TIME_PER_BIN = 12/72_ s and _counts_ are produced from L1 science processing.

torimarbois: @bishwassth can you look into the [scipy `curve_fit` function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html) and tell me how you'd like me to incorporate the count rate errors as weights? I tried to pass them in as the `sigma` parameter, but when I did that, the function errored by reaching the max recursive function call number without finding the optimal parameters.

I'm not sure if `sigma` is the right parameter to assign the errors to, or if it's some other issue.

Ex:
I did `sol = curve_fit(
            f=count_rate,
            xdata=energy_passbands,
            ydata=current_sweep_count_rates,
            p0=initial_param_guess,
            sigma=current_sweep_count_rate_errors,
        )`

Where the count rate value is `4167.6646` and the error value is `157.97491`.

And got this error: `RuntimeError: Optimal parameters not found: Number of calls to function has reached maxfev = 800.`.

bishwassth: I am not much familiar with the Python libraries, but according to the scipy curve_fit documentation, _sigma=count_rate_error_ seems the correct way to use count rate error as weight. I found [this page](https://stackoverflow.com/questions/27696324/using-scipy-optimize-curve-fit-with-weights), where an example of this type of implementation is explained.

Could you pass the following array values to curve_fit with the same initial parameter guess you are using, and let me know the output (_sol_)?

_xdata = [1257.56, 1369.26, 1489.38, 1621.08, 1764.61, 1921.41]
ydata = [245.95, 671.87, 1253.75, 1373.73, 869.83, 257.95]
sigma = [38.34, 63.37, 86.57, 90.62, 72.11, 39.27]_

torimarbois: @bishwassth in that case, `sol = ([5.45346497e+02 4.68051016e+00 1.18456907e+05], [[ 1.32580794e-01 -1.83099091e-03 -1.87799233e+02], [-1.83099091e-03  2.47663386e-03  1.78912082e+01], [-1.87799233e+02  1.78912082e+01  5.83434512e+06]])`.

I suspect that this works because it's a small set of data with variety, whereas the test data I've been given is all one value, which I listed above. It could be that I just need to change my test data, which is unfortunate because it's the test data I've been given.

bishwassth: @torimarbois In this example, the pseudo speed obtained from fitting looks very close to what I reported in the algorithms document. I am not sure which test data you are referring to. The subset of data (6 energy bins around the proton peak) I provided in the previous comment is extracted from the test data: "ialirt_test_data.csv". Note that the zero count rates in the test data and some non-zero count rates don't belong to the solar wind protons, and we need to exclude them during the fitting. Therefore, we choose only 6 energy bins around the peak count rates for the fitting. I hope this makes sense.

torimarbois: @bishwassth should the fitting for each sweep always only be the 6 energy bins around the peak count rate?

bishwassth: @torimarbois Yes.

torimarbois: @bishwassth sorry to rehash this one more time, but I want to make sure we're on the same page - you're saying that for every fitting, even during real-time processing (when we're NOT using test data), only the 6 points around the peak should be included in the model input. Right?

bishwassth: @torimarbois No worries. Yes, only 6 energy bins around the peak count rate should be included in the fitting. However, I would note that the position of the peak count rate in the ESA energy bins changes over time during real-time processing because of the variability in the solar wind conditions. So we have to find the position of the peak count rate in SWAPI's coarse energy bins in order to select the 6 energy bins used for fitting.

laspsandoval (approved):
LGTM!

PR #1706, opened 2025-05-07
1685 hi l2   l2 processing structure

### LINKED ISSUE #1685: Hi L2 - L2 processing structure
### Algorithm Description:

Create the structure needed to produce Hi L2 products

tmplummer:
# Change Summary

## Overview
This adds the initial work toward producing Hi L2 Maps.
Still to go:
- Use geometric factor and esa energy calibration data in intensity calculations
- Correctly combine calibration products at the same ESA
- Finalize Dataset for CDF writing

## New Files
- imap_processing/hi/l2/__init__.py
   - l2 module file
   - Add function for generating Hi L2 map given pointing sets and map configuration
   - Add function for computing ENA signal rates
   - Add initial function for computing ENA intensities and associated uncertainties
   - Add test coverage for the above new functions

## Updated Files
   - Rename some Hi PSET variables in `__init__` method of HiPointingSet
   - Check that variable renames worked.

Closes: #1685

### DIFF
--- a/imap_processing/hi/l2/hi_l2.py
+++ b/imap_processing/hi/l2/hi_l2.py
@@ -13,14 +13,14 @@ from imap_processing.spice.geometry import SpiceFrame
 logger = logging.getLogger(__name__)
 
 
-def hi_l2(
+def generate_hi_map(
     psets: list[str | Path],
     geometric_factors_path: str | Path,
     esa_energies_path: str | Path,
     cg_corrected: bool = False,
     direction: Literal["ram", "anti-ram", "full"] = "full",
     map_spacing_deg: int = 4,
-) -> list[xr.Dataset]:
+) -> xr.Dataset:
     """
     High level IMAP-Hi L2 processing function.
 
@@ -63,9 +63,8 @@ def hi_l2(
 
         # Background rate and uncertainty are exposure time weighted means in
         # the map.
-        pset.data.rename({"exposure_times": "exposure_factor"})
-        pset.data["background_rates"] *= pset.data["exposure_factor"]
-        pset.data["background_rates_uncertainty"] *= pset.data["exposure_factor"]
+        pset.data["bg_rates"] *= pset.data["exposure_factor"]
+        pset.data["bg_rates_unc"] *= pset.data["exposure_factor"]
 
         # Project (bin) the PSET variables into the map pixels
         rect_map.project_pset_values_to_map(
@@ -73,8 +72,8 @@ def hi_l2(
             [
                 "counts",
                 "exposure_factor",
-                "background_rates",
-                "background_rates_uncertainty",
+                "bg_rates",
+                "bg_rates_unc",
             ],
         )
 
@@ -84,8 +83,8 @@ def hi_l2(
     # Finish the exposure time weighted mean calculation of backgrounds
     # Allow divide by zero to fill set pixels with zero exposure time to NaN
     with np.errstate(divide="ignore"):
-        map_ds["background_rates"] /= map_ds["exposure_factor"]
-        map_ds["background_rates_uncertainty"] /= map_ds["exposure_factor"]
+        map_ds["bg_rates"] /= map_ds["exposure_factor"]
+        map_ds["bg_rates_unc"] /= map_ds["exposure_factor"]
 
     map_ds.update(calculate_ena_signal_rates(map_ds))
     map_ds.update(
@@ -117,7 +116,7 @@ def calculate_ena_signal_rates(map_ds: xr.Dataset) -> dict[str, xr.DataArray]:
     with np.errstate(divide="ignore"):
         # Calculate the ENA Signal Rate and Signal Rate Uncertainties
         signal_rate_vars["ena_signal_rates"] = (
-            map_ds["counts"] / map_ds["exposure_factor"] - map_ds["background_rates"]
+            map_ds["counts"] / map_ds["exposure_factor"] - map_ds["bg_rates"]
         )
         signal_rate_vars["ena_signal_rate_stat_unc"] = np.sqrt(
             map_ds["counts"] / map_ds["exposure_factor"]
--- a/imap_processing/hi/l2/hi_l2.py
+++ b/imap_processing/hi/l2/hi_l2.py
@@ -1,0 +1,193 @@
+"""IMAP-HI L2 processing module."""
+
+import logging
+from pathlib import Path
+from typing import Literal
+
+import numpy as np
+import xarray as xr
+
+from imap_processing.ena_maps.ena_maps import HiPointingSet, RectangularSkyMap
+from imap_processing.spice.geometry import SpiceFrame
+
+logger = logging.getLogger(__name__)
+
+
+def hi_l2(
+    psets: list[str | Path],
+    geometric_factors_path: str | Path,
+    esa_energies_path: str | Path,
+    cg_corrected: bool = False,
+    direction: Literal["ram", "anti-ram", "full"] = "full",
+    map_spacing_deg: int = 4,
+) -> list[xr.Dataset]:
+    """
+    High level IMAP-Hi L2 processing function.
+
+    Parameters
+    ----------
+    psets : list of str or Path
+        List of input PSETs to make a map from.
+    geometric_factors_path : str or Path
+        Where to get the geometric factors from.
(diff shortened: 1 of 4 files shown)


imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,192 @@
...
+    rect_map = RectangularSkyMap(
+        spacing_deg=map_spacing_deg, spice_frame=SpiceFrame.ECLIPJ2000
+    )
+
+    # TODO: Implement Compton-Getting correction
+    if cg_corrected:
+        raise NotImplementedError
+    # TODO: Implement directional filtering
+    if direction != "full":
+        raise NotImplementedError
```
nkerman, line 60: Nice hooks! 

This isn't necessary until you start implementing it, but for direction/spin_phase, maybe have an enum?

imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,192 @@
...
+
+    for pset_path in psets:
+        logger.info(f"Processing {pset_path}")
+        pset = HiPointingSet(pset_path)
+
+        # Background rate and uncertainty are exposure time weighted means in
+        # the map.
+        pset.data["bg_rates"] *= pset.data["exposure_factor"]
+        pset.data["bg_rates_unc"] *= pset.data["exposure_factor"]
```
nkerman: Maybe do this with a list of `VARS_TO_WEIGHT_BY_EXPOSURE`? Then you can do `pset.data[VARS_TO_WEIGHT_BY_EXPOSURE] *= pset.data["exposure_factor"]`. Not mandatory, just something I think Greg suggested in the Ultra code.

imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    signal_rate_vars = {}
+    # Allow divide by zero to fill set pixels with zero exposure time to NaN
+    with np.errstate(divide="ignore"):
+        # Calculate the ENA Signal Rate and Signal Rate Uncertainties
+        signal_rate_vars["ena_signal_rates"] = (
+            map_ds["counts"] / map_ds["exposure_factor"] - map_ds["bg_rates"]
+        )
+        signal_rate_vars["ena_signal_rate_stat_unc"] = np.sqrt(
+            map_ds["counts"] / map_ds["exposure_factor"]
```
nkerman: Is it possible this should be 
$\sqrt{counts}/ exposure$

rather than 
$\sqrt{counts/ exposure}$
?

I don't think the latter makes sense for this sort of calculation.

imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+    # Statistical fluctuations may result in a negative ENA signal rate after
+    # background subtraction. A negative signal rate is nonphysical. See IMAP Hi
+    # Algorithm Document section 3.1.1
+    signal_rate_vars["ena_signal_rates"].values[
+        signal_rate_vars["ena_signal_rates"].values < 0
+    ] = 0
```
nkerman, line 139: You can also use `np.maximum(signal_rate_vars["ena_signal_rates"].values, 0)` here. Not sure if that would be any more efficient. 🤷

imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,192 @@
...
+    # Minimum count uncertainty is 1 for any pixel that has non-zero exposure time
+    # See IMAP Hi Algorithm Document section 3.1.1
+    signal_rate_vars["ena_signal_rate_stat_unc"].values[
+        (signal_rate_vars["ena_signal_rate_stat_unc"].values == 0)
+    ] = 1
```
nkerman: Is this currently limited to non-zero exposure pixels?

imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,192 @@
...
+        coords=[map_ds["esa_energy_step"], map_ds["calibration_prod"]],
+    )
+    # TODO: Implement esa energies lookup
+    if esa_energies_path is not None:
+        raise NotImplementedError
+    esa_energy = xr.ones_like(map_ds["esa_energy_step"])
+
+    # Convert ENA Signal Rate to Flux
+    intensity_vars = {
+        "ena_intensity": map_ds["ena_signal_rates"] / geometric_factor / esa_energy,
```
nkerman: Nit: for legibility, I strongly prefer `A / (B * C)` over `A / B / C`.

imap_processing/hi/l2/hi_l2.py
```diff
@@ -0,0 +1,192 @@
...
+        / geometric_factor
+        / esa_energy,
+        "ena_intensity_sys_err": map_ds["bg_rates_unc"] / geometric_factor / esa_energy,
+    }
+
+    # TODO: Correctly implement combining of calibration products. For now, just sum
+    intensity_vars["ena_intensity"] = intensity_vars["ena_intensity"].sum(
+        dim="calibration_prod"
+    )
+    intensity_vars["ena_intensity_stat_unc"] = np.sqrt(
```
nkerman, line 195: Can you clarify in a comment somewhere what this "calibration_product" axis represents?

imap_processing/tests/hi/test_hi_l2.py
```diff
@@ -0,0 +1,120 @@
+"""Test coverage for imap_processing.hi.l2.hi_l2.py"""
```
nkerman, line 1: It probably won't fit neatly into a unit test, but out of curiosity, have you tried taking synthetic data and running it through? I find that a super useful way to find bugs.

nkerman (commented):
Looks good, Tim! A couple comments

imap_processing/hi/l2/hi_l2.py
tmplummer: Yep, you're right. Good catch.

imap_processing/hi/l2/hi_l2.py
tmplummer: Yes, I can add a comment, but the above division by `exposure_factor` will result in `ena_signal_rate_stat_unc` being NaN anywhere that there is zero exposure.

imap_processing/tests/hi/test_hi_l2.py
tmplummer: Yes, the first test added here is an integration level test that runs a simulated PSET file to make sure that this code plays nice with the xr.Dataset from a PSET CDF file. The other tests use synthetic data. While I could certainly design the synthetic data to more exhaustively test the functions, I think that I am hitting the main points. The verification process should help identify additional bugs/issues.

imap_processing/hi/l2/hi_l2.py
tmplummer: It looks like `np.maximum` creates a new array where the boolean indexing modifies the values in place. I doubt that the performance is much different for this size of array, but I generally lean toward modifying in place.

imap_processing/hi/l2/hi_l2.py
tmplummer: Actually, thanks for this comment. It made me realize that the correct value to set is `1/exposure_factor`.

imap_processing/tests/hi/test_hi_l2.py
nkerman: Yep! I guess I mean also taking a look at the output as a plot on the sky. 

E.g., Put in 5 psets taken 10 days apart, then plotting the counts or flux, do you get 5 great circles on the sky?

nkerman (approved):
LGTM! I added one more comment about qualitative testing (not a pytest runnable test, but just trying to get some output map plots from realistic-ish input data)

You might have already done this though, or may not find it necessary- it's just a helpful "sanity check"

imap_processing/tests/hi/test_hi_l2.py
tmplummer: Ah, yes. I have played around with the simulated data from Nick and have run one, a few, and a full year of data through. The mapping part looks as expected, though I am not yet getting the IMAP logo as expected. I suspect that L1B and L1C validation will reveal some issues and 🤞 result in the correct map.

PR #1707, opened 2025-05-07
Ultra l1c - cdf updates

### LINKED ISSUE #1608: Ultra l1b - make cdf instead of l1b de in conftest.py
### Description of the issue

 - Take out l1b de cdf and put it in s3 so it doesn't need to be generated in the conftest.py
 - Add efficiency equations

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

laspsandoval:
# Change Summary

## Overview
Tying up some loose ends. Adding l1b de cdf as external download file. Adding "Data_version" to the test to be able to populate global attributes. Creating a single FILL value in badtimes.py and cullingmask.py since an empty cdf was causing problems.

## Updated Files
- badtimes.py
   - Nick was having problems with this cdf because it was empty. Added an if statement that populates the cdf with a single FILL value if it is empty.
- cullingmask.py
   - Same as what was done with badtimes.py.
- de.py
   - Adding the event efficiency in.
   - Added default float 64 so I would have a proper FILL value for the cullingmask.py and badtimes.py
- ultra_l1_utils.py
   - From Tenzin: we need to make check_schema=False to properly populate the attributes

## Testing
- test_ultra_l1a.py
   - Added a “Data_version” to the attributes here since global attributes do not get added if this value is none. And also set ISTP=True
- test_ultra_l1b.py
   - Added a “Data_version” to the attributes here since global attributes do not get added if this value is none. 
- conftest.py
   - Added the l1b de cdf to the files available to download from s3.
- test_de.py
   - download l1b_de cdf instead of relying on data generated in conftest.py
- test_badtimes.py
   - Test for empty dataset issue
- test_culingmask.py
   - Test for empty dataset issue

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1a.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1a.py
@@ -70,7 +70,8 @@ def test_cdf_aux(ccsds_path_theta_0):
     """Tests that CDF file can be created."""
 
     test_data = ultra_l1a(ccsds_path_theta_0, apid_input=ULTRA_AUX.apid[0])
-    test_data_path = write_cdf(test_data[0])
+    test_data[0].attrs["Data_version"] = "v999"
+    test_data_path = write_cdf(test_data[0], istp=True)
 
     assert test_data_path.exists()
     assert test_data_path.name == "imap_ultra_l1a_45sensor-aux_20240207_v999.cdf"
@@ -79,6 +80,7 @@ def test_cdf_aux(ccsds_path_theta_0):
 def test_cdf_rates(ccsds_path_theta_0):
     """Tests that CDF file can be created."""
     test_data = ultra_l1a(ccsds_path_theta_0, apid_input=ULTRA_RATES.apid[0])
+    test_data[0].attrs["Data_version"] = "v999"
     test_data_path = write_cdf(test_data[0], istp=False)
 
     assert test_data_path.exists()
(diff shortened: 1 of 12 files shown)


imap_processing/tests/ultra/unit/test_ultra_l1a.py
```diff
@@ -79,6 +80,7 @@ def test_cdf_aux(ccsds_path_theta_0):
 def test_cdf_rates(ccsds_path_theta_0):
     """Tests that CDF file can be created."""
     test_data = ultra_l1a(ccsds_path_theta_0, apid_input=ULTRA_RATES.apid[0])
+    test_data[0].attrs["Data_version"] = "v999"
```
laspsandoval, line 83: I need to come back here and figure out why istp=True causes problems. It has to do with trying to save a binary blob to a cdf. I think I'll do that in another PR though.

laspsandoval:
pre-commit.ci autofix

imap_processing/tests/ultra/unit/test_ultra_l1a.py
nkerman: I think it's fine to have `istp=False` at L1.

nkerman (approved):
This all looks reasonable to me.

imap_processing/tests/ultra/unit/test_ultra_l1a.py
laspsandoval: We had an email chain going with Larry and for some reason setting istp=True puts the data in a format that Larry couldn't read. I'll see if I can forward it to you. I usually put a breakpoint in my tests to create my data products so this was causing problems.

PR #1708, opened 2025-05-07
Ultra L2: define `delta_minus`/`delta_plus` for energy bins WRT geometric mean

### LINKED ISSUE #1697: Ultra L2: Set `energy_delta_minus/plus` correctly
See TODO in ultra_l2.py

nkerman:
# Change Summary

Calculate the asymmetric `delta_minus`/`delta_plus` of Ultra's energy bins.

Closes #1697 

## Overview

Ultra uses a geometric mean of energy bin the energy bin:

$E_{\mu,geom} = \sqrt{E_{min} \times E_{max}}$

So we need to calculate the asymmetric `delta_minus`/`delta_plus` of Ultra's energy bins

$\Delta_- = | E_{\mu,geom} - E_{min}| $
$\Delta_+ = | E_{max} - E_{\mu,geom}| $

This is probably retrievable from the geometric mean energy and the total energy range of the energy bin, but it's far simpler to just define a function which calls `build_energy_bins()` and does the arithmetic above. Because `build_energy_bins` is set to provide a constant output, this should be safe. 

## Updated Files

   - Define `get_energy_delta_minus_plus()` function
   - Use the new function to get the energy deltas

## Testing

- Test the function using mocked outputs of `build_energy_bins()`

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
@@ -10,6 +10,7 @@ from imap_processing.ultra.l1c import ultra_l1c_pset_bins
 from imap_processing.ultra.l1c.ultra_l1c_pset_bins import (
     build_energy_bins,
     get_background_rates,
+    get_energy_delta_plus_minus,
     get_helio_exposure_times,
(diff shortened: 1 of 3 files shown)


lacoak21 (approved):
LGTM!

Issue #1641, opened 2025-04-22
Hi PSET - epoch should be pointing start time


tmplummer:
### Algorithm Description:
Currently, the epoch time is the first CCSDS time. Use pointing start time instead.

As part of this, also add the pointing midpoint time. Make it consistent with Lo and Ultra.

PR #1714, opened 2025-05-08
Lo L1C - Initialize PSET

### LINKED ISSUE #1713: Initialize Lo L1C Pointing Set
- Epoch should be set to the midpoint of all the events
- goodtimes need to be filtered using the sweep table's good times list

sdhoyt:
# Change Summary

## Overview
This PR initlalizes the Lo L1C Pointing Set, sets the epoch to the mid-point of the direct events in the pointing and filters for good times (I know the format of the ancillary file, but need to create a fake one to use for testing)

## Updated Files
   - added pset dataset initialization that sets the epoch to the midpoint
   - added good times filtering using the sweeptable's good times list. I will probably revisit the ancillary file handling once I get to the background rates file
   - added sci/anc inputs for the l1c processing function

## Testing
Added unit tests for pset init and good times filtering

Closes #1713


imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -31,34 +32,96 @@ def lo_l1c(dependencies: dict) -> list[Path]:
...
+
+
+def initialize_pset(
+    l1b_de: xr.Dataset, attr_mgr: ImapCdfAttributes, logical_source: str
+) -> xr.Dataset:
+    """
+    Initialize the PSET dataset and set the Epoch.
+
+    The Epoch time is set to the mid-point of the L1B
+    Direct Event times. There is one Epoch per PSET file.
```
greglucas: I don't think this is what we should do because this is dependent on the distribution of events you are getting and doesn't have a real "meaning". My initial take is that this should be either the start of the pointing / collection period, or the center of that period. i.e. we want the center of the accumulation/integration period not the central point measured.

I'm not sure how others are handling this though, so would be good to get their feedback on what epoch they are setting on psets @nkerman, @laspsandoval , @subagonsouth?

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -31,34 +32,96 @@ def lo_l1c(dependencies: dict) -> list[Path]:
...
+
+    # Create a mask for epochs within any of the start/end time ranges
+    mask = np.zeros_like(l1b_de["epoch"], dtype=bool)
+
+    # Iterate over the good times and create a mask
+    for start, end in zip(goodtimes_start, goodtimes_end):
+        mask |= (l1b_de["epoch"] >= start) & (l1b_de["epoch"] < end)
+
+    # Filter the dataset using the mask
+    filtered_epochs = l1b_de.sel(epoch=mask)
```
greglucas: It looks like this is a "good" mask? If so, could we label this as something like mask_good, mask_selection to indicate it is selecting the good values and not pruning the bad ones? I think that would make it more obvious to me at first glance.

imap_processing/tests/lo/test_anc/imap_lo_l1c-sweeptable_20250415_v001.csv
```diff
@@ -0,0 +1,3 @@
+YYYYDDD,GoodTime_strt,GoodTime_end
```
greglucas: Is it a typo with `strt` missing the "a" or that is expected?

The table header says "DDD" but it looks like you are putting in "MMDD" in the values below.

imap_processing/lo/l1c/lo_l1c.py
nkerman: ~~I was planning to make this the earliest epoch of the pointing set - this is essentially the start time\*. I vaguely recall that being what the IMWG had been talking about, but I'm not sure if anyone had a definite answer as to what is best.~~

~~\* This isn't technically the very start of the collection period, as it'll be the midpoint of the earliest pointing.~~

---
~~Key to note is that for the ENA maps, we will have pixel-specific observation date, which (while problematic in their own way) can better describe this quantity, especially for Lo/Hi.~~

---

I'm sorry I misread this as the epoch of the L2 Map- Ignore my previous comments!

tmplummer: SPDF provided feedback that for these long bin-times (~24-hours), `epoch` should be the start of the data collection time. For Hi, I am currently using the time of the first DE in the PSET, but... I have a ticket to change that to the start of the Pointing as defined by the repoint-table (#1641).
Ideally, one of us will implement a common function in `spice.repoint.py` that gets the start time of the Pointing for an input Pointing #.

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: Yeah, _good_ point

imap_processing/tests/lo/test_anc/imap_lo_l1c-sweeptable_20250415_v001.csv
sdhoyt: That's how it's spelled in the sample ancillary file that Lo sent. I'll fix the date issue. I got a new file from them that has some changes so I'll add those changes to this PR

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: @subagonsouth Just to make sure I understand, are you planning to use the repoint number from the L1B DE product to determine which pointing to reference in the repointing table? Then use the first time in the table for that repointing as your PSET epoch time?

tmplummer: @sdhoyt , that seems reasonable. An alternate implementation would be to have a function that gets the pointing start time for any MET time within the pointing time range. So, you could use the MET time of any telemetry (for example, your first DE). I can't think of any advantages to either of those implementations.

sdhoyt: @subagonsouth  Thanks! I think using one of the DE epochs as an input would be easier. I created this issue: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1733

 I think I'm going to use the first DE epoch for now and then when I have time I can work on this issue and switch to using the pointing start.

tmplummer (approved):
LGTM

PR #1715, opened 2025-05-08
add kernel furnishing to InstrumentProcessing.pre_processing

### LINKED ISSUE #1683: Add kernel furnishing to preprocessing
### Algorithm Description:

Furnish spice kernels in pre_processing

tmplummer:
# Change Summary

## Overview
This PR adds the automatic furnishing of spice files in the `pre_processing` step that come in as dependencies. The changes in `cli.py` are straightforward, but the added test in `test_cli.py` is not. See change description below.

## Updated Files
- imap_processing/cli.py
   - Log the files that will be furnished
   - Furnish the files listed as source='spice'
   - The test has lots of mocks in order to get the right coverage and test what I wanted to test.
   - The imap_data_access.processing_input.download is mocked to copy files from the `tests/spice/test_data` path to the expected `DATA_DIR/spice` path.
   - The `do_processing` step is mocked to check that the expected kernels are furnished at the time that instrument processing is done.
   - `load_cdf` and `post_processing` are mocked so that they do nothing.

Closes: #1683

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -21,9 +21,11 @@ from pathlib import Path
 from typing import final
 
 import imap_data_access
+import spiceypy
 import xarray as xr
 from imap_data_access.processing_input import (
     ProcessingInputCollection,
+    SPICESource,
 )
 
 import imap_processing
@@ -404,7 +406,8 @@ class ProcessInstrument(ABC):
         Complete pre-processing.
 
         For this baseclass, pre-processing consists of downloading dependencies
-        for processing. Child classes can override this method to customize the
+        for processing and furnishing any spice kernels in the input
+        dependencies. Child classes can override this method to customize the
         pre-processing actions.
 
         Returns
@@ -415,6 +418,12 @@ class ProcessInstrument(ABC):
         dependencies = ProcessingInputCollection()
         dependencies.deserialize(self.dependency_str)
         dependencies.download_all_files()
+
+        # Furnish spice kernels
+        kernel_paths = dependencies.get_file_paths(source=SPICESource.SPICE.value)
+        logger.info(f"Furnishing kernels: {kernel_paths}")
+        spiceypy.furnsh([str(kernel_path) for kernel_path in kernel_paths])
+
         return dependencies
 
     @abstractmethod
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -422,7 +422,7 @@ class ProcessInstrument(ABC):
         # Furnish spice kernels
         kernel_paths = dependencies.get_file_paths(source=SPICESource.SPICE.value)
         logger.info(f"Furnishing kernels: {kernel_paths}")
-        spiceypy.furnsh([str(kernel_path) for kernel_path in kernel_paths])
+        spiceypy.furnsh([str(kernel_path.resolve()) for kernel_path in kernel_paths])
 
         return dependencies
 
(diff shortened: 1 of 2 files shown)


imap_processing/cli.py
```diff
@@ -415,6 +418,12 @@ def pre_processing(self) -> ProcessingInputCollection:
         dependencies = ProcessingInputCollection()
         dependencies.deserialize(self.dependency_str)
         dependencies.download_all_files()
+
+        # Furnish spice kernels
+        kernel_paths = dependencies.get_file_paths(source=SPICESource.SPICE.value)
+        logger.info(f"Furnishing kernels: {kernel_paths}")
+        spiceypy.furnsh([str(kernel_path) for kernel_path in kernel_paths])
```
greglucas: Does spiceypy not accept Path objects?
Do you need to resolve the path before turning it to a string?

imap_processing/tests/test_cli.py
```diff
@@ -316,6 +320,54 @@ def test_hit_l1a(mock_hit_l1a, mock_instrument_dependencies):
...
+def test_pre_processing(spice_test_data_path):
+    """Test coverage for ProcessInstrument.pre_processing method()."""
+    kernels_to_furnish = ["naif0012.tls", "imap_sclk_0000.tsc"]
+    dependency_str = (
+        "["
+        '{"type": "science","files": ["imap_hi_l2a_sensor45-de_20100105_v001.cdf"]},'
+        f'{{"type":"spice","files": ["{'", "'.join(kernels_to_furnish)}"]}},'
+        '{"type": "spice","files": ["imap_2010_104_01.repoint.csv"]}'
+        "]"
+    )
```
greglucas: This is pretty confusing to follow with all the interwoven single and double quotes, what about creating the object and calling `json.dumps()` on it?

`json.dumps([{"type": "science", "files": ["imap_hi_l2a_sensor45-de_20100105_v001.cdf"]}, ...])`

imap_processing/tests/test_cli.py
```diff
@@ -316,6 +320,54 @@ def test_hit_l1a(mock_hit_l1a, mock_instrument_dependencies):
...
+    ):
+        mock_download.side_effect = download_side_effect
+        mock_do_processing.side_effect = do_processing_side_effect
+
+        instrument = Hi(
+            "l1b", "sensor45-de", dependency_str, "20100105", None, "v001", True
+        )
+        # Verify no kernels are furnished prior to calling process
+        assert spiceypy.ktotal("ALL") == 0
+        instrument.process()
```
greglucas: Do you want to use the verification on the number after processing too? Not necessary, but would you want to say anything about what happens if you call `instrument.process()` again or with some kernels already pre-loaded?
```suggestion
        # Verify no kernels are furnished prior to calling process
        assert spiceypy.ktotal("ALL") == 0
        instrument.process()
        assert spiceypy.ktotal("ALL") == len(kernels_to_furnish)
```

greglucas (approved):
This is nice and simple as you mentioned! Only a few minor suggestions that I think might help with the failing test too. The logic looks good to me.

imap_processing/tests/test_cli.py
tmplummer: Great question. I just tested calling `instrument.process()` again and SPICE happily re-furnishes the same files. I hadn't really thought about the desired behavior for that case. It would be easy to clear the furnished kernels prior to the `furnsh()` call to ensure that only what is passed in are used 🤷

tmplummer: I decided that it makes sense to clean up the furnished kernels in `post_processing()`. It isn't guaranteed to run but as we discussed probably isn't that important. If we find the need to do it differently, it's easy to change.

PR #1723, opened 2025-05-12
Updates to pointing kernel generation

### LINKED ISSUE #739: Pointing Frame Kernel Generation Algorithm - per pointing kernel

tmplummer:
# Change Summary

## Overview
Sorry for the rather large PR.
This is mostly a restructuring for flexibility. I broke up the code that computes the average quaternion for a Pointing and the code that writes the CK. This will give us some flexibility in what gets written to the Pointing CK.

## Updated Files
   - Broke apart functions that compute the mean quaternion for a pointing and write the CK file.
   - Added comments section to the CK file
   - Added test coverage that checks that one or more segments can get written to a CK
   - Added test coverage that the comments section exists in the CK file

Closes: #739

### DIFF
--- a/imap_processing/spice/pointing_frame.py
+++ b/imap_processing/spice/pointing_frame.py
@@ -4,16 +4,28 @@ import logging
 import typing
 from collections.abc import Generator
 from contextlib import contextmanager
+from datetime import datetime, timezone
 from pathlib import Path
 
 import numpy as np
 import spiceypy
 from numpy.typing import NDArray
 
+from imap_processing.spice.geometry import SpiceFrame
 from imap_processing.spice.kernels import ensure_spice
-from imap_processing.spice.time import met_to_sclkticks, sct_to_et
+from imap_processing.spice.repoint import get_repoint_data
+from imap_processing.spice.time import et_to_utc, met_to_sclkticks, sct_to_et
 
 logger = logging.getLogger(__name__)
+
+POINTING_SEGMENT_DTYPE = np.dtype(
+    [
+        ("sclk_start", np.float64),
+        ("sclk_end", np.float64),
+        ("quaternion", np.float64, (4,)),
+        ("pointing_id", np.int32),
+    ]
+)
 
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


imap_processing/spice/pointing_frame.py
```diff
@@ -4,17 +4,29 @@
...
+from imap_processing.spice.time import et_to_utc, met_to_sclkticks, sct_to_et
 
 logger = logging.getLogger(__name__)
 
+POINTING_SEGMENT_DTYPE = np.dtype(
+    [
+        ("sclk_start", np.float64),
+        ("sclk_end", np.float64),
+        ("quaternion", np.float64, (4,)),
+        ("pointing_id", np.int32),
```
greglucas: Neither of these may matter, but figured I'd mention them in case.

Should sclk be a uint? I thought MET was the float version.

Should pointing_id be uint?

imap_processing/spice/pointing_frame.py
```diff
@@ -41,54 +53,125 @@ def open_spice_ck_file(pointing_frame_path: Path) -> Generator[int, None, None]:
...
+                # End times of individual pointing records within segment.
+                # Since there is only a single record this is equal to sclk_endtim.
+                np.array([segment["sclk_end"]]),  # Single stop time
+                # Average quaternion.
+                segment["quaternion"],
+                # 0.0 Angular rotation terms.
+                np.array([0.0, 0.0, 0.0]),
+                # Rates (seconds per tick) at which the quaternion and
+                # angular velocity change.
+                np.array([1.0]),
```
greglucas: I thought we were getting 10Hz quaternions? Does this need to be given to us or can we derive it from the kernel somehow?

imap_processing/spice/pointing_frame.py
```diff
@@ -41,54 +53,125 @@ def open_spice_ck_file(pointing_frame_path: Path) -> Generator[int, None, None]:
...
-    "imap_wkcp.tf",
-    "naif0012.tls"
-
-    Assumptions:
-    - The MOC has removed timeframe in which nutation/procession are present.
-    TODO: We may come back and have a check for this.
-    - The pointing frame kernel is made based on the most recent ck kernel.
-    In other words 1:1 ratio.
+
+    - Latest NAIF leapseconds kernel (naif0012.tls
```
greglucas:
```suggestion
    - Latest NAIF leapseconds kernel (naif0012.tls)
```

imap_processing/spice/pointing_frame.py
```diff
@@ -100,72 +183,71 @@ def create_pointing_frame(
...
+    repoint_df = repoint_df[
+        (repoint_df["repoint_end_et"] >= et_start)
+        & (repoint_df["repoint_start_et"] <= et_end)
+    ]
+    n_pointings = len(repoint_df) - 1
+
+    sclk_start = np.zeros(n_pointings, dtype=POINTING_SEGMENT_DTYPE["sclk_start"])
+    sclk_end = np.zeros(n_pointings, dtype=POINTING_SEGMENT_DTYPE["sclk_end"])
+    quaternion = np.zeros(n_pointings, dtype=POINTING_SEGMENT_DTYPE["quaternion"])
+    pointing_id = np.zeros(n_pointings, dtype=POINTING_SEGMENT_DTYPE["pointing_id"])
```
greglucas: Not sure if this would be helpful or not, but you could use structured or record arrays here I think with passing in the multiple dtype field. https://numpy.org/doc/stable/user/basics.rec.html

```
pointing_segments = np.zeros(n_pointings, dtype=POINTING_SEGMENT_DTYPE)
```

imap_processing/spice/pointing_frame.py
```diff
@@ -100,72 +183,71 @@ def create_pointing_frame(
...
-                # 0.0 Angular rotation terms.
-                np.array([0.0, 0.0, 0.0]),
-                # Rates (seconds per tick) at which the quaternion and
-                # angular velocity change.
-                np.array([1.0]),
-            )
+    return np.array(
+        list(zip(sclk_start, sclk_end, quaternion, pointing_id)),
+        dtype=POINTING_SEGMENT_DTYPE,
+    )
```
greglucas: If you are creating this structured array here already could you create this above initially like I mentioned and then fill in the pieces as you go?

```python
pointing_segments = np.zeros(n_pointings, dtype=POINTING_SEGMENT_DTYPE)

pointing_segments["sclk_start"][i_pointing] = spiceypy.sce2c(int(id_imap_sclk), pointing_start_et)

return pointing_segments
```

imap_processing/tests/spice/test_pointing_frame.py
```diff
@@ -79,107 +143,92 @@ def test_create_rotation_matrix(et_times, pointing_frame_kernels):
...
     rotation_matrix_expected = np.array(
         [[0.0000, 0.0000, 1.0000], [0.9104, -0.4136, 0.0000], [0.4136, 0.9104, 0.0000]]
     )
-    np.testing.assert_allclose(rotation_matrix_1, rotation_matrix_expected, atol=1e-4)
-
-    # Verify imap_dps.bc has been created.
-    assert (tmp_path / "imap_dps.bc").exists()
+    np.testing.assert_allclose(
+        spiceypy.q2m(segment_data["quaternion"][0]), rotation_matrix_expected, atol=1e-4
+    )
```
greglucas: I realize this was here before, but should we switch it to checking for decimal places since that seems to be the intention above with only 4 values?
https://numpy.org/doc/2.2/reference/generated/numpy.testing.assert_almost_equal.html

imap_processing/spice/pointing_frame.py
```diff
@@ -4,17 +4,29 @@
...
 
+from imap_processing.spice.geometry import SpiceFrame
 from imap_processing.spice.kernels import ensure_spice
-from imap_processing.spice.time import met_to_sclkticks, sct_to_et
+from imap_processing.spice.repoint import get_repoint_data
+from imap_processing.spice.time import et_to_utc, met_to_sclkticks, sct_to_et
 
 logger = logging.getLogger(__name__)
 
+POINTING_SEGMENT_DTYPE = np.dtype(
```
nkerman, line 26: Since we also refer to segments as parts of each spin, it may be worth a comment explaining that this file will be using the term to refer to SPICE segments (I think? 😁)

imap_processing/spice/pointing_frame.py
tmplummer: Looking at the documentation, this value is supposed to be:

> the number of seconds per encoded spacecraft clock tick for each interval.

I updated it to be `TICK_DURATION` from `spice.time.py`

imap_processing/spice/pointing_frame.py
tmplummer: This is great! Thanks for the suggestion.

greglucas (approved):

nkerman (approved):
Looks like I just missed my chance to comment on this 😁 but at least this PR got me to do a (somewhat) deep dive into understanding quaternions!

PR #1728, opened 2025-05-13
Ancillary file combiner

maxinelasp:
# Change Summary
Tool to combine ancillary files together. This was developed against requirements for most of MAG's calibration files but can easily be adapted to other instruments if they have different requirements, or extended. 

The tool takes in:
- N files with the same internal structure and data vars
  - These files don't have time varying data in them. Data can be multi-dimensional

And outputs:
- One dataset which creates a time span of days across the full range of dates covered in the files
- It fills in the data var at every day with the data from the input file with the highest version

So, you can pass in a ProcessingInput object and recieve a dataset, where you can then request the valid data for any given day. It should work with any kind of input data, although it always assigns anonymous dimensions. The version number of the input file is assigned to each date so you can track backwards.  

## Testing

Testing uses the MAG calibration tests, which are exactly how we expect to receive them.


tools/ancillary/ancillary_dataset_combiner.py
```diff
@@ -0,0 +1,243 @@
...
+
+    To change the behavior of the class, the methods "convert_file_to_dataset" and
+    "get_combined_dataset" can be overridden. If using a file format that is different
+    than CDF, override "convert_file_to_dataset" to convert the file to an xarray
+    dataset. If the structure of the input file or output file needs to be changed, or
+    they need to be combined in a different way, "get_combined_dataset" should be
+    overridden.
+
+    Some ancillary files can extend indefinitely. Therefore, the class requires the user
+    to provide an end time, which will
```
lacoak21: continue sentence.

tools/ancillary/ancillary_dataset_combiner.py
```diff
@@ -0,0 +1,243 @@
...
+        if isinstance(expected_end_date, str):
+            expected_end_date = np.datetime64(
+                f"{expected_end_date[:4]}-"
+                f"{expected_end_date[4:6]}-"
+                f"{expected_end_date[6:]}"
+            )
+
+        self.expected_end_date = expected_end_date
+
+        # TODO NEXT STEP: WRITE SOME TESTS
```
lacoak21: remove todo?

tools/ancillary/ancillary_dataset_combiner.py
```diff
@@ -0,0 +1,243 @@
...
+        if filepath.end_date is not None:
+            # Convert end_date to np.datetime64
+            formatted_str = (
+                f"{filepath.end_date[:4]}-"
+                f"{filepath.end_date[4:6]}-"
+                f"{filepath.end_date[6:]}"
+            )
+            end_dt = np.datetime64(formatted_str, "D")
+        else:
+            end_dt = self.expected_end_date
```
lacoak21, line 115: ah so that is why end_date is a required parameter. If the file doesn't have it the input. Could you use datetime.today() as default?

lacoak21: oops I see you already added an explanation for that I just missed it somehow.

tools/tests/ancillary/test_ancillary_dataset_combiner.py
```diff
@@ -0,0 +1,205 @@
...
+    with mock.patch(
+        "tools.ancillary.ancillary_dataset_combiner.cdf_to_xarray"
+    ) as read_cdf:
+        mocks = {
+            "read_cdf": read_cdf,
+        }
+        yield mocks
+
+
+@pytest.fixture
```
lacoak21, line 28: Maybe add a test for using the latest version. (Maybe you already have that and I missed it)

lacoak21 (approved):
Looks good Maxine - I just had a couple of questions

tools/ancillary/ancillary_dataset_combiner.py
maxinelasp: Haha, I got halfway through the doc and realized I was missing a feature. Thanks for catching this!

tools/ancillary/ancillary_dataset_combiner.py
maxinelasp: Yes, I think it would be fine for the end user of this class to decide "datetime.today()" is a fine input, but I want to make sure the user does include an end_date explicitly and intentionally.

tools/tests/ancillary/test_ancillary_dataset_combiner.py
maxinelasp: This is tested in most of the tests, as `test_ancillary_input` fixture has a file that isn't used at all because it's overwritten by higher versions (is that what you mean?)

lacoak21: Ah got it yes I see now.

tools/ancillary/ancillary_dataset_combiner.py
lacoak21: Haha it gave me a laugh because I totally knew what happened.

PR #1731, opened 2025-05-14
Ultra l1c SIT4 Prep

laspsandoval:
# Change Summary

## Overview
Setting up for creating l1c pset cdfs in the spacecraft frame.

## Updated Files
- spacecraft_pset.py
   - Small change in data structure so that cdf creation wouldn't throw an error

## Testing
- conftest.py
   - Added external file with l1a 1/2 data that also contains the epoch and pointing number columns added.
- test_spacecraft_pset.py
   - Created place where l1c spacecraft psets will be tested once we have the histogram data in healpix form.
- test_ultra_l1a.py
   - Added pointing numbers to attributes.
- test_ultra_l1b.py
   - Added pointing numbers to attributes.
- test_ultra_l1c.py
   - Checked that cdf was able to be created.

## Notes
A dataframe was created from Ultra test data in the following branch:
ultra_l1c_put_cdf_in_s3

This resulted in a .csv with combined pointings, epoch calculated, and pointing numbers assigned.

### DIFF
--- a/imap_processing/tests/ultra/unit/test_spacecraft_pset.py
+++ b/imap_processing/tests/ultra/unit/test_spacecraft_pset.py
@@ -88,48 +88,48 @@ def test_calculate_spacecraft_pset_with_cdf():
 
     df = pd.read_csv(TEST_PATH / "IMAP-Ultra45_r1_L1_V0.csv")
 
-    # Select a single pointing number
-    pointing = 0
-    df_subset = df[df["pointing_number"] == pointing].copy()
+    # Loop over all unique pointing numbers
+    for pointing in df["pointing_number"].unique():
+        df_subset = df[df["pointing_number"] == pointing].copy()
 
-    de_dict = {}
+        de_dict = {}
 
-    de_dict["epoch"] = df_subset["epoch"].values
-    species_bin = np.full(len(df_subset), 1, dtype=np.uint8)
+        de_dict["epoch"] = df_subset["epoch"].values
+        species_bin = np.full(len(df_subset), 1, dtype=np.uint8)
 
-    start_type = np.where(df_subset["PosYSlit"].values, 2, 1)
-    d, yf = get_front_y_position(start_type, df_subset["StopY"].values)
-    v, vhat, r = get_de_velocity(
-        (df_subset["StartX"].values, yf),
-        (df_subset["StopX"].values, df_subset["StopY"].values),
-        d,
-        df_subset["TOF"].values,
-    )
-    de_dict["direct_event_velocity"] = v.astype(np.float32)
+        start_type = np.where(df_subset["PosYSlit"].values, 2, 1)
+        d, yf = get_front_y_position(start_type, df_subset["StopY"].values)
+        v, vhat, r = get_de_velocity(
+            (df_subset["StartX"].values, yf),
+            (df_subset["StopX"].values, df_subset["StopY"].values),
+            d,
+            df_subset["TOF"].values,
+        )
+        de_dict["direct_event_velocity"] = v.astype(np.float32)
 
-    ultra_frame = SpiceFrame.IMAP_ULTRA_45
-    _, sc_dps_velocity, _ = get_annotated_particle_velocity(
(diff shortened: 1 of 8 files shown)


greglucas:
Minor nit, but could you delete the unused template items when you are opening the PRs. Otherwise it is hard to parse what text you add vs what is boilerplate. (I often delete a large chunk of it if it isn't directly relevant or necessary, so don't feel like you have to address all of those sections IMO)

laspsandoval:
> Minor nit, but could you delete the unused template items when you are opening the PRs. Otherwise it is hard to parse what text you add vs what is boilerplate. (I often delete a large chunk of it if it isn't directly relevant or necessary, so don't feel like you have to address all of those sections IMO)

yeah. It was in draft status I think when you looked at it.

imap_processing/ultra/l1c/spacecraft_pset.py
```diff
@@ -69,7 +69,7 @@ def calculate_spacecraft_pset(
     exposure_pointing = get_spacecraft_exposure_times(df_exposure)
 
     # For ISTP, epoch should be the center of the time bin.
-    pset_dict["epoch"] = de_dataset.epoch.data[0].astype(np.int64)
+    pset_dict["epoch"] = de_dataset.epoch.data[:1].astype(np.int64)
```
nkerman, line 72: I think you had it closer before - the PSET epochs are supposed to be the start of the pointing, right?

nkerman: Ideally this should be the start of the first included spin.

imap_processing/tests/ultra/unit/test_ultra_l1c.py
```diff
@@ -102,3 +117,67 @@ def test_ultra_l1c_error(mock_data_l1b_dict):
         ValueError, match="Data dictionary does not contain the expected keys."
     ):
         ultra_l1c(mock_data_l1b_dict)
+
+
+@pytest.mark.xfail(reason="IMAP-Ultra45_r1_L1_V0.csv too large to download.")
```
nkerman: Is it easy to add a test with a simple dataset you build within the test?

imap_processing/ultra/l1c/spacecraft_pset.py
laspsandoval: It's the same thing, but just formatted differently:

de_dataset.epoch.data
array([797949074578373376, 797949083468373376, 797949088588373376, ...,
       798031864898378880, 798031865538378880, 798031868218378880],
      shape=(20159,))
de_dataset.epoch.data[:1].astype(np.int64)
array([797949074578373376])
de_dataset.epoch.data[0].astype(np.int64)
np.int64(797949074578373376)

imap_processing/tests/ultra/unit/test_ultra_l1c.py
laspsandoval: I just took a sample of the .csv and created a test that way. Thanks for the idea!

imap_processing/ultra/l1c/spacecraft_pset.py
nkerman: Ah sorry of course they're the same, and they'll both be the epoch of the first recorded event, right?

I think it still should be the start of the first included spin, but if there will be plenty of counts per spin (I think true for Ultra), this difference should be very small.

imap_processing/tests/ultra/unit/test_spacecraft_pset.py
```diff
@@ -75,3 +77,63 @@ def test_pset():
     assert "healpix" in spacecraft_pset.coords
```
nkerman: Is this PR a good place to switch the coordinate to "pixel_index"?

nkerman (approved):
This all seems good to me!

imap_processing/tests/ultra/unit/test_spacecraft_pset.py
laspsandoval: Yes! Made the change.

PR #1732, opened 2025-05-14
1730 ena skymap inheritance improvements

### LINKED ISSUE #1730: ENA SkyMap inheritance improvements
### Algorithm Description:

Similar to the changes made to the PointingSet and it's sub-classes, do some refactoring of the SkyMap class to be more OO.

tmplummer:
# Change Summary

## Overview
Some cleanup and move toward more OO code for the SkyMap classes.

## Updated Files
   - Move shared __init__ code into base class
   - Define required class variables in class definitions
   - Make `to_dataset()` a required `@abstractmethod` needed in each subclass of AbstractSkyMap
   - Make `binning_grid_shape` an abstract property
   - Make `num_points` a property instead of a variable
   - Make `to_dict()` an abstract method to move class specific implementation into subclasses
   - Make `tiling_type` a class variable
   - Make `RectangularSkyMap.spacing_deg` a property

Closes: #1730

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -681,8 +681,8 @@ class AbstractSkyMap(ABC):
         """
         raise NotImplementedError("AbstractSkyMap.to_dataset() not implemented.")
 
-    @abstractmethod
     @property
+    @abstractmethod
     def binning_grid_shape(self) -> tuple[int]:
         """
         Shape of the binning grid.
@@ -923,6 +923,7 @@ class AbstractSkyMap(ABC):
         skymap.values_to_pull_project = properties.get("values_to_pull_project", [])
         return skymap
 
+    @abstractmethod
     def to_dict(self) -> dict:
         """
         Convert the SkyMap object to a dictionary of properties.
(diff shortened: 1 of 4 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -1076,6 +1051,101 @@ def __init__(
...
+        -------
+        dict
+            Dictionary containing the map properties.
+        """
+        map_properties_dict = {
+            "sky_tiling_type": "RECTANGULAR",
+            "spice_reference_frame": self.spice_reference_frame.name,
+            "spacing_deg": self.spacing_deg,
+            "values_to_push_project": (
+                self.values_to_push_project if self.values_to_push_project else []
```
greglucas: These are initialized to empty lists now, so I think you can simplify this to `"values_to_push_project": self.values_to_push_project`?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -1155,6 +1226,43 @@ def __init__(
...
 
+    @property
+    def binning_grid_shape(self) -> tuple[int]:
+        """
+        Binning_grid_shape for consistency with RectangularSkyMap.
+
+        Returns
+        -------
+        binning_grid_shape : tuple[int]
+            Shape of the AzElSkyGrid (num_az_bins, num_el_bins).
```
greglucas: I think this is under the Healpix version now and needs the docstring updated.
```suggestion
            Shape of the HealpixSkyGrid (num_points,).
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -1076,6 +1051,101 @@ def __init__(
...
+                    for coord in self.data_1d[key].coords
+                    if coord != CoordNames.GENERIC_PIXEL.value
+                }
+            )
+        return xr.Dataset(
+            rewrapped_data,
+            coords={**self.non_spatial_coords, **self.spatial_coords},
+        )
+
+    def to_dict(self) -> dict:
```
greglucas: Should this be "to_properties_dict()" or something to indicate that this isn't round-trippable. I generally think of the `to_X()` as having a corresponding `from_X()` or outputting the data as well like in the `to_dataset()` method above.

greglucas (approved):
Looks good to me to follow more of this inheritance approach. Just minor comments from me.

imap_processing/ena_maps/ena_maps.py
nkerman: I think that's a good idea too. Now that it's not under the `AbstractSkyMap` class, I agree that it would be good to make it clear that it's a dict of the map properties, rather than a dict formatted map with all the data already present.

nkerman (commented):
Looks good to me as well! Thanks for cleaning up the inheritance structures.

nkerman (approved):

PR #1734, opened 2025-05-14
Ultra l1b - cdf formatting

laspsandoval:
# Change Summary

## Overview
Formatting changes for extendedspin data products so that epoch has a coordinate of spin and will thus be filtered when the spins are filtered.

## Updated Files
- ultra_utils.py
   - Formatting changes for extendedspin data products so that epoch has a coordinate of spin and will thus be filtered when the spins are filtered.
- badtimes.py
   - Formatting changes for extendedspin data products so that epoch has a coordinate of spin and will thus be filtered when the spins are filtered.
- cullingmask.py
   - Formatting changes for extendedspin data products so that epoch has a coordinate of spin and will thus be filtered when the spins are filtered.


nkerman (approved):
LGTM

PR #1737, opened 2025-05-15
CoDICE L1a hi-omni validation

bourque:
This PR makes changes to the CoDICE L1a `hi-omni` implementation and completes the validation for the data product.

In doing so, I created a `create_binned_dataset` function (and some other supporting methods/code changes) to handle how the `hi-omni` data product defines its CDF variables via energy bins. I think this will also eventually apply to the `hi-sectored` data product in a future PR, though I need to confirm with Joey some subtleties about how `hi-sectored` data are shaped (I added some TODOs for this).

Pertains to #251


imap_processing/codice/constants.py
```diff
@@ -104,7 +103,7 @@
 
 # CDF variable names used for hi data products
 HI_COUNTERS_SINGLES_VARIABLE_NAMES = ["tcr", "ssdo", "stssd"]
-HI_OMNI_VARIABLE_NAMES = ["h", "he3", "he4", "c", "o", "ne_mg_si", "fe", "uh"]
+HI_OMNI_VARIABLE_NAMES = ["h", "he3", "he4", "c", "o", "ne_mg_si", "fe", "uh", "junk"]
```
bourque, line 106: Yes, the CoDICE instrument team literally expects a variable named "junk"

imap_processing/codice/constants.py
```diff
@@ -210,6 +209,158 @@
     if is_active
 ]
 
+# Energy tables for CoDICE-Hi data products. These values represent the edges
```
bourque, line 212: Just moving this up higher in the file so that it can be referenced in that data product configuration dictionary

imap_processing/codice/codice_l1a.py
```diff
@@ -558,6 +569,140 @@ def set_data_product_config(self, apid: int, dataset: xr.Dataset) -> None:
...
+    # Run some of the pipeline methods to set configs and decompress
+    # the data
+    pipeline = CoDICEL1aPipeline(table_id, plan_id, plan_step, view_id)
+    pipeline.set_data_product_config(apid, dataset)
+    pipeline.decompress_data(science_values)
+
+    # hi-omni data gets reshaped a bit differently than other products,
+    # so we need to stray away from the nominal pipeline
+    stacked_data = np.stack(
+        [np.array(item, dtype=np.uint32) for item in pipeline.__dict__["raw_data"]]
```
greglucas: Why are you going into the `__dict__` here? Can you access the attribute `pipeline.raw_data` instead?

imap_processing/codice/codice_l1a.py
```diff
@@ -558,6 +569,140 @@ def set_data_product_config(self, apid: int, dataset: xr.Dataset) -> None:
...
+                # We only need one set of support variables in the CDF,
+                # so just iterate using one species for these
+                if species == "h":
+                    # For each spin, we add <spin_period>*<num_spins> to the epoch value
+                    spin_period = (
+                        dataset.spin_period.data[i] * constants.SPIN_PERIOD_CONVERSION
+                    )
+                    epoch_value = current_epoch + np.int64(
+                        (spin_period * num_spins) * 1e9
+                    )
```
greglucas: Could you add a quick comment about spin_period being in seconds and you converting this to ns (if I understand what is going on here).

imap_processing/codice/codice_l1a.py
```diff
@@ -558,6 +569,140 @@ def set_data_product_config(self, apid: int, dataset: xr.Dataset) -> None:
         self.cdf_attrs.add_instrument_variable_attrs("codice", "l1a")
 
 
+def create_binned_dataset(apid: int, dataset: xr.Dataset) -> xr.Dataset:
+    """
+    Create dataset for data that is binned by energy.
```
greglucas, line 576: Could you add a comment here about what the general shape/contents you expect are. Reading through the code I'm still not entirely clear what will be in the Dataset in terms of shapes. Something like:

```
Dataset:
  - epoch (nepochs)
  - spin_period(epoch)
  - h(epoch, energy)
  - ...
  - other shapes?
```

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -244,7 +266,7 @@ def test_l1a_validate_data_arrays(test_l1a_data: xr.Dataset, index):
     able_to_be_validated = [
         "hi-counters-aggregated",
         "hi-counters-singles",
-        "hi-priority",
+        "hi-omni",
```
greglucas: Do you want to add `hi-omni` in addition to `hi-priority`? Or is this a rename?

greglucas (approved):
I think this looks good, I didn't review everything here but the overall addition of the extra function looks like a nice abstraction. Most of my comments are about comments ;)

imap_processing/codice/codice_l1a.py
bourque: Good catch. I think this is leftover from some debugging I was doing.

imap_processing/tests/codice/test_codice_l1a.py
bourque: Oops, thanks for catching this!

PR #1738, opened 2025-05-15
BUG: setting Data_version in write_cdf

tech3371:
# Change Summary

## Overview
small change to fix Data_version=None throwing error on loading CDF file.

## Updated Files
   - small fix

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -116,6 +116,7 @@ def write_cdf(
             stacklevel=2,
         )
         version = "v999"
+        dataset.attrs["Data_version"] = version
     elif not re.match(r"v\d{3}", version):
         raise ValueError(
             f"The Data_version attribute {version} does not match expected format vXXX."
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -115,11 +115,11 @@ def write_cdf(
             "No Data_version attribute found in dataset. Using default v999.",
             stacklevel=2,
         )
-        version = "v999"
+        version = "999"
         dataset.attrs["Data_version"] = version
-    elif not re.match(r"v\d{3}", version):
+    elif not re.match(r"\d{3}", version):
         raise ValueError(
-            f"The Data_version attribute {version} does not match expected format vXXX."
+            f"The Data_version attribute {version} does not match expected format XXX."
         )
 
     repointing = dataset.attrs.get("Repointing", None)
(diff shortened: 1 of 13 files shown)


maxinelasp (commented):
Good catch! Would you also be able to update the CLI and this code to set the version to "001" instead of "v001" as requested by Andriy? Should just be able to chop it to the last thee characters in those two spots.

greglucas (approved):

greglucas:
> Would you also be able to update the CLI and this code to set the version to "001" instead of "v001" as requested by Andriy? Should just be able to chop it to the last thee characters in those two spots.

Oof :( Do we need to update this in imap-data-access too so we can reference the version everywhere as just 3 numbers and only our filenames would print out `v` as a prefix in that field?

tech3371:
> > Would you also be able to update the CLI and this code to set the version to "001" instead of "v001" as requested by Andriy? Should just be able to chop it to the last thee characters in those two spots.
> 
> Oof :( Do we need to update this in imap-data-access too so we can reference the version everywhere as just 3 numbers and only our filenames would print out `v` as a prefix in that field?

@greglucas Can you help me with this work?

maxinelasp:
@greglucas AFAIK it is just in the attribute within the file, but Veronica brought it up. No way should we change the filename again - they had their chance to say no!

maxinelasp:
I think imap-data-access and everything else uses the filename, so it shouldn't affect anything except (maybe?) the reading in and writing out of files

greglucas:
> No way should we change the filename again - they had their chance to say no!

💯 I completely agree :)

I think in all of our `ScienceFile.version`, `Database.version`, etc... we are using the "vXXX" format as a string with the v prefix. Is it OK if we have a `file.global_attribute["Data_version"] == "001"` but the filename would be access as a version with "v001"? I don't think we are really using the Data_version global attribute at all ourselves, just putting that into the file, so perhaps we just go with the minimal update @maxinelasp suggests and don't worry about our internal versioning because that is consistent on our end already?

maxinelasp:
@greglucas  yes that was exactly what I was thinking, just change the attribute. It wasn't even added into files consistently until recently so I don't think it's used anywhere.

greglucas:
@tech3371, I just pushed a commit that updates the `Data_version` attribute for everyone automatically and updates all of the corresponding tests that expect that value. Filenames still have "vXXX", but the attribute only has "XXX" now.

tech3371:
> @tech3371, I just pushed a commit that updates the `Data_version` attribute for everyone automatically and updates all of the corresponding tests that expect that value. Filenames still have "vXXX", but the attribute only has "XXX" now.

Thank you @greglucas. Changes looks good to me. Would you like me to merge it?

greglucas:
> Thank you @greglucas. Changes looks good to me. Would you like me to merge it?

Yes, if you are OK with those changes then please merge.

PR #1739, opened 2025-05-15
Add spice time conversions

maxinelasp:
# Change Summary
ET is a convenient transition point when converting between time formats, but it doesn't have some of the ET -> other format conversions that I wanted. This meant I had to go from J2000->ET->UTC string -> np.datetime64 which is annoying. There was also no way to go from ET to J2000ns, so for example when using the start_time input from the CLI, I couldn't go from a string -> ET -> J2000ns. 

This change adds both of those functionalities. The to datetime64 is mostly for convenience, with the other one bridging a gap in functionality.

## Overview

- Added et_to_ttj2000ns()
- Added et_to_datetime64()

## Testing

For testing, I basically duplicated the tests for ttj2000ns_to_et and added in a round-trip conversion using both methods. For et_to_datetime64 I kept it very simple since it's just a wrapper around et_to_utc anyway.

Other notes: I found the function added by @lacoak21 , `epoch_to_doy()` threw errors when I called it with J2000 ns... not sure if I was using it incorrectly, just wanted to note that.

### DIFF
--- a/imap_processing/spice/time.py
+++ b/imap_processing/spice/time.py
@@ -131,6 +131,33 @@ def ttj2000ns_to_et(tt_ns: npt.ArrayLike) -> npt.NDArray[float]:
 
 
 @typing.no_type_check
+@ensure_spice
+def et_to_ttj2000ns(et: npt.ArrayLike) -> npt.NDArray[float]:
+    """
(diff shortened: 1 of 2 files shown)


tmplummer (approved):
Thanks for adding this, Maxine!

PR #1741, opened 2025-05-15
Allow for JSON dependency in parse_args

### LINKED ISSUE #1740: CLI:  allow for JSON `--dependency` instead of only strings

nkerman:
# Change Summary

Small change to the CLI to allow for JSON `--dependency` instead of only strings. 

Closes #1740 

## Overview

If `--dependency` ends with ".json", we can assume it's a json filepath. 

## Updated Files

- cli.py
   - check for and handle JSON filepath dependency inputs

## Testing

- Test `cli._parse_args()`


lacoak21 (approved):
Look great Nat! Thanks for doing this. Super helpful

tech3371:
Sorry Nat for being late on this review. I know I asked you to add this feature for us but I was late to review it. I will review this right now.

imap_processing/cli.py
```diff
@@ -243,6 +246,14 @@ def _parse_args() -> argparse.Namespace:
     )
     args = parser.parse_args()
 
+    # If the dependency argument was passed in as a json file, read it into a string
+    if args.dependency.endswith(".json"):
+        logger.info(
+            f"Interpreting dependency argument as a JSON file: {args.dependency}"
+        )
+        with open(args.dependency) as f:
+            args.dependency = f.read()
```
tech3371, line 259: This looks great. I know we didn't have support for this at the time of our discussion but we added some supported for those and I am wondering if you can go one step further. Right now, we don't know a way of knowing how to download and then read the json file in batch job or local. Can you do these?
1. update `imap-data-access` to `v0.27.0`. This should now support [cadence file validator](https://github.com/tech3371/imap-data-access/blob/7223e726401ac560a6477c8cd06407231281477b/imap_data_access/file_validation.py#L1009) we recently added.
2. Then call [download()](https://github.com/tech3371/imap-data-access/blob/7223e726401ac560a6477c8cd06407231281477b/imap_data_access/io.py#L42) which will take care of rest of things(such as download path or where to download from) for you. Then read downloaded file like you have currently. This will help line `imap_cli` with how we want it to behave in local testing and in batch job testing.

imap_processing/tests/test_cli.py
```diff
@@ -93,6 +95,54 @@ def test_main(mock_instrument):
...
+        },
+        {
+            "type": "science",
+            "files": [
+                "imap_idex_l2_sci_20240312_v000.cdf",
+                "imap_idex_l2_sci_20240312_v001.cdf",
+            ],
+        },
+    ]
+    test_json_dst = "test_cli_dependency_json.json"
```
tech3371: Once you add above changes, you will need to make sure this line up with filename convention as science but with json extension now. Please don't hesitate to ask me and happy to walk through that.

imap_processing/tests/test_cli.py
```diff
@@ -93,6 +95,54 @@ def test_main(mock_instrument):
...
+    with open(test_json_dst, "w") as f:
+        f.write(json.dumps(test_json_content))
+
+    test_args = [
+        "imap_cli",
+        "--instrument",
+        "mag",
+        "--dependency",
+        str(test_json_dst),
```
tech3371, line 132: This looks great! Thank you for doing that. I takes one load off from @lacoak21 and I. Really appreciate the help!

nkerman: Of course! Thank you for all the infrastructure you've built!

imap_processing/tests/test_cli.py
nkerman: I think I've done this now! At some point, we should change all the paths from "candence" to "cadence", but that's a minor bugfix.

imap_processing/cli.py
nkerman: Done! I'm not sure if the tests will pass on github (is imap-data-access required to be v0.27.0 or does it still need to be manually installed until the version update?) I'll also increment the version in pyproject.toml.

imap_processing/tests/test_cli.py
tech3371: Oh no. did I still mess up the spelling?

nkerman: Yeah, but I don't think it's a huge issue to prioritize at the moment/before SIT4. Just need to change the path from "candence" to "cadence".

imap_processing/cli.py
tech3371: yep. only `v0.27.0` contains support for cadence. What you did in this PR looks great! Test did pass and should pass once you update version.

tech3371 (approved):
Thank you for doing this. It looks great!

imap_processing/tests/test_cli.py
tech3371: Let me fix that. Thank you for spotting that! but it won't block your work here. We can open new PR with new version of this fix.

PR #1742, opened 2025-05-15
I-ALiRT - Create a way to mimic attitude kernels

laspsandoval:
# Change Summary

## Overview
This creates a way to mimic attitude kernel so we can transform the MAG data into different reference frames.

## Updated Files
- imap_ialirt_sim_metakernel.template
   - Added metakernel that will be used for I-ALiRT
- ialirt_spice.py
   - Creates matrices for each transformation
   - Multiplies matrices to get into inertial reference frame

## Testing
- test_ialirt_spice.py
   - Tests against what is produced using SPICE attitude kernels

### DIFF
--- a/imap_processing/ialirt/l0/ialirt_spice.py
+++ b/imap_processing/ialirt/l0/ialirt_spice.py
@@ -79,7 +79,9 @@ def transform_instrument_vectors_to_inertial(
     spin_phase: NDArray,
     sc_inertial_right: NDArray,
     sc_inertial_decline: NDArray,
-    et: np.ndarray,
+    et: NDArray,
+    instrument_frame: SpiceFrame = SpiceFrame.IMAP_MAG,
+    spacecraft_frame: SpiceFrame = SpiceFrame.IMAP_SPACECRAFT,
 ) -> NDArray:
     """
     Transform instrument-frame vectors into the inertial frame (ECLIPJ2000).
@@ -96,6 +98,10 @@ def transform_instrument_vectors_to_inertial(
         Spacecraft declination in radians. Shape: (N,).
     et : np.ndarray
         Ephemeris time. Shape: (N,).
+    instrument_frame : SpiceFrame
+        Instrument frame.
+    spacecraft_frame : SpiceFrame
+        Spacecraft frame.
 
     Returns
     -------
@@ -113,20 +119,22 @@ def transform_instrument_vectors_to_inertial(
     vectors_urf = get_instrument_vector(
         et,
         instrument_vectors,
+        instrument_frame,
+        spacecraft_frame,
     )
 
     vectors_inertial = np.array(
-        [spice.mxv(r.T, v) for r, v in zip(rot_matrices, vectors_urf)]
+        [spice.mxv(r.T.copy(), v) for r, v in zip(rot_matrices, vectors_urf)]
     )
 
     return vectors_inertial
 
 
 def get_instrument_vector(
-    et: np.ndarray,
-    vector: np.ndarray,
-    instrument_frame: SpiceFrame = SpiceFrame.IMAP_MAG,
-    spacecraft_frame: SpiceFrame = SpiceFrame.IMAP_SPACECRAFT,
+    et: NDArray,
+    vector: NDArray,
+    instrument_frame: SpiceFrame,
+    spacecraft_frame: SpiceFrame,
 ) -> np.ndarray:
     """
     Get the vectors wrt the spacecraft.
@@ -149,10 +157,10 @@ def get_instrument_vector(
     """
     # Instrument frame → SC frame (URF)
     vector_urf = frame_transform(
-        et=et,
-        position=vector,
-        from_frame=instrument_frame,
-        to_frame=spacecraft_frame,
+        et,
+        vector,
+        instrument_frame,
+        spacecraft_frame,
     )
 
     return vector_urf
--- a/imap_processing/ialirt/l0/ialirt_spice.py
+++ b/imap_processing/ialirt/l0/ialirt_spice.py
@@ -101,9 +101,6 @@ def compute_sc_to_inertial_rotation_matrix_from_z(
         # Columns = spacecraft axes in inertial frame:
         # SPICE: Zsc → X, Ysc → Y, Xsc → Z
         R = np.stack([x_rot, y_rot, z], axis=1)
-        # Orthonormalize
-        u, _, vh = np.linalg.svd(R)
-        R = u @ vh
         R_all.append(R)
 
     return np.stack(R_all)
@@ -133,7 +130,7 @@ def transform_instrument_vectors_to_inertial(
     R_mount = spice.pxform("IMAP_MAG", "IMAP_SPACECRAFT", 0.0)
 
     # SC → inertial
-    R_sc_to_inertial = spice.pxform("IMAP_SPACECRAFT", "ECLIPJ2000", et[0])
+    #R_sc_to_inertial = spice.pxform("IMAP_SPACECRAFT", "ECLIPJ2000", et[0])
 
     R_sc_to_inertial_test = compute_sc_to_inertial_rotation_matrix_from_z(
         z_axis,
--- a/imap_processing/ialirt/l0/ialirt_spice.py
+++ b/imap_processing/ialirt/l0/ialirt_spice.py
@@ -4,7 +4,11 @@ import numpy as np
 import spiceypy as spice
 from numpy.typing import NDArray
 
-from imap_processing.spice.geometry import spherical_to_cartesian
+from imap_processing.spice.geometry import (
+    SpiceFrame,
+    frame_transform,
+    spherical_to_cartesian,
+)
 
 
 def get_z_axis(sc_inertial_right: NDArray, sc_inertial_decline: NDArray) -> NDArray:
@@ -108,3 +112,39 @@ def transform_instrument_vectors_to_urf(
     )
 
     return vectors_urf
+
+
+def get_instrument_vector(
+    et: np.ndarray,
+    vector: np.ndarray,
+    instrument_frame: SpiceFrame = SpiceFrame.IMAP_MAG,
+    spacecraft_frame: SpiceFrame = SpiceFrame.IMAP_SPACECRAFT,
+) -> np.ndarray:
+    """
+    Get the vectors wrt the spacecraft.
+
+    Parameters
+    ----------
+    et : np.ndarray
+        Ephemeris time.
+    vector : np.ndarray
+        Vector in the instrument frame.
+    instrument_frame : SpiceFrame
+        Instrument frame.
+    spacecraft_frame : SpiceFrame
+        Spacecraft frame.
+
+    Returns
+    -------
+    vector_urf : np.ndarray
+        Transformed vector(s) in the spacecraft frame.
+    """
+    # Instrument frame → SC frame (URF)
+    vector_urf = frame_transform(
+        et=et,
+        position=vector,
+        from_frame=instrument_frame,
+        to_frame=spacecraft_frame,
+    )
+
+    return vector_urf
--- a/imap_processing/ialirt/l0/ialirt_spice.py
+++ b/imap_processing/ialirt/l0/ialirt_spice.py
@@ -74,7 +74,7 @@ def get_rotation_matrix(z_axis: NDArray, spin_phase: NDArray) -> NDArray:
     return rot_matrices
 
 
-def transform_instrument_vectors_to_urf(
+def transform_instrument_vectors_to_inertial(
     instrument_vectors: NDArray,
     spin_phase: NDArray,
     sc_inertial_right: NDArray,
@@ -82,7 +82,7 @@ def transform_instrument_vectors_to_urf(
     et: np.ndarray,
 ) -> NDArray:
     """
-    Transform instrument-frame vectors into the spacecraft URF frame.
+    Transform instrument-frame vectors into the inertial frame (ECLIPJ2000).
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


laspsandoval:
pre-commit.ci autofix

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -42,69 +45,158 @@ def get_z_axis(sc_inertial_right: NDArray, sc_inertial_decline: NDArray) -> NDAr
...
     """
-    Rotate a spacecraft frame about the spin axis by the given spin phase angle.
+    Create rotation matrices for spin about the spacecraft Z-axis.
 
     Parameters
     ----------
     z_axis : NDArray
-        Unit vector spacecraft Z-axis.
+        Unit vector of spacecraft spin axis (Z-axis).
     spin_phase : NDArray
```
tmplummer: I didn't catch this last time, but @nkerman , @greglucas , and I decided on an unofficial standard of always having angles input and output from functions in degrees. I know it isn't part of these changes, but I would suggest refactoring of this function.
Perhaps draw from the spice documentation to make this function more general.
- Description: `Construct a rotation matrix that rotates vectors by a specified angle about a specified axis.`
- Parameters:
   - axis: Rotation axis.
   - angle: Rotation angle, in degrees.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
```diff
@@ -68,28 +73,137 @@ def test_get_rotation_matrix():
...
+    z_axis = get_z_axis(np.array([ra]), np.array([dec]))[0]  # extract the single row
+
+    # Test that our get_z_axis code is returning what SPICE returns.
+    np.testing.assert_allclose(
+        z_axis,
+        sc_z_inertial,
+        atol=1e-9,
+    )
+
+    spin_phase = np.array([0.0])
```
tmplummer: It would be good to test non-zero spin-phase. Otherwise there is no rotation happening about the z-axis.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
```diff
@@ -68,28 +73,137 @@ def test_get_rotation_matrix():
...
     )
 
-    np.testing.assert_allclose(result, expected, atol=1e-8)
+    # Pick midpoint of first coverage interval
+    et_start = ck_cover[0]
+    et_end = ck_cover[1]
+    et = (et_start + et_end) / 2.0
+
+    # Assume IMAP_MAG +X is boresight
+    instrument_vector = np.array([[1.0, 0.0, 0.0]])
```
tmplummer: I think it is a better test to choose a vector with non-zero x, y, and z components. No need to make it a unit vector either.

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -42,69 +45,158 @@ def get_z_axis(sc_inertial_right: NDArray, sc_inertial_decline: NDArray) -> NDAr
...
+    # Pick a fixed reference vector.
+    v_ref = np.array([0, 0, 1])
+
+    # Detect if z_axis is nearly aligned with v_ref.
+    dot_products = np.dot(z_axis, v_ref)
+    too_parallel = np.abs(dot_products) > 0.99
+
+    # Use alternate reference vector where needed.
+    v_refs = np.tile(v_ref, (z_axis.shape[0], 1))
+    v_refs[too_parallel] = np.array([1, 0, 0])
```
tmplummer, line 88: This code still bothers me. Where did this come from? I would like to have a reason that randomly picking a z-axis vector or x-axis vector to use to define the y-axis works. I have a suspicion that the `too_parallel` condition will never happen because the spacecraft z-axis will never point close to the HAE poles (z-axis).

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
```diff
@@ -68,28 +73,137 @@ def test_get_rotation_matrix():
...
-        ]
+    # Check cross(X, Y) = Z.
+    reconstructed_z = np.cross(x_axis, y_axis)
+    assert np.allclose(reconstructed_z, z_axis, atol=1e-6)
+
+
+def test_compute_total_rotation():
+    """Test compute_total_rotation function."""
+
+    r_sc = spin = mount_matrix = [
```
tmplummer: I'm not sure that testing that 3 identity matrices multiplied together gives an identity matrix is accomplishing much.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
```diff
@@ -68,28 +73,137 @@ def test_get_rotation_matrix():
...
+    instrument_vector = np.array([1.0, 2.0, 3.0])
+    output_vector = spiceypy.mxv(total_rotations[0], instrument_vector)
+
+    np.testing.assert_allclose(output_vector, instrument_vector, atol=1e-9)
+
+
+@pytest.mark.use_test_metakernel("imap_ena_sim_metakernel.template")
+@pytest.mark.external_kernel
+@ensure_spice
+def test_transform_instrument_vectors_to_inertial(
```
tmplummer, line 155: This is a fantastic test! I have a couple of comments to maybe improve the coverage it provides.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
```diff
@@ -68,28 +73,137 @@ def test_get_rotation_matrix():
...
+        v_manual[0],
+        v_spice,
+        atol=1e-9,
+    )
+
+
+@pytest.mark.use_test_metakernel("imap_ialirt_sim_metakernel.template")
+@pytest.mark.external_kernel
+@ensure_spice
+def test_no_attitude():
```
tmplummer, line 224: So I think that the point of this test is to make sure that you aren't somehow cheating and using the s/c attitude kernel to compute the transform matrix? An alternate way to accomplish that in the above test is to just not furnish kernels while you call the `transform_instrument_vector_to_intertial()` function then manually furnish the kernels for the call `v_spice = spiceypy.mxv(rot_inst_to_inertial, instrument_vector[0])`.

imap_processing/tests/spice/test_data/imap_ialirt_sim_metakernel.template
tmplummer, line 1: Is this the minimum set of kernels needed to call `transform_instrument_vectors_to_intertial`? I would expect to maybe only need `imap_wkcp.tf`?

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -42,69 +45,158 @@ def get_z_axis(sc_inertial_right: NDArray, sc_inertial_decline: NDArray) -> NDAr
...
+        rotation_inst_to_inertial = rotation_sc @ spin @ mount_matrix
+
+        total_rotations.append(rotation_inst_to_inertial)
+
+    total_rotations = np.array(total_rotations)
+
+    return total_rotations
+
+
+def transform_instrument_vectors_to_inertial(
```
tmplummer, line 128: Can you define what inertial reference frame this is?

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -42,69 +45,158 @@ def get_z_axis(sc_inertial_right: NDArray, sc_inertial_decline: NDArray) -> NDAr
...
-        Vectors in the instrument frame. Shape: (N, 3).
-    spin_phase : np.ndarray
-        Spin phase angle(s) in radians. Shape: (N,).
-    sc_inertial_right : np.ndarray
-        Spacecraft right ascension in radians. Shape: (N,).
-    sc_inertial_decline : np.ndarray
-        Spacecraft declination in radians. Shape: (N,).
+    instrument_vectors : NDArray
+        Array of instrument-frame vectors, shape (N, 3).
+    spin_phase : NDArray
```
tmplummer, line 143: Similar to my other comment, we should only pass around variables with angle in degrees.

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -42,69 +45,158 @@ def get_z_axis(sc_inertial_right: NDArray, sc_inertial_decline: NDArray) -> NDAr
...
+    total_rotations = []
+
+    for rotation_sc, spin in zip(inertial_frames, spin_rotations):
+        # Multiply the three matrices: inertial, spin, and mount.
+        # instrument → spacecraft → rotated spacecraft → inertial
+        rotation_inst_to_inertial = rotation_sc @ spin @ mount_matrix
+
+        total_rotations.append(rotation_inst_to_inertial)
+
+    total_rotations = np.array(total_rotations)
```
tmplummer: I believe that numpy will correctly to the broadcasting of these arrays such that matrices of sizes `(N, 3, 3) @ (N, 3, 3) @ (3, 3) = (N, 3, 3)`
```suggestion
    for rotation_sc, spin in zip(inertial_frames, spin_rotations):
        # Multiply the three matrices: inertial, spin, and mount.
        # instrument → spacecraft → rotated spacecraft → inertial
        rotation_inst_to_inertial = rotation_sc @ spin @ mount_matrix

        total_rotations.append(rotation_inst_to_inertial)

    total_rotations = rotation_sc @ spin @ mount_matrix
```
From the documentation:

> The behavior depends on the arguments in the following way.
    - If both arguments are 2-D they are multiplied like conventional matrices.
    - If either argument is N-D, N > 2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.

https://numpy.org/doc/2.1/reference/generated/numpy.matmul.html

tmplummer (commented):
Really good test coverage! I have some comments to improves some details.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
laspsandoval: I'm having a difficult time with this. When I furnish sim_1yr_imap_attitude.bc and back-calculate I see only 0 degrees. And I tried:

et = 797949054.185627 + 10  

frame_id = int(spiceypy.gipool("FRAME_IMAP_SPACECRAFT", 0, 1)[0])
et_out, quat, av, found = spiceypy.ckgpav(frame_id, et, 0.0, "ECLIPJ2000")

I see spiceypy.utils.exceptions.NotFoundError: Spice returns not found for function: ckgpav

which I think indicates that we don't have angular rotation in our ck kernel.

imap_processing/ialirt/l0/ialirt_spice.py
laspsandoval: It is almost the same as _create_rotation_matrix in pointing_frame.py. That code was written in MATLAB by Nick Dutton. As I understand it, it is a standard way to create an orthonormal frame.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
laspsandoval: That is true. But then I would need to furnish them for ra, dec, and et above. If ok, I think I'd like to keep it as is. I'll want to add on to the other function later. This seems like a structure preference comment. But I do appreciate it the idea.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
laspsandoval: Figured out a way! Ok. Updated.

imap_processing/ialirt/l0/ialirt_spice.py
tmplummer: Right, but in `pointing_frame._create_rotation_matrix()`, the ECLIPJ2000 z-axis is chosen because the pointing frame y-axis is defined to be the closest vector to the Ecliptic z-axis that is normal to the Pointing z-axis. Thus, `Y=Z_pointing cross Z_ecliptic`.
This code makes me think that the x and y axes don't matter, they just need to be orthogonal. But I have a hard time reconciling that with the fact that this matrix is later used as a rotation matrix which, to me, means that the x and y axes should matter.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
```diff
@@ -68,28 +70,179 @@ def test_get_rotation_matrix():
...
+    """Test transform_instrument_vectors_to_inertial function."""
+
+    ck_path = spice_test_data_path / "sim_1yr_imap_attitude.bc"
+    id_imap_spacecraft = spiceypy.gipool("FRAME_IMAP_SPACECRAFT", 0, 1)
+
+    ck_cover = spiceypy.ckcov(
+        str(ck_path), int(id_imap_spacecraft), True, "INTERVAL", 0, "TDB"
+    )
+
+    # Pick midpoint of first coverage interval
```
tmplummer, line 167: This comment is no longer correct.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
```diff
@@ -68,28 +70,179 @@ def test_get_rotation_matrix():
...
+    )
+    v_manual_1 = transform_instrument_vectors_to_inertial(
+        instrument_vector,
+        np.array([240.0]),
+        np.array([np.degrees(ra)]),
+        np.array([np.degrees(dec)]),
+    )
+
+    rot_inst_to_inertial_0 = spiceypy.pxform("IMAP_MAG", "ECLIPJ2000", et_start + 10)
+    rot_inst_to_inertial_1 = spiceypy.pxform("IMAP_MAG", "ECLIPJ2000", et_start + 20)
```
tmplummer, line 204: Can you add a comment on why `et_start + 10` is 120 degrees and `et_start + 20` is 240 degrees?

tmplummer (approved):
Nice job Laura. Thanks for making those changes! I left just a couple of nits about code comments.

imap_processing/ialirt/l0/ialirt_spice.py
laspsandoval: This is definitely the reason I needed your review! Thank you.

I made some minor changes and I believe that with this calculation I am setting Y ≈ ecliptic north while still being perpendicular to z. I THINK that's right, but I'm still learning here. What gives me comfort is that I am matching the attitude kernel. I feel confident enough to merge, but if you think of anything different that I should do I can open another PR.

PR #1743, opened 2025-05-15
1727 hi l2   dataset metadata for cdf

### LINKED ISSUE #1727: Hi L2 - dataset metadata for CDF
### Algorithm Description:

Remove unneeded variables from the `SkyMap` object then take the dataset returned by the `to_dataset()` method and add the global attributes and variable attributes such that it is ready for `write_cdf`.

tmplummer:
# Change Summary

## Overview
This PR covers work to get data from the RectangularSkyMap CDF ready. It adds this to the Hi L2 code.

## Updated Files
   - Add global attributes for hi_l2_enamap-sf data products. These are placeholders, not finalized
   - Add "obs_date" variable to HiPointingSet for tracking obs_date when projecting
   - Add `min_epoch` and `max_epoch` as instance variables to `AbstractSkyMap` class. These are used to store the min/max epochs from any PSET that gets projected into the SkyMap.
   - Add `solid_angle` to data_1d in RectangularSkyMap __init__ method
   - Add `build_cdf_dataset` method to `RectangularSkyMap` that gets the 1D data ready to write to a CDF.
   - Test coverage for new `build_cdf_dataset()` method.
      - Add `hi_l2` function that will be called by cli.py module.
      - Add `obs_date` to values to project
      - Add code to get RectangularSkyMap data ready for writing to CDF
      - Add test coverage for updates to hi_l2 module

Closes: #1727

### DIFF
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -167,7 +167,7 @@ obs_date_energy_dependent:
 
 solid_angle:
   <<: *default_float32
-  CATDESC: Solid angle of subtended by each pixel.
+  CATDESC: Solid angle subtended by each pixel.
   FIELDNAM: Solid Angle
   DEPEND_0: epoch
   UNITS: sr
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -164,3 +164,13 @@ exposure_factor_energy_dependent:
 
 obs_date_energy_dependent:
   <<: *obs_date_energy_independent
+
+solid_angle:
+  <<: *default_float32
+  CATDESC: Solid angle of subtended by each pixel.
+  FIELDNAM: Solid Angle
+  DEPEND_0: epoch
+  UNITS: sr
+  VAR_TYPE: data
+  LABLAXIS: Solid Angle
+  DISPLAY_TYPE: image
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -134,6 +134,16 @@ ena_intensity_stat_unc:
   LABLAXIS: Statistical Unc
   DISPLAY_TYPE: image
 
+ena_intensity_sys_err:
+  <<: *default_float32
+  CATDESC: ENA intensity non-statistical error.
+  FIELDNAM: Intensity non-stat error
+  UNITS: counts/(s * cm^2 * Sr * KeV)
+  DEPEND_0: epoch
+  VAR_TYPE: data
+  LABLAXIS: Non-Statistical Err
+  DISPLAY_TYPE: image
+
 sensitivity:
   <<: *default_float32
   CATDESC: Calibration/sensitivity factor calculated from multiple pointing sets.
@@ -159,6 +169,17 @@ obs_date: &obs_date_energy_independent
   datatype: int64
   CATDESC: Exposure time weighted mean collection date of data in a pixel.
   FIELDNAM: J2000 Nanoseconds
+  UNITS: ns
+  DEPEND_0: epoch
+  VAR_TYPE: data
+  LABLAXIS: epoch
+  DISPLAY_TYPE: image
+
+obs_date_range:
+  <<: *default_int64
+  datatype: int64
+  CATDESC: Standard deviation of the observation date.
+  FIELDNAM: Observation date range
   UNITS: ns
   DEPEND_0: epoch
   VAR_TYPE: data
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -33,6 +33,15 @@ default_float32_attrs: &default_float32
   VALIDMAX: 3.4028235e+38
   dtype: float32
 
+# Define epoch_delta attributes
+epoch_delta:
+  <<: *default_int64
+  CATDESC: Number of nanoseconds covered by data included in this map product.
+  DISPLAY_TYPE: no_plot
+  VALIDMIN: 0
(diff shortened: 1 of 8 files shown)


imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -164,3 +164,13 @@ exposure_factor_energy_dependent:
 
 obs_date_energy_dependent:
   <<: *obs_date_energy_independent
+
+solid_angle:
+  <<: *default_float32
+  CATDESC: Solid angle of subtended by each pixel.
```
nkerman:
```suggestion
  CATDESC: Solid angle subtended by each pixel.
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -664,6 +673,11 @@ def __init__(self, spice_frame: geometry.SpiceFrame) -> None:
         self.values_to_push_project = []
         self.values_to_pull_project = []
 
+        # Initialize min and max epoch variables that are used to keep track of
+        # PSET epochs that get binned to the map
+        self.min_epoch = np.iinfo(np.int64).max
+        self.max_epoch = np.iinfo(np.int64).min
```
nkerman, line 679: Took me a second to understand this with the min/max-ing later on in the code, but I think it's clever!

imap_processing/ena_maps/ena_maps.py
```diff
@@ -819,6 +833,9 @@ def project_pset_values_to_map(
             # that correspond to each map pixel as the weights.
             self.data_1d[value_key] += pointing_projected_values
 
+        self.min_epoch = min(self.min_epoch, pointing_set.epoch)
+        self.max_epoch = max(self.max_epoch, pointing_set.epoch)
```
nkerman, line 839: I think the max epoch is a bit trickier, since it should include the time duration _of the last PSET_ as well, (~1 day). A good approximation until we have that info in the PSET would be to add `(86400 * 1E9)` nanoseconds to the max epoch.

imap_processing/hi/l2/hi_l2.py
```diff
@@ -12,7 +12,56 @@
...
+    Returns
+    -------
+    l2_dataset : list[xr.Dataset]
+        Level 2 IMAP-Hi dataset ready to be written to a CDF file.
+    """
+    # TODO: parse descriptor to determine map configuration
+    sensor = "45" if "45" in descriptor else "90"
+    direction: Literal["full"] = "full"
+    cg_corrected = False
+    map_spacing = 4
```
nkerman, line 49: I have a backburnered PR updating the naming utils that will allow for parsing like this. I'll try to get it out soon!

imap_processing/hi/l2/hi_l2.py
```diff
@@ -71,32 +125,39 @@ def generate_hi_map(
...
+
+    rect_map.data_1d.update(calculate_ena_signal_rates(rect_map.data_1d))
+    rect_map.data_1d.update(
+        calculate_ena_intensity(
+            rect_map.data_1d, geometric_factors_path, esa_energies_path
+        )
+    )
+
+    # TODO: Figure out how to compute obs_date_range (stddev of obs_date)
+    rect_map.data_1d["obs_date_range"] = xr.zeros_like(rect_map.data_1d["obs_date"])
```
nkerman, line 140: I think that Jon sent us a link to a way we can do this in his email titled "iterative standard deviation". 

[StackExchange thread](https://math.stackexchange.com/questions/2148877/iterative-calculation-of-mean-and-standard-deviation/2148949#2148949)

Looks like we can project 3 values separately: `all ones, obs_date (per-pixel), obs_date**2` and then compute $\sigma_{obs\_date}$

imap_processing/ena_maps/ena_maps.py
```diff
@@ -1128,6 +1151,136 @@ def to_dataset(self) -> xr.Dataset:
             coords={**self.non_spatial_coords, **self.spatial_coords},
         )
 
+    def build_cdf_dataset(
```
nkerman, line 1156: This is a great function! I'm not sure if the Ultra L2 code will switch to using this yet, as it needs to be Healpix compatible too, but if I have time I will extend this to HealpixSkyMaps.

imap_processing/tests/hi/test_hi_l2.py
```diff
@@ -38,8 +40,23 @@ def empty_rectangular_map_dataset() -> xr.Dataset:
...
+    assert len(l2_dataset.data_vars) == 15
+    np.testing.assert_array_equal(
+        l2_dataset["ena_intensity"].dims, ["epoch", "energy", "longitude", "latitude"]
+    )
+
+
+@pytest.mark.external_test_data
+def test_genarate_hi_map(hi_l1_test_data_path):
+    """Test coverage for genarate_hi_map()"""
+    pset_path = hi_l1_test_data_path / "imap_hi_l1c_45sensor-pset_20250415_v999.cdf"
```
nkerman, line 54: Might be worth testing output values are as expected/non-zero.

nkerman (commented):
This is great and 99% ready. I want to hear your thoughts on the epoch end before hitting approve.

imap_processing/ena_maps/ena_maps.py
tmplummer: Good point. I would say that even better would be to query the repoint table to get the start epoch and end epoch. To do that we could keep track of the min/max pointing that is used. Another possibility is to add a epoch delta to our PointintSets. 

I'm going to add a TODO and write a ticket for this.

imap_processing/ena_maps/ena_maps.py
tmplummer: Yeah, I really wanted to implement base functionality in the `AbstractSkyMap.build_cdf_dataset()` function and likewise in `HealpixSkyMap` and `RectangularSkyMap` but simply don't have time right now to do it the right way :(.

imap_processing/ena_maps/ena_maps.py
tmplummer: I created this issue: #1747

tmplummer:
Hey @nkerman, can you give this another look when you get a chance.

nkerman (approved):
Thanks for pinging on this, Tim! It got a bit buried on my end. 

It looks great.

PR #1745, opened 2025-05-16
Mag l2 24hours

maxinelasp:
# Change Summary
Add the ability to use ancillary files and truncate data at MAG L2. Completes most of MAG L2, just verification and CDF attributes left.

## Overview

- Add the ancillary combiner into MAG CLI
- Update code to retrieve the calibration from the combined ancillary file (offsets are not affected and should be one file)
- add code to truncate timestamps to exactly one day
- add magnitude calculation

**Important note**: This PR depends on the [ancillary combiner PR](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1728). Don't review the ancillary combiner files here, that branch is just a parent of this one.


tools/ancillary/ancillary_dataset_combiner.py
```diff
@@ -0,0 +1,243 @@
+"""File for taking in multiple ancillary files and creating a combined dataset."""
```
maxinelasp: DON'T REVIEW THIS FILE HERE!! Ancillary combiner PR open now :)

tools/tests/ancillary/test_ancillary_dataset_combiner.py
```diff
@@ -0,0 +1,205 @@
+from pathlib import Path
```
maxinelasp: DON'T REVIEW THIS FILE HERE!! Ancillary combiner PR open now :)

imap_processing/cli.py
```diff
@@ -941,26 +944,53 @@ def do_processing(
...
+                    f"and at least one calibration file."
+                    f"All ancillary dependencies: "
+                    f"{anc_dependencies}"
+                )
+
+            # If the calibration files have no end date on them, we need to designate
+            # one. This ensures we have 3 days past the processing day in the
+            # calibration file.
+
+            if self.start_date is not None:
```
sdhoyt, line 1014: Is there a specific situation you're thinking of where the ancillary start date could be none?

sdhoyt (approved):
Looks good to me

imap_processing/cli.py
maxinelasp: This is looking at the start date from the CLI, which is not a required input. So although it is always set from batch_starter, it is possible to trigger processing without including --start-date in the CLI call. Mostly I'm thinking this might happen when running locally.

PR #1749, opened 2025-05-16
Ultra l1a - Add apids and modify the rates so that it is istp compliant

laspsandoval:
# Change Summary

## Overview
Add l1a data and update for rates apid to make it ISTP compliant.

## Updated Files
- ultra_l1a.py
   - Remove fastdata_00 since it isn’t required for the rates packet (it is already parsed in the code)
   - Add logic for commant text apid to make it a single ascii string, as requested by the IT
- ultra_utils.py
   - Remove apids that we haven’t finished yet so we don’t throw an error during processing
   - Attribute update

## Testing
- test_ultra_l1a.py
   - Added tests and data for new apids
- confest.py (both)
   - New test data for new apids


imap_processing/tests/ultra/unit/conftest.py
```diff
@@ -60,6 +60,45 @@ def ccsds_path_theta_0():
     )
 
 
+@pytest.fixture
+def ccsds_path_hk():
+    """Returns the ccsds directory."""
```
nkerman, line 65: Nit: maybe slightly flesh out the docstrings of these, or have an overarching comment above them like "# These fixtures return the paths to ... files"

nkerman (approved):
LGTM! One nit/comment request

PR #1755, opened 2025-05-19
MapDescriptor Naming Class and Descriptor Parsing in Ultra L2

### LINKED ISSUE #1752: Refactor Map Naming tools into a class
Dan R and the IMWG seem to approve of the current scheme of ENA map names, so I think we're ready to lock in the naming convention. Move the current rough function to a neater class based system and implement parsing of map descriptors.

### LINKED ISSUE #1753: Parse map descriptors and create output map structures from them.
Create a function/class/method to parse a map descriptor (e.g., `"u45-ena-h-sf-nsp-full-hae-6deg-1yr"`) according to the IMWG's naming conventions. It should be able to return the instrument, sensor (if one exists), time duration, ram/antiram/full, etc.

Also create a way to use these to make an output map structure for the ena map code, matching the map descriptor properties (here just tiling type and spacing are relevant.)

E.g., 
```
if "6deg" is in the descriptor, it should make a `RectangularSkyMap` with spacing_deg == 6
if "nside32" is in the descriptor, it should make a `HealpixSkyMap` with nside == 32
```

### LINKED ISSUE #1754: Ultra L2: Pass descriptor string from CLI and parse for output map shape.

nkerman:
# Change Summary

## Overview

Most of this PR is about revamping the functions used to build and parse the ENA map names. I refactored the existing build_map_descriptor into methods of a class, which allows for easy parsing of the descriptor. So, for instance, we can call 

```
descriptor = MapDescriptor.from_string("u45-ena-h-sf-nsp-full-hae-6deg-1yr")
descriptor.instrument => "ultra"
descriptor.sensor => "45"
descriptor.resolution  => "6deg"
```

This allows us to use the `get_output_map_structure_from_descriptor_string(descriptor)` function to create output map structures based on the `descriptor` passed in as a string.

Closes #1752
Closes #1753

The other component of this PR adds a `descriptor` parameter to the Ultra L2 main function, which uses the `get_output_map_structure_from_descriptor_string` method mentioned above to take a descriptor passed in with the CLI arguments to determine the output map style (`Healpix`/`Rectangular`) and resolution (`nside32`/`2deg`). 

Closes #1754

## Updated Files

- imap_processing/cli.py
   - Pass in  `self.descriptor` to ultra L2
- imap_processing/ena_maps/utils/naming.py
   - Class based approach to name creation and parsing 
   - Use the descriptor provided by the CLI to override the default output map parameters

## Testing

- Tests for all the naming methods and for Ultra L2 with descriptor parameter.


imap_processing/ena_maps/utils/naming.py
```diff
@@ -23,8 +26,350 @@ class MappableInstrumentShortName(Enum):
...
+            sensor = f"{sensor:03}"
+        else:
+            raise ValueError("Integer sensor values are only valid for LO instruments.")
+    # Hi and Ultra may be either "45", "90", or "combined", in which case
+    # Hi should get the sensor "ic" and Ultra should get the sensor "lc"
+    elif sensor == "combined":
+        if instrument is MappableInstrumentShortName.ULTRA:
+            sensor = "lc"
+        elif instrument is MappableInstrumentShortName.HI:
+            sensor = "ic"
```
tmplummer: I thought it was supposed to be "hic" or "ulc" for combined.

.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 15
+    timeout-minutes: 20
```
nkerman: The ubuntu 3.10 test runner that downloads data and runs some extra tests was _just_ timing out at 15 minutes (it succeeds in 15 min, 0 seconds 🤦)
Screenshot 2025-05-20 at 11 19 23 AM

greglucas: I think we should remove the timeout-minutes, not sure why it was added as I didn't follow that addition. We are on the free tier and not paying for these minutes.

imap_processing/ena_maps/utils/naming.py
```diff
@@ -23,8 +26,350 @@ class MappableInstrumentShortName(Enum):
     GLOWS = "glx"
 
 
-# Ignore too many branches warning
-def build_l2_map_descriptor(  # noqa: PLR0912
+def get_instrument_descriptor(
+    instrument: MappableInstrumentShortName,
+    sensor: int | Literal["45", "90", "combined"] | str = "",
```
greglucas: Having both Literal and str for a type seems a bit odd. Do you need to accept generic strings too?

imap_processing/ena_maps/utils/naming.py
```diff
@@ -23,8 +26,350 @@ class MappableInstrumentShortName(Enum):
...
+        The duration to parse. This can be a string in the format "1yr", "6mo", etc.,
+        an integer representing the number of days, or a timedelta object.
+
+    Returns
+    -------
+    str
+        The parsed duration string in the format "1yr", "6mo", etc.
+    """
+    if isinstance(duration, timedelta):
+        # Convert timedelta to a string representation of number of 28.5 day months
```
greglucas: Where does 28.5 days come from? Is that something from IBEX? We should document why we are choosing that number somewhere here because it is non-obvious to me.

imap_processing/ena_maps/utils/naming.py
```diff
@@ -23,8 +26,350 @@ class MappableInstrumentShortName(Enum):
...
+    instrument: MappableInstrumentShortName
+    frame: str | SpiceFrame
+    resolution_str: str
+    duration: str | int | timedelta
+    sensor: int | str = ""
+    principal_data: str = "ena"
+    species: str = "h"
+    survival_corrected: str = "nsp"
+    spin_phase: str = "full"
+    coordinate_system: str = "hae"
```
greglucas: I am a bit confused here why there are so many options for these types. I think this class should be the defacto storage type that will be used everywhere else. So you should dictate what you want these to be. Then your from_string() and to_string() methods would do the conversion of your storage to the desired input/output.

My thought is that you could have your frame be only a SpiceFrame here. Then you can have a post_init method that would do some of these conversions for you if you are coming directly from an init member. Although it actually looks like you'll almost always be coming from the classmethod?
https://docs.python.org/3/library/dataclasses.html#post-init-processing

imap_processing/ena_maps/utils/naming.py
```diff
@@ -23,8 +26,350 @@ class MappableInstrumentShortName(Enum):
...
+
+    # Quantities parsed into strings that will fit in the descriptor
+    @property
+    def frame_str(self) -> str:
```
greglucas: I'm a bit confused with these properties. They are piping out into the other functions you defined. Can you put the functions into the definitions here on this class instead since they seem like they are only useful for the class?

greglucas (commented):
A few broad comments.
* I think we should be more strict in the typing of this class and what you want the internal storage mechanism to be and cast to/from that in the methods.
* The extra functions should be methods on the class I think instead of standalone which would encapsulate some of the logic a little better and avoid some indirection.

.github/workflows/test.yml
nkerman: Got it - rerunning tests now without this

imap_processing/ena_maps/utils/naming.py
nkerman: I'll try to get rid of the str type here - it was an attempt to appease mypy, but there's probably a cleaner way

imap_processing/ena_maps/utils/naming.py
nkerman: Tim and I were just discussing this in [another PR](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1756#discussion_r2098254613).

Short answer: I vaguely recall finding an edge case where 28.5 day months was a better simple assumption than 30 day months; however I can't replicate it, so I'll switch to 30 day months here.

imap_processing/ena_maps/utils/naming.py
nkerman: I was able to remove this - it just required parsing 3 digit sensor numbers to ints and 2 digit ones to Literal of ["45", "90"]

imap_processing/ena_maps/utils/naming.py
nkerman: > I am a bit confused here why there are so many options for these types. I think this class should be the defacto storage type that will be used everywhere else. So you should dictate what you want these to be. Then your from_string() and to_string() methods would do the conversion of your storage to the desired input/output.

I'm trying to tighten it up a bit. Because of the way the IMWG wants to define these descriptor strings, certain variables are a bit messy (especially sensor, which for the LO instrument types can  be any 3 digit integer, but for the other ENA instruments can only be "45" or "90", OR a combined indicator string, which is not the same for Ultra(`lc`)/Hi(`ic`)).

> Although it actually looks like you'll almost always be coming from the classmethod?

Yes, I expect we'll pretty much always go from the `from_string` classmethod. I don't think this should always be a `SpiceFrame`, because there's no obvious `SpiceFrame` for the 'hk' frame yet.

imap_processing/ena_maps/utils/naming.py
nkerman: Done (moved the functions into static methods of the class)

For the irregular parts of the descriptor (especially the instrument and sensor component which get combined in a complicated way, which differs in structure based on the instrument), I think we need a method to both go from the 2 values e.g., `(LO_HIGH_THROUGHPUT, 75)` to the string representation which will go in the map descriptor `("t075")`. We also need to be able to do the opposite, parsing `("t075")` into those components.

I used the property syntax for parsing the attributes into the short strings which will fit into the `MapDescriptor.to_string` method, while still letting us access `MapDescriptor.instrument` and `MapDescriptor.sensor` separately. I don't think the staticmethod itself should be a property though, as it is useful to be able to parse a "t123" into `(LO_HIGH_THROUGHPUT, 123)` outside of a full `MapDescription` object.

nkerman:
Thanks for the detailed comments, @greglucas ! I removed the `build_l2_map_descriptor` function to enforce the object-oriented way of building descriptors, moved all the functions into static methods of the class, and tightened up the types. 

I also fixed a bunch of rebase conflicts I let sneak in 🤦

imap_processing/ena_maps/utils/naming.py
```diff
@@ -26,148 +29,369 @@ class MappableInstrumentShortName(Enum):
...
+        Get the instrument descriptor string.
+
+        Returns
+        -------
+        str
+            The instrument descriptor string.
+        """
+        return MapDescriptor.get_instrument_descriptor(self.instrument, self.sensor)
+
+    @classmethod
```
tmplummer: Just a personal preference...
I like to organize methods in a class somewhat in the order that one would use them. For example, `__init__` and `@classmethod`s go at the top, followed by `@property` methods, then other methods. I think this helps someone looking at the class for the first time understand it from the high-level down.

In this case, it seems like a user will mostly be using the `from_string` and `to_string` methods so they should come first followed by all of the functions that support these methods.

imap_processing/ena_maps/utils/naming.py
```diff
@@ -26,148 +29,369 @@ class MappableInstrumentShortName(Enum):
...
+            The parsed frame string.
+        """
+        if isinstance(frame, SpiceFrame):
+            match frame:
+                case SpiceFrame.IMAP_DPS.value:
+                    return "sf"
+                case SpiceFrame.ECLIPJ2000.value:
+                    return "hf"
+                case _:
+                    raise NotImplementedError(f"Frame {frame} is not yet implemented.")
```
tmplummer: I'm a bit confused by this method. It seems to be serving multiple purposes. I would expect a method named `parse_xxxx` to take a string as input and return something else. Could this piece of code be moved into another method? Something like `spice_frame_to_str`?

imap_processing/ena_maps/utils/naming.py
```diff
@@ -203,6 +427,54 @@ def ns_to_duration_months(ns: int) -> int:
...
+        If the descriptor string is invalid.
+    """
+    map_descriptor = MapDescriptor.from_string(descriptor)
+
+    # If "deg" is in the resolution string, then this is a rectangular map
+    # (e.g., '2deg')
+    if "deg" in map_descriptor.resolution_str:
+        return ena_maps.RectangularSkyMap(
+            spacing_deg=float(map_descriptor.resolution_str.split("deg")[0]),
+            spice_frame=MapDescriptor.get_map_coord_frame(
```
tmplummer: Can you add a `@property(spice_frame)` so that this can just be:
`spice_frame=map_descriptor.spice_frame`?

tmplummer (commented):
Thanks for putting this together! Another useful piece of common code.
Just a couple of comments to follow up on.

imap_processing/ena_maps/utils/naming.py
nkerman: That makes sense from an end user perspective. I tend to start with the basic methods at the bottom of the tree, and then write the ones which call them; however, the methods in this PR got a bit jumbled and so I should probably refactor the order. 

I'll move those methods up like you suggest!

imap_processing/ena_maps/utils/naming.py
nkerman: The way this `instrument_descriptor` is getting built connects an instrument short name ("u" for Ultra, "h" for Hi, "idx" for IDEX) to a sensor ("lc" for combined ultra, "ic" for combined Hi, "" for IDEX). In these three examples, you'll get out: "ulc", "hic", "idx" for the `instrument_descriptor`, that first part of the description string

imap_processing/ena_maps/utils/naming.py
nkerman: Yeah - since in the descriptor string, we have both the inertial/velocity frame `(hf, sf, hk)` and the coordinate system which we and the instrument teams generally use interchangeably with the term "frame", I'm using the property name "map_spice_coord_frame" to make it a bit more explicit.

imap_processing/ena_maps/utils/naming.py
nkerman: Makes sense - I was thinking of it as more like "parse an input frame", which would get a consistent representation from either a SpiceFrame (preferable for a PSET/existing file) or its string (preferable for the descriptor string). 

I can extract that out into a function like you're talking about.

nkerman: On closer examination, I think there's too much risk here of mixing up the inertial frame and coordinate frame - e.g., a helio (hf) pset can still be in the coordinates of `IMAP_DPS`. 

I'm going to make it so that only strings are valid for the inertial frame.

nkerman:
Finished addressing Greg + Tim's comments.

imap_processing/ena_maps/utils/naming.py
```diff
@@ -203,6 +365,50 @@ def ns_to_duration_months(ns: int) -> int:
     return int(months)
 
 
+def get_output_map_structure_from_descriptor_string(
```
greglucas: Should this be a method on the class?

```python
MapDescriptor.from_string(descriptor).to_empty_map()
```

greglucas (approved):
Looks good to me. Just one minor suggestion for the extra hanging function you can take/leave as you see fit. I'm fine with either way.

imap_processing/ena_maps/utils/naming.py
nkerman: Yeah, that works! I'll change it here and in Ultra L2

Issue #1747, opened 2025-05-16
Map epoch and epoch_delta


tmplummer:
### Algorithm Description:

A temporary method for setting the map `epoch` and `epoch_delta` values has been implemented in the ena_maps.AbstractSkyMap.project_pset_values_to_map()` method. It just keeps track of the min and max epoch from psets that get projected to the map.

@nkerman pointed out that the end epoch should really be the end of the final PSET that is included instead of the start which is what is in the `epoch` variable. This ticket should fix that by either adding `epoch_delta` to all PSET products and getting the end epoch from `end_epoch = epoch + epoch_delta`, or we should keep track of the min and max pointing numbers. Using those we can query for the min pointing start time and max pointing end time to set the map `epoch` and `epoch_delta` values.

See PR discussion here:
_Originally posted by @nkerman in https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1743#discussion_r2093412782_

PR #1756, opened 2025-05-19
Ultra L2: Add epoch_delta and map duration

### LINKED ISSUE #1696: Ultra L2: Track map duration
We need to know the duration of the map when setting the logical source. It's currently defaulting to "99mo"

See TODO in ultra_l2.py

nkerman:
# Change Summary

## Overview

Ultra L2: calculate an epoch_delta in nanoseconds based on: `(max PSET epoch + 1 day) - the min PSET epoch`. Add a TODO to replace this 1 day with the true duration of the final PSET, but that is not yet available. See #1747. This should always be a small difference compared to the duration of an L2 map, so a day is a good starting assumption.  

Also add a function to take such a time delta in ns and return a month number calculated by a very simple 1month = 28.5 days, then use that function to set the map duration. 

Closes #1696 

## Updated Files

   - Add epoch_delta to the common map parameters
-  imap_processing/ena_maps/utils/naming.py
   - define ns_to_duration_months function
   - Ensure we get the epoch = min PSET time
   - Calculate epoch_delta and add it to the dataset
   - Set the map duration based on epoch_delta. 

## Testing

Tests for new function, duration of output L2 map

### DIFF
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -55,14 +55,11 @@ energy_label:
 
 epoch_delta:
   <<: *default_int64
-  CATDESC: Time difference between the start and end of the exposure.
+  CATDESC: Number of nanoseconds covered by data included in this map product.
   FIELDNAM: epoch_delta
   UNITS: ns
   VAR_TYPE: support_data
   DISPLAY_TYPE: no_plot
-  DEPEND_0: epoch
-  dtype: int64
-  TIME_BASE: J2000
   TIME_SCALE: Terrestrial Time
 
 energy_delta_minus:
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -53,6 +53,18 @@ energy_label:
   FORMAT: A16
   DEPEND_1: energy
 
+epoch_delta:
+  <<: *default_int64
+  CATDESC: Time difference between the start and end of the exposure.
+  FIELDNAM: epoch_delta
+  UNITS: ns
+  VAR_TYPE: support_data
+  DISPLAY_TYPE: no_plot
+  DEPEND_0: epoch
+  dtype: int64
+  TIME_BASE: J2000
+  TIME_SCALE: Terrestrial Time
+
 energy_delta_minus:
   <<: *default_float32
   VAR_TYPE: support_data
--- a/imap_processing/ena_maps/utils/naming.py
+++ b/imap_processing/ena_maps/utils/naming.py
@@ -8,6 +8,9 @@ from typing import Literal
 
 from imap_processing.spice.geometry import SpiceFrame
 
+# Set a constant number of days in a month to calculate the duration of maps
+DAYS_IN_MONTH = 28.5
+
 
 class MappableInstrumentShortName(Enum):
     """Enumeration of the short names of the ENA and other mappable instruments."""
@@ -133,12 +136,12 @@ def build_l2_map_descriptor(  # noqa: PLR0912
 
     # Handle duration
     if isinstance(duration, timedelta):
-        # Convert timedelta to a string representation of number of 28.5 day months
-        num_months = int(duration.days // 28.5)
+        # Convert timedelta to str representation of number of DAYS_IN_MONTH-day months
+        num_months = int(duration.days // DAYS_IN_MONTH)
         duration = f"{num_months}mo"
     elif isinstance(duration, int):
-        # Assume number of days and convert to 28.5-day months
-        duration = f"{int(duration // 28.5)}mo"
+        # Assume number of days and convert to DAYS_IN_MONTH-day months
+        duration = f"{int(duration // DAYS_IN_MONTH)}mo"
     elif isinstance(duration, str):
         pass
     # Replace 12mo with 1yr
@@ -169,7 +172,7 @@ def build_l2_map_descriptor(  # noqa: PLR0912
 
 def ns_to_duration_months(ns: int) -> int:
     """
-    Convert nanoseconds to months using 28.5 days per month approximation.
+    Convert nanoseconds to months using DAYS_IN_MONTH days per month approximation.
 
     Parameters
     ----------
@@ -186,7 +189,7 @@ def ns_to_duration_months(ns: int) -> int:
     This can be used to convert from the difference between two epochs in ns to the
     number of months between them.
 
-    This is a very simple estimate, which assumes that a month is 28.5 days and
+    This is a very simple estimate, which assumes that a month is DAYS_IN_MONTH days and
     floors the result.
 
     This successfully yields:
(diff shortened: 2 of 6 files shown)


imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -53,6 +53,18 @@ energy_label:
   FORMAT: A16
   DEPEND_1: energy
 
+epoch_delta:
+  <<: *default_int64
+  CATDESC: Time difference between the start and end of the exposure.
```
tmplummer: I have a slight opposition to calling the map duration an "exposure".

We are going to have a merge conflict. I added these variable attributes to my PR yesterday. My CATDESC is `Number of nanoseconds covered by data included in this map product.`

The description isn't too important at this point b/c I want the science team to edit all of the descriptions anyway.

imap_processing/ena_maps/utils/naming.py
```diff
@@ -167,6 +167,39 @@ def build_l2_map_descriptor(  # noqa: PLR0912
...
+
+    This successfully yields:
+    - 12 months for 365.25 days in ns
+    - 6 months for 182.625 days (365.25/2) in ns
+    - 4 months for 121.75 days (365.25/3) in ns
+    - 3 months for 91.3125 days (365.25/4) in ns
+    """
+    days = ns / (1e9 * 60 * 60 * 24)
+    months = days // 28.5
+    return int(months)
```
tmplummer: No problem with this, but just FYI, numpy has a `timedelta64` class that makes conversions easy:
```
days = np.timedelta64(int(ns), "ns").astype("timedelta64[D]")
```
It looks like it uses the generic definition of 1 month = 30 days.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -360,10 +366,25 @@ def ultra_l2(
...
+            [
+                pset_epochs.min(),
+            ],
+        ),
+    )
+
+    # TODO: replace 1 day in ns below with the actual end time of the last PSET.
+    # Currently assumes the end time of the last PSET is 1 day after its start.
+    map_duration_ns = (pset_epochs.max() + (86400 * 1e9)) - pset_epochs.min()
+    map_duration_months_int = naming.ns_to_duration_months(map_duration_ns)
```
tmplummer, line 386: I think that ultimately we will be able to parse the map duration from the descriptor input to the processing job.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -53,6 +53,18 @@ energy_label:
...
+  FIELDNAM: epoch_delta
+  UNITS: ns
+  VAR_TYPE: support_data
+  DISPLAY_TYPE: no_plot
+  DEPEND_0: epoch
+  dtype: int64
```
tmplummer: These are defined in `default_int64`.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -53,6 +53,18 @@ energy_label:
...
+  TIME_BASE: J2000
```
tmplummer: I don't think we actually want to specify this time base as the true time base is whatever is in the `epoch` variable.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -221,8 +225,10 @@ def generate_ultra_healpix_skymap(
             f"PULL Variables: {output_map_structure.values_to_pull_project}"
         )
 
+    all_pset_epochs = []
     for ultra_l1c_pset in ultra_l1c_psets:
         pointing_set = ena_maps.UltraPointingSet(ultra_l1c_pset)
+        all_pset_epochs.append(pointing_set.epoch)
```
tmplummer, line 231: This is a much easier to understand implementation than what I did with keeping track of the min/max each iteration through the loop.

tmplummer (approved):
The only change requests are regarding the `epoch_delta` attributes. Otherwise, looks good.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
nkerman: Agreed - I'll switch to the string you wrote out here.

~~Out of curiosity, why is exposure not a valid concept here? It seems pretty much totally analogous to the concept of exposure time with light~~
On reflection, I see the problem given the uneven spun exposures across different pixels.

imap_processing/ena_maps/utils/naming.py
nkerman: Thanks Tim, I didn't investigate that one. The reason I chose 28.5 days is that it gives the 'correct' number of months for these key conversions, based on a full 365.25 day orbital period of IMAP\*
```
# 365.25 days  should be 12 months
(365.25/1)/28.5 = 12.81578947 --> 12
# half a year should be 6 months
365.25 * (1/2) = 182.625 ; 182.5 / 28.5 = 6.4 --> 6
# 1/3 of a year should be 4 months
(365.25/3) / 28.5 = 4.27 --> 4
# 1/4 of a year should be 3 months
(365.25/4) / 28.5 = 3.2 --> 3
# 3/4 of a year should be 9 months
(365.25 * (3/4)) / 28.5 = 9.6 --> 9
```

\* While the IMWG hasn't decided on any definition of their "month", I think we will need to always think about things in terms of fractions of a 365.25 day period, rather than Jan-Mar, Apr-Jul, etc.

nkerman: Well apparently I did the math wrong last time, because it looks like this also works for 30... I was sure there was a reason to use 28.5, but I can't seem to find/remember it. if you'd prefer 30, we can switch to it.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
nkerman: Got it, so just cut the TIME_BASE here?

imap_processing/ultra/l2/ultra_l2.py
nkerman: I like yours though too - it's elegant and it should be more efficient, I just didn't think we needed efficiency here.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Agreed - I feel more comfortable outputting the measured time duration at present, but, especially with the new descriptor string parsing, this will be easy.

imap_processing/ena_maps/utils/naming.py
nkerman: For now, I just factored out this to a `DAYS_IN_MONTH` constant, so we can easily change it if we decide to later.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
tmplummer: Correct.

PR #1758, opened 2025-05-20
Lo L1C - PSET Exposure Time

### LINKED ISSUE #1757: Lo L1C - PSET Exposure Times
Add calculation for L1C PSET Exposure Times. This should be a 1 x 3600 x 40 x 7 array. If there are more than one exposure time in a bin the average of the exposure times should be taken.

sdhoyt:
# Change Summary

## Overview
This PR adds the exposure time calculation to the L1C PSET and adds a small function to L1B DE to retain the average spin duration for each event which is needed for the exposure time calculation.

## Updated Files
   - Added set_avg_spin_durations_per_event function to retain the avg spin duration for each event
   - changed name of get_avg_spin_durations to get_avg_spin_durations_per_cycle to differentiate from new function
   - added the exposure time calculation

## Testing
Added unit test for new L1B function and new unit test for L1C function.

Closes #1757

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -52,7 +52,9 @@ def lo_l1b(dependencies: dict) -> list[Path]:
         # Get the start and end times for each spin epoch
         acq_start, acq_end = convert_start_end_acq_times(spin_data)
         # Get the average spin durations for each epoch
-        avg_spin_durations = get_avg_spin_durations(acq_start, acq_end)
+        avg_spin_durations_per_cycle = get_avg_spin_durations_per_cycle(
+            acq_start, acq_end
+        )
         # get spin angle (0 - 360 degrees) for each DE
         spin_angle = get_spin_angle(l1a_de)
         # calculate and set the spin bin based on the spin angle
@@ -63,9 +65,15 @@ def lo_l1b(dependencies: dict) -> list[Path]:
         # get spin start times for each event
         spin_start_time = get_spin_start_times(l1a_de, l1b_de, spin_data, acq_end)
         # get the absolute met for each event
-        l1b_de = set_event_met(l1a_de, l1b_de, spin_start_time, avg_spin_durations)
+        l1b_de = set_event_met(
+            l1a_de, l1b_de, spin_start_time, avg_spin_durations_per_cycle
+        )
         # set the epoch for each event
         l1b_de = set_each_event_epoch(l1b_de)
+        # Set the average spin duration for each direct event
+        l1b_de = set_avg_spin_durations_per_event(
+            l1a_de, l1b_de, avg_spin_durations_per_cycle
+        )
         # calculate the TOF1 for golden triples
         # store in the l1a dataset to use in l1b calculations
         l1a_de = calculate_tof1_for_golden_triples(l1a_de)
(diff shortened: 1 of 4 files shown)


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -407,6 +415,36 @@ def set_each_event_epoch(l1b_de: xr.Dataset) -> xr.Dataset:
...
+        The average spin duration for each spin epoch.
+
+    Returns
+    -------
+    l1b_de : xarray.Dataset
+        The L1B DE dataset with the average spin duration added.
+    """
+    # repeat the average spin durations for each cycle based on the direct event count
+    # to get an average spin duration for each direct event. This will be used in L1C
+    # to calculate the exposure time for each direct event.
```
greglucas, line 446: This could be in the extended docstring too so it would be rendered in the documentation.

Is my understanding correct that `l1a_de["de_count"]` is number of de's per cycle? So this calculation is saying if I had 2 cycles of 3 and 4 events the output array would be: "cycle1, cycle1, cycle1, cycle2, cycle2, cycle2, cycle2"?

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -241,6 +246,52 @@ def create_pset_counts(
...
+    exposure_time : xarray.DataArray
+        The exposure times for the L1B Direct Event dataset.
+    """
+    # Create bin edges
+    lon_edges = np.arange(3601)
+    lat_edges = np.arange(41)
+    energy_edges = np.arange(8)
+
+    data = np.column_stack(
+        (l1b_de["pointing_bin_lon"], l1b_de["pointing_bin_lat"], l1b_de["esa_step"])
```
greglucas, line 275: just to clarify since I haven't looked through the code in detail. The `pointing_bin_lon` is the bin number and not the center longitude of the bin? i.e. it is the integer index corresponding to your int range of 0-3600 above and not 0-360.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -241,6 +246,52 @@ def create_pset_counts(
...
+    )
+
+    result = binned_statistic_dd(
+        data,
+        l1b_de["avg_spin_durations"].to_numpy(),
```
greglucas: Wouldn't average spin duration be covering the entire 3600 points, so do you need to divide this up to get the exposure time per-pixel based upon the avg_spin_durations? Something like this over 3600 instead.

greglucas (approved):
Looks good overall. One question on what exposure time means in the context of per-pointing-pixel.

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: That's correct. I'll add it to the docstring

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: Correct. The pointing_bin_lon is the bin index for each event

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: Yeah, not sure how I missed that. Thanks for catching that!

PR #1762, opened 2025-05-20
Adding example for spin file in CLI

maxinelasp:
# Change Summary
I need the spin data for GLOWS L1B, so I thought I would add it using the new input types/SPICE stuff as an example for other instruments (and to confirm I am doing this the correct way.)

I also updated all the tests to line up with the format of the file most recently uploaded May 8th: s3://sds-data-449431850278/imap/spice/spin/imap_2025_122_2025_122_01.spin.csv

I'm not sure if the format of that file is correct or if we expect it to change again, but it sounded like it was newer and more accurate than the test file as I believe that one came out of MSIM3. 

## Overview

- Add file path option to the existing spin file
- Change all the union types to use `|` because mypy is annoying
- add example in GLOWS L1B of separating out and processing spin file
- Updated spin.py to use the new file format
- Updated test data and tests to use the new format

## Testing

Tested using the command:
``` --instrument glows --data-level l1b --repointing repoint00001 --dependency "[{\"type\": \"science\", \"files\": [\"imap_glows_l1a_hist_20110920-repoint00001_v002.cdf\"]}, {\"type\": \"spin\", \"files\": [\"imap_2025_122_2025_122_01.spin.csv\"]}]" --version "003"```


tmplummer (commented):
This incorrectly changes the format of the spin-table. For the correct format, see: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/173715570/Spin+and+Repoint+MOC+Data+product#SpinandRepointMOCDataproduct-SpinDataProductPacket/FileContentsTable
Recent examples from MSIM3 are in s3 and have dates around 2026_268/269

One other comment, and this was probably not communicated well by me...
I am working on a PR that will cache the location of the spin and repoint tables in the pre_processing step so that, similar to spice kernels, you will not need to pass them around in your processing.

maxinelasp:
@subagonsouth Ok, that sounds great! In that case I'll close this PR and wait to see what yours ends up looking like. This took a short amount of time so no worries about any wasted work.

PR #1763, opened 2025-05-20
I-ALiRT - SIT Prep Draft

laspsandoval:
# Change Summary

## Overview
Starting to test with the SC packet instead of instrument packets.

## Updated Files
- process_hit.py
   - Use a timestamp in the sc packet since it is not available in the instrument packet
- grouping.py
   - Address packets with no data

## Testing
- test_parse_mag.py
- test_process_hit.py
- test_process_swe.py
- conftest.py

### DIFF
--- a/imap_processing/ialirt/l0/parse_mag.py
+++ b/imap_processing/ialirt/l0/parse_mag.py
@@ -1,6 +1,7 @@
 """Functions to support I-ALiRT MAG packet parsing."""
 
 import logging
+from decimal import Decimal
 from typing import Union
 
 import numpy as np
@@ -389,16 +390,14 @@ def process_packet(
         mag_data.append(
             {
                 "apid": 478,
-                # Start of the group.
-                "met": met.values.min(),
+                "met": int(met.values.min()),
                 "utc": met_to_utc(met.values.min()),
-                "ttj2000ns": met_to_ttj2000ns(met.values.min()),
-                # Vectors.
-                "mag_4s_b_gse": [0, 0, 0],
-                "mag_4s_b_gsm": [0, 0, 0],
-                "mag_4s_b_rtn": [0, 0, 0],
-                "mag_phi_4s_b_gsm": 0,
-                "mag_theta_4s_b_gsm": 0,
+                "ttj2000ns": int(met_to_ttj2000ns(met.values.min())),
+                "mag_4s_b_gse": [Decimal("0.0") for _ in range(3)],
+                "mag_4s_b_gsm": [Decimal("0.0") for _ in range(3)],
+                "mag_4s_b_rtn": [Decimal("0.0") for _ in range(3)],
+                "mag_phi_4s_b_gsm": Decimal("0.0"),
+                "mag_theta_4s_b_gsm": Decimal("0.0"),
             }
         )
 
--- a/imap_processing/ialirt/l0/parse_mag.py
+++ b/imap_processing/ialirt/l0/parse_mag.py
@@ -391,7 +391,7 @@ def process_packet(
             {
                 "apid": 478,
                 "met": int(met.values.min()),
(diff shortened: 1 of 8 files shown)


imap_processing/tests/ialirt/unit/test_parse_mag.py
```diff
@@ -253,7 +235,7 @@ def test_process_packet(xarray_data, mag_test_data, calibration_dataset):
     xarray_data["sc_sclk_sec"] = xarray_data["mag_acq_tm_coarse"]
     xarray_data["sc_sclk_sub_sec"] = xarray_data["mag_acq_tm_fine"]
 
-    parsed_packets = process_packet(xarray_data, calibration_dataset)
+    _, parsed_packets = process_packet(xarray_data, calibration_dataset)
```
torimarbois, line 238: Why are you now throwing away the `mag_data` list?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -381,6 +383,22 @@ def process_packet(
             }
         )
 
-        mag_data.append({**status_data, **science_data, **time_data})
+        l1b_data.append({**status_data, **science_data, **time_data})
 
-    return mag_data
+        # Placeholder for real data.
+        met = grouped_data["met"][(grouped_data["group"] == group).values]
+        mag_data.append(
```
torimarbois, line 390: I'm curious about the type of `mag_data` - why is it a list of dictionaries, instead of just a dictionary? I'm assuming it's because eventually, it will actually be a list of dictionaries, but with only fake data available for the moment, it only contains one group/dictionary? Just checking.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -462,6 +465,14 @@ def process_swe(accumulated_data: xr.Dataset, in_flight_cal_files: list) -> list
         accumulated_data["swe_acq_sec"], accumulated_data["swe_acq_sub"]
     )
     accumulated_data["time_seconds"] = time_seconds
+    # Subsecond time conversion specified in 7516-9054 GSW-FSW ICD.
```
torimarbois, line 468: The time conversion for SWE and HIT are specified in the same document?

imap_processing/tests/ialirt/unit/test_process_hit.py
```diff
@@ -114,20 +127,26 @@ def test_prefixes():
     assert HIT_PREFIX_TO_RATE_TYPE["SLOW_RATE"] == expected_slow_rate
 
 
-def test_find_groups(xarray_data):
-    """Tests find_groups"""
-
-    filtered_data = find_groups(xarray_data)
+def test_create_l1(xarray_data):
+    """Tests create_l1"""
```
torimarbois: Super nit: can you add `()` at the end of the function name to make this slightly easier to read?

imap_processing/ialirt/l0/process_swe.py
laspsandoval: Yep!

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: It is fake data at the moment. We need to start testing and MAG isn't finished yet. I'm going back and forth between a PR in sds-data-manager and this PR to make certain the formatting is aligned. I'll add you to that one too sometime today.

imap_processing/tests/ialirt/unit/test_parse_mag.py
laspsandoval: Yes, so this is a temporary fix with the fake data. I'll need to go back and make this more modular after the SIT.

torimarbois (approved):
LGTM!

PR #1765, opened 2025-05-21
1760 cache spin and repoint dependencies in pre processing

### LINKED ISSUE #1132: Write utils function to query and load spin data

### LINKED ISSUE #1133: Write utils function to query and load repoint data

### LINKED ISSUE #1735: Join data from multiple spin table csvs
This is great! I will use this comment to generate a new ticket.

The spin data functionality will need to join data from multiple spin-tables.

_Originally posted by @subagonsouth in https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/590#discussion_r2089652274_

### LINKED ISSUE #1760: Cache spin and repoint dependencies in pre_processing
### Algorithm Description:

Similar to how spice kernels are furnished in the pre_processing step, we need to cache the locations of the spin and repoint tables.

In addition, we need to add logic to join multiple spin tables into a single dataframe in the function that reads that data in.

tmplummer:
# Change Summary

## Overview
This PR adds automatic handling of spin and repoint dependencies in the `pre_processing` step. Rather than loading tables from an environment variable, mutable module level attributes are set up to hold the path or paths to the CSV files. When the `get_spin_data()` or `get_repoint_data()` functions are called, the path(s) are pulled from those attributes.

Additionally, I added caching of the dataframes produced so that subsequent calls don't require reloading the table from disk.

## Updated Files
- imap_processing/cli.py
   - Set the spin and repoint paths in the pre_processing step
   - Add mutable module level attribute for storing the path to the repoint table
   - Add setter for the above attribute
   - Use the new attribute when loading the repoint table from disk
   - Add caching of the repoint table to the `get_repoint_data()` function
- imap_processing/spice/spin.py
   - Add mutable module level attribute for storing the path to the spin tables
   - Add setter for the above attribute
   - Use the new attribute when loading the spin tables from disk
   - Add caching of the spin table dataframe to the `get_spin_data()` function
   - Add merging of multiple input spin tables
   - Fix up spin and repoint table fixtures to set module global attributes instead of environment variables
   - Update tests for repoint changes
   - Add test coverage for attribute setter
- imap_processing/tests/spice/test_spin.py
   - Update tests for spin changes
   - Add test coverage for attribute setter
   - Add test for merging multiple spin tables
   - 
   - Add test coverage for setting spin and repoint paths from dependencies

Closes: #1760


imap_processing/spice/repoint.py
```diff
@@ -1,16 +1,49 @@
...
 
+# Copy-on-write will be enabled by default in pandas 3.0
+# It is recommended to enable it now. See:
+# https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
+pd.options.mode.copy_on_write = True
+
 logger = logging.getLogger(__name__)
 
+# Use a mutable module level attribute to store the location of the repoint table
+_repoint_table_path: Path | None = None
```
maxinelasp: I really don't like the global variable... why not stick this stuff into a class so we can pass around an instance? You could also put it in the `SpinInput` class (as some kind of set_dataframe() method) and then it will operate in the same kind of way as this does. 

If you don't want to update imap_data_access you could extend ProcessingInput into an imap_processing specific class and replace the dependency with a new ProcessingInput with the dataframe included.

imap_processing/spice/spin.py
```diff
@@ -13,14 +14,45 @@
...
+    """
+    Set the paths to input spin-table csv files.
+
+    Parameters
+    ----------
+    paths : list[Path]
+        List of paths to spin-table csv files that will be used to supply
+        spin-table data.
+    """
+    global _spin_table_paths  # noqa: PLW0603
```
maxinelasp: I don't think we should `noqa` this particular ruff rule - I agree with ruff that global variables can cause a lot of difficult-to-understand issues

maxinelasp (changes requested):
I think these changes should actually probably be in imap-data-access - since we're passing around classes with dependencies in them, we can put the spin pandas dataframe in there. I can see these global variables causing lots of difficult to diagnose problems for us...

imap_processing/spice/repoint.py
tmplummer: I generally agree that global variables should be avoided. I think this is an acceptable time to use them. I view the spin table query functions as being very similar to the SPICE queries. I don't think we want to have to pass spin and repoint paths, or dataframe objects, through many layers of functions in order to make the data available where it is needed. SPICE works in a very similar way where we tell the SPICE library about the data in pre-processing and then make queries throughout our code to retrieve what we want.

maxinelasp: We're already passing around ancillary files and I think this falls in the same category - ie it is a data source that code needs to access and so it should be explicitly passed around into places that use it. I dislike having implicit dependencies like this, even if it means we need to pass around more objects. I also think this design would be difficult to test against - i.e. not difficult to test, but difficult if you depend on it. 

I am also worried about the global object getting mutated and messing up processing at later steps. Those kinds of bugs can be really difficult to track down. If we could make the object an immutable type I would feel better about it. 

I took a peek into the Spiceypy code, and the `furnsh` method actually calls directly out to SPICE so I think that's a different case. 

Things that would make me feel better about this design:
- Putting the variable into a singleton class so we can name the class something like `SpinSingleton` and make it clear this is globally shared. I also worry about people by-passing the `get()` method you have and using the variable directly. You can then use the `@property` decorator to override the class method for `get()` and people can then use `SpinSingleton.spin` and always use the get() method you wrote.
- if the spin table path is ever updated, we should invalidate the cache

tmplummer: This seems like an impasse. I simply don't have the time to totally re-architect this. Can global variables cause hard to debug issues? Sure. I think that the following mitigate the hazards sufficiently:
- The global is named starting with an underscore indicating it is private so no one should directly mutate it.
- Even if someone did change the spin or repoint files pointed to by the global, things would be OK.
   - The `get_spin_data()` function sorts the files to the appropriate order so that when they are combined, the resulting dataframe is correct. The only side effect would be that there might not be the coverage needed in which case an exception is raised saying that the query time is outside of the spin-table data.
   - Changing the repoint file will simply mean that the time coverage is not the same. Each repoint table file is the same as the prior version with additional data appended. Again, worst case, an exception gets raised.

> I took a peek into the Spiceypy code, and the furnsh method actually calls directly out to SPICE so I think that's a different case.

Sure, but just like someone can mutate a private module variable, someone can also call `spiceypy.furnish()` with some random kernel, or even a reordered list of the original kernels that were furnished, and royally screw things up.

## Path forward?
I guess that I will cherry pick the changes to the `get_spin_data()` class that add the functionality to combine multiple spin files and then make the current environment variable implementation carry a json string with the multiple file paths. It won't change any of your concerns, but perhaps we can get past SIT4 with it?

maxinelasp: Well, I hate the way SPICE does it, but necessary evils!

Since we're under a tight deadline for SIT-4, I am ok with this being merged. Maybe we can re-visit after the SIT.

maxinelasp (approved):

imap_processing/spice/repoint.py
greglucas: I think this is a place where we do want to somehow load these automagically for users without needing to pipe them through everywhere similar to SPICE, so I think this makes sense. Maybe a middle-ground would be to remove the global calls by moving these global items to a separate singleton module that can be referenced here instead? I think that would be similar to Maxine's suggestion of the singleton class, but without the need for defining a class and rather just defining your attribute at the module level instead of class level.

Something like this setup:

**spice/config.py**
```python
_repoint_table_path = None
_spin_table_paths = None
```

**spice/repoint.py**
```python
from imap_processing.spice import config

def set_repoint_path(path):
    config._repoint_table_path = path

...
```

imap_processing/spice/repoint.py
```diff
@@ -1,16 +1,49 @@
...
+# cache repoint-table data to avoid reloading from disk every time
+_repoint_df_cache: dict[str, pd.DataFrame] = {}
+
+
+def set_repoint_table_paths(paths: list[Path]) -> None:
```
greglucas: Do we want to name these similar to SPICE so call it furnish_repoint_file() or something like that? Maybe set_global_repoint_table() to add in global and indicate it is updating this globally.

imap_processing/spice/repoint.py
```diff
@@ -42,16 +75,27 @@ def get_repoint_data() -> pd.DataFrame:
...
+        )
+
+    logger.debug(
+        f"Reading in the following repoint table file: {_repoint_table_path.name}"
+    )
+
+    cache_key = str(_repoint_table_path.resolve())
+    if cache_key in _repoint_df_cache:
+        logger.debug("Returning cached repointing data")
+        return _repoint_df_cache[cache_key]
```
greglucas: I don't think this will work equivalently to what you have here since there is some extra processing involved, but wanted to mention that sometimes I have made a separate function that can be cached internal to your external function so that I don't have to make a cache and can just leverage the functools implementation. Something like:

```python
def get_repoint_data():
    path = _from_global
    repoint_df = _load_repoint_data_with_cache(path)

@functools.cache
def _load_repoint_data_with_cache(path):
    return pd.read_csv(path)
```

greglucas (approved):
I agree that this is something that will be helpful to get in for SIT-4, and hopefully any clean-up would only improve on the approach but wouldn't mean this was wasted effort / wrong. I have a few minor suggestions/comments, but they are non-critical and can be merged over if you don't have time/desire to look into them right now.

imap_processing/spice/repoint.py
tmplummer: Good call. I was thinking along these lines, but couldn't quite make the leap to this pattern.

Issue #1108, opened 2024-11-01
I-ALiRT: codice lo algorithm


laspsandoval:
### Requirements:

> **Specify: (optional)**

### Algorithm Code Information:
> Input data:
> > **Specify details:**
>
> Algorithm steps/psuedocode:
> > **Specify details:**
>
> Output data:
> > **Specify details:**

### Code:

```python
# Code
```

tech3371:
I believe this is blocked by new SCI-LUT upload to onboard by CoDICE team. New SCI-LUT details been reviewed by project CCB. This task is blocked by that. Then if and when SDC receive new SCI-LUT, it may be blocked for a bit when we need to create the new SCI-LUT json.

laspsandoval:
Instrument team approved that CoDICE-Lo data may be released. There was a single data point out of ~10^6 that is slightly off. A ticket was created in response, but marked as very low priority.

PR #1768, opened 2025-05-22
Implement/Validate CoDICE L1a lo-ialirt

### LINKED ISSUE #785: CoDICE: Implement L1a i-alirt processing

bourque:
This PR implements and validates the CoDICE L1a `lo-ialirt` data product.

A few things to note:
- "Level L1a" doesn't really mean anything from the perspective of I-ALiRT, as that team just needs the end products, not intermediary products. But for the sake of my own pipeline and milestones, I am producing an intermediary product for this. The L1a dataset will be used in the future to calculate rates and the final I-ALiRT products.
- When I visited SwRI to work with Joey, we spent a lot of time working through the intricacies/nuances of I-ALiRT processing. We took an approach that started from scratch, and as such we came up with an algorithm that significantly differs from from @laspsandoval originally wrote in `ialirt.l0.process_codicelo.py`. As such, I removed a lot of code from that module that is now implemented within the CoDICE L1a pipeline. I just wanted to note that I do feel bad that I removed a lot of Laura's original work, but that code was a great starting point for discussion and definitely served as inspiration for the solution we came up with.
- There are a few additions in the PR for `hi-ialirt` (namely CDF attribute and configuration stuff), but I still need to fully implement and validate that produce, which I will do in a future PR.
- Keep in mind that the `lo-ialirt` product is basically a subset of the `lo-sw-species` L1a product, so I tried my best to re-use code for that product.
- With implementing the `lo-ialirt` product, I had to implement a new decompression method (the CoDICE team calls this "pack 24 bit compression") that was introduced relatively recently. So there are some code changes to parts of code that haven't been touched in some time.

Pertains to #785, #1108, #251 and #252


imap_processing/codice/codice_l1a.py
```diff
@@ -93,22 +93,28 @@ def calculate_epoch_values(self) -> NDArray[int]:
         Calculate and return the values to be used for `epoch`.
 
         On CoDICE, the epoch values are derived from the `acq_start_seconds` and
-        `acq_start_subseconds` fields in the packet. Note that the
-        `acq_start_subseconds` field needs to be converted from microseconds to
-        seconds
+        `acq_start_subseconds` fields in the packet. The exception to this is
+        are the I-ALiRT packets, which use "acquisition_time".
```
```suggestion
        `acq_start_subseconds` fields in the packet. The exception to this is
        the I-ALiRT packets, which use "acquisition_time".
```

imap_processing/codice/codice_l1a.py
```diff
@@ -129,18 +135,31 @@ def decompress_data(self, science_values: list[str]) -> None:
...
+                )
+                decompressed_values = decompress(values, compression_algorithm)
+                self.raw_data.append(decompressed_values)
+
+        else:
+            for packet_data, byte_count in zip(
+                science_values, self.dataset.byte_count.data
+            ):
+                # Convert from numpy array to byte object
+                values = ast.literal_eval(str(packet_data))
```
greglucas: I don't think you need the ast literal_eval here and could get the values in some other way?

imap_processing/codice/codice_l1a.py
```diff
@@ -118,9 +124,9 @@ def decompress_data(self, science_values: list[str]) -> None:
 
         Parameters
         ----------
-        science_values : list[str]
-            A list of byte strings representing the science values of the data
-            for each packet.
+        science_values : list[str] | str
+            A list of (or a single) byte string(s) representing the science
+            values of the data for each packet.
```
greglucas: Would it be easier to wrap the single i-alirt string in a list before this so you don't have to account for both cases? It'd just be a one-element list.

imap_processing/codice/constants.py
```diff
@@ -719,6 +764,29 @@
...
 }
 
+# Define the packet fields needed to be stored in segmented data and their
+# corresponding bit lengths for direct event data products
+IAL_BIT_STRUCTURE = {
+    "SHCOARSE": 32,
+    "PACKET_VERSION": 16,
+    "SPIN_PERIOD": 16,
+    "ACQ_START_SECONDS": 32,
+    "ACQ_START_SUBSECONDS": 20,
```
greglucas, line 774: I thought you had only a acquisition_time field per your above comments/code?

Should these be in Laura's root packet definition instead so you can unpack them as `CODICE_PLAN_ID` etc.?

Do you have SHCOARSE transmitted in the CoDICE portion of the packet? I thought the entire packet just had one SHCOARSE associated with it for all instrument to reference.

imap_processing/codice/codice_l1a.py
bourque: Thanks for pointing that out, I can just use the native `eval()` instead.

greglucas: I'm confused, why would you need to use "eval" at all. We aren't getting something like "a < b" are we? i.e. I think we should be able to create the bytes directly through normal methods/conversions and not go through the human-readable inputs.

imap_processing/codice/codice_l1a.py
bourque: Upon further examination I actually had this wrong. Even in my corner case of I-ALiRT data this is a `list[str]`, so thanks for commenting here!

imap_processing/codice/constants.py
bourque: Thanks for pointing this out. It got me thinking more about this and indeed I should have been calculating `epoch` based on `acq_start_seconds` and `acq_start_subseconds` instead of `shcoarse`.

Each CoDICE I-ALiRT packet has all three of `SHCOARSE`, `ACQ_START_SECONDS`, and `ACQ_START_SUBSECONDS`, but the CoDICE team prefers to se the acq starts to determine epoch instead of `SHCOARSE`.

Since these are all stored in the `DATA_xx` fields of the packets themselves, they need to be parsed directly from the bit strings of the `DATA_xx` fields, and thus I am not sure that could be easily captured in a XML packet definition(?)

imap_processing/codice/codice_l1a.py
bourque: The starting point for this (what I call `science_values`) is a list of 0-dimensional numpy arrays that represent the `packet.data` field for each `packet`. This is what comes comes out of the `utils.packet_file_to_datasets` function. (I realize now that my type hints and doc strings don't reflect this, so I will have to update that).

So I need to go from something like `np.array(b']\x00\x00\x08 ...')` to a `<class 'bytes'>` type object so that I can apply the `byte_count` cutoff and then go on to decompress the bytes to a list of integers.

I am a bit lost on this now so I am not sure how best to do that conversion. The `eval()` method has been working for me since the start so I hadn't put much thought into it.

If i try to use `packet_data.tobytes()` instead, I notice that I get a wildly different value than if I did `eval()`, for example:

`eval()` method yields `b']\x00\x00\x08\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x00\x82T\x1e\xf1\xfd\x8f%\x960\xc0d?\x19\xd8\xcd\xf0G\x94\x04\xde\xc0\x90\x1fL\n\xf2\xe0\xab>\x05\xd4\xbe$\xbd\xb3ll\xbc\xcf\x96\x82/Tcn\xe1\xac;\x9a\x93\xc1A\xc3\x7f\x8e\xc6\xdc\xb0B\xde\xa1\x0e lco2c\x97M\x9d\x92\x1c\x0c\xb4Y\xe7\xfct\xb8\xe7\xcb\xfd\x99\x86\x03O\xd3\x0f\xc2\xc8\xb8\x0c\xc2\xe0^p\x7f\xff\xe4*0\x00\x00'`

`tobytes()` method yields `b'0J:\x1d\x01\x00\x00\x00'`

bourque: On a side note, I now remember why I was using `ast.literal_eval`; it is because using just `eval()` throws a `S307` warning in `ruff`: 

```
S307 Use of possibly insecure function; consider using `ast.literal_eval`
```

greglucas: Yes, I would agree with using literal_eval over eval, but I think we shouldn't be using it at all. I'm not sure exactly what is going on, but I can't replicate that.

```python
>>> x = np.array(b']\x00\x00\x08\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x00\x82T\x1e\xf1\xfd\x8f%\x960\xc0d?\x19\xd8\xcd\xf0G\x94\x04\xde\xc0\x90\x1fL\n\xf2\xe0\xab>\x05\xd4\xbe$\xbd\xb3ll\xbc\xcf\x96\x82/Tcn\xe1\xac;\x9a\x93\xc1A\xc3\x7f\x8e\xc6\xdc\xb0B\xde\xa1\x0e lco2c\x97M\x9d\x92\x1c\x0c\xb4Y\xe7\xfct\xb8\xe7\xcb\xfd\x99\x86\x03O\xd3\x0f\xc2\xc8\xb8\x0c\xc2\xe0^p\x7f\xff\xe4*0\x00\x00')

>>> x[()].tobytes()
b']\x00\x00\x08\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x00\x82T\x1e\xf1\xfd\x8f%\x960\xc0d?\x19\xd8\xcd\xf0G\x94\x04\xde\xc0\x90\x1fL\n\xf2\xe0\xab>\x05\xd4\xbe$\xbd\xb3ll\xbc\xcf\x96\x82/Tcn\xe1\xac;\x9a\x93\xc1A\xc3\x7f\x8e\xc6\xdc\xb0B\xde\xa1\x0e lco2c\x97M\x9d\x92\x1c\x0c\xb4Y\xe7\xfct\xb8\xe7\xcb\xfd\x99\x86\x03O\xd3\x0f\xc2\xc8\xb8\x0c\xc2\xe0^p\x7f\xff\xe4*0'

>>> x[()].tobytes() == eval(str(x[()]))
True
```

bourque: Wow, I never have seen that before. Today I learned that `[()]` is how you access a value stored in a 0-dimensional array.

So I can just use `packet_data[()]` then

greglucas: Thanks to @subagonsouth since I saw him using that a lot in the SPICE stuff :)

greglucas (approved):

imap_processing/codice/constants.py
greglucas: This is slightly unclear to me as well. I-ALiRT is a bit odd because the packet definition is per a single event, but you are stacking DATA_xx fields across multiple packets. So the definitions are indeed a bit odd and likely not helpful. So I agree with your approach here for now 👍

PR #1769, opened 2025-05-22
IDEX L2B Add science acquisition times from IDEX housekeeping dataset 

lacoak21:
# Change Summary

## Overview

This update comes from an IDEX SCR request. Idex l2b needs instrument logs of IDEX science acquisition start and stop times. The IDEX team queries oasis for this but we decided to decom the event message housekeeping packets that we expect to receive once a day and add them as a dependency to l2b. Since we expect weekly IDEX science packets, IDEX l2b will expect all the idex l1b event message packets decommed since the last science packet. The science acquisition event messages will then be decoded and saved to the L2b CDF along with the timestamps. 

## Updated Files
- imap_processing/cdf/config/imap_idex_l2b_variable_attrs.yaml
   - Add message and timestamp attrs
- imap_processing/cli.py
   - Add l1b hk event message dependency to idex l2b processing call 
   - enum for event message codes
  - function for get science acquire times. 

## Testing
Add test and update fixture for l2b


imap_processing/cli.py
```diff
@@ -826,10 +826,10 @@ def do_processing(
             science_files = dependencies.get_file_paths(source="idex")
             datasets = PacketParser(science_files[0]).data
         elif self.data_level == "l1b":
-            if len(dependency_list) > 3:
+            if len(dependency_list) > 2:
```
greglucas: If you are expecting an explicit number you should test against that number. Here and below.
```suggestion
            if len(dependency_list) != 2:
```

imap_processing/cli.py
```diff
@@ -846,19 +846,20 @@ def do_processing(
             dependency = load_cdf(science_files[0])
             datasets = [idex_l2a(dependency)]
         elif self.data_level == "l2b":
-            if len(dependency_list) > 2:
+            if len(dependency_list) > 3:
                 raise ValueError(
                     f"Unexpected dependencies found for IDEX L2B:"
-                    f"{dependency_list}. Expected only two dependency."
+                    f"{dependency_list}. Expected only three dependency."
```
```suggestion
                    f"{dependency_list}. Expected only three dependencies."
```

imap_processing/cli.py
```diff
@@ -846,19 +846,20 @@ def do_processing(
...
             sci_files = dependencies.get_file_paths(
                 source="idex", descriptor="sci-1week"
             )
-            dependency = load_cdf(sci_files[0])
-            # TODO update l2b to use hk files
-            # hk_files = dependencies.get_file_paths(source="idex", descriptor="evt")
-            # hk_dependency = [load_cdf(dep) for dep in hk_files]
-            datasets = [idex_l2b(dependency)]
+            # There should only be one file in the list but just in case sort them
+            # by the start date and load the most recent.
```
greglucas: If this happens do you want to raise instead? Sorting something that shouldn't need to be sorted seems like a recipe for confusion later.

imap_processing/idex/idex_constants.py
```diff
@@ -85,3 +85,11 @@ class ConversionFactors(float, Enum):
...
 # Define the pointing reference frame for IDEX
 IDEX_EVENT_REFERENCE_FRAME = SpiceFrame.ECLIPJ2000
+
+
+class IDEXEvtAcquireCodes(IntEnum):
+    """Create ENUM for event message ints that signify science acquire events."""
+
+    ACQSETUP = 2
+    ACQ = 3
+    CHILL = 5
```
greglucas: 😎

imap_processing/idex/idex_l2b.py
```diff
@@ -14,7 +14,7 @@
     l1a_data = PacketParser(l0_file)
     l1b_data = idex_l1b(l1a_data)
     l1a_data = idex_l2a(l1b_data)
-    l2b_data = idex_l2b(l2a_data)
+    l2b_data = idex_l2b(l2a_data, [evt_data, evt_data2])
```
greglucas: Where did these `evt_data` come from in this example? I think you need to load them somewhere above.

imap_processing/idex/idex_l2b.py
```diff
@@ -50,29 +53,29 @@ def idex_l2b(l2a_dataset: xr.Dataset) -> xr.Dataset:
...
+
     # Create l2b Dataset
-    l2b_dataset = xr.Dataset(
-        coords={"epoch": epoch_da},
-        attrs=idex_attrs.get_global_attributes("imap_idex_l2b_sci"),
+    prefixes = ["latitude", "longitude", "_dust_mass_estimate", "_impact_charge"]
+    l2b_dataset = setup_dataset(l2a_dataset, prefixes, idex_attrs)
+    l2b_dataset.attrs = idex_attrs.get_global_attributes("imap_idex_l2b_sci")
+
+    # Get science acquisition start and stop times from evnt dataset
```
greglucas:
```suggestion
    # Get science acquisition start and stop times from event dataset
```

imap_processing/idex/idex_l2b.py
```diff
@@ -120,3 +122,55 @@ def round_spin_phases(spin_phases: xr.DataArray) -> xr.DataArray:
     # Calculate nearest quadrant value.
     # Use mod to wrap values > 315 to 0.
     return (quadrant_size * (shifted_spin_phases / quadrant_size).astype(int)) % 360
+
+
+def get_science_acquire_timestamps(
```
greglucas: This is the only place you use "acquire" so maybe change it to "acquisition" for consistency with everywhere else?
```suggestion
def get_science_acquisition_timestamps(
```

imap_processing/idex/idex_l2b.py
```diff
@@ -120,3 +122,55 @@ def round_spin_phases(spin_phases: xr.DataArray) -> xr.DataArray:
...
+        if (v1, v2) in [
+            (IDEXEvtAcquireCodes.ACQSETUP, IDEXEvtAcquireCodes.ACQ),
+            (IDEXEvtAcquireCodes.ACQ, IDEXEvtAcquireCodes.CHILL),
+        ]:
+            message = (
+                f"SCI state change: {IDEXEvtAcquireCodes(v1).name} to"
+                f" {IDEXEvtAcquireCodes(v2).name}"
+            )
+            event_logs.append(message)
+            event_timestamps.append(epoch)
```
greglucas: Rather than storing this only as text, do you want to also store a boolean or state change variable indicating what it was as a value? `if (v1, v2) == (IDEXEvtAcquireCodes.ACQSETUP, IDEXEvtAcquireCodes.ACQ): event_values.append(True)`

imap_processing/idex/idex_constants.py
lacoak21: 😆

imap_processing/cli.py
lacoak21: Actually I handled this case in batch starter so I can remove the sorting.

imap_processing/idex/idex_l2b.py
lacoak21: True!

imap_processing/idex/idex_l2b.py
lacoak21: Hm that is a good idea. So True if a "science acquire start" and false for "science acquire stop"?

greglucas: That is what I would go with, but that is totally up to you and the instrument team with whether this is even helpful or not. You could also make it a uint8 flag with multiple options too if you want to store any other state changes or think you might expand it in the future etc..

lacoak21: OK I think for now I will stick with the boolean values for start and stop. Its a good idea rather than relying on just text. The programmer part of me did not like that 😄

lacoak21: Actually for ISTP reasons I made it start = 1 stop = 0

lacoak21:
pre-commit.ci autofix

greglucas (approved):

PR #1772, opened 2025-05-22
Document Update For Query '--table' Param

daralynnrhode:
# Adding the --table Parameter to Documentation

## Overview

Updated the openapi.yml file to include the new parameter. I believe this is the only location it needs updated for, but correct me if I am wrong. 

## Updated Files

- openapi.yml


maxinelasp (approved):
looks good, thanks for adding this! After merging, just check the documentation to make sure it looks good and was uploaded properly

PR #1776, opened 2025-05-23
1729 add pointing frame kernel job to clipy

### LINKED ISSUE #1729: Add pointing frame kernel job to cli.py
### Algorithm Description:

The functional pieces have been written. We need to add the job to cli.py so that when the batch starter triggers it, we will produce the new Pointing kernels.

tmplummer:
# Change Summary

## Overview
This adds the final pieces for the Pointing attitude kernel generation job. I have added the job level `spice` to the allowed inputs for `spacecraft`. It also required a custom `post_processing` method to handle the Path object(s) returned by the function that generates the file.

## Updated Files
- imap_processing/__init__.py
- imap_processing/cli.py
   - Added another @final method, `cleanup` to the processing workflow. This allows for custom `post_processing` methods while still performing required cleanup actions.
   - Added the pointing frame creation job to `Spacecraft.do_processing`
   - Added custom `Spacecraft.post_processing` method.
   - Added the high level function that generates the pointing attitude kernel. The naming just matches the input CK file that the new kernel is derived from.
   - Added test coverage for the new kernel generation function
   - Added test coverage for the new Spacecraft code
- pyproject.toml
  - Bumped the imap-data-access version

Closes: #1729


tmplummer:
pre-commit.ci autofix

imap_processing/cli.py
```diff
@@ -540,7 +543,7 @@ def post_processing(
         # If it is start_date, skip repointing in the output filename.
 
         products = []
-        for ds in datasets:
+        for ds in processed_data:
```
greglucas, line 557: Another option here would be to put your isinstance check in this loop.

```python
if isinstance(ds, xr.Dataset):
    ... do current attribute injection and writing cdf
else:
    # A path to a product that was already written out
    products.append(ds)
```

imap_processing/cli.py
```diff
@@ -1060,25 +1066,70 @@ def do_processing(
...
+            A list of either datasets (products) produced by do_processing or
+            Paths to pointing_kernels that have been generated.
+        dependencies : ProcessingInputCollection
+            Object containing dependencies to process.
+        """
+        # If the datasets list contains a xr.Dataset, call the super method
+        if isinstance(processed_data[0], xr.Dataset):
+            super().post_processing(processed_data, dependencies)
+            return
+        # Otherwise, we need to upload the pointing attitude kennels
```
greglucas: Just one now?
```suggestion
        # Otherwise, we need to upload the pointing attitude kernel
```

greglucas (approved):
One minor suggestion that I think might help avoid needing to override post_processing for the spacecraft job. But, looks good to me either way.

imap_processing/cli.py
tmplummer: 😆  I went back and forth several times on which implementation to choose. I'm happy to do it that way if you like it better.

greglucas: 🤷 I don't really care either, maybe a +0.1 to this other way (less code and avoids potentially missing call to super), but you have my approval either way you choose.

imap_processing/spice/pointing_frame.py
```diff
@@ -35,6 +37,39 @@
...
+    # TODO: For now just use the input CK start/end dates. It is possible that
+    #    the end date is incorrect b/c the repoint table determines the last
+    #    segment in the pointing kernel.
+    input_file_parts = re.match(
+        SPICEFilePath.attitude_file_pattern, imap_attitude_ck.name
+    ).groupdict()
+    pointing_kernel_path = (
+        imap_attitude_ck.parent / f"imap_dps_{input_file_parts['start_year_doy']}_"
+        f"{input_file_parts['end_year_doy']}_{input_file_parts['version']}.ah.bc"
+    )
```
tech3371: Does setting `spice_obj=SPICEFilePath()` and then using `spice_obj.spice_metadata["start_date"].strftime("%Y_%j")` or `spice_obj.spice_metadata["version"]` help here?

imap_processing/__init__.py
```diff
@@ -27,7 +27,7 @@
     "idex": ["l1a", "l1b", "l2a", "l2b"],
     "lo": ["l1a", "l1b", "l1c", "l2"],
     "mag": ["l1a", "l1b", "l1c", "l1d", "l2"],
-    "spacecraft": ["l1a"],
+    "spacecraft": ["l1a", "spice"],
```
tech3371, line 30: nice!

tech3371 (approved):
lgtm

imap_processing/spice/pointing_frame.py
tmplummer: Oh, yes. That is much cleaner. Thanks!

PR #1777, opened 2025-05-23
I-ALiRT - cdf creation

laspsandoval:
# Change Summary

## Overview
Takes the output of a DynamoDB query and converts it into a xarray that will be written to cdf.

## New Files
- create_xarray.py
   - Takes the output of a DynamoDB query and converts it into a xarray that will be written to cdf.
- imap_ialirt_global_cdf_attrs.yaml
   - Adds attributes
- imap_ialirt_l1_variable_attrs.yaml
   - Adds attributes

## Testing
- test_create_xarray.py
   - Tests the creation of the xarray.

### DIFF
--- a/imap_processing/cdf/config/imap_ialirt_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_ialirt_global_cdf_attrs.yaml
@@ -1,0 +1,18 @@
+instrument_base: &instrument_base
+  Descriptor: IALiRT>IMAP IALiRT (IMAP Active Link in Real Time)
+  TEXT: >
+    The Interstellar Mapping and Acceleration Probe (IMAP) mission includes the Active Link
+    for Real-Time (I-ALiRT) system, which provides continuous, low-latency space weather data.
+    I-ALiRT broadcasts real-time measurements from IMAP via NASA’s Deep Space Network and
+    international ground stations. It has data from 5 of the IMAP instruments:
+    SWE (Solar Wind Electron instrument), CoDICE (Compact Dual Ion Composition Experiment),
+    HIT (High-energy Ion Telescope), MAG (Magnetometer),
+    and SWAPI (Solar Wind and Pickup Ion instrument).
+    See https://imap.princeton.edu/instruments/ultra for more details.
+  Instrument_type: "Multiple Instruments"
+
+imap_ialirt_l1_realtime:
+    <<: *instrument_base
+    Data_type: L1_REALTIME>Level-1 Realtime
+    Logical_source: imap_ialirt_l1_realtime
+    Logical_source_description: IMAP-IALiRT Instrument Level-1 Realtime Data.
--- a/imap_processing/cdf/config/imap_ialirt_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_ialirt_global_cdf_attrs.yaml
@@ -1,14 +1,14 @@
 instrument_base: &instrument_base
-  Descriptor: IALiRT>IMAP IALiRT (IMAP Active Link in Real Time)
+  Descriptor: IALiRT>IMAP I-ALiRT (IMAP Active Link in Real Time)
   TEXT: >
     The Interstellar Mapping and Acceleration Probe (IMAP) mission includes the Active Link
     for Real-Time (I-ALiRT) system, which provides continuous, low-latency space weather data.
     I-ALiRT broadcasts real-time measurements from IMAP via NASA’s Deep Space Network and
     international ground stations. It has data from 5 of the IMAP instruments:
-    SWE (Solar Wind Electron instrument), CoDICE (Compact Dual Ion Composition Experiment),
+    SWE (Solar Wind Electron), CoDICE (Compact Dual Ion Composition Experiment),
     HIT (High-energy Ion Telescope), MAG (Magnetometer),
-    and SWAPI (Solar Wind and Pickup Ion instrument).
-    See https://imap.princeton.edu/instruments/ultra for more details.
+    and SWAPI (Solar Wind and Pickup Ion).
+    See https://imap.princeton.edu for more details.
   Instrument_type: "Multiple Instruments"
 
 imap_ialirt_l1_realtime:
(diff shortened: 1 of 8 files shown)


imap_processing/cdf/config/imap_ialirt_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,18 @@
+instrument_base: &instrument_base
+  Descriptor: IALiRT>IMAP IALiRT (IMAP Active Link in Real Time)
```
greglucas: Do you need to include an extra hyphen in the name?
```suggestion
  Descriptor: IALiRT>IMAP I-ALiRT (IMAP Active Link in Real Time)
```

imap_processing/cdf/config/imap_ialirt_global_cdf_attrs.yaml
```diff
@@ -0,0 +1,18 @@
...
+  TEXT: >
+    The Interstellar Mapping and Acceleration Probe (IMAP) mission includes the Active Link
+    for Real-Time (I-ALiRT) system, which provides continuous, low-latency space weather data.
+    I-ALiRT broadcasts real-time measurements from IMAP via NASA’s Deep Space Network and
+    international ground stations. It has data from 5 of the IMAP instruments:
+    SWE (Solar Wind Electron instrument), CoDICE (Compact Dual Ion Composition Experiment),
+    HIT (High-energy Ion Telescope), MAG (Magnetometer),
+    and SWAPI (Solar Wind and Pickup Ion instrument).
```
greglucas: For consistency I'd remove the extra "instrument" from SWE and SWAPI.
```suggestion
    SWE (Solar Wind Electron), CoDICE (Compact Dual Ion Composition Experiment),
    HIT (High-energy Ion Telescope), MAG (Magnetometer),
    and SWAPI (Solar Wind and Pickup Ion).
```

imap_processing/cdf/config/imap_ialirt_l1_variable_attrs.yaml
```diff
@@ -0,0 +1,370 @@
...
+  DISPLAY_TYPE: time_series
+  FILLVAL: -9223372036854775808
+  FORMAT: I12
+  VALIDMIN: -9223372036854775808
+  VALIDMAX: 9223372036854775807
+  VAR_TYPE: data
+  UNITS: " "
+
+component:
+  CATDESC: Velocity components (vx, vy, vz)
```
greglucas: Component is a direction, the velocity could have components, but a component shouldn't reference velocity.
```suggestion
component:
  CATDESC: Component direction (x, y, z)
```

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -0,0 +1,96 @@
...
+    cdf_manager.add_instrument_global_attrs("ialirt")
+    cdf_manager.add_instrument_variable_attrs("ialirt", "l1")
+
+    instrument_prefixes = ("swe", "hit", "mag", "codicelo", "codicehi", "swapi")
+    instrument_keys: set[str] = set()
+
+    for record in records:
+        instrument_keys.update(
+            key for key in record if key.startswith(instrument_prefixes)
+        )
```
greglucas: I'm not totally following this portion here. You are looping through your records twice. Can you bring this logic into the for-loop below this?

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Convert to columns, add fillvals, and associate with datatype.
+    data_dict = defaultdict(list)
+    for record in records:
+        for key in instrument_keys:
```
greglucas: Shouldn't you be looping over everything within the records. What happens if you have records that have content that you miss here, that would silently be dropped.

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+    for key in sorted(instrument_keys):
+        data = data_dict[key]
+        dims = ["epoch", "component"] if key.startswith("mag") else ["epoch"]
+        dataset[key] = xr.DataArray(
+            data,
+            dims=dims,
+            attrs=cdf_manager.get_variable_attributes(key, check_schema=False),
+        )
```
greglucas: There are a lot of duplicate loops that could be combined into one loop over the records.

My suggestion would be to create an empty dataset first based on the length of records, but it would have `np.empty(nrecords, dtype)` as appropriate for all the dataarrays you expect. Then within the loop you can assign `ds[variable].data[i] = record[variable]` and the casting will happen for you I believe.

imap_processing/ialirt/utils/create_xarray.py
laspsandoval: Here I'm collecting all keys that start with the instrument prefixes.

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -0,0 +1,90 @@
...
+    attrs = cdf_manager.get_variable_attributes("default_int64_attrs")
+    fillval = attrs.get("FILLVAL")
+    ttj2000ns_values = np.full(n, fillval, dtype=np.int64)
+
+    # Collect all keys that start with the instrument prefixes.
+    for i, record in enumerate(records):
+        ttj2000ns_values[i] = record["ttj2000ns"]
+        instrument_keys.update(
+            key for key in record if key.startswith(instrument_prefixes)
+        )
```
greglucas: Do the keys vary for every record or is there a way you can just look at the first record to get this information? Then below you could set the epoch's data to be `np.empty(len(records))` and fill in the ttj2000 in your final loop like the other variables.

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+    coords = {"epoch": epoch, "component": component}
+    dataset = xr.Dataset(
+        coords=coords,
+        attrs=cdf_manager.get_global_attributes("imap_ialirt_l1_realtime"),
+    )
+
+    # Create empty dataset for each key.
+    for key in instrument_keys:
```
greglucas, line 58: The one thing I'm slightly worried about here is that we are iterating over what you've found. If a bad/anomalous variable crops up or a variable goes missing, then this will dynamically change sizes throughout the mission. I think SPDF said they wanted the variables consistent across the data products, so it would be nice if you could make this a hard-coded list of fields and then fill them in that way. If you don't get data for a particular instrument that product would be all nan's then and not missing.

torimarbois (approved):

PR #1781, opened 2025-05-27
Update imap-data-access version

sdhoyt:
# Change Summary

## Overview
This PR updates the version of imap-data-access to v0.30.0 which now works with SPICE science frames


poetry.lock
tech3371, line 1: Did you mean to add pyproject.toml file too? Looks like it's missing.

sdhoyt: I think I'm missing something. What needs to change with pyproject.toml?

tech3371: This file is generally produced when you update `imap-data-access` version and commit it. They usually go in pair. On this PR, I only see one file changed.

sdhoyt: This is what is looks like on this PR branch: https://github.com/sdhoyt/imap_processing/blob/imap-data-access/pyproject.toml

And this is the section with imap-data-access. Do I need to change that to be ">=0.30.0"?
```python
[tool.poetry.dependencies]
astropy-healpix = ">=1.0"
cdflib = "^1.3.1"
imap-data-access = ">=0.29.0"
python = ">=3.10,<4"
space_packet_parser = "^5.0.1"
spiceypy = ">=6.0.0"
xarray = '>=2024.10.0'
numpy = "<=3"
sammi-cdf = "^1.0"
scipy = "^1.13"
```

tech3371: yep

greglucas: Yes, agreed. The `poetry.lock` file should never be updated manually, pre-commit should take care of it for us. Update the pyproject.toml as Tenzin suggests and that should update the poetry.lock file during pre-commit. Then commit that changed file and then the check should pass for you.

greglucas (approved):

poetry.lock
greglucas: Looks like you updated this back to the pre-commit version as seen because it was automatically generated by 1.8.0 rather than 1.8.2 like it was previously 👍 
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/a39262fd56acfd794e16a4cf5d20131db6ff7937/.pre-commit-config.yaml#L30

PR #1784, opened 2025-05-27
Mag ancillary files for L1B

maxinelasp:
# Change Summary
Updates to MAG to use ancillary files for L1B. Tests required some updates to make the types all work.

This is very similar to the changes for MAG L2, and tidies up some stuff. 

This still leaves the option for an ancillary file as a default, but that will likely be removed once I can verify this works for SIT-4.


laspsandoval (commented):
Reviewed and approved I-ALiRT changes only.

imap_processing/ancillary/ancillary_dataset_combiner.py
```diff
@@ -72,8 +74,15 @@ def __init__(
         self.expected_end_date = expected_end_date
 
         self.timestamped_data = []
-        for file in ancillary_input.filename_list:
-            self.timestamped_data.append(self.convert_to_timestamped_data(file))
+        if isinstance(ancillary_input, ProcessingInput):
+            for file in ancillary_input.filename_list:
+                self.timestamped_data.append(self.convert_to_timestamped_data(file))
```
greglucas: Can you push this conversion down into the `convert_to_timestamped_data()` function so that it can accept a path and do this logic for you there?

imap_processing/ancillary/ancillary_dataset_combiner.py
```diff
@@ -132,6 +141,7 @@ def convert_file_to_dataset(self, filepath: str | Path) -> xr.Dataset:
         xr.Dataset
             The converted xarray dataset.
         """
+        print(f"converting filepath: {filepath}")
```
greglucas: logger

imap_processing/ancillary/ancillary_dataset_combiner.py
```diff
@@ -188,8 +198,13 @@ def _combine_input_datasets(self) -> xr.Dataset:
         # dimension, named like {datavar}_dim_0, {datavar}_dim_1, etc.
         for data_var in self.timestamped_data[0].dataset.data_vars:
             shape = self.timestamped_data[0].dataset[data_var].shape
+            var_type = self.timestamped_data[0].dataset[data_var].dtype
+            if issubclass(var_type.type, np.integer):
+                maxval = np.iinfo(var_type).max
+            else:
+                maxval = np.iinfo(np.int32).max
```
greglucas, line 205: If it is an integer you set it to the maximum of that int type which makes sense, but should the else: clause be `np.finfo(var_type.type).max`? I'm just not following the fallback and what could be passed in.

maxinelasp: Yeah I wasn't sure if I should use the max float because I was just worried that would cause issues in weird corner cases. The int max felt safer, I guess? But I can change it if you think I should make it the max float for consistency.

imap_processing/ancillary/ancillary_dataset_combiner.py
maxinelasp: Actually, since the iteration is happening here, it would make `convert_to_timestamped_data()` much more complicated (and change the method signature a lot which has ripple effects) to make that change. I also think it would be helpful now to be able to override exactly what's inside `convert_to_timestamped_data` to change it for other use cases. 

If you'd prefer to keep this logic out of init (understandable) I could move it to a new function, something like `convert_input`?

greglucas: This just looks like duplicated code to me with one minor change of `file` -> `file.name` being passed in. I figured you could invert the logic within the function and do something like:
```python
if isinstance(file, Path):
    file = file.name
...
```

It is fine to keep it here if that doesn't work.

imap_processing/ancillary/ancillary_dataset_combiner.py
greglucas: I'm not sure what maxval is used for, so I'm not sure what safety would mean, but I would want the maxval associated with whatever dtype you are provided, so if you get a float32 or float64 here I would want the corresponding max of those and not the int32 max.

maxinelasp: It's just a placeholder/default that I expect to get overridden in almost every case. I can update it to use float for those dtypes.

imap_processing/ancillary/ancillary_dataset_combiner.py
maxinelasp: Due to the way ancillary_input works (in particular, it stores filenames and not full Paths) making that change is annoying. I agree that it would be preferable, but there are some weird API/type issues here because this isn't quite what ProcessingInput was intended for, so some of the pieces are a little awkward unfortunately. I did look to see if it could be improved and IMO it's as simple as I can reasonably make it (ProcessingInput works great for the actual CLI calls but is extremely annoying in tests, which is why I added the option to just use Paths directly.)

maxinelasp: Let me look again with your isinstance idea... maybe that piece can be simplified

maxinelasp: `AncillaryFilePath` can accept a path or a str and does the check already, so I was able to get it a little cleaner! Thanks for the comment, I think that is better!

greglucas (approved):

PR #1789, opened 2025-05-28
Ultra L2: obs_date_range with iterative standard deviation

### LINKED ISSUE #1746: Maps - Use iterative standard deviation to compute obs_date_range
I think that Jon sent us a link to a way we can do this in his email titled "iterative standard deviation". 

[StackExchange thread](https://math.stackexchange.com/questions/2148877/iterative-calculation-of-mean-and-standard-deviation/2148949#2148949)

Looks like we can project 3 values separately: `all ones, obs_date (per-pixel), obs_date**2` and then compute $\sigma_{obs-date}$

_Originally posted by @nkerman in https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1743#discussion_r2093426871_

nkerman:
# Change Summary

Closes #1746

I'm hoping to make a version of this which is more shared across the ENA instruments, but I want to be sure to get as much of the code I've got on this over the fence before I leave next week. 

Basic idea of this: 
$$\sigma_{obs\ date} = \sqrt{\Sigma{(obs\ date^2) / N} - \Sigma{(obs\ date / N)^2}}$$

I've done some initial tests and the mean `obs_date_delta` is very close to the `std(pset.epoch for pset in input_psets)`.

### DIFF
--- a/imap_processing/ultra/l2/ultra_l2.py
+++ b/imap_processing/ultra/l2/ultra_l2.py
@@ -77,6 +77,8 @@ VARIABLES_TO_DROP_AFTER_INTENSITY_CALCULATION = [
     "pointing_set_exposure_times_solid_angle",
     "num_pointing_set_pixel_members",
     "corrected_count_rate",
+    "obs_date_for_std",
+    "obs_date_squared_for_std",
 ]
 
 # These variables may or may not be energy dependent, depending on the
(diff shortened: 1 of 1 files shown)


tmplummer (approved):
LGTM. Thanks for getting this in.

As an aside, I was wondering about doing this as a weighted standard deviation using exposure_time rather than `num_pointing_set_pixel_members`.

nkerman:
> LGTM. Thanks for getting this in.
> 
> As an aside, I was wondering about doing this as a weighted standard deviation using exposure_time rather than `num_pointing_set_pixel_members`.

Thanks Tim - That does make sense to me. As you know, I'm fairly skeptical about the usefulness of `obs_date_delta` as a whole, and given the formula we're drawing from is only for unweighted std, I think we should look for a proof of this before implementing it.

PR #1791, opened 2025-05-29
CoDICE L1a Updates for ISTP compliance

bourque:
This PR makes updates to the CoDICE L1a data products and their attributes to adhere to ISTP compliance. There are some more updates to make for this, but I figured this was a good point to make a PR. The biggest change here is adding labels (as CDF coordinates) for corresponding data variables (e.g. `esa_step_label`, `inst_az_label` etc.). I think this is the preferred method for SPDF?


imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml
```diff
@@ -16,196 +16,196 @@ instrument_base: &instrument_base
 # L1a
 imap_codice_l1a_hskp:
     <<: *instrument_base
-    Data_type: L1A_hskp->Level-1A Housekeeping Data
+    Data_type: L1A_hskp>Level-1A Housekeeping Data
```
bourque, line 19: These changes remove a `-` to avoid a SPDF API error in the `Logical_source` check

imap_processing/codice/codice_l1a.py
```diff
@@ -165,19 +165,17 @@ def define_coordinates(self) -> None:
...
-        # These are labels unique to lo-counters products coordinates
-        if self.config["dataset_name"] in [
-            "imap_codice_l1a_lo-counters-aggregated",
-            "imap_codice_l1a_lo-counters-singles",
-        ]:
-            coord_names.append("spin_sector_pairs_label")
+        coord_names = [
+            "epoch",
+            *self.config["output_dims"].keys(),
+            *[key + "_label" for key in self.config["output_dims"].keys()],
```
bourque, line 172: Added a line here to include `*_label` coordinates for each of the expected dimension coordinates

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -152,11 +194,12 @@ counters_attrs: &counters
     CATDESC: Fill in at creation
     DISPLAY_TYPE: time_series
     FIELDNAM: Fill in at creation
+    FILLVAL: *uint32_fillval
     FORMAT: I7
     SCALETYP: linear
     UNITS: counts
     VALIDMIN: 0
-    VALIDMAX: 8388607  # max value for a signed 24-bit integer
```
bourque, line 159: Not sure why I had a `uint24` value by default here. This is a corner-case for a few CoDICE data products, but not common.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -2315,103 +2424,120 @@ esaa_dac:
...
     <<: *hskp_default
     LABLAXIS: SENSOR_HV_DAC_STOP_OPTICS_GRID
     FIELDNAM: Sensor HV - Stop Optics Grid DAC
+    FILLVAL: *uint16_fillval
     CATDESC: Sensor HV - Stop Optics Grid DAC
 
 sbulk_vmon:
     <<: *hskp_default
     LABLAXIS: SBULK_VMON
-    FIELDNAM: HVPS – V1 -- Sensor Bulk Voltage Monitor
```
bourque, line 2388: A lot of these changes here are because the dash (`-`) that I had was some sort of copy/paste issue that was not of supported unicode by SDPF.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -1,30 +1,42 @@
...
+max_epoch: &max_epoch 3155630469184000000
+
 
 # <=== Defaults ===>
 default_attrs: &default
     DEPEND_0: epoch
     DISPLAY_TYPE: no_plot
     FIELDNAM: " "
-    FILLVAL: -9223372036854775808
+    FILLVAL: *uint8_fillval
```
greglucas: The fillval should align with the min/max datatype you've defined above, so i think you probably want to change this to int_fillval. Alternatively, you could change the min/max to be uint8 values.
```suggestion
    FILLVAL: *int_fillval
```

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
greglucas: If the value can only be 24 bits, that is an acceptable VALIDMAX. For example, you might save out a uint8 value that is just a boolean flag and thus can only be 0, 1 so the VALIDMAX would be 1 in that case and not 255.

imap_processing/codice/codice_l1a.py
```diff
@@ -198,11 +197,22 @@ def define_coordinates(self) -> None:
...
                 )
+                dims = [name]
+            elif name in [
+                "spin_sector_label",
+                "esa_step_label",
+                "inst_az_label",
+                "spin_sector_index_label",
+                "ssd_index_label",
+            ]:
+                key = name.split("_label")[0]
```
```suggestion
                key = name.remove_suffix("_label")
```

greglucas (approved):

imap_processing/codice/codice_l1a.py
bourque: I spent so many of my formative python years dealing with native string methods that I always forget about these `Pathlib` convenience methods 😅

bourque: Oh nvm, this is a native string method. I guess I should rephrase -- I spent many of my formative years dealing with Python 2.7 😂

PR #1792, opened 2025-05-29
Explicitly define table for querying

maxinelasp:
# Change Summary
Hit a minor bug in imap-data-access, thought we should explictly say we're lookign at the science table anyway.


bourque (approved):
LGTM!

PR #1793, opened 2025-05-29
Final l2 changes for menlo compatibility

nkerman:
# Change Summary

I'm out of office and don't have time to write this up yet, but Menlo may need it ASAP, so I'm opening this as a PR.

### DIFF
--- a/imap_processing/ultra/l2/ultra_l2.py
+++ b/imap_processing/ultra/l2/ultra_l2.py
@@ -549,6 +549,12 @@ def ultra_l2(
                 name=f"{coord_var}_label",
             )
 
+    # Add systematic error as all zeros with shape matching statistical unc
+    # TODO: update once we have information from the instrument team
+    map_dataset["ena_intensity_sys_err"] = xr.zeros_like(
+        map_dataset["ena_intensity_stat_unc"],
+    )
+
     # Add epoch_delta
     map_dataset.coords["epoch_delta"] = xr.DataArray(
         [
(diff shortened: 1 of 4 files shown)


imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -458,6 +458,16 @@ def ultra_l2(
                 dims=(CoordNames.GENERIC_PIXEL.value,),
             )
 
+        # Add solid angle to the map dataset with dims (epoch=1, pixel=num_points)
+        healpix_skymap.data_1d["solid_angle"] = xr.DataArray(
```
tmplummer: Similar to `RectangularSkyMap`, I think that this should be done in the `HealpixSkyMap.to_dataset()` function.

tmplummer (approved):
This looks reasonable, and I am OK with this implementation if it is needed ASAP, but see comment about location of adding `solid_angle`.

pleasant-menlo (approved):
This looks reasonable to us. Thanks!

tmplummer:
@pleasant-menlo, Do you need this for SIT4? If not, I think that we can wait to merge and make the changes that I have recommended.

pleasant-menlo:
> @pleasant-menlo, Do you need this for SIT4? If not, I think that we can wait to merge and make the changes that I have recommended.

We will be okay without it; we currently have a fallback with NaN arrays.

nkerman:
I should have put it in to_dataset() in the first place, but it was very simple to move there. If @subagonsouth has no objections, I'll merge this in, but it could also wait until post SIT4, if we don't want to risk throwing anything off.

imap_processing/ultra/l2/ultra_l2.py
nkerman: Good point - it was very simple to move it there!

tmplummer (approved):
LGTM

PR #1798, opened 2025-05-30
CoDICE L1a: More ISTP fixes

bourque:
Similar to #1791 , this PR makes some more updates to CoDICE L1a products for ISTP compliance. I'm trying to split these changes into multiple PRs, so I expect one or two more of these PRs to come through (I still need to make updates for a few more products).

### DIFF
--- a/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
@@ -35,6 +35,8 @@ hskp_attrs: &hskp_default
     VAR_TYPE: support_data
 
 hi_energies_attrs: &hi_energies_default
+    DEPEND_0: epoch
+    DISPLAY_TYPE: no_plot
     FIELDNAM: Energy Table
     FILLVAL: *real_fillval
     FORMAT: F10.7
@@ -237,7 +239,6 @@ direct_events_attrs: &direct_events
     FIELDNAM: Fill in at creation
     FILLVAL: 255
     FORMAT: I5
-    LABLAXIS: unitless
     UNITS: unitless
     VALIDMIN: 0
     VALIDMAX: 10000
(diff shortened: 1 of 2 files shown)


sdhoyt (approved):
Looks good to me

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -1333,798 +1332,1036 @@ lo-nsw-species-cnoplus:
...
     FIELDNAM: Priority 0 Multi Flag
+    LABL_PTR_1: event_num
     VALIDMAX: 1
 
 P0_NumEvents:
     <<: *direct_events
     CATDESC: Priority 0 Number of Events
     FIELDNAM: Priority 0 Number of Events
     FILLVAL: 65535
+    LABLAXIS: unitless
```
greglucas: I thought we were putting `" "` for the UNITS field on things like this?

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -1333,798 +1332,1036 @@ lo-nsw-species-cnoplus:
...
+    DEPEND_1: event_num
     FIELDNAM: Priority 2 APD Gain
+    LABL_PTR_1: event_num
     VALIDMAX: 1
 
 P2_DataQuality:
     <<: *direct_events
     CATDESC: Priority 2 Data Quality
     FIELDNAM: Priority 2 Data Quality
+    LABL_PTR_1: event_num
```
greglucas: Does this need a DEPEND_1 field too? Seems odd to have a label pointer but no depends. I see this is also done in the lower one too.

greglucas (approved):

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: Yeah, good call. This is what Joey has in his validation data but I will change it to `" "` for consistency on our end.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
bourque: Good catch, thanks! This was a copy/paste overlook on my part. The `P{n}_DataQuality` and `P{n}_NumEvents` variables don't have a `DEPEND_1` (and thus not a `LABL_PTR_1` either).

tech3371 (approved):

PR #1800, opened 2025-05-31
Lo L2 - Added L2 CLI processing and ISTP fixes

### LINKED ISSUE #1799: Lo L2 - Add L2 processing to CLI
Add L2 processing function to the CLI so L2 processing is run when batch starter triggers the job.

sdhoyt:
# Change Summary

## Overview
This PR adds L2 processing to the Lo CLI and makes ISTP fixes

## Updated Files
- imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
   - changed to do a 3mo map for SIT-4
- imap_processing/cli.py
   - added L2 processing
   - created a new function to handle adding the attributes to the map
   - removed the dimension transposing which is no longer needed. The dimension ordering was fixed in L1C in a previous PR.
   
## Testing
- Added unit test for attribute function
- Updated existing L2 unit tests to account for dimension ordering fix in previous PR

Closes #1799


laspsandoval (approved):
Looks good.

Issue #405, opened 2024-04-18
Produce CoDICE L1B CDFs


bourque:
- [ ] Convert data arrays to rates for CoDICE-lo
- [ ] Convert data arrays to rates for CoDICE-hi
- [ ] Convert data arrays to rates for Event Data
- [ ] Convert data arrays to rates for i-alirt Data

PR #1806, opened 2025-06-03
Finish implementing CoDICE L1b for SIT-4

bourque:
This PR finishes "implementing" CoDICE L1b products. "Implementing" is in quotes here because these products are not implemented in the sense of being ready to be validated. There are still some fixes to be made in L1a processing (e.g. centering epoch time bins, updated compression lookup tables, etc.).  For the purposes of creating ISTP compliant CDFs for SIT-4, this L1b pipeline currently just divides data arrays by a float to convert from counts to rates.

There were also some leftover ISTP compliance issues in L1a that I also fixed in this PR, so there are some changes to the L1a pipeline as well.

Pertains to #405


imap_processing/cdf/config/imap_codice_l1b_variable_attrs.yaml
bourque, line 1: This file does not need to be reviewed throroughly, because the vast majority of it is just copy/pasted from the L1a variable attributes YAML file. The only differences is that "Counts" labels are not "Rates", and some of the `FORMAT` values are set to floats instead of ints.

imap_processing/tests/codice/conftest.py
```diff
@@ -15,22 +15,25 @@
...
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_hskp_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-counters-aggregated_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-counters-singles_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-ialirt_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-nsw-angular_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-nsw-priority_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-nsw-species_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-sw-angular_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-sw-priority_20241110_v999.cdf",
+    TEST_L1A_DATA_PATH / "imap_codice_l1a_lo-sw-species_20241110_v999.cdf",
```
bourque: Just putting things in alphabetical order here

imap_processing/codice/codice_l1b.py
```diff
@@ -22,6 +26,67 @@
...
+        "lo-sw-angular",
+        "lo-nsw-priority",
+        "lo-sw-priority",
+        "lo-nsw-species",
+        "lo-sw-species",
+        "lo-ialirt",
+    ]:
+        rates_data = dataset[variable_name].data / (
+            acq_times
+            * 10e-6
```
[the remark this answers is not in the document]
greglucas: Agreed, or add a comment indicating the units of `acq_times` and your factor here.

Do you mean `1e-6` here too? i.e. is it a conversion of microseconds to seconds, or 10 microseconds to seconds.

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -205,7 +205,8 @@ def test_l1a_logical_sources(test_l1a_data, index):
     expected_logical_source = f"imap_codice_l1a_{DESCRIPTORS[index]}"
 
     # Write the dataset to a file to set the logical source attribute
-    _ = write_cdf(processed_dataset)
+    filename = write_cdf(processed_dataset)
+    logger.warning(f"Wrote CDF file to {filename}")
```
greglucas, line 209: I dislike logs/prints in tests in general because they spam the output, but I think this should at least be an info level instead of a warning level if you are going to add it.
```suggestion
    logger.info(f"Wrote CDF file to {filename}")
```

bourque: I will be removing this -- I am just using it to debug some GitHub Actions build errors.

bourque: @greglucas Speaking of which -- I was hoping to use the L1a CDFs generated by my tests to be used as input files for L1b testing, but I can't figure out where exactly in the GitHub Actions directory structure these L1a files are being written to. And I can't seem to get the `logger.info` or `logger.warning` messages (which I was hoping would tell me where the file is being written to) to even appear in the build output.  Do you have any insight on how I could figure this out?

greglucas: They should be created in temporary directories (We set IMAP_DATA_DIR to a temp path before every single test for isolation). This pattern will only work if you create them within the same test and not across test functions because the cleanup happens at the end of every test to consider them isolated.

bourque: Thanks, good to know. Sounds like I should go down the 'download from S3' route for these then.

imap_processing/codice/codice_l1b.py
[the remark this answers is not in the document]
bourque: Thanks, I will add a note. I am indeed trying to convert from microseconds to seconds here.

greglucas (approved):

PR #1808, opened 2025-06-03
Tiny bugfix: fix energy independent `obs_date_range` metadata

nkerman:
# Change Summary

## Overview
Small bugfix: In a recent PR, the default was changed so that variables are expected to be energy dependent. The obs_date_range metadata needs to be slightly updated to work with this new organization.


laspsandoval (approved):
Looks good.

tmplummer (approved):
Thanks for cleaning that up.

Issue #873, opened 2024-09-24
Produce CoDICE L2 CDFs


bourque:
- [ ] Implement Lo Species Intensities processing
- [ ] Implement Lo Angular Intensities processing
- [ ] Implement Hi Omni-directional Intensities processing
- [ ] Implement Lo Sectored Intensities processing

PR #1810, opened 2025-06-03
Initial CoDICE L2 processing for SIT-4

bourque:
This PR implements a very rudimentary initial L2 pipeline for CoDICE. It is basically the bare-bones minimum needed to produce an ISTP-compliant L2 data product for lo- and hi- direct event data (these products are used as input for L3).

In order to pass the ISTP compliance tests, I also had to fix some lingering issues in L1A and L1B, so there are a few changes for those level sprinkled into this PR, but nothing major.

Pertains to #873


imap_processing/cli.py
```diff
@@ -635,6 +635,16 @@ def do_processing(
             science_files = dependencies.get_file_paths(source="codice")
             datasets = [codice_l1b.process_codice_l1b(science_files[0])]
 
+        if self.data_level == "l2":
+            if len(dependency_list) > 1:
```
```suggestion
            if len(dependency_list) != 1:
```

imap_processing/codice/codice_l2.py
```diff
@@ -0,0 +1,74 @@
...
+    ----------
+    file_path : pathlib.Path
+        Path to the CoDICE L1 file to process.
+
+    Returns
+    -------
+    l2_dataset : xarray.Dataset
+        The``xarray`` dataset containing the science data and supporting metadata.
+    """
+    logger.info(f"\nProcessing {file_path}")
```
greglucas: Usually this starts on the same line as the potential timestamps etc. do you want the explicit newline here?
```suggestion
    logger.info(f"Processing {file_path}")
```

imap_processing/codice/codice_l2.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # Open the l1 file
+    l1_dataset = load_cdf(file_path)
+
+    # Use the logical source as a way to distinguish between data products and
+    # set some useful distinguishing variables
+    dataset_name = l1_dataset.attrs["Logical_source"]
+    data_level = dataset_name.removeprefix("imap_codice_").split("_")[0]
+    descriptor = dataset_name.removeprefix(f"imap_codice_{data_level}_")
+    dataset_name = dataset_name.replace(data_level, "l2")
```
greglucas: Rather than all of this parsing I think you could take advantage of the objects in imap-data-access like `ScienceFilePath(file_path)` where you could then do something like: `filepath.descriptor`. Or maybe those are on ProcessingInput instead.

imap_processing/codice/codice_l2.py
```diff
@@ -0,0 +1,74 @@
...
+    # Get the L2 CDF attributes
+    cdf_attrs = ImapCdfAttributes()
+    cdf_attrs.add_instrument_global_attrs("codice")
+    cdf_attrs.add_instrument_variable_attrs("codice", "l2")
+
+    # Update the global attributes
+    l2_dataset.attrs = cdf_attrs.get_global_attributes(dataset_name)
+
+    # TODO: Add L2-specific algorithms/functionality here. For SIT-4, we can
+    #       just keep the data as-is.
```
greglucas, line 71: It might be nice to change all of the l2_dataset variables to nan's or fill values to indicate it is incomplete data and not potentially produce values that aren't quite right here.

Are the variables all consistent between l1b and l2, just different factors/calibrations applied, or are they a completely different shape and meaning in l2? I am mostly getting at whether this is all even worth doing here just to get a product produced.

greglucas (commented):
Have you added this L2 product to the `sds-data-manager` dependencies so it would get triggered?

imap_processing/codice/codice_l2.py
bourque: Yeah that's a good idea. I will add a TODO for this and tackle it in the future.

imap_processing/codice/codice_l2.py
bourque: I believe the variables are all consistent, with just factors/calibrations applied.

I was just hoping to get these (totally not valid) L2 products in before SIT-4 so we can show that our infrastructure can go from L1 -> L2 -> L3 for at least the `direct-event` data products.

bourque:
> Have you added this L2 product to the `sds-data-manager` dependencies so it would get triggered?

Yes, just took care of that in https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/674

docs/source/code-documentation/codice.rst
```diff
@@ -19,6 +19,7 @@ products can be found below:
     codice_l0
     codice_l1a
     codice_l1b
+    codice_l2
```
lacoak21, line 22: nice - always forget this

imap_processing/cdf/config/imap_codice_l2_variable_attrs.yaml
```diff
@@ -0,0 +1,994 @@
...
+    DEPEND_1: event_num
+    FIELDNAM: Priority 1 APD Gain
+    LABL_PTR_1: event_num_label
+    VALIDMAX: 1
+
+P1_DataQuality:
+    <<: *direct_events
+    CATDESC: Priority 1 Data Quality
+    FIELDNAM: Priority 1 Data Quality
+    LABLAXIS: " "
```
lacoak21, line 209: Is this LABAXIS mean to be blank?

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -150,6 +150,13 @@ esa_step_label:
     FORMAT: A3
     VAR_TYPE: metadata
 
+event_num_label:
+    CATDESC: Event Number
+    DEPEND_1: event_num
```
tech3371, line 155: I don't think you need `DEPEND_1` in label. It could be from not turning `check_schema=False`. Some of us had that issue. But looks like you did set it to False. May be, we can revisit that after SIT.

imap_processing/codice/codice_l2.py
```diff
@@ -0,0 +1,75 @@
...
+        The``xarray`` dataset containing the science data and supporting metadata.
+    """
+    logger.info(f"Processing {file_path}")
+
+    # Open the l1 file
+    l1_dataset = load_cdf(file_path)
+
+    # Use the logical source as a way to distinguish between data products and
+    # set some useful distinguishing variables
+    # TODO: Could clean this up by using imap-data-access methods?
```
lacoak21, line 45: I think you could do something like 

```
input_obj = imap_data_access.processing_input.generate_imap_input(file_path)
data_level = file_obj.data_type
descriptor = file_obj.descriptor
```

tech3371 (approved):
It looks good to me on high-level. Had one comment that we can revisit after SIT-4.

lacoak21 (approved):
LGTM!

imap_processing/cdf/config/imap_codice_l2_variable_attrs.yaml
bourque: Probably should be something other than blank, but I will cross that bridge when I validate this / talk to the CoDICE instrument team about it.

imap_processing/codice/codice_l2.py
bourque: Thanks! I will add this in a future PR. I think I have some code in the L1b pipeline that could use this too.

PR #1811, opened 2025-06-03
Fix GLOWS CLI to work with extra input files

maxinelasp:
# Change Summary
Avoid the crash that occurred in today's dry run and limit the CLI to expecting only one science file.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -661,33 +661,32 @@ class Glows(ProcessInstrument):
         print(f"Processing GLOWS {self.data_level}")
         datasets: list[xr.Dataset] = []
 
-        dependency_list = dependencies.processing_input
         if self.data_level == "l1a":
-            if len(dependency_list) > 1:
-                raise ValueError(
-                    f"Unexpected dependencies found for GLOWS L1A:"
-                    f"{dependency_list}. Expected only one input dependency."
-                )
             science_files = dependencies.get_file_paths(source="glows")
+            if len(science_files) != 1:
+                raise ValueError(
+                    f"GLOWS L1A requires exactly one input science file, received: "
+                    f"{science_files}."
+                )
             datasets = glows_l1a(science_files[0])
 
         if self.data_level == "l1b":
-            if len(dependency_list) > 1:
-                raise ValueError(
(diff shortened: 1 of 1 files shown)


imap_processing/cli.py
```diff
@@ -661,33 +661,32 @@ def do_processing(
         print(f"Processing GLOWS {self.data_level}")
         datasets: list[xr.Dataset] = []
 
-        dependency_list = dependencies.processing_input
         if self.data_level == "l1a":
-            if len(dependency_list) > 1:
+            science_files = dependencies.get_file_paths(source="glows")
+            if len(science_files) != 1:
```
tech3371, line 666: I am looking at dependency_config.csv and you might want to update this to be 2 since you are receiving time kernels too.

imap_processing/cli.py
```diff
@@ -661,33 +661,32 @@ def do_processing(
...
+                    f"{science_files}."
                 )
-            science_files = dependencies.get_file_paths(source="glows")
             input_dataset = load_cdf(science_files[0])
             datasets = [glows_l1b(input_dataset)]
 
         if self.data_level == "l2":
-            if len(dependency_list) > 1:
+            science_files = dependencies.get_file_paths(source="glows")
+            if len(science_files) != 1:
```
tech3371, line 685: This one, I am unsure since I don't see that in the dep config file.

imap_processing/cli.py
maxinelasp: This is filtering only by the GLOWS files, so there shouldn't be any spice files included in `science_files` (since those have a source of `spice`)

imap_processing/cli.py
maxinelasp: Oh wow I am missing GLOWS L2 from the dependency config. I'll add it. It does have some SPICE dependencies but I won't be including them in the SIT. But, this specific variable should only have science files.

tmplummer (approved):
LGTM

imap_processing/cli.py
tech3371: Ah true!

tech3371 (approved):

PR #1812, opened 2025-06-04
Fix CoDICE CLI to work with extra input files

bourque:
Similar to #1811, this PR updates the CoDICE CLI to expect one _science_ file (instead of one dependency), which hopefully will solve issues encountered during the SIT-4 dry run.


tmplummer (approved):
Cross checked against `sds-data-manager` dependencies_config.csv. LGTM!

bourque:
Thanks @subagonsouth!

PR #1813, opened 2025-06-04
Bugfixes for delta_energy with epoch as first dim

nkerman:
# Change Summary

In pre-SIT4 testing, found bug with ena_intensity calculation. Coming from an unexpected "epoch" dim on the energy_delta

Also fix metadata bug


laspsandoval (approved):
Looks good.

PR #1816, opened 2025-06-04
CoDICE L2 lo/hi direct-event SPASE metadata

bourque:
This PR adds SPASE metadata `DICT_KEY` attributes to CoDICE L2 `direct-event` data products. I wasn't sure what to chose for these, but I think this will work for the time being, until I can get a better understanding on how to decide on these.


tech3371 (approved):

tech3371:
merging on Matthew's request.

PR #1818, opened 2025-06-05
Lo and HIT - SIT-4 Bug fixes

sdhoyt:
# Change Summary

## Overview
These are (hopefully) the final bug fixes for Lo and HIT for SIT-4

# Updated Files
- imap_processing/cli.py
   - Fixed dependency check issues for HIT that were caused by the new addition of SPICE dependencies
   - Similar fixes for Lo and updates to handle ancillary files
   - added logs
   - added placeholder value for epoch in function instead of spice call that was causing issues. The function needs updating in the future and is currently not being called. Will find a better solution for this after SIT-4

# Testing
Ran this is the dev account and all Lo and HIT files were produced


imap_processing/cli.py
```diff
@@ -791,25 +791,23 @@ def do_processing(
...
-            if len(dependency_list) > 1:
-                raise ValueError(
-                    f"Unexpected dependencies found for HIT L1B:"
-                    f"{dependency_list}. Expected only one dependency."
-                )
             data_dict = {}
-            # TODO: Check this and update with new features as needed.
+            # TODO: Sean removed the file number error handling to work with the
+            #  new SPICE dependencies for SIT-4. Need to review and make changes
+            #  if needed.
```
greglucas, line 808: In IMAP-Lo below you add the above check below the l0_files rather than to the initial dependencies. That makes sense to me, but this all seems like we should really rethink how we handle this later too. There were a lot of issues with these checks it looked like, so how much value do they add? If they do add value, can we somehow consolidate them better for all instruments to take advantage of rather than individually like this.

These are just some quick comments for myself, no ned to update anything here.

greglucas (approved):

PR #1819, opened 2025-06-05
CoDICE: Some hotfixes to get things working with MSIM3 data and SIT-4

bourque:
This PR makes a few small changes to get the L1a, L2b, and L2 pipelines working with CoDICE MSIM3 data, to be used in SIT-4.


lacoak21 (approved):

tech3371 (approved):
don't know enough but since it works, it looks good to me :)

PR #1827, opened 2025-06-13
HIT L1B - Handle data gaps in sectored counts

### LINKED ISSUE #1826: Add handling for data gaps in sectored counts
The `subset_data_for_sectored_counts` function that filters for complete sets of sectored counts data (i.e. major frame = 10 science frames = 10 minutes of data) only excludes partial major frames at the beginning and end of the dataset, but doesn't exclude partial major frames in the middle of the dataset due to the use of slicing when filtering the data. 

Slicing uses an index range. To exclude data gaps in the middle of the dataset, the dataset needs to be filtered by providing an array of valid indices instead.

vmartinez-cu:
This PR updates the `subset_data_for_sectored_counts` function to properly filter out all partial major frames in a dataset. A major frame = a set of complete sectored counts, which spans 10 minutes

Previously, the function was only filtering out partial major frames found at the beginning and end of the dataset since it was filtering data by slicing which includes partial major frames within the slice range. 

The code updates replace filtering by slicing with filtering by an array of exact indices needed, resulting in a dataset with only complete major frames of sectored counts data.

Closes #1826 

## Updated Files
- imap_processing/hit/l1b/hit_l1b.py
   - Replace filtering by slicing with filtering by an array of valid indices
   - Fix error in filtering start_indices to include index that is equal to the bin_size of 10

The unit test for the function was updated to handle different cases of partial data in a dataset

### DIFF
--- a/imap_processing/hit/l1b/hit_l1b.py
+++ b/imap_processing/hit/l1b/hit_l1b.py
@@ -403,13 +403,20 @@ def subset_data_for_sectored_counts(
     else:
         start_indices = start_indices[start_indices >= bin_size]
 
-    # Subset data to include only complete sets of sectored counts
+    # Subset data for complete sets of sectored counts.
+    # Each set of sectored counts is 10 minutes long, so we take the indices
+    # starting from the start indices and extend to the bin size of 10.
+    # This creates a 1D array of indices that correspond to the complete
+    # sets of sectored counts which is used to filter the L1A dataset and
+    # create the L1B sectored rates dataset.
     data_indices = np.concatenate(
         [np.arange(idx, idx + bin_size) for idx in start_indices]
     )
     l1b_sectored_rates_dataset = l1a_counts_dataset.isel(epoch=data_indices)
 
-    # Subset livetime staggered from sectored counts by 10 minutes
+    # Subset livetime values corresponding to the previous 10 minutes
+    # for each start index. This ensures the livetime data aligns correctly
+    # with the sectored counts for rate calculations.
     livetime_indices = np.concatenate(
         [np.arange(idx - bin_size, idx) for idx in start_indices]
     )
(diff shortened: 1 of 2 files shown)


imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -392,16 +393,27 @@ def subset_data_for_sectored_counts(
...
+            "No data to process - valid start indices not found for "
+            "complete sectored counts."
+        )
+        raise ValueError("No valid start indices found for complete sectored counts.")
+    else:
+        start_indices = start_indices[start_indices >= bin_size]
 
     # Subset data to include only complete sets of sectored counts
-    l1b_sectored_rates_dataset = l1a_counts_dataset.isel(epoch=data_slice)
+    data_indices = np.concatenate(
```
sdhoyt, line 412: Can you add a little more of an explanation for what this is doing and why?

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -392,16 +393,27 @@ def subset_data_for_sectored_counts(
...
+        [np.arange(idx, idx + bin_size) for idx in start_indices]
```
sdhoyt, line 413: If a major frame is missing the 4th science frame, would that still be included in this since `np.arange()` is using the start index and assuming it's that index plus the bin_size?

sdhoyt (approved):
Thanks for explaining the start_indicies! Nice work!

imap_processing/hit/l1b/hit_l1b.py
vmartinez-cu: `start_indices` only contain the start indices of complete major frames. In that case, a missing 4th frame would not have a corresponding start index in the start_indices array

PR #1828, opened 2025-06-13
Improved external test data file handling

### LINKED ISSUE #1537: Store/download CoDICE validation files from S3
Now that we have the mechanics in place to download files from S3 for individual tests during PR builds, I should move all of the CoDICE validation files to S3 and use that process, instead of storing them all in the repository.

bourque:
This PR changes how external test data files are defined -- I moved the list (which has grown quite large and may continue to grow) to its own file which is then imported into `conftest.py`. I also added some CoDICE validation files so that they can be stored in S3 instead of within the repo.

Closes #1537


tech3371 (approved):
LGTM

imap_processing/tests/external_test_data_config.py
```diff
@@ -0,0 +1,67 @@
+from imap_processing import imap_module_directory
+
+# ruff: noqa: E501
+EXTERNAL_TEST_DATA = [
+
+    # CoDICE
+    ("imap_codice_l0_raw_20241110_v001.pkts", imap_module_directory / "tests" / "codice" / "data" / "imap_codice_l0_raw_20241110_v001.pkts"),
```
tmplummer: Overall, this looks like a nice improvement. I'm wondering if it would be worth taking it a step further to make the tuple for a single file contain `(<s3_filename>, <parent_test_dir_relative_to_imap_module_dir>)`. That would clean up this configuration by removing the `imap_module_directory` as well as the target path filename. The later part would require that the filename remains the same when downloaded.

For this line, it would look like:
```suggestion
    ("imap_codice_l0_raw_20241110_v001.pkts", "tests/codice/data"),
```

tmplummer (approved):
I am happy with this as is, but made one recommendation for a slight tweak to further reduce clutter in the new config file.

imap_processing/tests/external_test_data_config.py
bourque: Good idea, I like it.

PR #1832, opened 2025-06-17
Changes to CoDICE I-ALiRT processing in prepration for I-ALiRT SIT

bourque:
This PR makes various changes to the CoDICE I-ALiRT processing so that the code works on I-ALiRT SIT input test data and creates a mock CoDICE I-ALiRT dataset.

### DIFF
--- a/imap_processing/cdf/config/imap_ialirt_l1_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_ialirt_l1_variable_attrs.yaml
@@ -76,7 +76,7 @@ codicehi_h:
   <<: *default_float32
   CATDESC: H intensities in 15 energy ranges and binned into 4 azimuths and 4 spin angle bins
   FIELDNAM: H intensities
-  LABLAXIS: H
+  LABLAXIS: Diff. Intensity
   UNITS: "# / cm2-sr-s- MeV"
   VALIDMIN: 0
   VALIDMAX: 100000000.0
@@ -84,8 +84,8 @@ codicehi_h:
 codicelo_c_over_o_abundance:
   <<: *default_float32
   CATDESC: C/O abundance ratio
-  FIELDNAM: C/O abundance ratio
-  LABLAXIS: C/O
+  FIELDNAM: C/O
+  LABLAXIS: ratio
   UNITS: " "
   VALIDMIN: 0
   VALIDMAX: 100000000.0
@@ -93,8 +93,8 @@ codicelo_c_over_o_abundance:
 codicelo_mg_over_o_abundance:
   <<: *default_float32
   CATDESC: Mg/O abundance ratio
-  FIELDNAM: Mg/O abundance ratio
-  LABLAXIS: Mg/O
+  FIELDNAM: Mg/O
+  LABLAXIS: ratio
   UNITS: " "
   VALIDMIN: 0
   VALIDMAX: 100000000.0
@@ -102,8 +102,8 @@ codicelo_mg_over_o_abundance:
 codicelo_fe_over_o_abundance:
   <<: *default_float32
   CATDESC: Fe/O abundance ratio
-  FIELDNAM: Fe/O abundance ratio
-  LABLAXIS: Fe/O
+  FIELDNAM: Fe/O
+  LABLAXIS: ratio
   UNITS: " "
   VALIDMIN: 0
   VALIDMAX: 100000000.0
@@ -111,8 +111,8 @@ codicelo_fe_over_o_abundance:
 codicelo_c_plus_6_over_c_plus_5_ratio:
   <<: *default_float32
   CATDESC: C+6/C+5 charge state ratio
-  FIELDNAM: C+6/C+5 charge state ratio
-  LABLAXIS: C+6/C+5
+  FIELDNAM: C+6/C+5
+  LABLAXIS: ratio
   UNITS: " "
   VALIDMIN: 0
   VALIDMAX: 100000000.0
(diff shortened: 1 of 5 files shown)


imap_processing/codice/codice_l1a.py
```diff
@@ -1003,72 +1021,79 @@ def create_ialirt_dataset(apid: int, packets: xr.Dataset) -> xr.Dataset:
...
+            # lo-sw-species data product
+            pipeline = CoDICEL1aPipeline(
+                metadata_values["TABLE_ID"][0],
+                metadata_values["PLAN_ID"][0],
+                metadata_values["PLAN_STEP"][0],
+                metadata_values["VIEW_ID"][0],
+            )
+            pipeline.set_data_product_config(apid, packets)
+            pipeline.decompress_data(science_values)
+            pipeline.reshape_data()
```
bourque, line 1063: Nothing much is changing here, I just put this all in an `if`/`else` statement to check if there are grouped datasets, and if there are none, throw a warning.

imap_processing/cdf/config/imap_ialirt_l1_variable_attrs.yaml
```diff
@@ -72,6 +72,69 @@ default_float64_attrs: &default_float64
   VALIDMAX: 1.7976931348623157e+308
   dtype: float64
 
+codicehi_h:
```
laspsandoval, line 75: Keep in mind that each of these variables will be a single column in the database. So there will not be dimensions here. For SWE, for example, I have a separate variable for each energy level. Does this need to be broken down further?

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -41,5 +50,39 @@ def process_codice(dataset: xr.Dataset) -> list[dict]:
...
+            "apid": int(dataset.pkt_apid[epoch].data),
+            "met": met,
+            "utc": utc,
+            "ttj2000ns": ttj2000ns,
+        }
+
+        # Add in CoDICE-Lo specific data
+        cod_lo_epoch_data = epoch_data.copy()
+        for field in constants.CODICE_LO_IAL_DATA_FIELDS:
+            cod_lo_epoch_data[f"codicelo_{field}"] = -1.0e31
```
laspsandoval: Consider putting the constant at the top:
FILLVAL_FLOAT32 = -1.0e31

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -31,8 +36,12 @@ def process_codice(dataset: xr.Dataset) -> list[dict]:
...
     - Calculate the public data products
     """
-    apid = dataset.pkt_apid.data[0]
-    codice_data = create_ialirt_dataset(apid, dataset)
+    # For I-ALiRT SIT, the test data being used has all zeros and thus no
+    # groups can be found, thus there is no data to process
+    # TODO: Once I-ALiRT test data is acquired that actually has data in it,
+    #       this can be turned back on
+    # codicelo_data = create_ialirt_dataset(CODICEAPID.COD_LO_IAL, dataset)
+    # codicehi_data = create_ialirt_dataset(CODICEAPID.COD_HI_IAL, dataset)
```
laspsandoval, line 47: Right after launch we will immediately begin to receive I-ALiRT packets that are empty. So we might have this scenario in real-life. Something to consider maybe in another PR, though, if it is too complicated to change now.

imap_processing/ialirt/l0/process_codice.py
laspsandoval: Because this is a float it has to be modified in order to insert into DynamoDB. This seemed to work for SWE:

Decimal(str( -1.0e31))

laspsandoval (approved):
Approving. I think once you have addressed the comments it is good to start testing.

imap_processing/cdf/config/imap_ialirt_l1_variable_attrs.yaml
bourque: Good question/good point -- For this particular data variable, there are 4 dimensions: `epoch`, `energy` (there are 15 levels), `ssd_index` (there are 4 of these), and `spin_sector` (there are 4 of these). So the dimensions of the data are (<# epochs>, 15, 4, 4). 

I am not sure how to break this down. Would this mean I need to make (`15*4*4=`) 240 separate variables?

imap_processing/ialirt/l0/process_codice.py
bourque: Is there a preferred way to handle empty packets? i.e. in that case, should the code return a completely empty list? Or a list with dicts that have all the expected fields but are empty?

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -41,5 +53,39 @@ def process_codice(dataset: xr.Dataset) -> list[dict]:
...
+        sc_sclk_sec = dataset.sc_sclk_sec.data[epoch]
+        sc_sclk_sub_sec = dataset.sc_sclk_sub_sec.data[epoch]
+        met = calculate_time(sc_sclk_sec, sc_sclk_sub_sec, 256)
+        utc = met_to_utc(met).split(".")[0]
+        ttj2000ns = int(met_to_ttj2000ns(met))
+
+        epoch_data = {
+            "apid": int(dataset.pkt_apid[epoch].data),
+            "met": met,
+            "utc": utc,
```
laspsandoval:
```suggestion
            "met_to_utc": utc,
```

laspsandoval (approved):
One more thing to add based on recent changes to db.

PR #1837, opened 2025-06-23
HIT L0 - Fix angle dimension order error

### LINKED ISSUE #1836: Transpose angle dimensions
Fix error identified during validation in the sectored products related to angle dimension ordering.

The order of the data is defined in the flight software as 8x15 (zenith x azimuth). It was later requested to flip this order to maintain consistency among the instruments. This was done by shaping the data as 15x8 at the very start when decommutating the raw binary data. However, since the data comes out of the flight software, which uses the original order, it needs to be processed in the original order (8x15) then transposed to 15x8 for structure in the CDFs. This ensures that the data end up where expected in the multidimensional data array.

vmartinez-cu:
This PR updates how the raw binary sectorates data is processed. It needs to be processed with a shape of 8x15 and then transposed to 15x8 to ensure that the data ends up in the correct spots in the data array. I was shaping the data as 15x8 from the start, but since the data comes out of flight software and it has the opposite order defined, this was not producing expected results during validation.

Closes #1836 

## Updated Files
- imap_processing/hit/l0/constants.py
   - Changed shape definition back to 8x15 for sectorates
- imap_processing/hit/l0/decom_hit.py
   - Transposed decommutated sectorates data to 15x8
- imap_processing/hit/l1a/hit_l1a.py
   - Replaced hard coded values for dimension sizes with length of constants for those dimensions

### DIFF
--- a/imap_processing/hit/l0/decom_hit.py
+++ b/imap_processing/hit/l0/decom_hit.py
@@ -92,8 +92,9 @@ def parse_count_rates(sci_dataset: xr.Dataset) -> None:
         # Get dims for data variables (yaml file not created yet)
         if len(field_meta.shape) > 1:
             if "sectorates" in field:
-                # Reshape data to 15x8 for azimuth and zenith look directions
+                # Transpose data to 15x8 for azimuth and zenith look directions
                 parsed_data = np.array(parsed_data).reshape((-1, *field_meta.shape))
+                parsed_data = np.transpose(parsed_data, axes=(0, 2, 1))
                 dims = ["epoch", "azimuth", "zenith"]
                 # Add angle values to coordinates
                 sci_dataset.coords["zenith"] = xr.DataArray(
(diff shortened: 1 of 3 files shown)


imap_processing/hit/l0/decom_hit.py
```diff
@@ -92,8 +92,11 @@ def parse_count_rates(sci_dataset: xr.Dataset) -> None:
         # Get dims for data variables (yaml file not created yet)
         if len(field_meta.shape) > 1:
             if "sectorates" in field:
-                # Reshape data to 15x8 for azimuth and zenith look directions
+                # Reshape data into (num_frames, 8, 15) for zenith and azimuth
+                # look directions
                 parsed_data = np.array(parsed_data).reshape((-1, *field_meta.shape))
+                # Transpose data to (num_frames, 15, 8) for flipped look directions
+                parsed_data = np.transpose(parsed_data, axes=(0, 2, 1))
```
sdhoyt, line 99: What would happen if you left L0 as (15, 8) and didn't transpose it here? Would that cause any issues?

vmartinez-cu: The values end up in a different arrangement in the data array which then fails HIT's validation tests. I guess transposing data is different from reshaping data and I didn't realize that before

sdhoyt: I just realized you added the transpose here as part as part of this change, so that was a dumb question

sdhoyt (approved):
Nice work!

PR #1843, opened 2025-06-24
Mock CoDICE I-ALiRT data now returns empty lists instead of a FILLVAL

bourque:
This change is requested for the I-ALiRT SIT


laspsandoval (approved):
Looks good. Thank you!

PR #1848, opened 2025-06-25
Mag l1a dupe packets

maxinelasp:
# Change Summary
Remove duplicate packets from MAG if they have the same SHCOARSE, SRC_SQ_CTR, and APID. 

Fixes a bug found in L1a. Also adds an error throw if the data is non-monotonically increasing, so bugs like this are easier to find (as MAG is prone to them.)

Note that we don't expect to need this in prod, but most other instruments do it in the packets_to_dataset function. 

## Overview

Added a step which adds the MagL0 object to a set and confirms it is unique. This required the __eq__ and __hash__ functions on the dataclass.

### DIFF
--- a/imap_processing/mag/l0/decom_mag.py
+++ b/imap_processing/mag/l0/decom_mag.py
@@ -43,7 +43,7 @@ def decom_packets(packet_file_path: str | Path) -> dict[str, list[MagL0]]:
 
     packet_definition = definitions.XtcePacketDefinition(xtce_document)
 
-    # We use a set for de-duplication, but we also keep a list to keep objects in order.
+    # Store in a dict for de-duplication. Only the keys are returned as a list.
     norm_dict: dict[MagL0, None] = {}
     burst_dict: dict[MagL0, None] = {}
 
--- a/imap_processing/mag/l0/decom_mag.py
+++ b/imap_processing/mag/l0/decom_mag.py
@@ -44,10 +44,8 @@ def decom_packets(packet_file_path: str | Path) -> dict[str, list[MagL0]]:
     packet_definition = definitions.XtcePacketDefinition(xtce_document)
 
     # We use a set for de-duplication, but we also keep a list to keep objects in order.
-    norm_set: set[MagL0] = set()
-    burst_set: set[MagL0] = set()
-    norm_list = []
-    burst_list = []
+    norm_dict: dict[MagL0, None] = {}
+    burst_dict: dict[MagL0, None] = {}
 
     with open(packet_file_path, "rb") as binary_data:
         mag_packets = packet_definition.packet_generator(binary_data)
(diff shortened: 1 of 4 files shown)


imap_processing/mag/l0/mag_l0_data.py
```diff
@@ -116,3 +119,46 @@ def __post_init__(self) -> None:
 
         self.PRI_VECSEC = 2**self.PRI_VECSEC
         self.SEC_VECSEC = 2**self.SEC_VECSEC
+
+    def __eq__(self, other: object) -> bool:
```
greglucas, line 123: Can't you set `eq=True` in the dataclass decorator to get this automatically for you?

imap_processing/tests/mag/test_mag_decom.py
```diff
@@ -118,3 +122,28 @@ def test_mag_raw_cdf_generation(cdf_attrs):
 
     input_xarray = load_cdf(output)
     assert input_xarray.attrs.keys() == burst_data.attrs.keys()
+
+
+def test_comparison():
```
greglucas, line 127: Would it make sense to also add a test that loads a ccsds file twice so that you know you are actually pruning duplicate packets? I think this is a nice unit test, but might also be nice to add the larger integration to show it is doing what you want and being used.

I pushed an example up here https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1850

imap_processing/mag/l0/decom_mag.py
```diff
@@ -43,8 +43,11 @@ def decom_packets(packet_file_path: str | Path) -> dict[str, list[MagL0]]:
 
     packet_definition = definitions.XtcePacketDefinition(xtce_document)
 
-    norm_data = []
-    burst_data = []
+    # We use a set for de-duplication, but we also keep a list to keep objects in order.
```
greglucas: This seems like a lot of workaround by adding two objects to keep track of things throughout. Can you use a dict which is guaranteed to preserve order instead so you only have one object to keep track of?

```python
norm_dict = {}
norm_dict[mag_l0] = None

# if you want a list at the end
return list(norm_dict.keys())
```

imap_processing/mag/l0/mag_l0_data.py
maxinelasp: Since I was only comparing SHCOARSE, APID, and SQ_CTR I thought this would be more reliable. I think your idea would probably also require freezing the CcsdsHeader class, which Lo uses as well.

imap_processing/tests/mag/test_mag_decom.py
maxinelasp: Sure, I can do that

imap_processing/mag/l0/decom_mag.py
maxinelasp: Sure, that seems like a good solution

imap_processing/mag/l0/decom_mag.py
```diff
@@ -43,8 +43,9 @@ def decom_packets(packet_file_path: str | Path) -> dict[str, list[MagL0]]:
...
```
greglucas: Need to update this comment to be a dict for de-duplication now.

greglucas (approved):

PR #1850, opened 2025-06-26
FIX: Remove duplicate packets from dataset

greglucas:
# Change Summary

## Overview

Our packet files can contain duplicates if the database we are drawing from does not remove the duplicates for us. We can remove the duplicate SHCOARSE values from each apid to make sure this does not happen.

### DIFF
--- a/imap_processing/tests/test_utils.py
+++ b/imap_processing/tests/test_utils.py
@@ -222,6 +222,35 @@ def test_packet_file_to_datasets(use_derived_value, expected_mode):
     np.testing.assert_array_equal(np.unique(data["mode"].data), expected_mode)
 
 
+def test_packet_file_to_datasets_duplicates(tmpdir, caplog):
+    """
+    Test that all datatypes aren't all int64 and that we get
+    uint8/uint16 from header items as expected.
+
+    Test that we get multiple apids in the output.
+    """
+    test_file = "tests/swapi/l0_data/imap_swapi_l0_raw_20240924_v001.pkts"
+    packet_file = imap_module_directory / test_file
+
+    # Write the file out twice to double the number of binary packets in
+    # a new file for testing
+    with open(two_files := tmpdir / "two_files.pkts", "wb") as f:
+        with open(packet_file, "rb") as original_file:
+            data = original_file.read()
+            f.write(data)
+            f.write(data)
+
+    packet_definition = (
+        imap_module_directory / "swapi/packet_definitions/swapi_packet_definition.xml"
+    )
+    ds_two_files = utils.packet_file_to_datasets(two_files, packet_definition)
+    ds_one_file = utils.packet_file_to_datasets(packet_file, packet_definition)
+    assert len(ds_two_files[1188]["epoch"]) == len(ds_one_file[1188]["epoch"])
+    assert len(ds_two_files[1188]["epoch"]) == 153
+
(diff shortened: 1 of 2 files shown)


imap_processing/tests/test_utils.py
```diff
@@ -222,6 +222,35 @@ def test_packet_file_to_datasets(use_derived_value, expected_mode):
...
+
+    packet_definition = (
+        imap_module_directory / "swapi/packet_definitions/swapi_packet_definition.xml"
+    )
+    ds_two_files = utils.packet_file_to_datasets(two_files, packet_definition)
+    ds_one_file = utils.packet_file_to_datasets(packet_file, packet_definition)
+    assert len(ds_two_files[1188]["epoch"]) == len(ds_one_file[1188]["epoch"])
+    assert len(ds_two_files[1188]["epoch"]) == 153
+
+    assert "Dropping duplicate packets" in caplog.records[0].message
```
tech3371, line 251: Good to know about `caplog`!

tech3371 (approved):
It's helpful to be doing this early on at unpacking packets. This helps with SWAPI concerns. I can remove my line that drops duplicate. Is this related and resolve @maxinelasp's other PR too?

greglucas:
> This helps with SWAPI concerns. I can remove my line that drops duplicate. Is this related and resolve @maxinelasp's other PR too?

Yeah, I thought that we had this already for all instruments, but then realized we only added it for SWAPI, so I added it here instead. Can you follow up with SWAPI removal after this?

@maxinelasp is doing the iteration differently and not using packet_file_to_datasets, so the other PR is needed still to update the MAG code as well.

tech3371:
> > This helps with SWAPI concerns. I can remove my line that drops duplicate. Is this related and resolve @maxinelasp's other PR too?
> 
> Yeah, I thought that we had this already for all instruments, but then realized we only added it for SWAPI, so I added it here instead. Can you follow up with SWAPI removal after this?
> 
> @maxinelasp is doing the iteration differently and not using packet_file_to_datasets, so the other PR is needed still to update the MAG code as well.

Sure thing! I will open a PR with that line removed after this.

I see. That's a bummer. This function has been so helpful and could unify same behavior across all.

maxinelasp:
@tech3371 Unfortunately that function won't really work for MAG because I'm producing some extra files for them. I did look into refactoring but it would require a major rehaul and doesn't seem worth it to me.

imap_processing/utils.py
```diff
@@ -347,6 +347,26 @@ def packet_file_to_datasets(
             coords={"epoch": time_data},
         )
         ds = ds.sortby("epoch")
+        # We may get duplicate packets within the packet file if packets were
+        # ingested multiple times by the POC. We want to drop packets where
+        # apid, epoch, and src_seq_ctr are the same.
+
+        # xarray only supports dropping duplicates by index, so we instead go
+        # to pandas multi-index dataframe to identify the unique positions
+        unique_indices = (
```
maxinelasp, line 356: Do we want to check SHCOARSE here too?

greglucas: That is epoch at this point, just renamed. So it is checking (shcoarse, src_seq_ctr) within the apid loop already. So the same as you have in MAG.

maxinelasp: Oh I see, ok. Just didn't fully comprehend that comment. Thanks for explaining

PR #1854, opened 2025-06-26
Ultra l1a - added cdf for apid 882

laspsandoval:
# Change Summary

## Overview
Added apid 882 (energy rates)

## Updated Files
- ultra_l1a.py
   - Updated logic to include the energy rates data
- ultra_utils.py
   - Added keys and packet properties
- decom_ultra.py
   - Added function to decom the packet
   - Modified global attributes
   - Modified variable attributes

## Testing
conftest.py
test_decom_apid_882.py
test_ultra_l1a.py

### DIFF
--- a/imap_processing/ultra/l0/decom_ultra.py
+++ b/imap_processing/ultra/l0/decom_ultra.py
@@ -262,8 +262,8 @@ def process_ultra_energy_rates(ds: xr.Dataset) -> xr.Dataset:
     """
     decom_data = defaultdict(list)
 
-    for fastdata in ds["ratedata"]:
-        raw_binary_string = convert_to_binary_string(fastdata.item())
+    for rate in ds["ratedata"]:
+        raw_binary_string = convert_to_binary_string(rate.item())
         decompressed_data = decompress_binary(
             raw_binary_string,
             cast(int, ULTRA_ENERGY_RATES.width),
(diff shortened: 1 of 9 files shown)


imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -242,3 +244,38 @@ def process_ultra_rates(ds: xr.Dataset) -> xr.Dataset:
         ds[key] = xr.DataArray(np.array(values), dims=["epoch"])
 
     return ds
+
+
+def process_ultra_energy_rates(ds: xr.Dataset) -> xr.Dataset:
```
greglucas, line 249: This is very close to the above function. If you are going to be doing this on other rate-type packets too it might be good to abstract the boilerplate stuff out and pass in the apid/enum to do the lookup of the packet-specific things within a common function instead of needing to duplicate all the looping/logic.

```python
def process_rates_l0(ds, packet_enum):
    # Lookup from whatever enum you passed in
    packet_enum.width, ...
```

greglucas (approved):
The branch is called ialirt_cdf_updates, but this does look like it only contains ultra l1a 882 changes. I have a minor suggestion for refactoring the function to be more general, but that might be a good follow-up task for @lacoak21 and would get her familiar with the unpacking L0 data.

While working through these, I don't think metadata on l1a products is important, so don't feel like we need to add all the variable information right now, that can easily come later IMO. If it is easiest just to do it at the same time then that is fine but just don't think you need to do it if it is slowing everything down.

imap_processing/ultra/l0/decom_ultra.py
laspsandoval: Yes. I know. This is the last rates packet, though.

PR #1864, opened 2025-06-27
Add Lo star tracker packet processing

### LINKED ISSUE #409: Create Star Sensor L1A CDF
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

### LINKED ISSUE #410: Validate Star Sensor Data
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

### LINKED ISSUE #422: Validate Raw Star Sensor Data

### LINKED ISSUE #488: Validate Processed Star Sensor Data

### LINKED ISSUE #939: Validate Lo L1A Star Sensor Data

### LINKED ISSUE #948: Validate Lo L1B Processed Star Sensor Data

greglucas:
# Change Summary

## Overview

This adds processing of the Lo Star Sensor packet apid. Since the code is already using `packet_file_to_datasets` we can also remove the previous star sensor dataclasses.

@sdhoyt, I am a bit confused reading through the algorithm document, so it would be good to have you double check this, but I am pretty sure we are guaranteed 720 values for the data field. Is that a true statement?

* I am doing the decompression right away. Do you want to make an L1a and L1b file where the L1a would not be decompressed?
* The csv file included in validation data was huge, so I parsed that down to just a few items. Let me know if this isn't acceptable and we could probably change this to use the s3 path instead.

closes #409 
closes #410
closes #422
closes #488 
closes #939 
closes #948


imap_processing/lo/l1a/lo_l1a.py
```diff
@@ -100,10 +101,24 @@ def lo_l1a(dependency: Path) -> list[xr.Dataset]:
...
-    good_apids = [LoAPID.ILO_SPIN, LoAPID.ILO_SCI_CNT, LoAPID.ILO_SCI_DE]
-    logger.info(f"\nReturning datasets: {[LoAPID(apid) for apid in good_apids]}")
-    return [datasets_by_apid[good_apid] for good_apid in good_apids]
+    good_apids = [
+        LoAPID.ILO_SPIN,
+        LoAPID.ILO_SCI_CNT,
+        LoAPID.ILO_SCI_DE,
+        LoAPID.ILO_STAR,
+    ]
+    apids_with_data = [apid for apid in good_apids if apid in datasets_by_apid]
```
greglucas: I reorganized this a bit because I think it would have failed if you had a packet file without some apids. We only want to look at the apids that were included in both of your lists, good and present.

imap_processing/tests/lo/test_star_sensor.py
```diff
@@ -1,35 +1,50 @@
...
-def test_science_counts(star_sensor):
-    ## Arrange
-    star_sensor.DATA_COMPRESSED = "0" * 5760
+def test_science_counts(star_sensor_ds):
+    validation_file = (
+        imap_module_directory
+        / "tests/lo/validation_data"
+        / "Instrument_FM1_T104_R129_20240803_ILO_STAR_EU_trimmed.csv"
+    )
+    validation_arr = np.loadtxt(validation_file, delimiter=",", skiprows=1, dtype=int)
```
greglucas, line 30: This is easier than pandas read_csv because there are so many columns, so just subset the fields ourselves instead.

greglucas:
The last two commits refactored this a bit and separated the values out into two datasets now, an l1a and l1b. I think it makes sense to do both of these at the same time since it is such a minor change from l1a to l1b so I don't think we need to create a separate processing job for it.

The l1b dataset is currently called `prostar`, is that what they want or does `l1b_star` indicate it is processed?

sdhoyt:
> The last two commits refactored this a bit and separated the values out into two datasets now, an l1a and l1b. I think it makes sense to do both of these at the same time since it is such a minor change from l1a to l1b so I don't think we need to create a separate processing job for it.
> 
> The l1b dataset is currently called `prostar`, is that what they want or does `l1b_star` indicate it is processed?

@greglucas I believe the L1A star data should be decompressed. We are decompressing the other data products in L1A as well. But yes, the plan was to use prostar for l1b

imap_processing/lo/l0/lo_star_sensor.py
```diff
@@ -0,0 +1,70 @@
...
+    Returns
+    -------
+    xr.Dataset
+        Processed dataset with a decompressed data field.
+    """
+    # Make one long flat buffer
+    # This assumes that all data_compressed entries are of the same length
+    # but allows for only one frombuffer call
+    buffer = b"".join(ds["data_compressed"].values)
+    data = np.frombuffer(buffer, dtype=np.uint8).reshape(-1, 720)
```
sdhoyt, line 35: I'll need to double check on the 720. Originally I thought it would always be that, but Colin told me at one point that it was 720 or less. I'll ask Lo at our tag-up on Monday

greglucas: :+1: Let me know if you need any clarification from me or want me to join that tag-up. It won't be hard to change this, but it is a bit unclear what it is stating currently. Section 8.3.5.3 has an explicit **720** listed in the output. Then the following section states:
> There is an exception in this case for the
DATA field, which contains is a N X 720 dimensional array, containing the 720 measurements
made by the star tracker for each packet.

Should these instead be `(N, COUNT)` dimensionality. For CDF we'd need to make arrays that are shape `(N, MAX_COUNT)` and fill the unused values.

greglucas: @sdhoyt any update on the 720 after your tag-up? Nick seemed to indicate he thought it would always be 720 for this packet.

sdhoyt: Not yet. I'm going to ask in my tag-up tomorrow morning so I'll let you know once I hear.

sdhoyt: @greglucas I confirmed that there will always be 720 data points in the packet

greglucas: OK, this is ready for review then!

sdhoyt (approved):
Thanks for doing this!

PR #1878, opened 2025-06-30
Ultra l1a - add more apids 898, 899, 900, 901

laspsandoval:
# Change Summary

## Overview
Adding events apids for priority data products.

## Updated Files
- ultra_l1a
   - Added way to create xarray for new apids
- ultra_utils
   - Added packet properties for apids
- decom_ultra
   - Added new apids
   - Added global attributes

## Testing
conftest.py
test_decom_apid_898.py
test_decom_apid_899.py
test_decom_apid_900.py
test_decom_apid_901.py
test_ultra_l1a.py
external_test_data_config.py


imap_processing/tests/external_test_data_config.py
```diff
@@ -67,6 +67,9 @@
     ("ultra45_raw_sc_imgpriority1evnt_FM45_UltraFM45Extra_TV_Tests_2024-01-22T0930_20240122T093008.csv", "ultra/data/l0/"),
     ("ultra45_raw_sc_ultranrgrates_FM45_UltraFM45_Functional_"
      "2024-01-22T0105_20240122T010548.csv", "ultra/data/l0/"),
+    ("ultra45_raw_sc_imgpriority2evnt_FM45_UltraFM45Extra_TV_Tests_2024-01-22T0930_20240122T093008.csv", "ultra/data/l0/"),
```
greglucas, line 69: Could you move this block before the `ultranrgrates` file so the `imgpriority1evnt` is with this block. When I first read this it looked like it was missing the 1evnt, but that is just right above this it seems.

imap_processing/tests/ultra/unit/test_decom_apid_900.py
```diff
@@ -0,0 +1,118 @@
...
+    decom_ultra = decom_test_data
+
+    df = pd.read_csv(priority_3_events_test_path, index_col="MET")
+
+    # # Check all values of each column are as expected,
+    # except for those set to fill value
+    np.testing.assert_array_equal(
+        df["SID"].values[df["SID"].values != -1],
+        decom_ultra["sid"].values[df["SID"].values != -1],
+    )
```
greglucas: As a general comment, this might be easier to read and keep track of with a loop over the variables. Rather than lots of individual checks.
```python
vars_to_compare = ["SID", "Spin", "AbortFlag"]
for var in vars_to_compare:
    good_values = df[var] != -1
    np.testing.assert_array_equal(df[var][good_values], decom_ultra[var.lower()].values[good_values])

File: imap_processing/ultra/packet_definitions/ULTRA_SCI_COMBINED.xml [hunk 3]
```diff
@@ -4726,15 +4726,15 @@
 			<xtce:IntegerParameterType name="U45_IMG_ENA_PHXTOF_HI_TIME.SPIN" signed="false">
 				<xtce:IntegerDataEncoding sizeInBits="8" encoding="unsigned" />
 			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="U45_IMG_ENA_PHXTOF_HI_TIME.ABORTFLAG" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="1" encoding="unsigned" />
+			</xtce:IntegerParameterType>
```
greglucas, line 4731: Looks like this was the only place this occurred? I checked the U90 and that looks good.

greglucas (commented):
Looks like the test files can't be found on the remote, did you push them to s3?

imap_processing/ultra/packet_definitions/ULTRA_SCI_COMBINED.xml
laspsandoval: Yes. I checked everything else. Odd that it was just right there that it was a problem.

laspsandoval:
> Looks like the test files can't be found on the remote, did you push them to s3?

Good catch. They are now in s3.

greglucas (approved):
Thanks for making all those updates, I find this easier to read :)

PR #1879, opened 2025-07-01
Ultra l1a 897 - update to packet definitions

laspsandoval:
# Change Summary

## Overview
Table 57 of the Ultra FSW was updated since the last test data was provided to me. Larry Brown confirmed the new values. This is just an update with the new test data and packet definitions.

## Updated Files
- ultra_utils.py
   - updated packet definitions

## Testing
test_decom_apid_897.py
external_test_data_config.py


greglucas (approved):
Interesting that there is a phase_angle now when there wasn't before.

PR #1882, opened 2025-07-01
SWAPI: adding sci start time and update epoch to center time

### LINKED ISSUE #1881: SWAPI: update epoch to center time and add sci start time
Per project guidance, update epoch to center time.
Per SWAPI request, keep science start time for L3 purposes.

tech3371:
# Change Summary
closes #1881 

## Overview
Update epoch to use center time and keep start time in new variable per SWAPI requests.

## Updated Files

   - Add attrs for new variable
   - Update L1 to do main part of this PR
   - put variable names in alphabetic order and added new variable to carry in L2 file which can be used in L3.

### DIFF
--- a/imap_processing/swapi/l1/swapi_l1.py
+++ b/imap_processing/swapi/l1/swapi_l1.py
@@ -631,7 +631,7 @@ def process_swapi_science(
         dims=["epoch"],
         attrs=cdf_manager.get_variable_attributes("plan_id"),
     )
-    # Store start time of L3 purposes per SWAPI requests
+    # Store start time for L3 purposes per SWAPI requests
     dataset["sci_start_time"] = xr.DataArray(
         good_sweep_sci["epoch"].data.reshape(total_full_sweeps, 12)[:, 0],
         name="sci_start_time",
--- a/imap_processing/swapi/l1/swapi_l1.py
+++ b/imap_processing/swapi/l1/swapi_l1.py
@@ -558,8 +558,11 @@ def process_swapi_science(
     # Step 3: Create xarray.Dataset
     # ===================================================================
 
-    # epoch time. Should be same dimension as number of good sweeps
-    epoch_values = good_sweep_sci["epoch"].data.reshape(total_full_sweeps, 12)[:, 0]
+    # epoch time. Should be same dimension as number of good sweeps.
+    # Use center time for epoch to line up with mission requests. Center time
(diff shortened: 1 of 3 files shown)


imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -628,6 +631,13 @@ def process_swapi_science(
         dims=["epoch"],
         attrs=cdf_manager.get_variable_attributes("plan_id"),
     )
+    # Store start time of L3 purposes per SWAPI requests
```
greglucas: Interesting, couldn't this be calculated by subtracting a fixed offset from the center time? 🤷

greglucas (approved):

imap_processing/swapi/l1/swapi_l1.py
tech3371: yeah, I didn't ask and I am guessing they didn't want to derived.

tech3371:
@greglucas Can I merge over codecov? I only added one new variable for this PR.

tech3371:
> Quote reply

Looks like codecov passed. I didn't do anything differently. I am going to merge this to keep things moving but @bishwassth don't hesitate to leave comment and I will address in new PR.

greglucas:
> Looks like codecov passed. I didn't do anything differently.

codecov currently fails until the test with remote data completes, so it lags behind. I'm guessing you were just looking before all jobs had finished.

One other thing I just thought about is do you need to add `BIN_LOCATION` to this new variable to indicate it is at the left edge 0.0 value rather than center 0.5? I'm also not sure if would need a DELTA_PLUS or DELTA_MINUS to be different on this epoch vs the new one... Things to keep in mind in the future.

tech3371:
> > Looks like codecov passed. I didn't do anything differently.
> 
> codecov currently fails until the test with remote data completes, so it lags behind. I'm guessing you were just looking before all jobs had finished.
> 
> One other thing I just thought about is do you need to add `BIN_LOCATION` to this new variable to indicate it is at the left edge 0.0 value rather than center 0.5? I'm also not sure if would need a DELTA_PLUS or DELTA_MINUS to be different on this epoch vs the new one... Things to keep in mind in the future.

Yeah, I thought about adding the DELTA but then realized that we all use common epoch attrs and I didn't want to mess with other's time attrs. That seem beyond this PR.

PR #1884, opened 2025-07-01
SWE: add uncertainty in L1B and L2

### LINKED ISSUE #1883: SWE: add uncertainty in L1B and L2
Add uncertainty in L1B and L2 per project requests and instrument requests

Notes from meeting:
Will need to add uncertainty to l1b and onward. This is coming from Matina and project.
Ruth wants to do statistical uncertainty which is sqrt(decompressed counts but not the deadtime correction. Come back to this). carry that through L2.
In L2,
do uncertainty data (phase space density calculation and the flux calculation), that gives the statistical uncertainty at level 2.
uncertainty calculation should go in the same bins as the data. Then average the unc data with sqrt(sum(uncertainty)) which is same as sqrt(sum((unc_1)**2, (unc_2)**2, so on)).
This is important in L3.
Get to this once we validate L2 product.

tech3371:
# Change Summary
closes #1883 

## Overview
Adding SWE uncertainty in L1B and L2 CDF files.

## Updated Files

- imap_processing/cdf/config/imap_swe_l1b_variable_attrs.yaml and imap_processing/cdf/config/imap_swe_l2_variable_attrs.yaml
   - add CDF attrs for new variable in L1B and L2
-  imap_processing/swe/l1b/swe_l1b.py and imap_processing/swe/l2/swe_l2.py
   - Adding L1B and L2 uncertainty.

### DIFF
--- a/imap_processing/swe/l1b/swe_l1b.py
+++ b/imap_processing/swe/l1b/swe_l1b.py
@@ -762,7 +762,7 @@ def swe_l1b_science(dependencies: ProcessingInputCollection) -> xr.Dataset:
     count_rate = convert_counts_to_rate(inflight_applied_count, acq_duration)
 
     # Statistical uncertainty is sqrt(decompressed counts)
-    # TODO: Update this if SWE like to include deatime correciton. 
+    # TODO: Update this if SWE like to include deadtime correciton.
     counts_stat_uncert = np.sqrt(populated_data["science_data"])
 
     # Store ESA energies of full cycle for L2 purposes.
--- a/imap_processing/swe/l1b/swe_l1b.py
+++ b/imap_processing/swe/l1b/swe_l1b.py
@@ -761,6 +761,10 @@ def swe_l1b_science(dependencies: ProcessingInputCollection) -> xr.Dataset:
 
     count_rate = convert_counts_to_rate(inflight_applied_count, acq_duration)
 
+    # Statistical uncertainty is sqrt(decompressed counts)
+    # TODO: Update this if SWE like to include deatime correciton. 
+    counts_stat_uncert = np.sqrt(populated_data["science_data"])
+
     # Store ESA energies of full cycle for L2 purposes.
     esa_energies = get_esa_energy_pattern(esa_lut_files[0])
     # Repeat energies to be in the same shape as the science data
@@ -893,6 +897,11 @@ def swe_l1b_science(dependencies: ProcessingInputCollection) -> xr.Dataset:
         count_rate,
         dims=["epoch", "esa_step", "spin_sector", "cem_id"],
         attrs=cdf_attrs.get_variable_attributes("science_data"),
+    )
+    science_dataset["counts_stat_uncert"] = xr.DataArray(
+        counts_stat_uncert,
+        dims=["epoch", "esa_step", "spin_sector", "cem_id"],
+        attrs=cdf_attrs.get_variable_attributes("counts_stat_uncert"),
     )
     science_dataset["acquisition_time"] = xr.DataArray(
         acq_time,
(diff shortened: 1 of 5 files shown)


imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -761,6 +761,10 @@ def swe_l1b_science(dependencies: ProcessingInputCollection) -> xr.Dataset:
 
     count_rate = convert_counts_to_rate(inflight_applied_count, acq_duration)
 
+    # Statistical uncertainty is sqrt(decompressed counts)
+    # TODO: Update this if SWE like to include deatime correciton.
```
```suggestion
    # TODO: Update this if SWE like to include deadtime correciton. 
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -155,6 +156,10 @@ def put_data_into_angle_bins(
     angle_bin_indices : numpy.ndarray
         Indices of angle bins to put data in. Shape:
         (full_cycle_data, N_ESA_STEPS, N_ANGLE_BINS).
+    is_unc : bool
+        Whether data is uncertainty data or not. If yes, sqrt(sum(data)).
```
greglucas: Where did this equation come from, could you reference this somewhere?

If you are summing all data values in a bin then I think this makes sense for reporting uncertainty, but aren't you calculating the mean number of counts in each bin? `sum(data) / len(data)`, So I'm wondering if the uncertainty needs to be scaled by the number of samples in your bin so it doesn't grow unbounded and increase just because you have more samples.

```
sum(sqrt(data)) / bin_counts
```

tech3371: From this meeting notes, https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/270122254/2025+06+30. I plan to follow-up with Ruth on this equation. I also unsure about formula and plan to ask Ruth.

greglucas: Oh... I see what I am missing now, thanks for sharing those notes! I was thinking you were still passing in "counts" and calculating uncertainty here. You are actually passing in the uncertainty to the function and summing the uncertainty here (not counts). Sorry for not understanding that at first! Then what you have here I think makes more sense now, I haven't given it a lot of thought, but seems like it could be right now.

imap_processing/swe/l2/swe_l2.py
```diff
@@ -49,27 +51,27 @@ def calculate_phase_space_density(l1b_dataset: xr.Dataset) -> npt.NDArray:
...
+    #   2 * ((C/tau) or uncertainty data) / (G * 1.237e31 * eV^2)
     # See doc string for more details.
-    density = (2 * l1b_dataset["science_data"]) / (
+    density = (2 * data) / (
         swe_constants.GEOMETRIC_FACTORS[np.newaxis, np.newaxis, np.newaxis, :]
         * swe_constants.VELOCITY_CONVERSION_FACTOR
         * particle_energy_data[:, :, :, np.newaxis] ** 2
     )
-    phase_space_density = density.data
+    phase_space_density = density
```
greglucas: Call the variable above this phase_space_density or return density below this, no need for this extra name.
```suggestion
```

imap_processing/swe/l2/swe_l2.py
```diff
@@ -177,7 +184,29 @@ def put_data_into_angle_bins(
     time_indices = np.arange(data.shape[0])[:, None, None]
     energy_indices = np.arange(swe_constants.N_ESA_STEPS)[None, :, None]
 
-    # Use np.add.at() to accumulate values into bins
+    if is_unc:
```
greglucas: This is a pretty detailed comment here. Do you want to make a different function for uncertainties so you don't have to call it with this keyword argument and do two different formulas in this one function? `put_uncertainty_into_angle_bins()` or similar.

greglucas (approved):

imap_processing/swe/l2/swe_l2.py
tech3371: sure thing

PR #1886, opened 2025-07-01
1885 write cdf   default compression to 6

### LINKED ISSUE #1885: write_cdf - default compression to 6
### Algorithm Description:

The cdflib function xarray_to_cdf defaults to no compression. We should override this default in the `write_cdf` function to the standard value of 6. According to cdflib documentation, this is the suggested variable compression that Andriy suggested.

Additionally since this will be a small PR, fix the bug with `**kwargs` not being passed to `xarray_to_cdf` correctly.

tmplummer:
# Change Summary

## Overview
This PR:
- Sets the default gzip compression for CDF variables to 6.
- Fixes a bug with how extra kwargs were being passed to `cdf_to_xarray`.

## Updated Files
   - Fix bug in how `kwargs` was being passed through to `cdf_to_xarray`
   - Set default compression in `write_cdf` to 6.
   - Add test coverage for kwargs passed to `cdf_to_xarray` and `xarray_to_cdf`

Closes: #1885

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -44,7 +44,7 @@ def load_cdf(
     if isinstance(file_path, imap_data_access.ImapFilePath):
         file_path = file_path.construct_path()
 
-    dataset = cdf_to_xarray(file_path, kwargs)
+    dataset = cdf_to_xarray(file_path, **kwargs)
 
     # cdf_to_xarray converts single-value attributes to lists
     # convert these back to single values where applicable
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -44,6 +44,10 @@ def load_cdf(
     if isinstance(file_path, imap_data_access.ImapFilePath):
         file_path = file_path.construct_path()
 
+    # By default, do not convert epoch to datetime64. This ensures that the
+    # round-trip of writing and then loading a cdf keeps the dataset the same.
+    if "to_datetime" not in kwargs:
+        kwargs["to_datetime"] = False  # type: ignore
     dataset = cdf_to_xarray(file_path, **kwargs)
 
     # cdf_to_xarray converts single-value attributes to lists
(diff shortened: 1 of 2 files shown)


imap_processing/cdf/utils.py
```diff
@@ -44,7 +44,7 @@ def load_cdf(
     if isinstance(file_path, imap_data_access.ImapFilePath):
         file_path = file_path.construct_path()
 
-    dataset = cdf_to_xarray(file_path, kwargs)
+    dataset = cdf_to_xarray(file_path, **kwargs)
```
greglucas, line 51: good catch!

greglucas (approved):
Did you test this on your datasets at all to see if it helped?

tmplummer:
@greglucas 

> Did you test this on your datasets at all to see if it helped?

Good idea... haha. Yes for a 34 MB file that I read in and wrote back out, the new file is 17 MB.

imap_processing/cdf/utils.py
```diff
@@ -44,7 +44,11 @@ def load_cdf(
     if isinstance(file_path, imap_data_access.ImapFilePath):
         file_path = file_path.construct_path()
 
-    dataset = cdf_to_xarray(file_path, kwargs)
+    # By default, do not convert epoch to datetime64. This ensures that the
+    # round-trip of writing and then loading a cdf keeps the dataset the same.
+    if "to_datetime" not in kwargs:
+        kwargs["to_datetime"] = False  # type: ignore
```
tmplummer, line 50: @greglucas, do you have any opinions about this implementation (mirrors what is also in the `write_cdf` function) of setting defaults in the kwargs dict vs. just exposing those keywords at the interface to the load and write functions?

greglucas: Do you mean having a signature like the following instead?
```python
def func(args, to_datetime=False, **kwargs):
    cdf_to_xarray(file_path, to_datetime, **kwargs)
```

I don't have a strong opinion here if that is what you are asking about. I would suggest doing that as a separate PR though.

PR #1904, opened 2025-07-02
Ultra l1a - adding cmdecho (apid 865)

laspsandoval:
# Change Summary

## Overview
Adding cmdecho apid.

## Updated Files
- decom_ultra.py
   - Added cmdecho code
- ultra_l1a.py
   - incorporate cmdecho into logic for l1a
   - attributes
   - attributes
- ultra_utils
   - Packet properties and fields

## Testing
test_decom_apid_865.py
test_ultra_l1a.py
conftest.py
ultra45_raw_hk_ultracmdecho_FM45_UltraFM45_Functional_2024-01-22T0105_20240122T010548.csv

### DIFF
--- a/imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_ultra_global_cdf_attrs.yaml
@@ -236,6 +236,18 @@ imap_ultra_l1a_90sensor-alarm:
     Logical_source: imap_ultra_l1a_90sensor-alarm
     Logical_source_description: IMAP-Ultra Instrument Level-1A Alarm Data for Ultra90.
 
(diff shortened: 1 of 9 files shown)


greglucas (commented):
Are strings of hex values actually what is desired here in the stored cdf, or is that just the representation in the csv file that was given for validation? i.e. I'm a bit confused whether you want to be doing this in the actual decom_ultra routine versus doing it in the tests so the comparisons work out there.

Should the `args` be a 2d array of values with an index of the number of maximum args with a fill value for args that weren't sent? i.e. each arg an individual entry instead of all combined into a single string.

greglucas (approved):
Looks good to me, thanks for making those quick updates!

Issue #1501, opened 2025-03-19
CoDICE: epoch times should reflect center of acquisition time bins


bourque:
A new convention came out of the Feb 2025 Science Team Meeting: The times listed in the `epoch` variable should reflect the center of the acquisition time period instead of the start. Additionally, this will require new variables that store the deltas. The CoDICE pipeline should be updated to support this.

tech3371:
Can we close this now that PR is merged?

bourque:
This still needs to be implemented for `hi-ialirt` and `hi-omni` data products. I will close this issue and open a new one for these.

tech3371:
We centered time for all product already but we are seeing 3sec difference in center time. Once that is fixed, we need to update all coordinate comparison in unittest to be equal once we match epoch center time.

tech3371:
Closing this because CoDICE is not using SPICE. We verified that the time delta matches.

PR #1906, opened 2025-07-02
Implement center/delta epoch values for most CoDICE L1a products

bourque:
This PR implements epoch values that are centered within the acquisition time window, as well as data variables for their corresponding `delta_minus` and `delta_plus` for most CoDICE L1a data products. These calculations for the `hskp`, `hi-omni`, and `hi-ialirt` are a bit trickier and will be addressed in a separate PR.

Another thing to note is that this PR will be the first of several that makes (relatively) small changes to CoDICE L1a data products. The goal is to eventually re-validate against updated validation data once all of those fixes are in. In the meantime, I may need to tweak or turn off certain unit tests to get tests passing in this middle-ground state between sets of validation data.

Partially completes #1501

### DIFF
--- a/imap_processing/codice/codice_l1a.py
+++ b/imap_processing/codice/codice_l1a.py
@@ -54,8 +54,6 @@ class CoDICEL1aPipeline:
 
     Methods
     -------
-    calculate_epoch_values()
-        Calculate and return the values to be used for `epoch`.
     decompress_data(science_values)
         Perform decompression on the data.
     define_coordinates()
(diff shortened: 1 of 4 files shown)


imap_processing/codice/codice_l1a.py
```diff
@@ -89,28 +87,6 @@ def __init__(self, table_id: int, plan_id: int, plan_step: int, view_id: int):
         self.plan_step = plan_step
         self.view_id = view_id
 
-    def calculate_epoch_values(self) -> NDArray[int]:
```
bourque, line 92: I moved this method outside of the class and made it a function so I can re-use it in multiple places in the code.

imap_processing/codice/codice_l1a.py
```diff
@@ -167,17 +143,30 @@ def define_coordinates(self) -> None:
...
+            ("epoch_delta_minus", epoch_delta_minus),
+            ("epoch_delta_plus", epoch_delta_plus),
+        ]:
+            coord = xr.DataArray(
+                var,
+                name=name,
+                dims=[name],
+                attrs=self.cdf_attrs.get_variable_attributes(name, check_schema=False),
+            )
+            self.coords[name] = var
```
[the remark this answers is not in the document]
bourque: Great catch robot overlord!

imap_processing/codice/codice_l1a.py
```diff
@@ -674,6 +663,52 @@ def set_data_product_config(self, apid: int, dataset: xr.Dataset) -> None:
...
+
+    # Apply correction to center the epoch bin
+    epoch = (acq_start[:-1] + acq_start[1:]) / 2
+    epoch_delta_minus = epoch - acq_start[:-1]
+    epoch_delta_plus = acq_start[1:] - epoch
+
+    # The end values are calculated differently
+    epoch = np.concatenate([epoch, [acq_start[-1]]])
+    epoch_delta_minus = np.concatenate([epoch_delta_minus, [epoch_delta_minus[-1]]])
+    epoch_delta_plus = np.concatenate([epoch_delta_plus, [epoch_delta_plus[-1]]])
```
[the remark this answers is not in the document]
bourque: I think just adding `.astype(int)` to the end of line 700 fixes all of this. I think the other changes are not necessary?

imap_processing/codice/codice_l1a.py
```diff
@@ -674,6 +663,52 @@ def set_data_product_config(self, apid: int, dataset: xr.Dataset) -> None:
...
+        start to the center of the acquisition time bin.
+    epoch_delta_plus: NDArray[int]
+        List of values that represent the length of time from the center of
+        the acquisition time bin to the end of acquisition.
+    """
+    # First calculate an epoch value based on the acquisition start
+    acq_start = met_to_ttj2000ns(acq_start_seconds + acq_start_subseconds / 1e6)
+
+    # Apply correction to center the epoch bin
+    epoch = ((acq_start[:-1] + acq_start[1:]) / 2).astype(int)
```
greglucas: Another option is to use integer division.
```suggestion
    epoch = ((acq_start[:-1] + acq_start[1:]) // 2)
```

imap_processing/codice/codice_l1a.py
```diff
@@ -674,6 +663,52 @@ def set_data_product_config(self, apid: int, dataset: xr.Dataset) -> None:
...
+    epoch_delta_minus = epoch - acq_start[:-1]
+    epoch_delta_plus = acq_start[1:] - epoch
+
+    # The end values are calculated differently
```
greglucas: How? Could you add a comment here saying what you are doing. I would have naively assumed you would handle the minus and plus differently because they are going off different ends of the array. (one prepend, the other append)

greglucas (approved):

PR #1907, opened 2025-07-02
Ultra l1a - add apid 872 (macroschecksum)

laspsandoval:
# Change Summary

## Overview
Adding another apid for hk apid 872

## Updated Files
- decom_ultra.py
   - Added code to parse the data
- ultra_utils.py
   - Added packet properties
- ultra_l1a.py
   - Added logic so it will be an l1a file
   - Added attributes
   - Added attributes

## Testing
conftest.py
ultra45_raw_hk_macrochecksumrpt_FM45_UltraFM45_Functional_2024-01-22T0105_20240122T010548.csv
test_decom_apid_872.py
test_ultra_l1a.py


imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -344,3 +344,38 @@ def process_ultra_energy_spectra(ds: xr.Dataset) -> xr.Dataset:
...
+    ds : xarray.Dataset
+        Dataset containing macro checksums.
+
+    Returns
+    -------
+    dataset : xarray.Dataset
+        Dataset with unpacked and decoded checksum values.
+    """
+    checksums = []
+    fill = np.iinfo(np.uint16).max
```
greglucas: It might help to bring your dtype below up here to define the fill value that way and then re-use the packed_dtype (or whatever name you prefer, maybe checksum_dtype) in the `from_buffer(, dtype=packed_dtype)` call below
```suggestion
    # big endian uint16
    packed_dtype = np.dtype(">u2")
    fill = np.iinfo(packed_dtype).max
```

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -344,3 +344,38 @@ def process_ultra_energy_spectra(ds: xr.Dataset) -> xr.Dataset:
...
+
+    for checksum in ds["checksums"]:
+        checksum_bytes = checksum.item()
+        arr = np.frombuffer(checksum_bytes, dtype=">u2")
+        checksums.append(arr)
```
greglucas: Rather than appending to a list here can you initialize an empty array above with the proper shape?

```python
nepochs = len(ds["epoch"])
nchecksums = 256
checksum_array = np.full((nepochs, nchecksums), fill, dtype=packed_dtype)
for i, checksum_row in enumerate(ds["checksums"]):
    checksum_array[i, :] = np.frombuffer(checksum_row.item(), dtype=packed_dtype)
```

imap_processing/ultra/l0/decom_ultra.py
```diff
@@ -344,3 +344,38 @@ def process_ultra_energy_spectra(ds: xr.Dataset) -> xr.Dataset:
...
+
+    checksum_array = np.array(checksums)
+    checksum_array[:, -1] = fill
```
greglucas: This doesn't seem quite right, why are we making it fill regardless of what came down? Do we want to append another byte or entry to the array for this extra fill instead?

laspsandoval: Yes. I don't know how this is supposed to be handled. I was mimicking the test data. I've reached out to Larry Brown to ask but he is out today. So I think I'll have to wait until next week.

greglucas (approved):

PR #1911, opened 2025-07-03
Created cdf for Ultra l1c pset helio frame

### LINKED ISSUE #1807: BUG - Ultra l1a de - Investigate reason for uint8 spin number
### Description of the issue

_No response_

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

laspsandoval:
# Change Summary

## Overview
This is taking the original code and formatting it so that it can be placed into a cdf. I also had feedback from Nick Dutton that I can delete the helio frame histogram function and instead use the spacecraft histogram function if I have appropriate inputs. I also wrote a test using fake data provided by Nick Dutton. 

## Updated Files
- helio_pset.py
   - Format existing code to create a cdf
- ultra_l1c_pset_bins.py
   - Remove helio histogram code per Nick Dutton’s recommendation. I am instead using the spacecraft histogram code.
- ultra_l1_utils.py
   - Update data array structure for histogram
- ultra_l1b_extended.py
   - Quick fix to calculate spin number for direct events data product.
- de.py
   - Include spin number calculation in direct events data product.

## Testing
test_ultra_l1c.py
test_ultra_l1c_pset_bins.py
test_ultra_l1b_extended.py

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
@@ -478,8 +478,12 @@ def test_get_spin_number(test_fixture, use_fake_spin_data_for_time):
 
     # Convert shcoarse (time) to MET relative to first value
     # Take off first partial spin.
-    de_met = de_dataset["shcoarse"].values[19::] - de_dataset["shcoarse"].values[19]
-    de_spin = de_dataset["spin"].values[19::]
+    first_spin_index = 19
+    de_met = (
+        de_dataset["shcoarse"].values[first_spin_index:]
+        - de_dataset["shcoarse"].values[first_spin_index]
+    )
+    de_spin = de_dataset["spin"].values[first_spin_index:]
 
     spin_number = get_spin_number(de_met, de_spin)
     # First spin number is 128.
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
@@ -473,23 +473,15 @@ def test_get_spin_number(test_fixture, use_fake_spin_data_for_time):
     """Tests that get_spin_number assigns the correct spin number."""
     df_filt, _, _, de_dataset = test_fixture
 
-    # Simulate a spin table from MET = 0 to MET = 141*15 seconds
-    use_fake_spin_data_for_time(start_met=0, end_met=141 * 15)
+    # Simulate a spin table from MET = 0 to MET = 500*15 seconds
+    use_fake_spin_data_for_time(start_met=0, end_met=500 * 15)
 
-    # Convert shcoarse (time) to MET relative to first value
-    # Take off first partial spin.
-    first_spin_index = 19
-    de_met = (
-        de_dataset["shcoarse"].values[first_spin_index:]
-        - de_dataset["shcoarse"].values[first_spin_index]
-    )
-    de_spin = de_dataset["spin"].values[first_spin_index:]
+    de_met = np.array([0, 0, 5760, 5760, 5760, 5760, 5760, 5760])
+    de_spin = np.array([128, 128, 129, 129, 130, 130, 131, 131], dtype=np.uint8)
 
     spin_number = get_spin_number(de_met, de_spin)
-    # First spin number is 128.
-    expected_spin_number = de_spin - 128
 
-    assert np.array_equal(spin_number, expected_spin_number)
+    assert np.array_equal(spin_number & 0xFF, de_spin)
 
 
 def test_get_eventtimes(test_fixture, use_fake_spin_data_for_time):
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
@@ -28,6 +28,7 @@ from imap_processing.ultra.l1b.ultra_l1b_extended import (
     get_path_length,
     get_ph_tof_and_back_positions,
     get_phi_theta,
+    get_spin_number,
     get_ssd_back_position_and_tof_offset,
     get_ssd_tof,
     interpolate_fwhm,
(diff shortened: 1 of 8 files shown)


imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -468,6 +469,25 @@ def test_get_phi_theta(test_fixture):
...
+    """Tests that get_spin_number assigns the correct spin number."""
+    df_filt, _, _, de_dataset = test_fixture
+
+    # Simulate a spin table from MET = 0 to MET = 141*15 seconds
+    use_fake_spin_data_for_time(start_met=0, end_met=141 * 15)
+
+    # Convert shcoarse (time) to MET relative to first value
+    # Take off first partial spin.
+    de_met = de_dataset["shcoarse"].values[19::] - de_dataset["shcoarse"].values[19]
+    de_spin = de_dataset["spin"].values[19::]
```
greglucas: It took me a second to figure out what 19 referred to, could you call that a specific variable name here? Also remove the second colon if it isn't used.
```suggestion
    first_spin_index = 19
    de_met = de_dataset["shcoarse"].values[first_spin_index:] - de_dataset["shcoarse"].values[first_spin_index]
    de_spin = de_dataset["spin"].values[first_spin_index:]
```

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -468,6 +469,25 @@ def test_get_phi_theta(test_fixture):
...
+
+    spin_number = get_spin_number(de_met, de_spin)
```
greglucas, line 482: Are you using the use_fake_spin_data? It looks like you are passing in your own spin table dataset here, so it is a bit unclear where that function takes effect.

imap_processing/ultra/l1c/helio_pset.py
```diff
@@ -36,17 +43,57 @@ def calculate_helio_pset(
...
+
+    # calculate background rates
+    background_rates = get_helio_background_rates()
+
+    efficiencies = ancillary_files["l1c-90sensor-efficiencies"]
+    geometric_function = ancillary_files["l1c-90sensor-gf"]
+
+    df_efficiencies = pd.read_csv(efficiencies)
+    df_geometric_function = pd.read_csv(geometric_function)
+    mid_time = sct_to_et(de_dataset["event_times"].data.mean())
```
greglucas: Should this be a mean (median?) of the exposure time below and not dependent on events?

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
laspsandoval: Here I am matching the Universal Spin table to the direct events. So I am getting the spin number for the direct events.

imap_processing/ultra/l1c/helio_pset.py
laspsandoval: The code that Nick gave me in rectangular coordinates used the median. I have asked Nick to review this code since it is not in the Algorithm Document.

imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
laspsandoval: The direct event spin number is uint8 (0-255) and the Universal Spin table spin number is uint32. So I have to figure out which spin number in the direct events corresponds to the Universal Spin table spin.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -805,6 +805,58 @@ def get_phi_theta(
...
+    # Determine the indices of new spins.
+    is_new_spin = np.concatenate([[True], de_spin_sorted[1:] != de_spin_sorted[:-1]])
+    spin_start_indices = np.where(is_new_spin)[0]
+    spin_end_indices = np.append(spin_start_indices[1:], len(de_met_sorted))
+
+    spin_start_mets = spin_df["spin_start_met"].values
+    spin_numbers = spin_df["spin_number"].values
+    assigned_spin_number_sorted = np.empty(de_spin_sorted.shape, dtype=np.uint32)
+    # Take last 8 bits of spin numbers.
+    possible_spins = spin_numbers & 0xFF
```
greglucas: I don't think I am following what is going on here with the uint8 vs uint32 discussion. Why do we want to get the last 8 bits from our spin table? If a pointing is roughly one day we will have 5760 spins in it. Way more than 256 for 8 bits.

Could you write a verbose description of what this function is doing because there is a lot of spins floating around here and I'm not sure whether it is the instrument's spin, the spin table id, etc...

The spin table has a `spin_start_met` column that I think you should be able to do something like digitize your input met to fit between the starts of the spins to get their index and then return that spin number.
```python
spin_indices = np.searchsorted(spin_df["spin_start_met"], de_met)
return spin_df["spin_number"].values[spin_indices]
```

But this may just be me not fully understanding what the function is intending to do where I think a verbose docstring could help.

laspsandoval: Thanks for honing in on this part, because it is a **very** important part of the processing and not something that should be done quickly without a thoughtful second-opinion. So I really appreciate it.

Agreed. I did what you described first. I took the median met of the de_spin and matched it to the Universal Spin Table. However, when I met with the Ultra team, they mentioned this idea, which I actually think makes things cleaner. The reason I get just a bit nervous matching the times is that the de_met is the packet creation time; not when the measurement was actually made. So it seemed like probably the median packet creation time of the spin would work, but I do think this is better. So let me add docstrings of what I am doing here so you'll see it clearer.

greglucas: Thank you for the detailed comments, that really helps! I can't say I fully understand what is going on still from a science perspective, but I can follow the code easier :)

I am still just a bit unclear why there is this 8bit spin ID onboard the spacecraft and we are trying to match that with many potential spins from a spin table and then find the closest time to figure out what spin things may have happened in by narrowing down groups.

> The reason I get just a bit nervous matching the times is that the de_met is the packet creation time; not when the measurement was actually made.

I agree with this concern. But where do you get the timing information about when the event happened? I feel like you need that information somewhere to be able to see what part of the sky the event came from because you need to know where the instrument was pointing. Right now all we have is the event happened in a generic spin you found, but that could have been any 15 seconds during that spin.

What I'm getting at is that if you have the actual time of the direct event, you can index directly into the spin table and not deal with any of this grouping etc. It may not be the "epoch" / packet creation time of this dataset, but maybe there is another field with that specific time information available?

greglucas (approved):

imap_processing/ultra/l1b/ultra_l1b_extended.py
laspsandoval: Thanks again for looking at this!

Unfortunately there is not another field with the specific time information in the direct events packet. The eventtime is calculate using the following equation:

event_times = spin_starts + spin_period_sec * (phase_angle / 720)

where spin_starts and spin_period_sec come from the Universal Spin Table and phase_angle comes from the direct events packet.

PR #1916, opened 2025-07-07
MAG L2 Validation

maxinelasp:
# Change Summary
Add MAG L2 validation using input data from mag team. 

## Overview

Basic usage of validation data is: 
- load example input files 
- run MAG L2 processing
- compare with example output files

In this case, the input files needed some rearranging to be in the correct format. 

These changes also include a few minor code changes to pass validation.  

## Testing

This required adding some additional download files.  

Also, the output validation data had some leapsecond related mismatches. I have shifted the validation data to exclude leap seconds here, and asked the MAG team to check if they want the 24 hour span to include or exclude timestamps.


tech3371 (approved):
I didn't looks closely at test code but the code changes looks to me.

PR #1918, opened 2025-07-08
IDEX l2c team updates 

lacoak21:
# Change Summary

## Overview
Modify the l2b code to calculate the rectangular maps. This was originally done in l2c, but since it uses so much of the l2b code, I decided to move it to l2b. L2c code now just copies the correct variables over from the l2b CDF and adds the global attributes. I kept it separate because there is a chance that the IDEX team wants this to be a 3month cadence job. I am still waiting to hear back from Ethan on that. 

## Updated Files
- imap_processing/cdf/config/imap_idex_l2b_variable_attrs.yaml
   - Add rectangular pixel bins
- imap_processing/cli.py
   - L2c processing now takes in a list of l2b datasets. 
   - remove old healpix parameters 
   - create monthly maps for counts and count rates
- imap_processing/idex/idex_l2c.py 
   - Most of l2c is done in l2b now. 
   

Add testing in l2b for the maps


imap_processing/cdf/config/imap_idex_l2b_variable_attrs.yaml
```diff
@@ -72,6 +85,30 @@ impact_charge_bins:
     UNITS: " "
     DICT_KEY: SPASE>Support>SupportQuantity:Other
 
+rectangular_lon_pixel:
+  CATDESC: Longitude pixel index for IDEX SkyMap
```
greglucas: If these are in units of degrees I think you should remove "index", maybe indicate it is the "center" of the pixel.
```suggestion
  CATDESC: Longitude pixel center for IDEX SkyMap
```

imap_processing/idex/idex_l2b.py
```diff
@@ -71,6 +73,13 @@
 )
 SPIN_PHASE_BIN_EDGES = np.array([0, 90, 180, 270, 360])
 
+# Get the rectangular map grid with the specified spacing
+SKY_GRID = AzElSkyGrid(IDEX_SPACING_DEG)
+LON_BINS_EDGES = SKY_GRID.az_bin_edges
+LAT_BINS_EDGES = SKY_GRID.el_bin_edges
+# Add one to the last bin edge to include the last bin (90 degrees)
+LAT_BINS_EDGES[-1] += 1
```
greglucas: 🤔 Something seems wrong here. Why would the SKY_GRID.el_bin_edges not give you the true edges of the map, is it doing centers instead or is there some calculation off there? It seems like this might be better fixed in the AxElSkyGrid class because I thought it should be edges from `[-90, -90 + spacing, ... 90 - spacing, 90]`

imap_processing/idex/idex_l2b.py
```diff
@@ -231,20 +328,38 @@ def compute_counts_by_charge_and_mass(
...
     counts_by_charge = np.zeros(
-        (len(epoch_doy_unique), len(CHARGE_BIN_EDGES), len(SPIN_PHASE_BIN_EDGES) - 1),
+        (*epoch_by_mass_shape, len(SPIN_PHASE_BIN_EDGES) - 1),
     )
     counts_by_mass = np.zeros(
-        (len(epoch_doy_unique), len(MASS_BIN_EDGES), len(SPIN_PHASE_BIN_EDGES) - 1),
+        (*epoch_by_mass_shape, len(SPIN_PHASE_BIN_EDGES) - 1),
+    )
+    # Initialize arrays to hold count maps. There should be 60 longitude bins,
+    # 30 latitude bins, 11 charge bins, and 11 mass bins.
```
greglucas: This makes it sound to me like (60, 30, 11, 11) for the shape. I think you mean you have two different arrays and each of them is (ntimes, 11 [charge or mass], 60 [nlons], 30 [nlats])? Could you try and rephrase it somehow.

imap_processing/idex/idex_l2b.py
```diff
@@ -266,6 +384,10 @@ def compute_counts_by_charge_and_mass(
         binned_charge = np.array(np.digitize(charge_vals, bins=CHARGE_BIN_EDGES))
         # Bin spin phases
         binned_spin_phase = bin_spin_phases(spin_phase_angles)
+        # Bin longitude and latitude into the rectangular grid.
+        binned_longitude = np.array(np.digitize(longitude, bins=LON_BINS_EDGES))
```
greglucas: digitize() should return an array already I think?

imap_processing/idex/idex_l2b.py
```diff
@@ -274,23 +397,37 @@ def compute_counts_by_charge_and_mass(
...
-        for mass_bin, charge_bin, spin_phase_bin in zip(
-            binned_mass, binned_charge, binned_spin_phase
+        # Count dust events for each spin phase, mass bin, charge bin, and bin into
+        # a rectangular grid
+        for mass_bin, charge_bin, spin_phase_bin, lon_bin, lat_bin in zip(
+            binned_mass,
+            binned_charge,
+            binned_spin_phase,
+            binned_longitude,
+            binned_latitude,
```
greglucas, line 405: Are these all guaranteed to be the same length? It is somewhat unclear to me based on the conditions above. Otherwise zip() will truncate the input lists and potentially lost some values you are interested in.

imap_processing/idex/idex_l2b.py
```diff
@@ -346,8 +489,24 @@ def compute_rates_by_charge_and_mass(
...
+        * epoch_doy_percent_on[non_zero_inds, np.newaxis, np.newaxis, np.newaxis]
+        * SECONDS_IN_DAY
+    )
 
-    return rate_by_charge, rate_by_mass, rate_quality_flags
+    return (
+        rate_by_charge,
+        rate_by_mass,
+        rate_by_charge_map,
+        rate_by_mass_map,
```
greglucas, line 525: Would it make sense to put these into a separate function? It looks like you are doing `rate_by_mass_map = rate_by_mass[..., np.newaxis]`. Could you rewrite it like that instead of repeating the same formula?

imap_processing/idex/idex_l2b.py
lacoak21: The edges are from [-90, -90 + spacing, ... 90 - spacing, 90] but np.digitize will consider the right edge closed. So if the lat value was 90 it would be placed in the right "outlier" bin. I want it to be placed in the last bin. There is a parameter "right=True" to indicate that the last bin edge is inclusive but then it makes the left bin edge exclusive.

imap_processing/idex/idex_l2b.py
lacoak21: Yes i can rephrase that.

imap_processing/idex/idex_l2b.py
lacoak21: I was having some precommit issue with these but I will try again.

imap_processing/idex/idex_l2b.py
lacoak21: Yes these should always be the same length. They are all 1d arrays with the dimension being "epoch".

imap_processing/idex/idex_l2b.py
lacoak21: Im not sure If I understand this. I dont think adding a new axis to the rates already calculated would work because the bins are different unless I am misunderstanding something. I did add a helper function though.

imap_processing/idex/idex_l2b.py
lacoak21: Ok I left it as is because I was still getting pre commit errors. I think it is because technically it could return a scalar.

greglucas: Was it mypy typing errors then? What about doing `np.asarray()` instead here to save a copy.

imap_processing/idex/idex_l2b.py
greglucas: Thanks for the explanation. Rather than changing the bin edges in place (which seems a bit worrisome if that were to cascade somewhere else to me) you did a `np.clip(result, a_max=nbins)`

lacoak21: Ok great, that is a much better solution.

imap_processing/idex/idex_l2b.py
lacoak21: OK good idea. Yes, I think it was mypy

imap_processing/idex/idex_l2b.py
```diff
@@ -377,18 +375,21 @@ def compute_counts_by_charge_and_mass(
         # Make sure longitude values are in the range [0, 360)
         longitude = np.mod(l2a_dataset["longitude"].data[current_day_indices], 360)
         latitude = l2a_dataset["latitude"].data[current_day_indices]
+        # Clip latitude values to the range [-90, 89]
+        # np.digitize does not include the right edge, so we clip to 89.
+        latitude = np.clip(np.atleast_1d(latitude), -90, 89)
```
greglucas: Can you clip the bins instead of the input? This would change the results if you went to say 0.5 degree bin spacing.

lacoak21: Im not sure how clipping the bins would work here because I want to extend the right bin rather than clipping it.

greglucas: Doesn't digitize give you nbins+1 if the result is to the right? I'm saying you would then clip that back down to nbins.

lacoak21: Oh I see now. Got it.

greglucas (approved):

PR #1921, opened 2025-07-09
Coverage plots

laspsandoval:
# Change Summary

## Overview
We have a requirement for the I-ALiRT Coordinator Study to generate coverage plots for the ground stations. This code finds out the time in UTC that each ground station (right now only for Kiel; later KSWC) will provide satellite coverage and calculates the percent coverage for the day. 

## New Files
- coverage.py
   - Calculate the coverage of the ground stations

## Updated Files
- constants.py
   - Added station properties for the ground stations

## Testing
test_coverage.py
coverage_plot

### DIFF
--- a/imap_processing/ialirt/constants.py
+++ b/imap_processing/ialirt/constants.py
@@ -45,11 +45,13 @@ class StationProperties(NamedTuple):
     longitude: float  # longitude in degrees
     latitude: float  # latitude in degrees
     altitude: float  # altitude in kilometers
+    min_elevation_deg: float  # minimum elevation angle in degrees
 
 
-# Define StationProperties instances
+# TODO: Verify longitude, latitude, altitude.
 KIEL = StationProperties(
-    longitude=-71.41,  # longitude in degrees
-    latitude=-33.94,  # latitude in degrees
-    altitude=0.157,  # altitude in kilometers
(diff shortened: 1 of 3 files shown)


imap_processing/ialirt/constants.py
```diff
@@ -36,3 +37,21 @@ class IalirtSwapiConstants:
...
+    altitude: float  # altitude in kilometers
+    min_elevation_deg: float  # minimum elevation angle in degrees
+
+
+# TODO: Verify longitude, latitude, altitude.
+KIEL = StationProperties(
+    longitude=10.122,  # degrees East
+    latitude=54.339,  # degrees North
+    altitude=0.043,  # approx 43 meters
+    min_elevation_deg=10,  # 10 degrees is the requirement
```
greglucas: Make this a dictionary up here instead of creating it below so that you don't have to keep track of it in multiple places?

```suggestion
# TODO: Verify longitude, latitude, altitude.
STATIONS = {"Kiel": StationProperties(
    longitude=10.122,  # degrees East
    latitude=54.339,  # degrees North
    altitude=0.043,  # approx 43 meters
    min_elevation_deg=10,  # 10 degrees is the requirement
}
```

greglucas (approved):
Just a minor nit, otherwise looks great!

PR #1922, opened 2025-07-09
Mag L2 completion

maxinelasp:
# Change Summary
This PR completes MAG L2 with the addition of SPICE transforms. 

This does not do a transform to RTN or GSE as we do not have those frames yet, but it is set up to do those transforms once the code is in place.

## Overview

- add a new rotate_frame method to apply spice functions to vectors
- created datasets for DSRF and SRF

## Testing

Tests simply mock the spice returns rather than using the kernels, as this is tested in the SPICE specific tests in this repo already.


imap_processing/mag/l2/mag_l2_data.py
```diff
@@ -310,3 +310,21 @@ def truncate_to_24h(self, timestamp: np.datetime64) -> None:
...
+        """
+        Rotate the vector data in the class to the output frame.
+
+        Parameters
+        ----------
+        end_frame : ValidFrames
+            The frame to rotate the data to. Must be one of the ValidFrames enum
+            values.
+        """
+        self.vectors = frame_transform(
```
tech3371, line 324: nice common tool!

tech3371 (approved):
LGTM

PR #1932, opened 2025-07-14
Default to mag frame, rotate into SRF

maxinelasp:
# Change Summary
MAG team pointed out that L2 starts in the MAG instrument frame and needs to be specifically rotated into the spacecraft frame SRF.


tech3371 (approved):
code changes looks good to me

tech3371:
looks like tests are failing.

PR #1939, opened 2025-07-15
I-ALiRT - update swe to be broken into half cycles

laspsandoval:
# Change Summary

## Overview
Had some feedback from Ruth about how to combine the data for the final data product. She wants:

1. Include only two quarter cycles to determine counterstreaming
2. Data products for each half-cycle

## Updated Files
- process_swe.py
   - Separate counterstreaming into half-cycles instead of combining them
   - Create separate entries for each half-cycle

## Testing
test_process_swe.py


tech3371 (approved):
Ruth helped review this in the meeting, she mentioned that this seems good.

PR #1940, opened 2025-07-15
I-ALiRT - SWPC Spacecraft Position and Velocity

### LINKED ISSUE #743: I-ALiRT SWPC Feedback Implementation
### Summary

met -> UTC (another secondary index)
Position, velocity from ephemeris kernel
gsm - position and trajectory

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
SWPC requested the following parameters be added to the database for each met. This code uses the new frame kernel sent by APL (imap_science_v100.tf) to find those parameters. Note that I will wait until @subagonsouth has reviewed this. The reason is that:

1. I am currently using the de440s.bsp kernel not the one sent in the email from APL (L1_de431.bsp) and I want to make certain that is right
2. Oddly the code functions fine without the ephemeris kernel and I am not certain why that is. That seems wrong
3. I committed the new frame kernel to the repo, but Tim may want it only in s3. 

spacecraft_position_gsm | float32 | Spacecraft position in GSM coordinates | -3.0e6 | 3.0e6 | km | epoch | direction (x,y,z) | 
spacecraft_velocity_gsm | float32 | Spacecraft velocity in GSM coordinates | -1.0e3 | 1.0e3 | km/s | epoch | direction (x,y,z)

## New Files
- spacecraft_state.py
   - Find the position and velocity of the spacecraft in GSM coordinates
- geometry.py
   - Added to the classes SpaceBody and SpaceFrame
- imap_science_v100.tf
   - New frame kernel

## Testing
test_spacecraft_state.py

### DIFF
--- a/imap_processing/ialirt/utils/spacecraft_state.py
+++ b/imap_processing/ialirt/utils/spacecraft_state.py
@@ -1,0 +1,36 @@
+"""Spacecraft state."""
+
+from numpy.typing import NDArray
+
+from imap_processing.spice.geometry import (
+    SpiceFrame,
+    imap_state,
+)
+from imap_processing.spice.time import met_to_sclkticks, sct_to_et
+
+
+def calculate_gsm_state(met: NDArray) -> tuple[NDArray, NDArray]:
+    """
+    Calculate the position and velocity of the spacecraft in GSM coordinates.
+
+    Parameters
+    ----------
+    met : NDArray
+        Start time in UTC.
+
+    Returns
+    -------
+    gsm_position: NDArray
+        Spacecraft position in GSM Coordinates.
+    gsm_velocity: NDArray
+        Spacecraft velocity in GSM Coordinates.
+    """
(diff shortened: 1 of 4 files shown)


greglucas:
Could you break this up into two separate PRs? I'm a bit confused what the purpose of this is because it isn't used.

`imap_state` is already tested, so I don't know that we need a separate function for this. So I'd suggest making a separate PR that is adding this state to the output file you expect to and just using these 3-4 lines directly.

```
    et = sct_to_et(met_to_sclkticks(met))
    gsm_state = imap_state(et, ref_frame=SpiceFrame.IMAP_GSM)

    gsm_position = gsm_state[:, :3]
    gsm_velocity = gsm_state[:, 3:]
```

Then a separate PR that adds the required Enums, since that is a small contained update that is needed outside of I-ALiRT too. We should add all states from that file I think

PR #1944, opened 2025-07-16
I-ALiRT - account for outages

laspsandoval:
# Change Summary

## Overview
Updated coverage code to account for outages from ground stations. Note that the plan is to:
1. Add a s3 directory for outage text files
2. Create a cron job that looks for new text files and generates coverage text files for a date 30 days into the future and a date in which there is a new outage

Or something like that. See the next PR in sds-data-manager.

## Updated Files
- generage_coverage.py
   - Updated coverage code to account for outages from ground stations.

## Testing
- test_generage_coverage.py

### DIFF
--- a/imap_processing/tests/ialirt/unit/test_generate_coverage.py
+++ b/imap_processing/tests/ialirt/unit/test_generate_coverage.py
@@ -2,6 +2,7 @@
 
 from datetime import datetime
 
+import numpy as np
 import pytest
 
 from imap_processing.ialirt.generate_coverage import generate_coverage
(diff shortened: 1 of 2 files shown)


imap_processing/tests/ialirt/unit/test_generate_coverage.py
```diff
@@ -32,3 +33,45 @@ def test_generate_coverage(furnish_kernels):
...
+        "naif0012.tls",
+        "pck00011.tpc",
+        "de440s.bsp",
+        "imap_spk_demo.bsp",
+        "earth_1962_240827_2124_combined.bpc",
+    ]
+
+    outages = {
+        "Kiel": [
+            ("2026-09-22T11:50:00.00Z", "2026-09-22T12:10:00Z"),
```
torimarbois, line 55: So the idea is that the mentioned cron job will look for outages in text files, format those in some kind of way, and then pass that into the `generate_coverage` function?

laspsandoval: Yes. I haven't worked out what sds-data-manager is doing exactly, but that's the idea.

torimarbois (approved):

PR #1970, opened 2025-07-21
MAG L1D

maxinelasp:
# Change Summary
This completes most of MAG L1D. 

Remaining work is to output a few additional non-science files that the MAG team requested, to add the quality flag calculation, and a few checks to make sure data is valid before processing.

There are no major changes in the L2 code, only rearranging how things are organized under classes. 

Sorry for the long PR, I was in a time crunch and wanted to get as much done as possible.

## Overview

- Added L1D framework
- Pulled out shared L2 processing pieces into a different superclass for L2 and L1D
- Added Spin averaging calculation and application
- Added gradiometry calculation and application
- Added L1D to CLI

## Testing

- Added some basic unit tests for each piece, but the majority of testing will be done with the MAG team. The current unit tests cover the basic cases for all the inputs.


imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -0,0 +1,572 @@
...
+            self.config.spin_average_application_factor,
+        )
+
+        # we need to be in DSRF for the gradiometry offsets calculation and application
+        self.rotate_frame(ValidFrames.DSRF)
+
+        if self.config.apply_gradiometry:
+            self.gradiometry_offsets = self.calculate_gradiometry_offsets(
+                self.vectors, self.epoch, self.magi_vectors, self.magi_epoch
+            )
```
maxinelasp, line 202: Note to self: add gradiometry application

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -0,0 +1,572 @@
...
+    config: MagL1dConfiguration
+    spin_offsets: xr.Dataset = None
+
+    def __post_init__(self) -> None:
+        """
+        Run all processing steps to generate L1d data.
+
+        This updates class variables to match L1D outputs.
+        """
+        self.vectors, self.magi_vectors = self._calibrate_and_offset_vectors(
```
maxinelasp, line 172: Note: truncate 24

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -0,0 +1,572 @@
...
+            axis=1,
+            arr=magi_vectors,
+            offsets=offsets,
+            is_magi=True,
+        )
+
+        return mago_vectors[:, :3], magi_vectors[:, :3]
+
+    @staticmethod
+    def offset_vector(
```
maxinelasp: Change to calibration_offsets

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -0,0 +1,572 @@
...
+        """
+        offset_value = gradiometry_offsets["gradiometer_offsets"].data
+        offset_value = np.apply_along_axis(
+            np.dot,
+            1,
+            offset_value,
+            gradiometer_factor,
+        )
+
+        print(offset_value.shape)
```
laspsandoval: Replace print with proper logging.

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -0,0 +1,572 @@
...
+            Whether the input vector is from MAGi (True) or MAGo (False).
+
+        Returns
+        -------
+        np.ndarray
+            The offset vector, shape (4,) where the last element is unchanged.
+        """
+        # Offsets are in shape (sensor, range, axis)
+        updated_vector = input_vector.copy().astype(np.int64)
+        range = int(input_vector[3])
```
laspsandoval: Rename range to something like range_idx or range_index to avoid confusion with Python's built-in range.

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -0,0 +1,572 @@
...
+            burst or norm. The same offsets file is applied to all.
+        spin_average_application_factor : np.float64
+            The spin average application factor from the configuration file.
+
+        Returns
+        -------
+        np.ndarray
+            The output vectors with spin offsets applied, shape (N, 3).
+        """
+        if spin_offsets is None:
```
laspsandoval, line 462: spin_offsets can be None when apply_gradiometry is False or in modes other than NORM. Are you certain you need to throw an error here? If so, that's fine. Just wanted to double check.

imap_processing/mag/l2/mag_l2_data.py
```diff
@@ -304,13 +200,97 @@ def truncate_to_24h(self, timestamp: np.datetime64) -> None:
...
+
+        This works by repeatedly calling the function calibrate_vector on the vectors
+        input.
+
+        Parameters
+        ----------
+        vectors : np.ndarray
+            Array of vectors to apply the calibration to. Should be of shape (n, 3)
+            where n is the number of vectors.
+        calibration_matrix : np.ndarray
```
laspsandoval, line 248: Is this shape correct?

This could either be (3, 3, 4) or (3,3) depending on the level.

imap_processing/mag/l2/mag_l2_data.py
```diff
@@ -328,3 +308,69 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
...
+
+        Returns
+        -------
+        np.ndarray
+            Array of vectors with offsets applied. Should be of shape (n, 3).
+        """
+        if vectors.shape[0] != offsets.shape[0]:
+            raise ValueError("Vectors and offsets must have the same length.")
+
+        offset_vectors: np.ndarray = vectors[:, :3] + offsets
```
laspsandoval: If vectors already has shape (n, 3), slicing with vectors[:, :3] will look exactly the same as vectors.

laspsandoval (commented):
Nicely done! Just some minor things and I'll approve.

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: Good catch - this one gets me all the time 😢

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: Yes, if we're applying offsets we need to have spin_offsets. if we're in BURST mode we need to pass the spin offsets in, so they shouldn't be none.

imap_processing/mag/l2/mag_l2_data.py
maxinelasp: Yes, good catch, probably some leftover from when vectors did include range. But here I would never expect them to so I can remove the slice.

laspsandoval (approved):
Nice work! Thanks Maxine. Amazing job.

PR #1977, opened 2025-07-22
Add k-factor to CoDICE L1 datasets

### LINKED ISSUE #1948: Add kfactor as variable to CoDICE L1 datasets
The kfactor is used to convert voltages to energies in L2 data products, so it would be convenient to have this as a variable in the dataset.

bourque:
This PR adds a `k_factor` data variable to CoDICE L1 datasets. This is a constant that will be used to convert voltages to energies in L2 processing.

Closes #1948

### DIFF
--- a/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
@@ -295,7 +295,7 @@ energy_table:
     VAR_TYPE: support_data
 
 k_factor:
-    CATDESC: K Factor
+    CATDESC: K Factor constant that is used to convert voltages to energies
     FIELDNAM: K Factor
     FORMAT: F5.2
     LABLAXIS: K Factor
--- a/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
@@ -294,6 +294,17 @@ energy_table:
     VALIDMAX: 14100.0
     VAR_TYPE: support_data
 
+k_factor:
+    CATDESC: K Factor
+    FIELDNAM: K Factor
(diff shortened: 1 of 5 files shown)


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -294,6 +294,17 @@ energy_table:
     VALIDMAX: 14100.0
     VAR_TYPE: support_data
 
+k_factor:
+    CATDESC: K Factor
```
```suggestion
    CATDESC: K constant factor that will be used to convert voltages to energies.
```

imap_processing/cdf/config/imap_codice_l1b_variable_attrs.yaml
```diff
@@ -295,6 +295,17 @@ energy_table:
     VALIDMAX: 14100.0
     VAR_TYPE: support_data
 
+k_factor:
+    CATDESC: K Factor
```
tech3371: same here as above

imap_processing/codice/codice_l1a.py
```diff
@@ -405,6 +405,12 @@ def define_support_variables(self, dataset: xr.Dataset) -> xr.Dataset:
                 dims = ["epoch"]
                 attrs = self.cdf_attrs.get_variable_attributes("spin_period")
 
+            # THe k-factor is a constant that maps voltages to energies
+            elif variable_name == "k_factor":
```
tech3371, line 409: can this be stored as global variable since it's a one value? But I see why you would want to store as data variable for easier access later. Just something that came to my mind.

tech3371 (approved):
Minor suggestion. But LGTM

imap_processing/codice/codice_l1a.py
bourque: Yeah, I think that would work too, but the CoDICE team prefers to store this in the CDF so that calculations can be done from the L1 file.

PR #1979, opened 2025-07-22
CoDICE: Change energy bin deltas to delta minus and delta plus

### LINKED ISSUE #1960: CoDICE: Energy-binned datasets should have delta minus/plus variables
Currently CoDICE L1a binned datasets (e.g. `hi-omni`) contain variables for the energy bin deltas (e.g. `energy_h_delta`). The CoDICE team would like these deltas to be broken up into delta `minus` and `plus`.

bourque:
This PR changes the defined energy bin deltas to separate delta minus and delta plus values in CoDICE L1a energy-binned datasets, calculated via the geometric mean, as preferred by the CoDICE instrument team.

Closes #1960

### DIFF
--- a/imap_processing/codice/codice_l1a.py
+++ b/imap_processing/codice/codice_l1a.py
@@ -313,7 +313,7 @@ class CoDICEL1aPipeline:
             ``xarray`` dataset for the data product, with added energy variables.
         """
         energy_bin_name = f"energy_{species}"
-        centers, deltas = self.get_hi_energy_table_data(
+        centers, deltas_minus, deltas_plus = self.get_hi_energy_table_data(
             energy_bin_name.split("energy_")[-1]
         )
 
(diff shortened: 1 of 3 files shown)


imap_processing/codice/codice_l1a.py
```diff
@@ -500,22 +508,22 @@ def get_hi_energy_table_data(
...
 
         # Find the centers and deltas of the energy bins
-        centers = np.array(
-            [
-                (energy_table[i] + energy_table[i + 1]) / 2
-                for i in range(len(energy_table) - 1)
-            ]
-        )
-        deltas = energy_table[1:] - centers
+        centers = np.sqrt(energy_table[:-1] * energy_table[1:])
```
tech3371, line 525: To me, this seems like the `delta` instead of center. May be rename this to `delta` and then minus and plus make sense. Also, should this be `deltas_plus = energy_table[1:] + centers`? if not, can you put an example of how delta is calculated?

tech3371 (approved):
I had one question, otherwise looks good to me.

PR #1985, opened 2025-07-23
Add ability to retreive spacecraft spin phase using J2000ns

maxinelasp:
# Change Summary
Extends the spin data interpolation functions to support both Mission Elapsed Time (MET) and J2000 nanoseconds (TTJ2000ns) time formats, addressing time conversion
  issues in MAG L1D processing where data uses TTJ2000ns but existing spin functions expected MET times.

I did use AI to write a lot of this PR (under very close supervision) which is why there are so many comments. 

## Updated Files

  - Enhanced interpolate_spin_data(): Added time_format parameter (defaults to "met" for backward compatibility) to accept both "met" and "j2000ns" time formats
  - Added get_spacecraft_spin_phase_j2000ns(): New function for direct J2000ns time input without conversion overhead
  - Updated _load_spin_data_with_cache(): Added precomputed spin_start_ttj2000ns column to spin data for efficient J2000ns operations
  - Maintained get_spacecraft_spin_phase(): Preserved existing MET-based function with identical behavior


imap_processing/spice/spin.py
```diff
@@ -126,70 +127,100 @@ def _load_spin_data_with_cache(csv_paths: tuple[Path]) -> pd.DataFrame:
     combined_df["spin_start_met"] = (
         combined_df["spin_start_sec_sclk"] + combined_df["spin_start_subsec_sclk"] / 1e6
     )
+    # Convert MET to TTJ2000ns for direct use with MAG data
+    combined_df["spin_start_ttj2000ns"] = met_to_ttj2000ns(
+        combined_df["spin_start_met"]
+    )
```
tmplummer: My main concern with this PR is that this adds a hidden SPICE kernel requirement to these functions. It means that to use this code, one must have the leapsecond and clock kernel furnished. I would advocate for the addition of a `ttj2000ns_to_met()` function to the `time.py` module so that your MAG call would be `spin_phase = get_spacecraft_spin_phase(ttj200ns_to_met(ttj200ns_times))`

imap_processing/spice/spin.py
```diff
@@ -263,12 +294,38 @@ def get_spacecraft_spin_phase(
...
         Spin phase for the input query times.
     """
-    spin_df = interpolate_spin_data(query_met_times)
+    spin_df = interpolate_spin_data(query_met_times, time_format="met")
     if np.asarray(query_met_times).ndim == 0:
         return spin_df["sc_spin_phase"].values[0]
     return spin_df["sc_spin_phase"].values
 
 
+def get_spacecraft_spin_phase_j2000ns(
```
tmplummer: If we are going to support `time_format="j2000ns"` in the `interpolate_spin_data` function, let's keep things consistent and add that keyword to `get_spacecraft_spin_phase()` rather than write a new function.

imap_processing/tests/spice/test_spin.py
```diff
@@ -277,3 +297,84 @@ def test_get_instrument_spin_phase(instrument, fake_spin_data):
...
+    with patch(
+        "imap_processing.spice.spin.met_to_ttj2000ns", side_effect=lambda x: x * 1e9
+    ):
+        spin_df = spin.interpolate_spin_data(query_times, time_format=time_format)
+        np.testing.assert_array_almost_equal(
+            spin_df["sc_spin_phase"].values, expected_phases
+        )
+
+
+def test_interpolate_spin_data_compatibility_and_errors(
```
tmplummer: Nice thorough test coverage!

tmplummer (commented):
As I said in one comment, my main concern is the addition of hidden SPICE kernel requirements that this PR brings into the `get_spin_data()` function. I think that it keeps things a bit more idiomatic to require the `query_time` to be MET and to manually do the conversion in the call to the `get_spacecraft_spin_phase` function.

imap_processing/spice/spin.py
maxinelasp: Yeah, I noticed that too, but I thought that all the instrument processing was using the time kernel anyway. I also was thinking most people would be using ttj2000ns to access spin data, rather than MET (I wasn't actually aware anyone was using MET past L1A, since I thought we were supposed to be entirely in J2000, but it does look like some functions use this method.) 

An alternative could be to check to see if the time kernel is included and only allow for querying off J2000 in that case...

imap_processing/spice/spin.py
maxinelasp: I was trying to avoid affecting existing use cases - the interpolate_spin_data isn't used outside these functions, but `get_spacecraft_spin_phase` is, which is why I created a new function. But, if you think it's ok to just add a keyword to those use cases (I think most of them are yours) then I'm happy to make the change.

imap_processing/spice/spin.py
tmplummer: Looking at the current usage of the `get_spin_data()`, `get_spacecraft_spin_phase()`, and `get_instrument_spin_phase()`, it looks like MAG is the only instrument that doesn't have MET readily available to pass in as an argument. I think that there is some heritage from IBEX of using MET. Hi, for example generally prefers to just work with MET.

The alternative proposed is not a good solution since the spin table data gets cached and the time kernels may be available for one call to that function, but not another.

I still have a strong preference to keep the dependence on spice out of the spin.py module and force the conversion to MET to be done outside the query. This also keeps the function signature cleaner. We went through a similar thing with many of the geometry functions allowing for input/output of radians or degrees and finally settled on sticking to one standard in those functions and doing the conversion outside of there when needed.

maxinelasp: Ok, if MAG is the odd one out then I'm happy to do the conversion for that code only. I thought it was the other way around, I'm surprised so many instruments use MET for higher levels!

imap_processing/tests/spice/test_spin.py
tmplummer, line 1: It looks like the changes to this file are not actually used anymore. `expected` is not used anywhere except in the generation of `expected_without_ttj2000ns`.

tmplummer (approved):
Thanks for digging in to the spice time conversions! This looks good other than removing the changes to the `test_spin.py` file that are not used.

PR #2000, opened 2025-07-31
Return paths to generated CDFs from post_processing

pleasant-menlo:
## Summary
Returns paths to CDFs written during ProcessInstrument post_processing.

## Testing
Added unit test verifying the path returned by write_cdf is returned by post_processing

- George and Ethan

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -18,7 +18,7 @@ import re
 import sys
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import final
+from typing import final, Optional
 
 import imap_data_access
 import numpy as np
@@ -517,7 +517,7 @@ class ProcessInstrument(ABC):
         self,
         processed_data: list[xr.Dataset | Path],
         dependencies: ProcessingInputCollection,
-    ) -> None:
+    ) -> Optional[list[Path]]:
         """
         Complete post-processing.
 
@@ -582,6 +582,7 @@ class ProcessInstrument(ABC):
                 products.append(ds)
 
         self.upload_products(products)
+        return products
 
     @final
     def cleanup(self) -> None:
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -545,8 +545,13 @@ class ProcessInstrument(ABC):
             method.
         dependencies : ProcessingInputCollection
             Object containing dependencies to process.
+
+        Returns
+        -------
+        list[Path]
+        List of paths to CDF files produced.
         """
-        products = []
+        products: list[Path] = []
 
         if len(processed_data) == 0:
             logger.info("No products to write to CDF file.")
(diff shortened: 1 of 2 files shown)


imap_processing/cli.py
```diff
@@ -517,7 +517,7 @@ def post_processing(
         self,
         processed_data: list[xr.Dataset | Path],
         dependencies: ProcessingInputCollection,
-    ) -> None:
+    ) -> Optional[list[Path]]:
```
greglucas: I don't love returning different values here. Could you change this to always return the list. I think if you move up `products = []` to the top, then in the early return case you can just `return products` there as well which would be an empty list.
```suggestion
    ) -> list[Path]:
```

maxinelasp (approved):
Looks good! I saw in your email you mentioned a failing codecov - sometimes the first codecov run fails and then once all the tests finish the re-run succeeds. We only have one test that uses all the remote data to run large/complicated tests, so the partial runs will fail the codecov. So, it looks like the PR is good as far as code coverage goes.

It does look like there are a few pre-commit bugs which should be fixed prior to merging.

greglucas (commented):
Seems fine to me 👍 , shouldn't really impact anything on our end since it was just returning None and we weren't doing anything with it.

imap_processing/cli.py
pleasant-menlo: Thanks for the callout! We've updated this and added corresponding tests in the second commit.

Ethan and George

pleasant-menlo:
> Looks good! I saw in your email you mentioned a failing codecov - sometimes the first codecov run fails and then once all the tests finish the re-run succeeds. We only have one test that uses all the remote data to run large/complicated tests, so the partial runs will fail the codecov. So, it looks like the PR is good as far as code coverage goes.
> 
> It does look like there are a few pre-commit bugs which should be fixed prior to merging.

@maxinelasp - to clarify, are there bugs we are missing in our work and should address, or do you mean prior to this PR? Thanks!

maxinelasp:
@pleasant-menlo If you look here: https://results.pre-commit.ci/run/github/654679818/1753982459.OnCeqyTeTPyVBjRYxc09Zw

There are a few minor changes to docs, typehints, etc that should be fixed before we can merge. Once the "pre-commit.ci" check passes, we can merge this in.

pleasant-menlo:
> @pleasant-menlo If you look here: https://results.pre-commit.ci/run/github/654679818/1753982459.OnCeqyTeTPyVBjRYxc09Zw
> 
> There are a few minor changes to docs, typehints, etc that should be fixed before we can merge. Once the "pre-commit.ci" check passes, we can merge this in.

@maxinelasp Thanks for pointing us in the correct direction on this! Sorry for all of the extra commits while we addressed the issues...

- George and Ethan

maxinelasp:
No problem! We squash before merging so it won't be clear :) Plus, dealing with pre-commit can be challenging...

We're planning on making a release today so this change should be quickly available.

PR #2001, opened 2025-07-31
HIT - implement handling for data spanning midnight

### LINKED ISSUE #1070: HIT L0 - Add handling for packets spanning multiple files
Science frames consist of 20 packets (1 minute). These packets can span midnight. 

- Need a time buffer on the L0 file to include these packets. 
- Use mean epoch time for each science frame to filter dataset for frames that belong in the processing day

### LINKED ISSUE #1867: Add handling for sectored rates that span midnight
A complete set of sectored rates data is 10 minutes and can span the midnight boundary. L1A processing will need to take in some data from the surrounding days to ensure that complete sets are being captured in processing.

- Need a 20 minute buffer on L0 files
- Find complete sets of sectored data
- Find corresponding livetime data that is shifted 10 minutes before sectored data
- Find the mean epoch for each complete set of sectored data
- Filter data for complete sets of sectored data that fall in the day being processed

vmartinez-cu:
This PR implements handling for data spanning midnight. HIT has data that spans midnight in three ways:

1.  A science frame consists of 20 packets that could span midnight. A frame is 1 minute of data
2. A major frame consists of 10 science frames that could span midnight. A major frame is 10 minutes and represents a complete set of sectored rates data.
3. Livetime data associated with the sectored data is 10 minutes before the sectored data's epoch values because sectored data is transmitted 10 minutes after it's collected.

This PR uses mean epoch values for the science frames and major frames to determine which day they belong to for processing. Livetime for sectored data is taken by shifting the epoch range by 10 minutes earlier. 

To reduce the number of epoch ranges being tracked in the L1A counts data product, this product has been split into two.

1. One for standard data which only consists of science frames and therefore can only span midnight by up to a minute
2. One for sectored data which consists of major frames and could span midnight by 10 minutes but also needs the livetime values from the previous 10 minutes to properly calculate rates in L1B. Thus, the sectored product can span midnight by up to 20 minutes. 

Splitting the L1A counts product into two required some updates to L1B to expect two science files instead of one. It also required updating a validation unit test for L1A science data.

Processing expects an L0 file with a 20 minute buffer to capture all these cases and outputs two science products trimmed to their appropriate time ranges that belong in the day being processed.

Closes #1070 
Closes #1867

### DIFF
--- a/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
@@ -23,6 +23,12 @@ imap_hit_l1a_counts:
     Logical_source: imap_hit_l1a_counts
     Logical_source_description: IMAP Mission HIT Instrument Level-1A Count Rates Data.
 
+imap_hit_l1a_counts_sectored:
+    <<: *instrument_base
+    Data_type: L1A_counts>Level-1A Sectored Counts
+    Logical_source: imap_hit_l1a_sectored_counts
+    Logical_source_description: IMAP Mission HIT Instrument Level-1A Sectored Count Rates Data.
+
 imap_hit_l1a_direct-events:
     <<: *instrument_base
     Data_type: L1A_pha>Level-1A Direct Events
--- a/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
@@ -17,11 +17,11 @@ imap_hit_l1a_hk:
     Logical_source: imap_hit_l1a_hk
     Logical_source_description: IMAP Mission HIT Instrument Level-1A Housekeeping Data.
 
-imap_hit_l1a_counts:
+imap_hit_l1a_counts_standard:
     <<: *instrument_base
-    Data_type: L1A_counts>Level-1A Counts
-    Logical_source: imap_hit_l1a_counts
-    Logical_source_description: IMAP Mission HIT Instrument Level-1A Count Rates Data.
+    Data_type: L1A_counts>Level-1A Standard Counts
+    Logical_source: imap_hit_l1a_counts_standard
+    Logical_source_description: IMAP Mission HIT Instrument Level-1A Standard Count Rates Data.
 
 imap_hit_l1a_counts_sectored:
     <<: *instrument_base
--- a/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
@@ -26,7 +26,7 @@ imap_hit_l1a_counts:
 imap_hit_l1a_counts_sectored:
     <<: *instrument_base
     Data_type: L1A_counts>Level-1A Sectored Counts
-    Logical_source: imap_hit_l1a_sectored_counts
+    Logical_source: imap_hit_l1a_counts_sectored
     Logical_source_description: IMAP Mission HIT Instrument Level-1A Sectored Count Rates Data.
 
 imap_hit_l1a_direct-events:
--- a/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
@@ -17,13 +17,13 @@ imap_hit_l1a_hk:
     Logical_source: imap_hit_l1a_hk
     Logical_source_description: IMAP Mission HIT Instrument Level-1A Housekeeping Data.
 
-imap_hit_l1a_counts_standard:
+imap_hit_l1a_counts-standard:
     <<: *instrument_base
     Data_type: L1A_counts>Level-1A Standard Counts
     Logical_source: imap_hit_l1a_counts-standard
     Logical_source_description: IMAP Mission HIT Instrument Level-1A Standard Count Rates Data.
 
-imap_hit_l1a_counts_sectored:
+imap_hit_l1a_counts-sectored:
     <<: *instrument_base
     Data_type: L1A_counts>Level-1A Sectored Counts
     Logical_source: imap_hit_l1a_counts-sectored
--- a/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
@@ -20,13 +20,13 @@ imap_hit_l1a_hk:
 imap_hit_l1a_counts_standard:
     <<: *instrument_base
     Data_type: L1A_counts>Level-1A Standard Counts
-    Logical_source: imap_hit_l1a_counts_standard
+    Logical_source: imap_hit_l1a_counts-standard
     Logical_source_description: IMAP Mission HIT Instrument Level-1A Standard Count Rates Data.
 
 imap_hit_l1a_counts_sectored:
     <<: *instrument_base
     Data_type: L1A_counts>Level-1A Sectored Counts
-    Logical_source: imap_hit_l1a_counts_sectored
+    Logical_source: imap_hit_l1a_counts-sectored
     Logical_source_description: IMAP Mission HIT Instrument Level-1A Sectored Count Rates Data.
 
 imap_hit_l1a_direct-events:
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -789,11 +789,11 @@ class Hit(ProcessInstrument):
 
         dependency_list = dependencies.processing_input
         if self.data_level == "l1a":
-            # 1 science files and 2 spice files
-            if len(dependency_list) > 3:
+            # 1 science file and 2 spice files
+            if len(dependency_list) > 2:
                 raise ValueError(
                     f"Unexpected dependencies found for HIT L1A:"
-                    f"{dependency_list}. Expected only one dependency."
+                    f"{dependency_list}. Expected only 2 dependencies."
                 )
             # process data to L1A products
             science_files = dependencies.get_file_paths(source="hit", descriptor="raw")
@@ -801,25 +801,29 @@ class Hit(ProcessInstrument):
 
         elif self.data_level == "l1b":
             data_dict = {}
-            # TODO: Sean removed the file number error handling to work with the
-            #  new SPICE dependencies for SIT-4. Need to review and make changes
-            #  if needed.
             l0_files = dependencies.get_file_paths(source="hit", descriptor="raw")
             l1a_files = dependencies.get_file_paths(source="hit", data_type="l1a")
-            if len(l0_files) > 0:
+            if len(l0_files) == 1:
                 # Add path to CCSDS file to process housekeeping
                 data_dict["imap_hit_l0_raw"] = l0_files[0]
             else:
+                # 1 science file
+                if len(l1a_files) > 1:
+                    raise ValueError(
+                        f"Unexpected dependencies found for HIT L1B:"
+                        f"{l1a_files}. Expected only one dependency."
+                    )
                 # Add L1A dataset to process science data
                 l1a_dataset = load_cdf(l1a_files[0])
                 data_dict[l1a_dataset.attrs["Logical_source"]] = l1a_dataset
             # process data to L1B products
             datasets = hit_l1b(data_dict)
         elif self.data_level == "l2":
+            # 1 science files and 4 ancillary files
             if len(dependency_list) != 5:
                 raise ValueError(
                     f"Unexpected dependencies found for HIT L2:"
-                    f"{dependency_list}. Expected only one dependency."
+                    f"{dependency_list}. Expected only five dependencies."
                 )
             # Add L1B dataset to process science data
             science_files = dependencies.get_file_paths(
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -797,7 +797,7 @@ class Hit(ProcessInstrument):
                 )
             # process data to L1A products
             science_files = dependencies.get_file_paths(source="hit", descriptor="raw")
-            datasets = hit_l1a(science_files[0])
+            datasets = hit_l1a(science_files[0], self.start_date)
 
         elif self.data_level == "l1b":
             data_dict = {}
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -789,11 +789,12 @@ class Hit(ProcessInstrument):
 
         dependency_list = dependencies.processing_input
         if self.data_level == "l1a":
-            # 1 science file and 2 spice files
+            # Two inputs - L0 and SPICE
             if len(dependency_list) > 2:
                 raise ValueError(
                     f"Unexpected dependencies found for HIT L1A:"
-                    f"{dependency_list}. Expected only 2 dependencies."
+                    f"{dependency_list}. Expected only 2 dependencies, "
+                    f"L0 and time kernels."
                 )
             # process data to L1A products
             science_files = dependencies.get_file_paths(source="hit", descriptor="raw")
--- a/imap_processing/hit/l1a/hit_l1a.py
+++ b/imap_processing/hit/l1a/hit_l1a.py
@@ -2,6 +2,7 @@
 
 import logging
 from datetime import datetime
+from pathlib import Path
 from typing import Optional
 
 import numpy as np
@@ -36,7 +37,7 @@ logger = logging.getLogger(__name__)
 fillval = -9223372036854775808
 
 
-def hit_l1a(packet_file: str, packet_date: Optional[str]) -> list[xr.Dataset]:
+def hit_l1a(packet_file: Path, packet_date: Optional[str]) -> list[xr.Dataset]:
     """
     Will process HIT L0 data into L1A data products.
 
--- a/imap_processing/hit/l1a/hit_l1a.py
+++ b/imap_processing/hit/l1a/hit_l1a.py
@@ -45,7 +45,8 @@ def hit_l1a(packet_file: str, packet_date: Optional[str]) -> list[xr.Dataset]:
     packet_file : str
         Path to the CCSDS data packet file.
     packet_date : str, optional
-        The date of the packet data in 'YYYYMMDD' format. This is used to filter.
+        The date of the packet data in 'YYYYMMDD' format. This is used to filter
+        data to the correct processing day since L0 will have a buffer around midnight.
 
     Returns
     -------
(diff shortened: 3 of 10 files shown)


imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
```diff
@@ -17,11 +17,17 @@ imap_hit_l1a_hk:
     Logical_source: imap_hit_l1a_hk
     Logical_source_description: IMAP Mission HIT Instrument Level-1A Housekeeping Data.
 
-imap_hit_l1a_counts:
+imap_hit_l1a_counts_standard:
```
sdhoyt: Might be good to make this `counts-standard` with a `-` instead of an `_` to keep it consistent with the other product variable names in the yaml and the logical source. Ex: `direct-events`

imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
```diff
@@ -17,11 +17,17 @@ imap_hit_l1a_hk:
...
     <<: *instrument_base
-    Data_type: L1A_counts>Level-1A Counts
-    Logical_source: imap_hit_l1a_counts
-    Logical_source_description: IMAP Mission HIT Instrument Level-1A Count Rates Data.
+    Data_type: L1A_counts>Level-1A Standard Counts
+    Logical_source: imap_hit_l1a_counts-standard
+    Logical_source_description: IMAP Mission HIT Instrument Level-1A Standard Count Rates Data.
+
+imap_hit_l1a_counts_sectored:
```
sdhoyt: Same comment as above with `-` instead of `_`

imap_processing/cli.py
```diff
@@ -789,37 +789,41 @@ def do_processing(
 
         dependency_list = dependencies.processing_input
         if self.data_level == "l1a":
-            # 1 science files and 2 spice files
-            if len(dependency_list) > 3:
+            # 1 science file and 2 spice files
+            if len(dependency_list) > 2:
```
sdhoyt, line 793: Why is this only 2? do the two SPICE files only found as one in the list?

imap_processing/cli.py
```diff
@@ -789,37 +789,41 @@ def do_processing(
```
sdhoyt, line 789: I really like how this file is organized!

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -308,25 +336,262 @@ def add_cdf_attributes(
...
+
+    # Trim livetime to the size of the sectored data but shifted 10 minutes earlier.
+    filtered_dataset = subset_livetime(filtered_dataset)
+
+    return filtered_dataset
+
+
+def update_livetime_coord(sectored_dataset: xr.Dataset) -> xr.Dataset:
+    """
+    Update livetime_counter to use a new epoch coordinate.
```
sdhoyt, line 454: Add comment for line 399 to docstring

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -308,25 +336,262 @@ def add_cdf_attributes(
...
+    end_idx = np.where(epoch_livetime_vals == epoch_vals[-1])[0][0]
+
+    if start_idx < 10:
+        logger.error(
+            "Start or end indices for livetime are less than 10. "
+            "This indicates that the dataset is too small to shift livetime correctly."
+        )
+        raise ValueError("Start or end indices for livetime are out of range.")
+
+    # Compute shifted indices
```
sdhoyt: Compute shifted indices 10minutes

sdhoyt: And maybe remind people that the time between epochs is always 1 minute

imap_processing/cli.py
sdhoyt: I meant to put that comment on the hit_l1a.py file. Not sure how it ended up on the cli.py file

imap_processing/cdf/config/imap_hit_global_cdf_attrs.yaml
vmartinez-cu: good catch!

imap_processing/cli.py
vmartinez-cu: this was confusing to me too. The set of spice files counts as one dependency in the list

tech3371: I can help add some clarification here. This is 2 because, we combine all metakernel files into one dependency input for easier SPICE loading. There, it's actually two dependency inputs, L0 and spice.

imap_processing/cli.py
```diff
@@ -789,37 +789,41 @@ def do_processing(
...
         dependency_list = dependencies.processing_input
         if self.data_level == "l1a":
-            # 1 science files and 2 spice files
-            if len(dependency_list) > 3:
+            # 1 science file and 2 spice files
+            if len(dependency_list) > 2:
                 raise ValueError(
                     f"Unexpected dependencies found for HIT L1A:"
-                    f"{dependency_list}. Expected only one dependency."
+                    f"{dependency_list}. Expected only 2 dependencies."
```
tech3371:
```suggestion
                    f"{dependency_list}. Expected only 2 dependencies,"
                    "L0 and time kernels."
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -28,41 +37,50 @@
...
-    datasets_by_apid = get_datasets_by_apid(packet_file)
+    if not packet_date:
+        raise ValueError("Packet date is required for processing L1A data.")
+    else:
+        # Unpack ccsds file to xarray datasets
+        datasets_by_apid = get_datasets_by_apid(str(packet_file))
 
-    # Create the attribute manager for this data level
-    attr_mgr = get_attribute_manager("l1a")
+        # Create the attribute manager for this data level
```
tech3371: should this tab be undone? It can be outside that else statement still.

imap_processing/hit/l1b/hit_l1b.py
```diff
@@ -106,18 +112,29 @@ def process_science_data(
...
-        ),
-        "imap_hit_l1b_sectored-rates": process_sectored_rates_data(
-            l1a_counts_dataset, livetime
-        ),
-    }
+    l1b_datasets = {}
+
+    if "imap_hit_l1a_counts-standard" in l1a_counts_dataset.attrs["Logical_source"]:
+        # Process counts data to L1B datasets
+        l1b_datasets.update(
```
greglucas: I think setting these directly is a little more standard and easier to follow than creating a dictionary and then merging.
```python
l1b_datasets["imap_hit_l1b_standard-rates"] = process_standard_rates_data(
                    l1a_counts_dataset, livetime
                )
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -308,25 +335,268 @@ def add_cdf_attributes(
...
+    window_size = len(MOD_10_PATTERN)
+
+    if mod_vals.size < window_size:
+        logger.warning(
+            "Mod 10 array is smaller than the required window size for "
+            "pattern matching."
+        )
+        return np.array([], dtype=int)
+    # Use sliding windows to find pattern matches
+    sw_view = np.lib.stride_tricks.sliding_window_view(mod_vals, window_size)
```
tech3371, line 362: don't know this approach. interesing.

imap_processing/cli.py
sdhoyt: Ah ok, gotcha. Can you add a comment so that's clear?

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -308,25 +335,268 @@ def add_cdf_attributes(
...
+
+    # Modify livetime_counter to use a new epoch coordinate
+    # that is aligned with the original epoch dimension. This
+    # ensures that livetime doesn't get filtered when the original
+    # epoch dimension is filtered for complete sets.
+    sectored_counts_dataset = update_livetime_coord(sectored_counts_dataset)
+
+    # Identify 10-minute intervals of complete sectored counts.
+    bin_size = 10
+    mod_10: np.ndarray = sectored_counts_dataset.hdr_minute_cnt.values % 10
```
tech3371, line 407: Can you add a comment about what kind of data is in this variable that we are taking mod?

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -308,25 +335,268 @@ def add_cdf_attributes(
...
+    ----------
+    mod_vals : np.ndarray
+        1D array of mod 10 values from the hdr_minute_cnt field in the L1A counts data.
+
+    Returns
+    -------
+    np.ndarray
+        Indices in mod_vals where the complete pattern [0, 1, ..., 9] starts.
+    """
+    window_size = len(MOD_10_PATTERN)
```
tech3371: Does each mod values correspond to each species? It seems like each mod value correspond to science frame. Is that right? If so, can we rename it to `SCI_FRAME_MOD_10_PATTERN` or something that indicates it's science frame pattern?

tech3371 (commented):
I looked at l1a code on high level. Please let me know if you want me to look at anything in detail.

Thank you for making those changes. It seems like it caused lots of changes. Nice work!

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -28,41 +37,51 @@
...
-    attr_mgr = get_attribute_manager("l1a")
-
-    l1a_datasets = []
-
-    # Process l1a data products
-    if HitAPID.HIT_HSKP in datasets_by_apid:
-        logger.info("Creating HIT L1A housekeeping dataset")
-        l1a_datasets.append(
-            process_housekeeping_data(
+    if not packet_date:
```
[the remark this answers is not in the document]
greglucas: I agree with this. You don't need this check I don't think since it is an argument with no default.

This will also avoid all the extra tab you did too. You don't need the else: block here because the if is raising. Overall, just get rid of this check though.

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -102,7 +120,16 @@ def subcom_sectorates(sci_dataset: xr.Dataset) -> xr.Dataset:
     sci_dataset : xarray.Dataset
         Xarray dataset with sectored rates data organized by species.
     """
-    updated_dataset = sci_dataset.copy()
+    # Initialize the dataset with the required variables
```
greglucas, line 124: Can you do the doubly nested variable syntax here or select them rather than creating a new dataset from scratch?

```
updated_dataset = sci_dataset[["sectorates", ...]]
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -308,25 +335,268 @@ def add_cdf_attributes(
...
+
+    if mod_vals.size < window_size:
+        logger.warning(
+            "Mod 10 array is smaller than the required window size for "
+            "pattern matching."
+        )
+        return np.array([], dtype=int)
+    # Use sliding windows to find pattern matches
+    sw_view = np.lib.stride_tricks.sliding_window_view(mod_vals, window_size)
+    matches = np.all(sw_view == MOD_10_PATTERN, axis=1)
```
greglucas: I find the MOD_10_PATTERN a bit confusing because it is really just a continuous range of numbers it looks like. What about doing directly the arange() call here? where window_size = 10 would be defined above.
```suggestion
    matches = np.all(sw_view == np.arange(window_size), axis=1)
```

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: Each mod value corresponds to a species and energy bin for a science frame. I can add a better comment for the constant

```
MOD_10_MAPPING = {
    0: {"species": "h", "energy_min": 1.8, "energy_max": 3.6},
    1: {"species": "h", "energy_min": 4, "energy_max": 6},
    2: {"species": "h", "energy_min": 6, "energy_max": 10},
    3: {"species": "he4", "energy_min": 4, "energy_max": 6},
    4: {"species": "he4", "energy_min": 6, "energy_max": 12},
    5: {"species": "cno", "energy_min": 4, "energy_max": 6},
    6: {"species": "cno", "energy_min": 6, "energy_max": 12},
    7: {"species": "nemgsi", "energy_min": 4, "energy_max": 6},
    8: {"species": "nemgsi", "energy_min": 6, "energy_max": 12},
    9: {"species": "fe", "energy_min": 4, "energy_max": 12},
}
```

imap_processing/hit/l1a/hit_l1a.py
```diff
@@ -308,25 +335,268 @@ def add_cdf_attributes(
...
+
+    Returns
+    -------
+    xarray.Dataset
+        The filtered dataset containing data within the processing day.
+    """
+    processing_day = datetime.strptime(packet_date, "%Y%m%d").strftime("%Y-%m-%d")
+
+    # Filter dataset by epoch
+    epoch = dataset["epoch"].values if epoch_array is None else epoch_array
```
greglucas: Rather than None as the default, do you want to pass in the name of the epoch variable instead?
Then this could be: `epoch = dataset[epoch_name].values`

greglucas (commented):
Looks OK, but there is a lot of diff here and I don't think all of it was actually changed but some of it just whitespace/tab changes.

One thing I'm a bit concerned about is how much there is going on in this PR. I thought that previously you had all of the times aligned and subsetted into the frames as necessary within each day, but this now seems to refactor/update that work? I thought that most of the work would have been just the new "subset_to_processing_day" function where you do the `isel()` based on the met.

vmartinez-cu:
> Looks OK, but there is a lot of diff here and I don't think all of it was actually changed but some of it just whitespace/tab changes.
> 
> One thing I'm a bit concerned about is how much there is going on in this PR. I thought that previously you had all of the times aligned and subsetted into the frames as necessary within each day, but this now seems to refactor/update that work? I thought that most of the work would have been just the new "subset_to_processing_day" function where you do the `isel()` based on the met.

My original code included grouping of science frames (done in L1A) and grouping of major frames for sectored data (done in L1B). This was sufficient for the single day of test data I had, but it didn't handle any data spanning midnight. Using a dataset provided by HIT spanning multiple days, I was able to begin tackling data spanning midnight.

My initial approach was to have a 1 minute buffer in L0 and just use isel to select the frames for the processing day in L1A. Then in L1B processing, take in 3 L1A products representing 3 days where I would handle the sectored data spanning midnight by combining the datasets and subsetting for the correct time range. I implemented this work locally but didn't put create a PR after learning that infrastructure couldn't support a dependency of 3 days of L1A products for L1B processing. 

This internal change in approach meant that all grouping of data, plus filtering by processing day for the different time ranges, needed to be handled in L1A to avoid including unnecessary data in higher products. I was able to move code from L1B that grouped 10 minutes of sectored data to L1A, but the refactor resulted in other updates as well (splitting L1A counts into two products, updating L1B dependencies accordingly, updating tests and validation). It was a significant amount of work which is why I was concerned about handling all of this in processing rather than on the infrastructure side.

Can you expand on the whitespaces/tab changes you're seeing? Are you able to see the code updates I made?

greglucas:
> Can you expand on the whitespaces/tab changes you're seeing? Are you able to see the code updates I made?

Related to this comment:
https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2001#discussion_r2248496925
where there is now an if/else added that moves everything over to the right and makes it look like more changes than there really are. I still vote for removing that if, or at least getting rid of the else block.

> My original code included grouping of science frames (done in L1A) and grouping of major frames for sectored data (done in L1B). This was sufficient for the single day of test data I had, but it didn't handle any data spanning midnight. Using a dataset provided by HIT spanning multiple days, I was able to begin tackling data spanning midnight.

OK, I think I'm following now. Thank you for writing that all out! My concern is that this is now doing l1b processing steps in l1a in a sense? I think I was assuming that all the subsetting would be done in l1b+ not necessarily in l1a. So you may have duplicates now across files in l1a (that is OK though and is the same as the packet files in L0 we are delivering) because you don't know whether that product would be used in day1 or day2 until you get to l1b.

imap_processing/hit/l1a/hit_l1a.py
vmartinez-cu: The name of the epoch variable is the same in all datasets (i.e. "epoch"). However, for the sectored dataset, I need to filter by an array of mean epoch values for sets of sectored data over 10 minute intervals. That array is not a coordinate in the sectored dataset since the data still needs a per minute epoch coordinate, but it is used for filtering because the mean values is repeated 10 times to match the size of the epoch coordinate in the dataset which is per minute. The function then uses the indices of mean epochs that fall in the processing day to filter the dataset on it's "epoch" coordinate. What I think I'll do, is replace the None default with an array of epoch values (i.e. dataset["epoch"].values or mean_epoch_array). Then the function will always filter datasets on the epoch coordinate, but will use the epoch array of values passed into the function for finding indices where values are in the processing day.

vmartinez-cu:
> OK, I think I'm following now. Thank you for writing that all out! My concern is that this is now doing l1b processing steps in l1a in a sense? I think I was assuming that all the subsetting would be done in l1b+ not necessarily in l1a. So you may have duplicates now across files in l1a (that is OK though and is the same as the packet files in L0 we are delivering) because you don't know whether that product would be used in day1 or day2 until you get to l1b.

My first approach was handling subsetting of sectored data in l1b, assuming that only 1 minute buffers was used for L0 in l1a processing to sufficiently cover all the non-sectored science data. Since infrastructure isn't able to pass 3 days of l1a products to my l1b processing pipeline, I don't see a way to isolate the subsetting to l1b. Not passing 3 days of l1a products to l1b means that a larger L0 buffer of 20 minutes is needed - with only up to a one minute buffer needed for most of the data, and up to 20ish minutes needed for sectored data. To avoid containing unnecessary data that don't belong in the processing day in l1a (i.e. 19 minutes of extra data for most of the science data and 5 or more minutes of extra data for the sectored data), all the subsetting needs to take place in l1a. The result then is only the data that belongs in the processing day is in the l1a products and the higher level products don't need to do any clean up of extra data. It also avoids duplicate data in l1a since the buffered L0 data is trimmed to exactly what is needed for the different science data. Happy to explain this further offline since it's a complicated problem to describe : )

imap_processing/hit/l1a/hit_l1a.py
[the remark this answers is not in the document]
vmartinez-cu: I think I need to keep the check. Packet date is either a string or None when passed in from cli.py. If it's None, then an error should be raised because l1a processing can't proceed without the date since it's needed to filter data in the L0 buffer that spans into the previous and next day.

sdhoyt (approved):

PR #2007, opened 2025-08-01
Fix running tests on Windows

pleasant-menlo:
## Overview

`get_test_kernels_to_load` failed to run on Windows because it assumed "/" for the path separator.

## Change Summary

Get the correct separator for the system using `os.path.sep`.

## New Dependencies

None

## New Files

None

## Deleted Files

None

## Updated Files
conftest.py: use (escaped) `os.path.sep` instead of a hardcoded "/".

## Testing
None

### DIFF
--- a/imap_processing/tests/conftest.py
+++ b/imap_processing/tests/conftest.py
@@ -265,7 +265,7 @@ def get_test_kernels_to_load(template_path, kernel_dir_path):
                     break
                 else:
                     slash_positions = np.array(
-                        [m.start() for m in re.finditer("/", kernel)]
+                        [m.start() for m in re.finditer(re.escape(os.path.sep), kernel)]
                     )
                     stop_idx = (
                         slash_positions[slash_positions < max_line_length - 1].max() + 1


greglucas:
@lacoak21 and @subagonsouth , is this no longer needed after you removed the ensure_spice work? I tried looking at the conflicts and it looks like this function has been removed altogether now.

Maybe @pleasant-menlo can confirm whether the latest `dev` still works for you all on Windows now?

lacoak21:
> @lacoak21 and @subagonsouth , is this no longer needed after you removed the ensure_spice work? I tried looking at the conflicts and it looks like this function has been removed altogether now.
> 
> Maybe @pleasant-menlo can confirm whether the latest `dev` still works for you all on Windows now?

Yes this fixture has been removed. Hopefully rerunning with the latest dev will pass.

greglucas:
OK, lets close this for now and open up another issue/PR if this still doesn't work because it will need a different fix.

pleasant-menlo:
Confirming that `dev` works for us on Windows. Thanks!

PR #2008, opened 2025-08-01
Ultra l1b - ebins

### LINKED ISSUE #1952: Ultra l1b - ebins - needs to be determined (requires luts)
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Added species bins to Ultra to check against the FSW.

## Updated Files
- ultra_l1b_extended.py
   - Added a validity flag for PH and SSD
   - Implemented the species lookup
- lookup_utils.py
   - Added a lookup table for the species
- de.py
   - Called on functions in ultra_l1b_extended.py

## Deleted Files (Old lookup tables not being used)
- dps_exposure_helio_45_E1.cdf
- dps_exposure_helio_45_E12.cdf
- dps_exposure_helio_45_E24.cdf

## Testing
external_test_data_config.py
conftest.py
test_lookup_utils.py
test_ultra_l1b_extended.py


imap_processing/ultra/l1b/de.py
```diff
@@ -174,16 +176,34 @@ def calculate_de(
...
     etof[ph_indices], xc[ph_indices] = get_coincidence_positions(
         de_dataset.isel(epoch=ph_indices),
         t2[ph_indices],
         f"ultra{sensor}",
         ancillary_files,
     )
+    backtofvalid = is_back_tof_valid(
+        de_dataset,
+        xf,
+        "ultra45",
```
greglucas: Do these need to be {sensor}?

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -348,3 +348,60 @@ def get_ph_corrected(
...
+    -------
+    ebins : NDArray
+        Energy bins from the lookup table.
+    """
+    with open(ancillary_files[lut]) as f:
+        all_lines = f.readlines()
+        pixel_text = "".join(all_lines[4:])
+
+    lut_array = np.fromstring(pixel_text, sep=" ", dtype=int).reshape((2048, 4096))
+    if lut == "l1b-tofxph":
```
greglucas, line 388: It might be good to add a quick note about energy being reversed here and ctof being reversed below.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -1194,29 +1215,172 @@ def determine_ebin_ssd(
...
+
+    steep_mask = path_length < get_image_params(
+        "PathSteepThresh", sensor, ancillary_files
+    )
+    medium_mask = (
+        path_length >= get_image_params("PathSteepThresh", sensor, ancillary_files)
+    ) & (path_length < get_image_params("PathMediumThresh", sensor, ancillary_files))
+    flat_mask = path_length >= get_image_params(
+        "PathMediumThresh", sensor, ancillary_files
+    )
```
greglucas: Since you are re-using some of these values it might be good to bring them out so you are only reading/loading these files once.

```python
steep_path_length = get_image_params("PathSteepThresh", sensor, ancillary_files)
medium_path_length = get_image_params("PathMediumThresh", sensor, ancillary_files)

steep_mask = path_length < steep_path_length
...
```

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -1194,29 +1215,172 @@ def determine_ebin_ssd(
...
+    etof_offset1 = get_image_params("eTOFOff1", sensor, ancillary_files)
+    etof_offset2 = get_image_params("eTOFOff2", sensor, ancillary_files)
+    etof_slope1 = get_image_params("eTOFSlope1", sensor, ancillary_files)
+    etof_slope2 = get_image_params("eTOFSlope2", sensor, ancillary_files)
+
+    t1 = (etof - etof_offset1) * etof_slope1 / 1024
+    t2 = (etof - etof_offset2) * etof_slope2 / 1024
+
+    condition_1 = (diff_x >= t1) & (diff_x <= t2)
+    condition_2 = (diff_x >= -t2) & (diff_x <= -t1)
```
greglucas, line 1388: Is this the same as using the absolute value? `(abs(diff_x) >= t1) & (abs(diff_x) <= t2`

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -161,7 +164,7 @@ def get_front_y_position(
 
 def get_ph_tof_and_back_positions(
     de_dataset: xarray.Dataset, xf: np.ndarray, sensor: str, ancillary_files: dict
-) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
```
greglucas: Remembering what is returned in what order is getting a bit tricky here with so many returns and putting tof first, but tofx and tofy last now. What about returning a namedtuple or something like that instead so you can be explicit with what you're getting later and only return one object. This can be a follow-up.

imap_processing/ultra/l1b/de.py
laspsandoval: Yes! Very nice catch. Thanks.

imap_processing/ultra/l1b/ultra_l1b_extended.py
laspsandoval: yep!

laspsandoval: Actually because of the | used below it isn't always the same.

greglucas (approved):

PR #2010, opened 2025-08-04
MNT: Add validation folders to gitignore

greglucas:
# Change Summary

## Overview

I was running the latest test suite locally and there were remote files that were added to my repository history when downloaded. This adds a general gitignore line for any validation files. The updates were to ignore: `mag/validiation` and `swe/l1_validation` and `swe/l2_validation` folders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,6 +178,7 @@ cython_debug/
 # Data that is downloaded
 data/
 test_data/
+*validation/
 
 # Ignore specific SPICE kernels that get downloaded from NAIF automatically for tests
 # marked with @pytest.mark.external_kernel


maxinelasp:
Most of my MAG validation files are local to github and small, there's only a few that are actually downloaded. I can move those into a different folder or add them specifically to the gitignore, but overall the MAG test files are small and fast to run so I would prefer to leave them in the local runs.

greglucas:
This doesn't remove them. It just gitignores them so they don't show up accidentally if someone does a `git add .` If you want them controlled you should add them to the repository. Otherwise if they are remote, this still works fine and brings them down just removes them from the `git status` logs.

maxinelasp:
It kind of goes against the spirit of the gitignore to have some files in it that are supposed to be tracked in my opinion. It's confusing. I'd much prefer specifically excluding subfolders or specific files.

greglucas:
I am not following what you mean. If you want them tracked they should get committed into the git repository. I agree we could have git lfs which would be nicer, but we don't have that. This is the same pattern as everything above this (see l.178 above), it is just that MAG and SWE called their folders "*validation" rather than "test_data" so it wasn't ignored.

tech3371 (approved):

maxinelasp:
Most of mag/validation/* does not have downloaded data in it. There are only a few files in there that are too big to be included in the repo. So, I think it's confusing that most of the folder should be committed, but it's included in the gitignore.

maxinelasp:
I also don't actually think you can add things manually if they are included in the gitignore. Instead, I would need to whitelist most of that folder to allow updates. It's more annoying to do that than to just explicitly add only the downloaded files to .gitignore.

greglucas:
> I also don't actually think you can add things manually if they are included in the gitignore. Instead, I would need to whitelist most of that folder to allow updates. It's more annoying to do that than to just explicitly add only the downloaded files to .gitignore.

You can definitely do `git add my_file.txt` and that will take precedence over the gitignore entry.

If you have an issue with this, please add the files to git or add them individually to gitignore. As it stands, this is an annoyance for an external developer to not have a clean repository after running `pytest`. That shouldn't add files to your repository locally that haven't been ignored or tracked in some other way.

maxinelasp:
I don't have an issue with adding the externally downloaded files to the .gitignore, I just have an issue with bulk-adding the entire validation file. If you want, you can close this PR and I can open one with the specific files ignored. You also need to specifically mark if you want to use external data files, so it's not that any person running `pytest` would get these files.

greglucas:
Feel free to open a separate PR. When I am on current dev branch after running pytest this is what is in my `git status` output:
image

Those should not be polluting my local workspace and should either be tracked or ignored. I chose to ignore validation in general and you can manually add explicit files to override it. If you want to take the time to individually ignore that is OK, I just didn't want to do that route.

maxinelasp:
Those files should not be downloaded unless you're specifically marking for external tests - so there must be something wrong with the way that's configured. I think Matthew may have updated it a little while ago, so it's possible the MAG files broke in that case - let me take a look. I can add them specifically to .gitignore to keep this problem from happening again. 

None of the other external data files are in the gitignore so that system isn't working for MAG. I remember trying to fix it when Matthew put some updates in a little while back but it looks like it's still downloading files regardless of if you're marking external data as true. I'll open a PR to fix that, and add all the external data files to the gitignore as specific files.

maxinelasp:
I see, most of the other external data files go into folders named data/ and then that is included in the .gitignore. That's why they aren't included here. I'll add the MAG files specifically to the gitignore - sorry about that! I still don't think it should be running unless you specifically mark external data as being downloaded so I'll still look at that and open a new PR.

tech3371:
All, I think Greg it try to solve this concerns. When I run test of everyone, it calls to download new files that Git thinks I want to commit but I don't. Is that right?
Screenshot 2025-08-06 at 1 40 08 PM

maxinelasp:
@tech3371 Yes, we kind of moved discussion over to another ticket: #2012. I think ultimately we should not implement either PR... I was going to close mine and take another try

tech3371:
> @tech3371 Yes, we kind of moved discussion over to another ticket: #2012. I think ultimately we should not implement either PR... I was going to close mine and take another try

@maxinelasp That sounds good. Should we close PR that is not relevant now.

greglucas:
Closing in favor of #2028

PR #2012, opened 2025-08-04
Fix external test data gitignore

maxinelasp:
# Change Summary
Having data/ included in the gitignore was disguising some issues with our external test data system. This explicitly adds only the external test files to .gitignore, while also correcting some of the issues that were disguised by the broad ignore. This still ignores the top-level data/ directory which is downloaded automatically when the CLI runs.

## Overview

- remove generic data/ ignore
- add specific ignores for only the external data
- add missing data files that are not automatically downloaded
- Added some ultra files to the external test files download step

## Testing

added in some new test files that were previously excluded by the gitignore or by accident.


tmplummer:
Can you elaborate on what the  disguised "issues with our external test data system" were?

I personally prefer to have files that I put in instrument specific `data` directory ignored by default because it forces me think about adding those files rather than being picked up by the often abused `git add -all` command. It acts as another guard against the accidental addition of large testing files.

greglucas:
I personally disagree with this and prefer the implementation in #2010 (it would be nice to crosslink these right away in the description and say this is an alternative approach) and I think this is adding more cognitive load to people to add files to a gitignore file as well. The gitignore should be rarely touched IMO.

> while also correcting some of the issues that were disguised by the broad ignore.

I don't think there were any issues? Could you elaborate. I think this is just an inverted way of thinking about the problem. Whether you explicitly ignore or explicitly add a file.

I won't block this, but won't approve either.

maxinelasp:
I agree that this adds maintenance burden.  

I really dislike having broad .gitignores because it completely hides files from the user, making it really easy to miss adding updates to files that should be tracked in git but are not. If, for example, a new developer came on, it's also confusing to have something as generic as data/ be untracked in the entire repo - that would not be behavior I expect or would think to check for. I'm surprised that this is counter to how other people on the team think about it - in other teams, the gitignore was treated as something rarely updated that should only have files that should _never_ be tracked. 

Git already has two checks against adding these big files - first, that they are not tracked by default, and the user needs to add them specifically. Two, that the pre-commit checks will block the commit of any large files. So you already need to explicitly add a file - but putting it in the gitignore completely hides it as well. Then, we also check during code review as well.

I don't like the way that the current large download system works for files. It's very confusing and prone to errors that the files are downloaded next to the rest of the data in most cases (including with MAG) which makes it hard to create an obvious gitignore (for example, downloaded_data/). Unfortunately it's very entrenched at this point. 

Adding broad files to the gitignore really breaks my workflow. It's challenging to remember what files I added when they don't turn up in `git status` and I really tend to view .gitignore as a hard "do not add". I don't think asking developers to add files with a `-f` flag should be a supported workflow (if you do git add on a file that is in the .gitignore, git blocks you and you have to force it.) I am open to adding more instrument-specific data/ ignores like Tim suggested, but I do think it is needlessly confusing for future developers. 

For the issues - I found some Ultra files which were not included in the download data. However, I think these were actually just moved and left behind in my local environment. More issues with the automatic downloader! 

So, my ideal fix would be updating the automatic download to download files into a specific folder with only downloaded files in it which can be added to "gitignore", but unfortunately that would require updating a lot of tests and at 50% time on IMAP I don't have the bandwidth to do it. Alternatively, we could gitignore specific instrument data/ directories (eg hi/data/). Or, we could go back to the previous method and I will gitignore specific MAG files. 

Ultimately, I think the main issue is that the automatic downloader downloads all test data by default and that these downloaded files aren't special or marked in any way. Maybe a shorter fix would be to add "downloaded_" in front of each file name? That would still require a lot of test updates, but could probably cover a  lot of them with the magic refactor button.

tmplummer:
I'm not sure how to reconcile these two approaches. I can certainly appreciate that the broad ignore hampers some developer's workflows. I also prefer to not need to remember that I need to add an external file name in more than one location in the repo.

I think that the ideal solution would be to have a single magic config file that provides 3 things:
- gitignore files listed
- Tells the downloader the external file names and where to download them. Where could be a common location like `./data/external_test_data`.
- Provides a fixture that can provide the downloaded location of the external file to a test that uses that file.

Due to the inflexibility of the naming and syntax of .gitignore, I'm not sure that there is a solution that satisfies all three bullets. The closest that I can come up with would be to move to a global location for external test files to be downloaded and modify the `external_test_data_config.py` to satisfy the second and third bullet above.

maxinelasp:
If we add a prefix on all the downloaded files, I think that would be the most minimal way to accomplish what you're suggesting. I agree that your suggestion is probably the ideal, but would be somewhat invasive on the test side.

maxinelasp:
Going to close this and take another pass at it.

PR #2018, opened 2025-08-04
Supporting maps in different coordinate systems

pleasant-menlo:
# Change Summary

## Overview
Updated the map processors to generate maps based off the descriptor the processor is called with, and added support for passing "hae", "hre", "hnu", "gcs" for the coordinate system of a map. This will allow the mapping tool to create L2 maps with lots of different configurations including these new coordinate systems just by changing the descriptor passed to the processor.

## Updated Files
- imap_processing/cli.py
   - Updated do_processing to pass in the descriptor to lo_l2 
- imap_processing/ena_maps/utils/naming.py
   - Added handling for mapping hre, hnu, and gcs frames 
 - imap_processing/hi/hi_l2.py
     - Now uses the passed in descriptor to generate a skymap with the desired shape (Pixelation, reference frame), instead of defaulting to 4 degree, SF frame, ECLIPJ2000
    - Now uses the passed in descriptor to generate a skymap with the desired shape (Pixelation, reference frame), instead of defaulting to 6 degree ECLIPJ2000.

## Testing
 - Updated tests for each processor
 - Added a new test of the Lo processor in test_cli.py


imap_processing/ena_maps/utils/naming.py
```diff
@@ -335,6 +335,12 @@ def get_map_coord_frame(frame_str: _coord_frame_str_types) -> SpiceFrame:
         """
         if frame_str == "hae":
             return SpiceFrame.ECLIPJ2000
```
greglucas: I realize this was already here, but should this be `IMAP_HAE`? Are they equivalent?

imap_processing/lo/l2/lo_l2.py
```diff
@@ -38,9 +41,7 @@ def lo_l2(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
         psets = sci_dependencies["imap_lo_l1c_pset"]
 
         # Create the rectangular sky map from the pointing set.
-        lo_rect_map = project_pset_to_rect_map(
-            psets, spacing_deg=6, spice_frame=geometry.SpiceFrame.ECLIPJ2000
-        )
+        lo_rect_map = project_pset_to_sky_map(psets, descriptor)
```
greglucas: You updated the hi variable name to remove the rect, should we also do that here so we are consistent across all of the instruments? (and elsewhere in this file)
```suggestion
        output_map = project_pset_to_sky_map(psets, descriptor)
```

greglucas (commented):
Overall I think this looks great! Thank you for taking this effort on. I'll let @subagonsouth and @sdhoyt comment on their specific instruments though since I'm not as familiar with it.

imap_processing/ena_maps/utils/naming.py
tmplummer: Yes, they are equivalent. `IMAP_HAE` is an alias for `ECLIPJ2000` but was not in the science frame kernel when this code was implemented. I agree that now we should switch to the `IMAP_HAE` frame here.

imap_processing/hi/hi_l2.py
```diff
@@ -43,23 +47,34 @@ def hi_l2(
...
     )
 
     # Get the map dataset with variables/coordinates in the correct shape
     # TODO get the correct descriptor and frame
-    l2_ds = rect_map.build_cdf_dataset("hi", "l2", "sf", descriptor, sensor=sensor)
+
+    if not isinstance(rect_map, RectangularSkyMap):
+        raise NotImplementedError("HEALPix map output not supported for Hi")
+    if not isinstance(map_descriptor.sensor, str):
+        raise NotImplementedError("Sensor must be 45 or 90")
```
tmplummer: Please change this to a `ValueError` and specify that it was not found to be a string.
```suggestion
        raise ValueError("Invalid map_descriptor. Sensor attribute must be of type str and be either '45' or '90'")
```

imap_processing/hi/hi_l2.py
```diff
@@ -43,23 +47,34 @@ def hi_l2(
         Level 2 IMAP-Hi dataset ready to be written to a CDF file.
     """
     # TODO: parse descriptor to determine map configuration
-    sensor = "45" if "45" in descriptor else "90"
     direction: Literal["full"] = "full"
     cg_corrected = False
-    map_spacing = 4
+    map_descriptor = MapDescriptor.from_string(descriptor)
 
     rect_map = generate_hi_map(
```
tmplummer: Please rename this variable to reflect that it might not be a `RectangularSkyMap`.
```suggestion
    sky_map = generate_hi_map(
```

tmplummer (commented):
This is great! Thanks for taking care of this. It was on my long list but you beat me to it.

Just a couple of nit picks.

pleasant-menlo:
We went ahead and made all the edits you all suggested. Give it another look, and let us know if there is anything else.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -40,16 +40,19 @@ def lo_l2(
         logical_source = "imap_lo_l2_l090-ena-h-sf-nsp-ram-hae-6deg-3mo"
         psets = sci_dependencies["imap_lo_l1c_pset"]
 
-        # Create the rectangular sky map from the pointing set.
-        lo_rect_map = project_pset_to_sky_map(psets, descriptor)
+        # Create an AbstractSkyMap (Rectangular or HEALPIX) from the pointing set
+        lo_sky_map = project_pset_to_sky_map(psets, descriptor)
+        if not isinstance(lo_sky_map, RectangularSkyMap):
```
pleasant-menlo, line 45: We also added a similar assert to Lo that the SkyMap is rectangular, because we believe the metadata and math would require more work for Lo to properly support Healpix

imap_processing/tests/hi/test_hi_l2.py
```diff
@@ -41,7 +41,7 @@ def empty_rectangular_map_dataset() -> xr.Dataset:
 
 
 @pytest.mark.external_test_data
-def test_hi_l2(hi_l1_test_data_path):
+def test_hi_l2(hi_l1_test_data_path, imap_ena_sim_metakernel):
```
tmplummer, line 45: The addition of this fixture requires the addition of `@pytest.mark.externel_kernel` as well.

tmplummer (approved):
Thanks for the changes. Looks good other than the one comment.

imap_processing/tests/hi/test_hi_l2.py
```diff
@@ -38,7 +41,7 @@ def empty_rectangular_map_dataset() -> xr.Dataset:
 
 
 @pytest.mark.external_test_data
```
greglucas, line 43:
```suggestion
@pytest.mark.external_test_data
@pytest.mark.external_kernel
```

imap_processing/tests/hi/test_hi_l2.py
greglucas: Just added this in.

greglucas:
Thank you very much for taking this on, it will be really helpful for the generalizations!

Issue #1945, opened 2025-07-17
Calculate CoDICE acquisition times instead of using a lookup table


bourque:
Currently the CoDICE pipeline uses a simple lookup table to define it's acquisition times, however, these may change in-flight and it may be better to implement the calculation so that only one or two parameters need to be tweaked instead of an entire table.

tech3371:
Was this done because of the closed PR?

bourque:
No, this issue was just referenced in that PR, but it is still not done. Currently we use a lookup table stored in `constants.py` for this, but ideally we should be calculating the acquisition times manually as described in Appendix B of the algorithm document.

tech3371:
This is completed now because we are reading from SCI-LUT json instead of constants.py

PR #2019, opened 2025-08-05
CoDICE: Updated acq time lookup table values

bourque:
This PR simply updates the CoDICE acquisition time lookup table values, as requested by the instrument team.


greglucas (approved):
A few questions:
* Would it be easier to have this as a calibration file you could read in rather than updating the actual Python lists every time?
* Should these be full floating point values because it looks like they have repeating decimals at the end. So I'm curious whether it should read more like a fraction that you then divide somewhere and then set as a variable here.

bourque:
@greglucas 
>     * Would it be easier to have this as a calibration file you could read in rather than updating the actual Python lists every time?
I think I will go down this road if we find that these values keep changing, but we don't anticipate that happening very often (or at all) in-flight.
> 
>     * Should these be full floating point values because it looks like they have repeating decimals at the end. So I'm curious whether it should read more like a fraction that you then divide somewhere and then set as a variable here.

These values are indeed calculated from a (rather complex) equation, which probably explains the repeating numbers. I have a 'nice to have' ticket about this (#1945), which will calculate the values on-the-fly instead of using a lookup table, but I am considering that lower priority.

PR #2021, opened 2025-08-05
I-ALiRT - MAG frame transformation to ECLIPJ2000 using Spacecraft Packet

laspsandoval:
# Change Summary

## Overview
The MAG vectors are calculated based on 4 packets, each of which contains its own spin phase, inertial right ascension, and inertial decline. This PR addresses that issue and performs a frame transformation from the MAG instrument frame to the ECLIPJ2000.

I also updated the docstring to include constants for KSWC (our future ground station partner).

## Updated Files
- parse_mag.py
   - Transform vector to ECLIPJ2000.
- constants.py
   - Add KSWC to the docstring

## Testing
test_parse_mag.py

### DIFF
--- a/imap_processing/ialirt/constants.py
+++ b/imap_processing/ialirt/constants.py
@@ -48,7 +48,15 @@ class StationProperties(NamedTuple):
     min_elevation_deg: float  # minimum elevation angle in degrees
 
 
-# Verified by Kiel Observatory staff.
+# Verified by Kiel and KSWC Observatory staff.
+# Notes: the KSWC station is not yet operational,
+# but will have the following properties:
+# "KSWC": StationProperties(
+#     longitude=126.2958,  # degrees East
+#     latitude=33.4273,  # degrees North
+#     altitude=0.1,  # approx 100 meters
+#     min_elevation_deg=5,  # 5 degrees is the requirement
+# ),
 STATIONS = {
     "Kiel": StationProperties(
         longitude=10.1808,  # degrees East
--- a/imap_processing/ialirt/l0/parse_mag.py
+++ b/imap_processing/ialirt/l0/parse_mag.py
@@ -27,11 +27,6 @@ from imap_processing.mag.l1d.mag_l1d_data import MagL1d
 from imap_processing.mag.l2.mag_l2_data import MagL2L1dBase
 from imap_processing.spice.time import met_to_ttj2000ns, met_to_utc
 
-# Range values (mago is 0 to 1, magi is 2 to 3)
-# Range values (0 to 3) represent MAG gain setting
-MAGO_RANGE = np.array([0, 1])
-MAGI_RANGE = np.array([2, 3])
-
 logger = logging.getLogger(__name__)
 
 
@@ -298,6 +293,7 @@ def calculate_l1b(
 
 def calibrate_and_offset_vectors(
     vectors: np.ndarray,
+    range_vals: np.ndarray,
     calibration: np.ndarray,
     offsets: np.ndarray,
     is_magi: bool = False,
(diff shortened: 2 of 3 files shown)


imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -309,6 +305,8 @@ def calibrate_and_offset_vectors(
     ----------
     vectors : np.ndarray
         Raw magnetic vectors, shape (n, 3).
+    range_vals : np.ndarray
```
maxinelasp, line 312: Can you add that you expect mago to be [0,1] and magi to be [2,3]?

maxinelasp (approved):
~~I'm not sure I understand why this change is necessary - but assuming it's for some future work, this looks good~~

Sorry, github actually just didn't show me all the changes! Let me properly review

imap_processing/ialirt/constants.py
```diff
@@ -48,7 +48,15 @@ class StationProperties(NamedTuple):
     min_elevation_deg: float  # minimum elevation angle in degrees
 
 
-# Verified by Kiel Observatory staff.
+# Verified by Kiel and KSWC Observatory staff.
+# Notes: the KSWC station is not yet operational,
+# but will have the following properties:
+# "KSWC": StationProperties(
```
maxinelasp, line 54: Any reason we can't add them now?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -338,6 +341,53 @@ def calibrate_and_offset_vectors(
...
+    Transform vector to ECLIPJ2000.
+
+    Parameters
+    ----------
+    sc_spin_phase_rad : numpy.ndarray
+        Spin phase for 4 packets 0 to 2π radians, shape (4).
+    sc_inertial_right : numpy.ndarray
+        Inertial right ascension for 4 packets 0 to 2π radians, shape (4).
+    sc_inertial_decline : numpy.ndarray
+        Inertial right ascension for 4 packets -π/2 to π/2 radians, shape (4).
```
maxinelasp: Is this comment correct? 
```suggestion
        Inertial decline for 4 packets -π/2 to π/2 radians, shape (4).
```

maxinelasp (approved):

imap_processing/ialirt/constants.py
laspsandoval: Yeah. This is used as a prediction. If we add it now then it will be predicted that they will provide coverage.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -379,6 +383,66 @@ def apply_gradiometry_correction(
...
+
+    Notes
+    -----
+    The MAG vectors are calculated based on 4 packets,
+    each of which contains its own spin phase,
+    inertial right ascension, and inertial decline.
+    """
+    # Interpolate spin phase, RA, and Dec at target_time
+    spin_phase_deg = np.degrees(
+        np.interp(target_time, attitude_time, sc_spin_phase_rad)
```
tmplummer: Any concern that `target_time` will be outside the interval `attitude_time[[0, -1]]`? If that happens, `np.interp()` will just return `sc_spin_phase_rad[0]` or `sc_spin_phase_rad[-1]` rather than raising an error.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -379,6 +383,66 @@ def apply_gradiometry_correction(
...
+    )
+    ra_deg = np.degrees(np.interp(target_time, attitude_time, sc_inertial_right))
+    dec_deg = np.degrees(np.interp(target_time, attitude_time, sc_inertial_decline))
```
tmplummer: Since RA and DEC have discontinuities, the interpolation between times when the RA or DEC go through the discontinuity will be incorrect. I think that you either need to unwrap `sc_intertial_right` and `sc_intertial_decline` or convert to Cartesian and then interpolate?

https://numpy.org/doc/stable/reference/generated/numpy.unwrap.html#numpy-unwrap

tmplummer (commented):
Just a couple of comments to check on.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: amazing catch. Thanks.

PR #2024, opened 2025-08-06
Replace earth pck spice kernel

torimarbois:
# Change Summary
Replace the large `earth_1962_240827_2124_combined.bpc` kernel with the smaller `earth_latest_high_prec.bpc` kernel.

## Overview
I've replaced the old kernel name with the new one in every place it was mentioned, and updated the date ranges within a few tests due to the shorter span of time that the new kernel covers.

## Updated Files

   - Replace kernel
- imap_processing/tests/ialirt/unit/test_process_ephemeris.py
   - Replace kernel
   - Update time range 
- imap_processing/tests/ultra/unit/test_ultra_l1c_culling.py
   - Replace kernel

### DIFF
--- a/imap_processing/tests/conftest.py
+++ b/imap_processing/tests/conftest.py
@@ -52,7 +52,6 @@ def _download_external_kernels(spice_test_data_path):
     kernel_urls = [
         "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/de440s.bsp",
         "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/pck00011.tpc",
-        "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/",
         "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_latest_high_prec.bpc",
     ]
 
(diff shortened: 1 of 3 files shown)


imap_processing/tests/conftest.py
```diff
@@ -52,8 +52,8 @@ def _download_external_kernels(spice_test_data_path):
     kernel_urls = [
         "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/de440s.bsp",
         "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/pck00011.tpc",
-        "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/"
-        "earth_1962_240827_2124_combined.bpc",
+        "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/",
```
tmplummer: It looks like this line was supposed to be removed?
```suggestion
```

tmplummer (commented):
Looks good to me other than the one change suggestion.

imap_processing/tests/conftest.py
torimarbois: No, I just added a comma at the end. It was there before, so I didn't remove it - is it necessary to keep? It does seem weird to me that it doesn't point to a specific kernel.

torimarbois: @subagonsouth just re-upping this^

tmplummer: So without the comma, it gets combined with the string on the next line so that there is one entry in the list that is `https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_1962_240827_2124_combined.bpc`. It is not necessary since that is the kernel being removed.

torimarbois: I see! My mistake. I'll fix it now.

laspsandoval (approved):
Looks good!

tmplummer (approved):
LGTM. Thanks!

PR #2025, opened 2025-08-06
Upgrade cdflib to 1.3.6

vmartinez-cu:
Upgrade `cdflib` to the latest version, 1.3.6, which fixes an issue where uncertainty variables in the "DELTA_VAR" attribute fields of other variables were being converted to coordinates when loading CDFs.


bourque:
Should `pyproject.toml` be updated as well?

greglucas:
> Should pyproject.toml be updated as well?

Yes, this is the only thing that should be updated locally. Then let pre-commit update `poetry.lock`, that step shouldn't be done manually.

greglucas:
@vmartinez-cu are you able to address the comments?

vmartinez-cu:
The pyproject.toml already has `cdflib = "^1.3.1"` which covers any versions above that. The lock file was updated by running `poetry update cdflib` in the terminal. Sounds like instead of doing this, I should update the toml to `cdflib = "^1.3.6"`?

greglucas (approved):

PR #2026, opened 2025-08-06
GLOWS: Adding SPICE parameters to L1B

### LINKED ISSUE #897: Integrate SPICE into GLOWS L1B
# Summary
Glows L1B requires some inputs from SPICE, including spacecraft location information. Once SPICE tools are completed, integrate them into GLOWS L1B. 

Depends on SPICE tooling, so blocked for now.

tech3371:
# Change Summary
closes #897 

## Overview

Added and updated SPICE related test in GLOWS L1B code.

## Updated Files

- imap_processing/glows/l1b/glows_l1b.py
   - add new variable and it's dimension
- imap_processing/glows/l1b/glows_l1b_data.py
   - main code that updates SPICE related stuff

## Testing

- imap_processing/tests/glows/conftest.py
- imap_processing/tests/glows/test_glows_l1b.py
- imap_processing/tests/glows/test_glows_l1b_data.py


imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -592,6 +592,78 @@ def __post_init__(
         # )
         self.flags = np.ones((FLAG_LENGTH,), dtype=np.uint8)
 
+    def update_spice_parameters(self) -> None:
+        """Update SPICE parameters based on the current state."""
+        data_start_met = self.imap_start_time
+        # TODO: should we be using glows_start_time instead?
```
maxinelasp: I'd say we can remove this TODO and say no, we should use imap_start_time.

imap_processing/tests/glows/test_glows_l1b.py
```diff
@@ -363,7 +375,10 @@ def test_glows_l1b(de_dataset, hist_dataset):
         assert key in de_output
 
 
-def test_generate_histogram_dataset(hist_dataset):
+@patch.object(HistogramL1B, "update_spice_parameters", autospec=True)
```
maxinelasp, line 403: I like this method of mocking, it's not one I've really used before but it's very clean

maxinelasp (approved):
Looks good, thank you so much!!

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -495,16 +503,18 @@ class HistogramL1B:
     # )  # Could be datetime TODO: Can't put a string in data
     imap_spin_angle_bin_cntr: np.ndarray = field(init=False)  # Same size as bins
     histogram_flag_array: np.ndarray = field(init=False)
-    spin_period_ground_average: np.double = field(init=False)  # retrieved from SPICE?
-    spin_period_ground_std_dev: np.double = field(init=False)  # retrieved from SPICE?
+    spin_period_ground_average: np.double = field(init=False)  # retrieved from SPICE
+    spin_period_ground_std_dev: np.double = field(init=False)  # retrieved from SPICE
```
tmplummer: Just a note that these are not retrieved from SPICE. They come from the universal spin table.

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -592,6 +592,78 @@ def __post_init__(
...
+        angle_offset = 360 - get_spin_angle(
+            get_instrument_spin_phase(
+                self.imap_start_time, instrument=geometry.SpiceFrame.IMAP_GLOWS
+            ),
+            degrees=True,
+        )
+        self.position_angle_offset_average = np.average(angle_offset)
+        self.position_angle_offset_std_dev = np.double(
+            0.0
+        )  # Set to zero per algorithm document
```
tmplummer: I'm not following this. Is `self.imap_start_time` a np.double or an np.ndarray? If it is just a double, then this is querying just a single time and the average has no effect.

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -592,6 +592,78 @@ def __post_init__(
...
+        # Calculate spin axis orientation
+        spin_axis_orientation = geometry.cartesian_to_latitudinal(
+            geometry.frame_transform(
+                time_range,
+                np.array([0, 0, 1]),
+                SpiceFrame.IMAP_DPS,
+                SpiceFrame.ECLIPJ2000,
+            )
+        )
+        self.spin_axis_orientation_average = np.average(spin_axis_orientation, axis=0)
```
tmplummer: Careful here. There are issues averaging the vectors when they are expressed in Lat,Lon due to the discontinuities in Lat and Lon. You actually need to do the average on each component of the Cartesian vectors and then convert to Lat/Lon coordinates.

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -592,6 +592,78 @@ def __post_init__(
...
+        spin_axis_all_times = geometry.cartesian_to_latitudinal(
+            geometry.frame_transform(
+                time_range,
+                np.tile([0, 0, 1], (len(time_range), 1)),
```
tmplummer: I think that `frame_transform` should allow you to just pass in a single position vector, `np.array([0, 0, 1])` and it will provide the postition vectors with shape: `(len(time_range), 3)` .

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -592,6 +592,78 @@ def __post_init__(
...
+                SpiceFrame.IMAP_DPS,
+                SpiceFrame.ECLIPJ2000,
+            )
+        )
+        self.spin_axis_orientation_std_dev = np.std(spin_axis_all_times, axis=0)
```
tmplummer: For this one, we need to ask what they want from the spin_axis_orientation_std_dev. Doing the standard deviation on the Lat/Lon coordinates has the same issue as the average above but it is also not clear what is desired from this standard deviation.

imap_processing/tests/glows/test_glows_l1b.py
```diff
@@ -376,3 +391,61 @@ def test_generate_de_dataset(de_dataset):
...
+        assert isinstance(hist_data.spin_period_ground_average, np.float64)
+        assert isinstance(hist_data.spin_period_ground_std_dev, np.float64)
+        assert isinstance(hist_data.position_angle_offset_average, np.float64)
+        assert isinstance(hist_data.position_angle_offset_std_dev, np.float64)
+        assert hist_data.spin_axis_orientation_std_dev.shape == (3,)
+        assert hist_data.spin_axis_orientation_average.shape == (3,)
+        assert hist_data.spacecraft_location_average.shape == (3,)
+        assert hist_data.spacecraft_location_std_dev.shape == (3,)
+        assert hist_data.spacecraft_velocity_average.shape == (3,)
+        assert hist_data.spacecraft_velocity_std_dev.shape == (3,)
```
tmplummer: Interesting, I was assuming these would be shape == (2,) since they are in Latitudinal coordinates and should always have radius of 1.

tmplummer (commented):
I have some comments on the averaging and std_dev of vectors.

imap_processing/glows/l1b/glows_l1b_data.py
tech3371: It's a double. Then don't need to average? what about std too?

imap_processing/glows/l1b/glows_l1b_data.py
tech3371: It's a 3d (radius, long, lat). So average/std over lat and log separately and then combine into 2d array?

imap_processing/glows/l1b/glows_l1b_data.py
tech3371: cool. like this?
```
geometry.frame_transform(
                time_range,
                np.array([0, 0, 1]),
                SpiceFrame.IMAP_DPS,
                SpiceFrame.ECLIPJ2000,
            )
        )
```

imap_processing/glows/l1b/glows_l1b_data.py
tech3371: I don't know much either. This is what algorithm document says
Screenshot 2025-08-07 at 2 46 38 PM

imap_processing/tests/glows/test_glows_l1b.py
tech3371: Maxine did tell me that it should be 2 but yeah, it did return 3. I saw in the function doc that it does return (radius, long, lat)

imap_processing/glows/l1b/glows_l1b_data.py
tech3371: nevermind on std. They want it to set to zero.

imap_processing/glows/l1b/glows_l1b_data.py
greglucas: If you have an array of angles of `[1, 359]`, is the average 180, or is it 0? Usually you want the smaller angle that wrapped around.

greglucas: @subagonsouth should we add a helper function to help with these averages? Or is this only needed in rare situations.

tmplummer: No, the problem is being in lat/lon coordinates. For example, given the following latitudinal vectors in degrees:
`[[1, 179, 0], [1, -179, 0]]` the average will be `[1, 0, 0]` which is 180 degrees away from the correct average of `[1, 180, 0]`. 

Asking Claude, came up with the following solution that I like better than what I was thinking. Define a circular mean and standard deviation function that is used on the Longitude coordinate instead of the standard (linear) mean and standard deviation.
```
def circular_mean_and_std(angles):
    """
    Calculate circular mean and standard deviation for angles.
    This is important for longitude since angles wrap around.
    
    Parameters
    ----------
    angles : array_like
        Array of angles in radians
        
    Returns
    -------
    circular_mean : float
        Circular mean in radians
    circular_std : float
        Circular standard deviation in radians
    """
    # Convert to unit vectors on the unit circle
    cos_angles = np.cos(angles)
    sin_angles = np.sin(angles)
    
    # Calculate mean of unit vectors
    mean_cos = np.mean(cos_angles)
    mean_sin = np.mean(sin_angles)
    
    # Calculate circular mean
    circular_mean = np.arctan2(mean_sin, mean_cos)
    
    # Calculate mean resultant length
    R = np.sqrt(mean_cos**2 + mean_sin**2)
    
    # Calculate circular standard deviation
    # Formula: sqrt(-2 * ln(R))
    if R > 0:
        circular_std = np.sqrt(-2 * np.log(R))
    else:
        circular_std = np.inf  # Maximum uncertainty
    
    return circular_mean, circular_std
```
Then, with the `spin_axis_all_times` as computed below, you can do:
```
# Calculate circular statistics for longitude (wraps around)
lon_mean, lon_std = circular_mean_and_std(spin_axis_all_times[..., 1])
lat_mean = np.mean(spin_axis_all_times[...,2]
lat_std = np.std(spin_axis_all_times[...,2])
```

@greglucas, we could certainly write a geometry function that calculates the latitudinal mean and stddev given a set of latitudinal vectors. GLOWS is the only instrument that I am aware of getting this geometry but ultimately that function belongs in the geometry module anyway.

imap_processing/glows/l1b/glows_l1b_data.py
tmplummer: Ok, so based on this document, I would expect 5 and 6 to be ndarrays with shape (2,).

imap_processing/tests/glows/test_glows_l1b.py
tmplummer: Yep, so just store the longitude and latitude components.

imap_processing/glows/l1b/glows_l1b_data.py
tech3371: Thank you for the code! That helps a lot since I don't know much there.

tech3371:
@subagonsouth I think I redid the calculation of spin axis orientation part and fixed test to look for (2,) now. Can you review again when you get a chance?

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -811,6 +809,75 @@ def __post_init__(
...
+                time_range,
+                np.array([0, 0, 1]),
+                SpiceFrame.IMAP_DPS,
+                SpiceFrame.ECLIPJ2000,
+            )
+        )
+        # Calculate circular statistics for longitude (wraps around)
+        lon_mean, lon_std = circular_mean_and_std(spin_axis_all_times[..., 1])
+        lat_mean = np.mean(spin_axis_all_times[..., 2])
+        lat_std = np.std(spin_axis_all_times[..., 2])
```
tmplummer: After going through an exercise of comparing the linear mean and stddev to circular mean and stddev, for a set of angles that don't wrap around and seeing that they are close but not identical, I think that we should use the same calculation for the latitude statistics. It is also worth bringing this up to GLOWS to make sure that they are happy with this implementation.
```suggestion
        lat_mean = circmean(spin_axis_all_times[..., 2], low=-np.pi, high=np.pi)
        lat_std = circstd(spin_axis_all_times[..., 2], low=-np.pi, high=np.pi)
```

imap_processing/tests/glows/conftest.py
```diff
@@ -183,3 +183,16 @@ def mock_ancillary_parameters():
...
     return AncillaryParameters(mock_table)
+
+
+def mock_update_spice_parameters(self, *args, **kwargs):
+    self.spin_period_ground_average = np.float64(0.0)
+    self.spin_period_ground_std_dev = np.float64(0.0)
+    self.position_angle_offset_average = np.float64(0.0)
+    self.position_angle_offset_std_dev = np.float64(0.0)
+    self.spin_axis_orientation_average = np.zeros(3, dtype=np.float64)
+    self.spin_axis_orientation_std_dev = np.zeros(3, dtype=np.float64)
```
tmplummer: Should these have just 2 elements now?
```suggestion
    self.spin_axis_orientation_average = np.zeros(2, dtype=np.float64)
    self.spin_axis_orientation_std_dev = np.zeros(2, dtype=np.float64)
```

imap_processing/spice/geometry.py
```diff
@@ -568,3 +568,46 @@ def solar_longitude(
     lat_coords = cartesian_to_latitudinal(imap_pos, degrees=degrees)[..., 1]
 
     return float(lat_coords) if lat_coords.size == 1 else lat_coords
+
+
+def circular_mean_and_std(angles: np.ndarray) -> tuple[float, float]:
```
tmplummer: Oh, hey... Sorry for not checking this earlier. scipy has circular statistics functions.... scipy.stats.circmean and scipy.stats.circstd. Since we have that as a dependency, lets use those functions.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.circmean.html
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.circstd.html

imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -811,6 +809,75 @@ def __post_init__(
...
+        spin_axis_all_times = geometry.cartesian_to_latitudinal(
+            geometry.frame_transform(
+                time_range,
+                np.array([0, 0, 1]),
+                SpiceFrame.IMAP_DPS,
+                SpiceFrame.ECLIPJ2000,
+            )
+        )
+        # Calculate circular statistics for longitude (wraps around)
+        lon_mean, lon_std = circular_mean_and_std(spin_axis_all_times[..., 1])
```
tmplummer: After finding that scipy has these functions, suggested change is:
```suggestion
    from scipy.stats import circmean, circstd
        lon_mean = circmean(spin_axis_all_times[..., 1], low=-np.pi, high=np.pi)
        lon_std = circstd(spin_axis_all_times[..., 1], low=-np.pi, high=np.pi)
```

tmplummer (commented):
Sorry that I didn't realize that scipy had these functions earlier. The good news is that we don't need to write tests for the `circular_mean_and_std` function.

tmplummer (approved):
LGTM

PR #2028, opened 2025-08-06
Minimal updates to gitignore

maxinelasp:
# Change Summary
Exclude downloaded SWE and MAG tests from git. 

I also investigated specifying downloaded test files by adding "downloaded_" as a prefix on the file names, but unfortunately, even with Claude there were way too many places to update for it to be a reasonable change.

I still strongly prefer specifically listing excluded files such as in #2012 and feel that bulk-excluding folders with mixed downloaded and tracked files is confusing. I also think it would be a good idea to indicate that a file is downloaded in the file name or file location.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,6 +178,18 @@ cython_debug/
 # Data that is downloaded
 data/
 test_data/
+/imap_processing/tests/mag/validation/L1c/T013/mag-l1b-l1c-t013-magi-burst-in.csv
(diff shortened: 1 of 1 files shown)


tmplummer (approved):
Should we capture a ticket to improve our external data download tools?

greglucas (approved):
I agree, updating the documentation would probably be good somewhere, but should be low priority I think.

tech3371 (approved):

PR #2029, opened 2025-08-06
Temporary fix for the external_kernel pytest marker

lacoak21:
# Change Summary

## Overview
The 'external_kernel' pytest marker assigns the fixture _download_external_kernels. This is failing when another fixture that expects local kernels is also assigned with the test. Im not sure what the benefits of using the marker vs pytest.uses_fixture are. That is why I am adding some temporary TODOs instead of removing them. I think @subagonsouth will know more. 

## Updated Files
All fixes are the same. 
- Add this fixture @pytest.mark.usefixtures("_download_external_kernels")

### DIFF
--- a/imap_processing/tests/conftest.py
+++ b/imap_processing/tests/conftest.py
@@ -515,3 +515,4 @@ if __name__ == "__main__":
     # This is to enable downloading files easier by letting us
     # run this file directly
     _download_external_data()
+    _download_external_kernels(spice_test_data_path)
--- a/imap_processing/tests/conftest.py
+++ b/imap_processing/tests/conftest.py
@@ -43,6 +43,10 @@ def clear_spin_and_repoint_paths(monkeypatch):
 
 
 @pytest.fixture(scope="session")
+def _download_kernels(spice_test_data_path):
+    _download_external_kernels(spice_test_data_path)
+
+
 def _download_external_kernels(spice_test_data_path):
     """This fixture downloads externally-located kernels into the tests/spice/test_data
     directory if they do not already exist there. The fixture is not intended to be
@@ -136,7 +140,7 @@ def pytest_collection_modifyitems(items):
     +---------------------+----------------------------+
     | pytest mark         | fixture added              |
     +=====================+============================+
-    | external_kernel     | _download_external_kernels |
+    | external_kernel     | _download_kernels          |
     | external_test_data  | _download_test_data        |
     +---------------------+----------------------------+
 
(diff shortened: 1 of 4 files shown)


imap_processing/tests/conftest.py
```diff
@@ -462,7 +462,7 @@ def wrapped_repoint_data_filepath(
 
 
 @pytest.fixture
-def imap_ena_sim_metakernel(furnish_kernels):
+def imap_ena_sim_metakernel(furnish_kernels, _download_external_kernels):
```
tmplummer: I think that this is probably the only fix needed?

Actually, now I see that @greglucas and @bourque set up the external data to download in github before the tests run. It would probably be good to add the same functionality to the external kernels. See:

- https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/tests/conftest.py#L514
- https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/.github/workflows/test.yml#L61

tmplummer (commented):
The `@pytest.mark.external_kernel` was implemented that way so that we can run pytest on the subset of tests that don't require external kernels or external data by doing: `pytest -m "not external_kernel and not external_test_data"`

I recommend:
- Leaving the `@pytest.mark.external_kernel` decorators in place
- Adding the dependency of `_download_external_kernels` to `imap_ena_sim_metakernel`, like you have done
- Pulling the code that downloads the kernels out into a function and calling that function in the `if __name__ == "__main__":` block at the end of the conftest.py file

I think that should take care of the issues. 🤞

imap_processing/tests/conftest.py
greglucas: The reason we did the external test data download separately was because in the pytest multi-processing environment there was some conflicts on whether the downloads were happening before tests requiring them would run. So it was easier to just stick that in front of everything to download apriori before running the tests in parallel. I agree we should add the kernel downloads there too.

lacoak21: OK sounds good. Do you think I still need to add the fixture to the  imap_ena_sim_metakernel fixture? Might be redundant now.

greglucas: No, that is still required because of running locally. We don't want to have to run the __main__ version if we don't need to.

lacoak21: Oh right 😆

tmplummer (approved):
Thanks! Looks good to me.

PR #2031, opened 2025-08-07
I-ALiRT - ingest plots

laspsandoval:
# Change Summary

## Overview
Our Kiel partner requested that we plot real-time ingest reports. To do that we are using logs created by the POC on an hourly basis. These logs contain to parts:

1. tcp connection information for each station
2. packet creation times (which do not have information for a single station)

This code creates a data structure that will be used to generate a json file every hour. That file will be stored in s3 and the WebTeam will use it to create plots on the IMAP website.

## New Files
- calculate_ingest.py
   - Creates the data that will be used for real-time ingest plots.

## Updated Files
- ialirt_test_data.csv
   - Relocated this file
- test_process_swapi.py
   - Result of relocated file
- constants.py
   - Updated list of DSN stations

## Testing
test_calculate_ingest.py
flight_iois_1.log.2025-212T16_55_27.531613
   - New test data


laspsandoval:
pre-commit.ci autofix

laspsandoval:
pre-commit.ci autofix

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,189 @@
...
+
+import logging
+from datetime import datetime, timedelta, timezone
+from typing import Any
+
+from imap_processing.ialirt.constants import STATIONS
+
+logger = logging.getLogger(__name__)
+
+ALL_STATIONS = [*STATIONS.keys()]
```
greglucas: I don't think you need this for how you are using it below. You can iterate directly through a dictionary, and you can call `list(STATIONS)` to create your list of keys.

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # TCP connection data for each station
+    for station in ALL_STATIONS:
+        tcp_ranges = find_tcp_connections(
+            start_of_time, end_of_time, all_lines, station
+        )
+        formatted["tcp"][station] = [
+            {"start": start.isoformat(), "end": end.isoformat()}
+            for start, end in tcp_ranges
+        ]
```
greglucas: You are iterating through the file multiple times with this approach and also requiring exactly some specific ground station names. I'd suggest inverting the way this is done and doing something like:

```python
for line in lines:
    if "antenna partner is" in line:
        add_connection_up_down(formatted["tcp"], line)
    elif "Renamed iois_1_packets" in line:
        add_packet_times(formatted["packet_ingest"], line)
```

So that your function would work on a single line and add the information from that line to the object that it needs to add it to. Additionally, I'd extract out the ground station name from the line rather than only keeping the ground stations you know about. Then if another ground station gets added or the text is off it won't completely get missed, but you'd still add it under that alternative name.

imap_processing/ialirt/generate_coverage.py
```diff
@@ -11,8 +11,21 @@
 # Logger setup
 logger = logging.getLogger(__name__)
 
-# TODO: get a list of all potential DSN stations.
-ALL_STATIONS = [*STATIONS.keys(), "DSS-55", "DSS-56", "DSS-74", "DSS-75"]
+ALL_STATIONS = [
+    *STATIONS.keys(),
+    "DSS-24",
```
greglucas, line 16: We don't actually connect directly to the DSN groundstations, so are we ever going to get these? I thought these would flow through the MOC so it would be a MOC-like ground station as the intermediary?

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,189 @@
...
+      ],
+      "time_range": [
+        "2025-07-30T23:00:00",
+        "2025-07-31T02:00:00"
+      ],
+      "packet_ingest": [
+        "2025-07-31T00:00:00",
+        "2025-07-31T02:01:00"
+      ],
+      "tcp": {
```
greglucas: Rather than "tcp" maybe call this "connection_times" or something like that which is more intuitive what is going on at first glance. tcp is an implementation detail I feel like.

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,189 @@
...
+def format_ingest_data(last_filename: str, all_lines: list) -> dict:
+    """
+    Format TCP connection and packet ingest data from multiple log files.
+
+    Parameters
+    ----------
+    last_filename : str
+        Log file that is last chronologically.
+    all_lines : list[str]
+        Combined lines from all log files (assumed already sorted by time).
```
greglucas, line 137: Are there ever multiple logs being written to at once where these might be interleaved and so they are in-order per file but not across multiple files?

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,189 @@
...
+      ],
+      "tcp": {
+        "Kiel": [
+          {
+            "start": "2025-07-30T23:00:00",
+            "end": "2025-07-31T00:15:00"
+          },
+          {
+            "start": "2025-07-31T02:00:00",
+            "end": "2025-07-31T02:00:00"
```
greglucas, line 171: I'm not sure how big this will get, but it might be worth thinking about whether a 2-tuple is OK here rather than repeating start/end every time. So every entry is a `(start, end)` combination.

imap_processing/ialirt/generate_coverage.py
laspsandoval: Yep. We will get these through the contact schedules that the POC will deliver to us via the upload api.

imap_processing/ialirt/calculate_ingest.py
laspsandoval: True. Thanks.

imap_processing/ialirt/calculate_ingest.py
laspsandoval: Response from Melissa:

"No! There's one flight log written per container so since we only have one container there aren't multiple"

imap_processing/ialirt/calculate_ingest.py
laspsandoval: I see what you are saying. But each ground station should only have a single start and end per day (much less each hour). It would be unlikely for it to be very large.

imap_processing/ialirt/calculate_ingest.py
laspsandoval: Nice ideas.

laspsandoval:
pre-commit.ci autofix

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,211 @@
...
```
torimarbois, line 171: It's not clear to me what these time ranges represent or how they're documenting something different than the 'time_range' and 'packet_ingest' attributes.

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,211 @@
...
+    Parameters
+    ----------
+    start_file_creation : datetime
+        File creation time of last file minus 48 hrs.
+    end_file_creation : datetime
+        File creation time of last file.
+    lines : list
+        All lines of log files.
+    formatted : dict
+        Input dictionary.
```
torimarbois: Could these two parameters have more descriptive names? Especially since `formatted` is used many other places in this file.

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,211 @@
...
+    ----------
+    start_file_creation : datetime
+        File creation time of last file minus 48 hrs.
+    lines : list
+        All lines of log files.
+
+    Returns
+    -------
+    packet_times : list
+        List of datetime objects when packets were finalized.
```
torimarbois: What does a packet being _finalized_ mean?

imap_processing/ialirt/calculate_ingest.py
torimarbois: I agree with Greg.

imap_processing/tests/ialirt/unit/test_calculate_ingest.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def test_packets_created():
+    """Test the packets_created function."""
+    with open(
+        TEST_PATH / "flight_iois_1.log.2025-212T16_55_27.531613", encoding="utf-8"
+    ) as f:
+        lines = f.readlines()
+
+    test = packets_created(datetime(2025, 7, 31, 16, 33, 39, 0), lines)
```
torimarbois: Not to be nit-picky, but can this variable be renamed to be more descriptive? Even `actual_output` would work.

imap_processing/tests/ialirt/unit/test_calculate_ingest.py
```diff
@@ -0,0 +1,177 @@
...
+    data = format_ingest_data(filenames[-1], log_lines)
+
+    assert data["packet_ingest"][0] == "2025-07-31T08:00:00"
+    assert data["packet_ingest"][-1] == "2025-07-31T15:00:00"
+    assert data["tcp"]["Kiel"][0]["start"] == "2025-07-31T08:00:00"
+    assert data["tcp"]["Kiel"][0]["end"] == "2025-07-31T16:00:00"
+
+
+def test_format_ingest_data_edge_cases():
+    """Test the format_ingest_data_edge_cases function."""
```
torimarbois: There isn't actually a function called `format_ingest_data_edge_cases`, right? I
assuming you meant something that said "Test the edge cases of the format_ingest_data function."

imap_processing/ialirt/calculate_ingest.py
```diff
@@ -0,0 +1,211 @@
+"""Packet ingest times for each station."""
```
torimarbois: Can you clarify this docstring? It doesn't read well/make sense to me.

torimarbois (commented):
This is a lot of funky time-centered logic, which is hard to piece together! Nice job Laura.

imap_processing/ialirt/calculate_ingest.py
laspsandoval: I added this to the notes:
where time_range is the overall time range of the data,
    packet_ingest contains timestamps when packets were finalized,
    and tcp contains connection windows for each station.

imap_processing/ialirt/calculate_ingest.py
laspsandoval: created. Made the change.

imap_processing/ialirt/calculate_ingest.py
laspsandoval: It's just the when the packets were created (ingested) into s3. I also added tcp times there too since that's what we are also using.

laspsandoval:
pre-commit.ci autofix

imap_processing/tests/ialirt/data/l0/ialirt_test_data.csv
torimarbois, line 1: Can you remind me why you decided to move this file here? Asking because I have very similar additional files to add (swapi validation test sets) and I want to know where I should put them.

laspsandoval: We want all test data to be in the same place. If you want to rename the current test directory that's fine. It's on my TODO since not everything is l0. But we need the test data to be together and not in separate diretories.

torimarbois (approved):

PR #2036, opened 2025-08-11
Add spin-phase filtering to Hi L2

### LINKED ISSUE #1690: Hi L2 - Add ram, anti-ram optional input to HiPointingSet
### Algorithm Description:

In order to make ram/anti-ram specific maps, we can just trim out the undesired pixels from the PSET data upon instantiation of the PointingSet. This works the same for Hi90 and Hi45. From meeting with Paul:

> Ram should be spin-phase 0 to 0.5 (assuming IMAP spins the same direction as IBEX right-handed spin direction)

This means that filtering the PSET for Ram only, just keeps `spin_angle_bin`s with indices [0:N_SPIN_BINS//2]. Anti-Ram is the complement set of `spin_angle_bins`.

tmplummer:
# Change Summary

## Overview
For Hi PSETs, filtering by ram or anti-ram is easy. Just keep the appropriate indices for the `spin_angle_bin` coordinate. This implements that filtering.

## Updated Files
   - Add `spin_phase` parameter to the `HiPointingSet._init_()` method.
   - Filter the xarray.Dataset based on input `spin_phase`
- imap_processing/hi/hi_l2.py
   - Rename `direction` to `spin_phase` to be consistent with the `MapDescriptor` class.
   - Pass `spin_phase` from the map_descriptor into the HiPointingSet instantiation call.
   - Add test coverage for `spin_phase` options.
   - Update tests with new `generate_hi_map` signature

Closes: #1690

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -590,12 +590,30 @@ class HiPointingSet(PointingSet):
 
     Parameters
     ----------
-    dataset : xarray.Dataset
-        Hi L1C pointing set data loaded in an xarray.DataArray.
+    dataset : xarray.Dataset | str | Path
+        Hi L1C pointing set data loaded in a xarray.DataArray.
+    spin_phase : str
+        Include ENAs from "full", "ram" or "anti-ram" phases of the spin.
     """
 
-    def __init__(self, dataset: xr.Dataset):
+    def __init__(self, dataset: xr.Dataset | str | Path, spin_phase: str):
         super().__init__(dataset, spice_reference_frame=geometry.SpiceFrame.ECLIPJ2000)
+
+        # Filter out ENAs from non-selected portions of the spin.
+        if spin_phase not in ["full", "ram", "anti-ram"]:
+            raise ValueError(f"Unrecognized spin_phase value: {spin_phase}.")
+        # ram only includes spin-phase interval [0, 0.5)
+        # which is the first half of the spin_angle_bins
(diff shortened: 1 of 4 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -590,13 +590,31 @@ class HiPointingSet(PointingSet):
...
     """
 
-    def __init__(self, dataset: xr.Dataset):
+    def __init__(self, dataset: xr.Dataset | str | Path, spin_phase: str):
         super().__init__(dataset, spice_reference_frame=geometry.SpiceFrame.ECLIPJ2000)
 
+        # Filter out ENAs from non-selected portions of the spin.
+        if spin_phase not in ["full", "ram", "anti-ram"]:
+            raise ValueError(f"Unrecognized spin_phase value: {spin_phase}.")
+        # ram only includes spin-phase interval [0, 0.5)
```
greglucas, line 605: I thought the spin-phase was as we are passing through the ecliptic, so I thought the relevant intervals would be (0.75 - 1, 0 - 0.25) and (0.25 - 0.75). i.e. rotated by 90 degrees off the ecliptic.

tmplummer: Yeah, I am going to double check this with Paul. I agree with you, but in one of our tag-ups he claimed that ram should be spin-phase [0, 0.5).

greglucas (approved):
This is what the team is expecting after feedback from them. We can update the spin_phase offset axis in a follow-up PR.

PR #2037, opened 2025-08-11
Mag l1d fixes

maxinelasp:
# Change Summary
Some fixes around SPICE and from comments from the MAG team on L1D. 

## Overview

Checks off the todos:
- check if all vectors are primary before computing offsets
- add attributes for L1D
- output additional ancillary files
- Output additioanl frames
- work with the OUTPUT_MAGO configuration option (swaps the output vectors)

This also adds a correction to the spin offset calculation. This calculation should include invalid spins in the chunk counter, and just re-use the average if this means more than half the spins are missing. this also adds time validity and counters to the output ancillary file. 

This does add a new dependency on the SLCK kernel for MAG L1D. All the dependencies are already included in dependency_config.csv. 

Remaining work on L1D: Attribute work (waiting to hear from MAG team), validation (also waiting on MAG team)

## Testing

I used Claude to help with the tests.


maxinelasp:
Sorry for the messy commits, this branch has kind of a weird history. Should be fixed by the squash merge. Also, the failing pre-commit check should be fixed by PR #2038.

imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -328,6 +328,9 @@ def generate_dataset(
     global_attributes = attribute_manager.get_global_attributes(logical_file_id)
     global_attributes["is_mago"] = str(bool(single_file_l1a.is_mago))
     global_attributes["is_active"] = str(bool(single_file_l1a.is_active))
+    global_attributes["all_vectors_primary"] = int(
```
greglucas: Should this be boolean or str(bool)? The naming of the variable makes it seem like it will be true/false and the return of the function is a boolean.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -124,6 +124,21 @@ def mag_l1c(
     try:
         global_attributes["is_mago"] = normal_mode_dataset.attrs["is_mago"]
         global_attributes["is_active"] = normal_mode_dataset.attrs["is_active"]
+
+        # Check if all vectors are primary in both normal and burst datasets
+        is_mago = normal_mode_dataset.attrs.get("is_mago", "False") == "True"
+        normal_all_primary = normal_mode_dataset.attrs.get("all_vectors_primary", 0)
+
+        # Default for missing burst dataset: 1 if MAGO (expected primary), 0 if MAGI
+        burst_all_primary = 1 if is_mago else 0
```
greglucas: Swapping between the 0/1 and string True/False is kind of confusing here, so I would suggest choosing one or the other. Alternatively, if you can store a boolean in the attributes directly maybe that is even easier/more straightforward without having to do all of these casts.

imap_processing/mag/l1d/mag_l1d.py
```diff
@@ -45,16 +45,18 @@ def mag_l1d(
...
-        elif "norm-mago" in source:
-            input_mago_norm = dataset
-        elif "burst-magi" in source:
-            input_magi_burst = dataset
-        elif "burst-mago" in source:
-            input_mago_burst = dataset
-        else:
-            raise ValueError(f"Input data has invalid logical source {source}")
+        ins_mode = source.split("_")[-1]
+        match ins_mode:
```
greglucas: instrument? Be explicit if it is just a few extra characters in this one spot.
```suggestion
        instrument_mode = source.split("_")[-1]
        match instrument_mode:
```

imap_processing/mag/l1d/mag_l1d.py
```diff
@@ -122,12 +134,43 @@ def mag_l1d(
...
+        l1d_burst.rotate_frame(ValidFrames.GSE)
+        burst_gse_dataset = l1d_burst.generate_dataset(attributes, day_to_process)
+        l1d_burst.rotate_frame(ValidFrames.RTN)
+        burst_rtn_dataset = l1d_burst.generate_dataset(attributes, day_to_process)
         output_datasets.append(burst_srf_dataset)
         output_datasets.append(burst_dsrf_dataset)
+        output_datasets.append(burst_gse_dataset)
+        output_datasets.append(burst_rtn_dataset)
+
+    # Output ancillary files
```
greglucas, line 152: Do we need to put ancillary somewhere in the filenames below or how does it know to put these output datasets into the ancillary location. Or are we saying they are ScienceFiles still just "extra" somehow?

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -541,10 +668,20 @@ def calculate_gradiometry_offsets(
 
         diff = aligned_magi - mago_vectors
 
+        # Calculate magnitude of gradiometer offset for each vector
+        magnitude = np.linalg.norm(diff, axis=1)
+
+        # Set quality flags: 0 = good data (below threshold), 1 = bad data
+        quality_flags = (magnitude > quality_flag_threshold).astype(int)
```
greglucas, line 675: Are you setting quality based on any other measures, so you want to reserve any specific bits to mean different reasons why it is a bad quality?

greglucas (commented):
There is a lot of stuff here to follow, so I wasn't able to review everything in-depth. My main comment towards the beginning of the review was just trying to see if we can align on (int, str, or bool) instead of casting between them somehow. I didn't follow everywhere all of those cases are used.

imap_processing/mag/l1a/mag_l1a.py
maxinelasp: Yeah good point, I'll make it a boolean everywhere to be consistent.

imap_processing/mag/l1d/mag_l1d.py
maxinelasp: These are ancillary files. They match the ancillary file format so they should be output into the ancillary spot (I will check on this.) However, if you find it confusing, I can explicitly indicate these are ancillary files in the filename. It is confusing to output ancillary and science files in the same step.

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: No, for L1D, this is the only indication of quality they want to include.

imap_processing/glows/l1b/glows_l1b_data.py
tech3371, line 1: was GLOWS changes intentional?

tech3371 (commented):
There was a lot going on. I am lacking context and understanding. They look good to me on high level. @greglucas, if you can review and approve, I appreciate it!

imap_processing/glows/l1b/glows_l1b_data.py
maxinelasp: Nope, that was a merge fail. I'll remove them 🤦

maxinelasp:
Maybe @laspsandoval can give a more in-depth review, since the changes are pretty similar to I-ALiRT?

greglucas (approved):

PR #2038, opened 2025-08-12
MNT: Ruff target update

maxinelasp:
# Change Summary
Clean-up work to update ruff for Python 3.10 since we dropped 3.9 support. These updates did change a lot of files but they were entirely done by ruff's automatic formatter, so only the pyproject.toml and .pre-commit-config.yaml files need review. 

## Overview

- Update ruff pre-commit checker version
- Update ruff standards to 3.10

I hit this issue in a MAG Pr but since it touches a lot of files I wanted to open a new PR. I would appreciate a quick review.

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -412,6 +412,7 @@ class RectangularPointingSet(PointingSet):
         for dim, constructed_bins in zip(
             [CoordNames.AZIMUTH_L1C.value, CoordNames.ELEVATION_L1C.value],
             [self.sky_grid.az_bin_midpoints, self.sky_grid.el_bin_midpoints],
+            strict=False,
         ):
             if not np.allclose(
                 sorted(constructed_bins),
@@ -522,6 +523,7 @@ class UltraPointingSet(HealpixPointingSet):
         for dim, constructed_bins in zip(
             [CoordNames.AZIMUTH_L1C.value, CoordNames.ELEVATION_L1C.value],
             [azimuth_pixel_center, elevation_pixel_center],
+            strict=False,
         ):
             if not np.allclose(
                 self.data[dim],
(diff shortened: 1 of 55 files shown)


tmplummer (approved):
Nice!

imap_processing/ena_maps/ena_maps.py
```diff
@@ -252,7 +252,7 @@ def __init__(
         """Abstract method to initialize the pointing set object."""
         self.spice_reference_frame = spice_reference_frame
 
-        if isinstance(dataset, (str, Path)):
+        if isinstance(dataset, str | Path):
```
greglucas: Can you ignore rule [UP038](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#non-pep604-isinstance-up038)
It is deprecated to keep the standard checking style in place. That will save a bit of this churn.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -944,7 +944,7 @@ def get_spin_number(de_met: NDArray, de_spin: NDArray) -> NDArray:
     possible_spins = spin_numbers & 0xFF
 
     # Assign each group based on time.
-    for start, end in zip(spin_start_indices, spin_end_indices):
+    for start, end in zip(spin_start_indices, spin_end_indices, strict=False):
```
greglucas, line 947: Why did add `strict=False` everywhere? I think we want `strict=True` if anything, or to just ignore this for now... For example, I think that these start/end indices had better be the same length or we should raise the ValueError from `strict=True` if we want it.

https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/

greglucas (commented):
The type unions are a nice benefit 👍 I have two other suggestions though to update the ignore codes on.

imap_processing/ultra/l1b/ultra_l1b_extended.py
maxinelasp: This just makes explicit the default behavior, so I think we should leave people to update it to `strict=True` if they need that. I like that it makes it explicit

imap_processing/ena_maps/ena_maps.py
maxinelasp: Well that's annoying that they left it on by default even though it's deprecated. I'll ignore it and I guess revert and re-do all the updates

maxinelasp: Should be fixed now

greglucas (approved):

PR #2041, opened 2025-08-13
2039 hi l1a   mem dump packet parsing

### LINKED ISSUE #2039: Hi L1A - MEM_DUMP packet parsing
### Algorithm Description:

Parse MEM_DUMP packet that starts at a specific memory address. Sample code is attached.

[memdmp_c.txt](https://github.com/user-attachments/files/21743262/memdmp_c.txt)

tmplummer:
# Change Summary

## Overview
This adds parsing of certain MEMDMP packets to Hi L1A.

## New Files
- imap_processing/tests/hi/data/l0/imap_hi_l0_raw-memdmp_20260926_v003.pkts
   - Small test file containing Hi 90 and Hi 45 MEMDMP packets

## Updated Files
   - Add global attributes for L1A MEMDMP products
- imap_processing/hi/hi_l1a.py
   - Add function for parsing MEMDMP packets dumped memory
- imap_processing/hi/utils.py
   - Add MEMDMP packet apids to HIAPID enum
   - Add test coverage for paring MEMDMP packets

Closes: #2039


imap_processing/hi/hi_l1a.py
```diff
@@ -445,3 +488,77 @@ def unpack_hist_counter(counter_bytes: bytes) -> NDArray[np.uint16]:
...
+        "pkt_len",
+        "ccsds_met",
+        "cksum",
+    ]:
+        attrs = attr_mgr.get_variable_attributes(f"hi_hist_{var_name}")
+        dataset.data_vars[var_name].attrs.update(attrs)
+
+    new_vars = dict()
+    # Convert all the dump_data from all packets into a single uint32 array
+    full_uint32_data = np.frombuffer(dataset["dump_data"].data.sum(), dtype=">u4")
```
greglucas, line 552: The "sum()" here is interesting, can you explain exactly what that is doing? That seems like it might have simplified some other areas of our code previously that are doing to/from string on those quantities I think.

imap_processing/hi/hi_l1a.py
```diff
@@ -445,3 +488,77 @@ def unpack_hist_counter(counter_bytes: bytes) -> NDArray[np.uint16]:
...
+    index_stride = int(dataset["num_bytes"].data[0] // 4)
```
greglucas, line 555: Is this guaranteed to be (521 - 6)? packet-length - header

imap_processing/hi/hi_l1a.py
```diff
@@ -54,6 +54,46 @@
 )
 TOTAL_COUNTERS = ("a_total", "b_total", "c_total", "fee_de_recd", "fee_de_sent")
 
+# MEMDMP Packet definition of uint32 fields
```
greglucas, line 57: Can you add a comment about what these numbers mean, something like: The offset index (assuming all 4-byte fields) of where these fields are located within the packet

imap_processing/tests/hi/test_l1a.py
```diff
@@ -103,6 +103,20 @@ def test_app_hist_decom(hi_l0_test_data_path):
     assert cem_raw_cdf_filepath.name.startswith("imap_hi_l1a_90sensor-hist_")
 
 
+def test_memdmp_decom(hi_l0_test_data_path):
+    """Test memdmp data"""
+    bin_data_path = hi_l0_test_data_path / "imap_hi_l0_raw-memdmp_20260926_v003.pkts"
```
greglucas, line 108: If you have a raw packet file, do you have any spot checks of values you can assert against too?

imap_processing/hi/hi_l1a.py
tmplummer: I had to modify the TLM spreadsheet definition of the MEMDMP packet a bit to work with space_packet_parser. The modification was to convert a big fixed number of uint32 fields to a single BYTE binary blob. This allows the MEMDMP packet to have variable length which it does depending on which chunk of memory is being dumped to the packet. I am not sure how the OPS tools work with the supplied definition. I tried to get an answer to that here: https://lasp.colorado.edu/flares/browse/IMAPCATS-2237

Anyway, dump_data is a ndarray of `bytes` objects. `sum()` just concatenates the entire array into a single bytes object so that the parsing can be done without looping through packets.

I can add a comment to that effect.

imap_processing/hi/hi_l1a.py
tmplummer: There are a few fixed length telemetry items other than the variable length `dump_data` in the packet so I think that `num_bytes = packet_length - header - non-header-items` = 521 - 6 - 12. But, yes it should be a fixed stride for this flavor of MEMDMP packet.

imap_processing/tests/hi/test_l1a.py
tmplummer: I will be requesting that Paul provide validation data. In the past I have typically had a separate validation ticket where I add the validation data and add checks that all values match up.

See #2042

imap_processing/hi/hi_l1a.py
greglucas: Ahh, very nice usee of .sum() to concatenate the bytes. That was what was escaping me here! Thanks for pointing it out, might help to put a quick comment for our future selves.

greglucas (approved):
Content looks good. Feel free to add the validation data in a follow-up PR once you have it.

PR #2045, opened 2025-08-15
Create text file generation function

torimarbois:
# Change Summary

## Overview
This code wraps the existing ephemeris processing functions to produce pointing schedule text files intended to be used by ground stations to move their antenna dish. 

## Updated Files

- process_ephemeris.py
   - Add the generate_text_files() function
   - Small docstring changes
- test_process_ephemeris.py
   - Add a test for the new text file generation function

## Testing

New unit test for the text file generation function, asserts that expected strings are in the output file in the expected places.

### DIFF
--- a/imap_processing/tests/ialirt/unit/test_process_ephemeris.py
+++ b/imap_processing/tests/ialirt/unit/test_process_ephemeris.py
@@ -141,7 +141,7 @@ def test_build_output(furnish_kernels):
 
 
 @pytest.mark.external_kernel
-def test_generate_text_files(furnish_kernels):
+def test_generate_text_files(furnish_kernels, tmpdir):
     kernels = [
         "naif0012.tls",
         "pck00011.tpc",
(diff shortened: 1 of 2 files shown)


imap_processing/tests/ialirt/unit/test_process_ephemeris.py
```diff
@@ -138,3 +138,23 @@ def test_build_output(furnish_kernels):
...
+        "de440s.bsp",
+        "imap_spk_demo.bsp",
+        "earth_latest_high_prec.bpc",
+    ]
+    with furnish_kernels(kernels):
+        process_ephemeris.generate_text_files(
+            station="Kiel", day="20250813", file_path="../test_text_files"
+        )
+
+    with open("../test_text_files/20250813_Kiel.txt") as file:
```
greglucas: This should use the pytest tmpdir fixture, that way you don't get dangling files left around after a test potentially fails.

torimarbois: Done, thanks for the suggestion!

greglucas (approved):

laspsandoval (approved):
Fix test and you are finished! Nice work.

torimarbois:
This is frustrating, looks like the test failure is caused by `datetime` not being updated to the version that can handle different isoformats (like "20250815" instead of "2025-08-15"). Assuming that we don't want to update the datetime module right now, I'll just change the day string format.

PR #2046, opened 2025-08-15
Add scattering calibration data LUT

lacoak21:
# Change Summary

## Overview
Add the scattering calibration LUT containing phi and theta values in the instrument frame and their corresponding coefficients for the FWHM scattering equation. Add relevant functions to extract the data from the csv and calculate which pixels are below the scattering threshold. 

These functions are needed for the l1b scattering culling and the l1c deadtime correction. For more info see: [ultra_scattering_algo_update.pptx](https://github.com/user-attachments/files/21800812/ultra_scattering_algo_update.pptx)

## Updated Files
   - Add function to extract the relevant data from the LUT and find the a and g coefficient values corresponding to the nearest phi and theta values given. 
   - Add function to determine, whether the pixel's scattering value is below the corresponding threshold. 

## Testing
Add tests relating to the FWHM scattering LUT and threshold culling.

### DIFF
--- a/imap_processing/tests/external_test_data_config.py
+++ b/imap_processing/tests/external_test_data_config.py
@@ -100,6 +100,7 @@ EXTERNAL_TEST_DATA = [
     ("imap_ultra_l1b-45sensor-spbtphcorr_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b-90sensor-sptpphcorr_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b-90sensor-spbtphcorr_20250101_v000.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1b-90sensor-scattering-calibration-data_20250101_v000.csv", "ultra/data/l1/"),
     ("ultra45_raw_sc_ultrarawimg_withFSWccs_FM45_40P_Phi28p5_BeamCal_LinearScan_"
      "phi2850_theta-000_20240207T102740_revised20250724.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b-45sensor-tofxeflat_20250101_v000.pgm", "ultra/data/l1/"),
--- a/imap_processing/tests/ultra/unit/test_lookup_utils.py
+++ b/imap_processing/tests/ultra/unit/test_lookup_utils.py
@@ -16,6 +16,7 @@ from imap_processing.ultra.l1b.lookup_utils import (
     get_image_params,
     get_norm,
     get_ph_corrected,
+    get_scattering_calibration_data,
     get_y_adjust,
 )
 
@@ -166,3 +167,17 @@ def test_get_ebins(ancillary_files):
     ebins = get_ebins("l1b-tofxph", energy, ctof, ebins, ancillary_files)
 
     np.testing.assert_array_equal(ebins, np.array([15, 19]))
+
+
+@pytest.mark.external_test_data
+def test_get_scattering_data(ancillary_files):
+    """Tests function get_scattering_data."""
+
+    (a_theta_val, g_theta_val, a_phi_val, g_phi_val) = get_scattering_calibration_data(
+        ancillary_files,
+        "l1b-90sensor-scattering-calibration",
+        np.array([47, 43]),
+        np.array([43, 42]),
+    )
+    np.testing.assert_array_equal(a_theta_val, np.array([np.nan, 35.23100]))
+    np.testing.assert_array_equal(g_theta_val, np.array([np.nan, -0.72148]))
--- a/imap_processing/tests/ultra/unit/test_lookup_utils.py
+++ b/imap_processing/tests/ultra/unit/test_lookup_utils.py
@@ -16,8 +16,9 @@ from imap_processing.ultra.l1b.lookup_utils import (
     get_image_params,
     get_norm,
     get_ph_corrected,
-    get_scattering_calibration_data,
+    get_scattering_coefficients,
     get_y_adjust,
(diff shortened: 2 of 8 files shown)


imap_processing/tests/external_test_data_config.py
```diff
@@ -100,6 +100,7 @@
     ("imap_ultra_l1b-45sensor-spbtphcorr_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b-90sensor-sptpphcorr_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b-90sensor-spbtphcorr_20250101_v000.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1b-90sensor-scattering-calibration-data_20250101_v000.csv", "ultra/data/l1/"),
```
greglucas, line 103: Do you need the 45 one as well?

imap_processing/tests/ultra/unit/test_lookup_utils.py
```diff
@@ -166,3 +168,51 @@ def test_get_ebins(ancillary_files):
...
+        ]
+    )
+    # Only pixels where both the theta and phi coefficients are below the FWHM
+    # threshold should be True.
+    expected_pixel_mask = np.array([False, True, False])
+    pixel_mask = pixels_below_fwhm_scattering_threshold(
+        theta_coeffs, phi_coeffs, energy
+    )
+    assert pixel_mask.shape == (3,)
+    np.testing.assert_array_equal(pixel_mask, expected_pixel_mask)
```
greglucas: The assert_array_equal takes care of the shape mismatch as well.

imap_processing/ultra/constants.py
```diff
@@ -129,3 +130,11 @@ class UltraConstants:
         341.989454569026,
         1e5,
     ]
+
+    ULTRA_FWHM_SCATTERING_CULLING_THRESHOLDS: ClassVar[dict] = {
```
greglucas, line 136: Might be nice to add a quick comment about what the indices are here for quick reference.

would this be better as a list that you could do a searchsorted on to find the bin?

```
CULLING_VALUES = np.array([-np.inf, 1, 5, 8, 10, 20])
CULLING_THRESHOLDS = np.array([np.nan, 12, 10, 8, 6, 4])
```

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -281,19 +282,144 @@ def get_geometric_factor(
...
+    # Calculate FWHM for theta and phi
+    fwhm_theta = theta_coeffs[:, 0] * energy ** theta_coeffs[:, 1]
+    fwhm_phi = phi_coeffs[:, 0] * energy ** phi_coeffs[:, 1]
+
+    # Get the scattering threshold based on the energy
+    threshold = next(
+        threshold
+        for energy_range, threshold in scattering_thresholds.items()
+        if energy_range[0] <= energy < energy_range[1]
+    )
```
greglucas, line 395: OK, I see where this is used now and it is only a single lookup based on an integer energy and not an array-based lookup. You can probably ignore my other comment unless you like that approach better, but this seems fine to me as-is.

greglucas (approved):

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -281,19 +282,144 @@ def get_geometric_factor(
...
+        Elevation angles in degrees.
+    phi : NDArray
+        Azimuth angles in degrees.
+
+    Returns
+    -------
+    tuple
+        Scattering a and g values corresponding to the given theta and phi values.
+    """
+    filename = f"l1b-{instrument_id}sensor-scattering-calibration"
```
laspsandoval: Really minor nit. This is descriptor and not filename.

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -281,19 +282,144 @@ def get_geometric_factor(
...
+    return np.logical_and(fwhm_theta <= threshold, fwhm_phi <= threshold)
+
+
+def is_inside_fov(phi: np.ndarray, theta: np.ndarray) -> np.ndarray:
+    """
+    Determine angles in the field of view (FOV).
+
+    More information can be found by looking at equation 19 in the Ultra Algorithm
+    Document.
+
```
laspsandoval, line 408: At each step, use the theta / phi dependence that defines the Field of View (FOV) as a selection criteria in which to accumulate time into that pixel or not.

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -281,19 +282,144 @@ def get_geometric_factor(
...
+    g_theta_val = g_theta[phi_idx, theta_idx]
+    a_phi_val = a_phi[phi_idx, theta_idx]
+    g_phi_val = g_phi[phi_idx, theta_idx]
+
+    return np.column_stack([a_theta_val, g_theta_val]), np.column_stack(
+        [a_phi_val, g_phi_val]
+    )
+
+
+def pixels_below_fwhm_scattering_threshold(
```
lacoak21: This will be direct events and not pixels for l1b

laspsandoval (commented):
Looks good. Minor comment.

imap_processing/tests/external_test_data_config.py
lacoak21: The 45 one has not been delivered yet. @laspsandoval should I update the function to return the 45 one temporarily for both ids? I think it would error out right now if it tried to get the 45 file.

PR #2049, opened 2025-08-15
I-ALiRT - MAG combination and minor updates

### LINKED ISSUE #1330: I-ALiRT - MAG Algorithm
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
I combined the algorithms I had finished for MAG and along the way found a few things that needed to be changed.

## Updated Files
- parse_mag.py
   - Updated the structure of the data that I passed to calculate_gradiometry_offsets and how i calculated the magntitude
   - Changed the order of when I transform to different frames (GSE, GSM, RTN) to the end
   - Added a new calibration file for input into the process_packet function
   - Took out l1b data being returned (cleanup)
   - Calculated phi and theta in GSM, GSE coordinates
   - Added relevant functions in the correct order

- mag_l1d_data.py
   - Added a boolean for cases which we are only finding the gradiometer for two vectors (each with a single timestamp)
  
## Testing
test_parse_mag.py

Ran the code and corrected for nans. The data looks good, but I'll request validation data from MAG.

### DIFF
--- a/imap_processing/ialirt/l0/parse_mag.py
+++ b/imap_processing/ialirt/l0/parse_mag.py
@@ -506,7 +506,7 @@ def transform_to_frames(
 
 def process_packet(
     accumulated_data: xr.Dataset, calibration_dataset: xr.Dataset
-) -> tuple[list[dict], list[dict]]:
+) -> list[dict]:
     """
     Parse the MAG packets.
 
@@ -541,7 +541,6 @@ def process_packet(
     grouped_data = find_groups(accumulated_data, (0, 3), "pkt_counter", "met")
 
     unique_groups = np.unique(grouped_data["group"])
-    l1b_data = []
     mag_data = []
 
     for group in unique_groups:
@@ -600,8 +599,6 @@ def process_packet(
             }
         )
 
-        l1b_data.append({**status_data, **science_data, **time_data})
-
         # Placeholder for real data.
         met = grouped_data["met"][(grouped_data["group"] == group).values]
         mag_data.append(
@@ -623,7 +620,7 @@ def process_packet(
             }
         )
 
-    return mag_data, l1b_data
+    return mag_data
 
 
 def retrieve_matrix_from_single_l1b_calibration(
--- a/imap_processing/ialirt/l0/parse_mag.py
+++ b/imap_processing/ialirt/l0/parse_mag.py
@@ -342,9 +342,10 @@ def calibrate_and_offset_vectors(
 
 
 def apply_gradiometry_correction(
-    mago_vector_eclipj2000: np.ndarray,
-    magi_vector_eclipj2000: np.ndarray,
-    time_data: dict,
+    mago_vectors_eclipj2000: np.ndarray,
+    mago_time_data: np.ndarray,
+    magi_vectors_eclipj2000: np.ndarray,
+    magi_time_data: np.ndarray,
     gradiometer_factor: np.ndarray,
 ) -> tuple[np.ndarray, np.ndarray]:
     """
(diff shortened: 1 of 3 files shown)


imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -328,11 +324,9 @@ def calibrate_and_offset_vectors(
     calibrated_and_offset_vectors : np.ndarray
         Calibrated and offset vectors, shape (n, 3).
     """
-    # Append range as 4th column
-    vec_plus_range = np.concatenate((vectors, range_vals[:, np.newaxis]), axis=1)
-
     # Apply calibration matrix -> (n,4)
-    calibrated = MagL2L1dBase.apply_calibration(vec_plus_range, calibration)
+    # apply_calibration_offset_single_vector
+    calibrated = MagL2L1dBase.apply_calibration(vectors.reshape(1, 4), calibration)
```
maxinelasp, line 329: Does this line drop the other n-1 vectors that are getting passed in?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -589,18 +609,61 @@ def process_packet(
...
-        science_data.update(
-            {
-                "calibrated_pri_x": updated_vector_mago[0],
-                "calibrated_pri_y": updated_vector_mago[1],
-                "calibrated_pri_z": updated_vector_mago[2],
-                "calibrated_sec_x": updated_vector_magi[0],
-                "calibrated_sec_y": updated_vector_magi[1],
-                "calibrated_sec_z": updated_vector_magi[2],
-            }
+        mago_calibration = l1d_calibration_dataset["URFTOORFO"][0]
```
maxinelasp, line 619: Does this end up in the right shape? This calibration matrix should look like (3, 3, 4), so I'm a little confused as to why you need to index in with `[0]`.

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -533,11 +537,14 @@ def calculate_gradiometry_offsets(
...
             - gradiometer_offsets: the offset values (MAGi - MAGo) for each axis
         """
-        aligned_magi = linear(
-            magi_vectors,
-            magi_epoch,
-            mago_epoch,
-        )
+        aligned_magi = magi_vectors
+
+        if align_timestamps:
```
maxinelasp: given what Alastair mentioned - is this still necessary?

maxinelasp (commented):
Just not approving since Alastair requested some changes in slack, and I had a few comments. Overall this looks pretty good to me though

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: There will only be a single vector being passed in here. The reason for this is that I am accumulating packets into a "group". We need 4 packets to get a complete set of data. So I have to do this for a single vector rather than a group, which is probably what you see in regular MAG processing.

imap_processing/mag/l1d/mag_l1d_data.py
laspsandoval: Nope. Changed it.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: l1d_calibration_dataset["URFTOORFO"].shape
(1, 3, 3, 4)
l1d_calibration_dataset["URFTOORFO"][0].shape
(3, 3, 4)

imap_processing/ialirt/l0/parse_mag.py
maxinelasp: Ok, can you update the docs to make that clear? Right now it's saying you're expecting vectors of shape (n, 4). Then, you shouldn't need to reshape here and you can just pass it along directly

maxinelasp (approved):
LGTM!

PR #2051, opened 2025-08-18
CoDICE: Add some placeholder code for L2 processing

bourque:
This PR adds some placeholder code for processing all of the L2 data products. I am hoping these changing will be helpful to other developers who may help out with implementing CoDICE code, so they know what needs to be implemented and where it should be implemented in the code.


vmartinez-cu (commented):
These are helpful placeholders. Thanks for adding this in!

vmartinez-cu:
Looks like a couple checks are failing

tech3371 (approved):
Thank you for detail notes and sections. That's gonna help us! I personally feel like we can merge over the test coverage since this is setting us up for future work and then future work can add tests.

Issue #1733, opened 2025-05-14
Create Utility for Repointing Start Time


sdhoyt:
The ENA Pointing Sets will contain 1 epoch per product. The epoch time for each pointing set should be the start of the pointing. A shared utility should be created that can be used by all the ENAs to get the pointing start time by passing in a DE epoch from that pointing. 

This issue is related to the discussion here: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1714#discussion_r2085137391

and the Hi issue: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1641

Added by @subagonsouth 
The current definition of a Pointing that we are using is the time from the end of one repoint to the end of the next repoint. So, the function for this ticket can use the existing `imap_processing.spice.repoint.interpolate_repoint_data()` function and simply return the `repoint_end_met` column (I think that returning MET time would be reasonable ~Tim).

sdhoyt:
Closed by #2027

PR #2061, opened 2025-08-18
Lo L1C - Pointing Set pointing times and spin numbers

### LINKED ISSUE #1900: Lo L1C - Add start and end spin numbers to PSET
The Lo PSET needs to have the Start/End Spin numbers added. Information on this can be found in the PSET section of the Lo algorithm document in section 10.1.1: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/146768943/IMAP+Algorithm+Document+from+Instrument+Teams

The start/end numbers are defined as follows:
Start Spin number--The spin number at the start of the pointing
End Spin number--The spin number at the end of the pointing

The start of the pointing can be determined by comparing finding the closes repointing end in the repointing table to the first epoch in the data and confirming that the epoch is between the end of the repointing and the next start pointing. Knowing the end of the repointing, you can find the start of the spin by doing the same thing with the spin table

### LINKED ISSUE #2027: Lo L1C - Add Epoch and Star/End Pointing Times
- There is one epoch time per pointing set CDF. The epoch time for the pointing set is currently set as the first epoch in the L1B direct event data. This needs to be changed to be the time when the pointing starts.

- The start/end pointing times needs to be set. The start of the pointing should be the end of the repointing in the repoint table and the end of the pointing is the repoint start in the repoint table that occurs after the start of the pointing. The epoch and the start pointing time should be equal.

The repoint table should be used for this.

Some related code has been produced for the spin number updates and should likely be pulled into these changes rather than the spin number branch:
- https://github.com/sdhoyt/imap_processing/blob/pset-spin/imap_processing/lo/l1c/lo_l1c.py

The unit tests have an example of how to use the conftest fixtures to create a fake repointing table to use in the unit test:
- https://github.com/sdhoyt/imap_processing/blob/pset-spin/imap_processing/tests/lo/test_lo_l1c.py

sdhoyt:
# Change Summary

## Overview
This PR adds code to create the pointing start/end times fields and the spin numbers fields.

The pointing start field contains the time that the pointing starts and the pointing end field contains the time that the pointing ends.

The spin start number contains the spin number at the start of the pointing and the spin end number contains the spin number at the end of the pointing.

## Updated Files
- lo_l1c.py
   - added functions for the spin start/end fields and the pointing start/end fields
   - Set the epoch as the pointing start field

## Testing
Add unit tests for each function added and updated the test_lo_l1c unit test.

Closes #2027
Closes #1900

### DIFF
--- a/imap_processing/lo/l1c/lo_l1c.py
+++ b/imap_processing/lo/l1c/lo_l1c.py
@@ -10,7 +10,10 @@ from scipy.stats import binned_statistic_dd
 
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.spice.time import met_to_ttj2000ns
-
+from imap_processing.spice.spin import (
+    get_spin_data,
+)
+from imap_processing.spice.repoint import get_repoint_data
 
 class FilterType(str, Enum):
     """
@@ -56,6 +59,7 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
         l1b_goodtimes_only = filter_goodtimes(l1b_de, anc_dependencies)
         pset = initialize_pset(l1b_goodtimes_only, attr_mgr, logical_source)
         full_counts = create_pset_counts(l1b_goodtimes_only)
+        start_spin_number, end_spin_number = get_spin_numbers(l1b_goodtimes_only)
         pset["triples_counts"] = create_pset_counts(
             l1b_goodtimes_only, FilterType.TRIPLES
         )
@@ -254,6 +258,27 @@ def create_pset_counts(
 
     return counts
 
+def get_spin_numbers(l1b_de: xr.Dataset) -> tuple[int, int]:
+    """
+    Get the start and end spin numbers from the L1B Direct Event dataset.
+
+    The spin numbers are used to identify the start and end of the
+    pointing periods in the dataset.
+
+    Parameters
+    ----------
+    l1b_de : xarray.Dataset
+        L1B Direct Event dataset.
+
+    Returns
+    -------
+    tuple[int, int]
+        The start and end spin numbers.
+    """
+    repoint_df = get_repoint_data()
+    print(repoint_df.info())
+
+    #return start_spin_number, end_spin_number
 
 def calculate_exposure_times(counts: xr.DataArray, l1b_de: xr.Dataset) -> xr.DataArray:
     """
--- a/imap_processing/lo/l1c/lo_l1c.py
+++ b/imap_processing/lo/l1c/lo_l1c.py
@@ -279,9 +279,12 @@ def get_spin_numbers(l1b_de: xr.Dataset) -> tuple[int, int]:
     spin_df = get_spin_data()
 
     # Convert repoint and spin MET to TTJ2000NS
-    repoint_df["repoint_end_ttj2000ns"] = met_to_ttj2000ns(repoint_df["repoint_end_met"])
+    repoint_df["repoint_end_ttj2000ns"] = met_to_ttj2000ns(
+        repoint_df["repoint_end_met"]
+    )
     repoint_df["repoint_start_ttj2000ns"] = met_to_ttj2000ns(
-        repoint_df["repoint_start_met"])
+        repoint_df["repoint_start_met"]
+    )
     spin_df["spin_start_ttj2000ns"] = met_to_ttj2000ns(spin_df["spin_start_met"])
 
     first_epoch = l1b_de["epoch"][0].item()
@@ -292,37 +295,75 @@ def get_spin_numbers(l1b_de: xr.Dataset) -> tuple[int, int]:
     pointing_start_spin_idx = find_spin_idx(spin_df, pointing_start)
     pointing_end_spin_idx = find_spin_idx(spin_df, pointing_end, start=False)
 
-    start_spin_number = spin_df["spin_number"].iloc[
-        pointing_start_spin_idx] if pointing_start_spin_idx is not None else None
-    end_spin_number = spin_df["spin_number"].iloc[
-        pointing_end_spin_idx] if pointing_end_spin_idx is not None else None
+    start_spin_number = (
+        spin_df["spin_number"].iloc[pointing_start_spin_idx]
+        if pointing_start_spin_idx is not None
+        else None
+    )
+    end_spin_number = (
+        spin_df["spin_number"].iloc[pointing_end_spin_idx]
+        if pointing_end_spin_idx is not None
+        else None
+    )
 
     print(f"Start spin number: {start_spin_number}")
 
     return start_spin_number, end_spin_number
 
+
 def pointing_times(df, epoch):
+    """
+    Get the pointing start and end times for a given epoch.
+    The pointing start time is the end of the repointing period
+    and the pointing end time is the start of the next repointing period.
+
+    Parameters
+    ----------
+    df : pandas.DataFrame
+        DataFrame containing the repointing data with columns:
+            - repoint_end_ttj2000ns
+            - repoint_start_ttj2000ns
+    epoch : int
+        The epoch time in TTJ2000NS to find the pointing start and end times for
+    Returns
+    -------
+    tuple[int, int]
+        The pointing start and end times in TTJ2000NS.
+        If no pointing times are found, both values will be None.
+    """
     mask_start = (df["repoint_end_ttj2000ns"] <= epoch) & (
-                df["repoint_start_ttj2000ns"].shift(-1) >= epoch)
+        df["repoint_start_ttj2000ns"].shift(-1) >= epoch
+    )
     mask_end = (df["repoint_end_ttj2000ns"] <= epoch) & (
-                df["repoint_start_ttj2000ns"].shift(-1) > epoch)
+        df["repoint_start_ttj2000ns"].shift(-1) > epoch
+    )
 
     if all(mask_start == mask_end):
         repoint_idx = mask_start.idxmax() if mask_start.any() else None
-        pointing_start = df["repoint_end_ttj2000ns"].iloc[
-            repoint_idx] if repoint_idx is not None else None
-        pointing_end = df["repoint_start_ttj2000ns"].iloc[
-            repoint_idx + 1] if repoint_idx is not None else None
+        pointing_start = (
+            df["repoint_end_ttj2000ns"].iloc[repoint_idx]
+            if repoint_idx is not None
+            else None
+        )
+        pointing_end = (
+            df["repoint_start_ttj2000ns"].iloc[repoint_idx + 1]
+            if repoint_idx is not None
+            else None
+        )
     return pointing_start, pointing_end
+
 
 def find_spin_idx(df, time, start=True):
     if start:
         mask = (df["spin_start_ttj2000ns"] <= time) & (
-                    df["spin_start_ttj2000ns"].shift(-1) > time)
+            df["spin_start_ttj2000ns"].shift(-1) > time
+        )
     else:
         mask = (df["spin_start_ttj2000ns"] <= time - 1) & (
-                    df["spin_start_ttj2000ns"].shift(-1) >= time - 1)
+            df["spin_start_ttj2000ns"].shift(-1) >= time - 1
+        )
     return mask.idxmax() if mask.any() else None
+
 
 def calculate_exposure_times(counts: xr.DataArray, l1b_de: xr.Dataset) -> xr.DataArray:
     """
--- a/imap_processing/lo/l1c/lo_l1c.py
+++ b/imap_processing/lo/l1c/lo_l1c.py
@@ -9,11 +9,10 @@ import xarray as xr
 from scipy.stats import binned_statistic_dd
 
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
-from imap_processing.spice.time import met_to_ttj2000ns
-from imap_processing.spice.spin import (
-    get_spin_data,
-)
 from imap_processing.spice.repoint import get_repoint_data
+from imap_processing.spice.spin import get_spin_data
+from imap_processing.spice.time import met_to_ttj2000ns
+
 
 class FilterType(str, Enum):
     """
@@ -258,6 +257,7 @@ def create_pset_counts(
 
     return counts
 
+
 def get_spin_numbers(l1b_de: xr.Dataset) -> tuple[int, int]:
     """
     Get the start and end spin numbers from the L1B Direct Event dataset.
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "poetry_dynamic_versioning.backend"
 [tool.poetry]
 name = "imap-processing"
 # Gets updated dynamically by the poetry-dynamic-versioning plugin
-version = "0.0.0"
+version = "0.6.0.post483.dev0+03229ea"
```
sdhoyt: I'm not sure why this changed. Could I go in and manually revert this or do I need to do something with poetry as well?

greglucas: No, this should not change and shouldn't be committed. It is dynamically resolved at runtime, feel free to manually change it back.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -56,6 +58,24 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
         l1b_goodtimes_only = filter_goodtimes(l1b_de, anc_dependencies)
         pset = initialize_pset(l1b_goodtimes_only, attr_mgr, logical_source)
         full_counts = create_pset_counts(l1b_goodtimes_only)
+
+        # Set the pointing start and end times based on the first epoch
```
greglucas, line 61: Can we use the pointing number instead? These products are all going to be built from products with a repointing number flag in it. So would it make sense to flow that through and do the lookup based on pointing number?

I think this is related to: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/1733
Is there a reason why we aren't using the interpolate_spin_data function like mentioned in that comment? I think your function below is somewhat duplicating that effort.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -255,6 +278,111 @@ def create_pset_counts(
...
+    # Get the spin table
+    spin_df = get_spin_data()
+
+    # create a mask for the spin that contains the start of the pointing
+    mask_start = (spin_df["spin_start_met"] <= pointing_start) & (
+        spin_df["spin_start_met"].shift(-1) > pointing_start
+    )
+    # create a masks for the spin that contains the end of the pointing
+    mask_end = (spin_df["spin_start_met"] <= pointing_end) & (
+        spin_df["spin_start_met"].shift(-1) >= pointing_end
```
greglucas: Does this need to be +1 to shift it to the right?
```suggestion
        spin_df["spin_start_met"].shift(1) >= pointing_end
```

Maybe more critically, do you need the second checks here, or can you just grab all values <= pointing_end and then do an idxmax() on it because that is the index you want.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -255,6 +278,111 @@ def create_pset_counts(
...
+    )
+
+    return (
+        xr.DataArray(np.array([pointing_start]), dims=["epoch"]),
+        xr.DataArray(np.array([pointing_end]), dims=["epoch"]),
+    )
+
+
+def get_spin_numbers(
+    pointing_start: int, pointing_end: int
```
tmplummer: I would expect the pointing start and end MET times to be floats.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -255,6 +278,111 @@ def create_pset_counts(
...
+    # create a mask for the spin that contains the start of the pointing
+    mask_start = (spin_df["spin_start_met"] <= pointing_start) & (
+        spin_df["spin_start_met"].shift(-1) > pointing_start
+    )
+    # create a masks for the spin that contains the end of the pointing
+    mask_end = (spin_df["spin_start_met"] <= pointing_end) & (
+        spin_df["spin_start_met"].shift(-1) >= pointing_end
+    )
+    # get the index for the spin that contains the start and end of the pointing
+    spin_start_idx = mask_start.idxmax() if mask_start.any() else None
```
greglucas: Shouldn't this be the minimum value of the condition to get the start?
```suggestion
    spin_start_idx = mask_start.idxmin() if mask_start.any() else None
```

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -255,6 +278,111 @@ def create_pset_counts(
...
+    spin_end_idx = mask_end.idxmax() if mask_end.any() else None
+
+    # get the spin numbers for the start and end of the pointing
+    start_spin_number = (
+        spin_df["spin_number"].iloc[spin_start_idx]
+        if spin_start_idx is not None
+        else None
```
greglucas: Is `None` OK here or do you want to put in a fill value instead? This changes the datatype of an array to use None

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -255,6 +278,111 @@ def create_pset_counts(
...
+    )
+
+    return (
+        xr.DataArray(np.array([pointing_start]), dims=["epoch"]),
+        xr.DataArray(np.array([pointing_end]), dims=["epoch"]),
+    )
+
+
+def get_spin_numbers(
```
tmplummer: I think that the existing `spin.interpolate_spin_data()` function could be used here. But your implementation also looks fine.
```
start_end_spin_data = interpolate_spin_data(np.array([pointing_start, pointing_end])
start_spin_number = start_end_spin_data["spin_number"].iloc[0]
end_spin_number = start_end_spin_data["spin_number"].iloc[1]
```

imap_processing/tests/conftest.py
```diff
@@ -528,7 +528,10 @@ def generate_repoint_data(
         Repoint dataframe with start and end repoint times provided and incrementing
         repoint_ids starting at 1.
     """
-    repoint_start_times = np.array(repoint_start_met)
+    if isinstance(repoint_start_met, float):
+        repoint_start_times = np.array(repoint_start_met)
+    elif isinstance(repoint_start_met, np.ndarray):
+        repoint_start_times = repoint_start_met
```
greglucas: I think this could be `np.atleast_1d(repoint_start_met)`?

tmplummer:
```suggestion
    repoint_start_times = np.atleast_1d(repoint_start_met)
```

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -255,6 +278,111 @@ def create_pset_counts(
     return counts
 
 
+def get_pointing_times(epoch: int) -> tuple[xr.DataArray, xr.DataArray]:
```
tmplummer: I think that this function could be moved into the repoint.py module if the query time was MET. The reason for MET is that it removes any SPICE dependency from that module.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -255,6 +278,111 @@ def create_pset_counts(
...
+    Returns
+    -------
+    tuple[xarray.DataArray, xarray.DataArray]
+        The pointing start and end times in TTJ2000NS.
+        If no pointing times are found, both values will be None.
+    """
+    # Get the repoint table
+    repoint_df = get_repoint_data()
+
+    # Convert repoint and spin MET to TTJ2000NS to compare with the epoch
```
tmplummer: Probably very minor, but it is cheaper to just convert the query time from TTJ2000ns to MET.
```
met_time = ttj2000ns_to_met(epoch)
```

tmplummer (commented):
Overall, this looks great. I have a few minor nit picks.

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: I forgot about the pointing number. I don't have that in my file names yet, but that would be better to use. 

I also forgot about that ticket. However, the interpolate_repoint_data uses MET times and I only have TTJ2000 at this point. I don't think we have function to convert from TTJ2000 to MET.

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: I don't see any combination of time functions to convert from TTJ200 to MET. Are you retaining MET times throughout processing? I don't have any MET data at this point in the processing.

tmplummer: The time.py module should be reorganized to make things easier to find, but see here: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/time.py#L244

sdhoyt: Oh man, I've looked for that before and have no idea how I missed that. Thank you!

tmplummer: Thank @maxinelasp, she implemented that. You probably missed it in the past because it was only merged three weeks ago.

If you ever finding yourself wishing there was a time conversion that isn't implemented, please ping me or write a ticket and assign it to me.

sdhoyt:
I think I made a mistake listening to git tell me to do `git config pull.rebase true` and pulling from my remote branch to fix some divergence. I'm working to remove all the other changes that were pulled into this PR

Update: I thought that would be harder to fix

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -52,10 +54,45 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
...
+        pset["pointing_start_met"] = xr.DataArray(
+            np.array([pointing_start_met]),
+            dims="epoch",
+            # attrs=attr_mgr.get_variable_attributes("pointing_start_met)"
+        )
+
+        # set the pointing end met as the start of the next repointing
+        pset["pointing_end_met"] = get_pointing_end_time(
+            pset["pointing_start_met"].item()
+        )
```
greglucas: This was a little confusing to create an xarray in one case, but not the other so they look like different call signatures. I think you previously had this as `get_pointing_start_end_time()` to return both the start and end of a pointing which I think makes sense for one call instead. Could you move the end calculation into the repoint.py function now?

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -52,10 +54,45 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
...
+
+        # pset["pointing_end_met"] = xr.DataArray(
```
greglucas:
```suggestion
```

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -52,10 +54,45 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
...
+        # Set the epoch to the start of the pointing
+        pset["epoch"] = xr.DataArray(
+            met_to_ttj2000ns(pset["pointing_start_met"].values),
+            attrs=attr_mgr.get_variable_attributes("epoch"),
+        )
+
+        # Get the start and end spin numbers based on the pointing start and end MET
+        pset["start_spin_number"] = xr.DataArray(
+            np.array([get_spin_number(pset["pointing_start_met"].item())]),
```
greglucas: What happens if you don't pass in `.item()` or pass in `.values` instead, would you be able to avoid the `np.array([])` cast?

This function seems good to call it separately to get the specific spin during that MET, so 👍 to not consolidating here.

imap_processing/spice/repoint.py
```diff
@@ -193,3 +193,21 @@ def interpolate_repoint_data(
     out_df["repoint_in_progress"] = query_met_times < out_df["repoint_end_met"].values
 
     return out_df
+
+
+def get_pointing_start_time(query_met_time: float) -> tuple[float, float]:
+    """
+    Get the start and end MET times for the pointing that contains the query MET time.
```
greglucas, line 200: start and end should both be here.

imap_processing/spice/repoint.py
```diff
@@ -193,3 +193,21 @@ def interpolate_repoint_data(
     out_df["repoint_in_progress"] = query_met_times < out_df["repoint_end_met"].values
 
     return out_df
+
+
+def get_pointing_start_time(query_met_time: float) -> tuple[float, float]:
```
greglucas: Remove "query" from the variable arguments to the function, you are just taking in a met_time.
```suggestion
def get_pointing_start_time(met_time: float) -> tuple[float, float]:
```

imap_processing/tests/spice/test_spin.py
```diff
@@ -97,6 +97,15 @@ def test_interpolate_spin_data(query_met_times, expected, fake_spin_data):
         )
 
 
+def test_get_spin_number(fake_spin_data):
```
greglucas: Parameterize over a few values so you can test a few different cases? (Same with repoint test possibly)
```suggestion
@pytest.mark.parameterize("met_time,spin_number", [(0, 0), (14.9, 0), (15.1, 1), (32, 2)]
def test_get_spin_number(fake_spin_data, met_time, spin_number):
```

imap_processing/spice/repoint.py
sdhoyt: I started doing start and end for this function, but realized the interpolate function it uses only returns the row with the start time. I didn't think I should change the function of interpolate for this which I think I'll need to do to get the pointing end time. I could move the pointing end time function as it is in my L1C code to pass in the pointing start_time and get the pointing end time. Would that work for you?

greglucas: > I didn't think I should change the function of interpolate for this which I think I'll need to do to get the pointing end time

I agree with this, don't change the `interpolate()` function.

> I could move the pointing end time function as it is in my L1C code to pass in the pointing start_time and get the pointing end time. Would that work for you?

I'm not sure I exactly follow this. I'm suggesting to move the logic within that separate function into this function here. So the logic in the other function is good, and you have the start time here already. So rather than two calls, it'd just be a single external call.

sdhoyt: Just to make sure I understand, you're talking about merging the end pointing function into the start pointing function in repoint.py?
```python
def get_pointing_start_time(met_time: float) -> tuple[float, float]:
    """
    Get the start and end MET times for the pointing that contains the query MET time.

    Parameters
    ----------
    met_time : float
        The MET time in a pointing.

    Returns
    -------
    pointing_start_time : float
        The MET time of the repoint maneuver that ends before the query MET time.
    pointing_end_time : float
        The MET time of the repoint maneuver that starts after the query MET time.
    """
    # Find the pointing start time by finding the repoint end time
    repoint_df = interpolate_repoint_data(met_time)
    pointing_start_met = repoint_df["repoint_end_met"].item()
    # Find the pointing end time by finding the next repoint start time
    repoint_df = get_repoint_data()
    pointing_idx = repoint_df.index[
        repoint_df["repoint_end_met"] == pointing_start_met
    ][0]
    pointing_end_met = np.array([repoint_df["repoint_start_met"].iloc[pointing_idx + 1]]),
    return pointing_start_met, pointing_end_met
```

greglucas: I think that is nicer to have things all consolidated in one location, so yes that makes sense to me. (I think we should really make a `pointing_function()` helper rather than a repointing helper in the future, and this would help us towards that potential future refactor.

sdhoyt:
I believe I've addressed all the comments now

greglucas (approved):

greglucas:
The Ultra failures are unrelated and I think will be fixed by the merge into dev. Merging over those failures.

PR #2070, opened 2025-08-20
MNT: Remove extraneous ultra import log

greglucas:
This was being printed to all instrument log streams, not just ultras. I noticed this in the idex logs.


lacoak21 (approved):
👍

PR #2072, opened 2025-08-20
IDEX handle missing science acquisition events 

lacoak21:
# Change Summary

## Overview
If there are no science acquisition events found, rate variables can not be computed. In this case only return the datasets with the counts variables. 

## Updated Files
   - Log warning if there are missing events 
   - return only counts if rates can not be computed

### DIFF
--- a/imap_processing/idex/idex_l2b.py
+++ b/imap_processing/idex/idex_l2b.py
@@ -141,9 +141,9 @@ def idex_l2b(
         daily_on_percentage,
     )
     # Create l2b Dataset
-    charge_bins = CHARGE_BIN_EDGES[1:]
-    mass_bins = CHARGE_BIN_EDGES[1:]
-    spin_phase_bins = SPIN_PHASE_BIN_EDGES[1:]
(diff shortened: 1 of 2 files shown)


imap_processing/idex/idex_l2b.py
```diff
@@ -127,20 +127,71 @@ def idex_l2b(
...
-        rate_quality_flags,
-    ) = compute_rates_by_charge_and_mass(
-        counts_by_charge,
-        counts_by_mass,
-        counts_by_charge_map,
-        counts_by_mass_map,
-        epoch_doy_unique,
-        daily_on_percentage,
-    )
+    if daily_on_percentage is None:
```
greglucas: What happens if you have a `daily_on_percentage == 0` instead of None? Would that allow you to keep going here and you'd just get some divide by zero warnings and it would be filled with nan's?

lacoak21: Good point. They requested fill vals of -1 for any missing rates - this is what I should have done in the first place.

greglucas (approved):
Nice, I like this approach better, thanks!

PR #1973, opened 2025-07-22
CoDICE: update compression lookup tables

bourque:
This PR updates the values of the CoDICE Lossy A and Lossy B compression lookup tables and corresponding test values.

Closes #1947

PR #2079, opened 2025-08-20
CoDICE L1a: Turn some validation steps back on

bourque:
This PR turns back on some validation steps to test the latest and greatest CoDICE L1a validation files. A lot of the variable and metadata names in the new validation files now use `lowercase_and_underscore` naming conventions instead of `CamelCase`, so a lot of these changes are a result of that.

The validation step to compare the data array values is still turned off -- there are some data products that don't currently pass this check that I need to sort out still (that will be a future PR).


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -1587,101 +1590,93 @@ P0_APDEnergy:
...
     LABL_PTR_1: event_num_label
     VALIDMAX: 128
 
-P0_ERGE:
-    <<: *direct_events
-    CATDESC: Priority 0 Energy Range
-    DEPEND_1: event_num
-    FIELDNAM: Priority 0 Energy Range
-    LABL_PTR_1: event_num_label
-    VALIDMAX: 4
```
bourque, line 1619: The `ERGE` variable was renamed to `gain`, and thus they can be removed

imap_processing/tests/codice/conftest.py
```diff
@@ -29,22 +31,22 @@
 
 # ruff: noqa
 VALIDATION_DATA = [
-    TEST_DATA_PATH / "validation" / "imap_codice_l1a_hi-counters-aggregated_20241110193900_v0.0.2.cdf",
```
bourque, line 32: These changes here are just to list things in a consistent order with what the unit tests expect

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -408,28 +404,30 @@ def test_l1a_validate_support_variables(test_l1a_data, index):
...
         )
 
         # Ensure that the acquisition times are (nearly) equal
-        np.testing.assert_almost_equal(
-            dataset.acquisition_time_per_step.data,
-            validation_dataset.AcquisitionTimePerStep.data,
-            decimal=3,
-        )
+        # TODO: Turn this back on when Joey supplies updated validation data with
+        #       updated acquisition times
```
greglucas, line 412: This is confusing to me. Does this mean that the current validation data is incorrect? I think you'd be able to assert against what you have currently, and if that is updated then the test would fail as expected and we'd need to make some kind of update. i.e. our updates should be in-sync with our tests. I'm not following why we are commenting/uncommenting tests instead.

greglucas (commented):
When are we expecting another validation data delivery? Do we need to wait on that or can we keep progressing with what we have. I think we need to avoid the situation where we are always waiting for the next best thing and keep moving forward here with what we have.

greglucas (approved):

bourque:
> When are we expecting another validation data delivery? Do we need to wait on that or can we keep progressing with what we have. I think we need to avoid the situation where we are always waiting for the next best thing and keep moving forward here with what we have.

I generally agree with you here, but I wanted to get this new set of validation data in because it has been several months of updates (#1906, #1917, #1920, #1928, #1973, #1977, #1979, #2019, #2022) that have gone into this set of validation data and code changes, so it seems like a good time to 'make sure L1A is mostly good before launch'. There may be some more minor tweaks down the road, but I am going to prioritize L1b and L2 development first.

greglucas:
There are KeyError failures still.

https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/17132513565/job/48600211176#step:7:4395

bourque:
I am aware, I wasn't ready for this PR to be reviewed yet 😅

Issue #1023, opened 2024-10-16
SPICE - Calculate instrument spin phase offset using spice


tmplummer:
### Requirements:
Currently the `imap_processing.spice.geometry.get_spacecraft_to_instrument_spin_phase_offset()` function is implemented with a hard coded offset table for each instrument. It would be preferable to implement this function such that it uses spice with the latest imap frame kernel to compute the offset for each instrument. The complicating factor is that each instrument frame is defined differently and thus, the boresight vector will likely need to be used. Some instrument's boresight vector may not give the desire result.

tmplummer:
Due to the added upstream dependency this would require, we chose not to implement this. Instead, we hard code the spin-phase offset.

PR #2089, opened 2025-08-21
2053 spin   update instrument spin phase definition

### LINKED ISSUE #2053: Spin - Update instrument spin phase definition
### Algorithm Description:

Instrument spin phase should be 0 at the point closest to the Ecliptic North Pole.

The change required is to add 90 degrees to each instrument spin phase offset.

tmplummer:
# Change Summary

## Overview
Updates the instrument spin phase 0 to be consistent with IT desires. That is where spin-phase 0 is when the instrument boresight is closest to the ecliptic north pole.

## Updated Files
   - Add 90 degrees and take the modulus with 360 to all instrument spin phase offsets
   - Update expected values by calculating a slightly different way than above.
   - Mock the SWE offset back to the previous value since validation of spin-phase was broken by above changes.

Closes: #2053 #1023

### DIFF
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -137,7 +137,9 @@ def get_spacecraft_to_instrument_spin_phase_offset(instrument: SpiceFrame) -> fl
     Get the spin phase offset from the spacecraft to the instrument.
 
     For now, the offset is a fixed lookup based on `Table 1: Nominal Instrument
-    to S/C CS Transformations` in document `7516-0011_drw.pdf`. These fixed
+    to S/C CS Transformations` in document `7516-0011_drw.pdf`. That Table
+    defines the angle from the spacecraft y-axis. We add 90 and take the modulous
+    with 360 in order to get the angle from the spacecraft x-axis. These fixed
     values will need to be updated based on calibration data or retrieved using
     SPICE and the latest IMAP frame kernel.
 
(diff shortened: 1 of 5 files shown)


imap_processing/spice/geometry.py
```diff
@@ -137,7 +137,9 @@ def get_spacecraft_to_instrument_spin_phase_offset(instrument: SpiceFrame) -> fl
     Get the spin phase offset from the spacecraft to the instrument.
 
     For now, the offset is a fixed lookup based on `Table 1: Nominal Instrument
-    to S/C CS Transformations` in document `7516-0011_drw.pdf`. These fixed
+    to S/C CS Transformations` in document `7516-0011_drw.pdf`. That Table
+    defines the angle from the spacecraft y-axis. We add 90 and take the modulous
+    with 360 in order to get the angle from the spacecraft x-axis. These fixed
     values will need to be updated based on calibration data or retrieved using
     SPICE and the latest IMAP frame kernel.
```
greglucas: Do you want to implement the retrieval from the SPICE instrument kernel now that we have official alignments there?

imap_processing/tests/swe/test_swe_l2.py
```diff
@@ -256,8 +256,13 @@ def test_put_data_into_angle_bins():
 
 
 @patch("imap_data_access.processing_input.ProcessingInputCollection.get_file_paths")
+@patch(
+    "imap_processing.spice.spin.get_spacecraft_to_instrument_spin_phase_offset",
+    return_value=153 / 360,
```
greglucas, line 261: This is a nice way to approach this for now. I think we might want to open an issue for @tech3371 to update the expected values at some point though...

greglucas (approved):

greglucas:
I also assume you know it is supposed to be +90 and not -90? I wasn't positive which direction of rotation you needed to rotate.

imap_processing/tests/swe/test_swe_l2.py
tech3371: Sounds good. I will work with Ruth on this after launch preparation.

tech3371 (approved):
The changes looks good to me. Thank you for having workaround for SWE.

imap_processing/spice/geometry.py
tmplummer: I went ahead and added the calculation using SPICE. @greglucas can you take another look when you get a chance?

imap_processing/tests/spice/test_geometry.py
```diff
@@ -46,27 +47,62 @@ def test_imap_state_ecliptic(imap_ena_sim_metakernel):
     assert state.shape == (6,)
 
 
+@pytest.mark.parametrize(
+    "instrument, expected_az_el",
+    [
+        # Expected spin-phase offsets based on 7516-0011_drw.pdf
+        (SpiceFrame.IMAP_LO_BASE, (60, 0)),  # (330 + 90) % 360 = 60
+        # Note HI_45 and HI_90 appear to be swapped in imap_wkcp.tf so the
+        # expected values are swapped here.
```
greglucas, line 56: This seems worth bringing up to someone...

greglucas (approved):
Looks good to me. The only downside is that if someone needs this functionality they have to add the frame kernel to their SPICE dependency chart, was this already the case or do we need to do an inventory there?

imap_processing/tests/spice/test_geometry.py
tmplummer: I am waiting for the updated imap frames kernel before I make any noise about this.

greglucas: I would message Nick on the side quickly about this so it gets taken care of before we get an "updated kernel" that doesn't have this update you want!

tmplummer:
> Looks good to me. The only downside is that if someone needs this functionality they have to add the frame kernel to their SPICE dependency chart, was this already the case or do we need to do an inventory there?

Agreed, this is a slight downside. I suspect all the processing getting spin-phase already required this kernel, but will need to let the team know about this change on Monday.

PR #2090, opened 2025-08-21
I-ALiRT SWAPI Validation

torimarbois:
# Change Summary

## Overview
Revamp SWAPI validation by running three new test datasets through the code, and make sure the results are within a tolerance specified by the SWAPI team.
Add logic to process the SWAPI esa-unit-conversion lookup table as an input calibration file to the processing function.

## New Files

- ialirt_test_data_u_sw....*
   - These are the 3 new test datasets
- swapi_ialirt_energy_steps.csv
   - This is a reduced version of the SWAPI lookup table, and it has realistic energy step values used for testing

## Updated Files

- constants.py
   - Replaced the Effective Area constant with a new value from the SWAPI team
   - Added new constants used to create the initial parameter guess for the optimization model
- process_swapi.py
   - Add input calibration file and the logic needed to extract energy steps
   - Add more robust initial parameter guess generation

## Testing

- test_process_swapi.py
   - Modify existing tests to check for new values which were impacted due to the change of the Effective Area constant
   - Add looping logic to test the 3 new datasets
   - Add calibration file parameter


imap_processing/ialirt/constants.py
```diff
@@ -33,10 +33,12 @@ class IalirtSwapiConstants:
     boltz = 1.380649e-23  # Boltzmann constant, J/K
     at_mass = 1.6605390666e-27  # atomic mass, kg
     prot_mass = 1.007276466621 * at_mass  # mass of proton, kg
-    eff_area = 3.3e-5 * 1e-4  # effective area, meters squared
+    eff_area = 1.633e-4 * 1e-4  # effective area, meters squared
```
greglucas: Why did this change so significantly?

Is there a reason to write it as Xe-4 * 1e-4 rather than Xe-8? Could you add a comment about cm2 -> m2 or similar if that is the reason.

imap_processing/tests/ialirt/unit/test_process_swapi.py
```diff
@@ -168,34 +172,74 @@ def test_count_rate():
...
+        count_rates = np.tile(count_rates, (2, 1))
+        count_rates_errors = energy_data["Count Rates Error [Hz]"].to_numpy()
+        count_rates_errors = np.tile(count_rates_errors, (2, 1))
+
+        result = optimize_pseudo_parameters(
+            count_rates, count_rates_errors, energy_passbands
+        )
+
+        for param in test_data[test_set]["expected_values"]:
+            assert np.allclose(
```
greglucas: Use the numpy testing assertion function so it does a pretty print for you on failure.
```suggestion
            np.testing.assert_allclose(
```

greglucas (approved):
minor nits is all, looks good generically 👍

imap_processing/ialirt/constants.py
bishwassth: @greglucas The previous value is from the heritage instrument (New Horizons SWAP), and the new one is for the SWAPI.

torimarbois: I don't know why it changed so significantly - it was a change given to me by the SWAPI team. 

The reason is because it's a conversion from cm2 to m2, so I will add a comment about that. Thanks!

bishwassth (approved):
Looks good.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -137,14 +129,18 @@ def optimize_pseudo_parameters(
...
+) -> list[dict]:
     """
     Extract I-ALiRT variables and calculate coincidence count rate.
 
     Parameters
     ----------
     unpacked_data : xr.Dataset
         SWAPI I-ALiRT data that has been parsed from the spacecraft packet.
+    calibration_lut_table : pd.DataFrame
+        DataFrame containing the contents of the SWAPI lookup table.
```
tech3371: I would mention something about which calibration file is used here since we have couple LUT. 
```suggestion
        DataFrame containing the contents of the SWAPI esa-unit-conversion lookup table.
```

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -193,7 +189,28 @@ def process_swapi_ialirt(unpacked_data: xr.Dataset) -> list[dict]:
...
+    # Find the sweep's energy data for the latest time, where sweep_id == 2
+    subset = calibration_lut_table[
+        (calibration_lut_table["timestamp"] == calibration_lut_table["timestamp"].max())
+        & (calibration_lut_table["Sweep #"] == 2)
+    ]
+    if subset.empty:
+        energy_passbands = np.full(63, np.nan, dtype=np.float64)
+    else:
+        subset = subset.sort_values(["timestamp", "ESA Step #"])
+        energy_passbands = subset["Energy"][0:63].to_numpy().astype(float)
```
tech3371, line 209: Logically, it makes sense to me. Can you add a test to check that it returns the expected 63 values since we know what to expect by manually looking in LUT. Eg. test that the subset matches the expected energy values. You can do that by refactoring energy look up code into it's function and test that way.

tech3371 (commented):
one minor request otherwise, looks good to me

imap_processing/ialirt/l0/process_swapi.py
torimarbois: Just logging our in-person conversation here: Tenzin wanted to make sure that the correct sweep & energy steps were being extracted form the lookup table. To accomplish that, I added another sweep to the test energy steps file (with a set of zeros corresponding to sweep # 1) and made sure that all the tests still passed, indicating that the right energy values were being extracted. 👍

tech3371 (approved):

PR #2091, opened 2025-08-21
MNT: Update Lo L1c pointing set dimensions and attribute naming

greglucas:
Based upon the pset variables named here: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/281316700/Remaining+PSET+Work

Updates all of the dimensions to be (epoch, esa_energy_step, spin_angle, off_angle). One question I still have is whether we want to fill in the actual spin_angle values instead of an integer bin number. This is what I would want, but I am unclear what the standard is for other instruments and the files provided had integer bin numbers (even though the variables don't have the `_bin` suffix). Thoughts on which approach to use?

### DIFF
--- a/imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
@@ -113,6 +113,10 @@ hae_longitude:
   FORMAT: I12
   UNITS: degrees
   LABLAXIS: Bin longitudes
+  DEPEND_1: spin_angle
+  DEPEND_2: off_angle
+  LABL_PTR_1: spin_angle_label
+  LABL_PTR_2: off_angle_label
 
 hae_latitude:
   <<: *default
@@ -123,6 +127,10 @@ hae_latitude:
   FORMAT: I12
   UNITS: degrees
   LABLAXIS: Bin latitudes
+  DEPEND_1: spin_angle
+  DEPEND_2: off_angle
+  LABL_PTR_1: spin_angle_label
+  LABL_PTR_2: off_angle_label
 
 exposure_time:
   <<: *default
--- a/imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
@@ -56,7 +56,7 @@ off_angle:
   UNITS: "degrees"
   FORMAT: I4
   VAR_TYPE: support_data
-  LABLAXIS: off axis angle
+  LABLAXIS: off angle
 
 pointing_start_met:
   <<: *default
@@ -64,7 +64,7 @@ pointing_start_met:
   FIELDNAM: Pointing start time
   FORMAT: I12
   UNITS: s
-  LABLAXIS: pointing start time
+  LABLAXIS: pointing start
 
 pointing_end_met:
   <<: *default
@@ -72,21 +72,21 @@ pointing_end_met:
   FIELDNAM: Pointing end time
   FORMAT: I12
   UNITS: s
-  LABLAXIS: pointing end time
+  LABLAXIS: pointing end
 
 start_spin_number:
   <<: *default
   CATDESC: The spin number at the start of the pointing
   FIELDNAM: Spin number at start of pointing
   FORMAT: I12
-  LABLAXIS: First spin number of pointing
+  LABLAXIS: Start spin
 
 end_spin_number:
   <<: *default
   CATDESC: The spin number at the start of the pointing
   FIELDNAM: Spin number at the end of the pointing
   FORMAT: I12
-  LABLAXIS: Last spin number of pointing
+  LABLAXIS: End spin
 
 pivot_angle:
   <<: *default
@@ -99,11 +99,11 @@ pivot_angle:
 
 esa_mode:
   <<: *default
-  DEPEND_0: esa_energy_step
+  DEPEND_1: esa_energy_step
   CATDESC: Science Mode for Pointing
   FIELDNAM: Science Mode
   FORMAT: I12
-  LABLAXIS: science mode
+  LABLAXIS: esa mode
 
 hae_longitude:
   <<: *default
@@ -112,7 +112,7 @@ hae_longitude:
   FIELDNAM: Bin longitudes
   FORMAT: I12
   UNITS: degrees
-  LABLAXIS: Bin longitudes
+  LABLAXIS: longitude
   DEPEND_1: spin_angle
   DEPEND_2: off_angle
   LABL_PTR_1: spin_angle_label
@@ -126,7 +126,7 @@ hae_latitude:
   FIELDNAM: Bin latitudes
   FORMAT: I12
   UNITS: degrees
-  LABLAXIS: Bin latitudes
+  LABLAXIS: latitude
   DEPEND_1: spin_angle
   DEPEND_2: off_angle
   LABL_PTR_1: spin_angle_label
@@ -164,7 +164,7 @@ doubles_counts:
   DEPEND_3: off_angle
   FIELDNAM: Double coincidence counts
   FORMAT: I12
-  LABLAXIS: double coincidence counts
+  LABLAXIS: doubles
   LABL_PTR_1: esa_energy_step_label
   LABL_PTR_2: spin_angle_label
   LABL_PTR_3: off_angle_label
@@ -192,6 +192,8 @@ h_background_rates:
   LABL_PTR_2: spin_angle_label
   LABL_PTR_3: off_angle_label
   FORMAT: I12
+  DELTA_PLUS_VAR: h_background_rates_stat_uncert
+  DELTA_MINUS_VAR: h_background_rates_stat_uncert
 
 h_background_rates_stat_uncert:
   <<: *default
@@ -240,6 +242,8 @@ o_background_rates:
   LABL_PTR_2: spin_angle_label
   LABL_PTR_3: off_angle_label
   FORMAT: I12
+  DELTA_PLUS_VAR: o_background_rates_stat_uncert
+  DELTA_MINUS_VAR: o_background_rates_stat_uncert
 
 o_background_rates_stat_uncert:
   <<: *default
--- a/imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
@@ -7,7 +7,7 @@ default_attrs: &default
   DEPEND_0: epoch
   VALIDMAX: 9223372036854775807
   VAR_TYPE: data
-  UNITS: ' '
+  UNITS: " "
 
 # Non-epoch Coordinates
 esa_energy_step_label:
@@ -21,7 +21,7 @@ esa_energy_step:
   VALIDMAX: 7
   CATDESC: Energy Step
   FIELDNAM: Energy Step
-  UNITS: ' '
+  UNITS: " "
   FORMAT: I1
   VAR_TYPE: support_data
   LABLAXIS: ESA
@@ -34,13 +34,13 @@ spin_angle_label:
 
 spin_angle:
   VALIDMIN: 0
-  VALIDMAX: 3599
-  CATDESC: Spin angle bins
-  FIELDNAM: Spin angle bins
-  UNITS: ' '
+  VALIDMAX: 360
+  CATDESC: Spin angle in despun pointing frame
+  FIELDNAM: Spin angle
+  UNITS: "degrees"
   FORMAT: I4
   VAR_TYPE: support_data
-  LABLAXIS: spin angle bins
+  LABLAXIS: spin angle
 
 off_angle_label:
   CATDESC: Off axis angles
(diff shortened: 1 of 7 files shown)


vmartinez-cu:
HIT uses actual angle values for it's zenith and azimuth angle coordinates. I'm not sure what other instruments are doing. If the angle values are known, it seems worth adding in

greglucas:
> HIT uses actual angle values for it's zenith and azimuth angle coordinates. I'm not sure what other instruments are doing. If the angle values are known, it seems worth adding in

Updated!

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
```diff
@@ -80,100 +97,170 @@ pivot_angle:
   UNITS: degrees
   LABLAXIS: pivot angle
 
+esa_mode:
+  <<: *default
+  DEPEND_0: esa_energy_step
```
vmartinez-cu: Is `esa_energy_step` a time varying variable? I believe only epoch or another time variable can be assigned to DEPEND_0. You might have to use DEPEND_1 here.  [ISTP guidance](https://spdf.gsfc.nasa.gov/istp_guide/vattributes.html#DEPEND_0:~:text=DEPEND_0%20%2D%2D%2D%20required%20for%20time%2Dvarying%20variables)

imap_processing/ena_maps/ena_maps.py
```diff
@@ -634,15 +634,10 @@ class LoPointingSet(PointingSet):
...
-        longitude_bin_centers = np.arange(
-            0 + inferred_spacing_deg / 2, 360, inferred_spacing_deg
-        )
-        latitude_bin_centers = np.arange(
-            -2 + inferred_spacing_deg / 2, 2, inferred_spacing_deg
-        )
+        # 0.1 degree bin spacing in both spin angle (longitude) and off angle (latitude)
+        bin_spacing_deg = 0.1
+        longitude_bin_centers = np.arange(0 + bin_spacing_deg / 2, 360, bin_spacing_deg)
+        latitude_bin_centers = np.arange(-2 + bin_spacing_deg / 2, 2, bin_spacing_deg)
```
tmplummer: Can't these be grabbed from the PSET Dataset?
```suggestion
        longitude_bin_centers = self.data["spin_angle"].data
        latitude_bin_centers = self.data["off_angle"].data
```

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
```diff
@@ -80,100 +97,170 @@ pivot_angle:
...
-  CATDESC: "Counts for double coincidence events"
-  DEPEND_1: pointing_bins
-  DEPEND_2: esa_step
+  CATDESC: Counts for double coincidence events
+  DEPEND_1: esa_energy_step
+  DEPEND_2: spin_angle
+  DEPEND_3: off_angle
   FIELDNAM: Double coincidence counts
   FORMAT: I12
   LABLAXIS: double coincidence counts
```
vmartinez-cu: [LABLAXIS](https://spdf.gsfc.nasa.gov/istp_guide/vattributes.html#LABLAXIS:~:text=LABLAXIS%20%2D%2D%2D%20required%20if%20not%20using%20LABL_PTR_1) should be short character strings (6-10 characters), though I'm not sure what would happen if it's longer. It's hard to tell from the ISTP guidance what's a recommendation vs. a hard requirement.

vmartinez-cu: HIT has some long values too, so maybe it's fine since Andriy didn't comment on it

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -68,15 +68,12 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
...
-    #  CDF won't open if DEPEND_0 is not deleted currently.
-    del pset["epoch"].attrs["DEPEND_0"]
 
     pset = pset.assign_coords(
         {
-            "energy": np.arange(1, 8),
-            "longitude": np.arange(3600),
-            "latitude": np.arange(40),
+            "esa_energy_step": np.arange(1, 8),
+            "spin_angle": np.arange(3600) / 3600 + 0.05,
```
tmplummer: These don't look right. I prefer the way you implemented it above...
```suggestion
            "spin_angle": np.arange(0 + bin_spacing_deg / 2, 360, bin_spacing_deg),
            "off_angle": np.arange(-2 + bin_spacing_deg / 2, 2, bin_spacing_deg),
```

Or from your suggestion on Hi code, should you be using np.linspace. I forget the exact reason that you made that suggestion.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -338,38 +333,54 @@ def create_datasets(
...
             attrs=attr_mgr.get_variable_attributes("esa_step_label"),
         )
-        pointing_bins_label = xr.DataArray(
-            pointing_bins.values.astype(str),
-            name="pointing_bins_label",
-            dims=["pointing_bins_label"],
-            attrs=attr_mgr.get_variable_attributes("pointing_bins_label"),
+
+        spin_angle = xr.DataArray(
+            data=np.arange(3600) / 3600 + 0.05,
```
tmplummer: Here is another instance where this array is being generated. Perhaps put this in a module level global variable. The implementation here is not correct.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -338,38 +333,54 @@ def create_datasets(
...
+        )
+        spin_angle_label = xr.DataArray(
+            spin_angle.values.astype(str),
+            name="spin_angle_label",
+            dims=["spin_angle_label"],
+            attrs=attr_mgr.get_variable_attributes("spin_angle_label"),
         )
+
+        off_angle = xr.DataArray(
+            data=np.arange(40) / 40 - 1.95,
```
tmplummer: Likewise, put this into a module level global and fix implementation?

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
```diff
@@ -80,100 +97,170 @@ pivot_angle:
...
   FORMAT: I12
   FIELDNAM: Oxygen count rates
-  LABLAXIS: O count rates
-  LABL_PTR_1: pointing_bins_label
-  LABL_PTR_2: esa_step_label
+  LABL_PTR_1: esa_energy_step_label
+  LABL_PTR_2: spin_angle_label
+  LABL_PTR_3: off_angle_label
+
+o_background_rates:
```
vmartinez-cu, line 234: You can add DELTA_MINUS_VAR and DELTA_PLUS_VAR attribute fields and assign the appropriate uncertainty variables to these. This is optional according to [ISTP guidance](https://spdf.gsfc.nasa.gov/istp_guide/vattributes.html#DELTA:~:text=DELTA_PLUS_VAR%20and%20DELTA_MINUS_VAR%20%2D%2D%2D%20optional). I added these for HIT using total uncertainty values.

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
```diff
@@ -7,69 +7,86 @@ default_attrs: &default
...
+spin_angle:
   VALIDMIN: 0
-  VALIDMAX: 3599
-  FILLVAL: -9223372036854775808
-  CATDESC: Pointing bins
-  DEPEND_1: pointing_bins
-  FIELDNAM: Pointing Bins
-  UNITS: ' '
+  VALIDMAX: 360
+  CATDESC: Spin angle in despun pointing frame
```
tmplummer: Might want to specify that these are bin centers.

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
```diff
@@ -7,69 +7,86 @@ default_attrs: &default
...
+off_angle_label:
+  CATDESC: Off axis angles
+  FIELDNAM: Off axis angles
+  FORMAT: A4
+  VAR_TYPE: metadata
+
+off_angle:
+  VALIDMIN: -2
+  VALIDMAX: 2
+  CATDESC: Off axis angle in despun pointing frame
```
tmplummer: Specify that these are bin centers?

tmplummer (commented):
I'd mostly like to see a common definition of the angular bin centers defined. For Hi I did it here: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/hi/hi_l1c.py#L38

tmplummer:
@greglucas 
> Updates all of the dimensions to be (epoch, esa_energy_step, spin_angle, off_angle). One question I still have is whether we want to fill in the actual spin_angle values instead of an integer bin number. This is what I would want, but I am unclear what the standard is for other instruments and the files provided had integer bin numbers (even though the variables don't have the _bin suffix). Thoughts on which approach to use?

I would also want the angle bins. For Lo where the off_angle dimension does not cover the full latitude range, I would argue it is critical to specify angle.

I'm sure Paul will be amenable to me changing Hi to specify angle instead of bin number in the PSET product.

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
vmartinez-cu: I forgot that this is L1, so it might not be worth the effort adding these in, but would be useful to include in L2

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
sdhoyt: It's not time varying and you're right, this should be DEPEND_1, not DEPEND_0

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
sdhoyt: @vmartinez-cu They want this product to be archived, so do you think given that, the DELTA_VAR attributes should be added?

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
greglucas: That is quite confusing to me even after reading it a few times. I think I agree with you, but that would mean that this has a DEPEND_1 without a DEPEND_0? 🤷

greglucas: Done

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
greglucas: Done

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
greglucas: I added these to point at the uncert variable for both plus and minus per the recommendation.

imap_processing/ena_maps/ena_maps.py
greglucas: I reworked this to use the hae_longitude/latitude variables now instead, grabbing those directly from the PSET. (brought in a previous commit I had into this PR)

imap_processing/lo/l1c/lo_l1c.py
greglucas: linspace is supposed to be more stable for floating point numbers according to the docs. I have upated this with globals to correspond with the Hi version.

imap_processing/lo/l1c/lo_l1c.py
greglucas: Yep, moved these up as globals, good suggestion!

imap_processing/cdf/config/imap_lo_l1c_variable_attrs.yaml
greglucas: Done

greglucas:
Thanks for the great reviews! I think I've addressed all comments now. I reworked the bin angles/centers code quite a bit, so would be good to review that and see if folks agree.

greglucas:
@sdhoyt It looks like codecov is failing because the `create_datasets` function is not called at all. Do you anticipate needing that or should I just remove it from here?

tmplummer (approved):
LGTM

sdhoyt:
> @sdhoyt It looks like codecov is failing because the `create_datasets` function is not called at all. Do you anticipate needing that or should I just remove it from here?

@greglucas That will be needed to set all the attributes or some of the code from that function will need to be pulled out and integrated into the other functions in the file. I don't think I've touched that function since I created an empty file for SIT-3, so it's out of date. I was leaving it there to use later, but the attributes weren't my priority so I stopped calling it rather than updating it as I made algorithm changes. Since the version history is there, it's probably fine to remove for now.

greglucas:
> @greglucas That will be needed to set all the attributes or some of the code from that function will need to be pulled out and integrated into the other functions in the file. I don't think I've touched that function since I created an empty file for SIT-3, so it's out of date. I was leaving it there to use later, but the attributes weren't my priority so I stopped calling it rather than updating it as I made algorithm changes. Since the version history is there, it's probably fine to remove for now.

OK, I'm fine merging over the codecov failures if you'll be using it in the future so I can just leave it for now would be my preference in that case.

Are you OK with me merging this now or do you want me to wait until yours is in so you don't have to deal with any conflicts? It looks like this is pretty separate from yours though, so merge conflicts wouldn't be too bad hopefully.

PR #2093, opened 2025-08-21
IDEX l2a calibration curve implementation 

lacoak21:
# Change Summary

## Overview
This was added as a change request:

"Calibration curve implementation: IDEX has obtained more calibration data over the last 2 months that we want to incorporate in our mass estimates. This is a 6th-order polynomial (p(x) = a6*x^6 + a5*x^5 + a4*x^4 + a3*x^3 + a2*x^2 + a1*x + a0) that will be used to estimate the mass from the impact charge and velocity."

Ethan has been working on this code for a while and supplied me with the power law functions and estimation code which I have used almost verbatim ( aside from organization and variable naming). 

## Updated Files
- imap_processing/cdf/config/imap_idex_l2a_variable_attrs.yaml
   - Add new attributes for the velocity estimates
- imap_processing/idex/idex_l2a.py
   - Calibration curve implementation

## Testing
Add a minimal test. I didnt test this that much due to time constraints. Please let me know if this needs more testing and I will be happy to add some.

### DIFF
--- a/imap_processing/idex/idex_l2a.py
+++ b/imap_processing/idex/idex_l2a.py
@@ -361,7 +361,7 @@ def log_smooth_powerlaw(log_v: float, log_a: float, params: np.ndarray) -> float
         Velocity.
     log_a : float
         Scale factor.
-    params : np.array
+    params : np.ndarray
         Calibration parameters for the power law.
 
     Returns
--- a/imap_processing/idex/idex_l2a.py
+++ b/imap_processing/idex/idex_l2a.py
@@ -342,7 +342,12 @@ def calculate_velocity_and_mass(
         )
         v_est = 10**root.root
     except Exception:
-        v_est = 0.0
+        logger.error(
+            "Unable to calculate velocity and mass estimate. "
+            "The root finding failed for power law function. "
+            "Returning nans for the estimate."
+        )
+        return np.nan, np.nan
 
     log_a_y: float = np.log10(yield_params[0])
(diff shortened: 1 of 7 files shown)


imap_processing/idex/idex_l2a.py
```diff
@@ -261,6 +304,84 @@ def idex_l2a(l1b_dataset: xr.Dataset) -> xr.Dataset:
...
+            lambda lv: log_smooth_powerlaw(lv, log_a_t, t_rise_params[1:])
+            - np.log10(t_rise),
+            bracket=[-1, 2],
+        )
+        v_est = 10**root.root
+    except Exception:
+        v_est = 0.0
+
+    log_a_y: float = np.log10(yield_params[0])
+    yield_val = 10 ** log_smooth_powerlaw(np.log10(v_est), log_a_y, yield_params[1:])
```
greglucas, line 353: np.log10(0) will give you a runtime warning and `-inf`, do you want to return in the exception above instead? something like `return 0.0, 0.0` or nan's or similar?

imap_processing/tests/idex/test_idex_l2a.py
```diff
@@ -267,6 +275,20 @@ def test_analyze_peaks_warning(caplog):
     np.testing.assert_array_equal(area_under_curve, np.zeros(area_under_curve.shape))
 
 
+def test_velocity_and_mass_estimate(ancillary_files):
```
greglucas, line 279: Needs to be marked external I think? Or perhaps you can mock the get_ancillary files to return some artificial arrays for you instead of needing the external file.

greglucas (approved):

PR #2096, opened 2025-08-22
Remove uses of spiceypy.kclear() and spiceypy.furnsh() from tests

### LINKED ISSUE #2095: BUG - Unpredictable test failures
### Description of the issue

From slack thread it seems tests fail unpredictably on the Ubuntu 3.10 job.

Suspicion is that the cause is the use of spiceypy.kclear() directly in tests. Tests should only use `furnish_kernels` fixture. If the desire is to have no kernels furnished, it can be used with an empty list to furnish.

Tim was able to reproduce by running the full ultra test suite locally.

_No response_

_No response_

### Code Snippet:

```python

```

_No response_

tmplummer:
# Change Summary

## Overview
I believe that the unpredictable test failures that were/are occurring on the Ubuntu 3.10 job are caused by the direct use of `spiceypy.kclear()` in tests. I removed those and also cleaned up a bunch of direct usages of `spiceypy.furnsh()`. In order to prevent tests from having SPICE related side effects, we should only be using the `furnish_kenrels` fixture. If a test requires no kernels be furnished, one can pass `furnish_kernels` an empty list.

## Updated Files
   - The biggest offender... remove `spiceypy.kclear`
   - Use `furnish_kernels` rather than `spiceypy.furnsh`
- imap_processing/tests/ultra/unit/test_ultra_l1b_annotated.py
   - Use `furnish_kernels` rather than `spiceypy.furnsh`
- imap_processing/tests/ultra/unit/test_ultra_l1c_culling.py
   - remove `spiceypy.kclear`

Closes: #2095

### DIFF
--- a/imap_processing/tests/spice/test_pointing_frame.py
+++ b/imap_processing/tests/spice/test_pointing_frame.py
@@ -22,7 +22,7 @@ from imap_processing.spice.time import TICK_DURATION
 
 
 @pytest.fixture
-def pointing_frame_kernels(furnish_kernels, spice_test_data_path):
+def furnish_pointing_frame_kernels(furnish_kernels, spice_test_data_path):
     """List SPICE kernels."""
     required_kernels = [
         "naif0012.tls",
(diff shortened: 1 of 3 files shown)


laspsandoval (approved):
Thanks for catching it Tim.

imap_processing/tests/spice/test_pointing_frame.py
```diff
@@ -36,23 +36,24 @@ def pointing_frame_kernels(spice_test_data_path):
 
 
 @pytest.fixture
-def et_times(pointing_frame_kernels):
+def et_times(furnish_kernels, pointing_frame_kernels):
```
greglucas: Would it be desirable to put these with blocks into the pointing_frame_kernels fixture itself?

In these fixtures, should you be yielding instead of returning so that the context is kept?

greglucas (approved):
Nice job tracking this down for sure!

PR #2097, opened 2025-08-22
Ultra l1c pset - Create a flag for pixels that "see" the Earth

laspsandoval:
# Change Summary

## Updated Files
- ultra_l1b_extended.py
   - Convert to et (instead of keeping it in SCT)
- quality_flags.py
   - created l1c pset quality flag for earth being in the fov
- ultra_l1c_culling.py
   - pass quality flag in that is updated for pixels in the fov
- spacecraft_pset.py
   -  add quality flags to the data product
- helio_pset.py
   -  add quality flags to the data product

## Testing
test_ultra_l1b_extended.py
test_ultra_l1c_culling.py

PR #2104, opened 2025-08-26
ULTRA L1C sensitivity calculation for helio and spacecraft psets

lacoak21:
# Change Summary

## Overview
This PR updates the spacecraft pset and helio pset to calculate sensitivity, efficiency, and geometric factor arrays as function of energy. This is required because maps require pixels over a certain uncertainty to be culled. Scattering culling has to be calculated and that is a function of spin phase. At each energy bin, there are different scattering thresholds so this means we have to calculate exposure pointing,  sensitivity, efficiency, and geometric factors in regards to the energy bin now. 

Helio psets also need to adjust the values above for the energy differences between the spacecraft and heliocentric frame. Laura already had code to do this thankfully. 

## New Files
   - move l1c lookups from l1b lookup utils file. 

## Updated Files
   - Update helio pset to calculate the eff, gf, exposure time, and sensitivity arrays as functions of energy. 
   - Adjust for heliocentric frame
   - fix scattering culling. There was a bug in the old way. I had made assumptions about what the lookup tables would look like. 

## Testing
Test lookup tables for l1c.

### DIFF
--- a/imap_processing/ultra/l1c/helio_pset.py
+++ b/imap_processing/ultra/l1c/helio_pset.py
@@ -6,6 +6,7 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 
+from imap_processing.spice.time import sct_to_et
 from imap_processing.ultra.l1c.l1c_lookup_utils import (
     calculate_pixels_within_scattering_threshold,
     get_spacecraft_pointing_lookup_tables,
@@ -105,8 +106,8 @@ def calculate_helio_pset(
     efficiencies, geometric_function = get_efficiencies_and_geometric_function(
         pixels_below_scattering, theta_vals, phi_vals, n_pix, ancillary_files
     )
-
-    mid_time = np.mean(de_dataset["event_times"].data)
+    # TODO fix midpoint time calculation
+    mid_time = sct_to_et(np.median(de_dataset["event_times"].data))
     exposure_time, efficiency, geometric_function = get_helio_adjusted_data(
         mid_time,
         exposure_time,
--- a/imap_processing/ultra/l1c/helio_pset.py
+++ b/imap_processing/ultra/l1c/helio_pset.py
@@ -8,10 +8,8 @@ import xarray as xr
 
 from imap_processing.spice.time import sct_to_et
 from imap_processing.ultra.l1c.l1c_lookup_utils import (
-    get_spacecraft_pointing_lookup_tables,
-)
-from imap_processing.ultra.l1c.spacecraft_pset import (
     calculate_pixels_within_scattering_threshold,
+    get_spacecraft_pointing_lookup_tables,
 )
 from imap_processing.ultra.l1c.ultra_l1c_pset_bins import (
     build_energy_bins,
--- a/imap_processing/ultra/l1c/helio_pset.py
+++ b/imap_processing/ultra/l1c/helio_pset.py
@@ -6,7 +6,13 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 
-from imap_processing.spice.time import sct_to_et
+from imap_processing.spice.repoint import get_pointing_times
+from imap_processing.spice.time import (
+    et_to_met,
+    met_to_ttj2000ns,
+    sct_to_et,
+    ttj2000ns_to_et,
+)
 from imap_processing.ultra.l1c.l1c_lookup_utils import (
     calculate_pixels_within_scattering_threshold,
     get_spacecraft_pointing_lookup_tables,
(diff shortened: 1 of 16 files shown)


imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -543,18 +560,18 @@ def get_helio_exposure_times(
...
+        A 2D array of shape (n_energy_bins, npix).
+    helio_geometric_factors : np.ndarray
+        A 2D array of shape (n_energy_bins, npix).
 
     Notes
     -----
     These calculations are performed once per pointing.
     """
     # Get energy midpoints.
     _, energy_midpoints, _ = build_energy_bins()
```
lacoak21: I have been using _, _, energy_geometric_means = build_energy_bins() instead of the midpoints. Should I switch to using these?

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -471,70 +476,82 @@ def get_efficiencies_and_geometric_function(
     """
     # Load callable efficiency interpolator function
     eff_interpolator = get_efficiency_interpolator(ancillary_files)
+    # load geometric factor lookup table
+    geometric_lookup_table = load_geometric_factor_tables(
+        ancillary_files, "l1b-sensor-gf-blades"
+    )
     # Get energy bin geometric means
     energy_bin_geometric_means = build_energy_bins()[2]
```
lacoak21, line 484: Here is where I use the energy bin means.

imap_processing/ultra/l1c/l1c_lookup_utils.py
```diff
@@ -0,0 +1,201 @@
+"""Contains tools for lookup tables for l1c."""
```
lacoak21, line 1: This looks like a lot of changes but most of these are just moving lookup utils from the l1b utils file and from the spacecraft pset file.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
laspsandoval: No. You are doing it correctly. I need to go in and change this.

lacoak21: I can also update this right now.

imap_processing/ultra/l1c/helio_pset.py
```diff
@@ -60,35 +78,61 @@ def calculate_helio_pset(
...
-    exposure_pointing = get_helio_exposure_times(mid_time, df_exposure)
+    exposure_time, deadtime_ratios = get_spacecraft_exposure_times(
+        df_exposure, rates_dataset, params_dataset, pixels_below_scattering
+    )
+    # calculate efficiency and geometric function as a function of energy
+    efficiencies, geometric_function = get_efficiencies_and_geometric_function(
+        pixels_below_scattering, theta_vals, phi_vals, n_pix, ancillary_files
+    )
+
+    mid_time = sct_to_et(np.median(de_dataset["event_times"].data))
```
laspsandoval: Why was the median event times used here? Also, the event times should be in et already. I am making that change in this PR:
https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2097

It's in draft form but you can see what I've done.

imap_processing/ultra/l1c/l1c_lookup_utils.py
```diff
@@ -0,0 +1,201 @@
...
+    phi_vals : np.ndarray
+         A 2D array of phi values for each HEALPix pixel at each spin phase step.
+    ancillary_files : dict
+        Dictionary containing ancillary files.
+    instrument_id : int,
+        Instrument ID, either 45 or 90.
+
+    Returns
+    -------
+    exposure_pointing_adjusted : list
```
laspsandoval: Change this part of the docstring.

laspsandoval (approved):
It looks good except for the one comment about the time (which is a mistake I just corrected in the other draft PR). Thanks Luisa. Nice work!

imap_processing/ultra/l1c/helio_pset.py
lacoak21: Ah got it. Im not sure it was already using the median time. I just went back to Nicks instructions and he said to " Grab the mid-point in time for a Pointing".

greglucas: You should use this new function that Sean just added yesterday! Pass in an event time to get back the start and end and then you can do the midpoint of that to get the middle of the pointing.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dc85ad38cb7439d96b2962dfb58f4e4d05f9dc3f/imap_processing/spice/repoint.py#L198

lacoak21: Oh nice! Thats perfect.

lacoak21: Im getting caught up in the time conversions. Would it be something like this?

```
    # TODO remove sct_to_et conversion
    pointing_start, pointing_stop = get_pointing_times(et_to_met(sct_to_et(de_dataset["event_times"].data[0])))
    mid_time =ttj2000ns_to_et( met_to_ttj2000ns((pointing_start + pointing_stop) / 2)) 
```



I need the mid_time to be in et

greglucas: I think that is right.

PR #2106, opened 2025-08-26
Ultra l1c load scattering thresholds from lookup table 

lacoak21:
# Change Summary

## Overview
Use lookup table instead of a constant variable to load energy ranges and their corresponding thresholds. 

-imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
   - Add attributes that I forgot to add in the last PR
   - lookup table function 

## Testing
Add a test for the expected lookup table. This will need to be updated eventually.


imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
```diff
@@ -17,6 +17,11 @@ default_float32_attrs: &default_float32
   VALIDMAX: 3.4028235e+38
   dtype: float32
 
+energy_dependent_attrs: &energy_dependent
+  <<: *default_float32
+  DEPEND_0: energy_bin_geometric_mean
```
greglucas: I have been told my @vmartinez-cu that DEPEND_0 must be epoch / time varying. So this should be DEPEND_1 instead I think.

greglucas (approved):

imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
lacoak21: Yeah I was wondering that. they specifically requested the dimensions to be energy by pixel index. Laura do you they will care if we go back to epoch, energy, pixel_index?

greglucas: I don't think you go back to the other version, you just label these DEPEND_1, DEPEND_2 instead so there is no DEPEND_0? That is what I did at least...

https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2091#discussion_r2294134720

lacoak21: Ah got it.

PR #2107, opened 2025-08-26
Ultra l1c - adding species filtering prior to pset calculations

laspsandoval:
# Change Summary

## Overview
We need to use only hydrogen when creating the psets. This simply filters out the other species prior to creating the psets.

## Updated Files
- spacecraft_pset.py
   - Filter out the other species prior to creating the psets.
- helio_pset.py
   - Filter out the other species prior to creating the psets.

### DIFF
--- a/imap_processing/ultra/l1c/spacecraft_pset.py
+++ b/imap_processing/ultra/l1c/spacecraft_pset.py
@@ -56,7 +56,7 @@ def calculate_spacecraft_pset(
         Ancillary files.
     instrument_id : int
         Instrument ID, either 45 or 90.
-    species_id : int, optional
+    species_id : int
         Species ID, default of 1 refers to Hydrogen.
 
     Returns
--- a/imap_processing/ultra/l1c/spacecraft_pset.py
+++ b/imap_processing/ultra/l1c/spacecraft_pset.py
@@ -33,6 +33,7 @@ def calculate_spacecraft_pset(
     name: str,
     ancillary_files: dict,
     instrument_id: int,
+    species_id: int = 1,
 ) -> xr.Dataset:
     """
     Create dictionary with defined datatype for Pointing Set Grid Data.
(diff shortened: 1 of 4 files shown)


imap_processing/ultra/l1c/spacecraft_pset.py
```diff
@@ -63,22 +66,26 @@ def calculate_spacecraft_pset(
     """
     pset_dict: dict[str, np.ndarray] = {}
     sensor = parse_filename_like(name)["sensor"][0:2]
+    indices = np.where(de_dataset["species"].values == species_id)[0]
```
greglucas, line 70: Add a quick comment about this being the place where you are only keeping this species_id.

greglucas (commented):
Do you ever calculate the species? It looks like it is set to ones everywhere and then filtering on non-ones but that can't happen. Is that coming in a follow-up PR to identify which species is which?

laspsandoval:
pre-commit.ci autofix

laspsandoval:
> Do you ever calculate the species? It looks like it is set to ones everywhere and then filtering on non-ones but that can't happen. Is that coming in a follow-up PR to identify which species is which?

That is correct. I'll check once more with the instrument team. In the past they have said that they want to wait until they are in-flight to filter out anything and that species should be set to Hydrogen for every event initially. I'll double check with them today though.

greglucas (approved):

greglucas:
I don't understand the doc failure for you...

```
/home/runner/work/imap_processing/imap_processing/docs/source/code-documentation/generated/imap_processing.ultra.l1c.spacecraft_pset.calculate_spacecraft_pset.rst:2: WARNING: py:class reference target not found: optional [ref.class]
```

But, it says that is how you should write it I believe.
https://numpydoc.readthedocs.io/en/latest/format.html

```
If it is not necessary to specify a keyword argument, use optional:

x : int, optional
```

Maybe just remove that for now...?

PR #2131, opened 2025-08-29
ULTRA l1c validation - write out scattering variables

lacoak21:
# Change Summary

## Overview
The ULTRA IT team requests that the scattering values for theta and phi and the scattering thresholds for each energy bin are stored in the l1c helio and spacecraft pset. 

## Updated Files
   - Write out scatter theta, phi, and thresholds
-  imap_processing/ultra/l1c/spacecraft_pset.py
     - Write out scatter theta, phi, and thresholds
-imap_processing/ultra/l1c/l1c_lookup_utils.py
     - Return averaged theta and phi scattering values over all the spin phases. 

## Testing
Update test data to have the same number of pixels.


laspsandoval (approved):
Pretty straightforward. Nice work.

PR #2132, opened 2025-08-29
ULTRA l1c rename lookup files

lacoak21:
# Change Summary

## Overview
Rename lookup tables. 

## Updated Files

- For all files remove -n32 from the filename. The instrument team will provide new lookup tables for nside(128) but I do not want to have to update the functions and dependency csv to account for this.


laspsandoval (approved):
Thanks Luisa!

PR #2134, opened 2025-08-29
CoDICE L2 - add function for geometric factors

vmartinez-cu:
This PR adds a function to compute geometric factors needed for L2 intensity algorithms. It also adds a hard coded value for efficiency, but this will need to be replaced at a later time. This work was done as part of implementing the lo-sw-species algorithm, but I didn't have enough time to complete it, so that it still a work in progress. The new function should help with multiple algorithms.


imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml
```diff
@@ -234,3 +234,9 @@ imap_codice_l2_hi-direct-events:
     Data_type: L2_hi-direct-events>Level-2 Hi Event Data
     Logical_source: imap_codice_l2_hi-direct-events
     Logical_source_description: IMAP Mission CoDICE Hi Level-2 Event Data.
+
+imap_codice_l2_lo-sw-species:
+  <<: *instrument_base
+  Data_type: L2_lo-sw-species>Level-2 Lo Sunward Species Counts Data
```
bourque:
```suggestion
  Data_type: L2_lo-sw-species>Level-2 Lo Sunward Species Intensities Data
```

imap_processing/tests/codice/test_codice_l2.py
bourque, line 1: These test additions are great!

imap_processing/codice/codice_l2.py
```diff
@@ -169,12 +177,97 @@ def add_dataset_attributes(
...
+    Geometric factors are determined by comparing the half-spin values per
+    esa_step in the HALF_SPIN_LUT to the rgfo_half_spin values in the provided
+    L2 dataset.
+
+    If the half-spin value is less than the corresponding rgfo_half_spin value,
+    the geometric factor is set to 0.75 (full mode); otherwise, it is set to 0.5
+    (reduced mode).
+
+    NOTE: Half spin values are associated with ESA steps which corresponds to the
+    index of the energy_per_charge dimension that is between 0 and 127.
```
bourque, line 209: Excellent description!

bourque (approved):
Nice work Veronica, thanks for adding this!

imap_processing/tests/codice/test_codice_l2.py
```diff
@@ -45,6 +50,25 @@ def mock_cdf_attrs():
...
+    """
+    mock_lut = {
+        1: list(range(0, 64)),
+        2: list(range(64, 128)),
+    }
+    monkeypatch.setitem(
+        compute_geometric_factors.__globals__,
+        "HALF_SPIN_LUT",
+        mock_lut,
+    )
```
greglucas: Rather than updating the globals which can be a bit dangerous can you update the attr instead?

`monkeypatch.setattr(compute_geometric_factors, "HALF_SPIN_LUT", mock_lut)`

One possible issue is that within the module itself you are importing the value directly rather than importing the module containing the value, so the value may already be loaded and in place. I think we should really try and get away from that and move more towards the
`constants.HALF_SPIN_LUT` style usage which I think then would allow the above to work better. If this doesn't work though, it probably isn't worth changing right now.

greglucas (commented):
Nice work cranking this out so quickly!

Looks like there are some potentially related failing tests:

```
ERROR imap_processing/tests/codice/test_codice_l2.py::test_l2_logical_sources[test_l2_data0-imap_codice_l2_hi-direct-events] - AttributeError: 'Dataset' object has no attribute 'rgfo_half_spin'
ERROR imap_processing/tests/codice/test_codice_l2.py::test_l2_logical_sources[test_l2_data1-imap_codice_l2_lo-direct-events] - AttributeError: 'Dataset' object has no attribute 'rgfo_half_spin'
```

tech3371 (approved):
I didn't pay close attention to the discussion because I was working on DE questions at the time. Thank you for setting these up so quick! Appreciate it!

imap_processing/tests/codice/test_codice_l2.py
vmartinez-cu: Updated, thanks!

Issue #1475, opened 2025-03-17
SPICE - spacecraft state at mid-point of pointing


tmplummer:
### Algorithm Description:

Implement a function in geometry.py that returns the s/c state (position and velocity) at the midpoint of a pointing.

Requires: #1474

tech3371:
Once we have the mid-point time from previous ticket. convert mid-point time to ET and then call geometry.imap_state function. reference frame is ECLIPJ2000, observer is SUN which are the defaults.

tech3371:
Completed in PR #2167

PR #2167, opened 2025-09-02
ENH: repoint mid-time utils function

### LINKED ISSUE #1474: SPICE repoint - get mid time between repoints
### Algorithm Description:

This is for a common function that will get the mid-point time between the end of one repoint and start of the next. Input could be a MET time or maybe a repoint ID (number).

tech3371:
# Change Summary
closes #1474 #1475

## Overview

Repoint mid-point utils functions.

## Updated Files

   - mid-time between repoints
   - repoint mid-time spacecraft state

## Testing

### DIFF
--- a/imap_processing/spice/repoint.py
+++ b/imap_processing/spice/repoint.py
@@ -225,7 +225,7 @@ def get_pointing_times(met_time: float) -> tuple[float, float]:
     return pointing_start_met, pointing_end_met
 
 
-def get_repoint_mid_time(met_time: float) -> float:
+def get_pointing_mid_time(met_time: float) -> float:
     """
     Get mid-point of the pointing for the given MET time.
 
@@ -264,7 +264,7 @@ def get_mid_point_state(met_time: float) -> npt.NDArray:
         The mid state of the pointing maneuver.
     """
     # Get mid point time in ET
-    mid_point_time = get_repoint_mid_time(met_time)
+    mid_point_time = get_pointing_mid_time(met_time)
     mid_point_time_et = sct_to_et(met_to_sclkticks(mid_point_time))
 
     # Convert mid point time to state
--- a/imap_processing/spice/repoint.py
+++ b/imap_processing/spice/repoint.py
@@ -9,6 +9,8 @@ import pandas as pd
 from numpy import typing as npt
 
 from imap_processing.spice import config
+from imap_processing.spice.geometry import imap_state
+from imap_processing.spice.time import met_to_sclkticks, sct_to_et
 
 logger = logging.getLogger(__name__)
 
@@ -221,3 +223,52 @@ def get_pointing_times(met_time: float) -> tuple[float, float]:
     ][0]
     pointing_end_met = repoint_df["repoint_start_met"].iloc[pointing_idx + 1].item()
     return pointing_start_met, pointing_end_met
+
+
+def get_repoint_mid_time(met_time: float) -> float:
+    """
+    Get mid-point of the pointing for the given MET time.
+
+    Get the mid-point time between the end of one repoint and
+    start of the next. Input could be a MET time.
+
+    TODO: Fix this if mid-point definition is changed later.
+
+    Parameters
+    ----------
+    met_time : float
+        The MET time in a repoint.
+
+    Returns
+    -------
+    repoint_mid_time : float
+        The mid MET time of the repoint maneuver.
+    """
+    pointing_start_met, pointing_end_met = get_pointing_times(met_time)
+    return float(np.mean([pointing_start_met, pointing_end_met]))
+
+
+def get_mid_point_state(met_time: float) -> float:
+    """
+    Get IMAP state for the mid-point.
+
+    Get IMAP state for the mid-point of the pointing in
+    reference frame, ECLIPJ2000 and observer, SUN.
+
+    Parameters
+    ----------
+    met_time : float
+        The MET time in a pointing.
+
+    Returns
+    -------
+    mid_point_state : float
+        The mid state of the pointing maneuver.
+    """
+    # Get mid point time in ET
+    mid_point_time = get_repoint_mid_time(met_time)
+    mid_point_time_et = sct_to_et(met_to_sclkticks(mid_point_time))
+
+    # Convert mid point time to state
+    pointing_state = imap_state(mid_point_time_et)
+    return float(pointing_state)
--- a/imap_processing/spice/repoint.py
+++ b/imap_processing/spice/repoint.py
@@ -248,7 +248,7 @@ def get_repoint_mid_time(met_time: float) -> float:
     return float(np.mean([pointing_start_met, pointing_end_met]))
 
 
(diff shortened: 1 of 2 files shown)


imap_processing/spice/repoint.py
```diff
@@ -221,3 +223,52 @@ def get_pointing_times(met_time: float) -> tuple[float, float]:
...
+
+
+def get_repoint_mid_time(met_time: float) -> float:
+    """
+    Get mid-point of the pointing for the given MET time.
+
+    Get the mid-point time between the end of one repoint and
+    start of the next. Input could be a MET time.
+
+    TODO: Fix this if mid-point definition is changed later.
```
greglucas: This isn't super helpful IMO because we won't look through all TODOs if/when we do change it later.

I think this is already handled by getting the pointing_start / pointing_end from a single function where that should update the values if needed, but the midpoint calculation will stay the same.

imap_processing/spice/repoint.py
```diff
@@ -221,3 +223,52 @@ def get_pointing_times(met_time: float) -> tuple[float, float]:
...
+    met_time : float
+        The MET time in a repoint.
+
+    Returns
+    -------
+    repoint_mid_time : float
+        The mid MET time of the repoint maneuver.
+    """
+    pointing_start_met, pointing_end_met = get_pointing_times(met_time)
+    return float(np.mean([pointing_start_met, pointing_end_met]))
```
greglucas:
```suggestion
    return (pointing_start_met + pointing_end_met) / 2
```

imap_processing/spice/repoint.py
```diff
@@ -221,3 +223,52 @@ def get_pointing_times(met_time: float) -> tuple[float, float]:
...
+
+
+def get_mid_point_state(met_time: float) -> npt.NDArray:
```
greglucas, line 249: This seems like we would want one or the other function, but possibly not both?

The above equation is very simple, so do we need a separate function call for that or can calling sites just do the math themselves?

This function calls `imap_state(midpoint_time)`, but takes a met_time as input, which isn't the midpoint so is a bit confusing. I'm curious how we can improve this. Perhaps something like `get_midpoint(pointing_number)` to do the table lookup? Keying this off of met just feels a bit off to me.

imap_processing/tests/spice/test_repoint.py
```diff
@@ -81,6 +82,41 @@ def test_get_pointing_times(fake_repoint_data):
...
+    kernels = [
+        "naif0012.tls",
+        "imap_sclk_0000.tsc",
+    ]
+    with furnish_kernels(kernels):
+        state = repoint.get_mid_point_state(met)
+        if hasattr(met, "__len__"):
+            np.testing.assert_array_equal(state.shape, (len(met), 6))
+        else:
+            assert state.shape == (6,)
```
greglucas: Why would we want to test this for an array? Your functions above say that it should be a float input. This function seems like a one-off that we may not want to vectorize, or do it in some other way. I just worry that someone is going to pass in all of their DE times, but they are all during one single pointing so we shouldn't actually recompute all these values every time, but rather re-use that first calculation throughout the code.

imap_processing/spice/repoint.py
tech3371: So remove TODO from this function at least?

imap_processing/spice/repoint.py
tech3371: Sounds good. This is what AI suggested but Tim and I discussed using np.mean.

imap_processing/tests/spice/test_repoint.py
tech3371: Input is float but return value is vector of 6 element indicating state. At least, that's my understanding

imap_processing/spice/repoint.py
tech3371: May be I misunderstood what Tim and I discussed. So you are saying we should do this instead?
```
def get_imap_state(mid_time):
       ....
met_time = 6
mid_time = get_repoint_mid_time(met_time)
```

tech3371: But I don't know who is the user and their use case. Tim knows better.

imap_processing/spice/repoint.py
greglucas: Yep

imap_processing/spice/repoint.py
greglucas: Same thing, but this is simpler and faster.

imap_processing/spice/repoint.py
greglucas: I don't think we should make a `get_imap_state()` functino, that is already the case with `imap_state()`, so we should just have the user make use of that function I think.

imap_processing/tests/spice/test_repoint.py
greglucas: Yes, but the first case is testing for multiple METs being input to the function which I don't think we want to allow. i.e. the only assertion should be the else-block.

imap_processing/spice/repoint.py
tech3371: Ok. I am going to ask Tim on what approach he likes and then make change here as needed.

tmplummer: Both of these functions were requested by GLOWS. They wanted all instruments to use the same function to get the midpoint time of a pointing and spacecraft state at that time. Maybe having a common function to get the midpoint time is sufficient.

Using pointing number as input may be more appropriate and less confusing. Is there a function that gets the pointing index given a MET time?

imap_processing/spice/repoint.py
```diff
@@ -221,3 +223,50 @@ def get_pointing_times(met_time: float) -> tuple[float, float]:
     ][0]
     pointing_end_met = repoint_df["repoint_start_met"].iloc[pointing_idx + 1].item()
     return pointing_start_met, pointing_end_met
+
+
+def get_repoint_mid_time(met_time: float) -> float:
```
tmplummer: This should be renamed to `get_pointing_mid_time` and comments below should clear up that this gets the midpoint of the pointing, not the repoint maneuver.

tmplummer (commented):
Sorry for the confusion. I think @greglucas is probably correct. Just having one function to get the pointing middle time for a given pointing number would suffice.

imap_processing/spice/repoint.py
tech3371: No we don't have that function yet. Can we do that later?

Also, do you both mind if I keep both functions for now to reduce duplicate code.

greglucas: Sure that is fine with me. Can you open a separate ticket that is something like: cleanup repointing calculations and handle integer repoint number to be passed in instead of an event MET

greglucas (approved):

imap_processing/spice/repoint.py
tech3371: Sure thing!

tmplummer (approved):

PR #2169, opened 2025-09-03
Lo - Added Pointing Set Background Fields

### LINKED ISSUE #2168: Lo - Add Background fields to pointing set
Add the Background rate, uncertainties, and error fields to the pointing set

sdhoyt:
# Change Summary

## Overview
- Added pointing set background fields (rates, uncertainties, errors)
- Added code to grab ESA mode from annotated direct events and temporary code to set it to 0 if it's not in the ADE (one of the fixes needed for the ADE is to add the ESA mode)
- updated the test background ancillary files to use the new binning

Closes #2168


imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -484,3 +491,98 @@ def create_datasets(
...
+    -------
+    background_rates : xarray.DataArray
+        The background rates for the specified species.
+    """
+    if species not in {FilterType.HYDROGEN, FilterType.OXYGEN}:
+        raise ValueError("Species must be 'h' or 'o'.")
+
+    bg_rates = np.zeros((7, 3600, 40), dtype=np.float16)
+    bg_stat_uncert = np.zeros((7, 3600, 40), dtype=np.float16)
+    bg_sys_err = np.zeros((7, 3600, 40), dtype=np.float16)
```
greglucas: There are now global variables at the top of this file you can use here instead.

```suggestion
    bg_rates = np.zeros((N_ESA_ENERGY_STEPS, N_SPIN_ANGLE_BINS, N_OFF_ANGLE_BINS), dtype=np.float16)
    bg_stat_uncert = np.zeros((N_ESA_ENERGY_STEPS, N_SPIN_ANGLE_BINS, N_OFF_ANGLE_BINS), dtype=np.float16)
    bg_sys_err = np.zeros((N_ESA_ENERGY_STEPS, N_SPIN_ANGLE_BINS, N_OFF_ANGLE_BINS), dtype=np.float16)
```

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -484,3 +491,98 @@ def create_datasets(
...
+
+    # read in the background rates from ancillary file
+    if species == FilterType.HYDROGEN:
+        background_df = lo_ancillary.read_ancillary_file(
+            next(s for s in anc_dependencies if "hydrogen-background" in s)
```
greglucas, line 545: I'm wondering if rather than passing in a list of ancillary files we should pass in the `ProcessingInputCollection` directly so that you could do the filtering through that class.

```
dependencies.get_file_paths(source="lo", descriptor="hydrogen-background")[0]
```

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -484,3 +491,98 @@ def create_datasets(
...
+    ]
+
+    # for each row in the bg ancillary file for this pointing
+    for _, row in pointing_bg_df.iterrows():
+        # convert the bin start and end resolution from 6 degrees to .1 degrees
+        bin_start = int(row["bin_strt"]) * 60
+        # The last bin end in the file is 0, which means 60 degrees. This is
+        # converted to 0.1 degree resolution of 3600
+        bin_end = int(row["bin_end"]) * 60
+        bin_end = np.where(bin_end == 0, 3600, bin_end)
```
greglucas: I don't think you need the where here and could just update those indices directly.
```suggestion
        bin_end[bin_end == 0] = 3600
```

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -484,3 +491,98 @@ def create_datasets(
...
     return dataset
+
+
+def set_background_rates(
+    pointing_start_met: float,
+    pointing_end_met: float,
+    species: FilterType,
+    anc_dependencies: list,
+    attr_mgr: ImapCdfAttributes,
+) -> xr.DataArray:
```
greglucas: I'm surprised this passed mypy since it is returning a tuple of DataArrays and not a single one.

As a potential follow-up later, I'm wondering if we should switch these over to be a `updated_dataset = process_function(input_dataset, dependencies)`. So you would do whatever you needed to the input dataset rather than having to extract the pointing start/end beforehand.

Also, you are passing in attr_mgr here, but in other places in Lo I think you're doing it all at the end. So create the entire dataset, then at the end iterate through the variables you created and grab the metadata associated with those fields.

```suggestion
    pointing_start_met: float,
    pointing_end_met: float,
    species: FilterType,
    anc_dependencies: list,
    attr_mgr: ImapCdfAttributes,
) -> Tuple[xr.DataArray, xr.DataArray, xr.DataArray]:
```

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -484,3 +491,98 @@ def create_datasets(
...
+    attr_mgr : ImapCdfAttributes
+        Attribute manager used to get the L1C attributes.
+
+    Returns
+    -------
+    background_rates : xarray.DataArray
+        The background rates for the specified species.
+    """
+    if species not in {FilterType.HYDROGEN, FilterType.OXYGEN}:
+        raise ValueError("Species must be 'h' or 'o'.")
```
greglucas: If you are going to put this check on, you should probably add a quick test for it too.

Add to the error message, f"got '{species}'" so that we know what was attempted to be passed.

```python
def test_background_species_raises():
    with pytest.raises(ValueError, "Species must be 'h'"):
        set_background_species(..., "abc")

**Review by greglucas — COMMENTED** [review]
Overall looks good, some minor nits is all.

File: imap_processing/lo/l1c/lo_l1c.py [hunk 6]
```diff
@@ -484,3 +491,111 @@ def create_datasets(
...
+        bin_end = int(row["bin_end"])
+        # for each energy step, set the background rate and uncertainty
+        for esa_step in range(0, 7):
+            value = row[f"E-Step{esa_step + 1}"]
+            if row["type"] == "rate":
+                bg_rates[esa_step, bin_start:bin_end, :] = value
+            elif row["type"] == "sigma":
+                bg_stat_uncert[esa_step, bin_start:bin_end, :] = value
+            else:
+                print("TYPE", row["type"])
```
greglucas, line 577: Stray print
```suggestion
```

imap_processing/tests/lo/test_lo_l1c.py
```diff
@@ -271,3 +313,46 @@ def test_calculate_exposure_times(l1b_de):
...
+
+def test_set_background_rates_species_error(anc_dependencies, attr_mgr):
+    # Arrange
+    pointing_start_met = 473389100.0
+    pointing_end_met = 473472100.0
+    species = FilterType.DOUBLES
+
+    # Act
+    with pytest.raises(
+        ValueError, match=f"Species must be 'h' or 'o', but got {species.value}."
```
greglucas: Just keep it simple for the match and look at the first few characters to generally say it was this error versus from somewhere else, we don't really care about the exact text.
```suggestion
        ValueError, match=f"Species must be 'h' or 'o'"
```

greglucas (approved):
A few minor things to clean up the failing test, otherwise good to go!

PR #2171, opened 2025-09-03
2114 hilo l2   flux correction predictor corrector scheme

### LINKED ISSUE #2114: Hi/Lo L2 - flux correction predictor corrector scheme
### Algorithm Description:

Implement the flux correction predictor corrector scheme from the Lo Mapping Algorithm Document

tmplummer:
# Change Summary

## Overview
Sorry for such a large PR :(. I would normally break this up but am trying to go fast.

Adds Hi/Lo L2 flux correction predictor-corrector algorithm code.

## New Files
   - Module for common Hi/Lo L2 correction code which will include flux and CG corrections.
- imap_processing/tests/ena_maps/data/imap_hi_90sensor-esa-eta-fit-factors_20240101_v001.csv
   - Polynomial factors for computing hi90 transmissions from energies and slopes
- imap_processing/tests/ena_maps/data/imap_lo_esa-eta-fit-factors_20240101_v001.csv
   - Polynomial factors for computing Lo transmissions from energies and slopes
   - Test coverage for corrections module

## Updated Files
- imap_processing/tests/ena_maps/conftest.py
   - add fixture for locating ena_maps test data
- updated file 2

Closes: #2114


imap_processing/ena_maps/utils/corrections.py
```diff
@@ -0,0 +1,299 @@
...
+    ----------
+    coeffs_file : str or Path
+        Location of CSV file containing ESA transmission coefficients.
+    """
+
+    def __init__(self, coeffs_file: str | Path):
+        """Initialize PowerLawFluxCorrector."""
+        self._load_coefficients(coeffs_file)
+
+    def _load_coefficients(self, coeffs_file: str | Path) -> None:
```
greglucas: Just inline this into the `__init__` function since it is a single line?

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        k = np.atleast_1d(k)
+        gamma = np.atleast_1d(gamma)
+        coeff_columns = ["M0", "M1", "M2", "M3", "M4", "M5"]
+        eta = np.empty_like(gamma)
+        for i, esa_step in enumerate(k):
+            # np.polyval expects polynomial coeffs in reverse order
+            eta[i] = np.polyval(
+                self.eta_coeffs_df.loc[esa_step, coeff_columns].values[::-1], gamma[i]
+            )
```
greglucas: Not sure if this would be better or not, but I think numpy would like to transition to the Polynomial class. So something like this to keep things in the same order might work too?
https://numpy.org/doc/stable/reference/routines.polynomials.html

```suggestion
            eta[i] = np.polynomial.Polynomial(
                self.eta_coeffs_df.loc[esa_step, coeff_columns].values)(gamma[i])
            )
```

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -0,0 +1,299 @@
...
+        right_log_fluxes = log_extended_fluxes[2:]  # indices 2 to n_levels+1
+        left_log_energies = log_extended_energies[:-2]
+        right_log_energies = log_extended_energies[2:]
+
+        # Compute power-law slopes for valid indices
+        central_valid = np.isfinite(left_log_fluxes) & np.isfinite(right_log_fluxes)
+        gamma[central_valid] = (
+            (right_log_fluxes - left_log_fluxes)
+            / (right_log_energies - left_log_energies)
+        )[central_valid]
```
greglucas, line 131: Will this spit out warnings? I'm wondering if you need to put any `with np.errstate(): ignore` block in here to try and minimize the extraneous warning.

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Handle one-sided differencing for points where central differencing failed
+        need_fallback = ~central_valid & np.isfinite(log_fluxes)
+        # Exclude first and last points since they already use the correct
+        # one-sided differencing
+        interior_fallback = (
+            need_fallback
+            & (np.arange(n_levels) > 0)
+            & (np.arange(n_levels) < n_levels - 1)
+        )
```
greglucas: What about setting the specific indices you need? This seems a little clearer to follow for me at first glance
```suggestion
        interior_fallback = np.zeros_like(need_fallback, dtype=bool)
        interior_fallback[1:-1] = need_fallback[1:-1]
```

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -0,0 +1,299 @@
...
+
+            for k in indices:
+                # For interior points: try forward first, then backward
+                if k < n_levels - 1 and np.isfinite(log_fluxes[k + 1]):
+                    gamma[k] = (log_fluxes[k + 1] - log_fluxes[k]) / (
+                        log_energies[k + 1] - log_energies[k]
+                    )
+
+                    # Compute uncertainty slope using same differencing
+                    if isinstance(delta_gamma, np.ndarray):
```
greglucas, line 161: You use `delta_gamme is not None` above I thought, so do you want to keep it consistent with that check?
```suggestion
                    if delta_gamma is not None:
```

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -0,0 +1,299 @@
...
+        corrected_flux = np.empty_like(flux)
+        corrected_flux_stat_unc = np.empty_like(flux_stat_unc)
+
+        # loop over spatial pixels (last dimension)
+        for i_pixel in range(flux.shape[-1]):
+            corrected_flux[:, i_pixel], corrected_flux_stat_unc[:, i_pixel], _ = (
+                self.predictor_corrector_iteration(
+                    flux[:, i_pixel], flux_stat_unc[:, i_pixel], energies
+                )
+            )
```
greglucas, line 289: I know you're going to hate me for saying this, but I am curious if this could be vectorized out as well... (Claude is quite good at these kinds of updates later too, so don't feel like you need to do it now)

You could keep iterating until the largest pixel error was below your threshold, or 20 iterations, but I don't think it matters if you over-converge. But that loop over 20 is much smaller than this exterior loop over 3600 / 1800.

imap_processing/tests/ena_maps/test_corrections.py
```diff
@@ -0,0 +1,231 @@
...
+        assert eta[1] == 1
+
+    def test_estimate_power_law_with_uncertainties(self):
+        """Test slope estimation with flux uncertainties."""
+
+        fluxes = np.array([10, 20, 40, 80, 160, 320, 640])
+        energies = np.arange(8) + 1
+        uncertainties = np.sqrt(fluxes)
+        corr = PowerLawFluxCorrector.__new__(PowerLawFluxCorrector)
+        gamma, delta_gamma = corr.estimate_power_law_slope(
```
greglucas: This seems like an odd pattern, why not initialize it right away `PowerLawFluxCorrect().estimate_power_law_slope(...)`, or make this a static method if self isn't needed/used.

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -0,0 +1,299 @@
...
+        -------
+        tuple[numpy.ndarray, numpy.ndarray]
+            Corrected fluxes and flux uncertainties.
+        """
+        corrected_flux = np.empty_like(flux)
+        corrected_flux_stat_unc = np.empty_like(flux_stat_unc)
+
+        # loop over spatial pixels (last dimension)
+        for i_pixel in range(flux.shape[-1]):
+            corrected_flux[:, i_pixel], corrected_flux_stat_unc[:, i_pixel], _ = (
```
greglucas, line 285: Is there any value in storing the energies returned as well? I thought we might want to look at those for what the distribution is across the map, and/or need them for other corrections later.

imap_processing/ena_maps/utils/corrections.py
tmplummer: Good point. The log function does spit out divide by zero warnings. Energies should never be zero, so this particular equation should be fine.

imap_processing/ena_maps/utils/corrections.py
tmplummer: Yep. I like that better. Forgot to clean up this Claude artifact.

imap_processing/ena_maps/utils/corrections.py
tmplummer: Yeah... I had it that way and Mypy complained about the assignment in this `if` block.

imap_processing/ena_maps/utils/corrections.py
tmplummer: Hmmmm, I think that algorithm is correcting the fluxes to the input energies. So, the energies don't change.

imap_processing/ena_maps/utils/corrections.py
tmplummer: Haha, I knew that was coming. I thought about that and worried that it would make the algorithm totally unreadable. There is also the issue that the polynomial evaluation is not vectorized. I suppose that I could vectorize that if I do the calculation manually rather than using np.polyval. I will also check if the Polynomial class can be vectorized.

imap_processing/ena_maps/utils/corrections.py
greglucas: :+1: So the next ticket is going to be the adjustment for energies then? I guess my ultiamte question was whether the `_` is even needed or if you can change the return signature to just return the fluxes without the energies.

imap_processing/ena_maps/utils/corrections.py
greglucas: Yeah, lets punt this. I think this could be a rabbit hole. A nice-to-have, but now that you have this all set up with nice tests I think it is easier to do this in a follow-up ticket if we find it is necessary.

imap_processing/ena_maps/utils/corrections.py
greglucas: Gotta love mypy...

greglucas (approved):
This is great! I am not sure how much smaller you could have made this. It would have been a decent amount of work to do that I think, so I understand here.

PR #2175, opened 2025-09-04
ULTRA l2 map attributes 

### LINKED ISSUE #2078: Ultra L2 validation - access what needs to be updated
### Summary

[l2_validation.pptx](https://github.com/user-attachments/files/21907229/l2_validation.pptx)

### Requirements

The more specific the better.
  - Using test data that I have created as an input: https://drive.google.com/file/d/1_D27ftk9ICe2bkmLUO_3holz-E0KwwR6/view?usp=sharing try to run the L2 code and produce a map
  - Look at the output cdf and compare with the validation slide (above) to see what we need to fix
  - Make more tickets to break down what we need to update/change

Instructions to run the L2 code were created by Nat and are located here:
https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/270110502/Running+Ultra+L2+Documentation

lacoak21:
# Change Summary

closes #2078

## Overview
Update ena map attributes to include efficiency, geometric function, and positional uncertainty. 

## Updated Files
   - Add shared attribute values
- imap_processing/cdf/config/imap_enamaps_l2-healpix_variable_attrs.yaml
   - Add healpix attrs
- imap_processing/cdf/config/imap_enamaps_l2-rectangular_variable_attrs.yaml
   - Add rectangular attrs
- imap_processing/tests/ultra/unit/test_ultra_l2.py
  - Keep obs date range and std ( IT request) 
- rename any scatter theta or scatter phi vars to positional uncertainty

### DIFF
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -220,7 +220,7 @@ geometric_function:
 
 efficiency:
   <<: *default_float32
-  CATDESC: Estimated event efficiency for particles path through the instrument calculated from multiple pointing sets.
+  CATDESC: Event efficiency calculated from multiple pointing sets.
   FIELDNAM: efficiency
   UNITS: " "
   DEPEND_0: epoch
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -229,7 +229,7 @@ efficiency:
   DISPLAY_TYPE: no_plot
   DICT_KEY: SPASE>Particle>ParticleType:Atom,ParticleQuantity:Other
 
-positional_uncertainty_theta:
+positional_uncert_theta:
   <<: *default_float32
   CATDESC: Positional uncertainty in theta direction calculated from multiple pointing sets.
   FIELDNAM: positional_uncertainty_theta
@@ -240,7 +240,7 @@ positional_uncertainty_theta:
   DISPLAY_TYPE: no_plot
   DICT_KEY: SPASE>Particle>ParticleType:Atom,ParticleQuantity:ArrivalDirection,Qualifier:Uncertainty,CoordinateSystemName:HAE,CoordinateRepresentation:Spherical
 
-positional_uncertainty_phi:
+positional_uncert_phi:
   <<: *default_float32
   CATDESC: Positional uncertainty in phi direction calculated from multiple pointing sets.
   FIELDNAM: positional_uncertainty_phi
@@ -266,20 +266,8 @@ obs_date: &obs_date
 obs_date_range:
   <<: *default_int64
   datatype: int64
-  CATDESC: Range of the observation dates.
-  FIELDNAM: Observation date range
-  UNITS: ns
-  DEPEND_0: epoch
-  VAR_TYPE: data
-  LABLAXIS: epoch
-  DISPLAY_TYPE: image
-  TIME_SCALE: Terrestrial Time
-
-obs_date_for_std:
-  <<: *default_int64
-  datatype: int64
   CATDESC: Standard deviation of the observation date.
-  FIELDNAM: Observation date standard deviation
+  FIELDNAM: Observation date range
   UNITS: ns
   DEPEND_0: epoch
   VAR_TYPE: data
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -207,6 +207,50 @@ exposure_factor: &exposure_factor
   DISPLAY_TYPE: no_plot
   DICT_KEY: SPASE>TemporalDescription:Exposure,Qualifier:Directional,CoordinateSystemName:HAE,CoordinateRepresentation:Spherical
 
+geometric_function:
+  <<: *default_float32
+  CATDESC: The geometric function calculated from multiple pointing sets.
+  FIELDNAM: geometric_function
+  UNITS: cm^2 sr
+  DEPEND_0: epoch
+  VAR_TYPE: data
+  LABLAXIS: Geometric Factor
+  DISPLAY_TYPE: no_plot
+  DICT_KEY: SPASE>Particle>ParticleType:Atom,ParticleQuantity:GeometricFactor,Qualifier:Directional,CoordinateSystemName:HAE,CoordinateRepresentation:Spherical
+
+efficiency:
+  <<: *default_float32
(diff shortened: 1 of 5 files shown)


imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -196,6 +196,50 @@ exposure_factor: &exposure_factor
   DISPLAY_TYPE: no_plot
   DICT_KEY: SPASE>TemporalDescription:Exposure,Qualifier:Directional,CoordinateSystemName:HAE,CoordinateRepresentation:Spherical
 
+geometric_function:
```
greglucas, line 210: I think this should be geometric factor?
```suggestion
geometric_factor:
```

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -196,6 +196,50 @@ exposure_factor: &exposure_factor
...
+  CATDESC: The geometric function calculated from multiple pointing sets.
+  FIELDNAM: geometric_function
+  UNITS: cm^2 sr
+  DEPEND_0: epoch
+  VAR_TYPE: data
+  LABLAXIS: Geometric Factor
+  DISPLAY_TYPE: no_plot
+  DICT_KEY: SPASE>Particle>ParticleType:Atom,ParticleQuantity:GeometricFactor,Qualifier:Directional,CoordinateSystemName:HAE,CoordinateRepresentation:Spherical
+
+efficiency:
```
greglucas, line 221: Some of these are ultra-specific and will be different for Lo/Hi I believe. Do we want to add a separate file like `enamaps_l2-ultra` and similar for some of these?

On Lo, our efficiency is per pointing (so not on the map), and it has a dependence on energy. I'm not sure if yours does or not (looks like no here)

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -541,6 +540,13 @@ def ultra_l2(
     # to "energy" for all instruments.
     map_dataset = map_dataset.rename({"energy_bin_geometric_mean": "energy"})
 
+    # Rename positional uncertainty variables if present
+    if "scatter_theta" in map_dataset and "scatter_phi" in map_dataset:
+        map_dataset = map_dataset.rename(
+            {"scatter_theta": "positional_uncertainty_theta"}
+        )
+        map_dataset = map_dataset.rename({"scatter_phi": "positional_uncertainty_phi"})
```
greglucas: This won't fail if the variables aren't present, so you can just call the rename directly and it will work if they are there.

```suggestion
    # Rename positional uncertainty variables if present
    map_dataset = map_dataset.rename({"scatter_theta": "positional_uncertainty_theta", "scatter_phi": "positional_uncertainty_phi"})
```

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
greglucas: I see, the energy depend was defined in the specific healpix/latlon files. I guess that could be brought out here since it is the same for all of them, but it doesn't really matter. I guess you can ignore this comment for now.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -220,6 +264,18 @@ obs_date_range:
   DISPLAY_TYPE: image
   TIME_SCALE: Terrestrial Time
 
+obs_date_for_std:
```
tmplummer: The product definition agreed to by the IMWG named this `obs_date_range`. Is Ultra requesting a change in that variable name? Here is the spreadsheet with the current definition: https://o365coloradoedu-my.sharepoint.com/:x:/g/personal/plummert_colorado_edu/EX8IRmgRKw5Hpik9wPAXZQABBgOemDqOv-uLMlRxZLJaTw

tmplummer: Oh, I guess I am confused what this is vs. `obs_date` and `obs_date_range`.

tmplummer (commented):
My only comment is that Ultra IT is adding some variables to the L2 definition a bit ad hoc. I guess in the name of progress, that is OK with me.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
lacoak21: I had the same thought but apparently they want geometric function. Will that be going against the grain? Maybe it is worth it to have a separate ultra l2 map yaml file.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
lacoak21: Oh interesting. So the efficiency just has energy as a depend for lo?

imap_processing/cdf/config/imap_enamaps_l2-rectangular_variable_attrs.yaml
```diff
@@ -134,6 +134,38 @@ exposure_factor:
...
+
+geometric_function:
+  DEPEND_1: energy
+  DEPEND_2: longitude
+  DEPEND_3: latitude
+  LABL_PTR_1: energy_label
+  LABL_PTR_2: longitude_label
+  LABL_PTR_3: latitude_label
+
+positional_uncertainty_theta:
```
laspsandoval: Based on the recent Slack messages should this now be uncert instead?

laspsandoval (approved):
Just one question about the uncertainty name. Everything looks good after you respond to Greg and Tim.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
lacoak21: I might have made some assumptions here. They made this note

Screenshot 2025-09-05 at 10 31 09 AM

tmplummer: Ok. I'm not sure what to do for this. Probably move forward with what you have. I will send out an email to the IMWG noting that I am seeing lots of variables being added to L2 products last minute here. All the ENA ITs agreed to the spreadsheet and this is causing quite a bit of churn.

lacoak21: OK I also just messaged Nick Dutton for more info. Yeah sorry about that - I didnt realize these are unexpected.

lacoak21: At least for this issue im wondering if they will be ok with just obs_date and obs_date range, knowing now that obs_date_range seems to be the std already.

PR #2177, opened 2025-09-04
2111 mag l1d l2   use correct instrument frame for geometry

tmplummer:
# Change Summary

## Overview
Updates the imap_frames kernel to the latest version that has been provided. This is not the final version that will use ground cal data. The main change is the removal IMAP_MAG frame and addition of the IMAP_BOOM, IMAP_MAG_I, and IMAP_MAG_O. Because that update broke some tests, I had to do the following:
- Keep the IMAP_MAG frame in geometry.py so that tests wouldn't fail at the top level collect.
- Add `@pytest.mark.xfail(reason="IMAP_MAG frame needs to be updated")` to any test that these changes broke.

There will be one or two follow on tickets to update the MAG and iALiRT MAG code with the appropriate frame.

## New Files
- imap_processing/tests/spice/test_data/imap_001.tf
   - New imap frames kernel

## Deleted Files
- imap_processing/tests/spice/test_data/imap_wkco.tf
   - Removing old imap frames kernel

## Updated Files
Too many to list one by one. The updates in each file should be self explanatory with the Overview in mind.


imap_processing/spice/geometry.py
```diff
@@ -50,13 +50,17 @@ class SpiceFrame(IntEnum):
     IMAP_HI_90 = -43160
     IMAP_ULTRA_45 = -43200
     IMAP_ULTRA_90 = -43210
-    IMAP_MAG = -43250
+    # TODO: remove IMAP_MAG frame once all usages have been removed
+    IMAP_MAG = -43999
+    IMAP_MAG_BOOM = -43250
+    IMAP_MAG_I = -43251
```
maxinelasp, line 56: nitpick but - would it be possible to name them `IMAP_MAGO` and `IMAP_MAGI`? That would match the naming scheme everywhere else.

maxinelasp (approved):

imap_processing/spice/geometry.py
tmplummer: No :( We don't have control over these. They need to match what is in the frames kernel.

maxinelasp: Ah well, c'est la vie! Thanks for making those changes :)

PR #2178, opened 2025-09-04
Lo L1B - add ESA mode and fix spin cycle in ADE

sdhoyt:
# Change Summary

## Overview

- ESA Mode field was added to the ADE from the sweep table
- Previous implementation of the spin cycle had the spin number resetting to 0 for each pointing. The code was updated to get the correct spin number from the spin table
- the `mode` field was also renamed to `mode_bit`

Follow-on PRs will be created to address the pointing bin structure and direction


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -275,19 +350,18 @@ def set_spin_cycle(l1a_de: xr.Dataset, l1b_de: xr.Dataset) -> xr.Dataset:
...
     spin_cycle = []
-    for i, esa_asc_group in enumerate(de_asc_groups):
-        # TODO: Spin Number does not reset for each pointing. Need to figure out
-        #  how to retain this information across days
-        # increment the spin_start by 28 after each aggregated science cycle
-        spin_start = i * 28
+    for esa_asc_group in de_asc_groups:
         # calculate the spin cycle for each DE in the ASC group
         # TODO: Add equation number in algorithm document when new version is
         # available. Add to docstring as well
```
greglucas: Do you still need this new equation, or can this be removed?

greglucas (approved):

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Yeah, I should be able to add that now. I'll update it

sdhoyt: Never mind, it looks like it hasn't been added to the ADE section yet. I'll talk to Lo about it and add it later.

PR #2180, opened 2025-09-04
ENH: Add intensity calculations to the Lo mapping code

### LINKED ISSUE #1888: Lo L2 - Create ENA Flux maps in the spacecraft frame, uncorrected
Create the ENA flux maps for Hydrogen and Oxygen. These CDFS include counts, rates, and fluxes as well as background rates and fluxes.

For more details see section 12.1 of the Lo algorithm document: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/146768943/IMAP+Algorithm+Document+from+Instrument+Teams

greglucas:
This is a major refactor of the L2 code to try and break out individual steps into easily testable blocks.

* I moved the geometric efficiencies into the main code area as these are fixed as I understand it for now. The issue is that these files are not named correctly for our ancillary file handler, so I couldn't download them on-demand like the other ancillary files. So this was the quick fix. Feel free to push back if you think we should approach this a different way, I'm not tied to it at all.

@subagonsouth, I think I unfortunately did even worse than your large PR!

The real work here is in the refactoring of L2 which was hard-coded logic before. Claude helped with the refactor and tests, but I went through and verified this with cli and read through the tests to check on things.

closes #1888


imap_processing/lo/l2/lo_l2.py
```diff
@@ -1,213 +1,694 @@
...
     attr_mgr = ImapCdfAttributes()
     attr_mgr.add_instrument_global_attrs(instrument="lo")
     attr_mgr.add_instrument_variable_attrs(instrument="enamaps", level="l2-common")
     attr_mgr.add_instrument_variable_attrs(instrument="enamaps", level="l2-rectangular")
 
-    # if the dependencies are used to create Annotated Direct Events
-    if "imap_lo_l1c_pset" in sci_dependencies:
-        logical_source = "imap_lo_l2_l090-ena-h-sf-nsp-ram-hae-6deg-3mo"
-        psets = sci_dependencies["imap_lo_l1c_pset"]
+    # Add global and variable attributes
```
tmplummer, line 145: This looks like it might duplicate some of the functionality in `RectangularSkyMap.build_cdf_dataset()`. Was it a conscious decision that you made to call `sky_map.to_dataset()` before adding geometric factors and calculating intensities? I haven't dug in to understand if there is a reason to do it in that order but you might be able to have the `add_geometric_factors()` and `calculate_all_rates_and_intensities()` functions operate on the `sky_map.data_1d` dataset and then leverage the `build_cdf_dataset()` method.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -1,213 +1,694 @@
...
+    # Drop the esa_energy_step coordinate first to avoid conflicts
+    if "esa_energy_step" in pset_renamed.variables:
+        pset_renamed = pset_renamed.drop_vars("esa_energy_step")
+
+    # Ensure the pset energy coordinates match the output map
+    if "energy" in output_map.data_1d.dims:
+        # Get the energy coordinates from the output map
+        map_energy_coords = output_map.data_1d.coords.get("energy", range(7))
+        # Align the pset energy coordinates to match the map
+        pset_renamed = pset_renamed.assign_coords(energy=map_energy_coords)
```
tmplummer, line 271: Does this actually re-align the data? Or is it assumed that the PSET will always contain energy steps 1-7 and those will always map to the map energy bins directly?

tmplummer (commented):
This all looks reasonable to me. I didn't review the tests at all. The organization of the processing steps look great. Just two questions to look at.

imap_processing/lo/l2/lo_l2.py
greglucas: Yeah, I couldn't get that to work for some reason, and @sdhoyt also had some comments in there previously about needing to add his own `add_attributes` function.

I think the issue is that the `build_cdf_dataset()` class doesn't like variables that aren't a function of angles if they aren't coordinates. Our geometric factors here are functions of `(epoch, energy)` and that was causing a reshaping error when I tried to use that. So my approach was to go to a 2d dataset sooner in the pipeline and then add geometric factors. Now that I'm writing this out, I am wondering if I could have called `build_cdf_dataset()` here and then had some automatic attributes and just added the few extras I needed at the end... I'll open a ticket for this, since I think it may be something to look into, but not critical for now.

tmplummer (approved):
LGTM

imap_processing/lo/l2/lo_l2.py
greglucas: I'm not sure "align" is the best word choice by me here. We are really just trying to rename the dimension from `esa_energy_step` which was in the pset to `energy` here which is what the map l2's want (need?). For some reason I also needed to reset the coordinates. Renaming the dimension was much more complicated than I was hoping... Maybe I am missing something though.

The energy_coords are a bit misleading, they are just energy "levels". There are separate variables for `h_energy` and `o_energy` that are functions of `(epoch, energy)` because those have different central energy channels apparently (obtained from the GF file).

imap_processing/lo/l2/lo_l2.py
sdhoyt: I don't think I was aware of that `build_cdf_dataset()`  method when I wrote the first L2 code. There are/were some naming differences between Lo and the other ENAs and, at least at the time, some fields that Lo was using that Hi wasn't, so that was probably where my comment came in about adding my own attributes.

PR #2184, opened 2025-09-05
Add Lo l2 backgrounds

### LINKED ISSUE #2181: Lo: Add background rates to the maps
### Summary

Currently, we only have intensities being projected, we need to do the same for background from the pointing set.

This will involve creating some new intermediate variables for projection as well to be cleaned up after the projection.

### Related tickets

#1888

greglucas:
# Change Summary

## Overview

This is a continuation of #2180 that added intensities. It adds in the backgrounds to the map file now following equations 6-8. The second commit is the only one new here.

There are some questions still:
* What do we want to name these quantities? `background_rates` following L1c or `bg_rate` following other L2 maps.
* Need to verify the equations and whether any geometric factors need to be applied. The quantities are called "rates" in the pset, but the equations use "J" which is intensity in the other reference frames. When do we need to apply that transformation?
* The output file from the cli does not have the proper dimension labels because the attributes need to be added still. I output a netcdf file and all looks good there to me.

closes #2181

### DIFF
--- a/imap_processing/lo/l2/lo_l2.py
+++ b/imap_processing/lo/l2/lo_l2.py
@@ -270,6 +270,16 @@ def normalize_pset_coordinates(
         # Align the pset energy coordinates to match the map
         pset_renamed = pset_renamed.assign_coords(energy=map_energy_coords)
 
+    # Rename the background rates variables for the maps
+    # TODO: Do we want to change either one to be consistent across all levels?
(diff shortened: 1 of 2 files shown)


tmplummer (approved):
Changes in the second commit look good to me. I think that ultra may now have backgrounds in their Map products so coordinating the name with Ultra is my only feedback on the naming.

PR #2188, opened 2025-09-05
DPS: fixing DPS code to match latest change in infrastructure

### LINKED ISSUE #2142: SPICE - Finalize poining kernel generation job
### Algorithm Description:

We need to go through all of the details for the DPS kernel creation job and test. This ticket captures that effort and any changes that are required.

tech3371:
# Change Summary
close #2142 

## Overview

Changes related to DPS job CLI inputs changes. And changes related to upload of SPICE file from batch job.

## Updated Files

- imap_processing/cli.py
   - CLI command fixes
   - post processing upload fixes
- updated file 2

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -24,7 +24,8 @@ import imap_data_access
 import numpy as np
 import spiceypy
 import xarray as xr
-from imap_data_access import ScienceFilePath
+from imap_data_access import ScienceFilePath, SPICEFilePath
+from imap_data_access.file_validation import generate_imap_file_path
 from imap_data_access.io import download
 from imap_data_access.processing_input import (
     ProcessingInputCollection,
@@ -407,32 +408,35 @@ class ProcessInstrument(ABC):
             A list of file paths to upload to the SDC.
         """
         if self.upload_to_sdc:
-            # Validate that the files don't already exist
-            for filename in products:
-                file_path = ScienceFilePath(filename)
-                existing_file = imap_data_access.query(
-                    instrument=file_path.instrument,
-                    data_level=file_path.data_level,
-                    descriptor=file_path.descriptor,
-                    start_date=file_path.start_date,
-                    end_date=file_path.start_date,
-                    repointing=file_path.repointing,
-                    version=file_path.version,
-                    extension="cdf",
-                    table="science",
-                )
-                if existing_file:
-                    raise ProcessInstrument.ImapFileExistsError(
-                        f"File {filename} already exists in the IMAP SDC. "
-                        "No files were uploaded."
-                        f"Generated files: {products}."
-                    )
-
-            if len(products) == 0:
+            if not products:
                 logger.info("No files to upload.")
+                return
+
             for filename in products:
-                logger.info(f"Uploading file: {filename}")
-                imap_data_access.upload(filename)
+                file_obj = generate_imap_file_path(filename)
+
+                if isinstance(file_obj, SPICEFilePath):
+                    # upload SPICE files (DPS)
+                    logger.info(f"Uploading SPICE file: {filename}")
+                    try:
+                        imap_data_access.upload(filename)
+                    except ProcessInstrument.ImapFileExistsError as e:
+                        logger.error(f"Failed to upload SPICE file {filename}: {e}")
+                        # Continue to upload next file
+                        continue
+
+                elif isinstance(file_obj, ScienceFilePath):
+                    logger.info(f"Uploading science file: {filename}")
+                    try:
+                        imap_data_access.upload(filename)
+                    except ProcessInstrument.ImapFileExistsError as e:
+                        logger.error(f"Failed to upload science file {filename}: {e}")
+                        # Continue to upload next file
+                        continue
+                else:
+                    logger.info(
+                        f"File type for {filename} is not supported for upload."
+                    )
 
     @final
     def process(self) -> None:
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -420,9 +420,9 @@ class ProcessInstrument(ABC):
                         logger.warning("Skipping upload of existing file, %s", filename)
                         continue
                     else:
-                        raise ValueError(f"Upload failed with error: {msg}") from e
+                        logger.error(f"Upload failed with error: {msg}")
                 except Exception as e:
-                    raise ValueError(f"Upload failed unknown error: {e!r}") from e
+                    logger.error(f"Upload failed unknown error: {e}")
 
     @final
     def process(self) -> None:
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -1263,7 +1263,6 @@ class Spacecraft(ProcessInstrument):
             raise NotImplementedError(
                 f"Spacecraft processing not implemented for level {self.data_level}"
             )
-        return []
 
 
 class Swapi(ProcessInstrument):
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -1239,9 +1239,8 @@ class Spacecraft(ProcessInstrument):
                     data_type=SPICESource.SPICE.value
                 )
                 ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
-                # TODO: refactor if multiple attitude history files are found
                 pointing_kernel_paths = (
-                    pointing_frame.generate_pointing_attitude_kernel(ah_paths[0])
+                    pointing_frame.generate_pointing_attitude_kernel(ah_paths[-1])
                 )
                 processed_dataset.extend(pointing_kernel_paths)
         else:
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -1223,26 +1223,25 @@ class Spacecraft(ProcessInstrument):
         """
         print(f"Processing Spacecraft {self.data_level}")
         processed_dataset = []
-        if self.data_level == "l1a":
-            if self.descriptor == "quaternions":
-                # File path is expected output file path
-                input_files = dependencies.get_file_paths(source="spacecraft")
-                if len(input_files) > 1:
-                    raise ValueError(
-                        f"Unexpected dependencies found for Spacecraft L1A: "
-                        f"{input_files}. Expected only one dependency."
-                    )
-                datasets = list(quaternions.process_quaternions(input_files[0]))
-                processed_dataset.extend(datasets)
-            elif self.descriptor == "pointing-attitude":
-                spice_inputs = dependencies.get_file_paths(
-                    data_type=SPICESource.SPICE.value
+        if self.descriptor == "quaternions":
+            # File path is expected output file path
+            input_files = dependencies.get_file_paths(source="spacecraft")
+            if len(input_files) > 1:
+                raise ValueError(
+                    f"Unexpected dependencies found for Spacecraft L1A: "
+                    f"{input_files}. Expected only one dependency."
                 )
-                ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
-                pointing_kernel_paths = (
-                    pointing_frame.generate_pointing_attitude_kernel(ah_paths[-1])
-                )
-                processed_dataset.extend(pointing_kernel_paths)
+            datasets = list(quaternions.process_quaternions(input_files[0]))
+            processed_dataset.extend(datasets)
+        elif self.descriptor == "pointing-attitude":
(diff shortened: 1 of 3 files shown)


imap_processing/cli.py
```diff
@@ -407,32 +408,35 @@ def upload_products(self, products: list[Path]) -> None:
             A list of file paths to upload to the SDC.
         """
         if self.upload_to_sdc:
-            # Validate that the files don't already exist
-            for filename in products:
-                file_path = ScienceFilePath(filename)
-                existing_file = imap_data_access.query(
```
tech3371, line 413: @maxinelasp I removed query lines with try and except statement below. I remember you needed this but wondering it new changes addresses what you needed.

imap_processing/cli.py
```diff
@@ -407,32 +408,35 @@ def upload_products(self, products: list[Path]) -> None:
...
+            if not products:
                 logger.info("No files to upload.")
+                return
+
             for filename in products:
-                logger.info(f"Uploading file: {filename}")
-                imap_data_access.upload(filename)
+                file_obj = generate_imap_file_path(filename)
+
+                if isinstance(file_obj, SPICEFilePath):
```
maxinelasp: Why is this functionality needed in this CLI?

greglucas: Do we need the different if blocks, or can we just try and upload regardless of what comes in?

The final else block here should probably raise or log a warning/error saying someone tried to upload something that doesn't work. Could you even bring this into the try/except?

imap_processing/cli.py
```diff
@@ -1234,30 +1238,27 @@ def do_processing(
...
-            )
-            ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
-            if len(ah_paths) != 1:
-                raise ValueError(
-                    f"Unexpected spice dependencies found for Spacecraft "
-                    f"pointing_kernel: {ah_paths}. Expected exactly one "
-                    f"attitude history file."
+                print("inputs: ", spice_inputs)
+                ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
+                print("attitude history paths: ", ah_paths)
```
greglucas: Extra debug prints left?

imap_processing/cli.py
maxinelasp: I think it was just to save the uploading attempt, but I don't remember why we needed that. Probably it's fine either way.

imap_processing/cli.py
```diff
@@ -407,32 +408,35 @@ def upload_products(self, products: list[Path]) -> None:
...
+
+                elif isinstance(file_obj, ScienceFilePath):
+                    logger.info(f"Uploading science file: {filename}")
+                    try:
+                        imap_data_access.upload(filename)
+                    except ProcessInstrument.ImapFileExistsError as e:
+                        logger.error(f"Failed to upload science file {filename}: {e}")
+                        # Continue to upload next file
+                        continue
+                else:
```
maxinelasp: I actually do need the capability to upload ancillary files to the SDC... I guess they've been quietly getting skipped this whole time and I didn't notice 😬

imap_processing/cli.py
```diff
@@ -407,32 +408,35 @@ def upload_products(self, products: list[Path]) -> None:
...
+                    logger.info(
+                        f"File type for {filename} is not supported for upload."
+                    )
```
maxinelasp: Why not make this totally generic:
```suggestion
                file_obj = generate_imap_file_path(filename)
                logger.info(f"Uploading {type(file_obj).__name__} object")
                try:
                    imap_data_access.upload(filename)
                except ProcessInstrument.ImapFileExistsError as e:
                    logger.error(f"Failed to upload file {filename}: {e}")
                    continue
```

Is there any reason why we can't just let the CLI upload any of the `ScienceFilePath, AncillaryFilePath, SPICEFilePath, QuicklookFilePath, DependencyFilePath` filetypes?

maxinelasp: Like, I wouldn't expect this code to upload quicklooks or dependency files, but is it that big a deal if it does?

imap_processing/cli.py
```diff
@@ -1234,30 +1238,27 @@ def do_processing(
...
+                ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
+                print("attitude history paths: ", ah_paths)
+                pointing_kernel_paths = (
+                    pointing_frame.generate_pointing_attitude_kernel(ah_paths[0])
                 )
-            pointing_kernel_paths = pointing_frame.generate_pointing_attitude_kernel(
-                ah_paths[0]
-            )
-            return pointing_kernel_paths
+                return pointing_kernel_paths
```
maxinelasp: Could we put the return at the bottom of the function? Nitpick...

maxinelasp (changes requested):

imap_processing/cli.py
maxinelasp: I understand now that I read further. I don't mind it being included

imap_processing/cli.py
maxinelasp: You also might want to expand the exception class here to cover any kind of generic "file couldn't upload" error, not just the file existing one.

tech3371: since `post_processing ` is only used in processing, we won't upload quicklooks or dependency file. So consolidating sounds ok to me. Sounds like you and Greg are on same page about just having one try/except. will do that.

greglucas: I think that Menlo will want to upload ancillary files from their processing as well.
https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/266

I have thought about doing quicklooks here for speed as well since those are high priority again apparently :( So keeping it open wider is appealing to me for all those reasons.

imap_processing/cli.py
```diff
@@ -407,32 +406,23 @@ def upload_products(self, products: list[Path]) -> None:
...
+                    imap_data_access.upload(filename)
+                except IMAPDataAccessError as e:
+                    msg = str(e)
+                    if "FileAlreadyExists" in msg and "409" in msg:
+                        logger.warning("Skipping upload of existing file, %s", filename)
+                        continue
+                    else:
+                        raise ValueError(f"Upload failed with error: {msg}") from e
+                except Exception as e:
+                    raise ValueError(f"Upload failed unknown error: {e!r}") from e
```
greglucas: I didn't think @maxinelasp ever wanted these to raise / fail? i.e. if there is a bad one somewhere in here, we don't want to not upload the others.

Can you change this to log an error/warning and continue on instead of raising?

greglucas (approved):

imap_processing/cli.py
tech3371: Interesting. so I understand this `if there is a bad one somewhere in here, we don't want to not upload the others.`  meaning don't upload others. That's why I added the raise to not upload others. Did I misunderstand?

tech3371: I am happy to change it to warning but little concerns with not knowing if we produced invalid file or because of different error.

greglucas: The double negative in there, sorry!

If we have 4 files to upload and the 1st one succeeds, but the second one fails, we still want to try and upload the 3rd and 4th is my understanding.

Feel free to put a `logging.error` in there. Maybe we should have a bug-snag checker on our logs that could notify us of those without crashing our apps. But that is not for now.

tech3371: ok. That makes sense. Will change to logger.error instead.

maxinelasp (approved):

imap_processing/spice/pointing_frame.py
```diff
@@ -208,9 +208,12 @@ def calculate_pointing_attitude_segments(
     # Check that the last loaded kernel matches it input kernel name. This ensures
     # that this CK take priority when computing attitude for it's time coverage.
     count = spiceypy.ktotal("ck")
+    # TODO: fix this for when we get multiple CK kernels
```
tmplummer: Reviewing this code reminded me of why I implemented it this way. We want 1-to-1 between the attitude history files and the pointing frame kernels. This implementation checks that the last loaded CK (attitude) kernel matches the name of the file input to this function. That is required in order to get that 1-to-1 correspondence. It is fine to have multiple CKs furnished, but we want to guarantee this match.

imap_processing/cli.py
```diff
@@ -407,32 +406,23 @@ def upload_products(self, products: list[Path]) -> None:
...
             for filename in products:
-                logger.info(f"Uploading file: {filename}")
-                imap_data_access.upload(filename)
+                try:
+                    logger.info(f"Uploading file: {filename}")
+                    imap_data_access.upload(filename)
+                except IMAPDataAccessError as e:
+                    msg = str(e)
+                    if "FileAlreadyExists" in msg and "409" in msg:
+                        logger.warning("Skipping upload of existing file, %s", filename)
```
tmplummer, line 420: Is this really just a warning? Do we have cases where we are OK producing a file that has already been uploaded to the SDC? I'm worried that this condition could imply bigger problems and bite us later when we realize a bunch of jobs have been running but not producing updated products as expected.

imap_processing/cli.py
```diff
@@ -1232,36 +1222,33 @@ def do_processing(
...
-            ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
-            if len(ah_paths) != 1:
-                raise ValueError(
-                    f"Unexpected spice dependencies found for Spacecraft "
-                    f"pointing_kernel: {ah_paths}. Expected exactly one "
-                    f"attitude history file."
+                ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
+                # TODO: refactor if multiple attitude history files are found
+                pointing_kernel_paths = (
+                    pointing_frame.generate_pointing_attitude_kernel(ah_paths[0])
```
tmplummer: I think that it is OK to get rid of the `raise ValueError` but we want to pass the last ah file path to the job.
```suggestion
                    pointing_frame.generate_pointing_attitude_kernel(ah_paths[-1])
```

imap_processing/cli.py
```diff
@@ -1232,36 +1222,33 @@ def do_processing(
...
-                data_type=SPICESource.SPICE.value
-            )
-            ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
-            if len(ah_paths) != 1:
-                raise ValueError(
-                    f"Unexpected spice dependencies found for Spacecraft "
-                    f"pointing_kernel: {ah_paths}. Expected exactly one "
-                    f"attitude history file."
+                ah_paths = [path for path in spice_inputs if ".ah" in path.suffixes]
+                # TODO: refactor if multiple attitude history files are found
```
tmplummer: I think that we can remove this TODO. It is OK if multiple `.ah` files are found but we want to generate the pointing kernel for the last file furnished which in theory is the one that triggered the job to run.

imap_processing/cli.py
```diff
@@ -1232,36 +1222,33 @@ def do_processing(
...
-
+        processed_dataset = []
         if self.data_level == "l1a":
-            # File path is expected output file path
-            input_files = dependencies.get_file_paths(source="spacecraft")
-            if len(input_files) > 1:
-                raise ValueError(
-                    f"Unexpected dependencies found for Spacecraft L1A: "
-                    f"{input_files}. Expected only one dependency."
+            if self.descriptor == "quaternions":
```
tmplummer: The quaternion job and the pointing-attitude job will never be triggered together so I am not sure I understand the reason for removing the direct return statements and instead extending the `processed_data` list to be returned at the end. I think keeping the return statements makes it clearer that only the products from that one job get returned at one time.

tmplummer (commented):
Couple of comments to review and respond to.

imap_processing/cli.py
tech3371: This is just a warning. Eg. MAG instrument could produce same file in different jobs because of soft dependency. This logs that upload failed happened and continues to upload others processed file still.

imap_processing/cli.py
tech3371: We need to distinguish because both are l1a but product name (descriptor) is different. Because of that, we need to add this. We query and call different processing function because of different descriptor.

imap_processing/cli.py
tech3371: yep. That fixed the concerns. Thank you!

tmplummer (approved):

imap_processing/cli.py
tech3371: I see what you mean now. No need to check for l1a. Just check descriptor. Got it!

PR #2189, opened 2025-09-05
Update frame transforms

### LINKED ISSUE #2111: MAG l1d l2 - Use correct instrument frame for geometry
### Algorithm Description:

A new version of the instrument frame kernel changed from having just one MAG frame to having 3. IMAP_MAG_BOOM, IMAP_MAG_I, and IMAP_MAG_O. The correct instrument frame needs to be used to transform MAG data into the s/c reference frames.

From slack convo with @maxinelasp about this:

> Tim Plummer
> At what level of MAG processing do you do that transform?
> 
> Maxine Hartnett
> Just in L1D and L2. I think it will be a simple change, if I set the frame specifically when creating the class instead of defaulting to IMAP_MAG, we can specify MAGO/MAGI at that point and then everything should just work
> 
> Tim Plummer
> Do you think that this will impact any validation for those products?
> 
> Maxine Hartnett
> Yes, probably
> L1D validation has data but isn’t complete. L2 validation is done and did use the SPICE kernels. Hopefully it’s as easy as asking the MAG team to re-generate the validation data, since they used the SPICE kernels we had.

maxinelasp:
# Change Summary
Follow on to #2177 and completes #2111 

Change all references to IMAP_MAG to either IMAP_MAG_O or IMAP_MAG_I and add a test to verify. 

Completes MAG L1D and MAG L2 changes for this update.

### DIFF
--- a/imap_processing/mag/l1d/mag_l1d_data.py
+++ b/imap_processing/mag/l1d/mag_l1d_data.py
@@ -166,6 +166,9 @@ class MagL1d(MagL2L1dBase):  # type: ignore[misc]
             The day we are processing, in np.datetime64[D] format. This is used to
             truncate the data to exactly 24 hours.
         """
+        # The main data frame is MAGO, even though we have MAGI data included.
+        self.frame = ValidFrames.MAGO
+
         # set the magnitude before truncating
         self.magnitude = np.zeros(self.vectors.shape[0], dtype=np.float64)  # type: ignore[has-type]
         self.truncate_to_24h(day)
@@ -272,8 +275,19 @@ class MagL1d(MagL2L1dBase):  # type: ignore[misc]
         end_frame : ValidFrames
             The frame to rotate to. Should be one of the ValidFrames enum.
         """
+        # If we're ever in a state where we rotated into the MAGI frame, update the
+        # start frame to be MAGO instead
+        if self.frame == ValidFrames.MAGI:
+            self.frame = ValidFrames.MAGO
+
         start_frame = self.frame
+
         super().rotate_frame(end_frame)
+        # If we were in MAGO frame, we need to rotate MAGI vectors from MAGI to
+        # end_frame
+        if start_frame == ValidFrames.MAGO:
+            start_frame = ValidFrames.MAGI
+
         self.magi_vectors = frame_transform(
             self.magi_epoch,
             self.magi_vectors,
--- a/imap_processing/mag/l1d/mag_l1d_data.py
+++ b/imap_processing/mag/l1d/mag_l1d_data.py
@@ -1,6 +1,7 @@
 # mypy: disable-error-code="unused-ignore"
 """Data classes for MAG L1D processing."""
 
+import logging
 from dataclasses import InitVar, dataclass
 
 import numpy as np
@@ -14,7 +15,9 @@ from imap_processing.mag.l2.mag_l2 import retrieve_matrix_from_l2_calibration
 from imap_processing.mag.l2.mag_l2_data import MagL2L1dBase, ValidFrames
 from imap_processing.spice import spin
 from imap_processing.spice.geometry import frame_transform
-from imap_processing.spice.time import ttj2000ns_to_met
+from imap_processing.spice.time import ttj2000ns_to_et, ttj2000ns_to_met
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -272,8 +275,19 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
         end_frame : ValidFrames
             The frame to rotate to. Should be one of the ValidFrames enum.
         """
+        # If we're ever in a state where we rotated into the MAGI frame, update the
```
laspsandoval: Wait, say it a different way please. This is confusing.

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -272,8 +275,19 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
...
+        # start frame to be MAGO instead
+        if self.frame == ValidFrames.MAGI:
+            self.frame = ValidFrames.MAGO
+
         start_frame = self.frame
+
         super().rotate_frame(end_frame)
+        # If we were in MAGO frame, we need to rotate MAGI vectors from MAGI to
```
laspsandoval, line 303: Same confusion :)

laspsandoval (commented):
A couple of questions. I also checked to make certain this doesn't affect I-ALiRT and it doesn't. Hooray! Nice work. I

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: This function can be called from any frame, to any frame. So if for some reason, we're in the MAGI frame, we need to fix that bug because the data is MAGO data and should never end up in the MAGI frame.

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: This class contains both MAGO and MAGI data. We're uisng self.frame=MAGO to indicate "all data is in the instrument reference frame", which means the self.vectors are rotated from the MAGO frame and the self.magi_vectors need to be rotated from the MAGI frame.

imap_processing/mag/l1d/mag_l1d_data.py
tmplummer: It seems like there is lots going on behind the scenes in this class. I agree with Laura that this is quite confusing. If `self.frame` shouldn't ever be `MAGI` shouldn't you be raising an exception here?

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -272,8 +275,19 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
...
             The frame to rotate to. Should be one of the ValidFrames enum.
         """
+        # If we're ever in a state where we rotated into the MAGI frame, update the
+        # start frame to be MAGO instead
+        if self.frame == ValidFrames.MAGI:
+            self.frame = ValidFrames.MAGO
+
         start_frame = self.frame
+
         super().rotate_frame(end_frame)
```
tmplummer: Reading the implementation of `super().rotate_frame()`, it only rotates `self.vectors` which is only MAGO vectors? Then below, `self.magi_vectors` are rotated by calling `frame_transform` directly. Would it help with clarity to just rotate the MAGO vectors directly in this function?

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -272,8 +275,19 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
...
+        # If we were in MAGO frame, we need to rotate MAGI vectors from MAGI to
+        # end_frame
+        if start_frame == ValidFrames.MAGO:
```
tmplummer, line 305: Doesn't the preceding code block:
```
if self.frame == ValidFrames.MAGI:
            self.frame = ValidFrames.MAGO

        start_frame = self.frame
```
guarantee this to be true? Why not just hard code the from frame below like:
```
        self.magi_vectors = frame_transform(
            self.magi_epoch,
            self.magi_vectors,
            from_frame=ValidFrames.MAGI.value,
            to_frame=end_frame.value,
        )
```

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -272,8 +275,19 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
...
+            start_frame = ValidFrames.MAGI
+
         self.magi_vectors = frame_transform(
             self.magi_epoch,
```
tmplummer: Double checking that `magi_epoch` is in ET since I see in `rotate_vectors()`, you set `self.epoch_et = ttj2000ns_to_et(self.epoch)`

tmplummer (commented):
After making the below comments, it occurred to me that maybe the problem is that the singular class instance attribute `self.frame` cannot handle the reality that when the vectors are in their native frames, there are actually 2 different frames, one for `self.vectors` and one for `self.magi_vectors`. If that is true, maybe adding another attribute that contains the `magi_frame` would clean things up? Just a suggestion from my incomplete picture of this class.

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: Yeah, that makes sense. Let me rewrite the function to just fully override this.

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: Great catch, that was a bug I fixed after the fact and I missed it in here

laspsandoval (approved):
Makes sense to me. Nice work

PR #2190, opened 2025-09-06
MNT: Convert multi-species map to single-species

greglucas:
Allow only one species to be present per map product. i.e. we are separating out the h_counts and o_counts into different data files now rather than combining them into the same file with different variable names. This will ensure consistency for all of the imagers on IMAP.

* Update the attribute descriptor field to use the same updating that Ultra is using, so we can pass in anything and update it on the fly.
* Get the species from the requested map descriptor and use that for making decisions on what to include in the datasets.

A lot of minor churn that makes this look bigger than it is. It was really just changing the naming conventions around from `h_counts`/`o_counts` to `counts` from `h_rates`/`o_rates` to `ena_rates`, etc...

### DIFF
--- a/imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
@@ -107,8 +107,9 @@ imap_lo_l1c_pset:
   Logical_source: imap_lo_l1c_pset
   Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1C Data
 
-imap_lo_l2_l090-ena-h-sf-nsp-ram-hae-6deg-3mo:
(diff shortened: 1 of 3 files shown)


tmplummer (approved):
This looks good to me.

greglucas:
@subagonsouth I just removed the extra commit and went back to the first commit here only. The second commit had an update to the "counts" variable to change that to an efficiency corrected counts. After thinking a bit more, I'm not sure that is right because counts to me means "integer value" and I read through the document and they mention applying efficiencies to rates. So I'm going to hold off on that commit and I put it on the agenda to bring up with the Lo team later. So that will be a separate commit / PR if it is needed.

greglucas:
Merging over the CoDICE failures which are being addressed elsewhere.

PR #2191, opened 2025-09-08
Lo - Minor fixes from pipeline testing L1A-L1C

sdhoyt:
# Change Summary

## Overview
Discovered some issues during pipeline testing of L1A to L1C:

- needed to convert some Path objects to strings
- get the correct goodtimes ancillary now that there are more than one being passed in for pset
- Updated column naming for latest ancillary
- H, O, U are capitalized in L1B, need to use capitals for L1B species filtering
- goodtimes anc changed descriptor name to good-times.

### DIFF
--- a/imap_processing/lo/l1c/lo_l1c.py
+++ b/imap_processing/lo/l1c/lo_l1c.py
@@ -93,7 +93,7 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
             )
         else:
             pset["esa_mode"] = xr.DataArray(
-                l1b_de["esa_mode"].values[0],
+                np.array([l1b_de["esa_mode"].values[0]]),
                 dims=["epoch"],
                 attrs=attr_mgr.get_variable_attributes("esa_mode"),
             )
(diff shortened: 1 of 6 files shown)


imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -93,7 +93,7 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
             )
         else:
             pset["esa_mode"] = xr.DataArray(
-                l1b_de["esa_mode"].values[0],
+                np.array([l1b_de["esa_mode"].values[0]]),
                 dims=["epoch"],
                 attrs=attr_mgr.get_variable_attributes("esa_mode"),
             )
```
greglucas, line 99: Can you copy this directly over from l1b to the pset? i.e. rather than creating a new dataarray. Then, it could theoretically bring everything along with it and act like a pass-through so if one place gets updated it automatically gets reflected in the second location too.

```python
pset["esa_mode"] = l1b_de["esa_mode"]
```

greglucas (approved):

imap_processing/lo/l1c/lo_l1c.py
sdhoyt: In L1B the ESA Mode is an array tied to the epoch and since L1B has an epoch for each DE, there will likely be more than one ESA mode in that array. I know this shouldn't change in a pointing, but the algorithm document says to have 1 ESA Mode for each direct event and I thought that'd be better than creating a new coordinate for a single ESA Mode. So I'm grabbing the first one out of the list to use in the pointing set (1 epoch in the pointing set), which is why I'm not using the L1B data array

greglucas: Ahh, that makes sense then! I thought it was just always a single value being passed through.

PR #2192, opened 2025-09-08
2088 hi l2   geometric factor

### LINKED ISSUE #2088: Hi L2 - geometric factor
### Algorithm Description:

Implement functions and design calibration files to for the geometric factor as described in 6.1.5 of the algorithm document.

tmplummer:
# Change Summary

## Overview
Another PR that is a bit too big here. A major part of the diff shown is just due to moving the `CalibrationProductConfig` class from hi_l1c.py into uitls.py

The major update is the addition of the geometric factor to the calibration product ancillary csv file and integration into Hi L2 processing. I also put the addition of the energy bin widths into this PR. This required a couple small changes to the RectangularSkyMap.build_cdf_dataset method.

## Updated Files
   - Don't try to rewrap non-spatial variables, such as energy_delta_plus/minus.
   - Instead of manually addint energy_delta_plus/minus variables filled with NaNs, require they are set upstream.
- imap_processing/hi/hi_l1c.py
   - Move `CalibrationProductConfig` class out of this module.
- imap_processing/hi/hi_l2.py
   - Return esa_energies_df rather than just central energy values
   - Use updated esa_energies_df to get L2 central energy coordinate and set associated energy_delat_plus/minus variables.
- imap_processing/hi/utils.py
   - Move CalibrationProductConfig class into this module
   - Update tests for above changes
- imap_processing/tests/hi/conftest.py
   - Add common fixture for locating ancillary test files
- imap_processing/tests/hi/data/l1/imap_hi_90sensor-cal-prod_20240101_v001.csv
   - Rename cal_prod_num column to calibration_prod to match CDF name
   - Add geometric_factor values
- imap_processing/tests/hi/data/l1/imap_hi_90sensor-esa-energies_20240101_v001.csv
   - Add bandpass_sigma and bandpass_fwhm columns
   - Move test coverage for CalibrationProductConfig out of this file
   - Add valid geometric_factors ancillary file to function callse
   - Add tests for CalibrationProductConfig to this file

Closes: #2088


imap_processing/ena_maps/ena_maps.py
```diff
@@ -1187,6 +1187,10 @@ def to_dataset(self) -> xr.Dataset:
         # Rewrap each data array in the data_1d to the original 2D grid shape
         rewrapped_data = {}
         for key in self.data_1d.data_vars:
+            # Don't rewrap non-spatial variables
+            if CoordNames.GENERIC_PIXEL.value not in self.data_1d[key].coords:
+                rewrapped_data[key] = self.data_1d[key]
+                continue
```
greglucas, line 1193: 🎉 I think this is what I was looking for the other day when I said I was having issues rewrapping coordinates that don't have angles. This is great!

imap_processing/ena_maps/ena_maps.py
```diff
@@ -1292,18 +1296,17 @@ def build_cdf_dataset(
...
-                    xr.full_like(cdf_ds[coord_name], np.nan),
-                    name=f"{coord_name}_delta",
-                    dims=[coord_name],
-                )
-                cdf_ds[f"{coord_name}_delta_plus"] = xr.DataArray(
-                    xr.full_like(cdf_ds[coord_name], np.nan),
-                    name=f"{coord_name}_delta",
-                    dims=[coord_name],
-                )
+                if f"{coord_name}_delta_minus" not in cdf_ds:
```
greglucas, line 1300: I thought on Lo we were going to have an `energy_stat_uncert` and I thought that the `delta_plus` / `delta_minus` could point to the same variable then? Do I need to break that up into a positive/negative range instead?

imap_processing/hi/hi_l2.py
```diff
@@ -153,14 +154,25 @@ def generate_hi_map(
...
     )
     output_map.data_1d = output_map.data_1d.rename({"esa_energy_step": "energy"})
-    output_map.data_1d = output_map.data_1d.assign_coords(energy=esa_energies)
+    output_map.data_1d = output_map.data_1d.assign_coords(
+        energy=esa_df["nominal_central_energy"].values
+    )
     # Set the energy_step_delta values
-    # TODO: get the correct energy delta values (they are set to NaN) in
-    #    output_map.build_cdf_dataset()
+    energy_delta = 2 * esa_df["bandpass_sigma"]
```
greglucas: Why the 2 *? Could you add a quick comment about that being single-sided to two-sided or similar?

imap_processing/hi/hi_l2.py
```diff
@@ -153,14 +154,25 @@ def generate_hi_map(
...
+    output_map.data_1d["energy_delta_minus"] = xr.DataArray(
+        energy_delta,
+        name="energy_delta_minus",
+        dims=["energy"],
+    )
+    output_map.data_1d["energy_delta_plus"] = xr.DataArray(
+        energy_delta,
+        name="energy_delta_minus",
```
greglucas: Naming
```suggestion
        name="energy_delta_plus",
```

imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -683,13 +690,36 @@ def test_build_cdf_dataset_key_error(
         skymap = ena_maps.RectangularSkyMap(6, geometry.SpiceFrame.ECLIPJ2000)
         skymap.min_epoch = 10
         skymap.max_epoch = 15
+        # Test that variables with not attributes defined raise KeyError
```
greglucas:
```suggestion
        # Test that variables with no attributes defined raise KeyError
```

greglucas (approved):
A few minor issues noted, but feel free to update those locations and merge at your leisure.

imap_processing/ena_maps/ena_maps.py
tmplummer: Is statistical uncertainty the same as delta? The CATDESC for `energy_delta_minus` is: `Lower bound of energy range`. Maybe that is the same thing. According to the product definition spreadsheet. I believe that the IMWG decided that all maps would have those variables. As far as CDF and ISTP compliance, it is fine to just point the `energy` delta attributes at one variable.

imap_processing/hi/hi_l2.py
tmplummer: Haha, good catch... I think this actually needs to be updated. Dan says we want to store the half-width-half-max here.

Issue #1894, opened 2025-07-01
Lo L2 - Create ENA Flux Maps, Sputtering correction in Sun Rest Frame


sdhoyt:
For more details, see section 12.2.4.2 of the Lo algorithm document: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/146768943/IMAP+Algorithm+Document+from+Instrument+Teams

greglucas:
This was completed by #2195

PR #2195, opened 2025-09-09
Lo l2 sputtering

greglucas:
# Change Summary

## Overview

The final commit is the sputtering correction (rebased onto the single-species update).

For now, I've decided for a basic implementation of the equations and pass-through of the oxygen dataset. I will need to think about this some more in terms of how we actually want to handle the h/o datasets. The major piece here is that this adds the equations and tests to calculate what the sputtering correction terms should be.

part of: #1894

### DIFF
--- a/imap_processing/lo/l2/lo_l2.py
+++ b/imap_processing/lo/l2/lo_l2.py
@@ -597,7 +597,7 @@ def populate_geometric_factors(
 
 
 def calculate_all_rates_and_intensities(
-    dataset: xr.Dataset,
+    dataset: xr.Dataset, sputtering_correction: bool = False
 ) -> xr.Dataset:
     """
     Calculate rates and intensities with proper error propagation.
@@ -606,6 +606,9 @@ def calculate_all_rates_and_intensities(
     ----------
     dataset : xr.Dataset
         Sky map dataset with count data and geometric factors.
+    sputtering_correction : bool, optional
+        Whether to apply sputtering corrections to oxygen intensities.
+        Default is False.
 
     Returns
     -------
@@ -622,7 +625,14 @@ def calculate_all_rates_and_intensities(
     # Step 3: Calculate background rates and intensities
     dataset = calculate_backgrounds(dataset)
 
-    # Step 4: Clean up intermediate variables
+    # Optional Step 4: Calculate sputtering corrections
+    if sputtering_correction:
+        # TODO: The second dataset is for Oxygen specifically,
+        #       if we get an H dataset in, we may need to calculate
+        #       the O dataset separately before calling here.
+        dataset = calculate_sputtering_corrections(dataset, dataset)
+
+    # Step 5: Clean up intermediate variables
     dataset = cleanup_intermediate_variables(dataset)
 
     return dataset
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l2/lo_l2.py
```diff
@@ -728,6 +738,91 @@ def calculate_backgrounds(dataset: xr.Dataset) -> xr.Dataset:
...
+
+    # Equation 10
+    j_o_prime_var = (
+        o_small_dataset["ena_intensity_stat_uncert"] ** 2
+        + o_small_dataset["bg_intensity_stat_uncert"] ** 2
+    )
+
+    # NOTE: From table 2 of the mapping document, for energy level 5 and 6
+    sputter_correction_factor = xr.DataArray(
+        [0.15, 0.01], dims=["energy"], coords={"energy": [4, 5]}
```
tmplummer: Any idea if these correction factors are likely to change? Do they need to be in an ancillary calibration file?

imap_processing/lo/l2/lo_l2.py
```diff
@@ -728,6 +738,91 @@ def calculate_backgrounds(dataset: xr.Dataset) -> xr.Dataset:
...
+        and uncertainties.
+
+    Returns
+    -------
+    xr.Dataset
+        Dataset with calculated sputtering-corrected intensities and their
+        uncertainties for hydrogen and oxygen.
+    """
+    logger.info("Applying sputtering corrections to oxygen intensities")
+    # Only apply sputtering correction to esa levels 5 and 6
```
tmplummer: You might consider stating that levels 5 and 6 correspond to indices 4 and 5.

I would also recommend storing the energy indices in a variable rather than hard coding them everywhere.

imap_processing/lo/l2/lo_l2.py
tmplummer: The more I use xarray, the more I like it! I didn't know that you could provide `coords` in a mapping like this.

imap_processing/tests/lo/test_lo_l2.py
tmplummer, line 1: Wow, that is some very thorough testing.

tmplummer (approved):
LGTM. Just a couple nits that you can decide what to do with.

imap_processing/lo/l2/lo_l2.py
greglucas: I don't think so because these are from an external paper reference, but I don't know for sure. For now, I'm taking the tact of putting these factors from the document as hard-coded arrays because they are from the table, but that it wouldn't be hard to switch in the future if we do find we need to do that.

imap_processing/tests/lo/test_lo_l2.py
greglucas: Claude is really helping me out on the tests! Honestly, it is a bit duplicative for my liking, so maybe in the future we could come back in and clean things up, but it did quite a good job with coverage and edge cases and much faster than what I could write.

imap_processing/lo/l2/lo_l2.py
greglucas: Good call, updated this!

greglucas:
Merging over the known failure again to not have to rerun the tests.

PR #2197, opened 2025-09-09
Lo l2 bootstrap

### LINKED ISSUE #1895: Lo L2 - ENA Flux Maps, Bootstrap Correction in Sun rest frame
For more details on the bootstrap correct, see 12.2.4.3 in the Lo algorithm document: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/146768943/IMAP+Algorithm+Document+from+Instrument+Teams

greglucas:
# Change Summary

## Overview

The final commit is the bootstrap correction code.

There is quite a bit of looping because there is spillover from higher energy channels to lower, so you have to compute the higher order terms first and cascade them down. There were a bunch of equations in the document that are all essentially the same, just with different indices of i, so I looped over that index as well to avoid needing to explicitly write out each individual equation.

closes #1895

### DIFF
--- a/imap_processing/lo/l2/lo_l2.py
+++ b/imap_processing/lo/l2/lo_l2.py
@@ -597,7 +597,9 @@ def populate_geometric_factors(
 
 
 def calculate_all_rates_and_intensities(
-    dataset: xr.Dataset, sputtering_correction: bool = False
+    dataset: xr.Dataset,
+    sputtering_correction: bool = False,
+    bootstrap_correction: bool = False,
 ) -> xr.Dataset:
     """
     Calculate rates and intensities with proper error propagation.
@@ -608,6 +610,9 @@ def calculate_all_rates_and_intensities(
         Sky map dataset with count data and geometric factors.
     sputtering_correction : bool, optional
         Whether to apply sputtering corrections to oxygen intensities.
+        Default is False.
+    bootstrap_correction : bool, optional
+        Whether to apply bootstrap corrections to intensities.
         Default is False.
 
     Returns
@@ -632,7 +637,11 @@ def calculate_all_rates_and_intensities(
         #       the O dataset separately before calling here.
         dataset = calculate_sputtering_corrections(dataset, dataset)
 
-    # Step 5: Clean up intermediate variables
+    # Optional Step 5: Clean up intermediate variables
+    if bootstrap_correction:
+        dataset = calculate_bootstrap_corrections(dataset)
+
+    # Step 6: Clean up intermediate variables
     dataset = cleanup_intermediate_variables(dataset)
 
     return dataset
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l2/lo_l2.py
```diff
@@ -823,6 +832,177 @@ def calculate_sputtering_corrections(
...
+    Returns
+    -------
+    xr.Dataset
+        Dataset with calculated bootstrap-corrected intensities and their
+        uncertainties for hydrogen.
+    """
+    logger.info("Applying bootstrap corrections")
+
+    # Table 3 bootstrap terms h_i,k
+    bootstrap_factor = np.array(
```
tmplummer, line 857: Same comment as the one I left for the sputtering correction. Is it worth putting this in an ancillary file?

imap_processing/lo/l2/lo_l2.py
```diff
@@ -823,6 +832,177 @@ def calculate_sputtering_corrections(
...
+
+    # Equation 17-23
+    # E8 is a virtual energy channel at 2.1 * E7
+    e8 = 2.1 * dataset["energy"].values[-1]
+
+    # Check for valid energy ratios and j_c_prime values for gamma calculation
+    j_c_5 = j_c_prime.isel(energy=5)
+    j_c_6 = j_c_prime.isel(energy=6)
+    e_5 = dataset["energy"].isel(energy=5)
+    e_6 = dataset["energy"].isel(energy=6)
```
tmplummer: Double check these indices. Isn't `isel(energy=6)` getting you `e_7`?

imap_processing/lo/l2/lo_l2.py
```diff
@@ -823,6 +832,177 @@ def calculate_sputtering_corrections(
...
+
+    # Handle division by zero and invalid ratios
+    ratio_valid = (j_c_6 > 0) & (j_c_5 > 0) & (e_6 > 0) & (e_5 != e_6)
+
+    if np.any(ratio_valid.values):
+        gamma = np.log(j_c_5 / j_c_6) / np.log(e_5 / e_6)
```
tmplummer: Put this in a `with np.errstate(divide='ignore'):`?

imap_processing/lo/l2/lo_l2.py
```diff
@@ -823,6 +832,177 @@ def calculate_sputtering_corrections(
...
+        # Handle NaN or infinite gamma values
+        gamma = gamma.where(np.isfinite(gamma), 0)
+        j_8_b = j_c_6 * (e8 / e_6) ** gamma
```
tmplummer: This should also be in the above recommended ignore divide by zero block.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -823,6 +832,177 @@ def calculate_sputtering_corrections(
...
+            "Cannot calculate valid gamma for E8 virtual channel, setting j_8_b to zero"
+        )
+        j_8_b = xr.zeros_like(j_c_6)
+
+    # Initialize bootstrap intensity and uncertainty arrays
+    dataset["bootstrap_intensity"] = xr.zeros_like(dataset["ena_intensity"])
+    dataset["bootstrap_intensity_stat_uncert"] = xr.zeros_like(dataset["ena_intensity"])
+    dataset["bootstrap_intensity_sys_err"] = xr.zeros_like(dataset["ena_intensity"])
+
+    for i in range(6, -1, -1):
```
tmplummer, line 912: 😱 Haha, j/k I read the overview justifying this.

tmplummer (approved):
Just a couple of minor suggestions.

imap_processing/lo/l2/lo_l2.py
greglucas: I'll defer this to later again. Agreed, it might be useful, but not a specific use-case immediately and we can change it in the future.

imap_processing/lo/l2/lo_l2.py
greglucas: Yes, confusing nomenclature with this off-by-one handling, sorry about that. Had a mix of myself and AI going at it... The logic was all correct, just confusing. I've rewritten this entire block accounting for your coments below as well. We can just do these calculations and ignore the bad values and then set them equal to zero after the loop and it simplifies a lot of this error handling.

imap_processing/lo/l2/lo_l2.py
greglucas: Yes. 👍

imap_processing/lo/l2/lo_l2.py
greglucas: Yes, 👍

imap_processing/lo/l2/lo_l2.py
greglucas: 😜

PR #2201, opened 2025-09-09
Fix bug in calculation of GLOWS spin-axis statistics

### LINKED ISSUE #2173: BUG - GLOWS L1B calculation of spin_axis_orientation
### Description of the issue

The spin axis is being taken from the DPS z-axis. This is an error due to a bad review from me (Tim). The code should really be using the z-axis of the IMAP_SPACECRAFT frame instead.

Due to recent emails regarding an error in the calculation and addition of ballast on the spacecraft which results in a true spin-axis farther from the spacecraft z-axis that desired, this may need to be updated again in the future to use a TBD frame or definition of the true spin-axis in the S/C frame.

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

tmplummer:
GLOWS L1B processing should use the spacecraft z-axis for computing the spin-axis statistics instead of DPS z-axis

# Change Summary

## Overview
When reviewing GLOWS code in order to respond to a question about how the spin-axis statistics were calculated, I found this bug in which z-axis was being used. For the L1b code, GLOWS wants to use the spacecraft z-axis instead of the DPS z-axis. This will capture any nutation, precession, or coning in the statistics. The DPS z-axis is an averaged spin-axis and thus would have stddev of 0 over a pointing.

## Updated Files
- updated file 1
   - Use the IMAP_SPACECRAFT z-axis instead of the IMAP_DPS z-axis for the spin-axis
   - Fix typo in comment

Closes: #2173

### DIFF
--- a/imap_processing/glows/l1b/glows_l1b_data.py
+++ b/imap_processing/glows/l1b/glows_l1b_data.py
@@ -851,7 +851,7 @@ class HistogramL1B:
             geometry.frame_transform(
                 time_range,
                 np.array([0, 0, 1]),
-                SpiceFrame.IMAP_DPS,
+                SpiceFrame.IMAP_SPACECRAFT,
                 SpiceFrame.ECLIPJ2000,
             )
         )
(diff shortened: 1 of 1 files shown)


greglucas (approved):

tech3371 (approved):
Thank you for doing these!

PR #2208, opened 2025-09-10
PERF: Vectorization k-loop in bootstrap calculations

### LINKED ISSUE #2207: Lo bootstrap correction add more vectorization
I think there is a possibilty to add more vectorization to the bootstrap algorithm. We are currently doing two for-loops, one over i and one over k. I think that the second loop could actually be vectorized with numpy.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/30df4c47118a92b9e4e3018ee4ea7dde1bad8662/imap_processing/lo/l2/lo_l2.py#L927-L939

Those are written as summations in the document, so I'm not sure why I didn't put that together, but I think we can do something like `np.sum(value[i+1:], axis=0)` to do that loop within numpy.

greglucas:
# Change Summary

## Overview

This updates the bootstrap code to do the summations as vectorizations over the k-loop. Staying in x-array and doing things over dimensions really helps here.

closes #2207

### DIFF
--- a/imap_processing/lo/l2/lo_l2.py
+++ b/imap_processing/lo/l2/lo_l2.py
@@ -853,8 +853,8 @@ def calculate_bootstrap_corrections(dataset: xr.Dataset) -> xr.Dataset:
     """
     logger.info("Applying bootstrap corrections")
 
-    # Table 3 bootstrap terms h_i,k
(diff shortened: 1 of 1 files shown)


tmplummer (approved):
Great! Hooray for xarray.

PR #2212, opened 2025-09-10
I-ALiRT - MAG CDF Bug Fix

laspsandoval:
# Change Summary

## Overview
The mag calibration cdf has the different dimensions that are labeled the same thing and it is causing problems. This fixes that issue by using the MAG tools.

## Updated Files
- parse_mag.py

## Testing
- test_parse_mag.py


laspsandoval:
Nevermind. Don't need this.

PR #2213, opened 2025-09-11
2209 bug   hi l1a de fails when no events are in data

### LINKED ISSUE #2209: BUG - Hi L1a DE fails when no events are in data
### Description of the issue

A bug in cdflib xarray_to_cdf with compression turned on results in a divide by zero error. Bug has been reported to cdflib.

https://github.com/MAVENSDC/cdflib/issues/311

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

Fix is to create a single event in such a dataset with fill values.

tmplummer:
# Change Summary

## Overview
I had been dragging my feed on handling empty DE datasets because I think the ultimate solution would be to fix cdflib's handling of empty variables. Alas, I had to fix the issue in Hi code by writing one event to the L1A and L1B files with all DE related variables containing fill values. Then in L1B and L1C I added code to handle these Fill events.

## Updated Files

   - Fix one warning about a missing depend variable
- imap_processing/hi/hi_l1a.py
   - If there are no events, create one artificial event with fill values that match the CDF FILLVAL
   - Fix one issue with MEMDMP parsing. cdflib doesn't like the ">u4" dtype, so convert to uint32.
- imap_processing/hi/hi_l1b.py
   - Add function that will check for any valid DEs in a dataset.
   - For all DE derived variables, return the filled dataarrays if there are no good DEs
- imap_processing/hi/hi_l1c.py
   - Filter out DEs with invalid trigger_id when binning counts
   - Keep object as xr.Dataset rather than converting to pd.Dataframe
   - Test new any_good_direct_events funciton
   - Add test for input dataset with single Filled DE
   - Add mocks for any_good_direct_event function where needed
   - Test input of "emtpy" l1b DE dataset
   - Update test to pass in xr.Dataset rather than pd.DataFrame
   - Add test for handling no direct events

Closes: #2209

### DIFF
--- a/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_hi_variable_attrs.yaml
@@ -252,6 +252,7 @@ hi_de_ccsds_index:
 # LABL_PTR_i expects VAR_TYPE of metadata with char data type
 hi_hist_angle_label:
   CATDESC: Angle bin centers for histogram data.
+  DEPEND_1: angle
   FIELDNAM: ANGLE
   FORMAT: A5
   VAR_TYPE: metadata
--- a/imap_processing/hi/hi_l1a.py
+++ b/imap_processing/hi/hi_l1a.py
@@ -264,25 +264,40 @@ def create_de_dataset(de_data_dict: dict[str, npt.ArrayLike]) -> xr.Dataset:
         attrs=epoch_attrs,
     )
 
-    event_met_attrs = attr_mgr.get_variable_attributes(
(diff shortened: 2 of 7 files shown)


greglucas (approved):
I hear you on not being sure where to apply these fixes. I think that Andriy did mention he also preferred for there to be a single epoch with all fill values in the dataset as well. Kind of interesting to think about what epoch to use though if you don't have any events...

Nice work figuring this all out, looks like it hit quite a few areas. I also worry that this is probably the only instrument robust to this case. I have no idea how other instruments will handle no data cases...

PR #2216, opened 2025-09-12
Lo L1B DE - Direction Vector and Pointing Bin Fixes

sdhoyt:
# Change Summary

## Overview
- Pointing bins
   - Updated pointing bins to be named off_angle_bin and spin_bin
   - Using HAE instead of DPS
- Direction Vector
   - using instrument_pointing function to get the cartesian coordinates and set variable
      - hae_x
      - hae_y
      - hae_z
- Removed spin_bin and spin_angle functions
   - spin_angle was used to get spin_bin (0-60 bins). Lo said these are no longer needed since we are using a 3600 bin spin_bin

A follow-on PR will be created to populate the badtimes variable using the bad-times ancillary file which will close out the L1B DE product.

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -15,7 +15,11 @@ from imap_processing.lo.l1b.tof_conversions import (
     TOF2_CONV,
     TOF3_CONV,
 )
-from imap_processing.spice.geometry import SpiceFrame, instrument_pointing
+from imap_processing.spice.geometry import (
+    SpiceFrame,
+    cartesian_to_latitudinal,
+    instrument_pointing,
+)
 from imap_processing.spice.repoint import get_pointing_times
 from imap_processing.spice.spin import get_spin_number
 from imap_processing.spice.time import met_to_ttj2000ns, ttj2000ns_to_et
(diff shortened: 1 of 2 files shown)


tech3371 (approved):
I don't know a lot in depth but changes looks good to me

greglucas (approved):

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -782,22 +728,31 @@ def set_pointing_direction(l1b_de: xr.Dataset) -> xr.Dataset:
     """
     # Get the pointing bin for each DE
     et = ttj2000ns_to_et(l1b_de["epoch"])
```
tmplummer, line 734: I assume that `l1b_de["epoch"]` contains the ttj2000ns times of each DE?

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -822,10 +777,14 @@ def set_pointing_bin(l1b_de: xr.Dataset) -> xr.Dataset:
     l1b_de : xarray.Dataset
         The L1B DE dataset with the pointing bins added.
     """
+    et = ttj2000ns_to_et(l1b_de["epoch"])
+
+    # get lon and lat in degrees
+    direction = instrument_pointing(et, SpiceFrame.IMAP_LO_BASE, SpiceFrame.IMAP_HAE)
```
tmplummer: Just a comment that it is likely faster to convert the Cartesian coordinates in the set_pointing_direction function to lat/lon using the geometry.cartesian_to_latitudinal function rather than making another spice call for each DE here.

tmplummer (approved):
This looks good. One recommendation for an improvement. Up to you if you want to use it or not.

frahmanif:
Looks good. Thanks!

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Correct

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Good call. I'll make that change

sdhoyt:
Thank you everyone for your reviews and help with implementation!

PR #2223, opened 2025-09-15
ULTRA L1c attr updates

lacoak21:
# Change Summary

## Overview
Change CATDESCs to reflect ULTRA IT CATDESCs. Add epoch as dim to new variables (I 

## Updated Files
   - new catdescs
- imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
   - new catdescs
- imap_processing/ultra/l1c/helio_pset.py ,  imap_processing/ultra/l1c/spacecraft_pset.py
   - add epoch as first dim. This is necessary for ultra l2 maps

### DIFF
--- a/imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
@@ -147,10 +147,10 @@ quality_flags:
   DEPEND_2: energy_bin_geometric_mean
   UNITS: " "
 
-bg_rate:
+background_rates:
   <<: *default_float32
   CATDESC: Background rates. Background dominated by accidentals caused by a combination of UV light and misregistered low energy ENA events.
-  FIELDNAM: bg_rate
+  FIELDNAM: background_rates
   LABLAXIS: background rates
   # TODO: come back to format
   UNITS: counts/second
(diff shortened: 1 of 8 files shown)


imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
```diff
@@ -27,6 +27,7 @@ default_uint16_attrs: &default_uint16
 
 energy_dependent_attrs: &energy_dependent
   <<: *default_float32
+  DEPEND_0: epoch
```
greglucas: Are these actually a function of "epoch" as well? In Lo, I was going to leave purely energy-dependent values off of "epoch" and label them support_data or similar. It seems odd to me to add the extra dimension otherwise.

lacoak21: OK you're right. I was thrown because the ULTRA team filled out the spreadsheet that these dimensions contain epoch but I just confirmed that was because it was copied over from other definitions! I will remove them. Exposure factor _should_ be pointing dependent though.

greglucas: Perfect, that is what I think as well and lines up with Lo's implementation!

lacoak21: Excellent.

lacoak21: (not to throw the ultra team under the bus I should have caught that 😄 )

laspsandoval (approved):
Thanks Luisa!

PR #2224, opened 2025-09-16
Mag l1c pre launch updates

maxinelasp:
# Change Summary
Updates L1C of MAG to manage gap filling at the beginning and end of the day.

I had Claude generate a lot of unit tests to try to figure out why the validation is not correct. Ulitmately I wasn't really able to figure it out, but I think this is closer. Hoping to come back and knock out a few additional TODOs before launch. 

Probably the unit tests don't need a close review because they are a lot of duplicates or trying to test EVERY edge case.


imap_processing/tests/mag/conftest.py
```diff
@@ -137,9 +138,13 @@ def mag_generate_l1b_from_csv(df, logical_source):
     epoch = [np.datetime64(t) - np.datetime64(TTJ2000_EPOCH) for t in df["t"]]
     epoch_ns = [(e / np.timedelta64(1, "ns")).astype(np.int64) for e in epoch]
     dataset.coords["epoch"] = xr.DataArray(epoch_ns, name="epoch", dims=["epoch"])
+    print(
+        f"Read out epoch values from CSV: {epoch_ns[0] / 1e9}s to {epoch_ns[-1] / 1e9}s"
+    )
```
tmplummer: Remove print statement?

imap_processing/mag/l1c/interpolation_methods.py
```diff
@@ -229,11 +229,13 @@ def cic_filter(
     cic1 = cic1 / decimation_factor
     cic2 = np.convolve(cic1, cic1)
     delay = (len(cic2) - 1) // 2
+
     input_filtered = input_timestamps
+    vectors_filtered = lfilter(cic2, 1, input_vectors, axis=0)
     if delay != 0:
         input_filtered = input_timestamps[:-delay]
```
tmplummer, line 236: Maybe add a comment about why input_timestamps get are removed from the end where vectors_filtered are removed from the start?

tmplummer (approved):
Looks good from a high level.

tech3371 (approved):
approving to unblock

imap_processing/mag/l1c/interpolation_methods.py
maxinelasp: It's a good question, but I don't know why. This was the only part of the algorithm where i was basically handed pseudocode.

PR #2233, opened 2025-09-17
ULTRA l2 handle pointing independent vars 

lacoak21:
# Change Summary

## Overview
I had misunderstood a conversation with the IT team (Nick) and thought that efficiency and sensitivity were pointing dependent. After a slack thread discussion about this (over 100 😄  - new record for me), I have all of the new variables set and feel confident. Since efficiency and sensitivity are no longer pointing dependent AND they need to be weighted by exposure factor, this required some slightly refactor to allow for broadcasting. 

I also misunderstood how pointing independent vars should be projected to the map. I had only pushed them once but they should also be accumulated and weighted: 
Screenshot 2025-09-17 at 3 30 55 PM

From Nick: 

"You have to sum all that stuff with exposure time (and solid angle if the maps / pixels are different sizes)
then divide out by the total
It's a weighted mean, where the exposure is the weighting factor
efficiency shouldn't change per Pointing, neither should the sensitivity.  I think the only thing that will really is the exposure time" 

## Updated Files
   - Add handling of pointing independent vars. 
   - Removed handling to only push pointing independent vars once. They should still be accumulated and weighted 

## Testing
Fix arrays to have correct dims

### DIFF
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -190,8 +190,7 @@ sensitivity:
   CATDESC: Averaged instrument sensitive area.
   FIELDNAM: Sensitivity
   UNITS: cm^2
-  DEPEND_0: epoch
-  VAR_TYPE: data
+  VAR_TYPE: support_data
   LABLAXIS: sensitivity
   DISPLAY_TYPE: image
   DICT_KEY: SPASE>Particle>ParticleType:Atom,ParticleQuantity:GeometricFactor,Qualifier:Directional,CoordinateSystemName:HAE,CoordinateRepresentation:Spherical
@@ -222,8 +221,7 @@ efficiency:
   CATDESC: Efficiency of the instrument in converting ENAs to valid events.
   FIELDNAM: Efficiency
   UNITS: " "
-  DEPEND_0: epoch
-  VAR_TYPE: data
+  VAR_TYPE: support_data
   LABLAXIS: Efficiency
   DISPLAY_TYPE: no_plot
   DICT_KEY: SPASE>Particle>ParticleType:Atom,ParticleQuantity:Other
--- a/imap_processing/ultra/l2/ultra_l2.py
+++ b/imap_processing/ultra/l2/ultra_l2.py
@@ -56,20 +56,18 @@ REQUIRED_L1C_VARIABLES_PUSH = [
 ]
 REQUIRED_L1C_VARIABLES_PULL = [
     "exposure_factor",
-    "sensitivity",
     "background_rates",
     "obs_date",
 ]
 # These variables are expected but not strictly required. In certain test scenarios,
 # they may be missing, in which case we will raise a warning and continue.
 # All psets must be consistent and either have these variables or not.
-EXPECTED_L1C_VARIABLES_PULL = [
-    "efficiency",
-]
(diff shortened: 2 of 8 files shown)


imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -190,8 +190,7 @@ sensitivity:
   CATDESC: Averaged instrument sensitive area.
   FIELDNAM: Sensitivity
   UNITS: cm^2
-  DEPEND_0: epoch
```
tmplummer, line 193: I think that there is some confusion around epoch in these L2 maps. For all of the maps, we only have epoch as a dimension because SPDF requires it. All of the maps will have an epoch dimension size of 1. I believe that is also true for Ultra PSETs. It certainly is for Hi PSETs. So, even though it is a singular dimension, I think that you want/need it on these variables.

tmplummer (commented):
Just the one comment about the epoch dimension for these L2 variables.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
greglucas: @subagonsouth, do you know if it is a requirement to have it on these variables? Lo's geometric factor is not a function of time, so it doesn't make sense to have it in there to me personally. But if it is a requirement I guess we would need to do the same for all the instruments. Lo's is slightly different though as that factor is not "projected".

lacoak21: Nick requested that these variables not contain epoch although he mentioned if it was required it would be ok. I did update them to be support_data so i was thinking that would pass SPDF requirements if I understand correctly.

tmplummer: I will leave it to SPDF to review and comment. I just wish we could remove the `epoch` dimension all together 😆 .

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -323,12 +327,11 @@ def generate_ultra_healpix_skymap(  # noqa: PLR0912
             pointing_set.data["exposure_factor"] * pointing_set.solid_angle
         )
 
-        # Get variables that should be weighted by exposure and solid angle
-        existing_vars_to_weight = []
-        for var in VARIABLES_TO_WEIGHT_BY_POINTING_SET_EXPOSURE_TIMES_SOLID_ANGLE:
-            if var in pointing_set.data:
-                existing_vars_to_weight.append(var)
-
+        # if the variable does not have an epoch dimension, add one temporarily
```
tmplummer, line 330: Ultimately, it would be nice to come up with a more general solution for this an put it in the common ena_maps code. Having to do these gymnastics to get the broadcasting to work isn't ideal. Just a though for a future improvement.

tmplummer (approved):

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
lacoak21: Me too 😄.

imap_processing/ultra/l2/ultra_l2.py
lacoak21: Agreed. Ill add a TODO in here.

PR #2236, opened 2025-09-18
2218 hi l2   integrate flux correction

### LINKED ISSUE #2218: Hi L2 - Integrate flux correction
### Algorithm Description:

Use the predictor-corrector code to make the flux correction

tmplummer:
# Change Summary

## Overview
This integrates the flux correction into the Hi L2 pipeline. As part of this, I refactored how ancillary files get passed around. In the cli `do_processing` method, I generate a dictionary that uses ancillary file descriptors as keys and the Path as the value. No attempt was made to test values produced by the added flux correction.

## Updated Files
- imap_processing/cli.py
   - Parse ancillary files into dictionary to pass into L2 processing.
- imap_processing/hi/hi_l2.py
   - Modify function signatures to pass dictionary of ancillary files around.
   - Add PowerLawFluxCorrector to final step of L2 processing.
   - Add fixtures for producing dictionary of L2 ancillary files.
   - Update tests to include esa-eta-fit_coeffs ancillary file
   - Update to L2 test to include the 3 required ancillary files.

Closes: #2218

### DIFF
--- a/imap_processing/hi/hi_l2.py
+++ b/imap_processing/hi/hi_l2.py
@@ -132,7 +132,7 @@ def generate_hi_map(
 
     output_map.data_1d.update(calculate_ena_signal_rates(output_map.data_1d))
     output_map.data_1d = calculate_ena_intensity(
-        output_map.data_1d, l2_ancillary_path_dict
+        output_map.data_1d, l2_ancillary_path_dict, descriptor
     )
 
     output_map.data_1d["obs_date"].data = output_map.data_1d["obs_date"].data.astype(
@@ -212,6 +212,7 @@ def calculate_ena_signal_rates(map_ds: xr.Dataset) -> dict[str, xr.DataArray]:
 def calculate_ena_intensity(
     map_ds: xr.Dataset,
     l2_ancillary_path_dict: dict[str, Path],
+    descriptor: MapDescriptor,
 ) -> xr.Dataset:
     """
     Calculate the ena intensities.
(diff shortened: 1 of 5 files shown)


tmplummer:
pre-commit.ci autofix

imap_processing/hi/hi_l2.py
```diff
@@ -47,29 +46,28 @@ def hi_l2(
...
             "and be either '45' or '90'"
         )
 
+    sky_map = generate_hi_map(
+        psets,
+        l2_ancillary_path_dict,
+        map_descriptor,
+    )
+    if not isinstance(sky_map, RectangularSkyMap):
+        raise NotImplementedError("Healpix map output not supported for Hi")
```
greglucas: Isn't this handled by your above nside check? It does seem better to do this right at the beginning rather than after the fact.

imap_processing/hi/hi_l2.py
```diff
@@ -268,6 +259,19 @@ def calculate_ena_intensity(
         esa_energy,
     )
 
+    # Flux correction
```
greglucas: Don't we only want to call this function if the descriptor doesn't have "raw" in it or similar? I think this needs to be gated somehow to handle maps that want this and maps that don't (unless Hi always wants the correction applied).

imap_processing/hi/hi_l2.py
tmplummer: Yep, good point. I added this because without it, one of the pre-commit checks was complaining that `AbstractSkyMap` doesn't have a `build_cdf_dataset` method. I'll figure out a better way to make the pre-commit happy here.

imap_processing/hi/hi_l2.py
tmplummer: Hi doesn't have any products defined that don't get the flux correction applied. I think that it is still a good idea to implement what you are talking about here in case there is a desire to run this code without the flux correction at some point.

greglucas: Good call, think about how you'd want to do this for Lo as well and whether that ties in here potentially because I know Lo wants this and I think it is a part of the map descriptor field for them.

greglucas (approved):
Approving as I think this is good now. Feel free to push the updates or leave them to another PR, my approval stands either way.

imap_processing/hi/hi_l2.py
greglucas: Oh yeah, I think I ran into that at some point too. Can you do something like this to indicate to the type checker what the return is expected to be?

```python
sky_map: RectangularSkyMap = processinging()
```

PR #2244, opened 2025-09-20
I-ALiRT - cdf bugfix

laspsandoval:
# Change Summary

## Overview
Realized that there were a number of items in the database that we don't necessarily want in the cdf.

## Updated Files
- create_xarray.py
   - Added keywords to prevent from those being in cdf.

### DIFF
--- a/imap_processing/ialirt/utils/create_xarray.py
+++ b/imap_processing/ialirt/utils/create_xarray.py
@@ -144,7 +144,17 @@ def create_xarray_from_records(records: list[dict]) -> xr.Dataset:  # noqa: PLR0
     # Populate the dataset variables
     for i, record in enumerate(records):
         for key, val in record.items():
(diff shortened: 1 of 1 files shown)


imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -144,7 +144,17 @@ def create_xarray_from_records(records: list[dict]) -> xr.Dataset:  # noqa: PLR0
...
     for i, record in enumerate(records):
         for key, val in record.items():
-            if key in ["apid", "met", "met_in_utc", "ttj2000ns", "last_modified"]:
+            if key in [
+                "apid",
+                "met",
+                "met_in_utc",
+                "ttj2000ns",
+                "last_modified",
+                "sc_position_GSM",
```
greglucas, line 153: I thought they did want the positions in the CDFs? They only want them in the web queries instead?

greglucas (approved):

PR #2254, opened 2025-09-29
Fix ialirt partial packet

laspsandoval:
# Change Summary

## Overview
The last packet of every I-ALiRT packets file is a partial packet. This change identifies the partial packet, logs a warning, but does not stop the code to throw an error.

## Updated Files
- utils.py
   - This change identifies the partial packet, logs a warning, but does not stop the code to throw an error.
## Testing
- Unit tests with I-ALiRT packets file

### DIFF
--- a/imap_processing/utils.py
+++ b/imap_processing/utils.py
@@ -288,9 +288,10 @@ def packet_file_to_datasets(
                         datatype_mapping[apid][key] = _get_minimum_numpy_datatype(
                             key, packet_definition, use_derived_value=use_derived_value
                         )
-        except Exception as e:
-            if isinstance(e, ValueError) and "Packet fields do not match" in str(e):
+        except ValueError as e:
+            if "Packet fields do not match" in str(e):
                 raise
+        except Exception as e:
             logger.warning(f"Truncated or malformed packet #{_i}. Skipping.")
             logger.warning(f"Details: {e}")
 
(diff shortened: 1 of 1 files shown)


imap_processing/utils.py
```diff
@@ -258,34 +258,42 @@ def packet_file_to_datasets(
...
+
+                for key, value in packet_content.items():
+                    val = value if use_derived_value else value.raw_value
+                    data_dict[apid][key].append(val)
+                    if key not in datatype_mapping[apid]:
+                        # Add this datatype to the mapping
+                        datatype_mapping[apid][key] = _get_minimum_numpy_datatype(
+                            key, packet_definition, use_derived_value=use_derived_value
+                        )
+        except Exception as e:
```
greglucas: Is there a different Exception you can catch here to not do the isinstance thing below?

imap_processing/utils.py
```diff
@@ -258,34 +258,42 @@ def packet_file_to_datasets(
 
     with open(packet_file, "rb") as binary_data:
         packet_generator = packet_definition.packet_generator(binary_data)
```
greglucas, line 260: You might be able to update this generator instantiation to skip the bad packets:
https://github.com/lasp/space_packet_parser/blob/552701b568750fd68298b66a1f813d18c43ac046/space_packet_parser/definitions.py#L481

```suggestion
        packet_generator = packet_definition.packet_generator(binary_data, parse_bad_pkts=False)
```

laspsandoval: I tried that and have the same error.

imap_processing/utils.py
laspsandoval: I made a change that I think is a bit of an improvement.

laspsandoval:
Not needed.

Issue #2237, opened 2025-09-18
CoDICE: L1A validation fixes


tech3371:
- [x] #2262
- [x] L1A lo angular sw didn't match - ["hplus", "heplusplus", "oplus6", "fe_loq"]
- [x] L1A lo nsw angular didn't match - ["heplusplus"]
- [x] #2257

- [x] #2259
> imap_processing/tests/codice/test_codice_l1a.py Variable 'voltage_table' not found in processed data.

- [x] #2258

- [x] L1A Hi sectored
- [x] #2261
- [x] #2260

tech3371:
Closing this since there is ticket for remaining work

PR #2264, opened 2025-09-29
CoDICE: Bug fix for angular and sectored L1A products

tech3371:
# Change Summary
related to #2237 

## Overview

This only fixes bugs for SW/NSW angular and hi sectored products. Remaining l1a will be fixed in their upcoming work.

## Updated Files

- imap_processing/codice/codice_l1a.py
   - fix bugs
   - update old variables and added missing information

## Testing

Fixed and refactor tests to be easier to fix in the upcoming works.


tech3371:
This is quick fix to l1a bugs. But lot of those will be refactored in upcoming work

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -1147,56 +1147,56 @@ lo-sw-angular-hplus:
...
     FIELDNAM: SW - He++
-    DEPEND_1: inst_az
-    DEPEND_2: spin_sector
-    DEPEND_3: esa_step
-    LABL_PTR_1: inst_az_label
-    LABL_PTR_2: spin_sector_label
-    LABL_PTR_3: esa_step_label
+    DEPEND_1: esa_step
+    DEPEND_2: inst_az
+    DEPEND_3: spin_sector
```
greglucas, line 1163: I thought we wanted the spin axis before the other one, but I am not sure exactly what values these are relating to here so I'm just noting this to indicate it might be worth looking into, not that it is absolutely wrong.

imap_processing/codice/constants.py
```diff
@@ -268,149 +269,158 @@
 
 OMNI_ENERGY_TABLE = {
```
greglucas, line 270: These are changing pretty significantly. Why are they different now?

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -62,33 +131,24 @@ def test_lo_ialirt():
...
-            assert processed_data[variable].shape == (1,)
-        else:
-            assert processed_data[variable].shape == (8, 128, 1)
+    for variable in val_data.data_vars:
+        if variable in EXPECTED_MISMATCHES or variable.startswith(
+            UNCERTAINTY_VARIABLES
+        ):
+            continue
+        assert processed_data[variable].shape == val_data[variable].shape, (
+            f"Shape mismatch for variable '{variable}'"
```
greglucas, line 149: Just to be clear, this is still only validating the shapes and not the values. Is that what you want for now because the values still don't match?

greglucas (approved):
Looks good for a quick fix, a few minor comments/clarifications to put on your radar.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
tech3371: you mean from project perspective, right? Looks like in SWE, I am doing this
```
(energy, spin angle, cem angle)
```
but CoDICE seems to be doing this instead. 
(energy, inst_az (instrument index), spin sector (spin angle))

darn another big change if we want to go in direction you are talking about...sigh

(energy, spin sector, inst_az)

imap_processing/codice/constants.py
tech3371: the LUT got updated. many values were added as constant which needs to change but it's going to add to existing work.

imap_processing/tests/codice/test_codice_l1a.py
tech3371: exactly!

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
tech3371: I will bring this up in today's meeting at 2pm and see how much work that will be.

tech3371: I will address this in refactor work

PR #2265, opened 2025-09-30
I-ALiRT - add mag hk status

laspsandoval:
# Change Summary

## Overview
The MAG team wants to put their HK data into the public database. I am adding it for now and will reaccess future options for extracting it out into a separate credentials.

## Updated Files
- parse_mag.py
   - Add HK data into output
- create_xarray.py
   - Do not use HK data in cdf creation


imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -719,6 +719,39 @@ def process_packet(
...
+                    "fib_saturated": int(status_data["fib_saturated"]),
+                    "mode": int(status_data["mode"]),
+                    "icu_temp": int(status_data["icu_temp"]),
+                    "hk2v5_warn": int(status_data["hk2v5_warn"]),
+                    "hk2v5_danger": int(status_data["hk2v5_danger"]),
+                    "hk2v5c_warn": int(status_data["hk2v5c_warn"]),
+                    "hk2v5c_danger": int(status_data["hk2v5c_danger"]),
+                    "hk3v3": int(status_data["hk3v3"]),
+                    "hk3v3_current": int(status_data["hk3v3_current"]),
+                    "pri_isvalid": int(status_data["pri_isvalid"]),
```
greglucas: Should any of these be booleans?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -719,6 +719,39 @@ def process_packet(
                 "mag_theta_B_GSM": Decimal(str(theta_gsm[i])),
                 "mag_phi_B_GSE": Decimal(str(phi_gse[i])),
                 "mag_theta_B_GSE": Decimal(str(theta_gse[i])),
+                "mag_hk_status": {
+                    "hk1v5_warn": int(status_data["hk1v5_warn"]),
```
greglucas: For the future, these duplicate things could be written with a dict-comprehension I think too.

```
{key: int(status_data[key]) for key in ["hk1v5_warn", "hk1v5_danger", ...]}
```

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -154,6 +154,7 @@ def create_xarray_from_records(records: list[dict]) -> xr.Dataset:  # noqa: PLR0
                 "sc_position_GSE",
                 "sc_velocity_GSM",
                 "sc_velocity_GSE",
+                "mag_hk_status",
```
greglucas, line 157: Does it need to be in the xarray CDF dataset? Maybe this line could be dropped so that it doesn't come in the CDF data, but rather only the direct API queries against the DB?

laspsandoval: Yep. These are the ones that will not be in the cdf.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: nice catch

greglucas (approved):

imap_processing/ialirt/utils/create_xarray.py
greglucas: Ahh, I misread that. Thank you for clarifying!

PR #2266, opened 2025-09-30
2253 ena maps   allow pset az el coordinates to have energy dimension

### LINKED ISSUE #2253: ENA Maps - Allow PSET az el coordinates to have energy dimension
### Algorithm Description:

For the Hi/Lo CG correction, we need to be able to bin each pset bin based on a az/el that is unique to the energy step. This ticket covers the changes needed to the mapping infrastructure to allow each pset energy level to get binned uniquely.

tmplummer:
# Change Summary

## Overview
The two changes here support the ability for PSET bins at different energy levels project (bin) into unique map pixels. This is needed for the Hi/Lo CG correction.

## Updated Files

   - Support input_obj with multi-dimensional azimuth and elevation coordinates in the `match_coords_to_indices` function. The leading dimension would correspond to energy. In this way, we can map a single pset bin to unique map bin at each energy level.
- imap_processing/ena_maps/utils/map_utils.py
   - Allow multidimensional inputs to `bin_single_array_at_indices` function and support broadcasting.
   - New `vectorized_bincount` function supports the above change. Multi-dimensional binning is supported by making it a 1D problem and adjusting the projection indices accordingly. This allows binning to be done in 1D objects which then get reshaped into 2D objects.
   - Add test coverage for changes to `match_coords_to_indices` function.
   - Add test coverage for new `vectorized_bincount` function
   - Update test coverage for `bin_single_array_at_indices` function to include many combinations of 1D and 2D inputs in order to check that broadcasting works as expected.

Closes: #2253

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -97,7 +97,7 @@ def match_coords_to_indices(
     Parameters
     ----------
     input_object : PointingSet | AbstractSkyMap
-        An object containing 1D spatial pixel centers in azimuth and elevation,
+        An object containing spatial pixel centers in azimuth and elevation,
         which will be matched to 1D indices of spatial pixels in the output frame.
         Must contain the Spice frame in which the pixel centers are defined.
     output_object : PointingSet | AbstractSkyMap
@@ -115,10 +115,13 @@ def match_coords_to_indices(
     Returns
     -------
     flat_indices_input_grid_output_frame : NDArray
-        1D array of pixel indices of the output object corresponding to each pixel in
-        the input object. The length of the array is equal to the number of pixels in
-        the input object, and may contain 0, 1, or multiple occurrences of the same
-        output index.
+        Array of pixel indices mapping each input object pixel center to a pixel
+        in the output object. If the input object has multi-dimensional coordinates
+        defines, the output indices with also be multi-dimensional. The shape of
+        the output array is (..., n) where ... matches the non-spatial dimensions
+        of the input object and n is the number of spatial pixels in the input
+        object. Output indices may contain 0, 1, or multiple occurrences of the
+        same output index.
 
     Raises
     ------
@@ -166,14 +169,14 @@ def match_coords_to_indices(
         # use ravel_multi_index to get the 1D indices of the pixels in the output frame.
         az_indices = (
             np.digitize(
-                input_obj_az_el_output_frame[:, 0],
+                input_obj_az_el_output_frame[..., 0],
                 output_object.sky_grid.az_bin_edges,
             )
             - 1
         )
         el_indices = (
             np.digitize(
-                input_obj_az_el_output_frame[:, 1],
+                input_obj_az_el_output_frame[..., 1],
                 output_object.sky_grid.el_bin_edges,
             )
             - 1
@@ -191,8 +194,8 @@ def match_coords_to_indices(
         # which directly returns the index on the output frame's Healpix tessellation.
         flat_indices_input_grid_output_frame = hp.ang2pix(
             nside=output_object.nside,
-            theta=input_obj_az_el_output_frame[:, 0],  # Lon in degrees
-            phi=input_obj_az_el_output_frame[:, 1],  # Lat in degrees
+            theta=input_obj_az_el_output_frame[..., 0],  # Lon in degrees
+            phi=input_obj_az_el_output_frame[..., 1],  # Lat in degrees
             nest=output_object.nested,
             lonlat=True,
         )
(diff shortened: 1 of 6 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -115,10 +115,13 @@ def match_coords_to_indices(
     Returns
     -------
     flat_indices_input_grid_output_frame : NDArray
-        1D array of pixel indices of the output object corresponding to each pixel in
-        the input object. The length of the array is equal to the number of pixels in
-        the input object, and may contain 0, 1, or multiple occurrences of the same
-        output index.
+        Array of pixel indices mapping each input object pixel center to a pixel
+        in the output object. If the input object has multi-dimensional coordinates
+        defines, the output indices with also be multi-dimensional. The shape of
```
greglucas:
```suggestion
        defined, the output indices with also be multi-dimensional. The shape of
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -166,14 +169,14 @@ def match_coords_to_indices(
         # use ravel_multi_index to get the 1D indices of the pixels in the output frame.
         az_indices = (
             np.digitize(
-                input_obj_az_el_output_frame[:, 0],
+                input_obj_az_el_output_frame[..., 0],
```
greglucas, line 172: Pretty cool you were able to make this pretty small change to support multi-dimensional!

imap_processing/ena_maps/utils/map_utils.py
```diff
@@ -10,6 +10,92 @@
...
+    # Each sample gets its own set of bins: sample 0 uses bins [0, minlength),
+    # sample 1 uses bins [minlength, 2*minlength), etc.
+    offsets = np.arange(n_samples).reshape(*output_shape, 1) * minlength
+    indices_offset = (indices_bc + offsets).ravel()
+
+    # Flatten weights if provided
+    if weights_bc is not None:
+        weights_flat = weights_bc.ravel()
+    else:
+        weights_flat = None
```
greglucas: Can this be brought up into the if condition above right below your broadcast arrays and then just use weights_bc below.

imap_processing/ena_maps/utils/map_utils.py
```diff
@@ -10,6 +10,92 @@
...
+
+    # Determine actual minlength if not specified
+    if minlength == 0:
+        minlength = int(np.max(indices_bc)) + 1
+
+    # Add offsets to indices to separate each sample's bins
+    # Each sample gets its own set of bins: sample 0 uses bins [0, minlength),
+    # sample 1 uses bins [minlength, 2*minlength), etc.
+    offsets = np.arange(n_samples).reshape(*output_shape, 1) * minlength
+    indices_offset = (indices_bc + offsets).ravel()
```
greglucas: indices_offset kind of confused me with the naming. Is it actually better specified as indices_flat? I understand that there are offsets, but it is really just a raveled nd componenets that you're lumping into a 1d call and then unflatting it at the end (I think?).

greglucas: May id_array which AI suggested to me... Do we need the minlength for users to pass in too that seems to add a bit of complexity?

```
import numpy as np

A = np.array([[0, 1, 0], [1, 2, 1]])
N = A.max() + 1  # Max value in A + 1
rows, cols = A.shape

# Create unique IDs for each element
# Offset each row's values by a multiple of N to distinguish them
id_array = A + (N * np.arange(rows))[:, None]

# Flatten the ID array and perform bincount
binned_flat = np.bincount(id_array.ravel(), minlength=N * rows)

# Reshape the result back to a 2D array
result = binned_flat.reshape(rows, N)
print(result)
```

greglucas (approved):
Looks good overall. Nice that this could fall out pretty easily.

imap_processing/ena_maps/utils/map_utils.py
tmplummer: I see what you mean regarding the name of `indices_offset`. `indices_flat` seems reasonable.

I'll try to take some pieces of your code to integrate and clarify comments.

`minlength` keeps the function signature the same as `np.bincount`. It also makes it so that the output from this function has the correct number of spatial pixels. Otherwise, we could end up with a binned array that doesn't have the same number of spatial pixels that we want in the map and we'd have to do extra contortions to get the binned data into the output map data arrays.

greglucas: > I'll try to take some pieces of your code to integrate and clarify comments.

Google AI's code ;)

All sounds good here. You still have my approval!

tmplummer: > ```
> # Create unique IDs for each element
> # Offset each row's values by a multiple of N to distinguish them
> id_array = A + (N * np.arange(rows))[:, None]
>```

This is a solution for one extra dimension, but Hi has arrays with 2 extra dimensions: (energy, calibration_product, spin_angle_bin) That is why I had a slightly more complicated way of computing this offset array.

PR #2268, opened 2025-10-01
ENH: Lo - Handle corrections based on map descriptor

greglucas:
Read the descriptor to determine which corrections to apply. Process the O dataset if needed for sputtering corrections.

The flow is:
If it is an ena corrected product with Hydrogen, we need to apply corrections. Thus, we need to make an O dataset first, so we recursively call the L2 code again to produce the oxygen dataset and then continue on with the corrections.


sdhoyt (approved):

tmplummer (approved):
This all looks good to me. Sorry for the slow review.

PR #2273, opened 2025-10-02
Lo L1C - Add HAE Lat/Lon to pointing set

sdhoyt:
# Change Summary

## Overview
This PR adds the HAE Longitude and Latitude fields to the pointing set. One additional function was added to set these values and this adds new SPICE dependencies to L1C PSET processing.

### DIFF
--- a/imap_processing/lo/l1c/lo_l1c.py
+++ b/imap_processing/lo/l1c/lo_l1c.py
@@ -581,7 +581,7 @@ def set_background_rates(
             if row["type"] == "rate":
                 bg_rates[esa_step, bin_start:bin_end, :] = value
             elif row["type"] == "sigma":
-                bg_stat_uncert[esa_step, bin_start:bin_end, :] = value
+                bg_sys_err[esa_step, bin_start:bin_end, :] = value
             else:
                 raise ValueError("Unknown background type in ancillary file.")
     # set the background rates, uncertainties, and systematic errors
@@ -628,7 +628,7 @@ def set_pointing_directions(epoch: float) -> tuple[xr.DataArray, xr.DataArray]:
     hae_latitude : xr.DataArray
         The HAE latitude for each spin and off angle bin.
     """
-    et = ttj2000ns_to_et(epoch)
+    et = ttj2000ns_to_et(epoch) + 1
     # create a meshgrid of spin and off angles using the bin centers
     spin, off = np.meshgrid(
         SPIN_ANGLE_BIN_CENTERS, OFF_ANGLE_BIN_CENTERS, indexing="ij"
@@ -638,7 +638,7 @@ def set_pointing_directions(epoch: float) -> tuple[xr.DataArray, xr.DataArray]:
     # Transform from DPS Az/El to HAE lon/lat
     hae_az_el = frame_transform_az_el(
         et, dps_az_el, SpiceFrame.IMAP_DPS, SpiceFrame.IMAP_HAE, degrees=True
-    ).transpose(1, 0, 2)
+    )
 
     return xr.DataArray(
         data=hae_az_el[:, :, 0].astype(np.float64),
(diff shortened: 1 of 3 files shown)


imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -597,3 +606,52 @@ def set_background_rates(
...
+    # create a meshgrid of spin and off angles using the bin centers
+    spin, off = np.meshgrid(
+        SPIN_ANGLE_BIN_CENTERS, OFF_ANGLE_BIN_CENTERS, indexing="ij"
+    )
+    dps_az_el = np.stack([spin, off], axis=-1)
+
+    # Transform from DPS Az/El to HAE lon/lat
+    hae_az_el = frame_transform_az_el(
+        et, dps_az_el, SpiceFrame.IMAP_DPS, SpiceFrame.ECLIPJ2000, degrees=True
+    ).transpose(1, 0, 2)
```
tmplummer: Can you add a comment on why this transpose is needed? Is it just because of what meshgrid returns?

tmplummer (approved):
LGTM. Nice test coverage! Just the one question about the transpose.

imap_processing/lo/l1c/lo_l1c.py
greglucas: This doesn't seem right because the az_el should be returned in that order regardless of what is passed in I thought. hae_lon == az I think?

Can you use the `SpiceFrame.IMAP_HAE` version as well, I think that is what was preferred moving forward?

In your tests below it looks like you set all of the angles to 1 with `np.full` along both the spin and off angles. Could you change those to vary by some amount? I think we want to make sure that the az/lon variable goes 0->360 and the el goes -90->90 and it isn't flipped.

sdhoyt: Yeah I'm not sure why. the dps_az_el is (3600, 40, 2), but the frame transform output is (40, 3600, 2) so I thought I needed to transpose. It does the same thing when I use IMAP_HAE. I'll update the unit test

greglucas: Oh, I'm following now. It is the meshgrid indexing="ij" that is doing this. Do you need "ij" indexing or can you just leave that out?

sdhoyt: I'm not sure. Without the "ij" spin and off mesh grids are 40, 3600. But thought I'd want that to be 3600, 40. For whatever reason the frame transform switches the indices so when I remove "ij" the transform is 3600, 40 without the transpose. I'm having trouble telling which is the one I want.

tmplummer: > Yeah I'm not sure why. the dps_az_el is (3600, 40, 2), but the frame transform output is (40, 3600, 2) so I thought I needed to transpose. It does the same thing when I use IMAP_HAE. I'll update the unit test

That sounds like a bug in `frame_transform_az_el`. It should output the same shape as is input.  That is very weird... There is a test for `frame_transform_az_el` that checks that it outputs the same shape as is input. Are you sure it is transposing the dimensions on you?

sdhoyt: @subagonsouth I think so. I ran this code:

```python
    dps_az_el = np.stack([spin, off], axis=-1)
    print("dps_az_el shape:", dps_az_el.shape)

    # Transform from DPS Az/El to HAE lon/lat
    hae_az_el = frame_transform_az_el(
        et, dps_az_el, SpiceFrame.IMAP_DPS, SpiceFrame.IMAP_HAE, degrees=True
    )
    print("hae_az_el shape:", hae_az_el.shape)
```

and got this from the print statements:
```
dps_az_el shape: (3600, 40, 2)
hae_az_el shape: (40, 3600, 2)
```

tmplummer: @sdhoyt 
Ah... I think that I just fixed that in an unrelated ticket. Can you rebase onto dev? I just tested the same code you provided and the shape remained the same after the transform.

sdhoyt: That fixed the issue. Thank you!

PR #2274, opened 2025-10-03
fix: Produce MAG L1c without Normal mode data

mfacchinelli:
# Change Summary

## Overview
Adds support for generating MAG L1c files even when no input Normal mode data is provided.

## New Dependencies
N/A

## New Files
Only new test data files.

## Deleted Files
N/A

## Updated Files
- `imap_processing/mag/l1c/mag_l1c.py`:
  - Interpolation can also take place if the input normal mode file is `None`. Reuses existing functionality to downsample the whole day +/- 30 minutes, without checking Normal mode for gaps (the whole day is the gap).

## Testing
- Added a new test for L1c with no Normal mode.
- Uncommented `test_missing_norm_file` in `imap_processing/tests/mag/test_mag_l1c.py`


imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -378,13 +395,15 @@ def fill_normal_data(
...
     filled_timeline[:, 5] = ModeFlags.MISSING.value
-    for index, timestamp in enumerate(normal_dataset["epoch"].data):
-        timeline_index = np.searchsorted(new_timeline, timestamp)
-        filled_timeline[timeline_index, 1:5] = normal_dataset["vectors"].data[index]
-        filled_timeline[timeline_index, 5] = ModeFlags.NORM.value
-        filled_timeline[timeline_index, 6:8] = normal_dataset["compression_flags"].data[
-            index
-        ]
+
+    if normal_dataset:
```
maxinelasp: Could we move this if statement out to where the function is called? So if we don't have normal mode data, we can just not call this method and instead just make norm_filled equal to filled_timeline and fill the last value with ModeFlags.MISSING. I think it's confusing to call a function called "fill_normal_data" if you don't have any normal data to fill.

imap_processing/tests/mag/test_mag_validation.py
```diff
@@ -256,11 +256,13 @@ def test_mag_l1b_validation(test_number, mocks):
         assert np.allclose(expected_time, mago_time, atol=1e-6, rtol=0)
 
 
-@pytest.mark.xfail(reason="All L1C edge cases are not yet complete")
-@pytest.mark.parametrize(("test_number"), ["013", "014", "015", "016"])
+@pytest.mark.parametrize(("test_number"), ["013", "014", "015", "016", "024"])
 @pytest.mark.parametrize(("sensor"), ["mago", "magi"])
 @pytest.mark.external_test_data
 def test_mag_l1c_validation(test_number, sensor):
+    if test_number != "024":
```
maxinelasp: Might as well put 013 and 014 in there as they pass!

```suggestion
    if test_number in ["013", "014", "024"]:
```

imap_processing/tests/mag/test_mag_validation.py
```diff
@@ -271,45 +273,64 @@ def test_mag_l1c_validation(test_number, sensor):
...
 
-    l1c = mag_l1c(norm, day_to_process, burst)
+    l1c = mag_l1c(first_dataset, day_to_process, second_dataset)
     expected_output = pd.read_csv(
         source_directory / f"mag-l1b-l1c-t{test_number}-{sensor}-normal-out.csv"
     )
 
     for index in expected_output.index:
+        if index == expected_output.index[-1]:
+            abs_tol = 1e-1  # TODO: solve boundary condition issue
```
maxinelasp: Why is this needed? I'd appreciate a more verbose comment here

maxinelasp (commented):
Overall looks good, just a few suggested changes. Thank you!!

imap_processing/mag/l1c/mag_l1c.py
mfacchinelli: Ok, I've moved the generation of `filled_timeline` outside the function, so that the function only gets called if there is norm data.

imap_processing/tests/mag/test_mag_validation.py
mfacchinelli: This was a little harder than expected...

The issue was that we were interpolating outside the range of allowed values (since the CIC filter removes the `delay` value from the `input_timestamps`, one of the `output_timestamps` was now falling outside the range of the inputs, meaning we were extrapolating). My computer came up with a slightly different extrapolation value and this is what caused the discrepancy in the last vector.

The fix is to not extrapolate, thus this change makes use of the already existing `remove_invalid_output_timestamps` to discard any output timestamp outside the input timestamp range.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -514,6 +533,20 @@ def interpolate_gaps(
                     "compression_flags"
                 ].data[burst_gap_start + index]
 
+        # for any timestamp that was not filled and is still missing, remove it
+        missing_timeline = np.setdiff1d(gap_timeline, adjusted_gap_timeline)
+
+        for timestamp in missing_timeline:
+            timeline_index = np.searchsorted(filled_norm_timeline[:, 0], timestamp)
+            if (sum(filled_norm_timeline[timeline_index, 1:4]) != 0) or (
```
maxinelasp: When we switch to FILLVAL this assumption will not be true anymore (which is why that bug still exists, too many places where this code assumes that missing data == 0 when it should be == FILLVAL) so we shouldn't make that assumption here. 

I also am not sure why these values should be removed - shouldn't they be left but marked as missing data?

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -305,50 +321,86 @@ def process_mag_l1c(
...
+    # TODO: fill with FILLVAL
+    norm_filled: np.ndarray = np.zeros((len(new_timeline), 8))
+    norm_filled[:, 0] = new_timeline
+    # Flags, will also indicate any missed timestamps
+    norm_filled[:, 5] = ModeFlags.MISSING.value
+
+    return norm_filled
+
+
 def fill_normal_data(
```
maxinelasp, line 392: Instead of changing the signature here and modifying filled_timeline in-place with no return (which I think is confusing), why not leave this method as-is (except for removing day_to_process which is no longer used) and use it like:
```
if normal_mode_dataset:
    norm_filled = fill_normal_data(normal_mode_dataset, new_timeline)
  else:
    norm_filled = generate_empty_norm_array(new_timeline)
```

I think this is more clear about what each function is doing and doesn't modify arrays inside the function call.

Alternatively, I think it would be ok to take in the filled_timeline like your changes here, but still return an array so it's clear what is being modified. Also, new_timeline isn't needed anymore, since the epoch values are in the first column of `filled_timeline`. 

```
fill_normal_data(normal_dataset, filled_timeline) -> np.ndarray:
  <.. code here>
  return filled_timeline
```

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -360,24 +412,12 @@ def fill_normal_data(
...
-    Returns
-    -------
-    np.ndarray
+    filled_timeline : np.ndarray
         An (n, 8) shaped array containing the timeline filled with normal mode data.
         Gaps are marked as -1 in the generated flag column at index 5.
         Indices: 0 - epoch, 1-4 - vector x, y, z, and range, 5 - generated flag,
         6-7 - compression flags.
     """
-    # TODO: fill with FILLVAL
```
maxinelasp, line 376: Rather than passing filled_timeline in, you could also call `generate_empty_norm_array()` in here so we don't duplicate code.

imap_processing/mag/l1c/mag_l1c.py
```diff
@@ -463,20 +501,15 @@ def interpolate_gaps(
         ]
 
         short = (gap_timeline >= burst_epochs[burst_start]) & (
```
maxinelasp, line 506: Rather than casting in the if statement we can cast to int here. I think this should always be an int anyway but I was using it primarily for debugging before so I think casting to be safe makes sense

maxinelasp:
@all-contributors please add @mfacchinelli for code, test

maxinelasp:
@all-contributors please add @mfacchinelli for code, test

imap_processing/mag/l1c/mag_l1c.py
mfacchinelli: Ok, makes sense! I think I've addressed this

I've opted to leave `new_timeline` as an input and removed `filled_timeline`, instead. This way it's only a return value.

imap_processing/mag/l1c/mag_l1c.py
mfacchinelli: I'm not sure if the indexing of `gap_timeline` will still work correctly if `short` is not an array of bools. I've created a new variable for the number of `short` (which is a terrible name, sorry) to use in the if-statement and log print.

mfacchinelli:
> all-contributors please add mfacchinelli for code, test

Thanks Maxine!

imap_processing/mag/l1c/mag_l1c.py
mfacchinelli: We want to remove them because there is no valid science for those points. If any vector is missing, it should probably be an error. 

I will remove the check that the values are zero, and only go on the missing tag.

imap_processing/mag/l1c/mag_l1c.py
maxinelasp: Ah ok, I see what you're saying! In that case either way is fine, thanks for clarifying that

maxinelasp (approved):
Looks good! Let me know when you're ready to merge and I can merge it for you @mfacchinelli

mfacchinelli:
> Looks good! Let me know when you're ready to merge and I can merge it for you @mfacchinelli

Thanks Maxine, ready to merge!

PR #2275, opened 2025-10-03
2270 hi lo l2 cg correction   fix binning when additional dimensions are present

### LINKED ISSUE #2270: Hi Lo L2 CG correction - Fix binning when additional dimensions are present
### Algorithm Description:

Hi has count variable with dimensions: (epoch, energy, calibration_product, pset_bin) as well as variables, such as exposure time, that does not have the calibration_product dimension. The binning indices for the CG correction have the shape (energy, pset_bin). This is causing the current binning scheme to fail. I believe the solution is to convert the PointingSet.az_el_points to an xr.DataArray so that it can have dimensions that get used when broadcasting the binning indices to the shape of the data being binned.

tmplummer:
# Change Summary

## Overview
In prototyping the Hi CG correction code, I identified one last deficiency in the ena_mapping code. The multidimensional binning was still not working because Hi input data has an extra calibration_prod dimension. The numpy broadcasting doesn't know how to insert a singular dimension where needed, but xarray does. For this reason, I added xarray broadcasting support to the binning methods. This also meant that I had to update the PointingSet.az_el_points to be an xarray.DataArray object rather than just a np.ndarray.

## Updated Files
   - Add xr.DataArray support to `match_coords_to_indices`. If the input coordinates are a DataArray, then output indices in a DataArray with the same non-spatial dimensions.
   - Change `PointingSet.az_el_points` instance variable to be an xr.DataArray.
   - Update property retriever function `num_points` to return second to last dimension size. This allows there to be preceding non-spatial dimensions.
   - Add `LoHiBasePointingSet`, common base class for Hi and Lo PointingSet
   - Use xarray broadcasting in `AbstractSkyMap.project_pset_values_to_map`
   - Claude helped me add this test coverage.

Close: #2270

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -87,7 +87,7 @@ def match_coords_to_indices(
     This function always "pushes" the pixels of the input object to corresponding pixels
     in the output object's unwrapped rectangular grid or healpix tessellation;
     however, by swapping the input and output objects, one can apply the "pull" method
-    of index  matching.
+    of index matching.
 
     At present, the allowable inputs are either:
     - A PointingSet object and a SkyMap object, in either order of input/output.
@@ -100,8 +100,6 @@ def match_coords_to_indices(
         An object containing spatial pixel centers in azimuth and elevation,
         which will be matched to 1D indices of spatial pixels in the output frame.
         Must contain the Spice frame in which the pixel centers are defined.
-        If the input object has multi-dimensional az_el_points (as xr.DataArray),
-        the dimensions will be preserved in the output indices.
     output_object : PointingSet | AbstractSkyMap
         The object containing a grid or tessellation of spatial pixels
         into which the input spatial pixel centers will 'land', and be matched to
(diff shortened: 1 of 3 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -589,7 +610,50 @@ def __repr__(self) -> str:
...
+    Base class for Lo and Hi pointing sets with HAE coordinate data.
+
+    This class provides common functionality for pointing sets that contain
+    hae_longitude and hae_latitude coordinates in the dataset.
+    """
+
+    tiling_type: SkyTilingType = SkyTilingType.RECTANGULAR
+
+    def update_az_el_points(
+        self, az_variable: xr.DataArray, el_variable: xr.DataArray
```
greglucas: since Lo/Hi pointing sets both come in with `self.data["hae_longitude"]` spelled the same, do you want to just have this be a plain method that references those specific variable names instead?

imap_processing/ena_maps/ena_maps.py
```diff
@@ -153,6 +155,12 @@ def match_coords_to_indices(
     # Az/El pixel center coords of the input object in its own frame
     input_obj_az_el_input_frame = input_object.az_el_points
 
+    # Check if az_el_points is an xarray.DataArray to preserve dimension information
```
greglucas: Do you need to handle both numpy and xarray here or can you update Ultra / others to also use xarray dimensions as well and avoid these special cases?

greglucas (approved):
Yeah, I was finding it much nicer to deal with xarray's nice dimension alignment too rather than figuring out how to insert dimensions arbitrarily too! This looks nice, a few minor comments that you can take/leave as you see fit.

imap_processing/ena_maps/ena_maps.py
tmplummer: Yeah, I think that is fine. For the CG correction, I was making new variables that contain the updated directions but I don't see why I shouldn't just overwrite the existing "hae_latitude/longitude" variables.

imap_processing/ena_maps/ena_maps.py
tmplummer: Good point. The only class not using xarray is RectangularPointingSet which was written by Nat and is not really being used other than quite extensively in testing. I'll see if that is an easy change.

Issue #2263, opened 2025-09-29
Slow packet parsing


greglucas:
I uploaded an 18MB MAG packet file for processing and it took ~25 minutes for L0->L1a to complete, but all the other steps l1a->l1c were less than 1 minute. We should look into the performance of the packet parsing and whether we can improve this at all.

I haven't looked at other instruments, this could be MAG specific or more general. So this is just an awareness issue at the moment.

greglucas:
I think this is specific to the MAG L0->L1a routine. I put similar sized HIT packet files through and they took ~1 minute, whereas MAG was ~1 hour. FYI @maxinelasp

greglucas:
I ran `packet_files_to_dataset` on the MAG packet file and it was a few seconds is all to create the dataset, so this is somewhere after space_packet_parser and might have to do with dataclass creation.

maxinelasp:
[mag_l1a_report_20251003_092714.txt](https://github.com/user-attachments/files/22686407/mag_l1a_report_20251003_092714.txt)

MAG L1A has a lot of complicated code in it due to the custom compression algorithm.

Here are some profiler results. As I expected, a lot of the high-usage functions are for processing the compressed vectors. Actually, the most time is spent in the call which appends and sorts the vectors into mago and magi, which uses a naive append/concatenate method rather than anything smarter. Nearly every other call after that is vectorized over numpy arrays or relies on numpy functions for the processing, and I don't expect them to be easy to improve on. Luckily the percall timing for most of the high ncall functions is low.

That append_vectors call should be a low-hanging fruit though. It happens at the end, so I think I can just collect all the vectors and do one concatenation call for all the packets rather than calling concatenate for each packet.

maxinelasp:
Fixed in #2276

PR #2276, opened 2025-10-03
Fix MAG L1A performance

maxinelasp:
# Change Summary
Fix bad MAG L1A performance by concatenating vector arrays at the end instead of at the end of each packet.

This brought processing time of a set of packets from 3891s (~1hr) to 390s (~6.5 minutes). 

This preserved the outside API of the dataclass by using properties and initvars, so none of the testing needed updating. All the tests pass. 

Pre-fix profile:
[mag_l1a_report_20251003_092714.txt](https://github.com/user-attachments/files/22687842/mag_l1a_report_20251003_092714.txt)
Post-fix profile:
[mag_l1a_report_20251003_140626.txt](https://github.com/user-attachments/files/22687843/mag_l1a_report_20251003_140626.txt)

Corrects #2263

### DIFF
--- a/imap_processing/mag/l1a/mag_l1a.py
+++ b/imap_processing/mag/l1a/mag_l1a.py
@@ -43,7 +43,7 @@ def mag_l1a(packet_filepath: Path) -> list[xr.Dataset]:
         A list of generated filenames.
     """
     packets = decom_mag.decom_packets(packet_filepath)
-
+    logging.info("Packet decoding complete, beginning L1A processing.")
     norm_data = packets["norm"]
     burst_data = packets["burst"]
 
--- a/imap_processing/mag/l1a/mag_l1a.py
+++ b/imap_processing/mag/l1a/mag_l1a.py
@@ -188,7 +188,7 @@ def process_packets(
             secondary_packet_data.start_time,
         )
 
-        # Sort primary and secondary into MAGo and MAGi by 24 hour chunks
+        # Sort primary and secondary into MAGo and MAGi
 
         if mago is None:
             mago = MagL1a(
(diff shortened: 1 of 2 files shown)


imap_processing/mag/l1a/mag_l1a.py
```diff
@@ -43,7 +43,7 @@ def mag_l1a(packet_filepath: Path) -> list[xr.Dataset]:
         A list of generated filenames.
     """
     packets = decom_mag.decom_packets(packet_filepath)
-
+    logging.info("Packet decoding complete, beginning L1A processing.")
```
greglucas:
```suggestion
    logger.info("Packet decoding complete, beginning L1A processing.")
```

imap_processing/mag/l1a/mag_l1a_data.py
```diff
@@ -248,30 +259,67 @@ class MagL1a:
...
         # the object
         self.most_recent_sequence = starting_packet.src_seq_ctr
         self.update_compression_array(starting_packet, self.vectors.shape[0])
 
+    @property
+    def vectors(self) -> np.ndarray:
+        """
+        Concatenate the internal vector list into a numpy array.
+
+        If the array has already been created, return the cached version.
```
greglucas, line 308: Will this ever be called in the middle of updating or will `_vector_list` ever change after this property is called? You might want to even call this property `cached_vectors()` or something similar to make it very clear there is some state behind this.

greglucas (approved):
Some minor comments, but looks good overall!

imap_processing/mag/l1a/mag_l1a_data.py
maxinelasp: If `_vector_list` is ever updated after the property is called then the cached version will be invalidated (see in `append_vectors()`. It also has to be named vectors to preserve the API and avoid changes to the tests and to where this class is used in MAG L1A.

maxinelasp: In normal usage, I also would not expect this property to be called unless _vector_list is fully updated, but it should still work ok even with incomplete vectors.

PR #2277, opened 2025-10-06
Update pointing_attitude job

### LINKED ISSUE #2272: Fix pointing attitude job
### Requirements:

The logic in the pointing attitude job will currently only calculate a segment for a pointing that is completely contained in the input S/C attitude kernel. This needs to be updated to produce a segment for any pointing that is partially covered by the input S/C attitude kernel.

This will also mean that there is not 1:1 correspondence between S/C attitude kernels and pointing attitude kernels. The logic for naming the output Pointing kernel will need to be updated to correctly reflect doy coverage which could be different than the input kernel... maybe???

tmplummer:
# Change Summary

## Overview
When looking into failures in the pointing_attitude jobs using early flight data, I identified a deficiency in the logic that selects which Pointings to generate DPS attitude for. The previous logic would miss any pointing that contains the end of one CK and start of the next. This issue is exacerbated by the fact that the CKs we are getting are not appending data to a fixed start date.

[pointing-attitude kernel creation.pdf](https://github.com/user-attachments/files/22725433/pointing-attitude.kernel.creation.pdf)

## Commits
-Update pointing attitude logic to generate data for any pointing partially covered by any of the CK attitude history files passed in as dependencies.

## Updated Files
   - Improve logging messages
   - Convert repoint times to Pointing times early
   - Update filtering to include all Pointings that are partially covered by any CK kernel passed in as a dependency but don't end after the end of CK coverage. This logic relies on the batch starter job to correctly identify the kernels needed to cover the time range between the new repoint table and the previous repoint table.
   - Improve logging
   - Update test to check that Pointing selection logic is working as expected.

Closes: #2272

### DIFF
--- a/imap_processing/spice/pointing_frame.py
+++ b/imap_processing/spice/pointing_frame.py
@@ -255,6 +255,10 @@ def calculate_pointing_attitude_segments(
     pointing_end_ets = pointing_end_ets[keep_mask]
 
     n_pointings = len(pointing_ids)
+    if n_pointings == 0:
+        logger.warning(
+            "No Pointings identified based on coverage of this CK file. Skipping."
+        )
 
     pointing_segments = np.zeros(n_pointings, dtype=POINTING_SEGMENT_DTYPE)
 
--- a/imap_processing/spice/pointing_frame.py
+++ b/imap_processing/spice/pointing_frame.py
@@ -269,9 +269,8 @@ def calculate_pointing_attitude_segments(
     n_pointings = len(pointing_ids)
     if n_pointings == 0:
         logger.warning(
-            "No Pointings identified based on coverage of this CK file. Skipping."
+            "No Pointings identified based on coverage of CK files. Skipping."
         )
-        return np.array([], dtype=POINTING_SEGMENT_DTYPE)
 
     pointing_segments = np.zeros(n_pointings, dtype=POINTING_SEGMENT_DTYPE)
 
--- a/imap_processing/spice/pointing_frame.py
+++ b/imap_processing/spice/pointing_frame.py
@@ -200,13 +200,16 @@ def calculate_pointing_attitude_segments(
     - IMAP historical attitude kernel from which the pointing frame kernel will
     be generated.
     """
-    logger.info(f"Extracting mean spin axes from CK kernel {ck_path.name}")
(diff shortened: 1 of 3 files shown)


laspsandoval (approved):
Looks good, Tim. Thanks.

tmplummer:
@laspsandoval, @greglucas, @tech3371,
Sorry for the churn on this PR. After finding out (unconfirmed) that Ian expects to continue to generate a spacecraft attitude history file for each calendar day, I had to re-work the logic in this job yet again. I will update the PR description to reflect the new logic. Please give it a fresh review.

I am also opening a ticket regarding how we can close the deficiency of reprocessing in the case of an updated s/c attitude history file. Currently, we only trigger off of new repoint files, so we will miss updating the DPS kernel if a past kernel gets updated say due to a data gap.

imap_processing/spice/pointing_frame.py
```diff
@@ -200,57 +213,72 @@ def calculate_pointing_attitude_segments(
...
+    pointing_ids = pointing_ids[keep_mask]
+    pointing_start_ets = pointing_start_ets[keep_mask]
+    pointing_end_ets = pointing_end_ets[keep_mask]
+
+    n_pointings = len(pointing_ids)
+    if n_pointings == 0:
+        logger.warning(
+            "No Pointings identified based on coverage of this CK file. Skipping."
+        )
+        return np.array([], dtype=POINTING_SEGMENT_DTYPE)
```
greglucas: Looking at this, I think this would all fall out below as well (create an empty array that gets returned and not iterate through the for-loop because it is zero) so it doesn't need a special case here. The warning message seems useful though.

greglucas (approved):

greglucas:
> I am also opening a ticket regarding how we can close the deficiency of reprocessing in the case of an updated s/c attitude history file. Currently, we only trigger off of new repoint files, so we will miss updating the DPS kernel if a past kernel gets updated say due to a data gap.

If handling some of this is overly complicated logic perhaps we could get notified when this happens and do some manual reprocessing of jobs that need to be updated instead of trying to engineer something in...
This seems like something we can push to the backlog to me and not a dealbreaker for implementing the right thing now.

Issue #2242, opened 2025-09-19
ENA Maps - Inject correct refernce frame into CDF attributes


tmplummer:
### Algorithm Description:
Reference frame should be mentioned in CATDESC, FIELDNAME, LABLAXIS, and DICT_KEY. We need to inject it into those attributes based on the map descriptor.

jtniehof:
I am guessing this is not a surprise, but the current Lo L2 has a CATDESC on the latitude variable of "Pixel center latitude in {frame} reference frame in the range [-90, 90]" where `{frame}` is literal.

I talked briefly with Dan and the thought is to use "Pixel center latitude in {frame}" (with appropriate substitution), and add VAR_NOTES "Reference frames are defined according to IMAP
mission SPICE kernels". Similar for longitude.

Rationale:
- The range is in the VALIDMIN/VALIDMAX and probably fairly evident from the map (I am sure the range was included because somebody made a request...)
- Saying "frame" doesn't add much and opens up the nomenclature argument of reference frame vs. coordinate system vs. ideal reference system....
- The VAR_NOTES give the concrete definition, and "reference frame" there is the SPICE terminology.

This isn't a high priority, just noting the discussion--I came across while working on some GLOWS L3.

tmplummer:
This may have been partially implemented by Menlo in PR: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/3255

PR #2279, opened 2025-10-07
Update common  map variable metadata

pleasant-menlo:
# Change Summary
Updated the common map metadata per the definition [here](https://o365coloradoedu-my.sharepoint.com/:x:/g/personal/plummert_colorado_edu/EX8IRmgRKw5Hpik9wPAXZQABBgOemDqOv-uLMlRxZLJaTw?rtime=8JL00J0F3kg)

Note that many of the variables in the green section of the spreadsheet were missing from the existing definition and these were not added in this PR. (bg_rate_uncert, bg_intensity, ...). We believe ena_rate, and ena_count have been renamed to ena_count_rate, and counts, but believe this change will require updates to the code so is not included here.

## Updated Files
- imap_enamaps_l2-common_variable_attrs.yaml
  - Changed metadata for already present variables

### DIFF
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -113,37 +113,22 @@ epoch_delta:
 # in the tiling-specific YAML files.
 longitude:
   <<: *default_float32
-  CATDESC: Pixel center longitude
   FIELDNAM: "{frame} Longitude"
-  LABLAXIS: Longitude
-  FORMAT: I3
+  LABLAXIS: "{frame} Longitude"
   UNITS: deg
   VAR_TYPE: support_data
   VALIDMIN: 0
   VALIDMAX: 360
-  LABL_PTR_1: longitude_label
-  SCALETYP: linear
-  MONOTON: INCREASE
-  DELTA_PLUS_VAR: longitude_delta
-  DELTA_MINUS_VAR: longitude_delta
   DICT_KEY: SPASE>Support>SupportQuantity:Orientation,Qualifier:DirectionAngle.AzimuthAngle,CoordinateSystemName:HAE,CoordinateRepresentation:Spherical
 
 latitude:
   <<: *default_float32
-  CATDESC: Pixel center latitude
   FIELDNAM: "{frame} Latitude"
-  LABLAXIS: Latitude
-  FORMAT: I2
+  LABLAXIS: "{frame} Latitude"
   UNITS: deg
   VAR_TYPE: support_data
   VALIDMIN: -90
   VALIDMAX: 90
-  LABL_PTR_1: latitude_label
-  SCALETYP: linear
-  BIN_LOCATION: 0.5
-  MONOTON: INCREASE
-  DELTA_PLUS_VAR: latitude_delta
-  DELTA_MINUS_VAR: latitude_delta
   DICT_KEY: SPASE>Support>SupportQuantity:Orientation,Qualifier:DirectionAngle.ElevationAngle,CoordinateSystemName:HAE,CoordinateRepresentation:Spherical
 
 # Define Data variables, but not their pixel-level DEPENDs
@@ -297,7 +282,7 @@ obs_date_range:
   <<: *default_int64
   datatype: int64
   CATDESC: Standard deviation of the observation date.
-  FIELDNAM: J2000 Nanoseconds
+  FIELDNAM: Observation date range
   UNITS: ns
   DEPEND_0: epoch
   VAR_TYPE: support_data
--- a/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
@@ -81,20 +81,6 @@ energy_delta_plus:
 epoch:
   CATDESC: Map time range start, number of nanoseconds since J2000 with leap seconds included
   FIELDNAM: J2000 Nanoseconds
-  LABLAXIS: epoch
-  FILLVAL: -9223372036854775808
-  FORMAT: " " # Supposedly not required, fails in xarray_to_cdf
-  VALIDMIN: 315576066184000000   # 2010-01-01T00:00:00 mission start (APL 0 epoch)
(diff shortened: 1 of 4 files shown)


greglucas:
Looks good to me. Ultra has some assertions on exact metadata it looks like, could you update those tests here too.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -72,16 +72,36 @@ energy_delta_plus:
...
   LABLAXIS: energy
   UNITS: KeV
-  FIELDNAM: energy_bin_delta_plus
+  FIELDNAM: energy delta plus
   DISPLAY_TYPE: no_plot
   DEPEND_1: energy
   LABL_PTR_1: energy_label
   DICT_KEY: SPASE>Particle>ParticleType:Atom,ParticleQuantity:Energy,Qualifier:Uncertainty
 
+epoch:
```
tmplummer, line 81: I will experiment with epoch inheritance. It seems like only specifying `DELTA_PLUS_VAR` and `BIN_LOCATION` here might work without duplicating the full epoch definition.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -93,24 +113,37 @@ epoch_delta:
 # in the tiling-specific YAML files.
 longitude:
   <<: *default_float32
-  CATDESC: "Pixel center longitude in {frame} reference frame in the range [0, 360]."
```
tmplummer, line 96: I specifically put this in so that the reference frame could be injected into this attribute. See: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/2242

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -93,24 +113,37 @@ epoch_delta:
...
+  CATDESC: Pixel center longitude
   FIELDNAM: "{frame} Longitude"
-  LABLAXIS: "{frame} Longitude"
-  UNITS: degrees
+  LABLAXIS: Longitude
+  FORMAT: I3
```
tmplummer: This is the common variable attributes file meaning that these attributes get included in the Ultra HealPix maps as well. Since HealPix maps don't have integer lat/lon, I think this belongs in the YAML file specifically for RectangularMaps.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -93,24 +113,37 @@ epoch_delta:
 # in the tiling-specific YAML files.
 longitude:
   <<: *default_float32
-  CATDESC: "Pixel center longitude in {frame} reference frame in the range [0, 360]."
+  CATDESC: Pixel center longitude
   FIELDNAM: "{frame} Longitude"
```
tmplummer, line 102: SPDF gave specific guidance on this:

>  For longitude, and latitude variables, please include coordinate system name (HAE ?) in all of their CATDESC, FIELDNAM and LABLAXIS, so that, for example, LABLAXIS="HAE latitude", so that immediately visible to user.

That is why I put these `{frame}` placeholders in here. They will eventually be populated.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -93,24 +113,37 @@ epoch_delta:
...
+  FORMAT: I3
+  UNITS: deg
   VAR_TYPE: support_data
   VALIDMIN: 0
   VALIDMAX: 360
+  LABL_PTR_1: longitude_label
+  SCALETYP: linear
+  MONOTON: INCREASE
+  DELTA_PLUS_VAR: longitude_delta
+  DELTA_MINUS_VAR: longitude_delta
```
tmplummer: Again, these need to be moved to the Rectangular specific YAML file: imap_enamaps_l2-rectangular_variable_attrs.yaml

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -144,10 +177,10 @@ ena_intensity_sys_err:
   <<: *default_float32
   CATDESC: ENA intensity non-statistical error.
   FIELDNAM: Intensity non-stat error
-  UNITS: counts/(s * cm^2 * Sr * KeV)
+  UNITS: cm -2 s -1 sr -1 keV -1
   DEPEND_0: epoch
   VAR_TYPE: support_data
-  LABLAXIS: Non-Statistical Err
+  LABLAXIS: Non-statistical Error
```
tmplummer, line 154: SPDF guidance for LABLAXIS:

> approximately 10 characters, but preferably 6 characters

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
```diff
@@ -261,7 +297,7 @@ obs_date_range:
   <<: *default_int64
   datatype: int64
   CATDESC: Standard deviation of the observation date.
-  FIELDNAM: Observation date range
+  FIELDNAM: J2000 Nanoseconds
```
tmplummer: Even though this is what is entered in the spreadsheet, I don't think this is correct for FIELDNAM

tmplummer (commented):
Thank you for propagating many changes from the spreadsheet into this YAML.  I have a handful of changes that are needed.

pleasant-menlo:
Hi Tim, thanks for looking at this. We made some updates. Let us know if anything else needs to change. We also updated the spreadsheet, updates are highlighted in red.

imap_processing/cdf/config/imap_enamaps_l2-common_variable_attrs.yaml
tmplummer: @pleasant-menlo ,
I confirmed that you can just have the two added attributes here. The rest will get inherited from the default `imap_constant_attrs.yaml` file.

tmplummer (approved):
Thanks! LGTM.

PR #2284, opened 2025-10-07
2282 pre binning calculations on PSET

### LINKED ISSUE #2282: Hi Lo L2 CG correction - pre binning calculations on PSET
### Algorithm Description:

This covers the CG correction calculations needed to prepare PSET data for binning. Needs to:
- Update `az_el_points` to reflect helio-frame source direction for each bin at each energy level
- Compute the S/C frame energy for each bin at each energy level

tmplummer:
# Change Summary

## Overview
This adds CG calculations that need to be done on each PSET prior to binning. I expect to need to make a few minor changes as I integrate this code into the Hi and Lo L2 pipelines.
I also snuck in a fix to Hi anti-ram filtering.

## Updated Files
   - fix anti-ram filtering by checking for "anti"
- imap_processing/ena_maps/utils/coordinates.py
   - add a generic coordinate for Cartesian vectors
   - Add CG calculations needed to be done for each PSET prior to binning to a SkyMap
   - Claude aided addition of test coverage for new functions

Closes: #2282

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -1315,7 +1315,7 @@ class RectangularSkyMap(AbstractSkyMap):
             if ("L2" in name)
         ]
         l2_coords.append(CoordNames.TIME.value)
-        for map_coord in cdf_ds.sizes.keys():
+        for map_coord in cdf_ds.dims:
             if map_coord not in l2_coords:
                 cdf_ds = cdf_ds.drop_dims(map_coord)
 
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -655,7 +655,7 @@ class HiPointingSet(LoHiBasePointingSet):
         super().__init__(dataset, spice_reference_frame=geometry.SpiceFrame.ECLIPJ2000)
 
         # Filter out ENAs from non-selected portions of the spin.
-        if spin_phase not in ["full", "ram", "anti-ram"]:
+        if spin_phase not in ["full", "ram", "anti"]:
             raise ValueError(f"Unrecognized spin_phase value: {spin_phase}.")
         # ram only includes spin-phase interval [0, 0.5)
         # which is the first half of the spin_angle_bins
@@ -665,7 +665,7 @@ class HiPointingSet(LoHiBasePointingSet):
             )
         # anti-ram includes spin-phase interval [0.5, 1)
         # which is the second half of the spin_angle_bins
-        elif spin_phase == "anti-ram":
+        elif spin_phase == "anti":
             self.data = self.data.isel(
                 spin_angle_bin=slice(self.data["spin_angle_bin"].data.size // 2, None)
             )
@@ -1315,7 +1315,7 @@ class RectangularSkyMap(AbstractSkyMap):
             if ("L2" in name)
         ]
         l2_coords.append(CoordNames.TIME.value)
-        for map_coord in cdf_ds.dims.keys():
+        for map_coord in cdf_ds.sizes.keys():
             if map_coord not in l2_coords:
                 cdf_ds = cdf_ds.drop_dims(map_coord)
 
--- a/imap_processing/ena_maps/utils/corrections.py
+++ b/imap_processing/ena_maps/utils/corrections.py
@@ -4,7 +4,16 @@ from pathlib import Path
 
 import numpy as np
 import pandas as pd
+import xarray as xr
 from numpy.polynomial import Polynomial
+
+from imap_processing.ena_maps.ena_maps import LoHiBasePointingSet
+from imap_processing.ena_maps.utils.coordinates import CoordNames
+from imap_processing.spice import geometry
+
+# Physical constants for Compton-Getting correction
+ERG_PER_EV = 1.6e-12  # erg per eV - unit conversion factor
+PROTON_MASS_GRAMS = 1.67e-24  # proton mass in grams
 
 
 class PowerLawFluxCorrector:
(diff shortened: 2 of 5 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -1315,7 +1315,7 @@ def build_cdf_dataset(
             if ("L2" in name)
         ]
         l2_coords.append(CoordNames.TIME.value)
-        for map_coord in cdf_ds.dims.keys():
+        for map_coord in cdf_ds.sizes.keys():
```
greglucas: I think you might be able to just get rid of the `.keys()` specifier here.
```suggestion
        for map_coord in cdf_ds.dims:
```

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -4,8 +4,17 @@
...
+import xarray as xr
 from numpy.polynomial import Polynomial
 
+from imap_processing.ena_maps.ena_maps import LoHiBasePointingSet
+from imap_processing.ena_maps.utils.coordinates import CoordNames
+from imap_processing.spice import geometry
+
+# Physical constants for Compton-Getting correction
+ERG_PER_EV = 1.6e-12  # erg per eV - unit conversion factor
+PROTON_MASS_GRAMS = 1.67e-24  # proton mass in grams
```
greglucas: Are these enough decimal places? If we have scipy as a dependency already should we use the `scipy.constants` module values here? No need to change, just making sure this was considered.

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -289,3 +298,241 @@ class or child classes.
...
+def _add_spacecraft_velocity_to_pset(pset: LoHiBasePointingSet, et: float) -> None:
+    """
+    Calculate and add spacecraft velocity data to pointing set.
+
+    Parameters
+    ----------
+    pset : LoHiBasePointingSet
+        Pointing set object to be updated.
+    et : float
+        Ephemeris time in seconds past J2000.
```
greglucas: Can't this be obtained from the pset["epoch"] rather than having it as another parameter?

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -289,3 +298,241 @@ class or child classes.
...
+        look_direction already added.
+    energies_hf : xr.DataArray
+        ENA energies in the heliosphere frame in eV.
+
+    Notes
+    -----
+    The algorithm is based on the "Appendix A. The IMAP-Lo Mapping Algorithms"
+    document.
+    Adds the following DataArrays to pset.data:
+    - "energy_sc": ENA energies in spacecraft frame (eV)
```
greglucas, line 408: minor nit, but you use `energy` here, but `energies` in the argument to the function. Can we make the input `energy_hf`?

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -289,3 +298,241 @@ class or child classes.
...
+    - "ena_source_hae_longitude": ENA source longitudes in heliosphere frame (degrees)
+    - "ena_source_hae_latitude": ENA source latitudes in heliosphere frame (degrees)
+    """
+    # Store heliosphere frame energies
+    pset.data["hf_energy"] = energies_hf
```
greglucas: Now it is `frame_energy` rather than `energy_frame`. I'd suggest trying to make it consistent one way or the other so that you use all `frame_` or `_frame` throughout on energy and directions.

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -289,3 +298,241 @@ class or child classes.
...
+    energy_u = (
+        0.5 * PROTON_MASS_GRAMS * (sc_velocity_km_per_sec * 1e5) ** 2 / ERG_PER_EV
+    )
+
+    # Calculate y values for each energy level (Equation 61)
+    # y_k = sqrt(E^h_k / E^u)
+    y = np.sqrt(pset.data["hf_energy"] / energy_u)
+
+    # Velocity magnitude factor calculation (Equation 62)
+    # x_k = (êₛ · û_sc) + sqrt(y² + (êₛ · û_sc)² - 1)
```
greglucas, line 456: I really like that you were able to get the UTF-8 math characters for all of this. How did you go about getting it formatted so nicely?

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -289,3 +298,241 @@ class or child classes.
...
+    x = dot_product + np.sqrt(y**2 + dot_product**2 - 1)
+
+    # Calculate ENA speed in the spacecraft frame
+    # |v⃗_sc| = x_k * U_sc
+    velocity_sc = x * sc_velocity_km_per_sec
+
+    # Calculate the kinetic energy in the spacecraft frame
+    # E_sc = (1/2) * M_p * v_sc² (convert km/s to cm/s with 1.0e5 factor)
```
greglucas, line 464: I was curious about the 1e5 above, maybe put this similar comment up there on l.424 as well.

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -289,3 +298,241 @@ class or child classes.
...
+
+    All calculations are performed using xarray DataArrays to preserve dimension
+    information throughout the computation.
+
+    Parameters
+    ----------
+    pset : LoHiBasePointingSet
+        Pointing set object containing HAE longitude/latitude coordinates.
+    et : float
+        Ephemeris time in seconds past J2000 for SPICE calculations.
```
greglucas: Similar to above, can we rely on the pset's epoch?

imap_processing/tests/ena_maps/test_corrections.py
```diff
@@ -228,3 +238,345 @@ def test_apply_flux_correction(self, mock_predictor_corrector, hi_coeffs_file):
...
+
+        # Set up simple look directions
+        _add_cartesian_look_direction(mock_hi_pset)
+
+        # Single energy level
+        energies_hf = xr.DataArray(np.array([1000.0]), dims=["esa_energy_step"])
+
+        _calculate_compton_getting_transform(mock_hi_pset, energies_hf)
+
+        # Physical checks:
```
greglucas, line 492: I like this approach 👍

imap_processing/ena_maps/utils/corrections.py
tmplummer: Claude helped with this 😆

greglucas: Did you put the pdf text into Claude or what was the process like from that standpoint?

tmplummer: Yes, I had Claude help me decipher Nathan's pdf algorithm document.

I don't know if you looked at the python code that he also sent, but it was unintelligible with zero comments. I also had Claude comment up that code based on the algorithm document. That was really the only way I could figure out what was supposed to be happening.

I'm still of the opinion that one of the steps is sort of being done backward. It seems like the flux correction step and interpolation step sort of hide that issue. Eventually I will bring that up with Dan, Paul, and Nathan.

greglucas: > I don't know if you looked at the python code that he also sent

I did not. That is also another great use of AI :)

> I'm still of the opinion that one of the steps is sort of being done backward. It seems like the flux correction step and interpolation step sort of hide that issue. Eventually I will bring that up with Dan, Paul, and Nathan.

Could you add a comment where you think that is so that future us knows exactly where to go and what lines are affected. NOTE: I think this might be wrong, we should look into it later

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -4,7 +4,18 @@
...
 from numpy.polynomial import Polynomial
+from scipy.constants import electron_volt, erg, proton_mass
+
+from imap_processing.ena_maps.ena_maps import LoHiBasePointingSet
+from imap_processing.ena_maps.utils.coordinates import CoordNames
+from imap_processing.spice import geometry
+from imap_processing.spice.time import ttj2000ns_to_et
+
+# Physical constants for Compton-Getting correction
+ERG_PER_EV = electron_volt / erg  # erg per eV - unit conversion factor
```
greglucas, line 20: This is inverse of the name. I think these should be swapped and then multipled rather than divided when used below.

greglucas (approved):
Looks good to me. One comment on the inverse of one of the unit quantities being named.

imap_processing/ena_maps/utils/corrections.py
tmplummer: I had to convince myself of this naming and calculation as well. The confusion comes from the fact that the constant values are units of `[J / eV]` and `[J / erg]`. So, to get units of `[erg / eV]`, you have to do the opposite of what you would think. I'll add a comment to clarify that.

greglucas: Ahh, I see what you're saying, I didn't realize it was energy per value. Thank you for adding a comment on this!

PR #2286, opened 2025-10-08
Fixing malformed JSON

maxinelasp:
Noticed all-contributors wasn't working


greglucas (approved):
Interesting, I figured all-contributors would have updated that correctly itself. I wonder if these were added manually instead of by the bot or if there is an issue upstream with the bot.

PR #2289, opened 2025-10-09
FIX: Explicitly choose SWAPI science/housekeeping path

greglucas:
If we didn't have science data, we'd fallback to the housekeeping path. This caused issues with a recursive loop where the science dataset created housekeeping datasets and triggered the job again.

### DIFF
--- a/imap_processing/swapi/l1/swapi_l1.py
+++ b/imap_processing/swapi/l1/swapi_l1.py
@@ -727,7 +727,7 @@ def process_swapi_science(
     return dataset
 
 
-def swapi_l1(dependencies: ProcessingInputCollection) -> xr.Dataset:
+def swapi_l1(dependencies: ProcessingInputCollection, descriptor: str) -> xr.Dataset:
     """
     Will process SWAPI level 0 data to level 1.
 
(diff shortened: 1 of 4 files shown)


imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -754,9 +757,11 @@ def swapi_l1(dependencies: ProcessingInputCollection) -> xr.Dataset:
         l0_files[0], xtce_definition, use_derived_value=False
     )
 
-    hk_files = dependencies.get_file_paths(descriptor="hk")
-    if hk_files and l0_unpacked_dict.get(SWAPIAPID.SWP_SCI, None) is not None:
+    if descriptor == "sci":
         logger.info(f"Processing SWAPI science data for {l0_files[0]}.")
+        if SWAPIAPID.SWP_SCI not in l0_unpacked_dict:
```
tech3371, line 762: should we raise error here since at this point, we should have science packets in it? also for below?

tech3371 (approved):
minor comment. Thank you for doing this!

imap_processing/swapi/l1/swapi_l1.py
greglucas: No, because we aren't guaranteed that packets of either type will be kicked off.

We get an L0 packet file that could have only HK data (current situation), only SCI data, or both and we don't know which one it is. So your dependency file is correct I think where both depend on the L0 file, there is just no way of knowing which situation we're in so we want to fail gracefully if we didn't get any science files.

greglucas:
@tech3371 can you build the SWAPI container and test in dev and release to prod?

PR #2290, opened 2025-10-09
CoDICE: Hi L2 omni and sectored work and validation

### LINKED ISSUE #2057: CoDICE: L2 Implement hi Omni-directional intensities data product

### LINKED ISSUE #2058: CoDICE: L2 Implement hi sectored intensities data product

tech3371:
# Change Summary
Related #2057 #2058 

## Overview

This does almost all of processing code for L2 omni and sectored. I will have following smaller PR for CDF attrs. I tested that CLI code works through both CLI command and unittests.

## Updated Files

- cli.py and imap_processing/codice/constants.py
   - updates related to new addition
- codice_l2.py
   - main file with L2 omni and sector code

## Testing

common ways to read test LUT files path
- imap_processing/tests/codice/conftest.py
- imap_processing/tests/codice/test_codice_hi_l2.py


imap_processing/codice/codice_l2.py
```diff
@@ -13,24 +13,180 @@
...
+        omni_direction_intensities = l1b_dataset[species] / (
+            L2_GEOMETRY_FACTORS
+            * L2_NUMBER_OF_SSD
+            * species_efficiencies
+            * energy_passbands
+        )
+        # Store by replacing existing species data with omni-directional intensities
+        l1b_dataset[species].values = omni_direction_intensities
+
+    return l1b_dataset
```
greglucas: The l1b return is a bit misleading here, because in the comment you say you are returning the L2 dataset. Do you want to say at the very top: l2 = l1b_dataset, or just call this dataset and say you are returning the modified dataset and disregard whether you're updating the l1/l2 dataset.

imap_processing/codice/codice_l2.py
```diff
@@ -13,24 +13,180 @@
...
+    efficiencies_file = dependencies.get_file_paths(
+        descriptor="l2-hi-sectored-efficiency"
+    )[0]
+    efficiencies_df = pd.read_csv(efficiencies_file)
+    # Similar to hi-omni, each species has different shape.
+    # Because of that, we need to loop over each species and calculate
+    # sectored intensities separately.
+    for species in HI_SECTORED_VARIABLE_NAMES:
+        # Efficiencies from dataframe maps to different dimension in L1B data.
+        # For example:
```
greglucas, line 138: Really helpful comments, thank you!

As an FYI, with all these reshapings, one thing we are finding on the mapping side of things is that xarray is really helpful to not have to worry about where the dimensions are and it will handle the broadcasting for you. So you can do math on arbitrary dimensions and the reshaping will be taken care of for you if the dimensions are named the same.

imap_processing/codice/constants.py
```diff
@@ -2259,3 +2259,23 @@
     30: [116, 117, 118, 119, 120, 121],
     31: [122, 123, 124, 125, 126, 127],
 }
+
+L2_GEOMETRY_FACTORS = 0.013
```
greglucas:
```suggestion
L2_GEOMETRIC_FACTOR = 0.013
```

greglucas (approved):
minor nits, but looks good otherwise 👍

imap_processing/codice/codice_l2.py
tech3371: Yep, that sounds good

imap_processing/codice/codice_l2.py
tech3371: I agree. In this case, we don't want to reshape input xarray shape. Instead, we are making sure that the efficiency data is broadcast to match correct dimension in input xarray data because the last two dimension are both 12 but efficiency's 12 maps to last one instead of second last. That way, it makes sure that we are dividing input xarray by the correct dimension. That's why I only broadcast efficiency data. I hope that helps.

greglucas: I am saying that you can turn the efficiency dataframe into an xarray dataarray that has dimensions ("a", "c") for example, and if the array you're multiplying it against has dimensions ("a", "b", "c", "d") it will just magically work. Also if you have a different array that is ("a", "c", "d"), that will also just work without needing to worry about where newaxis goes for the various cases.

tech3371: I see. That makes more sense. I will look into that.

tech3371: done!

imap_processing/codice/codice_l2.py
```diff
@@ -13,24 +13,194 @@
...
+        species_data = efficiencies_df[efficiencies_df["species"] == species].values
+        species_efficiencies = xr.DataArray(
+            species_data[:, 2:].astype(
+                float
+            ),  # Skip first two columns (species, energy_bin)
+            dims=(f"energy_{species}", "ssd_index"),
+            coords={
+                f"energy_{species}": l2_dataset[f"energy_{species}"],
+                "ssd_index": l2_dataset["ssd_index"],
+            },
```
greglucas: Out of curiosity, can you simplify this by indexing into the original dataset itself? Maybe even removing the dims above if that gets carried along automatically too?
```suggestion
            coords=l2_dataset[[f"energy_{species}", "ssd_index"]]
            },
```

imap_processing/codice/codice_l2.py
greglucas: Nice! That looks good to me. I am hopeful this pattern will help in the future by not having to worry about the shapes across all of the products as that seems like a lot of bookkeeping with the newaxis changing places everywhere.

tech3371: totally. It's much cleaner. and I am learning more about xarray!

imap_processing/codice/codice_l2.py
tech3371: yep. That works!

PR #2291, opened 2025-10-10
I-ALiRT - CoDICE l1a

laspsandoval:
# Change Summary

## Overview
Initial steps for l1a decompression (need to add test data).

## Updated Files
- process_codice.py

From our meeting (Note that the number of packets has changed)
L1a - put 15 bytes together 232 times (0-231) - coming from counter (concatenate) - Lo; 232 is number of packets
L1a - put 5 bytes together 240 times (0-239)  - Hi; 240 is number of packets

232 * 15 * 8 = 27840 bits - Lo
Lo - 24 bits to 32 bits int
Hi - lossy A only

Binary stream will need to be decompressed (input to code)
Lo - Use function that Joey wrote: _apply_pack_24_bit() in  decompress.py
Hi - Lossy A (use the function from Matthew) : _apply_loggy_a() in decompress.py

## Testing
- test_process_codice.py

### DIFF
--- a/imap_processing/ialirt/l0/process_codice.py
+++ b/imap_processing/ialirt/l0/process_codice.py
@@ -6,6 +6,8 @@ from typing import Any
 
 import xarray as xr
 
+from imap_processing.ialirt.utils.grouping import find_groups
+
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


tech3371 (approved):
Looks good to me!

PR #2293, opened 2025-10-13
ULTRA L1C and L2 Flux Validation

lacoak21:
# Change Summary

## Overview
Fixes/ bugs found during l1c/l2 validation with the ULTRA IT team. 

## Updated Files
   - Fixed incorrect unit
- imap_processing/cli.py
   - simplify / clear up logic 
   - return min and max values of theta and phi to add the capability of clipping. During validation there were a few values that were slightly greater/less than and that was causing the fill value to be used. 
   - gf and eff vals were switched. 
   - Add background rates placeholder
  - I read in the numpy documentation that when using the "where" parameter with np.divide it is preferable to use the "out" parameter too. I was seeing some strange results when skipping the "out" parameter.
  - flip eff and gf vals
  - exposure factor must be multiplied by the total number of valid spins in a pointing because it is only calculating values for one simulated spin. 

## Testing
Fix tests that broke with these updates


imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -1119,8 +1121,12 @@ def get_efficiency_interpolator(ancillary_files: dict) -> RegularGridInterpolato
 
     Returns
     -------
-    efficiency : NDArray
-        Interpolated efficiency values.
+    interpolator : RegularGridInterpolator
+        Callable function to interpolate efficiency values.
+    heta_min_max : tuple
```
greglucas:
```suggestion
    theta_min_max : tuple
```

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -440,13 +451,30 @@ def get_spacecraft_exposure_times(
...
+        (spin_data["spin_start_met"] >= pointing_start_met)
+        & (spin_data["spin_start_met"] <= pointing_stop_met)
+    ]
+    # Get only valid spin data
+    valid_mask = (spin_data["spin_phase_valid"].values == 1) & (
+        spin_data["spin_period_valid"].values == 1
+    )
+    total_spin: float = np.sum(
+        spin_data[valid_mask].spin_period_sec / nominal_spin_seconds
+    )
```
greglucas: I'm not following this calculation. Why do you need a nominal spin seconds when you have the actual data? You can get the total spinning time by summing the spin_period_sec column, and you can get the number of spins by doing a "count" to get an integer value here I think or just sum the number of entries in valid_mask.

Could you name it "total_spins" or "n_spins_in_pointing" to be more explicit about what this variable is.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -483,20 +511,27 @@ def get_efficiencies_and_geometric_function(
...
     # load geometric factor lookup table
     geometric_lookup_table = load_geometric_factor_tables(
         ancillary_files, "l1b-sensor-gf-blades"
     )
     # Get energy bin geometric means
     energy_bin_geometric_means = build_energy_bins()[2]
     energy_bins = len(energy_bin_geometric_means)
+    # clip arrays to avoid out of bounds errors
+    theta_vals = np.clip(theta_vals, theta_min_max[0], theta_min_max[1])
+    phi_vals = np.clip(phi_vals, phi_min_max[0], phi_min_max[1])
```
greglucas: Is there an issue with our interpolator that may have caused this out of bounds error? Putting it at the lowest/highest values of the grid seem like we might be putting too many items on the boundary then rather than avoiding them. I'm not sure how the math works out here, but just wanted to say that we could be getting a lot of accumulation into the edge bins with this clipping approach.

Do we want to log anything about how many values were outside the expected range?

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
lacoak21: The exposure time above was calculated for one nominal spin so here is where we scale it by the actual spins. 

From the algorithm document: 

```
We propose the following solution. For each pointing set frame definition, the
exposure time will be approximately the same on a per spin basis assuming no major attitude
instabilities. We precompute and tabulate a static exposure time 𝜏'i
$ for each pixel in the sky in the Pointing frame and correct the total exposure time by the number of valid spins.
```

greglucas: :+1: Could you add some comments about the assumption in there.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
lacoak21: Yeah good idea - Ill add some logs for visibility. Only a very small amount of values should be clipped and it should be a tiny adjustment. The issue was that were were some edge theta values that were -52.72 degrees but the table only went to - 52.70 and the interpolator returned the fill value. I talked to the ULTRA team about this and they wanted to clip the values instead of changing the "out of bounds" handling of the interpolator.

greglucas (approved):

PR #2303, opened 2025-10-15
Update imap data access version to v0.37.0

### LINKED ISSUE #2300: make pointing attitude job fail if no file is produced
### Algorithm Description:

We have seen some jobs that produce nothing because the attitude history and repoint table inputs don't provide combined coverage of any full pointing. Currently this is allowed and the job returns nothing. This behavior screws up the batch starter. This ticket is for making the change such that the job will fail when this condition occurs.

tmplummer:
# Change Summary

## Overview
Bump imap-data-access version

Probably bad form, but I also put a change to the pointing attitude job in here to make it fail when no pointings are covered. This will fix an issue with the batch starter.

Closes: #2300


tmplummer:
pre-commit.ci autofix

tmplummer:
I obviously still don't understand how to get the poetry.lock file to auto-update.

lacoak21 (approved):

tech3371 (approved):

PR #2305, opened 2025-10-16
CODICE l2 angular intensities 

lacoak21:
# Change Summary

## Overview
CODICE l2 angular intensities processing for SW and NSW. Validation data has not been provided yet so that will be in a future PR as well as metadata fixes. 

## Updated Files
   - Compute angular intensity 
   - position to elevation lookup table
- imap_processing/tests/codice/test_codice_l2.py  
   - add test

### DIFF
--- a/imap_processing/codice/codice_l2.py
+++ b/imap_processing/codice/codice_l2.py
@@ -312,11 +312,6 @@ def process_lo_angular_intensity(
     else:
         raise ValueError("Unknown positions for elevation angle mapping.")
 
-    # Create a new coordinate for spin angle based on spin_sector_index
-    # Use equation from section 11.2.2 of algorithm document
-    dataset = dataset.assign_coords(
-        spin_angle=("spin_sector_index", dataset["spin_sector_index"].data * 15.0 + 7.5)
-    )
     # Create a new coordinate for elevation_angle based on azimuth_index
     dataset = dataset.assign_coords(
         elevation_angle=(
@@ -330,6 +325,11 @@ def process_lo_angular_intensity(
         .groupby("elevation_angle")
         .sum(keep_attrs=True)  # One position should always contain zeros so sum is safe
         .transpose("epoch", "energy_table", "spin_sector_index", "elevation_angle", ...)
+    )
+    # Create a new coordinate for spin angle based on spin_sector_index
+    # Use equation from section 11.2.2 of algorithm document
+    dataset = dataset.assign_coords(
+        spin_angle=("spin_sector_index", dataset["spin_sector_index"].data * 15.0 + 7.5)
     )
 
     dataset = dataset.drop_vars(species_list).merge(dataset_converted)
--- a/imap_processing/codice/codice_l2.py
+++ b/imap_processing/codice/codice_l2.py
@@ -324,6 +324,8 @@ def process_lo_angular_intensity(
         dataset[species_list]
         .groupby("elevation_angle")
         .sum(keep_attrs=True)  # One position should always contain zeros so sum is safe
+        # Restore original dimension order because groupby moves the grouped
+        # dimension to the front
         .transpose("epoch", "energy_table", "spin_sector_index", "elevation_angle", ...)
     )
     # Create a new coordinate for spin angle based on spin_sector_index
--- a/imap_processing/codice/codice_l2.py
+++ b/imap_processing/codice/codice_l2.py
@@ -98,7 +98,7 @@ def get_efficiency_lut(dependencies: ProcessingInputCollection) -> pd.DataFrame:
     return pd.read_csv(dependencies.get_file_paths(descriptor="l2-lo-efficiency")[0])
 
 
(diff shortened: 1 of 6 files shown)


imap_processing/codice/codice_l2.py
```diff
@@ -233,11 +237,107 @@ def process_lo_species_intensity(
...
+            continue
+        # Shape: (epoch, esa_steps, positions)
+        denominator = (
+            geometric_factors
+            * species_eff
+            * dataset["energy_table"].data[:, np.newaxis]
+        )
+        # convert denominator to DataArray to leverage xarray broadcasting
+        denominator = xr.DataArray(
+            denominator, dims=("epoch", "energy_table", "azimuth_index")
```
greglucas: Nice!

This entire for loop looks basically identical to the function above (you don't use the xr.DataArray above though. Would it make sense to make any of these things into helper functions or are they just subtly different enough it wouldn't help?

imap_processing/codice/codice_l2.py
```diff
@@ -233,11 +237,107 @@ def process_lo_species_intensity(
...
+        )
+    )
+    # Take the mean across elevation angles and restore the original dimension order
+    dataset_converted = (
+        dataset[species_list]
+        .groupby("elevation_angle")
+        .sum(keep_attrs=True)  # One position should always contain zeros so sum is safe
+        # Restore original dimension order because groupby moves the grouped
+        # dimension to the front
+        .transpose("epoch", "energy_table", "spin_sector_index", "elevation_angle", ...)
```
tech3371: Just curious, is previous dimension before transpose this (epoch, elevation, spin sector)? If so, I plan to switch that at L1A in my refactor to line up in previous level. We should note to revisit this.

imap_processing/codice/codice_l2.py
```diff
@@ -233,11 +237,107 @@ def process_lo_species_intensity(
...
+    )
```
greglucas, line 380: Can you do a `.sum(dim="elevation_angle", keep_attrs=True)` so you are summing over the dimension instead of needing to groupby?

greglucas (approved):
Minor nit is all, looks good.

Do you need to upload/rename the external tests for this to pass the remote runner?

imap_processing/tests/external_test_data_config.py
```diff
@@ -60,10 +60,10 @@
     ("imap_codice_l1b_lo-counters-aggregated_20250814211100_v0.0.5.cdf", "codice/data/l1b_validation"),
     ("imap_codice_l1b_lo-counters-singles_20250814211100_v0.0.5.cdf", "codice/data/l1b_validation"),
     ("imap_codice_l1b_lo-ialirt_20250814211100_v0.0.5.cdf", "codice/data/l1b_validation"),
-    ("imap_codice_l1b_lo-nsw-angular_20250814211100_v0.0.5.cdf", "codice/data/l1b_validation"),
+    ("imap_codice_l1b_lo-nsw-angular_20250814_v005.cdf", "codice/data/l1b_validation"),
```
tech3371: I believe this needs to be v006

imap_processing/codice/codice_l2.py
```diff
@@ -233,11 +237,107 @@ def process_lo_species_intensity(
...
+    # Create a new coordinate for spin angle based on spin_sector_index
+    # Use equation from section 11.2.2 of algorithm document
+    dataset = dataset.assign_coords(
+        spin_angle=("spin_sector_index", dataset["spin_sector_index"].data * 15.0 + 7.5)
```
tech3371: is that the final L2 dimension name? If so, looks like we got different names. In hi L2 products, it's called 'spin_sector' but for lo, looks like it's called 'spin_sector_index'. Can we ask Joey to match?

tech3371 (approved):
minor comments, otherwise, looks good to me

imap_processing/codice/codice_l2.py
lacoak21: No- what happened is that groupby moves that dimension to the front so it was (elevation, epoch, energy, spin_sector)

imap_processing/codice/codice_l2.py
lacoak21: Yes they have subtle differences but ill look into it because the basic operation is the same. There are just different dimensions at play.

lacoak21: But now that I converted it to a DataArray I think that should handle those differences nicely.

greglucas: Yeah, agreed, that is the hope at least! Even if you have slightly different ordering of dimensions, that will 🤞 get taken care of under the hood.

imap_processing/codice/codice_l2.py
lacoak21: Sum collapses the entire dimension but I need that dimension still just with some positions summed which converts it to elevation angle dimension

imap_processing/codice/codice_l2.py
lacoak21: Ok that worked smoothly!

PR #2308, opened 2025-10-16
2249 hi lo cg correction   bin pset data 1

### LINKED ISSUE #2249: Hi Lo CG Correction - Bin PSET data
### Requirements:

For each helio-frame energy step, bin corresponding PSET bins into the AbstractSkyMap object. Bin the following values:
- exposure weighted counts
- exposure weighted backgrounds
- exposure weighted uncertainties
- exposure weighted S/C energy
- exposure time

tmplummer:
# Change Summary

## Overview
This PR includes changes needed to get Hi PSET data through the projection step of L2 processing with a CG correction. Major updates are:
- Add a ram_mask boolean variable to the PSET dataset which is used to filter ram/anti-ram pixels during the projection process.
- Enable xarray broadcasting of valid_pixel_mask during map projection. This is to support having a ram_mask that is different for each energy step.
- Modify CG correction functions to return the modified LoHiBasePointingSet object to make it obvious that they mutate the input PSET object.
- Update Hi L2 code to call CG correction function if the map descriptor calls for it.

## Updated Files
   - Add `ram_mask` variable to HiPointingSet class.
   - support xarray.DataArray input and xarray broadcasting for pset_valid_mask argument in `project_pset_values_to_map` function.
   - Return mutated PointingSet object from CG correction functions
   - Acrobatics to get mypy type checking happy
 - imap_processing/hi/hi_l2.py
   - Add check that all input pointing sets contain the same set of esa_energy_steps.
   - Add call to CG correction functions that act on PSET data before projecting.
   - Set pass appropriate `valid_pixel_mask` in to projection function so that ram/anti-ram filtering will work.
   - update calls to CG correction functions with mutated pset as output
   - Add test coverage for projecting with xarray.DataArray pset_valid_pixel mask.
   - Add testing of ram, anti, and helio frame maps

Closes: #2249

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -652,15 +652,16 @@ class HiPointingSet(LoHiBasePointingSet):
     def __init__(self, dataset: xr.Dataset | str | Path):
         super().__init__(dataset, spice_reference_frame=geometry.SpiceFrame.IMAP_HAE)
 
+        self.spatial_coords = ("spin_angle_bin",)
+
         # Naively generate the ram_mask variable assuming spacecraft frame
         # binning. The ram_mask variable gets updated in the CG correction
         # code if the CG correction is applied.
-        self.data["ram_mask"] = xr.zeros_like(self.data["spin_angle_bin"], dtype=bool)
+        ram_mask = xr.zeros_like(self.data["spin_angle_bin"], dtype=bool)
         # ram only includes spin-phase interval [0, 0.5)
         # which is the first half of the spin_angle_bins
-        self.data["ram_mask"][slice(0, self.data["spin_angle_bin"].data.size // 2)] = (
-            True
-        )
+        ram_mask[slice(0, self.data["spin_angle_bin"].data.size // 2)] = True
+        self.data["ram_mask"] = ram_mask
 
         # Rename some PSET vars to match L2 variables
         self.data = self.data.rename(
@@ -675,8 +676,6 @@ class HiPointingSet(LoHiBasePointingSet):
         self.data["obs_date"] = xr.full_like(
             self.data["exposure_factor"], self.data["epoch"].values[0]
         )
-
-        self.spatial_coords = ("spin_angle_bin",)
 
         # Update az_el_points using the base class method
         self.update_az_el_points()
@@ -802,12 +801,12 @@ class AbstractSkyMap(ABC):
         """
         return self.az_el_points.shape[0]
 
-    def project_pset_values_to_map(
+    def project_pset_values_to_map(  # noqa: PLR0912
         self,
         pointing_set: PointingSet,
         value_keys: list[str] | None = None,
         index_match_method: IndexMatchMethod = IndexMatchMethod.PUSH,
-        pset_valid_mask: NDArray | None = None,
+        pset_valid_mask: NDArray | xr.DataArray | None = None,
     ) -> None:
         """
         Project a pointing set's values to the map grid.
@@ -829,7 +828,7 @@ class AbstractSkyMap(ABC):
         index_match_method : IndexMatchMethod, optional
             The method of index matching to use for all values.
             Default is IndexMatchMethod.PUSH.
-        pset_valid_mask : NDArray, optional
+        pset_valid_mask : xarray.DataArray or NDArray, optional
             A boolean mask of shape (number of pointing set pixels,) indicating
             which pixels in the pointing set should be considered valid for projection.
             If None, all pixels are considered valid. Default is None.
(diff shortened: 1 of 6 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -849,9 +840,6 @@ def project_pset_values_to_map(
         """
         if value_keys is None:
             value_keys = list(pointing_set.data.data_vars.keys())
-        for value_key in value_keys:
-            if value_key not in pointing_set.data.data_vars:
```
tmplummer, line 853: This check was moved into the for loop below.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -934,10 +934,6 @@ def project_pset_values_to_map(
                 self.data_1d[value_key].values[..., valid_map_mask] += (
                     pointing_projected_values
                 )
-            else:
-                raise NotImplementedError(
-                    "Only PUSH and PULL index matching methods are supported."
-                )
```
tmplummer, line 940: This check was redundant. An identical check happens earlier in this same function.

imap_processing/ena_maps/ena_maps.py
greglucas: I think this could be beneficial to keep outside so that you don't do all the computations of the first 5 value_keys and then raise on the 6th. Instead of the for-loop, you could do something like this:
```
if missing_keys := set(value_keys) - set(pointing_set.data.data_vars):
    raise KeyError(f"Value keys not found in dataset: {missing_keys}")
```

imap_processing/hi/hi_l2.py
```diff
@@ -98,34 +101,64 @@ def generate_hi_map(
...
     for pset_path in psets:
         logger.info(f"Processing {pset_path}")
-        pset = HiPointingSet(pset_path, spin_phase=descriptor.spin_phase)
-
-        # Background rate and uncertainty are exposure time weighted means in
-        # the map.
+        pset = HiPointingSet(pset_path)
+
+        # Store the first PSET esa_energy_step values and make sure every PSET
+        # contains the same set of esa_energy_step values.
```
greglucas, line 129: Is this a hard requirement? I'm wondering if you can just recompute what you need to if the cache changes.

```
if not np.array_equal(cached_esa_steps, pset.data["esa_energy_step"].values):
    # recompute cache
```

imap_processing/hi/hi_l2.py
```diff
@@ -98,34 +101,64 @@ def generate_hi_map(
...
-            ["counts", "exposure_factor", "bg_rates", "bg_rates_unc", "obs_date"],
+            pset, vars_to_bin, pset_valid_mask=pset_valid_mask
         )
 
     # Finish the exposure time weighted mean calculation of backgrounds
     # Allow divide by zero to fill set pixels with zero exposure time to NaN
     with np.errstate(divide="ignore"):
         for var in VARS_TO_EXPOSURE_TIME_AVERAGE:
-            output_map.data_1d[var] /= output_map.data_1d["exposure_factor"]
+            if var in output_map.data_1d:
```
greglucas: Is this because your new variable isn't always in every output dataset? i.e. in a non-CG map you don't have your CG variable and therefore you need to skip it here? I wonder if it would make sense to dynamically update that global variable to add something to the VARS_TO_EXPOSURE_TIME_AVERAGE in the CG case instead of this if condition here. I'm not sure I have a major preference, this just sticks out to me as "what if we forget to add our var in the output_map and we skip this if check in the normal situation.

imap_processing/hi/hi_l2.py
```diff
@@ -441,4 +462,6 @@ def esa_energy_df(
     esa_energies_lut = pd.read_csv(
         esa_energies_path, comment="#", index_col="esa_energy_step"
     )
+    if esa_energy_steps is None:
+        return esa_energies_lut
```
greglucas: Another option to not have to special case None I think is to use `esa_energy_steps = slice(None)` upon entry. So in your code above set the slice().

greglucas (approved):
Overall looks like some nice consolidation again! A few minor suggestions, but the meat of it looks great.

imap_processing/hi/hi_l2.py
tmplummer: It is currently a hard requirement because the projection code only bins by spatial pixel. It requires that the variable being projected has the same dimension sizes as the map variable that it projects into. Now that I am writing this, I actually don't know if it is true. I don't know enough about the xarray broadcasting to know what will happen if one of the PSETs is missing an energy step. I know that it can't handle the case where the first PSET is missing an energy step because the map variables are instantiated based on the coordinates of the first PSET.

imap_processing/hi/hi_l2.py
tmplummer: Yes. On lines 104-107 I define what variables to project to the map. The CG map has the added `energy_sc` variable that gets projected and needs to be exposure time averaged. I try really hard not to mutate global variables but also see you point about a missing variable not getting caught here.

imap_processing/ena_maps/ena_maps.py
tmplummer: Nice walrus operator!

imap_processing/hi/hi_l2.py
tmplummer: I'm going to leave this for now and write a ticket to handle when a PSET doesn't have the same set of esa_energy_step.

PR #2311, opened 2025-10-18
Lo L1B - Re-Sweep Counts and Calculate Histogram Rates

sdhoyt:
# Change Summary

## Overview
This PR takes the L1A histogram counts and re-sweeps the ESA steps based on the sweep table and LUT ancillary files. Those counts are then divided by exposure time which is calculated from the L1A spin data.

(Sorry for the size of this PR, I put both the counts and rates in here to move things along quicker.)

Histogram Rate future PRs:
- Apply Badtime filtering
- Add Spin Cycle field
- Add ESA Mode field


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -988,3 +1026,270 @@ def create_datasets(
...
+    # Get epoch values
+    epochs = l1b_histrates["epoch"].values
+
+    # Initialize corrected data arrays
+    h_counts_reswept = np.zeros_like(l1b_histrates["h_counts"].values)
+    o_counts_reswept = np.zeros_like(l1b_histrates["o_counts"].values)
+
+    # Process each epoch
+    for epoch_idx, epoch_value in enumerate(epochs):
+        epoch_date = et_to_utc(ttj2000ns_to_et(epoch_value))
```
greglucas: I think these are vectorized functions, so we should calculate the list of dates all at once up front.

```suggestion
    epoch_dates = et_to_utc(ttj2000ns_to_et(epochs))
    for epoch_idx, epoch_date in enumerate(epoch_dates):
        
```

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -988,3 +1026,270 @@ def create_datasets(
...
+            # The output array still has 7 elements representing energy steps 1-7
+            for energy_level, h_count in energy_level_counts_h.items():
+                if 1 <= energy_level <= 7:  # Ensure valid energy level
+                    reswept_idx = energy_level - 1  # Convert to 0-based index
+                    h_reswept_az[reswept_idx] = h_count
+                    o_reswept_az[reswept_idx] = energy_level_counts_o[energy_level]
+
+            # Store corrected data
+            h_counts_reswept[epoch_idx, az_idx, :] = h_reswept_az
+            o_counts_reswept[epoch_idx, az_idx, :] = o_reswept_az
```
greglucas: I find this block a little bit hard to follow. If I'm understanding correctly you've already identified your mapping above. Rather than creating new intermediate arrays here can you add directly to the mapping indices you've defined previously?

Rather than `h_counts_reswept[epoch_idx, az_idx, :] = h_reswept_az` Could you update the specific reswept level you need as you're iterating through the steps/levels: `h_counts_reswept[epoch_idx, az_idx, reswept_idx]`. Currently it seems like there is one more intermediate loop than their needs to be.

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -988,3 +1026,270 @@ def create_datasets(
...
+        exposure_time = 4 * avg_spin_durations_per_cycle.values[spin_cycle_idx] / 60
+        exposure_times[epoch_idx] = exposure_time
+
+        # Calculate rates by dividing counts by exposure time
+        # Avoid division by zero
+        if exposure_time > 0:
+            h_rates[epoch_idx] = h_counts[epoch_idx] / exposure_time
+            o_rates[epoch_idx] = o_counts[epoch_idx] / exposure_time
+        else:
+            raise ValueError(f"Zero exposure time at epoch {epoch_idx}")
```
greglucas: Do we want to error here or set it to np.nan?

You might want to put this in a `np.seterr()` context manager to ignore divide-by-zero or just let it warn and produce infs? That way we can still produce data products but with easily findable bad values.

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -988,3 +1026,270 @@ def create_datasets(
...
+            o_rates[epoch_idx] = np.nan
+            continue
+
+        # Use the first matching spin_cycle
+        spin_cycle_idx = spin_cycle_indices[0]
+
+        # Calculate exposure time for 6-degree bins
+        # Lo Algorithm document V10 Section 9.6.4
+        # 4 * avg_spin_duration / 60
+        exposure_time = 4 * avg_spin_durations_per_cycle.values[spin_cycle_idx] / 60
```
sdhoyt: Exposure time needs to scale based on sweep table ESA steps

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -988,3 +1026,270 @@ def create_datasets(
...
+    # Initialize corrected data arrays
+    h_counts_reswept = np.zeros_like(l1b_histrates["h_counts"].values)
+    o_counts_reswept = np.zeros_like(l1b_histrates["o_counts"].values)
+
+    # Process each epoch
+    for epoch_idx, epoch_value in enumerate(epochs):
+        epoch_date = et_to_utc(ttj2000ns_to_et(epoch_value))
+        epoch_date_only = epoch_date.split("T")[0]
+        # Find matching LUT table for this date
+        if epoch_date_only not in date_to_lut:
```
tmplummer: It looks like this is the only place that this dictionary is used. You could skip creating the dictionary and do
```
if epoch_date_only not in sweep_dates:
    ....
```

imap_processing/lo/l1b/lo_l1b.py
tmplummer: +1 on this comment. I think that the loop over energy_step_mapping can just be:
```
for orig_idx, true_esa_step in energy_step_mapping.items():
                if orig_idx < len(h_original):  # Ensure index is valid
                    if true_esa_step not in energy_level_counts_h:
                        energy_level_counts_h[true_esa_step] = 0
                        energy_level_counts_o[true_esa_step] = 0

                    h_counts_reswept[epoch_idx, az_idx, true_esa_step - 1] += h_original[orig_idx]
                    o_counts_reswept[epoch_idx, az_idx, true_esa_step - 1] += o_original[orig_idx]
```

tmplummer (commented):
Just a couple of suggestions for small improvements.

sdhoyt:
I added scaling for the exposure time based on the resweeping. This complicated the code a bit and it could benefit from some refactoring, but it seems to be working. Based on schedule issues I'm thinking we'd want to come back to any major refactoring that is suggested.

sdhoyt:
Ignore the latest changes. Somehow my rebase wiped away some of my changes.

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1004,3 +1045,247 @@ def create_datasets(
...
+        dims=["epoch"],
+        # TODO: Add hydrogen to YAML file
+        # attrs=attr_mgr.get_variable_attributes("hydrogen"),
+    )
+    l1b_histrates["o_counts"] = xr.DataArray(
+        l1a_hist["oxygen"].values,
+        dims=["epoch"],
+        # TODO: Add oxygen to YAML file
+        # attrs=attr_mgr.get_variable_attributes("oxygen"),
+    )
```
greglucas, line 1207: Here, if there are already attributes set in L1a I think you can do a direct insertion and get those transferred as well potentially.

```suggestion
    # Copy over fields from L1A DE that will not change in L1B processing
    l1b_histrates["h_counts"] = l1a_hist["hydrogen"]
    l1b_histrates["o_counts"] = l1a_hist["oxygen"]
```

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1004,3 +1045,247 @@ def create_datasets(
...
+            esa_idx = int(row["E-Step_Idx"]) - 1
+            true_esa_step = int(row["E-Step_lvl"])
+            energy_step_mapping[esa_idx] = true_esa_step
+
+        for az_idx in range(num_azimuth):
+            h_original = l1b_histrates["h_counts"].values[epoch_idx, az_idx, :]
+            o_original = l1b_histrates["o_counts"].values[epoch_idx, az_idx, :]
+
+            for orig_idx, true_esa_step in energy_step_mapping.items():
+                if orig_idx < len(h_original) and 1 <= true_esa_step <= 7:
```
greglucas, line 1318: Do we want to log if this isn't the case?

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1004,3 +1045,247 @@ def create_datasets(
...
+                if scaled_exposure > 0:
+                    h_rates[epoch_idx, az_idx, esa_idx] = (
+                        h_counts[epoch_idx, az_idx, esa_idx] / scaled_exposure
+                    )
+                    o_rates[epoch_idx, az_idx, esa_idx] = (
+                        o_counts[epoch_idx, az_idx, esa_idx] / scaled_exposure
+                    )
+                else:
+                    h_rates[epoch_idx, az_idx, esa_idx] = 0
+                    o_rates[epoch_idx, az_idx, esa_idx] = 0
```
greglucas: You can get rid of the two for loops here pretty easily so we should do this.

If we don't have any exposure time do we actually want 0 rates? I would have assumed we want NaNs there since a 0 rate could be a real value. I'd suggest just ignoring the divide by zero warning and let the nan show up.
```suggestion
                scaled_exposure = (
                    base_exposure_time * exposure_factor[epoch_idx, ...]
                )
                exposure_times[epoch_idx, ...] = scaled_exposure

                with np.errstate(divide='ignore'):
                    h_rates[epoch_idx, ...] = (
                        h_counts[epoch_idx, ...] / scaled_exposure
                    )
                    o_rates[epoch_idx, ...] = (
                        o_counts[epoch_idx, ...] / scaled_exposure
                    )
```

greglucas (approved):
I do have one minor refactor suggestion that I think should be done but not worth blocking over and I think the logic is good.

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1004,3 +1045,247 @@ def create_datasets(
...
+    h_rates = np.zeros_like(h_counts, dtype=float)
+    o_rates = np.zeros_like(o_counts, dtype=float)
+    num_azimuth = h_counts.shape[1]
+    exposure_times = np.zeros((len(epochs), num_azimuth, 7), dtype=float)
+
+    for epoch_idx, epoch in enumerate(epochs):
+        spin_cycle_mask = (epoch >= met_to_ttj2000ns(acq_start.values)) & (
+            epoch <= met_to_ttj2000ns(acq_end.values)
+        )
+        spin_cycle_indices = np.where(spin_cycle_mask)[0]
```
tmplummer: `np.nonzero()` is preferred when used like this. https://numpy.org/doc/2.3/reference/generated/numpy.where.html
```suggestion
        spin_cycle_indices = np.nonzero(spin_cycle_mask)[0]
```

tmplummer (commented):
Just the one suggestion from me.

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Agreed, but I'll create a ticket and do that in a separate PR. I need to change all of L1A histogram to use spin_bin instead of azimuth. To avoid having azimuth carry over into L1B right now, I'll make this change in that PR.

https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/2462

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Yeah I'll add a log

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Good point. I'll fix this

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Oh yeah, I think you told me this before and I forgot. Thanks!

sdhoyt:
pre-commit.ci autofix

Issue #2269, opened 2025-10-01
CoDICE: L1A Lo species refactor


tech3371:
- [ ] refactor l1a species and both SW and NSW

tech3371:
Waiting on Time corrected files and validate again.

PR #2313, opened 2025-10-20
CoDICE: L1a first refactor of species and L1A utils functions

### LINKED ISSUE #1501: CoDICE: epoch times should reflect center of acquisition time bins

tech3371:
# Change Summary
Related to #2269

## Overview

First take on L1A refactoring. I am checking few things with Joey and Michael. Few things to come still.

## Updated Files

- codice_new_l1a.py
   - species processing code for both SW and NSW
- utils.py
   - This is where I added utils function to read SCI_LUT and calculated common utils functions for l1a.

## Testing

Test updates and refactoring
- imap_processing/tests/codice/conftest.py
- imap_processing/tests/codice/test_codice_hi_l2.py
- imap_processing/tests/codice/test_codice_l1a_lut.py
- imap_processing/tests/codice/test_codice_l2.py

### DIFF
--- a/imap_processing/codice/codice_l1a_lo_species.py
+++ b/imap_processing/codice/codice_l1a_lo_species.py
@@ -1,0 +1,275 @@
+"""CoDICE Lo Species L1A processing functions."""
+
+import logging
+from pathlib import Path
+
+import numpy as np
+import xarray as xr
+
+from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
+from imap_processing.codice import constants
+from imap_processing.codice.decompress import decompress
+from imap_processing.codice.utils import (
+    CODICEAPID,
+    ViewTabInfo,
+    calculate_acq_time_per_step,
+    get_codice_epoch_time,
+    get_collapse_pattern_shape,
+    get_view_tab_info,
+    read_sci_lut,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def l1a_lo_species(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
+    """
+    L1A processing code.
+
+    Parameters
+    ----------
+    unpacked_dataset : xarray.Dataset
+        The decompressed and unpacked data from the packet file.
+    lut_file : pathlib.Path
+        Path to the LUT (Lookup Table) file used for processing.
+
+    Returns
+    -------
+    xarray.Dataset
+        The processed L1A dataset for the given species product.
+    """
+    # Get these values from unpacked data. These are used to
+    # lookup in LUT table.
+    table_id = unpacked_dataset["table_id"].values[0]
+    view_id = unpacked_dataset["view_id"].values[0]
+    apid = unpacked_dataset["pkt_apid"].values[0]
+    plan_id = unpacked_dataset["plan_id"].values[0]
+    plan_step = unpacked_dataset["plan_step"].values[0]
+
+    logger.info(
+        f"Processing species with - APID: {apid}, View ID: {view_id}, "
+        f"Table ID: {table_id}, Plan ID: {plan_id}, Plan Step: {plan_step}"
+    )
+
+    # ========== Get LUT Data ===========
+    # Read information from LUT
+    sci_lut_data = read_sci_lut(lut_file, table_id)
+
+    view_tab_info = get_view_tab_info(sci_lut_data, view_id, apid)
+    view_tab_obj = ViewTabInfo(
+        apid=apid,
+        view_id=view_id,
+        sensor=view_tab_info["sensor"],
+        three_d_collapsed=view_tab_info["3d_collapse"],
+        collapse_table=view_tab_info["collapse_table"],
+    )
+
+    if view_tab_obj.sensor != 0:
+        raise ValueError("Unsupported sensor ID for Lo species processing.")
+
+    # ========= Decompress and Reshape Data ===========
+    # Lookup SW or NSW species based on APID
+    if view_tab_obj.apid == CODICEAPID.COD_LO_SW_SPECIES_COUNTS:
+        species_names = sci_lut_data["data_product_lo_tab"]["0"]["species"]["sw"][
+            "species_names"
+        ]
+        logical_source_id = "imap_codice_l1a_lo-sw-species"
+    elif view_tab_obj.apid == CODICEAPID.COD_LO_NSW_SPECIES_COUNTS:
+        species_names = sci_lut_data["data_product_lo_tab"]["0"]["species"]["nsw"][
+            "species_names"
+        ]
+        logical_source_id = "imap_codice_l1a_lo-nsw-species"
+    else:
+        raise ValueError(f"Unknown apid {view_tab_obj.apid} in Lo species processing.")
+
+    compression_algorithm = constants.LO_COMPRESSION_ID_LOOKUP[view_tab_obj.view_id]
+    # Decompress data using byte count information from decommed data
+    binary_data_list = unpacked_dataset["data"].values
+    byte_count_list = unpacked_dataset["byte_count"].values
+
+    # The decompressed data in the shape of (epoch, n). Then reshape later.
+    decompressed_data = [
+        decompress(
+            packet_data[:byte_count],
+            compression_algorithm,
+        )
+        for (packet_data, byte_count) in zip(
+            binary_data_list, byte_count_list, strict=False
+        )
+    ]
+
+    # Look up collapse pattern using LUT table. This should return collapsed shape.
+    # For Lo species, it will be (1,)
+    collapsed_shape = get_collapse_pattern_shape(
+        sci_lut_data, view_tab_obj.sensor, view_tab_obj.collapse_table
+    )
+
+    # Reshape decompressed data to:
+    #   (num_packets, num_species, esa_steps, *collapsed_shape)
+    # where collapsed_shape is usually (1,) for Lo species.
+    num_packets = len(binary_data_list)
+    num_species = len(species_names)
+    esa_steps = constants.NUM_ESA_STEPS
+    species_data = np.array(decompressed_data).reshape(
+        num_packets, num_species, esa_steps, *collapsed_shape
+    )
+
+    # ========== Get Voltage Data from LUT ===========
+    # Use plan id and plan step to get voltage data's table_number in ESA sweep table.
+    # Voltage data is (128,)
+    esa_table_number = (
+        sci_lut_data["plan_tab"].get(f"({plan_id}, {plan_step})").get("lo_stepping")
+    )
+    voltage_data = sci_lut_data["esa_sweep_tab"].get(f"{esa_table_number}")
+
+    # ========= Get Epoch Time Data ===========
+    # Epoch center time and delta
+    epoch_center, deltas = get_codice_epoch_time(
+        unpacked_dataset["acq_start_seconds"].values,
+        unpacked_dataset["acq_start_subseconds"].values,
+        unpacked_dataset["spin_period"].values,
+        view_tab_obj,
+    )
+
+    # ========== Create CDF Dataset with Metadata ===========
+    cdf_attrs = ImapCdfAttributes()
+    cdf_attrs.add_instrument_global_attrs("codice")
+    cdf_attrs.add_instrument_variable_attrs("codice", "l1a")
+
+    l1a_dataset = xr.Dataset(
+        coords={
+            "epoch": xr.DataArray(
+                epoch_center,
+                dims=("epoch",),
+                attrs=cdf_attrs.get_variable_attributes("epoch", check_schema=False),
+            ),
+            "epoch_delta_minus": xr.DataArray(
+                deltas,
+                dims=("epoch",),
+                attrs=cdf_attrs.get_variable_attributes(
+                    "epoch_delta_minus", check_schema=False
+                ),
+            ),
+            "epoch_delta_plus": xr.DataArray(
+                deltas,
+                dims=("epoch",),
+                attrs=cdf_attrs.get_variable_attributes(
+                    "epoch_delta_plus", check_schema=False
+                ),
+            ),
+            "esa_step": xr.DataArray(
+                np.arange(128),
+                dims=("esa_step",),
+                attrs=cdf_attrs.get_variable_attributes("esa_step", check_schema=False),
+            ),
+            "esa_step_label": xr.DataArray(
+                np.arange(128).astype(str),
+                dims=("esa_step",),
+                attrs=cdf_attrs.get_variable_attributes(
+                    "esa_step_label", check_schema=False
+                ),
+            ),
+            "spin_sector": xr.DataArray(
+                np.array([0], dtype=np.uint8),
+                dims=("spin_sector",),
+                attrs=cdf_attrs.get_variable_attributes(
+                    "spin_sector", check_schema=False
+                ),
+            ),
+            "spin_sector_label": xr.DataArray(
+                np.array(["0"]).astype(str),
+                dims=("spin_sector",),
+                attrs=cdf_attrs.get_variable_attributes(
+                    "spin_sector_label", check_schema=False
+                ),
+            ),
+        },
+        attrs=cdf_attrs.get_global_attributes(logical_source_id),
+    )
+    # Add first few unique variables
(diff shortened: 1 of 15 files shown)


imap_processing/codice/utils.py
```diff
@@ -57,3 +96,205 @@ class CoDICECompression(IntEnum):
...
+        # all non-zero values are identical means -> fully collapsed
+        return (1,)
+
+    # If not fully collapsed, find repeated patterns in rows and columns
+    # to reduce shape further.
+    unique_rows = np.unique(non_zero_reformatted, axis=0)
+    unique_columns = np.unique(non_zero_reformatted, axis=1)
+    # TODO: flip this dimension for final dataset to be below?
+    # (spin_sector, position or ssd id)
+    return (unique_rows.shape[0], unique_columns.shape[1])
```
tech3371: column is spin sector and row is az.
```suggestion
    return (unique_columns.shape[1], unique_rows.shape[0])
```

imap_processing/codice/codice_l1a_lo_species.py
```diff
@@ -0,0 +1,285 @@
...
+        sci_lut_data, view_tab_obj.sensor, view_tab_obj.collapse_table
+    )
+
+    # Reshape decompressed data to:
+    #   (num_packets, num_species, esa_steps, *collapsed_shape)
+    num_packets = len(binary_data_list)
+    num_species = len(species_names)
+    esa_steps = 128
+    species_data = np.array(decompressed_data).reshape(
+        num_packets, num_species, esa_steps, *collapsed_shape
```
tech3371, line 114:
```suggestion
        num_packets, num_species, esa_steps, (spin_sector, az)
```

tech3371: decompressed data (epoch, n) where n = num_species * 128 * 1

imap_processing/codice/codice_l1a_lo_species.py
```diff
@@ -0,0 +1,286 @@
...
+    xarray.Dataset
+        The processed L1A dataset for the given species product.
+    """
+    # Get these values from unpacked data. These are used to
+    # lookup in LUT table.
+    table_id = np.unique(unpacked_dataset["table_id"].values)[0]
+    view_id = np.unique(unpacked_dataset["view_id"].values)[0]
+    apid = np.unique(unpacked_dataset["pkt_apid"].values)[0]
+    plan_id = np.unique(unpacked_dataset["plan_id"].values)[0]
+    plan_step = np.unique(unpacked_dataset["plan_step"].values)[0]
```
greglucas: Why do the unique call here if you are only taking the first value? I think these can be `.values[0]` since that is guaranteed to be your first unique value too.

imap_processing/codice/codice_l1a_lo_species.py
```diff
@@ -0,0 +1,286 @@
...
+    logger.info(
+        f"Processing species with - APID: {apid}, View ID: {view_id}, "
+        f"Table ID: {table_id}, Plan ID: {plan_id}, Plan Step: {plan_step}"
+    )
+
+    # ========== Get LUT Data ===========
+    # Read information from LUT
+    sci_lut_data = read_sci_lut(lut_file, table_id)
+    if sci_lut_data is None:
+        raise ValueError(f"SCI-LUT file does not have data for table ID {table_id}.")
```
greglucas: Raise this error from within the read_sci_lut function instead of returning None and requiring handling externally.

imap_processing/codice/codice_l1a_lo_species.py
```diff
@@ -0,0 +1,286 @@
...
+        collapse_table=view_tab_info["collapse_table"],
+    )
+
+    if view_tab_obj.sensor != 0:
+        raise ValueError("Unsupported sensor ID for Lo species processing.")
+
+    # ========= Decompress and Reshape Data ===========
+    # Lookup SW or NSW species based on APID
+    if view_tab_obj.apid == CODICEAPID.COD_LO_SW_SPECIES_COUNTS:
+        species_names = sci_lut_data["data_product_lo_tab"]["0"]["species"]["sw"][
```
greglucas, line 73: string "0" here, should that be an integer instead?

imap_processing/codice/codice_l1a_lo_species.py
```diff
@@ -0,0 +1,286 @@
...
+            "species_names"
+        ]
+        logical_source_id = "imap_codice_l1a_lo-sw-species"
+    elif view_tab_obj.apid == CODICEAPID.COD_LO_NSW_SPECIES_COUNTS:
+        species_names = sci_lut_data["data_product_lo_tab"]["0"]["species"]["nsw"][
+            "species_names"
+        ]
+        logical_source_id = "imap_codice_l1a_lo-nsw-species"
+    else:
+        logger.error("Unknown apid in Lo species processing.")
```
greglucas: raise here instead of log. I'm guessing this won't succeed after this. Also include what the unknown apid is in the message.
```suggestion
        raise ValueError(f"Unknown apid {view_tab_obj.apid} in Lo species processing.")
```

imap_processing/codice/codice_l1a_lo_species.py
```diff
@@ -0,0 +1,286 @@
...
+    collapsed_shape = get_collapse_pattern_shape(
+        sci_lut_data, view_tab_obj.sensor, view_tab_obj.collapse_table
+    )
+
+    # Reshape decompressed data to:
+    #   (num_packets, num_species, esa_steps, *collapsed_shape)
+    # where collapsed_shape is usually (1,) for Lo species.
+    num_packets = len(binary_data_list)
+    num_species = len(species_names)
+    esa_steps = 128
```
greglucas: Is this a constant that should be added to the constants area?

imap_processing/codice/codice_l1a_lo_species.py
```diff
@@ -0,0 +1,286 @@
...
+                dims=("spin_sector",),
+                attrs=cdf_attrs.get_variable_attributes(
+                    "spin_sector_label", check_schema=False
+                ),
+            ),
+            "epoch": xr.DataArray(
+                epoch_center,
+                dims=("epoch",),
+                attrs=cdf_attrs.get_variable_attributes("epoch", check_schema=False),
+            ),
```
greglucas: nit: Can you move this up to the top so that these variables are in the order of dimensions in the dataset.

imap_processing/codice/codice_l1a_lo_species.py
```diff
@@ -0,0 +1,286 @@
...
+        direction = (
+            "Sunward"
+            if view_tab_obj.apid == CODICEAPID.COD_LO_SW_SPECIES_COUNTS
+            else "Non-Sunward"
+        )
+        # Replace {species} and {direction} in attrs
+        species_attrs["CATDESC"] = (
+            species_attrs["CATDESC"]
+            .replace("{species}", species)
+            .replace("{direction}", direction)
```
greglucas: I think you can call the format function on these.
```suggestion
            species_attrs["CATDESC"]
            .format(species=species, direction=direction)
```

imap_processing/codice/codice_new_l1a.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    # Decom packet
+    datasets_by_apid = packet_file_to_datasets(
+        science_file,
+        xtce_file,
+    )
+
+    datasets = []
+    for apid in datasets_by_apid:
+        if CODICEAPID.COD_LO_SW_SPECIES_COUNTS == apid:
```
greglucas: nit: swap these so your variable you are testing is first (I think you are doing it that way in other places)
```suggestion
        if apid == CODICEAPID.COD_LO_SW_SPECIES_COUNTS:
```

imap_processing/codice/utils.py
```diff
@@ -5,7 +5,46 @@
...
+import json
 from enum import IntEnum
+from pathlib import Path
+
+import numpy as np
+
+from imap_processing.spice.time import met_to_ttj2000ns
+
+
+class ViewTabInfo:
```
greglucas, line 19: Looks like a dataclass or namedtuple?
```suggestion
@dataclass
class ViewTabInfo:
        apid: int
        view_id: int
        sensor: int
        three_d_collapsed: int
        collapse_table: int
```

imap_processing/codice/utils.py
```diff
@@ -57,3 +96,203 @@ class CoDICECompression(IntEnum):
...
+
+    if np.unique(non_zero_reformatted).size == 1:
+        # all non-zero values are identical means -> fully collapsed
+        return (1,)
+
+    # If not fully collapsed, find repeated patterns in rows and columns
+    # to reduce shape further.
+    unique_rows = np.unique(non_zero_reformatted, axis=0)
+    unique_columns = np.unique(non_zero_reformatted, axis=1)
+    return (unique_columns.shape[1], unique_rows.shape[0])
```
greglucas: Why return this as (columns, rows) instead of (rows, columns)?

I remember talking through this function, and I'm still lost so can't do a great review on this. It looks alright, but seems like a lot of unique calls. Maybe there is a clever way to do this, but it if gives you what you want this seems pretty reasonable.

imap_processing/codice/utils.py
```diff
@@ -57,3 +96,203 @@ class CoDICECompression(IntEnum):
...
+        # Use constant 16 spins for Hi PHA
+        num_spins = 16.0
+    # If Non-Direct Event Hi product
+    else:
+        # Use 3d_collapsed value from LUT for other Hi products
+        num_spins = view_tab_obj.three_d_collapsed
+
+    # Units of 'spin ticks', where one 'spin tick' equals 320 microseconds
+    spin_period_ns = spin_period.astype(np.float64) * 320 * 1e3  # Convert to ns
+    delta_times = (num_spins * spin_period_ns) / 2
```
greglucas: This confused me a bit at first, but I think makes sense to me now. Maybe put a comment in about something having to do with the data being collected over multiple spins, so we are finding the center of the collection period not the center of the spin period.

imap_processing/codice/utils.py
```diff
@@ -57,3 +96,203 @@ class CoDICECompression(IntEnum):
...
+    np.ndarray
+        Array of acquisition times per step of shape (num_esa_steps,).
+    """
+    # These tunable values are used to calculate acquisition time per step
+    tunable_values = low_stepping_tab["tunable_values"]
+
+    # pre-calculate values
+    sector_time = tunable_values["spin_time_ms"] / tunable_values["num_sectors_ms"]
+    sector_margin_ms = tunable_values["sector_margin_ms"]
+    dwell_faction = tunable_values["dwell_fraction_percentage"]
```
greglucas: typo
```suggestion
    dwell_fraction = tunable_values["dwell_fraction_percentage"]
```

imap_processing/codice/utils.py
```diff
@@ -57,3 +96,203 @@ class CoDICECompression(IntEnum):
...
+    min_hv_settle_ms = tunable_values["min_hv_settle_ms"]
+    max_hv_settle_ms = tunable_values["max_hv_settle_ms"]
+    num_steps_data = np.array(
+        low_stepping_tab["num_steps"].get("data"), dtype=np.float64
+    )
+    # Total non-acquisition time is in column (BD) of science LUT
+    total_non_acq_time = float(sector_time) * (100.0 - dwell_faction) / 100.0
```
greglucas: Can you rename this to "dwell_percentage" or fully spelled out "dwell_fraction_percentage". Otherwise I thought there was an error here because you are subtracting from 100 rather than 1.

imap_processing/codice/utils.py
```diff
@@ -57,3 +96,203 @@ class CoDICECompression(IntEnum):
...
+    # Calculate HV settle time per step not adjusted for Min/Max.
+    # It's in column (BF) of science LUT.
+    non_adjusted_hv_settle_per_step = (
+        total_non_acq_time - sector_margin_ms
+    ) / num_steps_data
+    hv_settle_per_step = np.minimum(
+        np.maximum(non_adjusted_hv_settle_per_step, min_hv_settle_ms), max_hv_settle_ms
+    )
+
+    # acquisition time per step in seconds
```
greglucas: Do you need to divide sector_margin_ms by 1000? Or update this comment to be in ms.

imap_processing/codice/codice_l1a_lo_species.py
tech3371: true. will change that!

imap_processing/codice/codice_l1a_lo_species.py
tech3371: Json is picky about key being int. I don't remember why. I can try again.

imap_processing/codice/codice_l1a_lo_species.py
tech3371: so create a function in here and call it?

imap_processing/codice/utils.py
tech3371: we need to unpack and keep data in the order of (spin_sector, inst_az). I checked with CoDICE team on this because I had same question yesterday.

imap_processing/codice/codice_l1a_lo_species.py
greglucas: No, it is already available, I think the comemnt I made should work.

imap_processing/codice/codice_l1a_lo_species.py
greglucas: I figured that might be the case. It depends on how it is saved in the json file itself then whether it is quoted or not I think. So you'd likely have to update the json you created I think.

imap_processing/codice/utils.py
greglucas: Maybe use those labels then? `unique_spin_sectors`, `unique_inst_azs` so it is clear what the rows/cols are referring to.

imap_processing/codice/utils.py
tech3371: I don't need to. Data comes in ms. That's why I added suffix `_ms` to indicate that. Hope that helps!

imap_processing/codice/codice_l1a_lo_species.py
tech3371: cool

imap_processing/codice/utils.py
greglucas: This comment needs to be updated then because it says "in seconds"

imap_processing/codice/utils.py
tech3371: sounds good!

imap_processing/codice/codice_l1a_lo_species.py
tech3371: I tried to convert to int but it doesn't like it.
```In JSON, keys must always be strings. This is part of the JSON specification (RFC 8259). Even if you write a key as a number (like 0), most JSON parsers will treat it as the string "0".```

Screenshot 2025-10-23 at 10 15 47 AM

imap_processing/codice/utils.py
tech3371: I see what you are saying! Let me check with Joey and Michael on this.

tech3371: I will address this in future PR after talking with them. Good catch!

imap_processing/codice/codice_l1a_lo_species.py
greglucas: OK, no problem then. I suppose if you want to use indices then this should be a list instead of a mappable type. This is fine to leave as is.

tech3371: I understand. I kept structure of that json in such away that it's human readable since it's done manually by us.

imap_processing/codice/utils.py
tech3371: Michael responded and I fixed comments and return value to be seconds

tech3371:
@greglucas I addressed all of the comments. Did I miss anything?

greglucas (approved):
Yep, looks good to me!

tech3371:
Thank you @greglucas for quick and helpful review! Appreciate it!

PR #2318, opened 2025-10-21
I-ALiRT - Updated frame of reference

laspsandoval:
# Change Summary

## Overview
I did a sanity check with APL for this pointing code since we are providing this information to international ground station partners. We had values that were different enough that I started to look into why it would be. When I changed the obsref to ITRF93 our values agreed much better:

Max azimuth difference compared w/ APL = 0.006 degrees
Max elevation difference compared w/ APL = 0.00009 degrees

Whereas before the elevation was off by 10ths of a degree and azimuth was off by 100ths of a degree. 

A ground station with a 7m dish needs the pointing to have an accuracy of < ~0.03 degrees to have strong signal strength. I think I now feel confident about these calculations, but want an opinion from @subagonsouth . Note that this will be sent to our ground station partner in Brazil after this review.

## Updated Files
- process_ephemeris.py
   - Changed to ITRF93

## Testing
[manaus_pointing_schedule_20251023.txt](https://github.com/user-attachments/files/23027455/manaus_pointing_schedule_20251023.txt)

### DIFF
--- a/imap_processing/ialirt/process_ephemeris.py
+++ b/imap_processing/ialirt/process_ephemeris.py
@@ -72,6 +72,7 @@ def calculate_azimuth_and_elevation(
     altitude: float,
     observation_time: float | np.ndarray,
     target: str = SpiceBody.IMAP.name,
+    obsref: str = "IAU_EARTH",
 ) -> tuple:
     """
     Calculate azimuth and elevation.
@@ -91,6 +92,9 @@ def calculate_azimuth_and_elevation(
         is to be computed. Expressed as ephemeris time, seconds past J2000 TDB.
     target : str (Optional)
         The target body. Default is "IMAP".
+    obsref : str (Optional)
+        Body-fixed, body-centered reference frame wrt
+        observer's center.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


tmplummer (approved):
Yes! Nice job tracking this down. I learned something here too. According to NAIF:

> The IAU_EARTH reference frame is implemented via approximate formulas
   provided by the IAU report [1] and is not suitable for high-accuracy
   work.

imap_processing/ialirt/process_ephemeris.py
```diff
@@ -72,6 +72,7 @@ def calculate_azimuth_and_elevation(
     altitude: float,
     observation_time: float | np.ndarray,
     target: str = SpiceBody.IMAP.name,
+    obsref: str = "IAU_EARTH",
```
tmplummer: Wait, wait. If you are going to add this optional input, I think you should have the default be "ITRF93" so that others don't make the mistake of using the low accuracy frame.
```suggestion
    obsref: str = "ITRF93",
```

laspsandoval: Right, the reason is that I also use this code to predict ground station coverage, which doesn’t require high accuracy. That coverage routine relies on the science frame kernel, so I would’ve had to update the science frame kernel to include ITRF93 unnecessarily. But I will make ITRF93 the default, as suggested.

PR #2323, opened 2025-10-21
MNT/ENH: Decommutate Ultra packets with derived values

### LINKED ISSUE #2076: Ultra L1A HK to EU unit
### Summary

Additional post-launch wrap up work
VAR_NOTES - Bits for bitwise flags need to be documented here.
l1a add engineering units for certain packets

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview

The Ultra packet definition has engineering unit conversions, so we can call the packet_file_to_datasets with `use_derived_value=True` to get the conversion. These are then l1b products to follow along with the mission convention. We handle that directly in the l1a processing by calling the routine again with a different flag. The metadata for l1a can be re-used, except that FILLVAL, MINVAL, and MAXVAL are all skipped for l1b because those values try to get converted to the datatype, but the datatype has changed with enumeration and conversions from uint8 to str or uint8 to float, etc. For ease of use, we just skip those metadata fields for now.

closes #2076

## Notes

The only update within the for-loop is skipping event datasets because those are handled separately in l1b and I didn't want to have any conflicts with those. Is this the only place where those would overlap, should I do something else to handle these events cases?

I'm not sure I love this approach here by adding another for loop over the potential datasets_by_apid cases, but it seemed like the easiest path forward without changing all tests to do this all the time. Let me know if you have other ideas or preferences and I'm happy to update this.

### DIFF
--- a/imap_processing/ultra/l1a/ultra_l1a.py
+++ b/imap_processing/ultra/l1a/ultra_l1a.py
@@ -43,7 +43,7 @@ logger = logging.getLogger(__name__)
 
 
 def ultra_l1a(  # noqa: PLR0912
-    packet_file: str, apid_input: int | None = None
+    packet_file: str, apid_input: int | None = None, create_derived_l1b: bool = False
 ) -> list[xr.Dataset]:
     """
     Will process ULTRA L0 data into L1A CDF files at output_filepath.
@@ -54,6 +54,8 @@ def ultra_l1a(  # noqa: PLR0912
         Path to the CCSDS data packet file.
     apid_input : Optional[int]
         Optional apid.
+    create_derived_l1b : bool
+        Whether to create the l1b datasets with derived values.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


imap_processing/ultra/l1a/ultra_l1a.py
```diff
@@ -109,77 +121,114 @@ def ultra_l1a(  # noqa: PLR0912
...
+                decom_ultra_dataset = process_ultra_energy_rates(datasets_by_apid[apid])
+                decom_ultra_dataset = decom_ultra_dataset.drop_vars("ratedata")
+                gattr_key = ULTRA_ENERGY_RATES.logical_source[
+                    ULTRA_ENERGY_RATES.apid.index(apid)
+                ]
+            elif apid in all_event_apids:
+                # We don't want to process the event l1b datasets since those l1b
+                # products need more information
+                if i == 1:
+                    continue
```
greglucas, line 149: This is the change within the for loop.

I did this because we are producing a `de` product already in l1b. Is this the right avoidance for that de product or do we need to do something else?
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/88979b9897a58dd97b460f92678686b06d237e8d/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv#L891

lacoak21: I think thats enough to skip it here. Good catch

lacoak21 (approved):

lacoak21:
Just tested it with this file : imap_ultra_l0_raw_20251015-repoint00018_v001.pkts
and the l1b data matches what George was expecting! Thanks Greg

PR #2324, opened 2025-10-22
ULTRA l1b filter out events that are during a repointing

lacoak21:
# Change Summary

## Overview
The first issue was the conversion of event times. Tenzin helped me me solve this 🎆 ! She had a similar issue with CODICE. The second issue: 
Event times are calculated using the universal spin table because they are not in the packets. This was causing a SPICE coverage issue if the spin start time was during a repointing. The SPICE conversion was using the pointing attitude kernel which has gaps during repointing. I talked to the ULTRA team and they said for now it would be ok to just filter events out that are during the repointing and then we can revisit this problem later. 

## Updated Files
   - Use the repoint kernel to find the start and stop times of the pointing and filter out events outside of that range. 
   - Fix time conversion

## Testing
Fix the time conversion

### DIFF
--- a/imap_processing/ultra/l1b/de.py
+++ b/imap_processing/ultra/l1b/de.py
@@ -9,6 +9,8 @@ from imap_processing.quality_flags import (
     ImapDEScatteringUltraFlags,
 )
 from imap_processing.spice.geometry import SpiceFrame
+from imap_processing.spice.repoint import get_repoint_data
+from imap_processing.spice.time import et_to_met
 from imap_processing.ultra.l1b.lookup_utils import get_geometric_factor
 from imap_processing.ultra.l1b.ultra_l1b_annotated import (
     get_annotated_particle_velocity,
@@ -74,6 +76,10 @@ def calculate_de(
     spin_number = get_spin_number(
         de_dataset["shcoarse"].values, de_dataset["spin"].values
     )
+    repoint_id = de_dataset.attrs.get("Repointing", None)
+    if repoint_id is not None:
+        repoint_id = int(repoint_id.replace("repoint", ""))
+
     de_dict["spin"] = spin_number
 
     # Add already populated fields.
@@ -150,14 +156,6 @@ def calculate_de(
         dtype=np.uint16,
     )
 
-    xf[valid_indices] = get_front_x_position(
-        de_dataset["start_type"].data[valid_indices],
-        de_dataset["start_pos_tdc"].data[valid_indices],
-        f"ultra{sensor}",
-        ancillary_files,
-    )
-    start_type[valid_indices] = de_dataset["start_type"].data[valid_indices]
-
     (
         event_times[valid_indices],
         spin_starts[valid_indices],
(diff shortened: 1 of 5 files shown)


imap_processing/ultra/l1b/de.py
```diff
@@ -312,6 +318,27 @@ def calculate_de(
...
     valid_events = event_times != FILLVAL_FLOAT32
+    # TODO - find a better solution than filtering out data from repointings?
+    if repoint_id is not None:
+        repoint_data = get_repoint_data()
+        # To find the pointing start and stop, get the end of the current repointing
+        # and the start of the next repointing
+        repoint_row = repoint_data[repoint_data["repoint_id"] == repoint_id]
+        next_repoint_row = repoint_data[repoint_data["repoint_id"] == repoint_id + 1]
+        pointing_start_met = repoint_row["repoint_end_met"].values[0]
+        pointing_end_met = next_repoint_row["repoint_start_met"].values[0]
```
greglucas: Is the reason you have to do this method because this `get_pointing_times()` function takes in an event MET rather than a pointing_id? I'm wondering if this is something we should add as a helper function over there, or if alternatively you can pass the first event time into that function here to take advantage of it? (I think the times in each data file should be (end maneuver -> end maneuver) so the portion that gets filtered is always at the end.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/4c4c1077330cffced22cdcdfddbb0589f243d290/imap_processing/spice/repoint.py#L200

imap_processing/ultra/l1b/de.py
```diff
@@ -312,6 +318,27 @@ def calculate_de(
...
+        # Create a boolean array for events within the pointing
+        in_pointing = np.zeros(len(event_times), dtype=bool)
+
+        # Check which events are within the pointing
+        in_pointing[valid_events] = (
+            et_to_met(event_times[valid_events]) >= pointing_start_met
+        ) & (et_to_met(event_times[valid_events]) <= pointing_end_met)
+
+        # Update valid_events to only include times within a pointing
+        valid_events = valid_events & in_pointing
```
greglucas:
```suggestion
        valid_events &= in_pointing
```

greglucas (approved):
Looks good. Just one minor arrangement question and where to put some of the new helpful utilities. We can do that rearrangement later too though and it shouldn't hold this up.

imap_processing/ultra/l1b/de.py
lacoak21: Yes exactly. I think adding a helper function is a good idea. Laura said that the science dataset epoch times are packet creation times so I wasn't sure about using the first one to get the pointing time. I thought the repoint id might be safer.

lacoak21: I will make a ticket to add a helper function later.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -996,17 +996,21 @@ def get_eventtimes(
...
         + spin_df["spin_start_subsec_sclk"].values[index] / 1e6
     )
 
     spin_period_sec = spin_df["spin_period_sec"].values[index]
-
     event_times = spin_starts + spin_period_sec * (phase_angle / 720)
 
-    return sct_to_et(event_times), sct_to_et(spin_starts), spin_period_sec
+    return (
+        ttj2000ns_to_et(met_to_ttj2000ns(event_times)),
```
tech3371, line 1010: I am wondering if you can use. Either way, it's both getting to ET which seems like what you wanted. bummer we don't have MET to ET directly.
```
    sclk_ticks = met_to_sclkticks(met)
    et = sct_to_et(sclk_ticks)
```

tech3371 (approved):
Looks good to me

PR #2326, opened 2025-10-22
Modify build_cdf_dataset to automatically drop variables that don't h…

### LINKED ISSUE #2325: ENA Maps - automatically drop variables in build_cdf_dataset
### Algorithm Description:

It seems like it would be useful to optionally remove variables from the dataset that don't have attributes defined. This makes it so that the upstream code doesn't need to clean up intermediates. Optionally, these variables can be included if the flag is set to allow them.

tmplummer:
…ave attributes defined.

# Change Summary

## Overview
Small change to the RectangularSkyMap.build_cdf_dataset function. I found it desirable to automatically remove variables that do not have CDF attributes defined. These are typically intermediate variables. Optionally, one can specify that they want to keep such variables. In that case, the variables are kept with no attributes assigned from the CDF Attribute Manager.

## Updated Files
   - Instead of raising a KeyError, either drop variables with no attributes, or keep them and log that they don't have attributes defined.
   - Make sure that variables with no attributes get removed when desired and left in when appropriate.

Closes: #2325

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -1265,12 +1265,13 @@ class RectangularSkyMap(AbstractSkyMap):
             coords={**self.non_spatial_coords, **self.spatial_coords},
         )
 
-    def build_cdf_dataset(
+    def build_cdf_dataset(  # noqa: PLR0912
         self,
         instrument: str,
         level: str,
         descriptor: str,
         sensor: str | None = None,
+        drop_vars_with_no_attributes: bool = True,
     ) -> xr.Dataset:
         """
         Format the data into a xarray.Dataset and add required CDF variables.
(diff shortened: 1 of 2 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -1388,13 +1395,18 @@ def build_cdf_dataset(
...
+                if drop_vars_with_no_attributes:
+                    logger.debug(
+                        f"Dropping variable '{name}' that has no attributes defined."
+                    )
+                    cdf_ds = cdf_ds.drop_vars(name)
+                else:
+                    logger.debug(
+                        f"Variable '{name}' has no attributes defined. It will "
+                        f"be included in the output dataset with no attributes."
+                    )
```
greglucas, line 1409: The issue I've had here is that sometimes it won't write the cdf output because of ISTP compliance, is that an issue here too?

greglucas (approved):

imap_processing/ena_maps/ena_maps.py
tmplummer: Yes, you are probably right that it will fail to write the file if cdflib fails to fix the ISTP compliance issue. I was thinking this feature would mostly be used for validation and debugging so maybe it is ok to need to do a manual `write_cdf(..., istp=False`?

PR #2329, opened 2025-10-22
I-ALiRT - add tcp modifications

laspsandoval:
# Change Summary

## Overview
We had made the assumption that the tcp connection would go down (and then up again) after a DSN contact, but it seems that Kiel's tcp connection is perpetually stable, which is great! The issue is that our code was created to search for those messages "tcp up/down" in the logs. So the plots shown on the imap website show that there is no tcp connection, when indeed it is actually stable:
https://imap-mission.com/ialirt/packets

This PR searches for another message and if the tcp up/down message is not in the logs, we populate the tcp connection so that it is from the start/end time of the logs' timestamps if the secondary message appears ("Opened raw record file for {partner}")

I also added the coordinates for Korea.

## Updated Files
- calculate_ingest.py
   - Add alternative message in case "tcp up/down" is not present in the logs.

## Testing
- test_calculate_ingest.py

### DIFF
--- a/imap_processing/ialirt/calculate_ingest.py
+++ b/imap_processing/ialirt/calculate_ingest.py
@@ -9,7 +9,7 @@ from imap_processing.ialirt.constants import STATIONS
 logger = logging.getLogger(__name__)
 
 
-def find_tcp_connections(
+def find_tcp_connections(  # noqa: PLR0912
     start_file_creation: datetime,
     end_file_creation: datetime,
     lines: list,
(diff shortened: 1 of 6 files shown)


greglucas (approved):
I'll just note that this all seems quite complicated to me with many different potential paths you need to keep track of. I kind of wonder if there is some other way of logging what is going on here rather than you having to parse these logs for various conditions.

laspsandoval:
> I'll just note that this all seems quite complicated to me with many different potential paths you need to keep track of. I kind of wonder if there is some other way of logging what is going on here rather than you having to parse these logs for various conditions.

Noted Greg. I'll merge for now, though.

PR #2330, opened 2025-10-22
ULTRA l1b flag dont filter events in a pointing 

lacoak21:
# Change Summary

## Overview
Instead of filtering the entire dataset for events within the pointing I should instead, use the fill value and flag those events. 

## Updated Files
- imap_processing/quality_flags.py
   - Add new quality flag
   - Do not filter the entire dataset!! Just flag the repointing events. 

## Testing
Fix test.

### DIFF
--- a/imap_processing/tests/ultra/unit/test_de.py
+++ b/imap_processing/tests/ultra/unit/test_de.py
@@ -139,6 +139,7 @@ def test_calculate_events_in_pointing(use_fake_repoint_data_for_time):
     in_pointing = calculate_events_in_pointing(
         repoint_id,
         ttj2000ns_to_et(met_to_ttj2000ns(event_times)),
+        valid_events=np.ones(len(event_times), dtype=bool),
     )
     # The first event should be False (not during a pointing), and the rest True.
     assert np.all(not in_pointing[0])
--- a/imap_processing/tests/ultra/unit/test_de.py
+++ b/imap_processing/tests/ultra/unit/test_de.py
@@ -136,14 +136,10 @@ def test_calculate_events_in_pointing(use_fake_repoint_data_for_time):
     event_times = np.arange(pointing_start, pointing_start + 10)
     # Edit the first event time to be during a repoint.
     event_times[0] = pointing_start - 1
-    valid_events = np.ones_like(event_times, dtype=bool)
-    quality_flags = np.zeros_like(event_times, dtype=np.uint32)
-    in_poiting = calculate_events_in_pointing(
+    in_pointing = calculate_events_in_pointing(
         repoint_id,
         ttj2000ns_to_et(met_to_ttj2000ns(event_times)),
-        valid_events,
-        quality_flags,
     )
     # The first event should be False (not during a pointing), and the rest True.
-    assert np.all(not in_poiting[0])
-    assert np.all(in_poiting[1:])
+    assert np.all(not in_pointing[0])
+    assert np.all(in_pointing[1:])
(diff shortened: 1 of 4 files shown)


imap_processing/ultra/l1b/de.py
```diff
@@ -428,4 +431,5 @@ def calculate_events_in_pointing(
         et_to_met(event_times[valid_events]) >= pointing_start_met
     ) & (et_to_met(event_times[valid_events]) <= pointing_end_met)
 
+    quality_flags[~in_pointing] |= ImapAttitudeUltraFlags.DURINGREPOINT.value
```
greglucas: These are only updated for valid_events. Can you do this for all times instead above? i.e. remove the valid_events from the in_pointing checks above. That way we can know if it is both an invalid event and during a repointing.

lacoak21: Great point 👍

greglucas: Now you don't use the valid_events and the docstring needs to be updated I think. :+1: on the change though!

imap_processing/tests/ultra/unit/test_de.py
```diff
@@ -137,8 +137,12 @@ def test_calculate_events_in_pointing(use_fake_repoint_data_for_time):
     # Edit the first event time to be during a repoint.
     event_times[0] = pointing_start - 1
     valid_events = np.ones_like(event_times, dtype=bool)
+    quality_flags = np.zeros_like(event_times, dtype=np.uint32)
     in_poiting = calculate_events_in_pointing(
```
greglucas: typo: "pointing"

This needs to drop the valid_events.

imap_processing/ultra/l1b/de.py
```diff
@@ -321,7 +322,7 @@ def calculate_de(
     # TODO - find a better solution than filtering out data from repointings?
     if repoint_id is not None:
         in_pointing = calculate_events_in_pointing(
-            repoint_id, event_times, valid_events
+            repoint_id, event_times, quality_flags
```
greglucas: Rather than calculate_events_in_pointing doing the quality_flags update internally, do you want to handle that outside here. I think then the function would be more self-contained and potentially usable in other instruments (future) and mean you don't need to add a zeros pass-through in the test.

Then below this, why do you have a "valid_events" array, isn't that the same as what quality_flags is meant to indicate? i.e. I think we should be updating the quality flags throughout all these routines and then at the end use that to figure out what events are valid or not based upon the quality flags.

imap_processing/tests/ultra/unit/test_de.py
lacoak21: Oops - i forgot about unit test. This is a good reminder to be more thorough

imap_processing/ultra/l1b/de.py
lacoak21: Sure I can make that update outside. Also looking back at the upstream code it seems that valid_events and quality flags are not the same and I am not sure why. For example the valid_events is derived from the valid event_times but the quality flags are NOT updated based on invalid event_times. Im wondering if it set up this way so that we can still calculate certain values but the quality flag lets the user filter out those values based on certain flags. E.g. using only the valid_events (valid times) we can still get the particle velocities and the user can decide if they want to use them if they are flagged.

greglucas: Yeah, I think I agree with that. But, I think my point being that don't we want to calculate these quantities all the time and then use the flags to identify valid events only at the end. i.e. not only calculate quantities at a subset of "valid" times that is changing as we go.

lacoak21: Oh I see. We have to mask out the event times that are fillvals because it will throw a SPICE error in the function `get_annotated_particle_velocity` below. I can handle that inside the function if that is better?

greglucas: Ahh, OK. This all looks good to me now! Looks cleaner and simpler to me at least.

lacoak21: OK darn. I think I have to mask out event_times with valid_events again inside calculate_events_in_pointing because of a spice error as well.

lacoak21: Because I use et_to_met()

greglucas (approved):

PR #2333, opened 2025-10-23
CoDICE: L1B species refactoring after L1A

tech3371:
# Change Summary

## Overview

First refactor of L1B. Validation not included in this since L1A still needs fix for epoch and acquisition time which is important to validation. But this setups initial refactoring work of L1B.

## Updated Files

- imap_processing/codice/codice_l1b.py
   - L1B refactoring

### DIFF
--- a/imap_processing/codice/codice_l1b.py
+++ b/imap_processing/codice/codice_l1b.py
@@ -56,17 +56,9 @@ def convert_to_rates(dataset: xr.Dataset, descriptor: str) -> np.ndarray:
         "lo-sw-priority",
         "lo-ialirt",
     ]:
-        # Solve denominator, Acquisition times, of above products. Acquisition times
-        # must reshape to match the data variable shape, (epoch, esa_step, sector)
-        acq_times = dataset.acquisition_time_per_step.expand_dims(
-            {
-                "epoch": dataset.sizes["epoch"],
-                "spin_sector": dataset.sizes["spin_sector"],
-            }
-        ).transpose("epoch", "esa_step", "spin_sector")
         # Denominator to convert counts to rates
         denominator = (
-            acq_times
+            dataset.acquisition_time_per_step
             * constants.L1B_DATA_PRODUCT_CONFIGURATIONS[descriptor]["num_spin_sectors"]
         )
 
@@ -83,31 +75,17 @@ def convert_to_rates(dataset: xr.Dataset, descriptor: str) -> np.ndarray:
         "lo-nsw-species",
         "lo-sw-species",
     ]:
-        # Solve denominator, Acquisition times, of above products. Acquisition times
-        # must reshape to match the data variable shape, (epoch, esa_step, sector)
-        acq_times = dataset.acquisition_time_per_step.expand_dims(
-            {
-                "epoch": dataset.sizes["epoch"],
-                "spin_sector": dataset.sizes["spin_sector"],
-            }
-        ).transpose("epoch", "esa_step", "spin_sector")
-        # Create n_sector with same shape as data.
+        # Create n_sector with 'esa_step' dimension. This is done by xr.full_like
+        # with input dataset.acquisition_time_per_step. This ensures that the resulting
+        # n_sector has the same dimensions as acquisition_time_per_step.
         # Per CoDICE, fill first 127 with default value of 12. Then fill last with 11.
         n_sector = xr.full_like(
             dataset.acquisition_time_per_step, 12.0, dtype=np.float64
         )
         n_sector[-1] = 11.0
-        n_sector = n_sector.expand_dims(
-            {
-                "epoch": dataset.sizes["epoch"],
-                "spin_sector": dataset.sizes["spin_sector"],
-            }
-        ).transpose("epoch", "esa_step", "spin_sector")
 
         # Denominator to convert counts to rates
-        denominator = (
-            acq_times * n_sector  # Spin sectors
-        )
+        denominator = dataset.acquisition_time_per_step * n_sector
 
         # Do not carry these variable attributes from L1a to L1b for above products
         drop_variables = [
@@ -147,7 +125,7 @@ def convert_to_rates(dataset: xr.Dataset, descriptor: str) -> np.ndarray:
         dataset[unc_variable].data = (
             dataset[unc_variable].astype(np.float64) / denominator
         )
-        dataset[unc_variable].attrs["UNITS"] = "sqrt(counts)/s"
+        dataset[unc_variable].attrs["UNITS"] = "1/s"
 
     return dataset
 
--- a/imap_processing/codice/codice_l1b.py
+++ b/imap_processing/codice/codice_l1b.py
@@ -15,20 +15,15 @@ from pathlib import Path
 import numpy as np
 import xarray as xr
 
-from imap_processing import imap_module_directory
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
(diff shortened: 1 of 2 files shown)


imap_processing/codice/codice_l1b.py
```diff
@@ -58,41 +56,42 @@ def convert_to_rates(
...
-        # In order to divide by acquisition times, we must reshape the acq
-        # time data array to match the data variable shape
-        dims = [1] * dataset[variable_name].data.ndim
-        dims[1] = 128
-        acq_times = dataset.acquisition_time_per_step.data.reshape(dims)  # (128)
-        # Now perform the calculation
-        rates_data = dataset[variable_name].data / (
+        # Solve denominator, Acquisition times, of above products. Acquisition times
+        # must reshape to match the data variable shape
+        acq_times = dataset.acquisition_time_per_step.data[np.newaxis, :, np.newaxis]
```
greglucas: Can you stay in the proper xarray dimensions here and let the xarray broadcasting work itself out? If your acquisition_time_per_step has dimension `esa_step`, then it should broadcast out across the other dimensions automatically when doing the later multiplications. This has simplified a lot of the work on other instruments to try and stay in xarray-land rather than switch over to numpy axes.

imap_processing/codice/codice_l1b.py
```diff
@@ -58,41 +56,42 @@ def convert_to_rates(
...
-        dims = [1] * dataset[variable_name].data.ndim
-        dims[1] = 128
-        acq_times = dataset.acquisition_time_per_step.data.reshape(dims)  # (128)
+        # Solve denominator, Acquisition times, of above products. Acquisition times
+        # must reshape to match the data variable shape, (epoch, esa_step, sector)
+        acq_times = dataset.acquisition_time_per_step.data[np.newaxis, :, np.newaxis]
         # acquisition time have an array of shape (128,). We match n_sector to that.
         # Per CoDICE, fill first 127 with default value of 12. Then fill last with 11.
-        n_sector = np.full(128, 12, dtype=int)
+        n_sector = np.full(constants.NUM_ESA_STEPS, 12, dtype=int)
```
greglucas: Something like this to make in an xarray version that can be broadccast out for you: `xr.full(..., dims=("esa_step"))`

imap_processing/codice/codice_l1b.py
```diff
@@ -101,15 +100,20 @@ def convert_to_rates(
...
             * constants.HI_ACQUISITION_TIME
         )
 
-    return rates_data
+    # For each variable, convert counts and uncertainty to rates
+    for variable in variables_to_convert:
+        dataset[variable].data = dataset[variable].data / denominator
+        # Uncertainty calculation
+        unc_variable = f"unc_{variable}"
+        dataset[unc_variable].data = dataset[unc_variable].data / denominator
```
greglucas: If you take the above this can be written as `dataset[variable] /= denominator` I think.

greglucas (commented):
One over-arching suggestion to try and do the xarray version in the refactor now if you can.

tech3371:
> One over-arching suggestion to try and do the xarray version in the refactor now if you can.

Try/Except in convert_to_rates instead of if/else statement?

greglucas:
>> One over-arching suggestion to try and do the xarray version in the refactor now if you can.

> Try/Except in convert_to_rates instead of if/else statement?

No, sorry, I meant to switch to using xarray data-arrays for everything so that you don't use any `np.newaxis`. I find the newaxis things to be the confusing portions here, so if you can create a denominator with the proper dimension it should be applied over, xarray will take care of the newaxis calls for you.

tech3371:
> > > One over-arching suggestion to try and do the xarray version in the refactor now if you can.
> 
> > Try/Except in convert_to_rates instead of if/else statement?
> 
> No, sorry, I meant to switch to using xarray data-arrays for everything so that you don't use any `np.newaxis`. I find the newaxis things to be the confusing portions here, so if you can create a denominator with the proper dimension it should be applied over, xarray will take care of the newaxis calls for you.

Sounds good. Good idea. I will revisit this PR after times are fixed and make changes requested and validate them. It's good to know feedback in advance to incorporate later. Thank you!

tech3371:
@greglucas I refactor to use data array. It's much more human readable.

imap_processing/codice/codice_l1b.py
```diff
@@ -58,41 +56,71 @@ def convert_to_rates(
...
+        acq_times = dataset.acquisition_time_per_step.expand_dims(
+            {
+                "epoch": dataset.sizes["epoch"],
+                "spin_sector": dataset.sizes["spin_sector"],
+            }
+        ).transpose("epoch", "esa_step", "spin_sector")
+        # Denominator to convert counts to rates
+        denominator = (
             acq_times
-            * 1e-3  # Converting from milliseconds to seconds
```
greglucas, line 71: Did you lose these multiplications now, or where do they come in?

imap_processing/codice/codice_l1b.py
```diff
@@ -101,15 +129,27 @@ def convert_to_rates(
...
+        dataset[variable].data = dataset[variable].astype(np.float64) / denominator
+        # Carry over attrs and update as needed
+        dataset[variable].attrs["UNITS"] = "counts/s"
+
+        # Uncertainty calculation
+        unc_variable = f"unc_{variable}"
+        dataset[unc_variable].data = (
+            dataset[unc_variable].astype(np.float64) / denominator
+        )
+        dataset[unc_variable].attrs["UNITS"] = "sqrt(counts)/s"
```
greglucas:
```suggestion
        dataset[unc_variable].attrs["UNITS"] = "1/s"
```

imap_processing/codice/codice_l1b.py
```diff
@@ -58,41 +56,71 @@ def convert_to_rates(
...
-        # document
-        # In order to divide by acquisition times, we must reshape the acq
-        # time data array to match the data variable shape
-        dims = [1] * dataset[variable_name].data.ndim
-        dims[1] = 128
-        acq_times = dataset.acquisition_time_per_step.data.reshape(dims)  # (128)
-        # Now perform the calculation
-        rates_data = dataset[variable_name].data / (
+        # Solve denominator, Acquisition times, of above products. Acquisition times
+        # must reshape to match the data variable shape, (epoch, esa_step, sector)
```
greglucas: I thought that if you have acquisition times as an xarray dataarray with dimension "esa_step" and multiply it with a dataarray that has 2D ("epoch", "esa_step"), the broadcasting would work out for you, so you don't have to do any of the expanding/transposing below. It knows to work over the esa_step coordinate.

greglucas (approved):
Looks good. One minor question/comment on the xarray broadcasting, but I think this will set it up for the future 👍

imap_processing/codice/codice_l1b.py
tech3371: I will try that.

imap_processing/codice/codice_l1b.py
tech3371: In L1A, I convert acquisition time per step to seconds now. So we don't need to convert anymore.

imap_processing/codice/codice_l1b.py
tech3371: Yes. That is much better. I am learning and loving xr more each day :)

greglucas: Very nice, looks really clean now!

PR #2334, opened 2025-10-23
Lo L1B DE - Binning fixes

sdhoyt:
# Change Summary

## Overview
This PR makes some fixes to the L1B DE binning:

- Convert to DPS frame for binning
- Fix goodtimes/badtimes bin window checks
- Update naming in pset code to align with updated DE code

These fixes are unrelated to the PSET validation currently happening with Lo. The DE CDF being used for that validation is prior to these bugs being introduced. I re-ran my checks of DE counts vs PSET counts and, like the previous DE CDF that Lo is working with, the DE/PSET comparisons are matching with the new set of DE/PSET with this update.

Thank you @subagonsouth for your help with the transform!

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -19,6 +19,7 @@ from imap_processing.lo.l1b.tof_conversions import (
 from imap_processing.spice.geometry import (
     SpiceFrame,
     cartesian_to_latitudinal,
+    frame_transform,
     instrument_pointing,
 )
 from imap_processing.spice.repoint import get_pointing_times
@@ -761,7 +762,7 @@ def set_bad_or_goodtimes(
     # 0.1 degree bins to align with the spin_bins, so multiply by 60
     time_mask = (epochs[:, None] >= times_start) & (epochs[:, None] <= times_end)
     bin_mask = (spin_bins[:, None] >= times_df["bin_start"].values * 60) & (
-        spin_bins[:, None] <= times_df["bin_end"].values * 60
+        spin_bins[:, None] < (times_df["bin_end"].values + 1) * 60
     )
 
     # Combined mask for epochs that fall within the time and bin ranges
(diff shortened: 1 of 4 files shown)


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -761,7 +762,7 @@ def set_bad_or_goodtimes(
     # 0.1 degree bins to align with the spin_bins, so multiply by 60
     time_mask = (epochs[:, None] >= times_start) & (epochs[:, None] <= times_end)
     bin_mask = (spin_bins[:, None] >= times_df["bin_start"].values * 60) & (
-        spin_bins[:, None] <= times_df["bin_end"].values * 60
+        spin_bins[:, None] < (times_df["bin_end"].values + 1) * 60
```
tmplummer, line 767: Can you explain why the `+ 1` is needed in a comment?

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -77,6 +77,10 @@ def attr_mgr_l1a():
     return attr_mgr
 
 
+@patch(
+    "imap_processing.lo.l1b.lo_l1b.frame_transform",
+    return_value=np.array([[0, 0, 0], [0, 0, 0], [0, 0, 0], [0, 0, 0]]),
```
tmplummer, line 82: Just a comment that these returned vectors are non physical values. I guess that doesn't matter for your test. I typically do something like the following when mocking `frame_transform`. This will just return same vectors that are passed in.
```
mock_frame_transform.side_effect = lambda a, b, c, d: b
```

tmplummer (approved):
This all looks good to me. Only change to request is adding a comment on the bin_mask calculation.

imap_processing/lo/l1b/lo_l1b.py
sdhoyt: Yeah, I'll add a comment. The ancillary file typically has the bin start at 0 and bin end at 59, so I need to add 1 here so the bin end is always at the top edge of the bin and I don't miss any data when the bin is > 3540 and < 3600.

Issue #2319, opened 2025-10-21
CoDICE: L1A Lo angular refactor


tech3371:
- [ ] refactor l1a angular products

PR #2336, opened 2025-10-24
CoDICE: New l1a angular refactor

tech3371:
# Change Summary
related to #2319
Not closing mentioned ticket until time mismatch is fixed.

## Overview

Adds Lo angular processing. Simplify despinning logics into one for-loop. This includes CDF attrs work as well.

## Updated Files

- imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
- imap_processing/codice/codice_l1a_lo_species.py
- imap_processing/codice/codice_new_l1a.py
   - In this files, added support for angular CDF attrs and processing code
- imap_processing/codice/codice_l1a_lo_angular.py
- imap_processing/codice/utils.py
   - In these files, main angular processing code and utils function for that.
   - Depinning is added as well.

## Testing


imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
```diff
@@ -298,7 +310,7 @@ voltage_table:
 
 k_factor:
     CATDESC: K Factor constant that is used to convert voltages to energies
-    DEPEND_1: spin_sector
+    DEPEND_1: k_factor
```
greglucas, line 313: I don't think this is right, you can't depend on yourself can you? Should this be `k_factor_labels` or something instead.

imap_processing/codice/codice_l1a_lo_angular.py
```diff
@@ -0,0 +1,359 @@
...
+        (num_packets, num_species, esa_steps, 24, inst_az).
+    """
+    # species_data shape: (num_packets, num_species, esa_steps, *collapsed_dims)
+    num_packets, num_species, esa_steps = species_data.shape[:3]
+    collapsed_dims = species_data.shape[3:]
+    inst_az_dim = collapsed_dims[-1]
+
+    # Prepare despinning output: (num_packets, num_species, esa_steps, 24, inst_az_dim)
+    # 24 is derived by multiplying spin sector dim from collapse table by 2
+    spin_sector_len = collapsed_dims[-2] * 2
```
greglucas: Why not set this to 24 directly?

I'm a bit confused about the "collapsed_dims" and why you don't just extract the spin_sector_len and inst_az_dim up above. Are you guaranteed to have a "-2" index available?

imap_processing/codice/codice_l1a_lo_angular.py
```diff
@@ -0,0 +1,359 @@
...
+    despun_shape = (num_packets, num_species, esa_steps, spin_sector_len, inst_az_dim)
+    despun_data = np.full(despun_shape, 0)
+
+    # Pixel orientation array and mapping positions
+    pixel_orientation = np.array(
+        sci_lut_data["lo_stepping_tab"].get("pixel_orientation").get("data")
```
greglucas: It might be better to avoid the get() call because then you'd get a nicer explicit KeyError if it doesn't exist.
```suggestion
        sci_lut_data["lo_stepping_tab"]["pixel_orientation"]["data"]
```

imap_processing/codice/codice_l1a_lo_angular.py
```diff
@@ -0,0 +1,359 @@
...
+                :, :, orientation_b, :, pos_idx
+            ]
+        else:
+            # A → last 12, B → first 12
+            despun_data[:, :, orientation_a, 12:, pos_idx] = species_data[
+                :, :, orientation_a, :, pos_idx
+            ]
+            despun_data[:, :, orientation_b, :12, pos_idx] = species_data[
+                :, :, orientation_b, :, pos_idx
+            ]
```
greglucas, line 91: This looks good, but one thing that I'm somewhat curious about and not sure if it makes it any more clear or not, is whether you could make the orientations first and then do the lookup.

```python
orientation_first_12 = np.concat([orientation_a, orientation_b])
orientation_last_12 = np.concat([orientation_b, orientation_a])

despun_data[:, :, orientation, :, pos_idx] = species_data[:, :, orientation, : pos_idx]
```
so that you've collapsed the two `12:` and `:12` pieces into a single lookup array.

No need to do this, just something that came to mind at a quick glance.

imap_processing/codice/codice_l1a_lo_angular.py
```diff
@@ -0,0 +1,359 @@
...
+
+    # Despinning
+    # ----------------
+    species_data = _despin_species_data(species_data, sci_lut_data, view_tab_obj)
+
+    # ========== Get Voltage Data from LUT ===========
+    # Use plan id and plan step to get voltage data's table_number in ESA sweep table.
+    # Voltage data is (128,)
+    esa_table_number = (
+        sci_lut_data["plan_tab"].get(f"({plan_id}, {plan_step})").get("lo_stepping")
```
greglucas: Rather than get() throughout could you do the Key access way? I think it will be easier to read KeyErrors than None type does not have attribute .get() and you don't know where it is failing.

imap_processing/codice/utils.py
```diff
@@ -197,6 +197,40 @@ def get_collapse_pattern_shape(
...
+
+    Returns
+    -------
+    np.ndarray
+        Array of indices corresponding to non-zero unique rows.
+    """
+    if sensor_id == 0:
+        collapse_tab = json_data.get("collapse_lo").get(f"{collapse_table_id}")
+    else:
+        collapse_tab = json_data.get("collapse_hi").get(f"{collapse_table_id}")
```
greglucas:
```suggestion
    sensor = "lo" if sensor_id == 0 else "hi"
    collapse_matrix = np.array(json_data[f"collapse_{sensor}"][f"{collapse_table_id}"]["matrix"])
```

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
tech3371: This is tricky because of CDF attrs restriction. k_factor should be its own thing. It's nor dimension or epoch dependent data.

imap_processing/codice/codice_l1a_lo_angular.py
tech3371: I can hard code it. I may have overdone it. I think think spin sector size will ever change. If it does, then it will a major thing before it get to us.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
tech3371: Before, it made sense to make depend 1 to be spin sector. But with angular work, spin sector is not (1) anymore. It's (24) after despinning. I believe hi spin sector is (24,) too.

imap_processing/codice/codice_l1a_lo_angular.py
tech3371: Yeah, we might loose readability. But I added comments to be little bit more clear and maps to discussion with CoDICE team.

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
greglucas: I still don't follow though. Does the "epoch" variable DEPEND_0 on "epoch" (itself)?

greglucas (approved):

imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
tech3371: zoom?

greglucas: We are good. I understand now that this is a CDF metadata thing to get the load/reload working right!

Issue #2332, opened 2025-10-23
SWAPI: convert sci_start_time to UTC


tech3371:
sci_start_time is uncorrect time. epoch is center time. SWAPI requested sci_start_time to be UTC format.

tech3371:
Addressed in linked PR

PR #2338, opened 2025-10-24
SWAPI: science time in UTC and additional checks on L2 energy lookup

### LINKED ISSUE #2256: BUG - SWAPI L2 Outputs large negative uncertainty values
### Description of the issue

When running the SWAPI L2 processor against L1 input that has fill value for the coincident counts and corresponding uncertainties, L2 outputs large negative uncertainties (i.e. 'swp_coin_rate_stat_uncert_minus' values of -6.896552e+31). We hypothesize this is because the L2 code is performing operations on fill values, -1e+31, from L1.

This causes L3 products to output fill, or not process altogether for the L3b product.

Run the L2 processor against this L1 input: [imap_swapi_l1_sci_20260923_v001_cdf_not_text.txt](https://github.com/user-attachments/files/22603946/imap_swapi_l1_sci_20260923_v001_cdf_not_text.txt)

The L2 Processor should propagate fill value from L1.

_No response_

### Code Snippet:

```python
Code
```

_No response_

tech3371:
# Change Summary
closes #2256 #2332 

## Overview

Fixing time format and energy checks per SWAPI requests.

## Updated Files

   - Time reformat per SWAPI requests
   - Add additional checks

## Testing

- imap_processing/tests/swapi/test_swapi_l2.py

### DIFF
--- a/imap_processing/swapi/l1/swapi_l1.py
+++ b/imap_processing/swapi/l1/swapi_l1.py
@@ -13,6 +13,7 @@ from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.cdf.utils import load_cdf
 from imap_processing.quality_flags import SWAPIFlags
 from imap_processing.spice.time import met_to_utc, ttj2000ns_to_met
+from imap_processing.swapi.constants import NUM_ENERGY_STEPS, NUM_PACKETS_PER_SWEEP
 from imap_processing.swapi.swapi_utils import SWAPIAPID, SWAPIMODE
 from imap_processing.utils import packet_file_to_datasets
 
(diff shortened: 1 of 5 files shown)


imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -638,8 +639,14 @@ def process_swapi_science(
         attrs=cdf_manager.get_variable_attributes("plan_id"),
     )
     # Store start time for L3 purposes per SWAPI requests
+    # Per SWAPI request, convert start time of sweep to UTC time.
+    sci_start_time = met_to_utc(
+        ttj2000ns_to_met(
+            good_sweep_sci["epoch"].data.reshape(total_full_sweeps, 12)[:, 0]
```
greglucas: Do you have a constant.NSWEEPS somewhere for the 12 to use rather than the number itself?

imap_processing/swapi/l1/swapi_l1.py
```diff
@@ -638,8 +639,14 @@ def process_swapi_science(
         attrs=cdf_manager.get_variable_attributes("plan_id"),
     )
     # Store start time for L3 purposes per SWAPI requests
+    # Per SWAPI request, convert start time of sweep to UTC time.
+    sci_start_time = met_to_utc(
```
greglucas, line 652: Do you know what precision they want? Currently it looks like we default to 9 digits! I don't think we need to be that precise and maybe it is better to just stick with 0 and land on seconds.

greglucas (approved):
A few comments that don't impact the overall structure, but perhaps something to ask SWAPI about.

imap_processing/swapi/l1/swapi_l1.py
tech3371: @bishwassth can you check with your team which precision they want? Here is two options that Greg is talking about:
1. '2024-09-24T13:21:45'
2. 2024-09-24T13:21:45.000000000'

Can you let me know? I can update in future PR based on which one you guys want.

tech3371: Three or four decimal might be good for SWAPI team - from our discussion

greglucas: It should be 3, 6, or 9 (ms, us, ns). I wouldn't do anything else as it gets confusing then.

tech3371: Good to know. Then I will do ms.

PR #2339, opened 2025-10-25
FIX: Correct Lo bootstrap calculations with energy dimension

greglucas:
# Change Summary

## Overview
The Lo bootstrap calculation was using the correct broadcasting, but the coordinates were not aligned between the bootstrap factors and the intensity variables. This adds the actual energy values to the bootstrap factors we used to align them and do the proper summation over energies.

The big changes are: use isel for the index selection when necessary, but add the energy "values" to the energy_k coordinates (which gets renamed to energy for alignment) so that the future summation over energy in xarray-space gets applied correctly.


tmplummer (approved):
LGTM

PR #2343, opened 2025-10-28
Lo flux correction

### LINKED ISSUE #953: Corrected ENA Fluxes in Sun rest frame

greglucas:
# Change Summary

## Overview

This adds the common flux correction scheme to the Lo L2 pipeline. I applied it to both oxygen and hydrogen datasets, but I wasn't clear whether this needs to be handled differently between them or not. @frahmanif do we need a different set of flux correction factors (different energy steps?) for the Oxygen corrections, or do you not want this applied to the Oxygen maps?

closes #953

### DIFF
--- a/imap_processing/lo/l2/lo_l2.py
+++ b/imap_processing/lo/l2/lo_l2.py
@@ -10,6 +10,7 @@ import xarray as xr
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.ena_maps import ena_maps
 from imap_processing.ena_maps.ena_maps import AbstractSkyMap, RectangularSkyMap
+from imap_processing.ena_maps.utils.corrections import PowerLawFluxCorrector
 from imap_processing.ena_maps.utils.naming import MapDescriptor
 from imap_processing.lo import lo_ancillary
 from imap_processing.spice.time import et_to_datetime64, ttj2000ns_to_et
@@ -77,7 +78,13 @@ def lo_l2(
     logger.info("Step 4: Calculating rates and intensities")
 
     # Determine if corrections are needed and prepare oxygen data if required
-    sputtering_correction, bootstrap_correction, o_map_dataset = _prepare_corrections(
+    (
+        sputtering_correction,
+        bootstrap_correction,
+        flux_correction,
+        o_map_dataset,
+        flux_factors,
+    ) = _prepare_corrections(
         map_descriptor, descriptor, sci_dependencies, anc_dependencies
     )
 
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l2/lo_l2.py
```diff
@@ -1082,6 +1121,58 @@ def calculate_bootstrap_corrections(dataset: xr.Dataset) -> xr.Dataset:
...
+        Dataset with calculated flux-corrected intensities and their
+        uncertainties for the specified species.
+    """
+    logger.info("Applying flux corrections")
+
+    # Flux correction
+    corrector = PowerLawFluxCorrector(flux_factors)
+    # FluxCorrector works on (energy, :) arrays, so we need to flatten the map
+    # spatial dimensions for the correction and then reshape back after.
+    input_shape = dataset["ena_intensity"].shape[1:]  # Exclude epoch dimension
```
greglucas, line 1153: @subagonsouth, I've worked these input datasets to be in "map" space rather than 1d-map space because of the issues with the Oxygen dataset being needed as well. This meant that this needed to be reshaped back and forth. I'm not sure if you have better ideas for how to handle this here or not...

imap_processing/lo/l2/lo_l2.py
```diff
@@ -145,7 +156,24 @@ def _prepare_corrections(
...
+                "No flux correction factor file found in ancillary dependencies"
+            ) from None
+
+    return (
+        sputtering_correction,
+        bootstrap_correction,
+        flux_correction,
+        o_map_dataset,
+        flux_factors,
+    )
```
greglucas, line 176: I don't love how this correction setup is growing here... I may refactor this in the future, but figure this is easiest for now.

imap_processing/lo/l2/lo_l2.py
tmplummer: Oh, bummer. In reality, the PowerLawFluxCorrector should be modified to allow multiple spatial dimensions. I can write up a ticket for that... I also have a ticket for vectorizing the correction. Both changes might go well together.

In the meantime, to move forward with the correction as is, I discovered the xarray.stack and xarray.unstack functions recently. I'm not sure what your spatial dimension names are but something like:
```
dataset = dataset.stack({CoordNames.GENERIC_PIXEL.value: ("longitude", "latitude")})

# Do flux correction

dataset = dataset.unstack(CoordNames.GENERIC_PIXEL.value)
```

I suppose you could also stack the epoch dimension and the energy dimension together.

tmplummer (approved):
Looks good to me. Sorry that the flux corrector isn't flexible enough to allow for the singular epoch dimension or multiple spatial dimensions.

imap_processing/lo/l2/lo_l2.py
greglucas: OK, I opened up #2345 for this work. No rush on that ticket, since it does work like this and I don't think we have any more implementations requiring it right now.

(The vectorization would be nice though because this did add quite a bit of time locally to the processing for me (a few seconds so I can't really complain))

greglucas:
> Looks good to me. Sorry that the flux corrector isn't flexible enough to allow for the singular epoch dimension or multiple spatial dimensions.

No worries @subagonsouth! Thank you for making that class, it was super easy to use and integrate in here.

PR #2346, opened 2025-10-28
1641 hi pset   epoch should be pointing start time

### LINKED ISSUE #1641: Hi PSET - epoch should be pointing start time

tmplummer:
# Change Summary

## Overview
- Use L1B met time to query repoint table data for repoint start/end time.
- Set PSET epoch value to repoint start time
- Store PSET duration in epoch_delta variable

Closes: #1641

### DIFF
--- a/imap_processing/hi/hi_l1c.py
+++ b/imap_processing/hi/hi_l1c.py
@@ -29,11 +29,12 @@ from imap_processing.spice.geometry import (
     frame_transform,
     frame_transform_az_el,
 )
+from imap_processing.spice.repoint import get_pointing_times
 from imap_processing.spice.spin import (
     get_instrument_spin_phase,
     get_spin_data,
 )
-from imap_processing.spice.time import ttj2000ns_to_et
+from imap_processing.spice.time import et_to_met, met_to_ttj2000ns, ttj2000ns_to_et
 
 N_SPIN_BINS = 3600
 SPIN_PHASE_BIN_EDGES = np.linspace(0, 1, N_SPIN_BINS + 1)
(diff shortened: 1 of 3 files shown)


imap_processing/hi/hi_l1c.py
```diff
@@ -121,15 +122,16 @@ def generate_pset_dataset(
...
     """
     Allocate an empty xarray.Dataset with appropriate pset coordinates.
 
     Parameters
     ----------
-    epoch_val : int
-        The starting epoch in J2000 TT nanoseconds for data in the PSET.
+    l1b_met : float
+        Any met from the input L1B DE dataset. This is used to query the
+        repoint-table data to get the start and end times of the pointing.
```
greglucas, line 134: I know that Luisa was having issues with supplying an l1b_met to the subsequent functions because it may/may-not be within the actual pointing because the l1b has unfiltered events during a repointing period. We discussed potentially making a different function that said "get pointing start and end for this pointing number" where you would supply the current pointing from the attributes of the previous dataset `dataset.attrs["Pointing"]` or how the cli was called instead. Thoughts on adding that instead of using a MET here because now you need to do time conversions.

greglucas (approved):
There are a lot of time conversions needed now. I made one comment to get your thoughts on trying to avoid those if possible.

lacoak21 (approved):

lacoak21:
I had just started working on this!  You just made it a lot easier for me :)

imap_processing/hi/hi_l1c.py
tmplummer: Yep, that would be a more robust approach. That is why the call to this function averages the CCSDS MET times from the L1B packets in the pointing... but going by ID would be even better.

tmplummer: Ticket #2353

PR #2349, opened 2025-10-29
Adjust pyproject.toml so it can be built by `uv`

pleasant-menlo:
Downstream consumers `imap_L3_processing` and the IMAP `mapping_tool` use `uv` for managing dependencies. When building a package from source, `uv` requires the `project.name` and `project.version` fields to be populated if the `project` table is present at all in the `pyproject.toml` file.  

Because the project version is determined dynamically by poetry, we have marked it as dynamic in the project table rather than setting a specific version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,9 @@ test = ["openpyxl", "pytest", "pytest-cov", "pytest-xdist", "requests", "netcdf4
 tools = ["openpyxl", "pandas"]
 map_visualization = ["healpy"]
 
+[project]
+name = "imap-processing"
+
 [project.urls]
 homepage = "https://github.com/IMAP-Science-Operations-Center"
 repository = "https://github.com/IMAP-Science-Operations-Center/imap_processing"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ map_visualization = ["healpy"]
 
 [project]
 name = "imap-processing"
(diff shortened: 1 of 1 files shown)


greglucas:
Will this cause issues with Poetry getting out of sync with its own metadata style? i.e. there are now two project "name" fields here, one for poetry and one in the project table. We had an open issue/PR with updating poetry to the new version that supports the proper metadata fields, I wonder if we need to follow through with that (probably a much larger effort involved).
https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/1250

I don't have enough patience for poetry, and would be very happy if we switch over to a different build backend (like uv), but it seems like I am alone in that preference :(

pleasant-menlo:
@greglucas We agree that it is not ideal to have the project name appear in two separate places in the `pyproject.toml` file. We tested building a wheel from the project in the same way the pipeline does; both the name and version appear to work the same as they did prior to this change. 

If this solution is not preferred, we can return to using the official releases downstream. @subagonsouth had recommended using the latest code in dev to make sure we were using up to date code.

greglucas (approved):
Cool, I just pulled it and built a wheel and it didn't raise any warnings or anything either, so this seems OK to me. We can improve it more later if we so choose.

I agree you all probably want to point to the bleeding edge for now since we don't make releases all the time.

pleasant-menlo:
Great, thanks for getting this through so quickly.

Issue #2353, opened 2025-10-29
repoint - function to get Pointing start and end times from pointing id


tmplummer:
### Algorithm Description:

Using packet or direct event MET times is a bit fragile because the MET time could potentially be outside of the pointing boundary. We should implement a function to get the pointing start/end MET for a pointing ID. This would allow ENA jobs to use the pointing number passed in by the batch starter to query for the pointing star/end MET.

PR #2355, opened 2025-10-29
ULTRA add epoch_delta to pset

lacoak21:
# Change Summary

## Overview
Add epoch_delta variable to spacecraft and heliopset. Basically I copied what Tim did in his PR. 

## Updated Files
- imap_processing/cdf/config/imap_ultra_l1c_variable_attrs.yaml
   - Add epoch delta attrs (same as Tims) 
- imap_processing/ultra/l1c/helio_pset.py, imap_processing/ultra/l1c/spacecraft_pset.py
   - make start and stop a tuple and use the diff for epoch_delta


imap_processing/ultra/l1c/helio_pset.py
```diff
@@ -128,7 +128,7 @@ def calculate_helio_pset(
     healpix = np.arange(n_pix)
 
     # Get midpoint timestamp for pointing.
-    pointing_start, pointing_stop = get_pointing_times(
+    pointing_range_met = get_pointing_times(
         et_to_met(species_dataset["event_times"].data[0])
```
tmplummer: Is this event time guaranteed to be in the Pointing? I took a mean of all the times just to make sure that an edge case didn't cause a failure.

imap_processing/ultra/l1c/helio_pset.py
```diff
@@ -128,7 +128,7 @@ def calculate_helio_pset(
     healpix = np.arange(n_pix)
 
     # Get midpoint timestamp for pointing.
-    pointing_start, pointing_stop = get_pointing_times(
+    pointing_range_met = get_pointing_times(
```
tmplummer, line 131: FYI, @greglucas mentioned that you had a case where the query time was not in the pointing. I agreed with Greg that querying by Pointing ID would be more robust. I wrote up a ticket to provide that function.
#2353

tmplummer (approved):
Looks good to me.

imap_processing/ultra/l1c/helio_pset.py
lacoak21: Good call. No it might not be.

imap_processing/ultra/l1c/helio_pset.py
lacoak21: Yes! Thanks for making the ticket. Yeah that is something I keep meaning to do.

PR #2358, opened 2025-10-30
CoDICE: more test refactor

tech3371:
# Change Summary

## Overview

More clean up

### DIFF
--- a/imap_processing/tests/codice/test_codice_l1a.py
+++ b/imap_processing/tests/codice/test_codice_l1a.py
@@ -13,7 +13,7 @@ from unittest.mock import patch
 
 import numpy as np
 import pytest
-from imap_data_access import AncillaryInput, ProcessingInputCollection, ScienceInput
+from imap_data_access import ProcessingInputCollection
 
 from imap_processing import imap_module_directory
 from imap_processing.cdf.utils import load_cdf, write_cdf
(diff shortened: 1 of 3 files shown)


imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -227,12 +227,8 @@ def test_lo_sw_species(mock_get_file_paths, codice_lut_path):
 
     val_data = load_cdf(val_path)
 
-    sci_input = ScienceInput("imap_codice_l0_lo-sw-species_20250814_v001.pkts")
-    sci_lut_input = AncillaryInput("imap_codice_l1a-sci-lut_20251007_v001.json")
-    dependency = ProcessingInputCollection(sci_input, sci_lut_input)
-
     # Process the input data
-    processed_data = process_l1a(dependency=dependency)[0]
+    processed_data = process_l1a(dependency=ProcessingInputCollection())[0]
```
greglucas, line 231: Does this work because you are setting the `side_effect` above to point at the files you need?

Do you want to make dependency optional to `process_l1a()` so you would only need to handle this in one place and not create the empty InputCollections?

greglucas (approved):

imap_processing/tests/codice/test_codice_l1a.py
tech3371: We do need input collection but since we are doing side effect in test, we don't need to pass those objects, just pass an empty collection. That's why I don't want to make it optional.

lacoak21 (approved):

tech3371:
pre-commit.ci autofix

PR #2360, opened 2025-10-31
Add Lo l2 map metadata

### LINKED ISSUE #2182: Lo: Add map variables to attributes
### Summary

I'm not sure whether these variable definitions should go in enamaps_l2-common, or if we should start breaking these out into separate instrument-specific locations.

### Requirements

_No response_

### Related tickets

_No response_

greglucas:
Take advantage of the build_cdf_dataset function to automatically handle the adding of metadata and variables. However, the Lo code has already switched to a 2d map dataset, so we don't want to transform the 1d dataset anymore, but rather allow for our own processed map dataset to be used as the base.

The second commit is largely renaming things to add background uncertainty to the dataset itself rather than having to calculate it as an intermediate product.

closes #2182

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -1273,6 +1273,7 @@ class RectangularSkyMap(AbstractSkyMap):
         descriptor: str,
         sensor: str | None = None,
         drop_vars_with_no_attributes: bool = True,
+        external_map_dataset: xr.Dataset | None = None,
     ) -> xr.Dataset:
         """
         Format the data into a xarray.Dataset and add required CDF variables.
@@ -1293,6 +1294,12 @@ class RectangularSkyMap(AbstractSkyMap):
             the output CDF doesn't have any of the intermedeiate variables left
             over from computations. Sometimes, it is useful to output the
             intermedeiate variables. To do so, set this to False.
+        external_map_dataset : xarray.Dataset, optional
+            If provided, this dataset will be used as the base dataset to
+            build the CDF dataset from, instead of using the internal map data.
+            This is useful if additional processing has been done to the map data
+            after projection, and those results need to be included in the CDF.
+            Default is None.
 
         Returns
         -------
@@ -1308,7 +1315,10 @@ class RectangularSkyMap(AbstractSkyMap):
 
         logger.info("Building CDF ready dataset from RectangularSkyMap data.")
 
-        cdf_ds = self.to_dataset()
+        if external_map_dataset is not None:
+            cdf_ds = external_map_dataset
+        else:
+            cdf_ds = self.to_dataset()
 
         # Set the value of the epoch coord
         cdf_ds = cdf_ds.assign_coords(**{CoordNames.TIME.value: [self.min_epoch]})
--- a/imap_processing/lo/l2/lo_l2.py
+++ b/imap_processing/lo/l2/lo_l2.py
@@ -98,7 +98,9 @@ def lo_l2(
     )
 
     logger.info("Step 5: Finalizing dataset with attributes")
-    dataset = finalize_dataset(dataset, descriptor)
+    dataset = sky_map.build_cdf_dataset(  # type: ignore[attr-defined]
+        instrument="lo", level="l2", descriptor=descriptor, external_map_dataset=dataset
+    )
 
     logger.info("IMAP-Lo L2 processing pipeline completed successfully")
     return [dataset]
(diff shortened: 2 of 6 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -1273,6 +1273,7 @@ def build_cdf_dataset(  # noqa: PLR0912
         descriptor: str,
         sensor: str | None = None,
         drop_vars_with_no_attributes: bool = True,
+        external_map_dataset: xr.Dataset | None = None,
```
greglucas, line 1276: I don't love this approach... One other idea was to make a separate class method on this for attributes that doesn't call the `to_dataset()`, but I wasn't sure we need another function here just to do the redirect.

Any other ideas for how to handle this? The problem is that I already have a `to_dataset()` object and that is the thing that I want to add attributes to.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -649,17 +652,20 @@ def populate_geometric_factors(
     if species == "h":
         gf_vars = {
             "energy": "Cntr_E",
-            "energy_stat_uncert": "Cntr_E_unc",
             "geometric_factor": "GF_Trpl_H",
             "geometric_factor_stat_uncert": "GF_Trpl_H_unc",
         }
+        # NOTE: From an e-mail from Nathan on 2025-09-11
+        energy_delta_hires_values = [5.43, 10.02, 18.61, 33.31, 64.98, 131.64, 262.35]
+        energy_delta_hithr_values = [8.81, 16.04, 28.50, 53.13, 105.60, 219.67, 413.60]
```
greglucas, line 660: Unclear to me how we want to handle these going forward... They aren't in an ancillary table. We can cross this hurdle later though IMO.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -850,6 +864,19 @@ def calculate_backgrounds(dataset: xr.Dataset) -> xr.Dataset:
         / dataset["geometric_factor"]
     )
 
+    # Background intensity
```
greglucas, line 867: It was decided to add intensities to the datasets since it is simple enough to do. I'm not sure why only rates were desired before, we can have both.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -1196,6 +1212,8 @@ def cleanup_intermediate_variables(dataset: xr.Dataset) -> xr.Dataset:
 
     # Only remove variables that exist in the dataset for the specific species
     potential_vars = [
+        "geometric_factor",
+        "geometric_factor_stat_uncert",
```
greglucas, line 1216: Do we want these in the map datasets? I removed it because it wasn't listed in the spreadsheet, only geometric_function was there.

imap_processing/ena_maps/ena_maps.py
tmplummer: In theory, if I have implemented everything correctly, Hi could follow how you have implemented Lo where the data gets unstacked just after binning.

I suppose we could also do something like:
```
class RectangularSkyMap(AbstracSkyMap):
    ...
    # Dataset to store the 'unstacked' dataset after the PSET -> map projection step has been completed
    data_2d: xr.Dataset | None

    def unstack_dataset():
        """Unstack the spatial coordinates"""
        self.data_2d = self.to_dataset()
        self.data_1d = None

    def build_cdf_dataset():
        if self.data_2d is None:
            self.unstack_dataset()
       # the rest of the existing code below here
```

One option that probably works better with your Lo workflow would be to make this a standalone function rather than a method on the RectangularSkyMap class.

imap_processing/lo/l2/lo_l2.py
tmplummer: It is unclear to me if geometric_factor would be useful to scientists using these products. It seems like Paul and Dan have somewhat of the opposite philosophy as the Ultra team. They are worried about the data being misused by scientists who don't understand the details, so the err on the side of including fewer variables arguing that then people can't misuse the data.

tmplummer (approved):
This all looks good to me.

imap_processing/ena_maps/ena_maps.py
greglucas: > In theory, if I have implemented everything correctly, Hi could follow how you have implemented Lo where the data gets unstacked just after binning.

Looking at your suggested code, I wonder if I could actually go in the reverse direction, where we create/load the 2D lon/lat dataset and then immediately convert it to a 1D SkyMap so that all of Lo's computations would be done in the 1D way like the other instruments.

> One option that probably works better with your Lo workflow would be to make this a standalone function rather than a method on the RectangularSkyMap class.

Maybe, but it also needs some skymap specific things, so it does kind of make sense to have it on the skymap.

PR #2365, opened 2025-11-03
2341 hi l2   improve ram mask for spacecraft frame products

### LINKED ISSUE #2341: Hi L2 - improve ram mask for spacecraft frame products
### Algorithm Description:

Current implementation just assumes PSET bins 0-1799 are ram and 1800-3599 are anti. The accuracy can be improved by using the spacecraft velocity vector. The ITF would like it implemented the more accurate way.

tmplummer:
# Change Summary

## Overview
Hi changed their mind and wanted to compute the ram mask for S/C frame PSETs explicitly rather than just using the "close enough" method. This PR adds an explicit `calculate_ram_mask` function to the ena_maps/corrections.py module. That function can then be used when preparing a PSET for projection in the L2 workflow.

### Details about the change in how the ram pixels are being identified
The "close enough" method initially implemented relied on the definition of the spin-phase and made the approximation that the spacecraft velocity vector is normal to the spacecraft spin axis. If that approximation were true, then PSET bins with spin-phase in the range [0, 0.5) would be ram bins and those with spin-phase in the range [0.5, 1) would be anti-ram. Because of how the PSET is defined this happens to mean that the first half of the PSET bins are ram and the second half are anti-ram.

The more exact way to determine if a PSET bin is part of the ram set of bins is to do some vector math. The vector math is to simply find the dot product of the look direction of each PSET bin with the spacecraft velocity vector. This calculates the component of the look direction that points along the s/c velocity vector. Any positive result means that the look direction is looking ram and any negative result indicates anti-ram. This is what the new implementation does.

### Notes for future changes
I don't love some of the organization of the Hi L2 code and will write up a ticket to extract some functions so that the workflow is a bit cleaner. I am also thinking that the CG corrections.py functions should just act on a xr.Dataset rather than on a LoHiBasePsetSubclass.

## Updated Files
   - Add logger message to LoHiBasePointingSet when updating az/el points
   - Remove HiPointingSet "close enough" calculation of ram_mask
   - Use middle of Pointing as defined by PSET epoch + epoch_delta / 2 to query spacecraft velocity
   - Add public `calculate_ram_mask` function.
   - Make `add_spacecraft_velocity_to_pset` a public function.
- imap_processing/hi/hi_l2.py
   - Add explicit calls to calculate ram mask for S/C frame PSETs
   - Add test coverage for calculate_ram_mask function

Closes: #2341


imap_processing/ena_maps/utils/corrections.py
```diff
@@ -532,18 +531,64 @@ def _calculate_compton_getting_transform(
...
+            np.ones_like(longitude.values),
+            longitude.values,
+            latitude.values,
+        ],
+        axis=-1,
+    )
+    cartesian_source_direction = xr.DataArray(
+        geometry.spherical_to_cartesian(spherical_coords),
+        dims=[*longitude.dims, CoordNames.CARTESIAN_VECTOR.value],
+    )
```
lacoak21, line 577: OK this makes way more sense (higher level) :). Thanks for adding that description in.

lacoak21 (approved):
LGTM! Great tests

PR #2366, opened 2025-11-03
Codice-hi l1b validation

laspsandoval:
# Change Summary

## Overview
Validating CoDICE-Hi using test data and Tenzin's code.

## Updated Files
constants.py
- Change based on description in Algorithm Document page 52.
external_test_data_config.py
- Updated to recent version of test data

## Testing
test_process_codice.py

### DIFF
--- a/imap_processing/codice/constants.py
+++ b/imap_processing/codice/constants.py
@@ -713,7 +713,7 @@ L1B_DATA_PRODUCT_CONFIGURATIONS: dict[str, dict] = {
         "num_spins": 16,
     },
     "hi-ialirt": {
-        "num_spin_sectors": 24,
+        "num_spin_sectors": 6,
         "num_spins": 4,
     },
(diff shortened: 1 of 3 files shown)


imap_processing/codice/constants.py
```diff
@@ -713,7 +713,7 @@
         "num_spins": 16,
     },
     "hi-ialirt": {
-        "num_spin_sectors": 24,
+        "num_spin_sectors": 6,
```
tech3371, line 716: interesting. I don't know enough about why your spin sector is 6 instead of 24.

tech3371 (approved):

imap_processing/codice/constants.py
laspsandoval: Having pretty much no understanding of CoDICE I also do not know. But it was a change based on description in Algorithm Document page 52.

PR #2373, opened 2025-11-03
ULTRA l1b use energy bin's geometric mean for scattering culling

lacoak21:
# Change Summary

## Overview
L1b code was using the "event energy" instead of the geometric mean of the energy bin of that event's energy. Nick was doing this in his code and noticed it was wrong and asked me to check our code. 

## Updated Files
    - Use the energy bin geometric mean.

### DIFF
--- a/imap_processing/quality_flags.py
+++ b/imap_processing/quality_flags.py
@@ -43,7 +43,7 @@ class ImapDEOutliersUltraFlags(FlagNameMixin):
     NONE = CommonFlags.NONE
     FOV = 2**0  # bit 0
     PHCORR = 2**1  # bit 1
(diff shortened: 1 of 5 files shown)


laspsandoval (approved):
Looks good to me.

imap_processing/quality_flags.py
```diff
@@ -44,6 +44,7 @@ class ImapDEOutliersUltraFlags(FlagNameMixin):
     FOV = 2**0  # bit 0
     PHCORR = 2**1  # bit 1
     COINPH = 2**2  # bit 4 # Event validity
```
greglucas: Update the comment above this to be bit 2.

greglucas (approved):

PR #2375, opened 2025-11-03
ULTRA l1b correct particle direction velocity 

lacoak21:
# Change Summary

## Overview
The psets created on Friday have counts where there was no exposure factor (they were on the opposite side of the map). This led the instrument team to believe there was a sign flip somewhere when calculation velocities in the instrument frame. Sure enough there was. Apparently at one point Larry was calculating it this way as well so the code was validating but it needs to be corrected now. The issue is in equation 3. That R factor should be negated and our code was not doing that. 

Screenshot 2025-11-03 at 12 28 09 PM

## Updated Files
   - fix velocity valculations

## Testing
Fix tests

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
@@ -267,19 +267,19 @@ def test_get_de_velocity(test_fixture):
 
     np.testing.assert_allclose(
         v_x[test_tof > 0],
-        df_ph["vx"].astype("float").values[test_tof > 0],
+        -df_ph["vx"].astype("float").values[test_tof > 0],
         atol=1e-01,
         rtol=0,
     )
     np.testing.assert_allclose(
         v_y[test_tof > 0],
-        df_ph["vy"].astype("float").values[test_tof > 0],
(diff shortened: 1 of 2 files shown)


laspsandoval (approved):
Nice catch.

greglucas (commented):
I think you can rewrite this to look more like the equations. I'm not sure why r_hat = -v_hat as that doesn't seem quite right and there should be a tof factor difference in there I think?

```python
    r = np.empty((len(d), 3), dtype=np.float32)
    r[:, 0] = (front_position[0] - back_position[0]) * 0.1
    r[:, 1] = (front_position[1] - back_position[1]) * 0.1
    r[:, 2] = d * 0.1

    v = -r / tof * 1e3

    # Ignore bad TOFs
    v[tof < 0, :] = FILLVAL_FLOAT32

    # Get the unit vectors
    r_hat = r / np.linalg.norm(r, axis=1)
    v_hat = v / np.linalg.norm(v, axis=1)

    return v, v_hat, r_hat
```

edit:
> I'm not sure why r_hat = -v_hat as that doesn't seem quite right and there should be a tof factor difference in there I think?

I think this is because they are the unit vectors, so one is the inverse and the tof is normalized out... I wasn't paying attention at first, so this is likely fine as-is.

PR #2376, opened 2025-11-03
ULTRA l2 remove background rate subtraction from corrected counts

lacoak21:
# Change Summary

## Overview
The ULTRA team would like the background rates to be reported but not subtracted when calculating the corrected counts. Once they get a sense of what the background rates are then they will ask to add it back. 

## Updated Files
   - Do not subtract bg_rates for now. 

## Testing
Fix test. I had to raise the tolerance up for the test that compares the helio intensity to the rectangular intensity.

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l2.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l2.py
@@ -178,9 +178,9 @@ class TestUltraL2:
         )
 
         # Estimate the expected ena_intensity and its uncertainty
-        expected_ena_intensity = (
-            (10 * solid_angle_ratio_map_to_pset / 1) - 1 * solid_angle_ratio_map_to_pset
-        ) / (1 * hp_skymap.solid_angle * 1)
+        expected_ena_intensity = (10 * solid_angle_ratio_map_to_pset / 1) / (
+            1 * hp_skymap.solid_angle * 1
+        )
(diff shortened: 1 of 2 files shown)


greglucas (approved):
This seems like something that will need to be remembered later to remove and how do we decide when to remove it... We should follow up with Ultra on Wednesday about how to go about this rather than one-off notes to us as this seems like something easy to overlook in future science papers to me.

lacoak21:
> This seems like something that will need to be remembered later to remove and how do we decide when to remove it... We should follow up with Ultra on Wednesday about how to go about this rather than one-off notes to us as this seems like something easy to overlook in future science papers to me.

Yeah that is a good point. I can definitely see this slipping through the cracks.

There are actually a lot of things like this that they have asked to be "turned off" for now that we need to add back later. I will bring these up Wednesday 

- energy bins (we are already going to talk about that Wednesday) 
- Boundary scale factors 
- scattering thresholds. 
- background rates.

greglucas:
> There are actually a lot of things like this that they have asked to be "turned off" for now that we need to add back later. I will bring these up Wednesday

I wonder if we can add a global flag somewhere that you can add so we can do something like `if ONLY_DURING_COMMISSIONING:` so that you can set that flag or something along those lines and it automatically happens the other way after that then. It will be good to bring up with them for discussion 👍

Issue #2297, opened 2025-10-15
BUG - MAG L1D jobs always fail


alastairtree:
### Description of the issue

MAG L1D processing jobs always fail with the following error:

```
Toolkit version: CSPICE_N0067
SPICE(NOFRAMECONNECT) --
At epoch 8.1332646918235E+08 TDB (2025 OCT 10 00:01:09.182 TDB), there is insufficient information available to transform from reference frame -43000 (IMAP_SPACECRAFT) to reference frame -43901 (IMAP_DPS). IMAP_SPACECRAFT is a CK frame; a CK file containing data for instrument or structure -43000 at the epoch shown above, as well as a corresponding SCLK kernel, must be loaded in order to use this frame. IMAP_DPS is a CK frame; a CK file containing data for instrument or structure -43901 at the epoch shown above, as well as a corresponding SCLK kernel, must be loaded in order to use this frame. Failure to find required CK data could be due to one or more CK files not having been loaded, or to the epoch shown above lying within a coverage gap or beyond the coverage bounds of the loaded CK files. It is also possible that no loaded CK file has required angular velocity data for the input epoch, even if a loaded CK does have attitude data for that epoch. You can use CKBRIEF with the -dump option to display coverage intervals of a CK file.
pxform_c --> PXFORM --> REFCHG
```

### Code Snippet:

```python
Code
```

greglucas:
@subagonsouth Do you know if this is due to the DPS kernel issues you're working on resolving?

@alastairtree out of curiosity why is the l1d data being produced in the DPS frame? I thought you'd want the data to be in the RTN, GSM, etc. frames as the DPS is really only relevant for the imagers.

maxinelasp:
@greglucas  DPS is needed for calculating L1D

greglucas:
> @greglucas DPS is needed for calculating L1D

OK, so you aren't producing data in the DPS frame, it is an intermediate for some reason? I guess I haven't read any of the algorithm documents so I'm just behind. I thought the ENAs were the only ones who would be interested in the DPS per-pointing frame. Also note that this is on a different cadence than the MAG daily data and also will have a gap during repointing maneuvers every single day, so if you need a continuous frame this might not provide that.

alastairtree:
[...]

Issue #2295, opened 2025-10-14
MAG: L2 produced in all reference frames


mhairifin:
### Summary

The MAG L2 Processing currently only produces data in the SRF (spacecraft reference frame) and DSRF (despun spacecraft reference frame). It should also produce L2 in RTN, GSE, and GSM

### Requirements

L2 produced in all required reference frames
  - [ ] Produced in RTN
  - [ ] Produced in GSE
  - [ ] Produced in GSM

alastairtree:
I think this is done and working now @maxinelasp ? @mhairifin has just been pushing through a bunch of data for us to check.

maxinelasp:
I expect this to be working now, so if it looks good on the MAG side then we can close this ticket as completed by #2378!

PR #2378, opened 2025-11-04
task: Add more SPICE frames for MAG L2 processing and fix issue with SPICE epochs missing

mfacchinelli:
# Change Summary

## Overview
This change:
1. Adds transformation of MAG L2 data to 3 new frames (GSE, GSM, RTN)
2. Fixes a few issues with processing MAG L2 data due to:
  a. DSRF transformation not being defined during periods of thruster firing (transformation matrix is now all `NaN`s)
  b. Truncates L2 data to the specified day at the beginning of L2 processing
  c. Truncation also crops `epoch_et` 

> [!NOTE]
> The current implementation will require a lot of memory for processing Burst mode in 128 cadence (as the data for 1 entire day will be held in memory for 5 different frames).

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
N/A

## Updated Files
* `imap_mag_global_cdf_attrs.yaml`
  - Adds definition for GSM reference frame
* `mag_l2.py`:
  - Adds conversion to all supported reference frames
* `mag_l2_data.py`:
  - Adds enum for GSM reference frame
  - Also crops `epoch_et` when truncating
* `geometry.py`:
  - If SPICE tranformation fails becuase of `SpiceNOFRAMECONNECT`, i.e., no SPICE data is available for a specific timestamp (this happens every day between 10 and 10:30 AM for the despun reference frame, due to thruster firing), a `NaN` matrix is returned, instead of an error being thrown
* `test_mag_l2.py`:
  - Modifies existing test to expect 5 datasets for each frame
  - Adds a test to make sure `NaN` matrices are supported
* `test_geometry.py`:
  - Adds a test to make sure `get_rotation_matrix` returns `NaN` matrix when despun frame not defined for input `et`

## Testing
Testing was added to cover all cases.

### DIFF
--- a/imap_processing/mag/l2/mag_l2.py
+++ b/imap_processing/mag/l2/mag_l2.py
@@ -1,5 +1,7 @@
 """Module to run MAG L2 processing."""
 
+import logging
+
 import numpy as np
 import xarray as xr
 
@@ -7,6 +9,8 @@ from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.mag import imap_mag_sdc_configuration_v001 as configuration
 from imap_processing.mag.constants import DataMode
 from imap_processing.mag.l2.mag_l2_data import MagL2, ValidFrames
+
+logger = logging.getLogger(__name__)
 
 
 def mag_l2(
@@ -105,17 +109,31 @@ def mag_l2(
         frame=instrument_frame,
     )
 
+    # L2 data should not include the extra 30 min padding either side
+    # Note: this must be done after applying offsets and timedeltas
+    l2_data.truncate_to_24h(day)
+
     attributes = ImapCdfAttributes()
     attributes.add_instrument_global_attrs("mag")
     attributes.add_instrument_variable_attrs("mag", "l2")
 
     # Rotate from the MAG frame into the SRF frame
-    l2_data.rotate_frame(ValidFrames.SRF)
-    imap_srf = l2_data.generate_dataset(attributes, day)
-    l2_data.rotate_frame(ValidFrames.DSRF)
-    imap_dsrf = l2_data.generate_dataset(attributes, day)
+    frames: list[xr.Dataset] = []
 
-    return [imap_dsrf, imap_srf]
+    for frame in [
+        ValidFrames.SRF,
+        ValidFrames.GSE,
+        ValidFrames.GSM,
+        ValidFrames.RTN,
+        ValidFrames.DSRF,  # should be last as some vectors may become NaN
+    ]:
+        try:
+            l2_data.rotate_frame(frame)
+            frames.append(l2_data.generate_dataset(attributes, day))
+        except Exception:
+            logger.exception(f"Error rotating to frame {frame.name}")
+
+    return frames
 
 
 def retrieve_matrix_from_l2_calibration(
--- a/imap_processing/mag/l2/mag_l2.py
+++ b/imap_processing/mag/l2/mag_l2.py
@@ -127,11 +127,8 @@ def mag_l2(
         ValidFrames.RTN,
         ValidFrames.DSRF,  # should be last as some vectors may become NaN
     ]:
-        try:
-            l2_data.rotate_frame(frame)
-            frames.append(l2_data.generate_dataset(attributes, day))
-        except Exception:
-            logger.exception(f"Error rotating to frame {frame.name}")
+        l2_data.rotate_frame(frame)
+        frames.append(l2_data.generate_dataset(attributes, day))
 
     return frames
 
--- a/imap_processing/mag/l2/mag_l2_data.py
+++ b/imap_processing/mag/l2/mag_l2_data.py
@@ -319,6 +319,7 @@ class MagL2L1dBase:
             self.vectors,
             from_frame=self.frame.value,
             to_frame=end_frame.value,
+            allow_spice_noframeconnect=True,
         )
         self.frame = end_frame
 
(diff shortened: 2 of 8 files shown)


maxinelasp:
Relates to: #2297, #2295

imap_processing/spice/geometry.py
```diff
@@ -388,8 +393,21 @@ def get_rotation_matrix(
         `et` is a np.ndarray, the returned rotation matrix is of shape `(n, 3, 3)`
         where `n` matches the number of elements in et.
     """
+
+    def pxform_error_handler(  # type: ignore[no-untyped-def]
+        *arg, **kwargs
+    ):  # numpydoc ignore=GL08
+        try:
+            return spiceypy.pxform(*arg, **kwargs)
+        except spiceypy.utils.exceptions.SpiceNOFRAMECONNECT:
```
tmplummer: This implementation concerns me. I want this function to fail loudly by default rather than just logging an exception level message. I would be more comfortable with adding a keyword argument called something like `allow_spice_noframeconnect` that defaults to False.
```
try:
    return spiceypy.pxform(*arg, **kwargs)
except spiceypy.utils.exceptions.SpiceNOFRAMECONNECT as e:
    if not allow_spice_noframeconnect:
        raise(e)
    logger.debug(f"Returning NaN rotation matrix due to spiceypy error: {e}")
    return np.full((3, 3), np.nan)
```

imap_processing/tests/spice/test_geometry.py
```diff
@@ -356,6 +356,25 @@ def test_get_rotation_matrix(furnish_kernels):
...
+        "imap_science_100.tf",
+        "sim_1yr_imap_attitude.bc",
+        "sim_1yr_imap_pointing_frame.bc",
+        "de440s.bsp",
+    ]
+    with furnish_kernels(kernels):
+        # Midnight is not defined in pointing frame
+        et = spiceypy.utc2et("2029-01-01T00:00:00.000")
+        rotation = get_rotation_matrix(et, SpiceFrame.IMAP_MAG_O, SpiceFrame.IMAP_DPS)
+        assert np.isnan(rotation).all()
```
tmplummer, line 385: It would be good to also check that the vectorized form works:
```
# one hour after midnight should have coverage
ets = np.array([et, et + 3600])
rotations = get_rotation_matrix(ets, SpiceFrame.IMAP_MAG_O, SpiceFrame.IMAP_DPS)
assert rotations.shape == (2, 3, 3)
assert np.isnan(rotations[0]).all()
assert np.isfinite(rotations[1]).all()
```

imap_processing/mag/l2/mag_l2.py
```diff
@@ -105,17 +109,31 @@ def mag_l2(
...
+        ValidFrames.SRF,
+        ValidFrames.GSE,
+        ValidFrames.GSM,
+        ValidFrames.RTN,
+        ValidFrames.DSRF,  # should be last as some vectors may become NaN
+    ]:
+        try:
+            l2_data.rotate_frame(frame)
+            frames.append(l2_data.generate_dataset(attributes, day))
+        except Exception:
```
tmplummer: Broad exceptions like this are generally bad and in this case all details about what caused the failure to rotate the data to the desired frame is lost.

Are there specific reasons you expect this to fail and can you handle those specific exceptions?

tmplummer (commented):
I know this is still just a draft, but I thought that I would leave some comments anyway.

imap_processing/spice/geometry.py
mfacchinelli: Makes sense, done!

imap_processing/tests/spice/test_geometry.py
mfacchinelli: I've added this test, and a test for the default `allow_spice_noframeconnect` value, where we throw an error.

imap_processing/mag/l2/mag_l2.py
mfacchinelli: I've removed this as this change addresses the failures we have seen in the past.

Btw, the `logger.exception` method does log the exception information automatically (including stacktrace), without having to pass the exception directly:
https://docs.python.org/3/library/logging.html#logging.Logger.exception

The same is true when passing `exc_info` to other logging methods, like I've done for the `debug` call.

mfacchinelli:
> I know this is still just a draft, but I thought that I would leave some comments anyway.

Thanks Tim, I've implemented the feedback. I'll get it tested in the dev environment with the files that have failed in the past.

tmplummer:
This all looks reasonable to me. @mfacchinelli, will you please convert this to a regular PR, as opposed to a Draft, and add @maxinelasp and @greglucas as additional reviewers?

mfacchinelli:
> This all looks reasonable to me. @mfacchinelli, will you please convert this to a regular PR, as opposed to a Draft, and add @maxinelasp and @greglucas as additional reviewers?

I can't add reviewers, but the tag should also work!

imap_processing/mag/l2/mag_l2_data.py
```diff
@@ -25,6 +25,7 @@ class ValidFrames(Enum):
     DSRF = SpiceFrame.IMAP_DPS
     SRF = SpiceFrame.IMAP_SPACECRAFT
     GSE = SpiceFrame.IMAP_GSE
+    GSM = SpiceFrame.IMAP_GSM
```
maxinelasp, line 28: Do we need to add GSM for the L1D processing as well?

maxinelasp (approved):
This looks good to me! I am adding the SPICE kernels as required for L2 that I missed (ephemeris kernels). Let me know when you're done testing and ready to merge it. Thanks for the fix!

imap_processing/mag/l2/mag_l2_data.py
mfacchinelli: No, GSM is not used for L1d

mfacchinelli:
> This looks good to me! I am adding the SPICE kernels as required for L2 that I missed (ephemeris kernels). Let me know when you're done testing and ready to merge it. Thanks for the fix!

Thanks @maxinelasp. I've tested it locally, and it generated all the expected L2 files. Mhairi had a look at them and she was happy.

For me this is ready to merge.

PR #2380, opened 2025-11-04
ULTRA l1b fix repoint filtering

lacoak21:
# Change Summary

## Overview
Events that happened during the repointing were not getting filtered out in l1c. `get_de_rejection_mask`  was only filtering `quality_scattering` and `quality_outliers` and i added the "during a repoint" flag to the `quality_attitude` flags. I actually think it should be moved to the ImapDEOutliersUltraFlags because these are related to the direct events. These flags are getting filtered out already. In the discussion about this with Bob, he mentioned that any spins during the repoint should not be used in exposure/sensitivity/dead time corrections. 

## Updated Files
- imap_processing/quality_flags.py
   - Move flag to de outliers flags. 
   - Skip calculating deadtimes/exposure factor with spins that happened during the repoint.

### DIFF
--- a/imap_processing/ultra/l1b/de.py
+++ b/imap_processing/ultra/l1b/de.py
@@ -325,10 +325,10 @@ def calculate_de(
         in_pointing = calculate_events_in_pointing(
             repoint_id, event_times[valid_events]
         )
+        events_to_flag = np.zeros(len(quality_flags), dtype=bool)
+        events_to_flag[valid_events] = ~in_pointing
         # Update quality flags for valid events that are not in the pointing
-        quality_flags[valid_events][~in_pointing] |= (
-            ImapDEOutliersUltraFlags.DURINGREPOINT.value
-        )
+        quality_flags[events_to_flag] |= ImapDEOutliersUltraFlags.DURINGREPOINT.value
         # Update valid_events to only include times within a pointing
         valid_events[valid_events] &= in_pointing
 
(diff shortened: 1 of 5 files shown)


imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -455,6 +455,12 @@ def get_spacecraft_exposure_times(
     nominal_deadtime_ratios : np.ndarray
         Deadtime ratios at each spin phase step (1ms res).
     """
+    # filter rates dataset to only include data during a pointing
```
lacoak21, line 458: Unsure if I need to use the goodtimes mask here. I will discuss this with the UTLRA team.

imap_processing/ultra/l1b/de.py
```diff
@@ -326,10 +325,10 @@ def calculate_de(
         in_pointing = calculate_events_in_pointing(
             repoint_id, event_times[valid_events]
         )
+        events_to_flag = np.zeros(len(quality_flags), dtype=bool)
+        events_to_flag[valid_events] = ~in_pointing
         # Update quality flags for valid events that are not in the pointing
-        quality_flags[valid_events][~in_pointing] |= (
```
lacoak21, line 330: This was such a sneaky bug. The quality flags were actually NOT getting updated because `quality_flags[valid_events]` was creating a copy. Ive had this issue with numpy before but somehow I end up doing it again eventually 😆

greglucas: Nice catch! That is sneaky.

greglucas (approved):
Is there any mock test you can provide to test the calculate_de updates?

PR #2381, opened 2025-11-04
2374 lo l2   integrate cg correction for hf maps - prepare for lo integration

tmplummer:
# Change Summary

## Overview
There are two main changes in this PR:
- Make CG related functions in corrections.py act on xarray.Dataset objects rather than LoHiBasePsetSubclass objects. This need was driven by how the Lo L2 workflow is implemented (which I like much better than how I did it in Hi).
- Move the function for calculating the pset ram_mask variable into the corrections.py module.

Other changes were cascaded from the above main changes:
- Update Hi L2 processing workflow to only instantiate the HiPointingSet object just for use with the SkyMap projection method. (I will write up a ticket to extract functions similar to the Lo workflow)
- Add more extensive tests for calculate_ram_mask function.


imap_processing/ena_maps/ena_maps.py
```diff
@@ -649,28 +653,27 @@ class HiPointingSet(LoHiBasePointingSet):
...
-            }
-        )
+        # Rename some PSET vars to match L2 variables (if necessary)
+        rename_dict = {
+            key: value
+            for key, value in self.l1c_to_l2_var_mapping.items()
+            if key in self.data
+        }
+        if rename_dict:
+            self.data = self.data.rename(rename_dict)
```
greglucas: Do you need the "if" or can you call rename with an empty mapping and just carry on?

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -321,72 +321,69 @@ class or child classes.
...
+    Adds the following DataArrays to input dataset:
     - "sc_velocity": Spacecraft velocity vector (km/s) with dims ["x_y_z"]
     - "sc_direction_vector": Spacecraft velocity unit vector with dims ["x_y_z"]
     """
     # Compute ephemeris time (J2000 seconds) of PSET midpoint time
-    # TODO: Use the Pointing midpoint time. Epoch should be start time
-    #     but use it until we can make Lo and Hi PSETs have a consistent
-    #     variable to hold the midpoint time.
-    et = ttj2000ns_to_et(pset.data["epoch"].values[0])
+    et = ttj2000ns_to_et(pset["epoch"].values[0] + pset["epoch_delta"].values[0] / 2)
```
greglucas, line 350: Might be useful to remind ourselves here with a comment that pset["epoch"] is the left edge and the delta is the full width.

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -399,7 +396,7 @@ def _add_cartesian_look_direction(pset: LoHiBasePsetSubclass) -> LoHiBasePsetSub
     )
 
     # Convert to Cartesian coordinates and store as DataArray
-    pset.data["look_direction"] = xr.DataArray(
```
greglucas, line 401: I really like the dropping of all these "data" and "values" :)

greglucas (approved):

PR #2384, opened 2025-11-05
FIX: Apply flux correction to background intensities for Lo

### LINKED ISSUE #2377: BUG - Lo L2 - Flux correction not applied to backgrounds
### Description of the issue

I noticed that the flux correction is not being applied to backgrounds or background uncertainties.

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

If I understand the alg. doc. correctly, I think that the PowerLawFluxCorrector.apply_flux_correction() method just needs to be called a second time with the background and background uncertainty intensities passed in instead of ena_intensities.

greglucas:
Lo carries their background intensities separately, so we need to correct those as well.

closes #2377


tmplummer (approved):
LGTM

PR #2385, opened 2025-11-05
Get pointing start and stop from repoint id

### LINKED ISSUE #2327: ULTRA l1b repoint helper function
Move the repoint_id -> pointing start and pointing stop function to repoint.py and clean up handling.

### LINKED ISSUE #2353: repoint - function to get Pointing start and end times from pointing id

lacoak21:
# Change Summary
closes #2353 
closes #2327 
## Overview
Get pointing times using repoint id. This is probably more reliable for ultra than using met times in case an event falls within a repoint. 

## Updated Files
   - Get pointing times from id. I made a new function for this instead of adding it to get_pointing_times because I felt that there was enough of a difference
   - use new helper function
-

### DIFF
--- a/imap_processing/spice/repoint.py
+++ b/imap_processing/spice/repoint.py
@@ -260,7 +260,7 @@ def get_pointing_times_from_id(repoint_id: int | str) -> tuple[float, float]:
     next_repoint_row = repoint_df[repoint_df["repoint_id"] == repoint_id + 1]
     if next_repoint_row.empty:
         raise ValueError(
-            "Pointing end time not found. Either current "
+            f"Pointing end time not found for repoint ID {repoint_id}. Either current "
             "pointing is ongoing or the repoint table is outdated."
         )
 
--- a/imap_processing/spice/repoint.py
+++ b/imap_processing/spice/repoint.py
@@ -2,6 +2,7 @@
 
 import functools
 import logging
+import re
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


imap_processing/spice/repoint.py
```diff
@@ -225,6 +226,49 @@ def get_pointing_times(met_time: float) -> tuple[float, float]:
...
+    repoint_df = get_repoint_data()
+    # To find the pointing start and stop, get the end of the current repointing
+    # and the start of the next repointing
+    repoint_row = repoint_df[repoint_df["repoint_id"] == repoint_id]
+    if repoint_row.empty:
+        raise ValueError(f"Repoint ID {repoint_id} not found in repoint table.")
+    next_repoint_row = repoint_df[repoint_df["repoint_id"] == repoint_id + 1]
+    if next_repoint_row.empty:
+        raise ValueError(
+            "Pointing end time not found. Either current "
```
tmplummer: It would be useful to log the repoint_id in this message too.

imap_processing/tests/spice/test_repoint.py
```diff
@@ -82,6 +82,38 @@ def test_get_pointing_times(fake_repoint_data):
     assert pointing_end_time == expected_times[1]
 
 
+def test_get_pointing_times_from_id(fake_repoint_data):
+    """Test coverage for get_pointing_times_from_id function."""
+    id = 0
+    expected_times = (5.1, 15.2)
```
tmplummer, line 93: Can you add a comment as to why these are the expected values?

tmplummer (approved):
Looks good. Just a couple of nits.

imap_processing/spice/repoint.py
lacoak21: ah yep

lacoak21:
closes #2327

PR #2386, opened 2025-11-05
Ultra ena rates l1b

lacoak21:
# Change Summary

## Overview
The spin edges were getting converted to uint16. This was causing the spins to roll over when it reached the max uint16 val. This was causing ena_rates to be an array of zeros because none of the spins were within the spin_edges. 

`unconverted spins [140309 140310 140311 ... 146033 146034 146035]
converted spins [ 9237  9238  9239 ... 14961 14962 14963]`

This was a remanent of when the spin_number was uint8. @laspsandoval helped me confirm this was the correct way to fix this issue. 

## Updated Files
   - Dont convert spin numbers to uint16.


laspsandoval (approved):
Thanks for doing this Luisa.

PR #2391, opened 2025-11-06
CoDICE l2 lo attributes 

lacoak21:
# Change Summary

## Overview
CoDICE lo l2 attrs. These have been validated by ISTP. There is a warning in the species cdfs that a dim is < 2 (spin sectors) I will have to let the CoDICE team know and see what they think. 

## Updated Files
- imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml
   - fix global attrs 
- imap_processing/cdf/config/imap_codice_l2-lo-angular_variable_attrs.yaml
   - Add attrs
- imap_processing/cdf/config/imap_codice_l2-lo-species_variable_attrs.yaml  
   - Add attrs
   - energy_Table dim -> esa_step

### DIFF
--- a/imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_global_cdf_attrs.yaml
@@ -249,13 +249,13 @@ imap_codice_l2_lo-nsw-species:
 
 imap_codice_l2_lo-sw-angular:
   <<: *instrument_base
-  Data_type: L2_lo-sw-species>Level-2 Lo Sunward Angular Intensities Data
+  Data_type: L2_lo-sw-angular>Level-2 Lo Sunward Angular Intensities Data
   Logical_source: imap_codice_l2_lo-sw-angular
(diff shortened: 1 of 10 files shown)


tech3371 (approved):

tech3371 (commented):
can we do this in different PR? My plan is not use L2 yaml files later and carry from L1A onwards. This way, we have one file to update and update attrs in code with units as such.

lacoak21:
> can we do this in different PR? My plan is not use L2 yaml files later and carry from L1A onwards. This way, we have one file to update and update attrs in code with units as such.

Are you referring to the spin_angles -> spin_angle?

tech3371:
> > can we do this in different PR? My plan is not use L2 yaml files later and carry from L1A onwards. This way, we have one file to update and update attrs in code with units as such.
> 
> Are you referring to the spin_angles -> spin_angle?

Ignore my comment. After talking, this approach made sense since L2 is very different from L1A and L1B

PR #2395, opened 2025-11-07
PERF: Improve Lo direct event packet parsing

greglucas:
Use raw bytes instead of going to string representation.

Access the .values numpy arrays outside of the for loops and store the views external to the hot-loops. This was causing a significant slowdown because of all the array creation within the loops.

Currently, this takes ~7 hours to process L0 data and times out on our resources. With this update, it takes a few minutes locally.

**Before**
I killed the process because I wasn't willing to wait 7 hours, but you can see that the parse_events was taking up longer and longer time compared to create_datasets.
image

**After**
I was able to run the entire processing and the parse_events is less than the create_datasets now.
image


sdhoyt (approved):
Thank you!

PR #2396, opened 2025-11-07
I-ALiRT - SC packet updates for MAG

laspsandoval:
# Change Summary

## Overview
We had made the assumption that the spacecraft packet spin period, RA, and Dec were in units of radians. This PR addresses that issue. It also pull the values out of the for loop so that all of the values can be used during interpolation. This vastly improved the z vector component stability, but after some analysis it does seem that more work is needed. Incremental progress.

Note, I also reduced the number of decimals after a value for the MAG data in hopes that that will improve the latency issue we are having with the database (in addition to the changes that we are making to the database itself).

## Updated Files
- parse_mag.py
   - Conversion of spacecraft packet

## Testing
See attached.
[mag_inertial_frame.xlsx](https://github.com/user-attachments/files/23425467/mag_inertial_frame.xlsx)

### DIFF
--- a/imap_processing/ialirt/l0/parse_mag.py
+++ b/imap_processing/ialirt/l0/parse_mag.py
@@ -568,6 +568,19 @@ def process_packet(
     pkt_counter = get_pkt_counter(accumulated_data["mag_status"])
     accumulated_data["pkt_counter"] = pkt_counter
 
+    # Convert from incrementing uint16 (0-65535) to radians.
+    sc_spin_phase_rad = accumulated_data["sc_spin_phase"].astype(float) * (
+        2 * np.pi / 65535.0
+    )
+    sc_inertial_right = accumulated_data["sc_inertial_right"].astype(float) * (
+        2 * np.pi / 65535.0
+    )
+    sc_inertial_decline = (
+        accumulated_data["sc_inertial_decline"].astype(float) / 65535.0
+    ) * np.pi - (np.pi / 2)
+
+    attitude_time = met_to_ttj2000ns(accumulated_data["met"])
+
     grouped_data = find_groups(accumulated_data, (0, 3), "pkt_counter", "met")
 
     unique_groups = np.unique(grouped_data["group"])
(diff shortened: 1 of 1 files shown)


imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -568,6 +568,19 @@ def process_packet(
     pkt_counter = get_pkt_counter(accumulated_data["mag_status"])
     accumulated_data["pkt_counter"] = pkt_counter
 
+    # Convert from incrementing uint16 (0-65535) to radians.
+    sc_spin_phase_rad = accumulated_data["sc_spin_phase"].astype(float) * (
+        2 * np.pi / 65535.0
```
greglucas, line 573: All of these should be 65536 since we have 2**16 possible values in the range I think.
```suggestion
        2 * np.pi / 65536.0
```

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -568,6 +568,19 @@ def process_packet(
...
+    )
+    sc_inertial_right = accumulated_data["sc_inertial_right"].astype(float) * (
+        2 * np.pi / 65535.0
+    )
+    sc_inertial_decline = (
+        accumulated_data["sc_inertial_decline"].astype(float) / 65535.0
```
greglucas, line 579: Rename to declination? I don't know that decline is the right wording.
```suggestion
        accumulated_data["sc_inertial_declination"].astype(float) / 65535.0
```

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -568,6 +568,19 @@ def process_packet(
     pkt_counter = get_pkt_counter(accumulated_data["mag_status"])
     accumulated_data["pkt_counter"] = pkt_counter
 
+    # Convert from incrementing uint16 (0-65535) to radians.
```
greglucas, line 571: Do you need/want to update the accumulated_data[var] variables themselves so that you are storing the values you want to use later?

imap_processing/ialirt/l0/parse_mag.py
greglucas: Actually, I think it will depend on which variables you are wanting to do the math over. Latitude you do want to go [-pi, pi] inclusive, but I think here for the longitude you would want to exclude the final point because it is a duplicate [0, 2pi). Maybe it is easiest/best to directly use the conversion factors the MOC provided rather than calculate it here?

greglucas (approved):

greglucas:
@laspsandoval I merged this in because my comments were minor and we can update more later if we want. I did check the numbers in the spreadsheet and the division you have is right. I'm just not sure it totally lines up with what I was expecting for the off-by-one issues. It will be a minor adjustment if anything though, so can be done later. I wanted to get this in with the SWAPI updates so we can test and see how things look and make minor updates later.

PR #2397, opened 2025-11-09
FIX: Lo l1b_de needs to expand l1a met to each direct event

greglucas:
Previously, it was only calculating this per ASC, where there are multiple direct events per ASC. We can repeat the individual MET entries by the number of direct events in that ASC.

The test was assuming MET in l1a was the number of direct events, this turns out not to be the case. I think this is right here, but we could also update the l1a dataset to put MET in terms of direct_events as well.


sdhoyt (approved):
Thanks for figuring this issue out!

PR #2399, opened 2025-11-09
FIX: SWAPI datetimes were not offset from epoch and ESA tables 0 padded

greglucas:
The ESA tables provided have 0 padding for the ESA levels. i.e. 0A12 rather than just A12, so we need to specify that as 04X in the format string to get the 0 padding from the hex conversion.

We were converting l1 epoch values directly to datetime64, but the datetime64 epoch is different than the CDF epoch. We need to convert to timedelta64ns and then offset from the 2000-01-01 epoch CDF uses. I think we should use the spice time routines, but when I did that quickly I got missing clock kernels, so we'll need to add those to the datamanager still. That can be a follow-up or updated later. Putting this up for a hotfix reference point right now.

### DIFF
--- a/imap_processing/swapi/l2/swapi_l2.py
+++ b/imap_processing/swapi/l2/swapi_l2.py
@@ -220,13 +220,19 @@ def swapi_l2(
     # -----------------------------------
     # Convert unpacked ESA_LVL5 values to hex to match the LUT table
     # value
-    esa_lvl5_hex = np.vectorize(lambda x: format(x, "X"))(l1_dataset["esa_lvl5"].values)
+    esa_lvl5_hex = np.vectorize(lambda x: format(x, "04X"))(
+        l1_dataset["esa_lvl5"].values
+    )
+
+    # Turn the string start times into numpy datetime64
+    sci_start_time = l1_dataset["sci_start_time"].values.astype("datetime64[ns]")
+
     esa_energy = solve_full_sweep_energy(
         esa_lvl5_hex,
         l1_dataset["sweep_table"].data,
         esa_table_df=esa_table_df,
         lut_notes_df=lut_notes_df,
-        data_time=np.array(l1_dataset["epoch"].data, dtype="datetime64[ns]"),
+        data_time=sci_start_time,
     )
 
     l2_dataset["swp_esa_energy"] = xr.DataArray(


imap_processing/swapi/l2/swapi_l2.py
```diff
@@ -220,13 +220,22 @@ def swapi_l2(
...
+        l1_dataset["epoch"].data
+    ).astype("timedelta64[ns]")
+
     esa_energy = solve_full_sweep_energy(
         esa_lvl5_hex,
         l1_dataset["sweep_table"].data,
         esa_table_df=esa_table_df,
         lut_notes_df=lut_notes_df,
-        data_time=np.array(l1_dataset["epoch"].data, dtype="datetime64[ns]"),
+        data_time=data_time,
```
tech3371: can you pass in ?`sci_start_time` instead? This is already in datetime format from l1?

imap_processing/swapi/l2/swapi_l2.py
```diff
@@ -220,13 +220,22 @@ def swapi_l2(
     # -----------------------------------
     # Convert unpacked ESA_LVL5 values to hex to match the LUT table
     # value
-    esa_lvl5_hex = np.vectorize(lambda x: format(x, "X"))(l1_dataset["esa_lvl5"].values)
+    esa_lvl5_hex = np.vectorize(lambda x: format(x, "04X"))(
+        l1_dataset["esa_lvl5"].values
```
tech3371, line 224: Nice catch!

tech3371: Actually, this is not padded. I remember there is a reason for that first digits. Let me look up email on that.

greglucas: I thought the same thing, @tech3371. But if you look at the files, this definitely needed that. Maybe they changed their ancillary file naming since then?

tech3371: I see. I just looked through and this is a different issue. What you had is great!

imap_processing/swapi/l2/swapi_l2.py
tech3371: Can you make this minor change? I checked with Margaret and this will simply above changes and need to SPICE n L2.
```suggestion
        data_time=l1_dataset["sci_start_time"],
```

tech3371 (approved):

imap_processing/swapi/l2/swapi_l2.py
greglucas: OK, I pushed up something similar to this. But it did need a `.astype()` case because that variable is currently a string and you want it datetime64 in the receiving method. Take a look if that is what you were thinking.

tech3371: yep. That looks good to me!

PR #2400, opened 2025-11-10
MNT/FIX: Update Lo ancillary file naming from strt to start

greglucas:
# Change Summary

## Overview

Processing L1c failed because of the naming for the background ancillary file columns. It is unclear to me what we are expecting going forward, so just putting this up here for awareness right now.

What would you think about adding a condition into the code to allow for both cases @sdhoyt? I am not sure how I feel about that, but it would make our lives easier potentially too.

### DIFF
--- a/imap_processing/lo/l1c/lo_l1c.py
+++ b/imap_processing/lo/l1c/lo_l1c.py
@@ -561,19 +561,19 @@ def set_background_rates(
 
     # find to the rows for the current pointing
     pointing_bg_df = background_df[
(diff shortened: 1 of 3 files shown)


sdhoyt:
@greglucas I think we should change everything to use "start" and let them know. They changed most of those columns in the last sample ancillary they sent me except for the background and I guess I just went with that. But I think it makes sense to force it to use "start" instead of "strt"

sdhoyt (approved):

PR #2401, opened 2025-11-10
Lo L1C - Added epoch dim to background and direction in PSET

sdhoyt:
# Change Summary

## Overview
Added the epoch dimension to the beginning of the direction and background variables


greglucas (approved):
It looks like we need to update the L1C attributes/DEPENDS as well. Do you want to push those changes here or follow up in a separate PR?

greglucas (approved):

PR #2405, opened 2025-11-10
ULTRA l1c fine grained energy bins

lacoak21:
# Change Summary

## Overview
ULTRA requested new energy bins to use at l1c. At l1c the energy bins should be fine grained and at l2, a super set of those bins will be used. 

ULTRA IT instructions : [l1c_energy_bins_update.pptx](https://github.com/user-attachments/files/23463707/l1c_energy_bins_update.pptx)

## Updated Files
   - define new bin edges as a constant. 

## Testing
test that new bins match the powerpoint values


imap_processing/ultra/constants.py
```diff
@@ -88,31 +88,54 @@ class UltraConstants:
...
+        142.043527244462,
+        156.995477480721,
+        173.521317215533,
+        191.786719027695,
+        211.974794714821,
+        234.287931000591,
+        258.949818474338,
+        286.207694103215,
+        316.334819798291,
+        349.633221882321,
```
greglucas: 1e5?

I thought they wanted a high-energy catch-all bin. That would also give you 48 bins which seems more in line with a nice divisible number for grouping later.

tmplummer (commented):
LGTM. I will let Greg approve when the question about the upper edge gets resolved.

imap_processing/ultra/constants.py
lacoak21: Nick confirmed that they want the last bin to be 349. 

`We'll probably just want to throw everything out above that [349]; especially for hydrogen
some of the ions might make good measurements above that, but we're not making maps with those`

lacoak21: As for the grouping size. They do not want to group by a single integer. They would prefer to have the group configuration set up so you specify exactly how the binning should work. 

e.g. for the general grouping by 4, the config would be something like: 

`group_sizes = [4,4,4,4,4,4,4,4,4,4,4, 3]`

sum(group_sizes) must equal 47.

imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
```diff
@@ -105,8 +105,8 @@ def test_get_spacecraft_histogram(test_data):
     assert latitude.shape == (n_pix,)
     assert longitude.shape == (n_pix,)
 
-    # Spot check that 1 count is in the first energy bin
-    assert np.sum(hist[1, :]) == 2
+    # Spot check that 0 counts are in the first energy bin
```
greglucas: We should probably check a bin with values here.

greglucas (approved):
Looks like the tests are still failing and I have one minor suggestion on making a test a little more robust, but looks good to me too.

imap_processing/ultra/constants.py
greglucas: I was thinking a bit about this. On another project we use `ds.coarsen()` on a regular grid to shrink a dataset down for visualization. I think that might be able to be used here, but possibly not as precise. Then I found `groupby_bins()` which seems ideal for this case actually. You'd receive the bin edges that they want to use as a single list and xarray would take care of the aggregation under the hood of each pset dataset.

```python
pset.groupby_bins('energy', bins=energy_bins).sum('energy')
```

Just commenting here because I was curious and looking this up, but don't implement this yet, keep it as a separate PR.

lacoak21: Oh nice groupby_bins seems perfect for this.

lacoak21:
> Looks like the tests are still failing and I have one minor suggestion on making a test a little more robust, but looks good to me too.

Yeah I cant get them to pass. It keeps failing on that one test and it runs in 3 seconds locally.

PR #2409, opened 2025-11-11
MNT: Turn down cdflib log messages while writing L1 CDFs

### LINKED ISSUE #2408: Ignore cdflib istp logger messages when writing L1 files
The L1 processing logs are often polluted with cdflib ISTP compliance warnings for L1 data products. We should turn down the warnings when processing L1 files. Unfortunately these are loggers and not userwarnings so we can't suppress them with warningsfilters, but have to update the cdflib logger.

greglucas:
There are a lot of messages in the stream about ISTP compliance that aren't relevant if the L1 files don't need to be ISTP compliant. This can hide useful information, so lets suppress the log messages for L1 files by default.

closes #2408

### DIFF
--- a/imap_processing/cdf/utils.py
+++ b/imap_processing/cdf/utils.py
@@ -11,6 +11,7 @@ from pathlib import Path
 import imap_data_access
 import numpy as np
 import xarray as xr
+from cdflib.logging import logger as cdflib_logger
 from cdflib.xarray import cdf_to_xarray, xarray_to_cdf
 from cdflib.xarray.cdf_to_xarray import ISTP_TO_XARRAY_ATTRS
 
@@ -166,7 +167,17 @@ def write_cdf(
     if "compression" not in extra_cdf_kwargs:
         extra_cdf_kwargs["compression"] = 6  # type: ignore
 
-    xarray_to_cdf(dataset, str(file_path), **extra_cdf_kwargs)
+    prev_cdflib_level = cdflib_logger.level
+    try:
+        if not extra_cdf_kwargs.get("istp", False):
+            # Suppress cdflib logging messages for data levels that don't need
+            # strict ISTP compliance
+            logger.info("Disabling cdflib ISTP logging for level 1 data products")
+            cdflib_logger.setLevel(logging.ERROR)
+        xarray_to_cdf(dataset, str(file_path), **extra_cdf_kwargs)
+    finally:
+        # Set back to the previous logging level
+        cdflib_logger.setLevel(prev_cdflib_level)
     return file_path
 
 


imap_processing/cdf/utils.py
```diff
@@ -166,7 +167,17 @@ def write_cdf(
     if "compression" not in extra_cdf_kwargs:
         extra_cdf_kwargs["compression"] = 6  # type: ignore
 
-    xarray_to_cdf(dataset, str(file_path), **extra_cdf_kwargs)
+    prev_cdflib_level = cdflib_logger.level
```
tech3371, line 170: can you rename to something like this?
```suggestion
    cdflib_log_level = cdflib_logger.level
```

imap_processing/cdf/utils.py
```diff
@@ -166,7 +167,17 @@ def write_cdf(
...
+    try:
+        if not extra_cdf_kwargs.get("istp", False):
```
tech3371, line 172: what about terminate_on_warnings logs?

imap_processing/cdf/utils.py
```diff
@@ -166,7 +167,17 @@ def write_cdf(
...
+            # Suppress cdflib logging messages for data levels that don't need
+            # strict ISTP compliance
+            logger.info("Disabling cdflib ISTP logging for level 1 data products")
+            cdflib_logger.setLevel(logging.ERROR)
+        xarray_to_cdf(dataset, str(file_path), **extra_cdf_kwargs)
+    finally:
+        # Set back to the previous logging level
+        cdflib_logger.setLevel(prev_cdflib_level)
```
tech3371, line 180: I don't know enough here. Does finally filters logs based on level at the end? For example, if write didn't fail, it will go into finally and then filter based on level?

tech3371 (approved):
Minor comments and questions

imap_processing/cdf/utils.py
greglucas: I want to keep the "prev" in there because the point of this is to keep the old version around so we can reset it after the function call later.

imap_processing/cdf/utils.py
greglucas: I am just trying to minimize the logs and this is the line that causes log output. We aren't changing any other features or modifying anything being sent to the functions only turning the logs on/off based on this flag.

imap_processing/cdf/utils.py
greglucas: The finally is always executed, so this is a way to reset the log level even if there was an exception within xarray_to_cdf. This is probably all unnecessary, but a nice-to-do thing so that if someone does want to get logs from cdflib after this section we don't poison their session.

PR #2411, opened 2025-11-11
2374 lo l2 part3  integrate cg correction for hf maps

### LINKED ISSUE #2374: Lo L2 - Integrate CG correction for hf maps
### Algorithm Description:

Integrate CG correction with Lo L2 processing.

tmplummer:
# Change Summary

## Overview
Sorry for another rather large PR. I would have liked to break this up but am trying to get the Lo L2 stuff out the door.

This adds ram/anti filtering and CG correction to the Lo L2 pipeline. See the Updated Files section for more details.

## Updated Files
   - Unrelated - add `with np.errstate():` to remove invalid divide in logs.
   - Add code to correctly calculate the Pointing midpoint time for Lo.
   - Update `interpolate_map_flux_to_helio_frame` method to be able to interpolate multiple sets of intensity/uncertainty variables independently. Lo needs the ability to interpolate both "ena_intensity" and "bg_intensity".
- imap_processing/hi/hi_l2.py
   - Update call to `interpolate_map_flux_to_helio_frame`.
   - Move `_prepare_corrections` earlier in order to reduce duplicate identification of `cg_correction` bool.
   - Add CG correction to`process_single_pset`
   - Add direction_mask to `project_pset_to_map` for ram/anti filtering
   - Set correct energy values to energy coordinate in `normalize_pset_coordinates` (needed for CG correction).
   - Add `reduce_geometric_factor_dataset` method that returns a dataset of reduced geometric factor data for only the species and esa_mode.
   - Add interpolation to helio-frame energies to `calculate_all_rates_and_intensities`
   - Add Lo specific testing
   - Add coverage for interpolating multiple intensity variables in `interpolate_map_flux_to_helio_frame`.
   - Minimal update to mock function due to signature change of `interpolate_map_flux_to_helio_frame`
- imap_processing/tests/lo/test_lo_l2.py
   - Add test coverage for new and updated lo_l2.py functions

Closes: #2374

### DIFF
--- a/imap_processing/ena_maps/utils/corrections.py
+++ b/imap_processing/ena_maps/utils/corrections.py
@@ -731,6 +731,10 @@ def interpolate_map_flux_to_helio_frame(
     energy_right = esa_energies_ev.isel({"energy": right_idx_da})
 
     for var_name in vars_to_interpolate:
+        logger.debug(
+            f"Interpolating {var_name}, {var_name}_stat_uncert, and"
+            f"{var_name}_sys_err to heliocentric frame energies"
+        )
         # Step 2: Extract flux values at bounding energy channels
         # Use xarray's advanced indexing to get fluxes at left and right indices
         intensity = map_ds[var_name]
--- a/imap_processing/ena_maps/utils/corrections.py
+++ b/imap_processing/ena_maps/utils/corrections.py
@@ -749,7 +749,7 @@ def interpolate_map_flux_to_helio_frame(
 
     for var_name in vars_to_interpolate:
         logger.debug(
-            f"Interpolating {var_name}, {var_name}_stat_uncert, and"
+            f"Interpolating {var_name}, {var_name}_stat_uncert, and "
             f"{var_name}_sys_err to heliocentric frame energies"
         )
         # Step 2: Extract flux values at bounding energy channels
--- a/imap_processing/ena_maps/utils/corrections.py
+++ b/imap_processing/ena_maps/utils/corrections.py
@@ -273,7 +273,8 @@ class PowerLawFluxCorrector:
             source_uncertainties = observed_uncertainties / eta_final
 
             # Check convergence
-            ratios_sq = (source_fluxes_n / source_fluxes_prev) ** 2
+            with np.errstate(divide="ignore", invalid="ignore"):
+                ratios_sq = (source_fluxes_n / source_fluxes_prev) ** 2
             chi_n = np.sqrt(np.mean(ratios_sq)) - 1
 
             if chi_n < convergence_threshold:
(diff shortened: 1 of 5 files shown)


imap_processing/ena_maps/utils/corrections.py
```diff
@@ -273,7 +273,8 @@ def predictor_corrector_iteration(
             source_uncertainties = observed_uncertainties / eta_final
 
             # Check convergence
-            ratios_sq = (source_fluxes_n / source_fluxes_prev) ** 2
+            with np.errstate(divide="ignore", invalid="ignore"):
```
greglucas, line 276: Much appreciated! Thank you ❤️

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -344,10 +345,26 @@ def add_spacecraft_velocity_to_pset(
     - "sc_velocity": Spacecraft velocity vector (km/s) with dims ["x_y_z"]
     - "sc_direction_vector": Spacecraft velocity unit vector with dims ["x_y_z"]
     """
-    # Compute ephemeris time (J2000 seconds) of PSET midpoint
-    # epoch contains Pointing start time, and epoch_delta indicates the total
-    # duration of the Pointing
-    et = ttj2000ns_to_et(pset["epoch"].values[0] + pset["epoch_delta"].values[0] / 2)
+    # Hi and Lo need to use different methods for computing the Pointing
+    # midpoint time.
```
greglucas, line 349: Is this something we can add to the Lo pointing set to make it align with Hi? It seems like these should be identical between the instruments to me.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -65,36 +72,40 @@ def lo_l2(
...
+    sky_map = create_sky_map_from_psets(
+        psets, map_descriptor, efficiency_data, cg_correction
+    )
+
+    logger.info("Step 3: Converting to dataset and adding geometric factors")
+    dataset = sky_map.to_dataset()
+    dataset = add_geometric_factors(dataset, map_descriptor.species)
+
+    logger.info("Step 4: Calculating rates and intensities")
+
```
greglucas: For consistency with the others without a blank line.
```suggestion
```

imap_processing/lo/l2/lo_l2.py
```diff
@@ -340,6 +372,27 @@ def process_single_pset(
...
+        # Convert energy coordinate from keV to eV for CG correction
+        # (energy coordinate was set in normalize_pset_coordinates in keV)
+        energy_values_ev: xr.DataArray = pset_processed["energy"] * 1000.0
+        # ram_mask variable is added as part of apply_compton_getting_correction
+        pset_processed = apply_compton_getting_correction(
+            pset_processed, energy_values_ev
+        )
+    else:
+        pset_processed = add_spacecraft_velocity_to_pset(pset_processed)
+        pset_processed = calculate_ram_mask(pset_processed)
```
greglucas, line 395: We don't need ram/anti-ram calculations on the non-HF maps?

Similarly, you don't need the spacecraft velocity for the CG correction? For some reason I thought we went through all corrections before the CG correction, which would mean your CG-pset would need these.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -359,16 +412,23 @@ def normalize_pset_coordinates(pset: xr.Dataset, species: str) -> xr.Dataset:
...
 
     # Drop the esa_energy_step coordinate first to avoid conflicts
     pset_renamed = pset_renamed.drop_vars("esa_energy_step")
 
-    # Ensure the pset energy coordinates match the output map
-    # TODO: Do we even need this if we are assigning the true
-    #       energy levels later?
-    pset_renamed = pset_renamed.assign_coords(energy=range(7))
+    # Assign TRUE energy values as coordinates (in keV, matching map convention)
+    pset_renamed = pset_renamed.assign_coords(energy=gf_dataset["Cntr_E"].values)
```
greglucas, line 432: Nice :)

imap_processing/lo/l2/lo_l2.py
```diff
@@ -587,6 +655,45 @@ def load_geometric_factor_data(species: str) -> pd.DataFrame:
...
+    # Filter for the specific ESA mode
+    if "esa_mode" in gf_data.columns:
+        gf_data = gf_data[gf_data["esa_mode"] == esa_mode].copy()
+
+    # Convert to xarray Dataset indexed by energy step for vectorized selection
+    gf_ds = gf_data.set_index("Observed_E-Step").to_xarray()
+
+    # TODO: The ancillary data has repeated Observed_E-Step values with different
+    #    incident_E-Step values. Figure out how we get the correct row. For now,
+    #    just remove duplicate Observed_E-Step values.
```
greglucas: They said that we should only use `observed == incident` so (1, 1), (2, 2), etc. Does this do that or does it keep the lowest pair instead so (1, 1), (1, 2), (1, 3), etc.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -709,6 +805,7 @@ def calculate_all_rates_and_intensities(
     flux_correction: bool = False,
     o_map_dataset: xr.Dataset | None = None,
     flux_factors: Path | None = None,
+    cg_correction: bool = False,
```
greglucas, line 808: You are calling this "cg_correct" in other functions, I'd suggest sticking to one or the other throughout if we can.

It does look like we are calling it "flux_correction" above, but I'd be fine changing that to "flux_correct" if that is easier to align with hi as well.

greglucas (approved):
Nice job feeding all of that through the Lo code. Thank you for doing a bunch of cleanup as you went along too, it reads cleaner with your updates.

imap_processing/lo/l2/lo_l2.py
tmplummer: Those functions are hidden in the CG correction algorithm. I will pull them out to make it more obvious.

imap_processing/ena_maps/utils/corrections.py
tmplummer: I suppose that we could potentially just add the "epoch_delta" variable to the Lo L1C. I have no idea if the Lo IT specifically requested the "pointing_start_met" and "pointing_end_met" variables. Maybe @sdhoyt would know or have an opinion on this?

PR #2412, opened 2025-11-11
CoDICE: L1A DE refactor

### LINKED ISSUE #2260: CoDICE: Lo/Hi L1A Direct Event re-validation
Fix Lo and Hi's direct events in this ticket

tech3371:
# Change Summary
closes #2260 

## Overview

L1A DE refactor

## Updated Files

- imap_processing/codice/codice_l1a_de.py
   - main changes for this PR


imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,464 @@
...
+    for field, num_bits in constants.DE_METADATA_FIELDS.items():
+        metadata_str += f"{packets[field].data[packet_index]:0{num_bits}b}"
+    metadata_chunks = [metadata_str[i : i + 8] for i in range(0, len(metadata_str), 8)]
+    metadata_ints = [int(item, 2) for item in metadata_chunks]
+    metadata = bytes(metadata_ints)
+
+    return metadata
+
+
+def group_data(packets: xr.Dataset) -> list[bytes]:
```
tech3371, line 50: This and above functions may get replaced in the future with space_packet_parser's functionality

imap_processing/tests/codice/test_codice_l1a.py
```diff
@@ -533,55 +470,107 @@ def test_hi_priority():
...
+            # Do string comparison for priority_label
+            assert np.array_equal(
+                processed_data[variable].values, val_data[variable].values
+            ), f"Mismatch in variable '{variable}'"
+            continue
 
-    cdf_file = write_cdf(processed_data)
-    assert cdf_file.name == "imap_codice_l1a_lo-direct-events_20250814_v999.cdf"
+        # Joey's has fillval of 65535 for uint16 for these variables.
+        # Others are uint8 fillval.
```
greglucas: Do we want this too? Initialize the arrays as `np.full(shape, fill_value, dtype)` which I think it might have been before. It seems reasonable that we might want the max rather than 0 here if 0 could be a potentially valid value.

imap_processing/codice/codice_l1a_de.py
greglucas: Yeah, it would be nice if we could get this to work, but I haven't had time to look into it more. Unfortunately, I think it is a little trickier with the packing design (checksums at the end of every packet) than I was hoping for.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,468 @@
...
+    #                    Bit 6 → 0
+    #                    Bit 5 → 0
+    #                    Bit 4 → 0
+    #                    Bit 3 → 0
+    #                    Bit 2 → 0
+    #                    Bit 1 → 1
+    #                    Bit 0 (LSB) → 1
+    #   bits chunks - [5, 1, ...., 7, 3, 16]
+    #   vars - ['gain', 'apd_id', ...., 'energy_step', 'priority', 'spare']
+    #   unpack data - [3, 0, 0, ....., 0, 0]
```
greglucas, line 149: Interesting, this is backwards of many other versions where the "spare" are the 0s on the right-hand-side so it is aligned with your array. Here your array is kind of backwards in a sense. I guess you could iterate through your inputs in reverse order and start at the far left edge if that woudl make anything easier too. But what you have seems OK, just slightly confusing at first glance because I'm not used to it.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,468 @@
...
+        | (packets.seq_flgs.data == SegmentedPacketOrder.FIRST_SEGMENT)
+    )[0]
+    num_events_arr = packets.num_events.data[packet_index_starts]
+    data_quality_arr = packets.suspect.data[packet_index_starts]
+    priorities_arr = packets.priority.data[packet_index_starts]
+
+    # Initialize other fields of l1a that we want to
+    # carry in L1A CDF file
+    de_data["num_events"] = xr.DataArray(
+        np.zeros((num_epochs, num_priorities), dtype=int),
```
greglucas: I think you want this to be an unsigned int and likely smaller in size than a full 64bit int. I commented on one of the tests, but do you want to keep this initialized to the maximum values as well to align with Joey's data?
```suggestion
        np.full((num_epochs, num_priorities), 65535, dtype=np.uint16),
```

imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,468 @@
...
+        epoch_start = packet_index_starts[epoch_index] * num_priorities
+        epoch_end = packet_index_starts[epoch_index + 1] * num_priorities
+        # Extract the decompressed data for current epoch.
+        # epoch_data should be of shape ((num_priorities * num_events),)
+        epoch_data = decompressed_data[epoch_start:epoch_end]
+
+        # Extract these other data
+        unordered_priority = priorities_arr[epoch_start:epoch_end]
+        unordered_data_quality = data_quality_arr[epoch_start:epoch_end]
+        un_ordered_num_events = num_events_arr[epoch_start:epoch_end]
```
greglucas:
```suggestion
        unordered_num_events = num_events_arr[epoch_start:epoch_end]
```

imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,468 @@
...
+        #   current group's start index * 8 to next group's start indices * 8
+        epoch_start = packet_index_starts[epoch_index] * num_priorities
+        epoch_end = packet_index_starts[epoch_index + 1] * num_priorities
+        # Extract the decompressed data for current epoch.
+        # epoch_data should be of shape ((num_priorities * num_events),)
+        epoch_data = decompressed_data[epoch_start:epoch_end]
+
+        # Extract these other data
+        unordered_priority = priorities_arr[epoch_start:epoch_end]
+        unordered_data_quality = data_quality_arr[epoch_start:epoch_end]
```
greglucas, line 268: For some of this I don't know if it would help, but I think you might be able to do a `small_ds = ds.isel(epoch=slice(epoch_start, epoch_end))` and that would slice the entire dataset at once `small_ds["priorities"]`. That might be slower though too and make lots of views, so only if you think that would help with the logic / flow to stay in xarray indexing.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,468 @@
...
+            priority_num_events = int(un_ordered_num_events[priority_index])
+            # Reshape epoch data into (num_events, 8). That 8 is 8-bytes that
+            # make up 64-bits. Therefore, combine last 8 dimension into one to
+            # get 64-bits event data that we need to unpack later.
+            events_in_bytes = np.array(epoch_data[priority_index]).reshape(
+                priority_num_events, 8
+            )
+
+            # combine last 8 dimension into one 64-bits value
+            combined_64bits = np.dot(events_in_bytes, 1 << (8 * np.arange(8)[::-1]))
```
greglucas: Can you do a view of the uint8 data instead so you don't need to create a new array?
`event_in_bytes.view(np.uint64)`

imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,468 @@
...
+            # Put event data into their respective priority number bins
+            priority_num = int(unordered_priority[priority_index])
+            de_data["64bits_event_data"][
+                epoch_index, priority_num, :priority_num_events
+            ] = combined_64bits
+
+    # Now unpack the 64-bits event data using bitwise operations
+    unpacked_vars = list(bit_structure.keys())
+    unpacked_bits = [bit_structure[field]["bit_length"] for field in bit_structure]
```
greglucas: Combine these into a single mapping {field: num_bits} that you can pass in as one value and iterate over within the unpack_bits function.
```suggestion
    unpacked_bits = {field: bit_structure[field]["bit_length"] for field in bit_structure}
```

imap_processing/tests/codice/test_codice_l1a.py
tech3371: I thought about that. Looks like I am using common CDF attrs for direct event data(s). So, if I use fillval for uint8 and uint6 separately but common CDF fillval is -9223372036854775808, will they conflict? If not, then I can do that. This is one I was not sure on.

imap_processing/codice/codice_l1a_de.py
tech3371: yep. It took me a while to grasp but I matched with Joey. They are confident with their output and that helped me feel good with this implementation.

imap_processing/codice/codice_l1a_de.py
tech3371: same as above fillval comment.

imap_processing/codice/codice_l1a_de.py
tech3371: There are small things that are gotcha in DE implementation that I had to revert to Matthew's code. For example, we had to filter segment start or unsegment before I this. And epoch data being variable length. That caused lot of things such as it doesn't allow data to convert to np.array type. I haven't tried xr though. I can look into that.

imap_processing/tests/codice/test_codice_l1a.py
greglucas: You'd have to update the CDF attributes then I think. I think that might have been what Matthew was doing dynamically by storing it in the old dictionary items where you removed the dtype/fillval entries. A negative fillval for an unsigned integer doesn't seem like it should be valid.

imap_processing/codice/codice_l1a_de.py
tech3371: I don't know what view is. I can look into it

imap_processing/tests/codice/test_codice_l1a.py
tech3371: SPDF suggested fillval is -1.0000000E+31.

imap_processing/codice/codice_l1a_de.py
tech3371: cool. sliced worked

imap_processing/codice/codice_l1a_de.py
tech3371: nice. view is really cool. I didn't know that's a thing

tech3371: It still gave me of shape (23, 8) ) when I want (23,)

tech3371: As I look more closely, np.dot might be a good approach. I makes sense to me.

imap_processing/tests/codice/test_codice_l1a.py
greglucas: That is for floats I think. For ints it has to be different and a valid value within the integer range.

imap_processing/codice/codice_l1a_de.py
greglucas: The dot is more confusing to me with all of the bitshifts and a range being multiplied by 8 which is then shifted.

This works as I was expecting, so I think you'd have to index into the 0th element or squeeze it if you wanted to remove that last singleton dimension.

```python
x = np.ones((22, 8), dtype=np.uint8)
x.view(np.uint64)  # shape (22, 1)
```

tech3371: I can't get view to work with mine :(
```
>>> x
array([1, 2])
>>> x.view(np.uint8)
array([1, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0], dtype=uint8)
```

imap_processing/tests/codice/test_codice_l1a.py
tech3371: I will try to revert the change

tech3371: I addressed this in latest push. Looks like what Matthew did originally was good. DE has many potholes that we can't much here.

tech3371:
@greglucas I addressed all comments besides view.

imap_processing/codice/codice_l1a_de.py
tech3371: The issue was that I was initializing 8-element as int64, because of that view didn't combine. We need to initialize array as dtype=uint8. Then it worked. Thanks Greg for helping with this!

imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,480 @@
...
+            )
+
+    # Get num_events, data quality, and priorities data for beginning of packet_indexs
+    packet_index_starts = np.where(
+        (packets.seq_flgs.data == SegmentedPacketOrder.UNSEGMENTED)
+        | (packets.seq_flgs.data == SegmentedPacketOrder.FIRST_SEGMENT)
+    )[0]
+    num_events_arr = packets.num_events.isel(epoch=packet_index_starts)
+    data_quality_arr = packets.suspect.isel(epoch=packet_index_starts)
+    priorities_arr = packets.priority.isel(epoch=packet_index_starts)
```
greglucas: You should be able to do one isel above this I think and then access these later.

```python
packets_subset = packets.isel(epoch=packet_index_starts)
```

Then don't take out the arrays into 3 separate variables, but rather within the loop do another subselection.

```python
second_packet_subset = packet_subset.isel(epoch=slice(start, end))

priorities = second_packet_subset["priority"]
```

But, I know that recently we had to avoid the `isel()` within a loop on Lo and pull out the arrays sooner, so this might be a bad suggestion here and we should go back to the original version you had :-/ 

So, my two suggestions for ideas going forward:
* Go back to pulling out a raw array as early as possible and do normal numpy indexing/slicing (I think what you were doing originally!)
* Avoid the 3 isel() calls and do only one on the dataset itself and then get the 3 arrays by name after that.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -0,0 +1,480 @@
...
+            # make up 64-bits. Therefore, combine last 8 dimension into one to
+            # get 64-bits event data that we need to unpack later.
+            events_in_bytes = np.array(epoch_data[i], dtype=np.uint8).reshape(
+                priority_num_events, 8
+            )
+            # combine last 8 dimension into one 64-bits value
+            #   we need to make a copy and reverse the byte order
+            #   to match LSB order before we use .view.
+            combined_64bits = np.array(events_in_bytes[:, ::-1])
+            combined_64bits = combined_64bits.view(np.uint64).flatten()
```
greglucas: * Can you move the [:, ::-1] up into the reshape command somehow above I wonder? Then we wouldn't need to reshape it again here or could get just one copy.

* The .flatten() creates another copy, we should avoid that one by indexing into the array if we can pull out that individual column because we know it is a one-entry array: `[:, 0]`

greglucas (approved):
Looks good to me. One minor comment where I think my previous comment may have been a bad one, so sorry for that suggestion!

imap_processing/codice/codice_l1a_de.py
tech3371: so go back to this?
```
    num_events_arr = packets.num_events.data[packet_index_starts]
    data_quality_arr = packets.suspect.data[packet_index_starts]
    priorities_arr = packets.priority.data[packet_index_starts]
```

greglucas: Yeah, I think so, at the very top of the file outside of the loops. Then the indexing inside the loop is faster.

imap_processing/codice/codice_l1a_de.py
tech3371: - view is quite picky. I moved reverse right after reshape. reverse needs to happen after reshape. Either way, looks like we need to make two copies for view.
- we need to flatten for unpacking bits. but `[:,0]` works

PR #2415, opened 2025-11-11
Lo L1A - Refactor combin_segemented_packets to handle MET

sdhoyt:
# Change Summary

## Overview

- Moved check to see if combining packets is need to inside combine_segmented_packets so MET can still be set in the case that there are no segmented packets
- Added logging per packet for parsing DEs

### DIFF
--- a/imap_processing/lo/l0/lo_science.py
+++ b/imap_processing/lo/l0/lo_science.py
@@ -358,12 +358,15 @@ def extract_bits_from_bytes(
 
 def combine_segmented_packets(dataset: xr.Dataset) -> xr.Dataset:
     """
-    Combine segmented packets.
+    Combine segmented packets and set MET field.
 
     If the number of bits needed to pack the direct events exceeds the
     maximum number of bits allowed in a packet, the direct events
     will be spread across multiple packets. This function will combine
     the segmented binary into a single binary string for each epoch.
+
+    This function also sets the MET field based on segment start times,
+    even when no segmentation is present.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


greglucas (approved):
Nice job tracking this down and sorry for the regression!

I guess one other potential option would have been to remove the "if needs_combine" check altogether and always go down that path? I'm not sure if that would work, but it might make the code simpler in the end.

greglucas:
> Added logging per packet for parsing DEs

Is that going to be a lot of logs in the output now? I'm wondering if we want these logged every time or if it should be a debug-level log so you can get it if you want it by debugging locally, but not necessarily in every data file we are producing.

sdhoyt:
> I guess one other potential option would have been to remove the "if needs_combine" check altogether and always go down that path? I'm not sure if that would work, but it might make the code simpler in the end.

@greglucas It does work if all of them are unsegmented. The seq flag of 3 is considered a start of a segment, so it would take all of the packets as a start seq, see that they're all sequential and set everything correctly. I figured we wanted to skip this in the name of speeding things up so I left the check in there. Not sure how much of a difference it will make though. If you don't think it saves us much, I'll remove the check

sdhoyt:
> > Added logging per packet for parsing DEs
> 
> Is that going to be a lot of logs in the output now? I'm wondering if we want these logged every time or if it should be a debug-level log so you can get it if you want it by debugging locally, but not necessarily in every data file we are producing.

@greglucas We had a little over 200 packets with the data I've seen so far which I didn't think was too many. I thought it could be helpful to see the number of events in each packet in the logs even if there isn't there isn't an issue. But I can switch this to debug if you think that would be better

greglucas:
I'm fine either way. I don't think it would be a large bottleneck is my untested assumption. You can probably run a file locally quick and check if it is much slower or not.

PR #2420, opened 2025-11-13
FIX: The I-ALiRT sc declination is an unsigned integer

greglucas:
Also, use the spacecraft conversion values directly so we are aligned with what the s/c is measuring onboard.

This is from an e-mail from the MOC on 11/13/2025.


laspsandoval (approved):
Great. Thanks!

PR #2422, opened 2025-11-13
BUG: CG corrected efficiency 

lacoak21:
# Change Summary

## Overview
Major bug found in CG corrected psets. Im hoping this was the final issue with getting those validated 🤞 . 

## Updated Files
   - the CG corrected `_efficiency`var was unused. It should be overwriting the `efficiencies` value. 
   - That means the efficiencies used in the sensitivity calculation and therefore the flux calculation was not cg corrected


greglucas:
Can you assert against a value in your sensitivities for a quick test since it looks like that is the value what wouldn't have been changed previously?

greglucas (approved):

PR #2437, opened 2025-11-17
I-ALiRT - MAG frame transformations

laspsandoval:
# Change Summary

## Overview
This is the correction to the frame transformation. The transformation matches the "gold standard" using SPICE kernels. It does appear that there is still scattering after the transformation which is likely attributed to the input data from l1b; not the frame transformation after these corrections.

## Updated Files
- ialirt_spice.py
   - Found bugs in frame transformation code and also adjusted the spin phase to match SPICE
- parse_mag.py
   - Moved spherical interpolation to its own function so I could test

## Notes
**Kernels used for gold standard:**
    kernels = [
        "imap_science_100.tf",
        "imap_120.tf",
        "naif0012.tls",
        "de440s.bsp",
        "imap_recon_od005_20250925_20251014_v01.bsp",
        "pck00011.tpc",
        "imap_sclk_0036.tsc",
        "imap_2025_283_2025_284_001.ah.bc",
    ]

**Equations for gold standard**
```
        mago_inertial_vector_gold = frame_transform(
            ttj2000ns_to_et(time_data["primary_epoch"]),
            mago_out,
            from_frame=SpiceFrame.IMAP_MAG_O,
            to_frame=SpiceFrame.ECLIPJ2000,
        )
```
Did above w/ magi equivalent

```
        mago_gse_vector_gold = frame_transform(
            ttj2000ns_to_et(time_data["primary_epoch"]),
            mago_out,
            from_frame=SpiceFrame.IMAP_MAG_O,
            to_frame=SpiceFrame.IMAP_GSE,
        )
```

## Testing
- test_parse_mag.py
   - Added testing for interpolation
- test_ialirt_spice.py
   - Added testing for frame transformation

- Added kernels: imap_120.tf, imap_sclk_0036.tsc

[my_mago_output.xls](https://github.com/user-attachments/files/23593115/my_mago_output.xls)
[my_magi_output.xls](https://github.com/user-attachments/files/23593185/my_magi_output.xls)
[mago_vectors_gse.xlsx](https://github.com/user-attachments/files/23593591/mago_vectors_gse.xlsx)

mago
Screenshot 2025-11-17 at 3 30 39 PM
magi
Screenshot 2025-11-17 at 3 36 44 PM

mago gse (note that the gradiometer offset was applied to the inertial frame prior to the gse transformation which is why the gold standard is not exactly the same as the mago_gse calculated value)
Screenshot 2025-11-17 at 4 17 00 PM


imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -9,6 +9,10 @@
     spherical_to_cartesian,
 )
 
+# Derived empirically from comparing the gold standard (SPICE) output vector
+# and calculated vector.
+SPIN_PHASE_OFFSET_DEG = -24.8
```
tmplummer, line 14: This is a really big red flag for me. Is there an explanation for this offset?

imap_processing/ialirt/l0/ialirt_spice.py
```diff
@@ -123,7 +127,7 @@ def compute_total_rotation(
     total_rotations : NDArray
         Instrument to inertial rotation matrices (N, 3, 3).
     """
-    total_rotations = mount_matrix @ spin_rotations @ inertial_frames
+    total_rotations = inertial_frames @ spin_rotations @ mount_matrix
```
tmplummer, line 130: This change makes sense to me since combining transform matrices goes from right to left. i.e. this transform now takes vectors in the MAG frame, transforms to the S/C frame, then to a despun S/C frame and finally to the inertial frame.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -387,6 +387,73 @@ def apply_gradiometry_correction(
...
+            np.ones_like(sc_inertial_right),
+            sc_inertial_right,
+            sc_inertial_decline,
+        ],
+        axis=-1,
+    )
+    vecs = spherical_to_cartesian(spherical_coords)
+
+    # Interpolate in Cartesian space
+    vx = np.interp(target_time, attitude_time, vecs[:, 0])
```
tmplummer: I'm concerned about using linear interpolation here. I don't know off the top of my head what type of interpolation to use, but we want something that will fit a curve to several of the points since the spin axis wobbles around roughly tracing a circle. We could probably look at what SPICE uses... perhaps b-splines?

Additionally, it would probably be good to check that target_time is contained within the range `(attitude_time[0], attitude_time[-1])`. Otherwise, `np.interp` is just going to return the start/end attitude vector without telling you it couldn't actually perform the interpolation.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -387,6 +387,73 @@ def apply_gradiometry_correction(
...
+    v_interp /= np.linalg.norm(v_interp)
+
+    # Convert back to spherical
+    ra_dec = cartesian_to_spherical(v_interp)
+    ra_deg = ra_dec[1]
+    dec_deg = ra_dec[2]
+
+    # Account for discontinuities in spin phase.
+    spin_phase_unwrapped = np.unwrap(np.radians(sc_spin_phase))
+    spin_phase_interp = np.interp(target_time, attitude_time, spin_phase_unwrapped)
```
tmplummer, line 460: Linear interpolation should be appropriate for determining spin phase.

imap_processing/tests/ialirt/unit/test_ialirt_spice.py
```diff
@@ -60,40 +54,22 @@ def test_get_rotation_matrix():
...
-    expected = np.array(
-        [
-            [1.0, 0.0, 0.0],  # Rotating around X leaves X unchanged
-            [0.0, 0.0, -1.0],  # Rotating around Y sends X → -Z
-            [0.0, 1.0, 0.0],  # Rotating around Z sends X → Y
-        ]
-    )
+    # X → Y
+    # Y → -X
+    # Z stays the same
```
tmplummer: I think this can be updated to just say "Rotating a unit vector pointing along x-axis 90-degrees about z-axis results in a unit vector pointing along y-axis".

imap_processing/tests/ialirt/unit/test_parse_mag.py
```diff
@@ -526,29 +575,82 @@ def test_transform_to_frames(furnish_kernels, spice_test_data_path):
...
+        sc_inertial_decline,
+        sc_spin_phase,
+        attitude_time,
+        target_time,
+    )
+
+    expected_ra = np.interp(target_time, attitude_time, sc_inertial_right)
+
+    assert ra_deg == expected_ra
+    assert spin_phase_deg == 335
```
tmplummer: Please add a comment as to why this is the expected value. It is not obvious.

imap_processing/tests/ialirt/unit/test_parse_mag.py
```diff
@@ -526,29 +575,82 @@ def test_transform_to_frames(furnish_kernels, spice_test_data_path):
     np.testing.assert_allclose(rtn_vector, expected_rtn, atol=1e-05)
 
 
+def test_interpolate_spherical():
+    """Test the interpolate_spherical function."""
+
+    attitude_time = np.array([2, 3])
+    # Can be 0 -> 360 degrees
+    sc_inertial_right = np.array([0, 180])
```
tmplummer: This is a bit of a degenerate case to test b/c 90 and 270 are both valid answers. I'm actually quite surprised that this works. I would expect the interpolation of the Cartesian vectors to result in a [0, 0, 0] vector.

imap_processing/ialirt/l0/ialirt_spice.py
greglucas, line 14: This is suspiciously close to 1/15 to me (24.8 / 360). Are we off by one second in the timestamping of the vectors? i.e. we are using one packet too early or too late for the timestamp.

This phase offset seems like a red flag to me and we shouldn't be setting this as a factor.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -387,6 +387,73 @@ def apply_gradiometry_correction(
...
+    attitude_time: np.ndarray,
+    target_time: float,
+) -> tuple:
+    """
+    Interpolate spherical coordinates.
+
+    Parameters
+    ----------
+    sc_inertial_right : numpy.ndarray
+        Inertial right ascension for 4 packets 0 to 360 degrees, shape (4).
```
greglucas, line 404: I think I must be behind on this, but I thought that the magnetic field vectors were taken at a specific snapshot in time and then transmitted down over 4 packets, so we wouldn't want to do any interpolation/combination, but rather just shifting. The values from packet 3 should be moved to packet 1 because that is when that measurement was made? Is there an I-ALIRT MAG algorithm document we should be looking at to review this?

tmplummer (commented):
A few comments. I am still pondering some of this.

imap_processing/ialirt/l0/ialirt_spice.py
laspsandoval: You are probably right on this. I do not know why there is an offset.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: The MAG measurements have their own timestamp represented by mag_acq_tm_coarse and mag_acq_tm_fine. Those are used to create "target_time". And that timestamp is not the same as the spacecraft timestamp. 

The spacecraft timestamp is associated with the right ascension/declination/spin phase values. So that is why we are interpolating here.

imap_processing/ialirt/l0/ialirt_spice.py
laspsandoval: Here is how I came to that value.Basically I just plotted different values and kept seeing the offset. It's a brutal method that I also don't like.

```
@pytest.mark.external_kernel
def test_transform_instrument_vectors_to_inertial_postlaunch(
    postlaunch_xarray_data, spice_test_data_path, furnish_kernels
):
    """Test real-world application of this function."""
    met = calculate_time(
        postlaunch_xarray_data["sc_sclk_sec"], postlaunch_xarray_data["sc_sclk_sub_sec"], 256
    )

    attitude_time = ttj2000ns_to_et(met_to_ttj2000ns(met))
    sc_spin_phase_rad = postlaunch_xarray_data["sc_spin_phase"].astype(float) * (
            2 * np.pi / 65535.0
    )
    sc_inertial_right = postlaunch_xarray_data["sc_inertial_right"].astype(float) * (
            0.0055 * np.pi / 180
    )
    sc_inertial_decline = postlaunch_xarray_data["sc_inertial_decline"].astype(float) * (
            0.0027 * np.pi / 180
    )

    kernels = [
        "imap_science_100.tf",
        "imap_120.tf.txt",
        "naif0012.tls",
        "de440s.bsp",
        "imap_recon_od005_20250925_20251014_v01.bsp",
        "pck00011.tpc",
        "imap_sclk_0036.tsc.txt",
        "imap_2025_283_2025_284_001.ah.bc",
    ]

    v_manual_0_all = []
    mago_inertial_vector_all = []

    with furnish_kernels(kernels):

        for i in range(len(attitude_time)):

            instrument_vector = np.array([[-2.525630188, -0.337087161, -4.523789905]])

            v_manual_0 = transform_instrument_vectors_to_inertial(
                instrument_vector,
                np.array([np.degrees(sc_spin_phase_rad[i])]), # spin phase
                np.array([np.degrees(sc_inertial_right[i])]), # right ascension
                np.array([np.degrees(sc_inertial_decline[i])]), # declination
                SpiceFrame.IMAP_MAG_O,
            )

            mago_inertial_vector = frame_transform(attitude_time[i], instrument_vector,
                                                   from_frame=SpiceFrame.IMAP_MAG_O,
                                                   to_frame=SpiceFrame.ECLIPJ2000)

            v_manual_0_all.append(v_manual_0)
            mago_inertial_vector_all.append(mago_inertial_vector)
```

laspsandoval: And then many plots later I decided to do an offset.

laspsandoval: See explanation below. Agreed. I don't like it. Any ideas though?

imap_processing/tests/spice/test_data/imap_120.tf
tmplummer, line 1: Not sure if you saw the PR I have up, but I added the latest imap frames kernel, `imap_130.tf`, so no need for this one here if you rebase.

imap_processing/ialirt/l0/parse_mag.py
laspsandoval: Added a spline, although the results are similar.

imap_processing/ialirt/l0/ialirt_spice.py
tmplummer: My first thought would be to do an analysis on the reported spin-phase from the packet and the spin-phase as computed by SPICE. If those two don't agree, then this needs to be chased upstream to figure out why they don't agree.

laspsandoval: In the get_x_y_axes function we arbitrarily select an x-axis and y-axis. And that is the problem. At some point we have to calibrate either the spin phase or the x or y axis with SPICE or we do not know where we are.

laspsandoval:
I'm taking a different stance on the spin phase offset:

In the get_x_y_axes function we arbitrarily select an x-axis and y-axis. And that is the problem. At some point we have to calibrate either the spin phase or the x or y axis with SPICE or we do not know where we are.

So we can either have the spin phase offset or we can match the x-axis to the SPICE axis in get_x_y_axes. But we will need a constant somewhere in the code to align with the inertial frame.

Furthermore, there is nothing special about spin_phase = 0 from the telemetry. It doesn't match the x-axis in the inertial frame. Example:

Time at which spin phase equals 0 : 813433294.5140676

ref = spice.pxform("IMAP_SPACECRAFT", "ECLIPJ2000", 813433294.5140676)
x_ref = ref[:, 0]  # inertial direction of S/C +X
x_ref
array([ 0.23525427, -0.96026672,  0.15014413])

What I could do (maybe better?) is to calibration the x, y, and z axis at spin phase  = 0 and remove the function get_x_y_axes altogether. Or I could just keep what I have. Up to you all but we will need a calibration.

imap_processing/ialirt/l0/ialirt_spice.py
laspsandoval: See comment below.

imap_processing/tests/ialirt/unit/test_parse_mag.py
laspsandoval: Oh true. I updated to it isn't antipodal.

tmplummer:
> In the get_x_y_axes function we arbitrarily select an x-axis and y-axis. And that is the problem. At some point we have to calibrate either the spin phase or the x or y axis with SPICE or we do not know where we are.

@laspsandoval, I don't think it is fair to say that the x and y axes are selected arbitrarily. In fact, the code is very similar to the code for generating the pointing_attitude kernel. With the exception of the "alternate reference vector" part, the `get_x_y_axes` function calculates a y-axis that is perpendicular to the s/c z-axis and the HAE z-axis (north pole). The result is that the y-axis is in the HAE ecliptic plane. Then the x-axis is computed such that it completes the right-hand orthogonal set of basis-vectors. This also means it should be the closest spin-vector to the HAE north pole.


> Furthermore, there is nothing special about spin_phase = 0 from the telemetry. It doesn't match the x-axis in the inertial frame. Example:
> 
> Time at which spin phase equals 0 : 813433294.5140676
> 
> ref = spice.pxform("IMAP_SPACECRAFT", "ECLIPJ2000", 813433294.5140676) x_ref = ref[:, 0] # inertial direction of S/C +X x_ref array([ 0.23525427, -0.96026672, 0.15014413])

Is the ET in that example a time that we expect the s/c to be at spin-phase 0? There better be something special about this time.
Looking at this example in more detail... The full transform matrix at ET = 813433294.5140676 is:
```
[[ 0.23525427, -0.03275916, -0.97138163],
 [-0.96026672,  0.14655792, -0.23750496],
 [ 0.15014413,  0.9886595 ,  0.00302085]]
```
Vertical columns are the s/c x, y, z axes in HAE coordinates. I see that:
- the s/c y-axis is very close to the HAE z-axis. This is 90-degrees off from what I would expect.
- The s/c z (spin-axis) is basically in the ecliptic plane as expected

It would be interesting to compute this for more spin-phase = 0 times. 

I understand MAG's desire to put the code changes in this PR into production, so I will approve this PR with the caveat that a ticket is created to run this fixed offset to ground.

imap_processing/tests/spice/test_data/imap_120.tf
tmplummer, line 1: Can you please remove this file and use `imap_130.tf` which is now in the repo?

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -387,6 +388,78 @@ def apply_gradiometry_correction(
...
+        [
+            np.ones_like(sc_inertial_right),
+            sc_inertial_right,
+            sc_inertial_decline,
+        ],
+        axis=-1,
+    )
+    vecs = spherical_to_cartesian(spherical_coords)
+
+    spline_x = CubicSpline(attitude_time, vecs[:, 0])
```
tmplummer, line 440: Can you add a comment about why CubicSpline interpolation was chosen here?

tmplummer (commented):
Will you please write up a ticket to cover the work needed to understand and either justify or get rid of the `SPIN_PHASE_OFFSET_DEG`. Two other comments that I would like to see addressed before merging this PR.

alastairtree:
I agree with what Tim said, we should carry one grinding this down so it is well understood, and that we have a well defined way to go from the I-ALiRT attitude information to a well defined frame of reference, and ideally that frame is aligned with something well understood in SPICE. Presumably the SPICE kernels are generated in a very similar way. 
Laura's offset demonstably method works pretty well, but I think for long term sanity we should have it match SPICE exactly if at all possible. I dont want to have to debug "why is the science and i-alirt off by 0.21245xxxxx degrees and 100ms" in 3 years time.
My other fear is that we pick something wich drifts and is accurate for some times and not others, hope I am wrong on that, but I think we need to be sure that cannot happen. 
This branch is better than what we have now, so we could take it, but like tim said, lets follow it on now we are all in the zone.

alastairtree (approved):

laspsandoval:
> Will you please write up a ticket to cover the work needed to understand and either justify or get rid of the `SPIN_PHASE_OFFSET_DEG`. Two other comments that I would like to see addressed before merging this PR.

https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/2459

tmplummer (approved):
Looks good to move forward with for now.

PR #2439, opened 2025-11-17
ULTRA l2 group fine pset energy bins 

lacoak21:
# Change Summary

## Overview
Bin fine grained energy bins in l2. 

Instructions from the ULTRA IT team: [nov5_algorithm_updates_v2 (1).pptx](https://github.com/user-attachments/files/23593853/nov5_algorithm_updates_v2.1.pptx)

## Updated Files
- imap_processing/cli.py
   - Pass in ancillary file for ultra l2 processing 
- imap_processing/ultra/l1c/l1c_lookup_utils.py, imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
   - allow for build_energy_bin and get_energy_delta_minus_plus function to receive a list of energy bin edges.
  - add helper function to bin psets before they get projected to the map 

## Testing
Coarse energy bin tests


imap_processing/ultra/l1c/l1c_lookup_utils.py
```diff
@@ -336,7 +343,9 @@ def build_energy_bins() -> tuple[list[tuple[float, float]], np.ndarray, np.ndarr
         Array of geometric means of energy bins.
     """
     # Create energy bins.
-    energy_bin_edges = np.array(UltraConstants.PSET_ENERGY_BIN_EDGES)
+    if energy_bin_edges is None:
+        energy_bin_edges = np.array(UltraConstants.PSET_ENERGY_BIN_EDGES)
```
greglucas, line 347: Might want to put a log in here indicating we are using the default.
```suggestion
        energy_bin_edges = np.array(UltraConstants.PSET_ENERGY_BIN_EDGES)
        logger.info(f"No energy bin file found, using default pointing bin energy edges {energy_bin_edges}")
```

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -49,6 +52,15 @@
 DEFAULT_L2_HEALPIX_NSIDE = 32
 DEFAULT_L2_HEALPIX_NESTED = False
 
+# Set the default energy bin size
+DEFAULT_GROUP_SIZE = 4
+# Define energy bin groups for binning L1C energy bins into coarser bins
+# Get number of fine energy bins used in L1C PSETs
+n_fine_energy_bins = len(build_energy_bins()[2])
+DEFAULT_BIN_GROUPS = np.arange(n_fine_energy_bins)[::DEFAULT_GROUP_SIZE]
```
greglucas: I'm not sure what "groups" means here. This looks like the group edges to me. I would have expected it to be `[(0, 1, 2, 3), (4, 5, 6, 7), ...] to enumeate the indices within the groups.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -148,12 +171,114 @@ def get_variable_attributes_optional_energy_dependence(
...
+            f" Adding another 'catch-all' bin. New coarse energy bin edges:"
+            f" {bin_groups}"
+        )
+    if bin_groups[0] > 0:
+        bin_groups = np.insert(bin_groups, 0, 0)
+        logger.info(
+            "The given bin_groups did not include the first fine energy bins."
+            f" Adding another 'catch-all' bin. New coarse energy bin edges:"
+            f" {bin_groups}"
+        )
```
greglucas: Aren't you doing all of this in the earlier utility function as well? I'm wondering if this is all necessary, or if you can raise and tell the user to provide the correct values instead of you guessing and doing a catch-all.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -148,12 +171,114 @@ def get_variable_attributes_optional_energy_dependence(
...
+    # energy bin index to sum/average over the fine bins.
+    pset = pset.assign_coords(
+        energy_bin_index=(
+            "energy_bin_geometric_mean",
+            np.digitize(np.arange(n_fine_bins), bin_groups, right=False),
+        )
+    )
+    # Count number of non zero pixels in each new energy bin
+    non_zero_pixels_per_group = (
+        (pset[vars_to_average] > 0).astype(int).groupby("energy_bin_index").sum()
```
greglucas: `np.nonzero`? Do you run into any issues if you have masked/filled quantities like `1e31` in these?

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -148,12 +171,114 @@ def get_variable_attributes_optional_energy_dependence(
...
+    )
+    # Make sure the variables use the new energy bin coordinate
+    pset = (
+        pset.swap_dims({"energy_bin_index": "energy_bin_geometric_mean"})
+        # Restore original dimension order because groupby moves the grouped
+        # dimension to the front
+        .transpose("epoch", "energy_bin_geometric_mean", ...)
+        .drop("energy_bin_index")
+    )
+    return pset, new_bin_edges
```
greglucas: Should the bin_edges be a part of the pset itself? I think this is the energy_delta_minus and energy_delta_plus variables that could be referenced/updated for this.

I think if you can put this information in the pset itself that would simplify a lot of your function calls and not require you to pass it through everywhere.

imap_processing/ultra/l2/ultra_l2.py
lacoak21: Oh yeah I will change the name. That was a remnant from earlier when I was doing something else.

imap_processing/ultra/l2/ultra_l2.py
lacoak21: Which util? I asked Nick about this senario and he mentioned that they would prefer a catch-all bin.

imap_processing/ultra/l2/ultra_l2.py
lacoak21: Oh good point. I just checked and the fill val is negative. So I think this as is is working alright here

lacoak21: Ill change the comment because I only want counts above 0

imap_processing/ultra/l2/ultra_l2.py
lacoak21: True!

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -148,11 +174,124 @@ def get_variable_attributes_optional_energy_dependence(
...
+    # energy_bin_index = [0,0,0,0,1,1,1,1,2,2,2,2...12] That way we can groupby the new
+    # energy bin index to sum/average over the fine bins.
+    pset = pset.assign_coords(
+        energy_bin_index=(
+            "energy_bin_geometric_mean",
+            np.digitize(np.arange(n_fine_bins), bin_groups, right=False),
+        )
+    )
+    # Count number of pixels
+    non_zero_pixels_per_group = (
```
tmplummer, line 244: It is sort of strange to me that the average computed across the fine energy bins excludes pixels with zeros.

lacoak21: Yeah these were the instructions: 

Average each data product directly. The averages should be computed such that each pixel bin is summed, then divided by the number of times that data product was summed. 

e.g If using a group of 4, if only one of the fine energy bins of the 4 had data, the average would be that that data point/1

I can ask more about this in a tagup I have with them today

tmplummer: No need to waste time. If that is what they told you to do, I bet there is a reason to do it that way.

imap_processing/tests/lo/test_cdfs/imap_lo_l1c_pset_20260101-repoint01261_v001.cdf
greglucas, line 1: You should rebase / bring in the latest dev updates as this has been gitignored now I think.

imap_processing/ultra/l2/ultra_l2.py
greglucas: I was thinking the `build_energy_bins()` function, but now that I look at it that is maybe just for the psets and not L2? But then above this you use: `build_default_coarse_bin_edges()`

If someone explicitly passes in the bin edges, then I don't think we should be adding the low/high into it for them, that was their perogative to ignore the low/high energies. Maybe someone wants a map with only 1 energy step from the mapper tool.

Basically, I think this is handled by your "default energy bins" function isn't it? And otherwise someone is explicit in what they asked for and we should honor that explicitness.

imap_processing/ultra/l2/ultra_l2.py
```diff
@@ -148,11 +167,143 @@ def get_variable_attributes_optional_energy_dependence(
...
+        ((pset[vars_to_average] != 0) | (pset[vars_to_average] != FILLVAL_FLOAT32))
```
greglucas: and? Otherwise your FILLVAL_FLOAT32 will match the first condition here, but you do want to ignore it.

imap_processing/ultra/l2/ultra_l2.py
lacoak21: OK I see your point 👍

tmplummer (approved):
This all looks good to me!

PR #2441, opened 2025-11-18
Mag L1D Repair attributes

### LINKED ISSUE #2297: BUG - MAG L1D jobs always fail

maxinelasp:
# Change Summary
Fixes #2297 along with another bug introduced in #2378. Also completes the attributes for L1D.

## Overview

- Add new attributes for L1D - which are very similar to the ones for L2
- Fix the ancillary files so they do not upload through the write_cdf and upload flow
- Fix the ancillary files so errors do not interrupt science processing upload

## Testing

Added a test for the bugs that I saw causing this issue


imap_processing/cdf/config/imap_mag_l2_variable_attrs.yaml
```diff
@@ -63,20 +89,29 @@ direction_attrs:
...
-  FIELDNAM: Compressed data bit width
-  LABLAXIS: Bit width
+  CATDESC: Magnitude of the magnetic field vector
+  FIELDNAM: Magnetic Field Magnitude
+  LABLAXIS: "|B|"
+  FORMAT: F12.5
+  FILLVAL: -1.0e+31
+  UNITS: nT
+  VALIDMIN: 0
+  VALIDMAX: 1.0e+31
```
greglucas: We will never get more than a few 10s of nT, maybe 50k when we are in Earth's magnetic field. I think Andriy said they wanted these to be semi-realistic estimates.
```suggestion
  VALIDMAX: 1.0e+5
```

imap_processing/cli.py
```diff
@@ -1206,6 +1207,80 @@ def do_processing(  # noqa: PLR0912
...
+            if logical_source in ancillary_identifiers:
+                # Skip write_cdf
+                instrument, _data_level, descriptor = dataset.attrs[
+                    "Logical_source"
+                ].split("_")[1:]
+                start_date = self.start_date
+                version = self.version
+
+                output_filename = (
+                    imap_data_access.AncillaryFilePath.generate_from_inputs(
```
greglucas: Rather than recreating this through the subclass are you able to put this logic into the `write_cdf()` function instead? It seems like maybe the only difference is in the filenaming convention here, so you could put a cut-out for that in the write_cdf() I think where you check for mag l1d in there. This just reads as a lot of duplicate code and potential for confusion by calling super in a few different paths here.

imap_processing/cli.py
```diff
@@ -1206,6 +1207,80 @@ def do_processing(  # noqa: PLR0912
...
+                    # Don't fail for any reason for ancillary files
+                    logger.warning(f"Hit error {e} when creating {output_filename}")
+                    continue
+            else:
+                science_datasets.append(dataset)
+
+        try:
+            self.upload_products(ancillary_files)
+        except Exception as e:
+            logger.warning(f"Failed to upload ancillary products due to error {e}")
```
greglucas: If these are output products don't we think they should be done with the same care as science products? I'm not sure why these would fail when a science product wouldn't.

imap_processing/cli.py
maxinelasp: That was my initial thought, but it would require skipping a lot of write_cdf, and most of this code would still exist because it actually is pretty different than what write_cdf does. write_cdf is reading these values from the attributes of the dataset rather than from the class definition, so write_cdf doesn't even have access to the values I use here. So, moving this into write_cdf would require assigning these attributes into the dataset anyway (which is about as much code) and then adding another special case into write_cdf which skips most of it. I didn't want to clutter up shared code with the special case when it's not needed. 

You have a good point about calling super(), I can move all the calls to the bottom so we have this cut out which pulls out the ancillary files and writes them, but then in most cases the call just routes through the parent classes function like normal.

imap_processing/cli.py
maxinelasp: Well, they are currently failing and blocking science product release. Since these are internal watchdog products, I do think they are a lower priority than the science files... but I can remove the excepts if that doesn't line up with the rest of the SDC or what MAG wants.

imap_processing/cdf/config/imap_mag_l2_variable_attrs.yaml
maxinelasp: Good point!

imap_processing/cli.py
maxinelasp: The only code that actually matches with write_cdf is the split() call on the Logical_source and the xarray_to_cdf call at the end (where I pass in different kwargs.)

imap_processing/cli.py
greglucas: I'm just curious if there is a reason they are failing. Can we identify this earlier in the process somehow and not include these in the list of data products to upload if there is bad content. i.e. the upload isn't what should fail and if it does we should fix that at the source.

imap_processing/cli.py
greglucas: > So, moving this into write_cdf would require assigning these attributes into the dataset anyway (which is about as much code) and then adding another special case into write_cdf which skips most of it. I didn't want to clutter up shared code with the special case when it's not needed.

This sounds like it might be nice anyways to add those attributes? I agree though, if it isn't a one/two-line cut-out for ancillary files then it probably doesn't make sense to put it into the write_cdf() function. I thought it just looked like it would only be this one naming line because they are cdf files as well.

imap_processing/cli.py
maxinelasp: That is what the test I added is checking! But, partly the problem is that they are trying to get uploaded as science files and they aren't. So, I don't anticipate this happening again, but if it does, I was thinking it would be preferable to NOT block science products. However, this does reduce visibility into these products if they start failing so I understand your concern. 
Really, these checks aren't for bad content, they're for A. is the filename constructed properly and B. is the ancillary file upload functioning properly. 

But I am definitely open to removing them if the increased stability isn't worth the risk of missing errors from the SDC perspective.

greglucas: In the `post_processing()` method there is a callout for Path vs Dataset difference in the list of products. So my personal preference would be to write the ancillary products here as you have (logging/ignoring errors if there are any if you want in that portion), then call super().post_process() with the mixed list of all science datasets and ancillary filepaths, and let the uploads happen there.

greglucas (approved):
minor nit on preferring a mixed `list[Dataset, Path]` to try and re-use the upload capability, but not blocking on that and looks good to me otherwise so you can take/leave that comment as you see fit.

imap_processing/cli.py
maxinelasp: Oh yeah, that's a good catch that makes this cleaner as well. I'll make that change.

imap_processing/cdf/config/imap_mag_l2_variable_attrs.yaml
```diff
@@ -52,6 +52,32 @@ vector_attrs: &vectors_default
...
+    DICT_KEY: "SPASE>Field>FieldQuantity:Magnetic,Qualifier:Vector,CoordinateSystemName:GSE,CoordinateRepresentation:Cartesian"
+
+vector_attrs_gsm:
+    <<: *vectors_default
+    CATDESC: Magnetic field vectors with x y z varying by time in Geocentric Solar Magnetospheric Reference Frame
+    DICT_KEY: "SPASE>Field>FieldQuantity:Magnetic,Qualifier:Vector,CoordinateSystemName:GSM,CoordinateRepresentation:Cartesian"
+
+vector_attrs_rtn:
+    <<: *vectors_default
+    CATDESC: Magnetic field vectors with x y z varying by time in Radial-Tangential-Normal Reference Frame
```
alastairtree: @mhairifin can you confirm if these are correct for me? Would not label these as r t and n not xyz?

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -222,19 +222,23 @@ def generate_dataset(
         self,
         attribute_manager: ImapCdfAttributes,
         day: np.datetime64,
+        data_level: str = "l1d",
```
alastairtree: Should this be a field on `MagL2L1dBase` that is overridden in each subclass - I cant see why you want to be able to do MagL1d(..).generate_dataset(attributes, day, "some-other-level")?

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -253,16 +257,16 @@ def generate_dataset(
             self.epoch = self.magi_epoch  # type: ignore[no-redef]
             self.range = self.magi_range  # type: ignore[no-redef]
 
-            # Call parent generate_dataset method
-            dataset = super().generate_dataset(attribute_manager, day)
+            # Call parent generate_dataset method with L1D data level
+            dataset = super().generate_dataset(attribute_manager, day, data_level="l1d")
```
alastairtree: Does not need the default value if the default is set by the method definition? Or as I said above move it to a field/property.

imap_processing/tests/mag/test_mag_l1d.py
```diff
@@ -98,9 +101,96 @@ def test_mag_l1d(mag_test_l1d_data, norm_dataset, furnish_kernels, fake_mag_spin
...
+        start_date="20000101",
+        repointing=None,
+        version="v001",
+        upload_to_sdc=False,
+    )
+
+    mock_dependencies = ProcessingInputCollection()
+
+    with patch("imap_processing.cdf.utils.xarray_to_cdf"):
+        mag_processor.post_processing(l1d_datasets, mock_dependencies)
```
alastairtree, line 220: Any value in tests for the the burst files? Parameterised test? Similar for L2.

imap_processing/cli.py
```diff
@@ -1206,6 +1207,77 @@ def do_processing(  # noqa: PLR0912
...
+                        # write file to CDF
+                        xarray_to_cdf(
+                            dataset,
+                            output_filepath,
+                            terminate_on_warning=False,
+                            istp=False,
+                        )
+                        # update the dataset in processed_data to point to a path
+                        processed_data[index] = output_filepath
+                    except Exception as e:
```
alastairtree: Can you catch anything more specific? I don't think you want to catch and continue on a `MemoryError` for example

alastairtree (commented):
Thanks Maxine! Added a couple of small inline bits and looking forward to a working L1D!

imap_processing/cdf/config/imap_mag_l2_variable_attrs.yaml
mhairifin: That's correct, RTN specifically should be labelled as r, t, and n

PR #2447, opened 2025-11-19
ULTRA l1c do not reject events due to scattering. 

lacoak21:
# Change Summary

## Overview
Ultra l1c should not have been rejecting due to scattering anyway because the ancillary file scattering thresholds are all inf but I realized that some might be getting rejected still if the FWHM values were nans. This explicitly make sure we do not reject events based on those flags. 

## Updated Files
   - add bool to not reject based on scattering 
- imap_processing/ultra/l1c/helio_pset.py, imap_processing/ultra/l1c/spacecraft_pset.py


laspsandoval (approved):
LGTM

PR #2452, opened 2025-11-20
Logger formatting

greglucas:
# Change Summary

## Overview

It is useful to see the timestamps of when the log message was written. Right now we have to get this from the cloudwatch logs themselves, but we can also print them out in our normal console output so that they will show up in the version on the website as well.

Right now this looks like:
```
2025-11-19 21:20:10 - INFO:imap_processing.lo.l1a.lo_l1a:
Decommutating packets and converting to dataset
2025-11-19 21:20:10 - INFO:space_packet_parser.generators.utils:Creating packet generator from a filelike object, <_io.BufferedReader name='/Users/grlu5547/code/imap/imap_processing/data/imap/lo/l0/2025/11/imap_lo_l0_raw_20251104-repoint00038_v001.pkts'>. Total length is 37795158 bytes
2025-11-19 21:20:48 - WARNING:imap_processing.utils:Found [4] duplicate packets for APID 676. Dropping duplicate packets and continuing processing.
2025-11-19 21:20:49 - WARNING:imap_processing.utils:Found [14231] duplicate packets for APID 677. Dropping duplicate packets and continuing processing.
```

I'm happy to update the formatting if someone else has preferences/ideas for better versions of this too.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -84,13 +84,6 @@ from imap_processing.ultra.l1b import ultra_l1b
 from imap_processing.ultra.l1c import ultra_l1c
 from imap_processing.ultra.l2 import ultra_l2
 
-# Set the basic logging configuration for all users
-# of the CLI tool.
-logging.basicConfig(
-    format="%(asctime)s - %(levelname)s:%(name)s:%(message)s",
-    level=logging.INFO,
-    datefmt="%Y-%m-%d %H:%M:%S",
-)
 logger = logging.getLogger(__name__)
 
 
@@ -204,15 +197,7 @@ def _parse_args() -> argparse.Namespace:
         action="store_const",
         dest="loglevel",
         const=logging.DEBUG,
-        default=logging.WARNING,
-    )
-    parser.add_argument(
-        "-v",
-        "--verbose",
-        help="Add verbose output",
-        action="store_const",
-        dest="loglevel",
-        const=logging.INFO,
+        default=logging.INFO,
     )
     parser.add_argument("--instrument", type=str, required=True, help=instrument_help)
     parser.add_argument("--data-level", type=str, required=True, help=level_help)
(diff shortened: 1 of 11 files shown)


imap_processing/codice/codice_l1a.py
```diff
@@ -29,7 +29,6 @@
 from imap_processing.spice.time import met_to_ttj2000ns
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
```
tech3371, line 32: are removing this because it's INFO by default?

tech3371 (approved):
Minor curious questions. Otherwise, looks good to me

imap_processing/codice/codice_l1a.py
greglucas: Sort of for consolidation, but also to potentially remove logs when unnecessary. For example, if we run the tests now and we don't import the `cli.py` module, you shouldn't be getting the CoDICE logs anymore. I was annoyed by the instrument logs in tests all the time too.

imap_processing/cli.py
```diff
@@ -84,6 +84,13 @@
 from imap_processing.ultra.l1c import ultra_l1c
 from imap_processing.ultra.l2 import ultra_l2
 
+# Set the basic logging configuration for all users
+# of the CLI tool.
+logging.basicConfig(
+    format="%(asctime)s - %(levelname)s:%(name)s:%(message)s",
+    level=logging.INFO,
```
tmplummer: Where does the loglevel from the parsed args get used? I don't actually see that being used anywhere and I think we need to override this level based on the parsed args.

tmplummer (commented):
Nice! Just a question about how the overall loglevel gets set using the command line args.

imap_processing/cli.py
greglucas: Oh wow, good catch! I thought we were setting it in our args. It looks like verbose/debug never did anything.

I moved this into the argparse area, so we will only set the basicConfig information when going through the CLI path now. I defaulted to `INFO` and if someone passes `--debug` then we bump the loglevel to `DEBUG`. Let me know your thoughts on this approach.

tmplummer (approved):
Looks good to me! Thanks for fixing the unused cli arguments. I had tried both `--debug` and `-v` in the past with no effect. Now I know why.

PR #2456, opened 2025-11-21
ENH: Add Lo L1b instrument status summary product

greglucas:
# Change Summary

## Overview

Adding the state vector data product to Lo, which is a combination of 3 different packets into one data product.

### DIFF
--- a/imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
@@ -95,6 +95,12 @@ imap_lo_l1b_shk:
   Logical_source: imap_lo_l1b_shk
   Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1B Data
 
+imap_lo_l1b_instrument-status-summary:
+  <<: *instrument_base
+  Data_type: L1B_star>Level-1B Status Summary
+  Logical_source: imap_lo_l1b_instrument-status-summary
+  Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1B Data
+
 imap_lo_l1c_goodtimes:
   <<: *instrument_base
   Data_type: L1C_goodtimes>Level-1C Goodtimes List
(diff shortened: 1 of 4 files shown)


imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
```diff
@@ -95,6 +95,12 @@ imap_lo_l1b_shk:
   Logical_source: imap_lo_l1b_shk
   Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1B Data
 
+imap_lo_l1b_statevector:
```
sdhoyt: I'm not sure why the algorithm document still says state vector, but I have instrument status summary in the pipeline diagram. We decided as a mission to call this an instrument status summary because people will get confused and think of state vector as position, velocity vectors. Sorry, I should have mentioned this to you before you started on this.

imap_processing/lo/l1a/lo_l1a.py
```diff
@@ -347,3 +356,65 @@ def add_dataset_attrs(
...
+        Combined dataset containing the instrument state vector.
+    """
+    datasets_to_merge = [
+        ds
+        for apid, ds in datasets_by_apid_derived.items()
+        if apid in [LoAPID.ILO_APP_NHK, LoAPID.ILO_APP_SHK, LoAPID.ILO_DIAG_PCC]
+    ]
+    combined_ds = xr.merge(datasets_to_merge, compat="override")
+    # Only keep the desired keys for the state vector
+    state_vector_keys = [
```
sdhoyt: Are the an_c_thr, tof3_thr, tof2_thr, tof1_thr, tof0_thr, ifb_hot_spot_t, coarse_pot_pri, and fine_pot_pri no longer needed in this?

sdhoyt (commented):
Thanks for doing this!

imap_processing/lo/l1a/lo_l1a.py
greglucas: !! Thank you for actually looking at this, I just didn't continue to the second page. Updating now...

sdhoyt (approved):

greglucas:
I have changed `shcoarse` attributes to have a `DEPEND_0` on epoch now, rather than a `DEPEND_1` on itself which makes all l1a data products have everything depend on epoch instead of a separate hanging coordinate with nothing else dependent on it.

PR #2457, opened 2025-11-21
MNT: Suppress log messages and filter SPP warnings during tests

greglucas:
Quality of life improvement here. If you go to even a succeeding test, we get flooded with logs and warnings. I think it would be helpful to filter these out. For example, in this succeeding test there are many of these Number of bits warnings that we just ignore anyways, so we might as well filter them.
https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/19574991215/job/56057805699?pr=2452#step:7:2395

There are many CDF schema logger.warnings filling the tests that are not relevant. We can use the command line option to filter these warnings.

There are also space packet parser warnings that are emitted for many of our packets that are the wrong size, but that is all we have so we can ignore them during tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,9 +62,9 @@ jobs:
             # Download external data before we run our tests so that
             # we don't try to download them in parallel
             poetry run python imap_processing/tests/conftest.py
-            poetry run pytest -vvv -n auto --color=yes --cov --cov-report=xml
+            poetry run pytest -vvv -n auto --color=yes --cov --cov-report=xml --log-disable=root
           else
-            poetry run pytest -vvv -n auto --color=yes --cov --cov-report=xml -m "not external_kernel and not external_test_data"
+            poetry run pytest -vvv -n auto --color=yes --cov --cov-report=xml  --log-disable=root -m "not external_kernel and not external_test_data"
           fi
 
       - name: Upload coverage reports to Codecov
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -62,9 +62,9 @@ jobs:
             # Download external data before we run our tests so that
             # we don't try to download them in parallel
             poetry run python imap_processing/tests/conftest.py
-            poetry run pytest -vvv -n auto --color=yes --cov --cov-report=xml
+            poetry run pytest -vvv -n auto --color=yes --cov --cov-report=xml --log-disable=root
```
greglucas, line 65: I didn't see a way to suppress this through pyproject.toml, so I guess we can only do it via cli for now.

https://docs.pytest.org/en/stable/how-to/logging.html

tmplummer: Does this just disable all logs in the github jobs? I'm not sure what the scope of the root logger is here.

greglucas: Yeah sort of... It suppresses their output in the terminal. Except that we do have some tests with caplog that are asserting on specific messages and those assertions work still. (I tried setting the basicLogLevel to ERROR in our conftest and then some test assertions failed because of that)

The basic idea here is that we have too many logs that are polluting the stream and I can't find the actual content of the failed messages / raised value error sometimes. This is only on pytest and only on GitHub so you would still have to apply this locally to suppress things.

tmplummer (approved):
I'm happy to try this out. Interested to see what happens when a test failure occurs. Presumably the traceback for that will get captured.

greglucas:
> Interested to see what happens when a test failure occurs. Presumably the traceback for that will get captured.

Absolutely! The difference is that you can see the traceback right away now and not have to scroll way up in your terminal. Some instruments were outputting so many logs it actually would go off of the lines my terminal saved and I couldn't even see the actual failure content.

PR #2460, opened 2025-11-21
ENH: Add Lo L1b badtimes data product

### LINKED ISSUE #1873: Lo L1B -Create Badtimes List Product
The bedtimes list is a time period product contains start/stop MET times for badtimes with a corresponding badtime quality flag to indicate the cause of the. badtime.

See section 9.3.2 in the Lo algorithm document for more details: https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/146768943/IMAP+Algorithm+Document+from+Instrument+Teams

greglucas:
# Change Summary

## Overview

This adds the Lo L1b badtimes data product as output during L1b processing. This is the start of the product, removing any times with thruster firings. It is formatted as a data product like our normal science products, so we will make one of these files for each l1b science product now, it is not an ever growing list of times like the ancillary products.

Future work:
* Make use of this in processing? For now it is considered a diagnostic product and we are going to use it as output only. A separate ticket would be to incorporate this into the processing pipeline.
* Incorporate other badtime algorithms.

closes #1873

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -1051,10 +1051,15 @@ def create_badtimes_dataset() -> xr.Dataset:
             data=np.array([], dtype=np.uint8), dims=["epoch"]
         )
 
-        for i in range(1, 8):
-            empty_ds[f"E-Step{i}"] = xr.DataArray(
-                data=np.array([], dtype=np.uint8), dims=["epoch"]
-            )
+        empty_ds["esa_step"] = xr.DataArray(
+            data=np.arange(1, 8, dtype=np.uint8),
+            name="esa_step",
+            dims=["esa_step"],
+        )
+        empty_ds["badtime_flag"] = xr.DataArray(
+            data=np.empty((0, len(empty_ds["esa_step"])), dtype=np.uint8),
+            dims=["epoch", "esa_step"],
+        )
 
         empty_ds["Comment"] = xr.DataArray(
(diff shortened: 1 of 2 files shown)


tmplummer (approved):
I can't find anything to comment on. Good to go!

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1003,3 +1013,98 @@ def create_datasets(
...
+    thruster_ds["bin_start"] = xr.DataArray(
+        data=np.zeros(len(thruster_ds["epoch"]), dtype=np.uint8),
+        dims=["epoch"],
+    )
+    thruster_ds["bin_end"] = xr.DataArray(
+        data=np.full(len(thruster_ds["epoch"]), 59, dtype=np.uint8),
+        dims=["epoch"],
+    )
+    for i in range(1, 8):
+        thruster_ds[f"E-Step{i}"] = xr.DataArray(
```
sdhoyt: should this use "esa_step" like the other L1B products or is this supposed to match the CSV column (E-Step#)?

greglucas: I was trying to match the csv formatting, that is a good question though. I see that it is actually called `badtime_flag` (N, 7) now that I read closer in the algorithm document.

sdhoyt: Ah yeah, you're right!

sdhoyt (approved):

PR #2461, opened 2025-11-21
MNT: Change Lo logging in loop from info to debug

greglucas:
There are a lot of packets, so this adds a lot of logs to the command line. We can add the log statements for debugging purposes, but don't need them all the time.

Noticed this while running `imap_cli` locally and it added a lot of extra prints. @sdhoyt are you OK with switching to debug if we need it?


sdhoyt (approved):
Sounds good. Thanks!

PR #2463, opened 2025-11-24
FIX: SWE l1b needs to handle all full-cycle case as well

greglucas:
Before we were only setting the local variable if there was a mismatch in the full cycle length compared to packets. We can ignore this if-block and just subset the dataset based on full cycle data regardless and the isel just won't do anything in the case of a full match.

@tech3371 I am not sure if there is a good way to test this case? It looks like you are using raw packets, so I'm not sure how I could easily modify those to produce the same length of completed packets.

### DIFF
--- a/imap_processing/swe/l1b/swe_l1b.py
+++ b/imap_processing/swe/l1b/swe_l1b.py
@@ -647,8 +647,6 @@ def swe_l1b_science(dependencies: ProcessingInputCollection) -> xr.Dataset:
     science_files = dependencies.get_file_paths(descriptor="sci")
     l1a_data = load_cdf(science_files[0])
 
-    total_packets = len(l1a_data["science_data"].data)
-
     l1a_data_copy = l1a_data.copy(deep=True)
 
     # First convert some science data to engineering units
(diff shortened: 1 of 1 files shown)


imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -695,25 +695,24 @@ def swe_l1b_science(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
-            f"{full_cycle_l1a_data['quarter_cycle'].data}"
+    logger.info(
+        f"Total packets before filtering [{total_packets}]"
+        f"Number of full cycles found [{len(full_cycle_data_indices)}]"
+    )
+    full_cycle_l1a_data = l1a_data_copy.isel({"epoch": full_cycle_data_indices})
+
+    # Update total packets
+    total_packets = len(full_cycle_data_indices)
+    logger.debug(
```
tech3371: good debug logs

tech3371 (approved):
Thank you for fixing this! LGTM

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -695,25 +695,24 @@ def swe_l1b_science(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+        f"Quarters cycle after filtering: {full_cycle_l1a_data['quarter_cycle'].data}"
+    )
+    if len(full_cycle_data_indices) != len(full_cycle_l1a_data["quarter_cycle"].data):
```
tech3371: This check may be irrelevant now. Can you remove this check and error raise? Can you remove line 650 where I assigned to total_packets. Line 707 assignment will be enough.

imap_processing/swe/l1b/swe_l1b.py
```diff
@@ -695,25 +695,24 @@ def swe_l1b_science(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
-
-        # Update total packets
-        total_packets = len(full_cycle_data_indices)
-        logger.debug(
-            "Quarters cycle after filtering: "
-            f"{full_cycle_l1a_data['quarter_cycle'].data}"
+    logger.info(
+        f"Total packets before filtering [{total_packets}]"
+        f"Number of full cycles found [{len(full_cycle_data_indices)}]"
+    )
```
tech3371, line 702: we can update this logs to print quarter cycle like you had below and remove this if we make below changes?

tech3371 (commented):
Couple changes after rethink.

imap_processing/swe/l1b/swe_l1b.py
greglucas: Yep, agreed this is redundant, so removed it.

imap_processing/swe/l1b/swe_l1b.py
greglucas: This is an info log, so just printing out the sizes of these. The other logs are debugs with more information.

tech3371 (approved):
looks great. Thank you!

Issue #2257, opened 2025-09-29
CoDICE: Lo L1A counters single and aggregated re-validation


tech3371:
- [x] Refactor Lo counters products
- [x] Validate. Main variables that are active are validated.
- [x] Double check Lo counters aggregated's decompressed data reshape order and update as noted in the comments
- [x] Update non-active variable's filval based on what Joey does. - Updated to use fillval of uint32

tech3371:
Counters aggregated reshape was correct but singles needed update instead. Push that change in PR #2465

tech3371:
Created ticket for other minor updates #2469  and closing this ticket.

PR #2465, opened 2025-11-24
CoDICE: Lo counters

tech3371:
# Change Summary
Related to #2257 

## Overview

Refactoring done. Validation of active variables are done. non-active variables need to be update in validation files and revalidate later.

## New Files

- two new files for Lo singles and aggregated counters processing function

## Updated Files

- Refactor of CDF attrs.
- Updates to include above new changes

## Testing

- tests update and captured TODO

### DIFF
--- a/imap_processing/codice/codice_l1a_lo_counters_aggregated.py
+++ b/imap_processing/codice/codice_l1a_lo_counters_aggregated.py
@@ -10,7 +10,6 @@ from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.codice import constants
 from imap_processing.codice.decompress import decompress
 from imap_processing.codice.utils import (
-    CODICEAPID,
     ViewTabInfo,
     calculate_acq_time_per_step,
     get_codice_epoch_time,
@@ -77,9 +76,6 @@ def l1a_lo_counters_aggregated(
     voltage_data = sci_lut_data["esa_sweep_tab"][f"{esa_table_number}"]
 
     # ========= Decompress and Reshape Data ===========
-    if view_tab_obj.apid != CODICEAPID.COD_LO_INST_COUNTS_AGGREGATED:
-        raise ValueError("Unsupported APID for Lo Counters aggregated processing.")
-
     logical_source_id = "imap_codice_l1a_lo-counters-aggregated"
     # Counters is little bit different in how CDF variables are derived.
     # For singles, CDF variables are coming from 'product' tab. But for
--- a/imap_processing/codice/codice_l1a_lo_counters_aggregated.py
+++ b/imap_processing/codice/codice_l1a_lo_counters_aggregated.py
@@ -1,4 +1,4 @@
-"""CoDICE L1A Hi Counters aggregated processing functions."""
+"""CoDICE L1A Lo Counters aggregated processing functions."""
 
 import logging
 from pathlib import Path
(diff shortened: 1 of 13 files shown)


imap_processing/codice/codice_l1a_lo_counters_aggregated.py
```diff
@@ -0,0 +1,258 @@
...
+    # Use plan id and plan step to get voltage data's table_number in ESA sweep table.
+    # Voltage data is (128,)
+    esa_table_number = sci_lut_data["plan_tab"][f"({plan_id}, {plan_step})"][
+        "lo_stepping"
+    ]
+    voltage_data = sci_lut_data["esa_sweep_tab"][f"{esa_table_number}"]
+
+    # ========= Decompress and Reshape Data ===========
+    if view_tab_obj.apid != CODICEAPID.COD_LO_INST_COUNTS_AGGREGATED:
+        raise ValueError("Unsupported APID for Lo Counters aggregated processing.")
```
greglucas: Didn't you pass in the apid to the view_tab_obj earlier, so can you ever encounter this case?

imap_processing/codice/codice_l1a_lo_counters_singles.py
```diff
@@ -0,0 +1,247 @@
...
+        l1a_dataset[var] = xr.DataArray(
+            unpacked_dataset[var].values,
+            dims=("epoch",),
+            attrs=cdf_attrs.get_variable_attributes(var),
+        )
+
+    # Finally, add species data variables and their uncertainties.
+    # Since singles only has one variable, we can directly add it here.
+    l1a_dataset["apd_singles"] = xr.DataArray(
+        counters_data[:, 0, :, :, :],
```
greglucas: Rather than indexing with the length-1 second index here, would it make more sense to do for singles above when creating `counters_data` as:
```
-1, esa_step, spin_sector_pairs, inst_az
```

dropping `len(variable_names)` because that is 1 for all singles here apparently? That way it would line up nicely with your dimensions here too.

greglucas (commented):
Minor nits is all. Looks good otherwise. We should probably keep this open until Joey gets back to us about the dimension reshaping for validation data?

imap_processing/codice/codice_l1a_lo_counters_aggregated.py
tech3371: yep. I may be overdoing this. Have been adding that for all other products as well.

tech3371: done. cleaned up other places that had this additional check.

tech3371:
> Minor nits is all. Looks good otherwise. We should probably keep this open until Joey gets back to us about the dimension reshaping for validation data?

@greglucas That sounds good. I can see us handle this two ways. I was planning on keeping this ticket, https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/2257, open until those comments are addressed. So address that in second half of that ticket. I captured notes on what is remaining and whoever pick will have some reference. or keep this PR open and you or @lacoak21 can add those remaining updates to this PR if I don't receive those updates today.

greglucas:
If you think we are good now, I'm happy to merge and do a small commit later if needed to change things around. Especially if this will get data flowing that the instrument teams want/need.

greglucas (approved):

PR #2466, opened 2025-11-24
BUG - Lo Hist Rate L1B - Swap ESA Step and Az dims, intialize exposure factor to 1 instead of 0

sdhoyt:
# Change Summary

## Overview
- I forgot I had already swapped the ESA Step and Azimuth dimensions in the L1A product and needed to do that now for the L1B product. The dimension ordering is now <epoch, esa_step, spin_bin> (azimuth names will be changed to spin bin in a later PR)

- The exposure factor is now initialized as 1 instead of 0. The factor is used to scale (multiple) the exposure times, so if there if a ESA step is not reswept, then the exposure time should be multiplied by 1, not 0. The remaining logic still works as it will increment each time a ESA field is reswept and scale as expected.

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -1254,10 +1254,12 @@ def resweep_histogram_data(
     o_counts_reswept = np.zeros_like(l1b_histrates["o_counts"].values)
 
     # Get the number of azimuth bins from the l1b_histrates dataset
-    num_azimuth = l1b_histrates["h_counts"].shape[2]
+    num_azimuth = l1b_histrates.sizes["spin_bin_6"]
     # initialize exposure factor to 1 as this will be used to scale (multiply)
     # the exposure time later
-    exposure_factor = np.full((len(epochs), 7, num_azimuth), 1, dtype=int)
+    exposure_factor = np.full(
+        (len(epochs), l1b_histrates.sizes["esa_step"], num_azimuth), 1, dtype=int
+    )
 
     for epoch_idx, epoch in enumerate(epoch_utc):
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1254,8 +1254,10 @@ def resweep_histogram_data(
     o_counts_reswept = np.zeros_like(l1b_histrates["o_counts"].values)
 
     # Get the number of azimuth bins from the l1b_histrates dataset
-    num_azimuth = l1b_histrates["h_counts"].shape[1]
-    exposure_factor = np.zeros((len(epochs), num_azimuth, 7), dtype=int)
+    num_azimuth = l1b_histrates["h_counts"].shape[2]
```
tmplummer: You could change this to access the dimension size directly:
```suggestion
    num_azimuth = l1b_histrates.sizes["spin_bin_6"]
```

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1254,8 +1254,10 @@ def resweep_histogram_data(
...
+    # initialize exposure factor to 1 as this will be used to scale (multiply)
+    # the exposure time later
+    exposure_factor = np.full((len(epochs), 7, num_azimuth), 1, dtype=int)
```
tmplummer: I guess that the number of ESA steps won't change, but that could be parameterized here too.
```suggestion
    exposure_factor = np.full((len(epochs), l1b_histrates.sizes["esa_step"], num_azimuth), 1, dtype=int)
```

tmplummer (commented):
Just a couple of suggestions on how to better leverage xarray.

greglucas (approved):

PR #2467, opened 2025-11-24
ULTRA l1b coinph valid and backtof valid flags

lacoak21:
# Change Summary

## Overview
Add a quality flag for BACKTOF and make sure to cull events with that flag and COINPH in l1c.

## Updated Files
   - Make sure to flag both coinphvalid and backtof valid events
- imap_processing/ultra/l1b/quality_flag_filters.py
   - Add quality outlier flags to rejection mask used at l1c
- fix backtof flags to follow pattern and not use numpy fancy indexing!! 

## Testing
Fix test

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
@@ -735,13 +735,14 @@ def test_is_back_tof_valid(test_fixture, ancillary_files):
     df_ph = df_filt[np.isin(df_filt["StopType"], [StopType.PH.value])]
 
     # Get the ph inds
-    ph_indices = np.nonzero(
-        np.isin(de_dataset["stop_type"], [StopType.Top.value, StopType.Bottom.value])
-    )[0]
-    quality_flags = np.zeros(len(ph_indices), dtype=np.uint16)
-    valid = is_back_tof_valid(
-        de_dataset.isel(epoch=ph_indices),
-        df_filt.Xf.astype("float").values[ph_indices],
+    ph_mask = (
+        (de_dataset["stop_type"] == StopType.Top.value)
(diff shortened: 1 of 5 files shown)


imap_processing/tests/ultra/unit/test_ultra_l1b_extended.py
```diff
@@ -734,12 +734,19 @@ def test_is_back_tof_valid(test_fixture, ancillary_files):
     df_filt, _, _, de_dataset = test_fixture
     df_ph = df_filt[np.isin(df_filt["StopType"], [StopType.PH.value])]
 
+    # Get the ph inds
+    ph_indices = np.nonzero(
+        np.isin(de_dataset["stop_type"], [StopType.Top.value, StopType.Bottom.value])
+    )[0]
```
greglucas: Could this be written as the following using concatenation instead of several numpy calls? I'm not sure it is more efficient, but reads a bit clearer to me than the nonzero version.
```suggestion
    ph_indices = (de_dataset["stop_type"] == StopType.Top.value) | (de_dataset["stop_type"] == StopType.Bottom.value])
```

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -1391,7 +1389,7 @@ def is_back_tof_valid(
     valid[bottom_mask] = (diff[bottom_mask] >= diff_bt_min) & (
         diff[bottom_mask] <= diff_bt_max
     )
-
+    quality_flags[~valid] |= ImapDEOutliersUltraFlags.BACKTOF.value
```
greglucas, line 1394: Should we be returning quality_flags if we are updating them in these routines to make it less cryptic where these updates are happening? i.e. you are mutating this here but I had no idea based on the call signature.

greglucas (approved):
A couple of nits, but the implementation looks good to me.

Issue #2471, opened 2025-12-01
L2 Hi Lo - Remove energy from esa fit factor ancillary file


tmplummer:
### Algorithm Description:

During flux correction validation, I identified that energies in the flux correction were coming from the ancillary file instead of the PSET. Use the energies in the PSET and assume that fit factors for each ESA step align.

Double check that this approach will also work for Hi.

PR #2474, opened 2025-12-02
2185 hi lo l2   vectorize predictor corrector

### LINKED ISSUE #2185: Hi Lo L2 - Vectorize predictor corrector
> I know you're going to hate me for saying this, but I am curious if this could be vectorized out as well... (Claude is quite good at these kinds of updates later too, so don't feel like you need to do it now)
> 
> 
> 
> You could keep iterating until the largest pixel error was below your threshold, or 20 iterations, but I don't think it matters if you over-converge. But that loop over 20 is much smaller than this exterior loop over 3600 / 1800. 

 _Originally posted by @greglucas in [#2171](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2171/files#r2320397772)_

Claude made this vectorization quite simple. Use it!

tmplummer:
# Change Summary

## Overview
This work vectorizes the flux correction. I checked that a Lo L2 CDF generated before and after the changes had identical ena and background intensities and intensity statistical uncertainty so I am confident that this provides the exact same results as before.

This also changes the input to require that xr.DataArrays be passed in and handles the stacking of multiple spatial dimensions which streamlines the Lo L2 code for applying the flux correction.

According to the logs, the flux correction for Lo went from taking ~6 seconds down to <1 second.

Closes: #2185 #2471


imap_processing/ena_maps/utils/corrections.py
```diff
@@ -250,76 +285,150 @@ def predictor_corrector_iteration(
         # Initial source flux estimate
         eta_initial = self.eta_esa(energy_levels, gamma_initial)
         source_fluxes_n = observed_fluxes / eta_initial
+        source_uncertainties = observed_uncertainties / eta_initial
+
+        # Track which pixels have converged and iteration count per pixel
+        converged = np.zeros(observed_fluxes.shape[1:], dtype=bool)
```
greglucas, line 287: Do we want/need to track this on a per-pixel basis? It seems like we could just keep iterating until all pixels met the threshold and if we are converged we would just return the same number at that pixel every time we iterated past convergence.

greglucas (commented):
Nice, looks like a lot of book-keeping, but still seems pretty readable to me.

Just one question on whether we want/need to track indivdiual pixel convergence or if we can leave out that detail and just keep converging until all pixels have converged to not keep track of that book-keeping step as well.

imap_processing/ena_maps/utils/corrections.py
```diff
@@ -213,20 +244,23 @@ def predictor_corrector_iteration(
...
         Implements the algorithm from Appendix A of the Mapping Algorithm Document.
+        Fully vectorized to process all spatial pixels simultaneously, with
+        per-pixel convergence tracking.
 
         Parameters
         ----------
         observed_fluxes : np.ndarray
-            Array of observed fluxes.
+            Array of observed fluxes. Shape (n_energy,) or
+            (n_energy, ...spatial_dims...).
```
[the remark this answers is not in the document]
tmplummer: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2474#discussion_r2582948568)

imap_processing/ena_maps/utils/corrections.py
tmplummer: Yeah, valid question. I was just trying to preserve the exact functionality that was already implemented so that I could definitively say that the algorithm produced exact results.

I am happy with either implementation. If we iterate until all pixels converge, it might have very slight consequences for validation.

greglucas: OK, I'll leave this to you then. I am also fine with either implementation and having this for validation purposes makes some good sense.

greglucas (approved):

PR #2479, opened 2025-12-03
ULTRA l1c scattering culling

lacoak21:
# Change Summary

## Overview
In ultra l1c we need to explicitly NOT reject pixels based on the FWHM theta and phi values. The ancillary file that controls the FWHM thresholds are currently all set to np.inf so this should not be skipping any pixels but I discovered today that some were still being skipped because the fwhm values were invalid. 

## Updated Files
   - Do not skip pixels based off of their FWHM value

### DIFF
--- a/imap_processing/ultra/l1c/l1c_lookup_utils.py
+++ b/imap_processing/ultra/l1c/l1c_lookup_utils.py
@@ -169,14 +169,12 @@ def calculate_fwhm_spun_scattering(
 
         for energy_idx in range(len(energy_bin_geometric_means)):
             if reject_scattering:
-                # If reject_scattering is True, only keep pixels below the scattering
-                # threshold
-                scattering_valid_pixels = scattering_mask[:, energy_idx]
-                valid_pixels = for_pixel_indices[scattering_valid_pixels]
+                valid_pixels = scattering_mask[:, energy_idx]
+                pixels_below_scattering_for_energy.append(
+                    for_pixel_indices[valid_pixels]
+                )
             else:
-                valid_pixels = for_pixel_indices
-
-            pixels_below_scattering_for_energy.append(valid_pixels)
+                pixels_below_scattering_for_energy.append(for_pixel_indices)
 
         pixels_below_scattering.append(pixels_below_scattering_for_energy)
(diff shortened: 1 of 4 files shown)


greglucas (commented):
Should this be adding in a check for `np.isfinite()` within the code that sets the scattering mask itself?

lacoak21:
> Should this be adding in a check for `np.isfinite()` within the code that sets the scattering mask itself?

I was thinking about doing that but it didnt make sense to me to have nan values "passing" the scattering check.

lacoak21:
> I was thinking about doing that but it didnt make sense to me to have nan values "passing" the scattering check.

Im talking to Nick about why we even have nan values in the first place. Seems like they are being introduced from an ancillary file so this issue may be resolved with an updated file.

imap_processing/ultra/l1c/l1c_lookup_utils.py
```diff
@@ -165,8 +168,13 @@ def calculate_fwhm_spun_scattering(
...
         for energy_idx in range(len(energy_bin_geometric_means)):
-            valid_pixels = scattering_mask[:, energy_idx]
-            pixels_below_scattering_for_energy.append(for_pixel_indices[valid_pixels])
+            if reject_scattering:
+                valid_pixels = scattering_mask[:, energy_idx]
+                pixels_below_scattering_for_energy.append(
+                    for_pixel_indices[valid_pixels]
+                )
+            else:
+                pixels_below_scattering_for_energy.append(for_pixel_indices)
```
greglucas: Could this be brought outside the loop in this case? This is all a little confusing to me what is going on in these various loops and subsetting of scattering and then by energy, so I'm just a little unclear whether this function could be restructured a little easier with numpy broadcasting lookups too.

```python
if not reject_scattering:
    pixels_below_scattering_for_energy = np.repeat(for_pixel_indices, len(energy_bin_geometric_means))
else:
    loop...
```

lacoak21: OH! Yes good point.

lacoak21:
Im closing this PR. It can be simplified much further.

PR #2480, opened 2025-12-04
I-ALiRT -  Codice L2

laspsandoval:
# Change Summary

## Updated Files
- process_codice.py
   - Puts together all levels of processing and creates L2 I-ALiRT data product for CoDICE-HI
- external_test_data_config.py
   -Added lookup and validation tables
- constants.py
   -Added constant number of SSDs per group for I-ALiRT
- codice_l1a_ialirt_hi.py
   - Added parameters that was required later for l2: energy_minus, energy_plus
   - Removed for loop since we only have a single species for Hi: "h"

## Testing
- test_process_codice.py

### DIFF
--- a/imap_processing/codice/codice_l1a_ialirt_hi.py
+++ b/imap_processing/codice/codice_l1a_ialirt_hi.py
@@ -72,6 +72,8 @@ def l1a_ialirt_hi(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
 
     species_data = sci_lut_data["data_product_hi_tab"]["0"]["ialirt"]
     species_names = species_data.keys()
+    first_species = next(iter(species_data))
+    centers, energy_minus, energy_plus = get_energy_info(species_data[first_species])
 
     compression_algorithm = constants.HI_COMPRESSION_ID_LOOKUP[view_tab_obj.view_id]
 
@@ -118,6 +120,14 @@ def l1a_ialirt_hi(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
             "epoch_delta_plus": xr.DataArray(
                 repeated_deltas,
                 dims=("epoch",),
+            ),
+            f"energy_{first_species}_minus": xr.DataArray(
+                energy_minus,
+                dims=(f"energy_{first_species}",),
+            ),
+            f"energy_{first_species}_plus": xr.DataArray(
+                energy_minus,
+                dims=(f"energy_{first_species}",),
             ),
         },
     )
--- a/imap_processing/codice/codice_l1a_ialirt_hi.py
+++ b/imap_processing/codice/codice_l1a_ialirt_hi.py
@@ -9,7 +9,6 @@ import xarray as xr
 from imap_processing.codice import constants
 from imap_processing.codice.decompress import decompress
 from imap_processing.codice.utils import (
-    CODICEAPID,
     ViewTabInfo,
     get_codice_epoch_time,
     get_collapse_pattern_shape,
@@ -61,14 +60,6 @@ def l1a_ialirt_hi(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
         three_d_collapsed=view_tab_info["3d_collapse"],
         collapse_table=view_tab_info["collapse_table"],
     )
-
-    if view_tab_obj.sensor != 1:
-        raise ValueError("Unsupported sensor ID for Hi processing.")
-
-    if view_tab_obj.apid != CODICEAPID.COD_HI_IAL:
-        raise ValueError(
-            f"Unknown apid {view_tab_obj.apid} in I-ALiRT omni processing."
-        )
 
     species_data = sci_lut_data["data_product_hi_tab"]["0"]["ialirt"]
     first_species = next(iter(species_data))
--- a/imap_processing/codice/codice_l1a_ialirt_hi.py
+++ b/imap_processing/codice/codice_l1a_ialirt_hi.py
@@ -147,13 +147,13 @@ def l1a_ialirt_hi(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
     )
 
     # Number of energy chunks.
-    species_chunk_sizes = [len(species_data[first_species]["min_energy"])]
+    num_energy_chunk = [len(species_data[first_species]["min_energy"])]
 
     l1a_dataset = l1a_dataset.assign_coords(
         {f"energy_{first_species}": (f"energy_{first_species}", np.array(centers))}
     )
 
-    chunk_size = species_chunk_sizes[0]
+    chunk_size = num_energy_chunk[0]
 
     # This is rearranging data from (epoch, energy, n_spins, spin_sector, inst_az)
     # -> (epoch, n_spins, energy, spin_sector, inst_az) ->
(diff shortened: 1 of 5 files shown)


imap_processing/codice/codice_l1a_ialirt_hi.py
```diff
@@ -71,7 +71,8 @@ def l1a_ialirt_hi(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
         )
 
     species_data = sci_lut_data["data_product_hi_tab"]["0"]["ialirt"]
-    species_names = species_data.keys()
+    first_species = next(iter(species_data))
+    centers, energy_minus, energy_plus = get_energy_info(species_data[first_species])
```
greglucas, line 66: Why only one species?

imap_processing/codice/codice_l1a_ialirt_hi.py
```diff
@@ -137,47 +146,34 @@ def l1a_ialirt_hi(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
         *collapse_shape,
     )
 
-    species_chunk_sizes = [
-        len(species_data[species]["min_energy"]) for species in species_names
-    ]
-
-    start_idx = 0
+    species_chunk_sizes = [len(species_data[first_species]["min_energy"])]
```
greglucas: Looking at the spreadsheet and algorithm document, aren't all chunk sizes 24-bit counters for I-ALiRT?

imap_processing/codice/codice_l1a_ialirt_hi.py
```diff
@@ -137,47 +146,34 @@ def l1a_ialirt_hi(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
...
-                    dims=(f"energy_{species_name}",),
-                )
-            }
-        )
+    chunk_size = species_chunk_sizes[0]
+    end_idx = chunk_size * n_spins
 
-        chunk_size = species_chunk_sizes[index]
-        end_idx = start_idx + chunk_size * n_spins
+    species_array = decompressed_data[:, 0:end_idx]
```
greglucas: Do you need to make sure you are starting at the beginning of a sequence here? I think decompressed_data is for every ialirt packet isn't it, but CoDICE is combining a bunch of packets together to get their full 3456 byte buffer. I am wondering if this should somehow be something like a 1d array that has been concatenated together so that you have a long byte chunk here.

imap_processing/codice/constants.py
```diff
@@ -1059,6 +1059,7 @@
 PUI_POSITIONS = SW_POSITIONS
 L2_GEOMETRIC_FACTOR = 0.013
 L2_HI_NUMBER_OF_SSD = 12.0
+IALIRT_HI_NUMBER_OF_SSD_PER_GROUP = 3.0
```
greglucas, line 1075: This should be 4 according to the algorithm document. It should also be an integer and not a float I think.
image

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -182,23 +188,88 @@ def create_xarray_dataset(
...
+        cod_hi_l1b_data[f"energy_{species}_plus"]
+        + cod_hi_l1b_data[f"energy_{species}_minus"]
+    ).values[np.newaxis, :]
+
+    # For omni over 3 SSDs:
+    g_g = constants.L2_GEOMETRIC_FACTOR * constants.IALIRT_HI_NUMBER_OF_SSD_PER_GROUP
+
+    # Now give it a column dimension for broadcasting
+    energy_passbands = np.asarray(energy_passbands).squeeze()
+    energy_passbands = energy_passbands[:, None]
```
greglucas: Pull the g_g up above or below so it isn't breaking up the energy passbands variable calculations.

You are adding a newaxis to a numpy array, then calling asarray (it already is one), then squeezing it to get rid of the new axis, then adding one at the end... This is confusing to follow. I think it is could be:

```python
(
        cod_hi_l1b_data[f"energy_{species}_plus"]
        + cod_hi_l1b_data[f"energy_{species}_minus"]
    ).values[:, np.newaxis]
```

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -182,23 +188,88 @@ def create_xarray_dataset(
...
+
+    denom = g_g * eps_ig * energy_passbands  # (15, 4)
```
greglucas, line 232: The equations in the algorithm document have this as `g_m`, is there a reason to use `g_g` instead?

imap_processing/codice/codice_l1a_ialirt_hi.py
laspsandoval: Hi only have species "h" for I-ALiRT

imap_processing/codice/constants.py
laspsandoval: Look at the table right before that. There are 3 SSDs in a group. And four SSD groups. So 12 in all.

imap_processing/ialirt/l0/process_codice.py
laspsandoval: I see: Gg⋅𝜖𝜖𝑖 𝑖𝑖⋅(Δ𝐸𝐸𝑖 ) on page 64

imap_processing/codice/codice_l1a_ialirt_hi.py
laspsandoval: This is after the first part of the decompression has happened. This is just selecting the indices for each energy bin. But since Hi only has a single energy bin then I actually don't even think that I need this. Removing it now.

imap_processing/codice/codice_l1a_ialirt_hi.py
laspsandoval: This is just rearranging the already decompressed data. It isn't decompressing it. And this part is not necessary once I looked at it again. I just removed it.

imap_processing/codice/codice_l1a_ialirt_hi.py
```diff
@@ -137,47 +146,32 @@ def l1a_ialirt_hi(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
...
-        l1a_dataset = l1a_dataset.assign_coords(
-            {
-                f"energy_{species_name}": xr.DataArray(
-                    np.array(centers),
-                    dims=(f"energy_{species_name}",),
-                )
-            }
-        )
+    # Number of energy chunks.
+    species_chunk_sizes = [len(species_data[first_species]["min_energy"])]
```
lacoak21: Super small nit but the var name made me thing it was supposed to be the delta or something not just number of energy chunks

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -209,62 +280,84 @@ def process_codice(
...
+            # Put in Decimal format so DynamoDB can read it.
+            dec_l2_hi = np.vectorize(lambda x: Decimal(f"{float(x):.3f}"))(
+                l2_hi
+            ).tolist()
+
+            codice_hi_data.append(
+                {
+                    "apid": 478,
+                    "met": int(met[0]),
+                    "met_in_utc": met_to_utc(met[0]).split(".")[0],
```
lacoak21, line 355: Why so many times? Is this what will go into the dynamo db?

lacoak21 (approved):

imap_processing/ialirt/l0/process_codice.py
laspsandoval: Yes. This is the format for DynamoDB. It will be a matrix of dimensions (4, 15, 4, 4).

PR #2483, opened 2025-12-04
UTLRA l1c refactor and validation test 

lacoak21:
# Change Summary

## Overview
Slight refactor in regards to the spun valid pixels. I made this into an xarray dataset which simplified other functions. 
Add validation test to make sure we match with ULTRA data. 

## Updated Files
- imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
   - Fix test. 
   - DO not apply boundary scale factors
  - simplify valid pixels if we are not culling pixels by FWHM values 
   - simplify exposure factor function and eff and gf calculation

## Testing
Add validation test for checking sensitivity and exposure factor

### DIFF
--- a/imap_processing/tests/external_test_data_config.py
+++ b/imap_processing/tests/external_test_data_config.py
@@ -183,18 +183,11 @@ EXTERNAL_TEST_DATA = [
     ("imap_ultra_l1c-90sensor-sc-pointing-phi-test_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1c-90sensor-sc-pointing-index-test_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1c-90sensor-sc-pointing-bsf-test_20250101_v000.csv", "ultra/data/l1/"),
-    ("imap_ultra_l1c-90sensor-sc-pointing-theta_20250101_v001.csv", "ultra/data/l1/"),
-    ("imap_ultra_l1c-90sensor-sc-pointing-phi_20250101_v001.csv", "ultra/data/l1/"),
-    ("imap_ultra_l1c-90sensor-sc-pointing-index_20250101_v001.csv", "ultra/data/l1/"),
-    ("imap_ultra_l1c-90sensor-sc-pointing-bsf_20250101_v001.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b-scattering-thresholds-per-energy_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b_45sensor-de_20240207-repoint99999_v999.cdf","ultra/data/l1/"),
     ("imap_ultra_l1c-45sensor-nominal-for-lookup_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1c-45sensor-static-dead-times_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1c-90sensor-static-dead-times_20250101_v000.csv", "ultra/data/l1/"),
-    ("Exposures-IMAP_ULTRA_90-IMAP_DPS-SC-nside32-ebin0.csv", "ultra/data/l1/"),
-    ("SENS-IMAP_ULTRA_90-IMAP_DPS-SC-nside32-ebin0.csv", "ultra/data/l1/"),
-    ("imap_ultra_l1b_45sensor-de_20000101-repoint00000_v000.cdf", "ultra/data/l1/"),
 
     # l2
     ("imap_ultra_l2-energy-bin-group-sizes_20250101_v000.csv", "ultra/data/l2/"),
--- a/imap_processing/tests/external_test_data_config.py
+++ b/imap_processing/tests/external_test_data_config.py
@@ -183,11 +183,19 @@ EXTERNAL_TEST_DATA = [
     ("imap_ultra_l1c-90sensor-sc-pointing-phi-test_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1c-90sensor-sc-pointing-index-test_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1c-90sensor-sc-pointing-bsf-test_20250101_v000.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1c-90sensor-sc-pointing-theta_20250101_v001.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1c-90sensor-sc-pointing-phi_20250101_v001.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1c-90sensor-sc-pointing-index_20250101_v001.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1c-90sensor-sc-pointing-bsf_20250101_v001.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b-scattering-thresholds-per-energy_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1b_45sensor-de_20240207-repoint99999_v999.cdf","ultra/data/l1/"),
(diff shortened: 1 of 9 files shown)


imap_processing/tests/external_test_data_config.py
```diff
@@ -183,11 +183,19 @@
     ("imap_ultra_l1c-90sensor-sc-pointing-phi-test_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1c-90sensor-sc-pointing-index-test_20250101_v000.csv", "ultra/data/l1/"),
     ("imap_ultra_l1c-90sensor-sc-pointing-bsf-test_20250101_v000.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1c-90sensor-sc-pointing-theta_20250101_v001.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1c-90sensor-sc-pointing-phi_20250101_v001.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1c-90sensor-sc-pointing-index_20250101_v001.csv", "ultra/data/l1/"),
+    ("imap_ultra_l1c-90sensor-sc-pointing-bsf_20250101_v001.csv", "ultra/data/l1/"),
```
greglucas: Above this there are `-test` files with these same names. Do you want to replace those or remove those now?

imap_processing/tests/ultra/unit/test_ultra_l1c_pset_bins.py
```diff
@@ -276,18 +276,65 @@ def test_apply_deadtime_correction(imap_ena_sim_metakernel, ancillary_files):
...
+
+
+@pytest.mark.external_kernel
+def test_apply_deadtime_correction_energy_dep(imap_ena_sim_metakernel, ancillary_files):
+    """Tests apply_deadtime_correction function when scattering rejection is on."""
+    nside = 8
+    pix = hp.nside2npix(nside)
+    steps = 500  # Reduced for testing
+    mock_theta = np.random.uniform(-60, 60, (pix, steps))
+    mock_phi = np.random.uniform(-60, 60, (pix, steps))
```
greglucas, line 315: We should be cautious about using random data in tests. Where is this used and does it even matter? We should either seed the random number generator above this, or use a linspace or similar function to sample the values in a reproducible way.

imap_processing/ultra/l1c/l1c_lookup_utils.py
```diff
@@ -243,6 +250,9 @@ def get_spacecraft_pointing_lookup_tables(
     for_indices_by_spin_phase = np.nan_to_num(index_grid[:, 2:], nan=0).astype(
         bool
     )  # Shape (npix, 15000)
+    boundary_scale_factors = xr.DataArray(
+        boundary_scale_factors, dims=("pixel", "spin_phase_step")
+    )
```
greglucas, line 255: There is a `to_xarray()` method on pandas dataframes too. But, that might actually be even more confusing, so just putting this out there in case you'd find that easier. This seems good too though.
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_xarray.html

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -359,74 +359,69 @@ def get_deadtime_ratios_by_spin_phase(
...
-
+    # and below the scattering threshold (if scattering rejection is on).
+    # Sum over the first dim of valid_spun_pixels is the spin phase step.
+    # This is like spinning the spacecraft by nominal 1 ms steps in the despun frame.
+    if apply_bsf:
+        counts = (valid_spun_pixels * deadtime_ratios * boundary_scale_factors).sum(
+            dim="spin_phase_step"
+        )
+    else:
+        counts = (valid_spun_pixels * deadtime_ratios).sum(dim="spin_phase_step")
```
greglucas: My personal preference is to extract the final sum since you are doing that over both of these variables. But, either way is fine and you can ignore this if you don't like it.
```suggestion
    all_counts = valid_spun_pixels * deadtime_ratios
    if apply_bsf:
        all_counts *= boundary_scale_factors)
    counts = all_counts.sum(dim="spin_phase_step")
```

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -359,74 +359,69 @@ def get_deadtime_ratios_by_spin_phase(
...
     # Multiply by the nominal spin step to get the exposure time in ms
-    exposure_pointing = counts * nominal_ms_step
+    exposure_pointing = counts.values * nominal_ms_step
```
greglucas: Do you not want to return the xarray dataarray here like you do in the other functions?

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -502,16 +501,24 @@ def get_spacecraft_exposure_times(
...
     # Adjust exposure time by the actual number of valid spins in the pointing
     exposure_pointing_adjusted = n_spins_in_pointing * exposure_time
-    return exposure_pointing_adjusted, nominal_deadtime_ratios
+    # Ensure exposure factor is broadcast correctly
+    if exposure_pointing_adjusted.shape[0] != n_energy_bins:
+        exposure_pointing_adjusted = np.repeat(
+            exposure_pointing_adjusted,
+            n_energy_bins,
+            axis=0,
+        )
```
greglucas, line 509: This is where it seems like it'd be nice to have a dataarray that has the proper dimensions associated with it so that the broadcasting "just works"

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -592,27 +604,38 @@ def get_efficiencies_and_geometric_function(
...
+        valid_at_spin = valid_spun_pixels.isel(
+            spin_phase_step=i
+        )  # shape: (energy, pixel)
+
         for energy_bin_idx in range(energy_bins):
-            pixel_inds = pixels_at_spin[energy_bin_idx]
+            # Determine pixel indices based on energy dependence
+            if valid_at_spin.sizes["energy"] == 1:
+                # No scattering rejection. Same pixels for all energies
+                pixel_inds = np.where(valid_at_spin.isel(energy=0))[0]
```
greglucas, line 616: Just an FYI that doing all of these `isel` calls within a hot loop was causing quite a bit of performance issues on Lo, so it might be worth looking at this later to try and avoid them or do them at the highest level if possible.

imap_processing/tests/external_test_data_config.py
lacoak21: I would like to keep them for the other tests. I added these real ones for the validation test but they are large and would cause the other tests to slow down significantly.

greglucas: Do the files still get downloaded by everyone because this is a generic external downloads? Or is each individual test requesting a single external download file so this doesn't impact everyone with the file sizes?

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
lacoak21: I broadcasted it here because this is the final step that uses the exposure_pointing and the next step is to pass it into the create_dataset function that expects numpy arrays. Maybe in another refactor I can update that function to accept xarrays already.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
lacoak21: OK good to know Ill add a todo here.

imap_processing/tests/external_test_data_config.py
lacoak21: I think they would be downloaded by everyone? Do you know how to configure the test to request the file? Ill go look into that now because that would be ideal.

greglucas: I think you'd request the download function explicitly from within the test itself. Since you are skipping it automatically already, that should work and not need the external test mark even.

greglucas (approved):

PR #2485, opened 2025-12-05
SWAPI: Fix I-ALiRT error resulting from new sweep table number

hafarooki:
# Change Summary

## Overview
I was shown in an email that I-ALiRT is broken because this portion of the code is failing to handle NaNs. I also found that an interval where the same solution kept being returned was having that issue because the zeros were not being excluded from `curve_fit`. I am working on another potential PR to improve this fitting procedure, but this PR should fix those particular bugs. However, I haven't tested it properly (I did validate that the code runs with a custom unit test but the test is not well-written and not included in the PR) and would like help doing that. **Also, there is a potential issue with how I did it: If ALL of the counts are excluded by the mask, there will be NO samples passed to curve fit.**

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
N/A

## Updated Files
Just updated `process_swapi.py` to exclude zeros (as otherwise the initial guess is left as the solution and the covariance is infinite) and nan/infinite points (as otherwise there is an error that breaks I-ALiRT)

- updated file 1
- updated file 2

## Testing
I think there should be a test or two added to ensure the new behavior is correct. I haven't gotten around to creating such tests yet as I am not sure how the software team would prefer them to be implemented and I understand that this is urgent..


hafarooki:
@laspsandoval

hafarooki:
Here is the test I wrote that shows that the code does at least run (I guess you can also just use the existing tests with simulated data...):
```python
def test_process_real_packet():
    from pathlib import Path
    packets_directory = Path('/home/hafarooki/projects/imap_processing/swapi-test/packets')
    
    xtce_ialirt_path = (
        imap_module_directory / "ialirt" / "packet_definitions" / "ialirt.xml"
    )
    
    calibration_file = pd.read_csv(
        f"{imap_module_directory}/tests/ialirt/data/l0/swapi_ialirt_energy_steps.csv"
    )

    lines = []

    for packet_path in packets_directory.iterdir():    
        sc_xarray_data = packet_file_to_datasets(
            packet_path, xtce_ialirt_path, use_derived_value=False
        )[478]
        lines.append(packet_path.name)
        
        lines.append('\tswapi_coin_cnt# sum: ' + ' '.join(f'{str(key).split("_")[-1]}:{sc_xarray_data[key].sum().item()}'
                            for key in sc_xarray_data.keys()
                            if str(key).startswith('swapi_coin_cnt')))
        lines.append('\ttotal count: ' + str(sum(sc_xarray_data[key].sum().item()
                                        for key in sc_xarray_data.keys()
                                        if str(key).startswith('swapi_coin_cnt'))))
    
        swapi_product = process_swapi_ialirt(sc_xarray_data, calibration_file)
        for i in range(len(swapi_product)):
            density = swapi_product[i]['swapi_pseudo_proton_density']
            temperature = swapi_product[i]['swapi_pseudo_proton_temperature']
            speed = swapi_product[i]['swapi_pseudo_proton_speed']
            lines.append(f'\ti={i} ({swapi_product[i]["met_in_utc"]})')
            lines.append(f'\t\tpseudo moments: density={density}, temperature={temperature}, speed={speed}')

    output_file = packets_directory.parent / 'output.txt'

    output_file.write_text('\n'.join(lines))
    
    assert False, 'need to add assertions comparing with omni'
    ```

File: imap_processing/ialirt/l0/process_swapi.py [hunk 1]
```diff
@@ -111,18 +111,19 @@ def optimize_pseudo_parameters(
                 60000 * (initial_speed_guess / 400) ** 2,
             ]
         )
+        five_point_range = range(max_index - 2, max_index + 2 + 1)
```
laspsandoval: Can we keep the range the same as before?

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -111,18 +111,19 @@ def optimize_pseudo_parameters(
...
+        xdata = energy_passbands.take(five_point_range, mode="clip")
+        ydata = current_sweep_count_rates.take(five_point_range, mode="clip")
+        sigma = current_sweep_count_rate_errors.take(five_point_range, mode="clip")
+        mask = (ydata > 0) & (sigma > 0) & np.isfinite(xdata) & np.isfinite(ydata) & np.isfinite(sigma)
```
laspsandoval: Does it matter if xdata is greater than 0? Should it be >= 0?

laspsandoval (commented):
Take a look at my comments.

imap_processing/ialirt/l0/process_swapi.py
hafarooki: The previous range was incorrect -- I should have put this in a separate PR. It was using 6 points instead of 5 points.

imap_processing/ialirt/l0/process_swapi.py
hafarooki: xdata is energy, right? physically, energy must always be >= 0, but I am not sure about in the context of SWAPI. It is probably not necessary to put a check in for that

imap_processing/ialirt/l0/process_swapi.py
hafarooki: I included this here by accident -- should I put this in a separate PR or just update the name/description of this PR?

laspsandoval: This might take more time if we change the range. Bishwas will need to review and we will need to fix the test. Could we leave it for now and update it in the next PR?

laspsandoval: Change to this:

five_point_range = range(max_index - 3, max_index + 3)

Bishwasls: Its okay to leave the energy range like this now.

Bishwasls (commented):
Its not critical to change the energy bins used for fitting.

imap_processing/ialirt/l0/process_swapi.py
laspsandoval: Ok. We will need to update the test that is failing:
test_optimize_parameters

hafarooki: In Figure 4 of Lee et al. 2025, the I-ALiRT paper, it looks like they actually used 6 points to fit.
https://link.springer.com/article/10.1007/s11214-025-01244-9

hafarooki:
I reverted to using 6 points for fitting. Locally, I confirmed that the failing test once again passes.

hafarooki:
One thing that should be discussed is how we handle cases where less than 3 points are available for fitting. With less than 3 points, the 3-parameter Maxwellian model cannot be constrained. Some sort of fill value should be used instead?

bishwassth (approved):
This version looks good to me.

bishwassth:
> One thing that should be discussed is how we handle cases where less than 3 points are available for fitting. With less than 3 points, the 3-parameter Maxwellian model cannot be constrained. Some sort of fill value should be used instead?

That's a good point. Did you see any case where less than three points are available? I always thought we will have at least three points available for fitting, but may be not due to compression of the counts in the I-ALiRT packets.

hafarooki:
So far I have seen a case with only 3 points available but not one with less than 3. However would be good to include the edge case in case it ever happens so that the I-ALiRT service is not interrupted again.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -111,18 +111,19 @@ def optimize_pseudo_parameters(
                 60000 * (initial_speed_guess / 400) ** 2,
             ]
         )
+        fitting_point_range = range(max_index - 3, max_index + 3)
+        xdata = energy_passbands.take(fitting_point_range, mode="clip")
+        ydata = current_sweep_count_rates.take(fitting_point_range, mode="clip")
+        sigma = current_sweep_count_rate_errors.take(fitting_point_range, mode="clip")
+        mask = (ydata > 0) & (sigma > 0) & np.isfinite(xdata) & np.isfinite(ydata) & np.isfinite(sigma)  
+        assert np.count_nonzero(mask) > 0
```
greglucas: Don't put `assert` in the actual code here because we can hit this path (you have a failing test with this exact failure now), we should check for it and raise or return nans or whatever you expect to happen if this is the case.

Do you want this to be greater than 2 so your curve fit has some values to fit per your discussion?

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -111,18 +111,19 @@ def optimize_pseudo_parameters(
...
-                range(max_index - 3, max_index + 3), mode="clip"
-            ),
-            sigma=current_sweep_count_rate_errors.take(
-                range(max_index - 3, max_index + 3), mode="clip"
-            ),
+            xdata=xdata[mask],
+            ydata=ydata[mask],
+            sigma=sigma[mask],
             p0=initial_param_guess,
+            full_output=True
```
greglucas: You aren't using the infodict or anything else returned after this, so we don't need the full_output.
```suggestion
```

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -111,18 +111,19 @@ def optimize_pseudo_parameters(
                 60000 * (initial_speed_guess / 400) ** 2,
             ]
         )
+        fitting_point_range = range(max_index - 3, max_index + 3)
+        xdata = energy_passbands.take(fitting_point_range, mode="clip")
+        ydata = current_sweep_count_rates.take(fitting_point_range, mode="clip")
+        sigma = current_sweep_count_rate_errors.take(fitting_point_range, mode="clip")
+        mask = (ydata > 0) & (sigma > 0) & np.isfinite(xdata) & np.isfinite(ydata) & np.isfinite(sigma)
```
greglucas: Can we call this `valid_data` or something similar. `mask` is typically the "hidden" values in my quick skims, so we should prefix it with something else to indicate this is the selection you are interested in.

This line is too long currently and failing our code quality checks in the `precommit` step.

greglucas (commented):
Do you have some example tests so that we can add some assertions on what you want to happen in the various cases.

Something like the following pseudo code to test just this curve fitting function.

```
x = np.arange(63)
# Add a nan to our x that should be ignored in the fits
x[50] = np.nan
output = optimize_pseudo_parameters(x, ...)
assert output ...
```

greglucas:
Here is a link to the failing test you should be able to run locally:
https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/19998051126/job/57349485125?pr=2485#step:7:2712

imap_processing/ialirt/l0/process_swapi.py
hafarooki: I think returning nan is a reasonable alternative. @Bishwasls can chime in. But would there be any errors further down the pipeline if nans are used? Or is the code already designed to accept nans?

imap_processing/ialirt/l0/process_swapi.py
hafarooki: Good idea!

imap_processing/ialirt/l0/process_swapi.py
greglucas: It looks like it will accept NaNs and pass those through, but I'm not sure what that means for returns to the website and such. But, this seems like a good thing to do so you know you tried to process data at that time and the values were just bad, not that processing failed.

It looks like we return/store it as a Decimal, which does pass NaN through.
```python
>>> Decimal(f"{np.nan:.3f}")
Decimal('NaN')
```

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/3fe84244317bd4970f45f9a24d83cc00c83406b8/imap_processing/ialirt/l0/process_swapi.py#L224-L248

hafarooki: Updating it to provide nans

hafarooki:
Something strange I just realized: Why is there even a nan/infinite value in the xdata (as the stack trace in the original email that raised this issue shows) in the first place?

In `process_swapi.py`, there is this bit of code:
```python
    # Find the sweep's energy data for the latest time, where sweep_id == 2
    subset = calibration_lut_table[
        (calibration_lut_table["timestamp"] == calibration_lut_table["timestamp"].max())
        & (calibration_lut_table["Sweep #"] == 2)
    ]
    if subset.empty:
        energy_passbands = np.full(NUM_IALIRT_ENERGY_STEPS, np.nan, dtype=np.float64)
    else:
        subset = subset.sort_values(["timestamp", "ESA Step #"])
        energy_passbands = (
            subset["Energy"][:NUM_IALIRT_ENERGY_STEPS].to_numpy().astype(float)
        )

```
Note in particular that when the data subset is empty, nans are provided as energy.

Is this appropriate behavior? Maybe instead of masking out nans in `optimize_pseudo_parameters`, we should not give it nans in the first place.

I am not sure what that line of code is meant to be doing exactly---under what circumstances would `subset` be empty? Clearly, this did end up happening in the real application. I think this implies that `calibration_lut_table` does not contain any entries for its last timestamp at `Sweep # = 2`. But I do not know under what circumstances it would be expected to happen.

greglucas:
> Something strange I just realized: Why is there even a nan/infinite value in the xdata (as the stack trace in the original email that raised this issue shows) in the first place?

> Note in particular that when the data subset is empty, nans are provided as energy.

> Is this appropriate behavior? Maybe instead of masking out nans in optimize_pseudo_parameters, we should not give it nans in the first place.

@hafarooki I agree with you here. If that is the case, we should probably just exit/return early because there is no point in even continuing on after that subset.empty call.

Another thing that is slightly confusing to me and might help is to make `optimize_pseudo_parameters()` function only take in one sweep at a time, it doesn't need to do any looping internally, then it returns only the 3 values it found. We move the loop over sweeps outside of this and populate the dictionary of return values during the loop in the main function rather than the internal function.

greglucas:
Why do we hardcode sweep # == 2? We have actually changed to sweep 3 now, so is there a way for I-ALiRT to know which sweep # to lookup or do we have to hardcode 3 now and make sure we are in-sync manually.

hafarooki:
Good catch! After consulting with the engineering team, I concluded that we can use `swapi_version` from the packet file to determine which table to use. For now, I just used the first one in the packet, but there are technically 60, one for each second, so we may have to account for potentially multiple in one file. For now this should be good enough

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -211,7 +222,7 @@ def process_swapi_ialirt(
     # Find the sweep's energy data for the latest time, where sweep_id == 2
     subset = calibration_lut_table[
         (calibration_lut_table["timestamp"] == calibration_lut_table["timestamp"].max())
-        & (calibration_lut_table["Sweep #"] == 2)
+        & (calibration_lut_table["Sweep #"] == sweep_table_version)
```
greglucas: I think this update is right, but now I've opened the can of worms and I don't think the other selection piece is correct... See here for what we are doing in L2 processing

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/7c03578253f16c33dac4a37006eee9bf307d121c/imap_processing/swapi/l2/swapi_l2.py#L72-L74

I think we should drop the timestamp comparison here. Then in the energy passbands below we need to get the **latest** energy passbands from the table. So we need to do a reverse index lookup to get the final 72 energy steps and then only select the first 63 from that point [-72:-9]

```
        energy_passbands = (
            subset["Energy"][-72:-72 + NUM_IALIRT_ENERGY_STEPS].to_numpy().astype(float)
        )

Comment (hafarooki) [reply imap_processing/ialirt/l0/process_swapi.py]: Could you suggest a unit test for this

Comment (greglucas) [reply imap_processing/ialirt/l0/process_swapi.py]: I'd suggest modifying the current LUT we have in the tests, similar to this PR: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2484

**Comment by laspsandoval** [comment]
pre-commit.ci autofix

File: imap_processing/ialirt/l0/process_swapi.py [hunk 7]
```diff
@@ -208,41 +216,38 @@ def process_swapi_ialirt(
...
-        )
-
-    solution = optimize_pseudo_parameters(
-        raw_coin_rate, count_rate_error, energy_passbands
-    )
-
     swapi_data = []
 
-    for entry in np.arange(0, len(solution["pseudo_speed"])):
+    # TODO why is it that len(raw_coin_rate) != len(met_values)
```
hafarooki: @greglucas In one of the test cases, `len(met_values)` was 90 but `len(raw_coin_rate)` was 2. Could you please clarify this?

imap_processing/ialirt/l0/process_swapi.py
hafarooki: Resolved by the latest commit

imap_processing/ialirt/l0/process_swapi.py
hafarooki: resolved -- this test had a fake packet. now that we have new tests with real packets, we decided to just remove that test, as the differing shape of the fake packet also made the test very slow

greglucas:
I'm sorry, but I'm not going to review this. You have touched too many things since the last time. Please keep updates minimal!

greglucas (changes requested):
This is touching L2 code now too. Please remove that

hafarooki:
I agree, the scope of this PR has changed because the issue turned out to be different from the original one, and we have identified some other issues along the way. I will split this into multiple PRs once we finalize the changes to be made.

ejzirn:
A lot has happened today for the SWAPI team in terms of how to process the I-ALiRT packets to get our pseudo moments. Because of this, I think we should pause with the PRs for now.

hafarooki:
Closing this PR as the main issue (sweep table version) was resolved in #2503. Working on a new PR for the other issue (handling zero counts)

Issue #2470, opened 2025-12-01
BUG - SWAPI L2 Outputs large negative uncertainty values


pleasant-menlo:
### Description of the issue

When running the SWAPI L2 processor against L1 input that has fill value for the coincident counts and corresponding uncertainties, L2 outputs large negative uncertainties (i.e. 'swp_coin_rate_stat_uncert_minus' values of -6.896552e+31). We hypothesize this is because the L2 code is performing operations on fill values, -1e+31, from L1.

This causes L3 products to output fill, or not process altogether for the L3b product.

### Steps to reproduce the issue

Run the L2 processor against a recent production L1 file, such as `imap_swapi_l1_sci_20251130_v003.cdf`.

### Expected behavior (What should happen)

The L2 Processor should propagate fill values from L1. Measurements of -1e31 in the L1 file should produce -1e31 in the L2 file as well.

### Actual behavior (What does happen)

```
swapi_l2["swp_coin_rate_stat_uncert_minus"][0]
array([-6.8965519e+31,  1.5421159e+01,  1.8246561e+01,  1.9506393e+01,
        1.5421159e+01,  2.5804535e+01,  1.6893034e+01,  1.5421159e+01,
...
```

There is a large negative value in the first cell that does not match the configured fill value.

### Additional notes, affected areas, and suggested fixes

This was reported earlier as https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/2256, but the fix does not seem to have addressed the issue.

jtniehof:
Short term, are we able to use VALIDMIN to screen those as fill? I'm a little nervous about the possibility of values maybe getting above VALIDMIN in the course of processing, so that's not a total fix. But presumably we aren't doing a direct comparison to -1e31 (float comparisons being odious).

greglucas:
I think the thing that is causing this issue is that we are not loading the CDF files with FILLVAL -> NaN, but rather loading the FILLVAL directly and then doing some computations on that. I think we want to propagate the NaNs and re-load data with NaNs when appropriate. cdflib has a flag for that but it needs to be enabled when we are loading the CDF files.

I opened a PR with this change, but I'm not sure if that will completely solve this issue and we might need to do some more work on SWAPI's nan handling in uncertainties.
https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2486

greglucas:
[...]

PR #2486, opened 2025-12-08
MNT/FIX: Load CDF files with NaN replacement for FILLVAL by default

greglucas:
We want to propagate NaNs in data products rather than propagating the fill value itself.

related to #2470

### DIFF
--- a/imap_processing/tests/codice/test_codice_l2.py
+++ b/imap_processing/tests/codice/test_codice_l2.py
@@ -343,6 +343,8 @@ def test_codice_l2_sw_species_intensity(mock_get_file_paths, codice_lut_path):
     l2_val_data = load_cdf(l2_val_data)
     for variable in l2_val_data.data_vars:
         processed_val = processed_2_ds[variable].values
+        # NOTE: Replace nan with 0 for comparison as the validation data uses 0
+        processed_val[np.isnan(processed_val)] = 0.0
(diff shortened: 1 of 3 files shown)


greglucas:
@lacoak21 There was some CoDICE validation file comparisons that would be changed with this. It looks like the validation files we got have  "0.0" in them rather than NaNs. The VALIDMIN is 1e-10, so 0.0 isn't a valid value either, so I assume this is meant to be a nan, but I wanted to verify with you and get your thoughts. See the latest commit to address those test failures.

https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/20035815937/job/57456901352?pr=2486#step:7:3104

tmplummer (approved):
LGTM

imap_processing/tests/codice/test_codice_l2.py
```diff
@@ -382,8 +384,11 @@ def test_codice_l2_nsw_species_intensity(mock_get_file_paths, codice_lut_path):
     )
     l2_val_data = load_cdf(l2_val_data)
     for variable in l2_val_data.data_vars:
+        # NOTE: Replace nan with 0 for comparison as the validation data uses 0
```
lacoak21, line 387: This looks good to me!

lacoak21 (approved):

PR #2491, opened 2025-12-09
MNT: Rename SWAPI variable swp_esa_energy to just esa_energy

### LINKED ISSUE #2421: SWAPI: Update L2 esa_step to be esa_energy
- [ ] update L2 esa_step to be esa_energy
- [x] move swp_esa_energy to be new dimension esa_energy

greglucas:
# Change Summary

This changes `swp_esa_energy` to be `esa_energy` (drop the swp_ prefix).

@bishwassth and @hafarooki I'm not sure what the first bullet in #2421 means. Could you let me know if this is the only change you need or if I'm missing something else here that was supposed to be done?

> update L2 esa_step to be esa_energy

`esa_step` is a dimension that is of shape 72. I don't think it can/should be `esa_energy` because it will be variable and change with time. So I think what we have is currently correct where the `esa_step` is the dimension (length 72), and then esa_energy is the variable with shape (nepochs, esa_step). 

closes #2421

### DIFF
--- a/imap_processing/swapi/l2/swapi_l2.py
+++ b/imap_processing/swapi/l2/swapi_l2.py
@@ -235,7 +235,7 @@ def swapi_l2(
         data_time=sci_start_time,
     )
 
-    l2_dataset["swp_esa_energy"] = xr.DataArray(
+    l2_dataset["esa_energy"] = xr.DataArray(
         esa_energy,
         name="esa_energy",
         dims=["epoch", "esa_step"],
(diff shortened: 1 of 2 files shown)


lacoak21 (approved):

greglucas:
@pleasant-menlo I tagged you here because you were requesting some updates with SWAPI. I think this is the final L2 work I had on the list for you, is this what you were expecting/needing from the energy variable updates or could you add some additional context if you need something else here?

pleasant-menlo:
> @pleasant-menlo I tagged you here because you were requesting some updates with SWAPI. I think this is the final L2 work I had on the list for you, is this what you were expecting/needing from the energy variable updates or could you add some additional context if you need something else here?

We aren't aware of needing variable updates but we can update our code accordingly to read from these variables once this is merged.

pleasant-menlo (approved):

PR #2492, opened 2025-12-09
ULTRA L1C HELIO index maps

lacoak21:
# Change Summary

## Overview
Create ultra helio index maps on the fly to be used in the calculation of the exposure factor, efficiencies and geometric function. This is an alternative to reading in many ancillary files. It was decided that writing the code to do this is best incase we need to change it. 

Most of the code to produce the index map was written by Claude after importing the Java code that nick uses. There were a few small issues to fix after this. 

## Updated Files
   - Add constants for helio index map creation 
   - Fix inside fov function to match Nicks version (from java code)
   - make index maps on the fly for helio psets. 
-imap_processing/ultra/l1c/l1c_lookup_utils.py
   - Fix valid pixel calculation to handle energy dependent index tables. 
 - imap_processing/ultra/l1c/make_helio_index_maps.py
    - code for calculating the helio index maps 
  - fix code to use energy dependent index maps 

## Testing
- validation test for the index maps 
- validation test for helio psets for efficiency and sensitivity values 
- fixed other tests that were effected by the changes


imap_processing/ultra/l1c/l1c_lookup_utils.py
```diff
@@ -231,27 +276,36 @@ def get_spacecraft_pointing_lookup_tables(
...
+    )
+    index_grid = (
+        pd.read_csv(ancillary_files[index_descriptor], header=None, skiprows=1)
+        .to_numpy(dtype=float)
+        .T
+    )
+    boundary_scale_factors = (
+        pd.read_csv(ancillary_files[bsf_descriptor], header=None, skiprows=1)
+        .to_numpy(dtype=float)[:, 2:]
+        .T
```
lacoak21, line 296: I decided to transpose all these arrays to make them follow the dimension order that everything else follows.

imap_processing/ultra/l1c/make_helio_index_maps.py
```diff
@@ -0,0 +1,328 @@
...
+)
+from imap_processing.ultra.constants import SIM_START_ET, UltraConstants
+from imap_processing.ultra.l1b.lookup_utils import is_inside_fov
+from imap_processing.ultra.l1c.l1c_lookup_utils import build_energy_bins
+
+logger = logging.getLogger(__name__)
+
+# TODO why this constant offset??? Ask Nick for his CoordConverters.convertToRaDec
+#  function
+ULTRA_90_PHI_CORRECTION_DEG = 139.609025
```
lacoak21: I can not find the reason for this constant offset. I have tried many things including making sure Nick and I used all of the same kernels. The index table match within .02% with it so it definitely is a real offset.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -647,124 +657,6 @@ def get_efficiencies_and_geometric_function(
     return gf_averaged, eff_averaged
 
 
-def get_helio_adjusted_data(
```
lacoak21, line 650: This whole function can be removed because it was the workaround we were using instead of using the helio index tables.

imap_processing/tests/ultra/unit/test_helio_pset.py
```diff
@@ -0,0 +1,120 @@
...
+    spin_data = pd.DataFrame(
+        {
+            "spin_start_met": np.linspace(
+                pointing_range_met[0], pointing_range_met[1], nspins
+            ),
+            "spin_period_sec": np.full(nspins, nominal_spin_seconds),
+            "spin_phase_valid": np.ones(nspins),
+            "spin_period_valid": np.ones(nspins),
+        }
+    )
```
greglucas, line 64: There is a utility function `use_fake_spin_data_for_time` and `generate_fake_spin_data` that youmight be able to use here?

imap_processing/ultra/l1b/lookup_utils.py
```diff
@@ -445,21 +446,24 @@ def is_inside_fov(phi: np.ndarray, theta: np.ndarray) -> np.ndarray:
...
+        Azimuth angles in radians.
 
     Returns
     -------
     numpy.ndarray
         Boolean array indicating if the angle is in the FOV, False otherwise.
     """
     numerator = 5.0 * np.cos(phi)
-    denominator = 1 + 2.80 * np.cos(phi)
-    # Equation 19 in the Ultra Algorithm Document.
```
greglucas, line 461: Leave this equation number?

imap_processing/ultra/l1c/helio_pset.py
```diff
@@ -156,8 +168,8 @@ def calculate_helio_pset(
     geometric_function, efficiencies = get_efficiencies_and_geometric_function(
         pixels_below_scattering,
         boundary_scale_factors,
-        theta_vals,
-        phi_vals,
+        theta_vals.values,
+        phi_vals.values,
```
greglucas, line 172: Do you need to pass numpy arrays or can you pass xarray arrays and let the math just work on these?

imap_processing/ultra/l1c/l1c_lookup_utils.py
greglucas: You are transposing after the indexing though. Aren't you getting different values now? Before you were getting `arr[:, :2]`, now you are doing `arr[:, 2:].T`.

imap_processing/ultra/l1c/make_helio_index_maps.py
greglucas: This is only with ULTRA_90 and there isn't one for ULTRA_45? Is this one of the mounting angles in the SpiceFrames?

imap_processing/ultra/l1c/make_helio_index_maps.py
```diff
@@ -0,0 +1,328 @@
...
+    )
+    if compute_bsf:
+        logger.info(
+            "Computing boundary scale factors with %d points per pixel",
+            boundary_points,
+        )
+
+    time_id = 0
+    t = start_et
+    while t < (end_et - dt_step / 2):
```
greglucas, line 208: You have 3 nested Python loops here. I think you might be able to vectorize some of this math out to make it more numpy-vectorization rather than copying the Java version. This can be a follow-up PR too.

imap_processing/ultra/l1c/make_helio_index_maps.py
```diff
@@ -0,0 +1,328 @@
...
+        Compute boundary scale factors (default False).
+    boundary_points : int, optional
+        Number of boundary points per pixel (default 8).
+
+    Returns
+    -------
+    xr.Dataset
+        Dataset with helio index maps.
+    """
+    with sp.KernelPool(kernel_paths):
```
greglucas, line 106: I think we were trying to have all Spice functions in the spice module. So if you need to furnish kernels etc. that would all be handled in one common location and not within any indivdiual instruments. I'm not entirely clear why you need this though and why we can't rely on the current furnishing of kernels. Do you want to calculate ancillary files based on simulation spinning or real spin data that we actually get? i.e. I'd assume you want the simulated values during testing but real values when you are running the pipeline. Overall, I'm not entirely following how all of these files are made and whether this is an OK use of the kernel pool or if it should be moved to the spice module or handled entirely differently.

@subagonsouth pinging you as the resident spice expert.

tmplummer: I must admit, I don't understand what is going on here. I'm not clear why the helio index maps get made using simulated kernels and a fixed time.

In terms of correctly furnishing kernels so that they are only furnished for the scope of this context manager after which the kernels automatically furnished in the pre-processing step get restored, this looks OK.

lacoak21: Yeah I was confused at first as well. These were originally supposed to be calibration files to be used for every pointing that UTLRA gave us but they made one for each energy bin so that would be ~90 calibration files. I thought about making a script that combines them but Nick mentioned it would be good to have SDC code to create them. Greg and I thought it might be worth it to just have the index creation algorithm be part of the pipeline. 

I asked about using the non simulated kernels and Nick said that extra care would need to be taken, because the spin duration isn't exactly 15 second so then you're picking a spin within the pointing and rotating the spacecraft based on that. When you go to scale it, you'd have to track that for every pointing

Maybe thats fine? Im also thinking about just making them once and reading them in as ancillary because this might be overkill to recalculate them every time especially since they will only change if the nside changes.

imap_processing/ultra/l1c/l1c_lookup_utils.py
lacoak21: I updated the code downstream to handle the transposed values.

imap_processing/ultra/l1c/helio_pset.py
lacoak21: True they dont need to be numpy.

imap_processing/ultra/l1c/l1c_lookup_utils.py
greglucas: I'm still confused though, you are getting fundamentally different values out of the file itself aren't you? `np.arange(4)[:2] != np.arange(4)[2:]`

lacoak21: OH! I see. Wow yikes good catch. I need to go understand why the test didnt catch this.

lacoak21: Ok I was doing that just for the ra_and_dec variable. 


The other variables were [:, 2:] and are still [:, 2:]

It used to be 
```
index_grid = pd.read_csv(
        ancillary_files[index_descriptor], header=None, skiprows=1
    ).to_numpy(dtype=float)
ra_and_dec = index_grid[:, :2]  
```
but now its 
```
index_grid = pd.read_csv(
        ancillary_files[index_descriptor], header=None, skiprows=1
    ).to_numpy(dtype=float).T
ra_and_dec = index_grid[:2, :]
```

imap_processing/ultra/l1c/make_helio_index_maps.py
lacoak21: Oh right I have to validate ultra 45. Im assuming there will be a new constant for 45

imap_processing/ultra/l1c/make_helio_index_maps.py
lacoak21: Ill add a TODO

imap_processing/ultra/l1c/make_helio_index_maps.py
tmplummer: Ok, I think that I am fine with this approach. Like I said, the kernel handling looks like it is done correctly and shouldn't cause any side-effects.

imap_processing/ultra/l1c/make_helio_index_maps.py
lacoak21: Actually Nick and I figured out what was causing this (different start times). There was a minute offset between et and utc. I am now using the start of the pointing kernel to make it far more clear. This hardcoded timestamp was a bad idea.

imap_processing/ultra/l1c/make_helio_index_maps.py
lacoak21: OK great. Im also going to use the start ET value of the pointing instead of the hardcoded value as well. That should have been done this way from the start.

tmplummer: But doesn't that mean you will run into kernel coverage issues with the simulated kernels?

lacoak21: I meant the start et of the simulated pointing kernel.

imap_processing/tests/ultra/unit/test_helio_pset.py
lacoak21: use_fake_spin_data_for_time is not giving me the correct number of nominal spins because it is marking some as invalid. I did it this way so I can explicitly set the number of valid spins.

imap_processing/ultra/l1c/make_helio_index_maps.py
```diff
@@ -0,0 +1,335 @@
...
+
+    Returns
+    -------
+    xr.Dataset
+        Dataset with helio index maps.
+    """
+    # Get all loaded SPK kernels
+    spk_kernels = [sp.kdata(i, "spk")[0] for i in range(sp.ktotal("spk"))]
+    # Find the de440s.bps kernel
+    de440s_file = next((k for k in spk_kernels if "de440s" in k), None)
```
tmplummer: There are both the `de440.bsp` and `de440s.bsp` kernels. The `de440s.bsp` kernel has the "s" for shortened and excludes some far past and distant future coverage in order to make a smaller file. We use the "s" version in testing, but the full `de440.bsp` kernel in production. You should make this code find either one.
```suggestion
    de440s_file = next((k for k in spk_kernels if "de440" in k), None)
```

imap_processing/ultra/l1c/make_helio_index_maps.py
```diff
@@ -0,0 +1,335 @@
...
+    imap_state,
+)
+from imap_processing.ultra.constants import UltraConstants
+from imap_processing.ultra.l1b.lookup_utils import is_inside_fov
+from imap_processing.ultra.l1c.l1c_lookup_utils import build_energy_bins
+
+logger = logging.getLogger(__name__)
+
+
+def vector_ijk_to_theta_phi(
```
tmplummer, line 23: This pretty much looks like the `geometry.cartesian_to_spherical` function: 
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/geometry.py#L521

imap_processing/ultra/l1c/make_helio_index_maps.py
```diff
@@ -0,0 +1,335 @@
...
+            )  # Galilean transform
+            helio_normalized = helio_velocity / np.linalg.norm(
+                helio_velocity, axis=1, keepdims=True
+            )
+
+            # Transform to inst
+            inst_vecs = helio_normalized @ rotation_matrix
+            theta, phi = vector_ijk_to_theta_phi(inst_vecs)
+
+            phi = np.where(phi > np.pi, phi - 2 * np.pi, phi)
```
tmplummer, line 235: I think that this is converting to what SPICE calls latitudinal coordinates. You could use this function.  The documentation is incomplete, but if you look at the SPICE docs, I think it returns what you want.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/geometry.py#L615
https://naif.jpl.nasa.gov/pub/naif/misc/toolkit_docs_N0067/C/cspice/reclat_c.html
```suggestion
            latitudinal = cartesian_to_latitudinal(inst_vecs)
            theta = latitudinal [..., 2]
            phi = latitudinal[..., 1]
```

imap_processing/ultra/l1c/make_helio_index_maps.py
```diff
@@ -0,0 +1,335 @@
...
+    # TODO vectorize loop
+    time_id = 0
+    t = start_et
+    while t < (end_et - dt_step / 2):
+        # Get rotation matrix for this time step
+        rotation_matrix = get_rotation_matrix(
+            np.array([t]),
+            from_frame=instrument_frame,
+            to_frame=SpiceFrame.IMAP_DPS,
+        )[0]
```
tmplummer: Instead of passing an array of length one for the `et` argument and taking the zeroth element of the result, I think you should be able to do:
```suggestion
        rotation_matrix = get_rotation_matrix(
            t,
            from_frame=instrument_frame,
            to_frame=SpiceFrame.IMAP_DPS,
        )
```

imap_processing/ultra/l1c/make_helio_index_maps.py
```diff
@@ -0,0 +1,335 @@
...
+            # Transform pixel vectors to heliocentric frame
+            helio_velocity = (
+                sc_vel.reshape(1, 3) + kps * pixel_vecs
+            )  # Galilean transform
+            helio_normalized = helio_velocity / np.linalg.norm(
+                helio_velocity, axis=1, keepdims=True
+            )
+
+            # Transform to inst
+            inst_vecs = helio_normalized @ rotation_matrix
```
tmplummer: Typically, rotations are applied in the opposite direction. That is you would see `output_vector = rotation_matrix @ input_vector`. It looks like the rotation_matrix is going from instrument frame -> DPS which I then means that this order is the equivalent of doing `output_vector = rotation_matrix.T @ input vector` and transposing a rotation is the same as inverting it. So I think this is doing as your comment says transforming from DPS to instrument frame but it is a bit confusing.

imap_processing/ultra/constants.py
```diff
@@ -157,3 +159,21 @@ class UltraConstants:
         "proton": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19],
         "non_proton": [20, 21, 22, 23, 24, 25, 26],
     }
+
+
+SPICE_DATA_TEST_PATH = imap_module_directory / "ultra/l1c/sim_spice_kernels"
```
tmplummer: I would rename this to `SPICE_DATA_SIM_PATH` so as to not confuse with test data.

tmplummer (commented):
Just a few things worth reviewing.

imap_processing/ultra/l1c/make_helio_index_maps.py
lacoak21: I had that in there but this is vectorized and far faster. It was taking a pretty long time with the cartesian_to_spherical call.

imap_processing/ultra/l1c/make_helio_index_maps.py
lacoak21: Ok that makes sense - Im going to switch this up.

greglucas (approved):
Do we need all of the `sim_spice_kernels` included in the released package because we want those sort of "frozen in time" in a sense? This means we wouldn't get any updated clock / instrument offset updates etc.

Looks good to me and I think you already discussed that with @subagonsouth so I'm fine with it all on the code front.

lacoak21:
> Do we need all of the `sim_spice_kernels` included in the released package because we want those sort of "frozen in time" in a sense? This means we wouldn't get any updated clock / instrument offset updates etc.
> 
> Looks good to me and I think you already discussed that with @subagonsouth so I'm fine with it all on the code front.

Yes exactly we want those never to change. Unless Nick requests they do.

imap_processing/ultra/l1c/make_helio_index_maps.py
tmplummer: Oh, weird. That function is also vectorized. I am surprised that this provided a speedup. I'll have to write a ticket to investigate.

tmplummer (approved):
LGTM

Issue #2489, opened 2025-12-08
Healpix <-> Rectangular needs to ignore nans


greglucas:
Currently, it sounds like if we get a nan value in the healpix subdivision, we are "poisoning" the entire rectangular grid cell with that result. We should ignore those nan values in our calculations, and only set the value to nan if _all_ healpix values are nan within that rectangular cell. i.e. if we have one good healpix value and 5 nan's, we should only use that one good value with the associated exposure time / angle etc. and not anything else.

I think this is the relevant area of the code.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/3fe84244317bd4970f45f9a24d83cc00c83406b8/imap_processing/ena_maps/ena_maps.py#L1672-L1679

Relevant tests here:

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/3fe84244317bd4970f45f9a24d83cc00c83406b8/imap_processing/tests/ena_maps/test_ena_maps.py#L1453

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/3fe84244317bd4970f45f9a24d83cc00c83406b8/imap_processing/tests/ena_maps/test_ena_maps.py#L1306

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/3fe84244317bd4970f45f9a24d83cc00c83406b8/imap_processing/tests/ena_maps/test_ena_maps.py#L1397

lacoak21:
@pleasant-menlo Is this complete?

pleasant-menlo:
Yes, it looks like the PR https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2493 addressed this.

PR #2493, opened 2025-12-09
Ignore fill values when converting healpix to rectangular

pleasant-menlo:
# Change Summary

## Overview
Addressing issue #2489 to ignore fill values when calculating the weighted averages of subpixels.

## Updated Files
- ena_maps.py
  - No longer weighting by the full rectangular pixel, we are taking a mask of the finite values and applying those to the numerator and denominator in the weighted averaging.

## Testing
- Modified `test_calculate_rect_pixel_value_from_healpix_map_n_subdivisions` to include nan value that should be ignored when performing subdivision averaging.

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -1633,13 +1633,6 @@ class HealpixSkyMap(AbstractSkyMap):
             + subpix_spacing / 2
         )
 
-        # We must weight by solid angle, which is not exactly equal for all subpixels
-        # Calculate the solid angle of the full rectangular pixel (sterad)
-        full_rect_pixel_solid_angle = np.deg2rad(rect_pix_spacing_deg) * (
-            np.sin(np.deg2rad(bottom_edge_lat + rect_pix_spacing_deg))
-            - np.sin(np.deg2rad(bottom_edge_lat))
-        )
-
         # Calculate solid angle of each subpix from the rect_subpix_lat_ctrs (sterad)
         all_edges_lat = bottom_edge_lat + np.arange(n_subpix_side + 1) * subpix_spacing
         sine_all_edges_lat = np.sin(np.deg2rad(all_edges_lat))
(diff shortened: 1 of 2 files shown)


imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -1327,18 +1327,18 @@ def test_calculate_rect_pixel_value_from_healpix_map_n_subdivisions(
...
-            (180.5, -1.5): 12,
+            (180.5, -1.5): 13,
             (180.5, -0.5): 14,
             (180.5, 0.5): 15,
             (180.5, 1.5): 16,
             (181.5, -1.5): 17,
             (181.5, -0.5): 18,
             (181.5, 0.5): 19,
-            (181.5, 1.5): 20,
+            (181.5, 1.5): 999,
```
tmplummer, line 1339: I'm not sure I understand this change. Can you add some comments about why this entry is out of family from the others?

imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -1354,18 +1354,19 @@ def mock_ang2pix_fn(nside, theta, phi, nest=True, lonlat=False):
...
+        hp_map.data_1d["counts"][0, 999] = np.nan
 
         for num_subdiv, (expected_value, atol) in enumerate(
             [
                 # The first subdivs have all the same solid angle
                 (expected_mean_0_subdivisions, 1e-9),
                 (expected_mean_1_subdivisions, 1e-9),
                 # Slight difference from not taking into account asym solid angle
-                (expected_mean_2_subdivisions, 0.1),
+                (expected_mean_2_subdivisions, 1e-4),
```
tmplummer, line 1371: Why was this value changed? The comment about not taking into account the asym solid angle still seems to apply.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -1669,13 +1662,16 @@ def calculate_rect_pixel_value_from_healpix_map_n_subdivisions(
         # Get the healpix values at the rectangular subpixel centers
         hp_vals_at_rect_pix_ctrs = value_array.values[..., hp_pix_at_rect_subpix_ctrs]
 
+        mask = np.isfinite(hp_vals_at_rect_pix_ctrs)
```
tmplummer: Just a nit-pick that `mask` is a very generic uninformative name here.

tmplummer (commented):
I would like some clarification about changes in the test as my comments reflect.

imap_processing/tests/ena_maps/test_ena_maps.py
pleasant-menlo: We made this entry out of family from the others to make it clear that this entry will be fill value.

imap_processing/tests/ena_maps/test_ena_maps.py
pleasant-menlo: We were seeing that the tolerance for the expected mean could be smaller after having fixing line 1330 that previously was pointing to the same pixel as line 1329 to instead point to a different pixel. We thought that this was likely a typo when setting up the test.

imap_processing/tests/ena_maps/test_ena_maps.py
```diff
@@ -1327,18 +1328,18 @@ def test_calculate_rect_pixel_value_from_healpix_map_n_subdivisions(
...
-            (180.5, -1.5): 12,
+            (180.5, -1.5): 13,
             (180.5, -0.5): 14,
             (180.5, 0.5): 15,
             (180.5, 1.5): 16,
             (181.5, -1.5): 17,
             (181.5, -0.5): 18,
             (181.5, 0.5): 19,
-            (181.5, 1.5): 20,
+            (181.5, 1.5): 999,
```
tmplummer, line 1339:
```suggestion
            # Set the final entry to 999 to identify a pixel that will be set to NaN
            (181.5, 1.5): 999,
```

tmplummer (approved):
Thanks for doing this!

PR #2497, opened 2025-12-10
Feat/update mag frame

mhairifin:
# Change Summary

## Overview
- Update MAGO and MAGI spice frames to use IMAP_MAG_BASE instead of the mechanical spice frames. The MAG team have agreed to use an ideal frame for the spice transformation as our in flight calibration already includes rotation into the idealised frame.
- Update MAG L2 and L1D variable names for the vectors

## Updated Files
- ialirt/l0/parse_mag
   - Change MAGO and MAGI frame to IMAP_MAG_BASE for I-ALiRT
- mag/l2/mag_l2_data
   - Introduce additional fields to ValidFrames enum to distinguish MAGI and MAGO despite sharing a spice frame, and add in the variable name and variable attributes identifier
   - Use these new fields to generate an L2 dataset
- mag/l1d/mag_l1d_data
   - Use new spice_frame formulation

## Testing
- Tests updated where needed to use `spice_frame`
- Frame rotation test updated to reflect that MAGI and MAGO share a spice frame
- L2 dataset generation tests updated to test for correct variable name


imap_processing/mag/l2/mag_l2_data.py
```diff
@@ -20,13 +20,33 @@
...
 
-    MAGO = SpiceFrame.IMAP_MAG_O
-    MAGI = SpiceFrame.IMAP_MAG_I
-    DSRF = SpiceFrame.IMAP_DPS
-    SRF = SpiceFrame.IMAP_SPACECRAFT
-    GSE = SpiceFrame.IMAP_GSE
-    GSM = SpiceFrame.IMAP_GSM
-    RTN = SpiceFrame.IMAP_RTN
+    MAGO = ("MAGO", SpiceFrame.IMAP_MAG_BASE, "vector_attrs", "vectors")
+    MAGI = ("MAGi", SpiceFrame.IMAP_MAG_BASE, "vector_attrs", "vectors")
```
alastairtree: If both sensors are in the same spice frame IMAP_MAG_BASE what are the consequences? I had assumed we needed both to be different to be able to handle the the slightly different positions of them

alastairtree (commented):
Looks very tidy to me!

imap_processing/mag/l2/mag_l2_data.py
mhairifin: We do need to transform them differently - but we already do so by providing separate calibration matrices which are used for L2 and L1D. We create those calibration matrices by using the magnetic field data to work out how non-aligned with the spacecraft reference frame the sensors are. So either we use an idealised transform here, or we would have to create a calibration matrix to align it, and then un-twist it by this spice transform before we provide it as ancillary data.

alastairtree: Based on our chat we have agreed
- Fix the lowercase MAGi=>MAGI
- Add comment explain why BASE frame is used for L2/L1D only
- Add frames for the unused, but still valid, SPICE magi/o frames and call them "ground calibration"

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -712,7 +712,7 @@ def process_packet(
             attitude_time,
             time_data["secondary_epoch"],
             magi_out,
-            SpiceFrame.IMAP_MAG_I,
+            SpiceFrame.IMAP_MAG_BASE,
```
maxinelasp: I think it would make sense to point this to `ValidFrames.IMAP_MAGI` so it can benefit from that abstraction layer if ValidFrames gets updated again.

imap_processing/mag/l2/mag_l2_data.py
```diff
@@ -20,13 +20,105 @@
...
+        "MAGI_GROUND_CAL",
+        SpiceFrame.IMAP_MAG_I,
+        "vector_attrs",
+        "vectors",
+    )
+
+    DSRF = ("DSRF", SpiceFrame.IMAP_DPS, "vector_attrs_dsrf", "b_dsrf")
+    SRF = ("SRF", SpiceFrame.IMAP_SPACECRAFT, "vector_attrs_srf", "b_srf")
+    GSE = ("GSE", SpiceFrame.IMAP_GSE, "vector_attrs_gse", "b_gse")
+    GSM = ("GSM", SpiceFrame.IMAP_GSM, "vector_attrs_gsm", "b_gsm")
```
maxinelasp, line 51: There are a few people who are using MAG data as inputs - we will need to reach out to them if we change variable names here to make sure they can access the correct variable. (Or open a PR against L3 code, I can take a look at this.)

The current users of MAG are:
CoDICE L3b
HIT L3
SWE L3

That's not too many, so I think it's fine to change the names - just something we should consider and check.

maxinelasp (commented):
Looks good! Nice improvement. I think we should hold back on merging this until I'm able to investigate L3 MAG usage, but I can deploy this to dev now for further testing. I poked around more and it doesn't look like anywhere in the MAG codebase assumes that the main variable is named "vectors".

imap_processing/mag/l2/mag_l2_data.py
greglucas: ping @jtniehof and @pleasant-menlo to see if this renaming of MAG frames impacts you.

pleasant-menlo: Just to confirm, this will not change the "vectors" variable name in the Mag L1d products? If not, this does not cause any issues for L3 processing.

Thanks! - Ethan and Peter

mhairifin: Hi - it does change the "vectors" variable name. It will be `b_gse`/`b_dsrf`/`b_srf` etc. dependent on the utilised reference frame

greglucas: @mhairifin do you want all of these variables in the same file now? i.e. we'll have one `l2_norm` file with all 3 of those vectors now, or will the filename be `l2_norm-dsrf` AND have a `b_dsrf` variable in it?

greglucas:
@mhairifin it looks like you have conflicts from the last merge now that need to be resolved.

imap_processing/mag/l2/mag_l2_data.py
mhairifin: We are happy with each reference frame in a different file - we would just like to change the variable name to use one that is more standard for magnetic fields, and to specify the co-ordinate system in the variable name for consistency with I-ALiRT and more clarity for science end users

greglucas: OK, I have a slight concern that this should have all been discussed long ago where @maxinelasp and Drew led a discussion at one of the science team meetings several years ago now I think to align all instruments on the naming conventions and settle on a consistent naming for variables. I personally prefer this update (I didn't like vectors as being too generic), but just want to put this down on asking why this is happening so late in the game here. Maybe it was just never updated by the SDC and this is what was settled on long ago.

mhairifin: I believe we did discuss changing the names of these variables over the summer with @maxinelasp but it never became the highest priority bit of work, and I don't think we were aware it would be a problem. I don't think it's an unreasonable concern, and if you consider it a blocking issue then we can discuss internally how to move forwards, but these variable names would be our preference.

pleasant-menlo: Hi, thanks for the clarification that it will change variable names. We created a branch in L3 that will use the new variable name if/when this change goes through. We will have to redeploy SWE, HIT, and CoDICE to prod when the L2 change gets deployed.

Thanks,
Patrick and Sebastian

alastairtree (approved):

greglucas (approved):
Lets get this in then if you all want it and L3 has been updated fairly easily. @pleasant-menlo feel free to make the changes then and we can deploy new containers all at the same time with this update and yours.

pleasant-menlo:
@greglucas We deployed the SWE, Hit, and CoDICE containers to dev with this change, they are ready to be deployed into production. 

CAVA expects variable names to be consistent across a whole dataset. Is it possible or planned to reprocess all of the mag data that has been produced so far in production after this change?

Thanks,
Patrick and Sebastian

greglucas:
> CAVA expects variable names to be consistent across a whole dataset. Is it possible or planned to reprocess all of the mag data that has been produced so far in production after this change?

Absolutely! This is an SPDF archiving requirement too I believe.

greglucas:
I have deployed this update in the latest MAG container release, and I have pushed the SWE, HIT, and CoDICE L3 containers to production to align with that update. Let me know if there are any other updates needed.

@maxinelasp is going to work on reprocessing the l1d MAG files later today/tomorrow after the pointing attitude jobs get reprocessed with recent updates from them that would be good to have in the MAG files.

PR #2501, opened 2025-12-10
I-ALiRT - Update codice l2

laspsandoval:
# Change Summary

## Overview
Formatting to get codice_lo into a format that can be inserted into DynamoDB.

## Updated Files
- process_codice.py
   - Formatting to get codice_lo into a format that can be inserted into DynamoDB.

## Testing
- test_process_codice.py

### DIFF
--- a/imap_processing/ialirt/l0/process_codice.py
+++ b/imap_processing/ialirt/l0/process_codice.py
@@ -317,7 +317,7 @@ def calculate_ratios(
         )  # (epoch,)
         pseudo_density_dict[species] = summed_pseudo_density.values
 
-    species = constants.LO_IALIRT_VARIABLE_NAMES
+    species_list = constants.LO_IALIRT_VARIABLE_NAMES
 
     # Denominator.
     # Note that outside of this test a zero value denominator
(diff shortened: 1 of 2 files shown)


greglucas (commented):
I have a few suggestions that I think might make this a little cleaner without all the case handlings.

* Stick to numpy arrays. So try to initialize an empty array with the expected shape.
`pseudo_density_arr = np.empty((len(constants.LO_IALIRT_VARIABLE_NAMES), len(intensity))`
Then you can ignore the division by zero and let nan's fall out automatically and replace with fills later if you want. (remove all if 0 branches)

* Use an IntEnum for your species names. Then you can lookup into arrays with `pseudo_density_arr[LoSpecies.O_7_plus, :]` rather than `pseudo_density_dict[species_list[3]]` with kind of two redirects in there which is confusing for me to follow without looking back up what index is what in the species list.

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -242,10 +257,141 @@ def convert_to_intensities(
...
+
+    # Denominator.
+    # Note that outside of this test a zero value denominator
+    # will lead to a null value.
+    # The use of zeros here is only to match the test data as
+    # confirmed by the instrument team.
+    o_abundance_denom = (
+        pseudo_density_dict[species_list[3]]
+        + pseudo_density_dict[species_list[4]]
+        + pseudo_density_dict[species_list[5]]
```
greglucas: I thought I commented on this somewhere else earlier but couldn't find it immediately. I think this would be much easier to follow with an `IntEnum` to index into the named species.

```python
class Species(IntEnum):
    OXYGEN = 0
    CARBON2 = 1
    ...

pseudo_density_arr[Species.OXYGEN]
```

laspsandoval: You are right. It's not too readable. I decided to just write out the names.

laspsandoval:
@greglucas sorry it took me so long to get back to this. Would you mind looking it over again?

greglucas (approved):
Yes, much easier to read with string names, thank you for updating that!

PR #2502, opened 2025-12-10
MNT/FIX: Update SWAPI L2 to handle generalized fine sweeps

### LINKED ISSUE #2500: Update SWAPI LUT
SWAPI is doing another LUT update.

They also provided the incorrect 0 energy background sweep numbers. Those are at the beginning of the fine sweeps, not the end.

greglucas:
We are not guaranteed to get 9 fine sweep values in the LUT, so we need to only look up values when "solve" is declared. There have been quite a few updates with different changes, so this should address things in a general way so that we only need to upload new ancillary files now and don't need to update the code anytime new fine stepping sizes change with sweep number.

For now, everything is in a loop per-sweep, I'm not sure off the top of my head if there is a way to do this more efficiently with numpy vectorization with all the potential changing of sizes and wanted to get this in and we can tackle that later if we have more time. It is not that slow as-is.

closes #2500

### DIFF
--- a/imap_processing/swapi/l2/swapi_l2.py
+++ b/imap_processing/swapi/l2/swapi_l2.py
@@ -65,9 +65,16 @@ def solve_full_sweep_energy(
         dtype="datetime64[ns]"
     )
 
-    first_63_energies = []
+    # Initialize the output energy array
+    # Each sweep will have different energies for each step.
+    # The first 63 energies are coarse steps, then followed by 9 fine steps.
+    # The 9 fine steps may be defined in the main table (fixed steps), or "solve"
+    # which requires a separate lookup in the lut-notes table.
+    energy_data = np.empty((len(sweep_table), NUM_ENERGY_STEPS), dtype=float)
 
-    for time, sweep_id in zip(data_time, sweep_table, strict=False):
+    for i_sweep, (time, sweep_id, esa_lvl5_val) in enumerate(
+        zip(data_time, sweep_table, esa_lvl5_data, strict=True)
+    ):
         # Find the sweep's ESA data for the given time and sweep_id
         subset = esa_table_df[
             (esa_table_df["timestamp"] <= time) & (esa_table_df["Sweep #"] == sweep_id)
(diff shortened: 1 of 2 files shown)


imap_processing/swapi/l2/swapi_l2.py
```diff
@@ -89,83 +96,56 @@ def solve_full_sweep_energy(
...
-    # Therefore, instead of relying on np.isin or similar, we explicitly use
-    # np.where in a loop to find the index of each value in esa_lvl5_data individually,
-    # ensuring the output array has the same shape as the input.
-
-    # Page 31 of algorithm document
-    last_energy_step_indices = np.array(
-        [
-            np.where(lut_notes_df["ESA DAC (Hex)"].values == val)[0][0]
-            for val in esa_lvl5_data
+        # Subset data can contain multiple sweeps of 72 energy values.
```
lacoak21, line 99: If each sweep will have different energy values, why are you only taking the last sweep?

lacoak21 (approved):
Looks good!

imap_processing/swapi/l2/swapi_l2.py
lacoak21: Nvm I missed that this is already a subset df for that sweep.

bishwassth (approved):
I checked the fine sweep values on CDF during the LUT changes and they look good to me.

imap_processing/swapi/l2/swapi_l2.py
greglucas: Yeah this is a slightly tricky one. A subset here contains all instances of `Sweep # == X`, but because the instrument team could have updated the values within that sweep number at a later date, there could be multiple dates associated with a specific sweep number being active. Something like this:
```
(sweep1, 2010-01)
(sweep2, 2010-01)
(sweep3, 2010-01)
(sweep2, 2020-01)
```

subset of `sweep == 2` would return 144 values associated with both of the sweep 2 time entries (72 each). And we only want to use the most recent one.

PR #2484, opened 2025-12-05
ENH: Add sweep table 3 processing handling to SWAPI

greglucas:
SWAPI updated their algorithms for sweep table 3 that uses a new fine stepping approach with different step offsets.

closes #2476

PR #2503, opened 2025-12-11
I-ALiRT - SWAPI Quick Fix for LUT

laspsandoval:
# Change Summary

## Overview
This is to accommodate the new lut.

## Updated Files
- process_swapi.py
   - Checks for sweep number and then latest timestamp.
- imap_swapi_esa-unit-conversion_20251201_v001.csv
   - New lut.

## Testing
- test_process_swapi.py
- conftest.py
- iois_1_packets_2025_343_00_00_17


imap_processing/tests/ialirt/unit/test_process_swapi.py
```diff
@@ -118,47 +116,6 @@ def test_decom_packets(xarray_data, swapi_test_data):
         )
 
 
-@pytest.mark.external_test_data
```
laspsandoval: This test really didn't do anything except check key names.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -202,19 +203,21 @@ def process_swapi_ialirt(
...
     # Extract energy values from the calibration lookup table file
     calibration_lut_table["timestamp"] = pd.to_datetime(
-        calibration_lut_table["timestamp"], format="%m/%d/%Y %H:%M"
-    )
-    calibration_lut_table["timestamp"] = calibration_lut_table["timestamp"].to_numpy(
-        dtype="datetime64[ns]"
+        calibration_lut_table["timestamp"]
     )
+    subset_sweep = calibration_lut_table[
+        calibration_lut_table["Sweep #"] == SWEEP_NUMBER
```
greglucas: I don't think this is right. We want the sweep id that is coming down with the ialirt packet, not fixed on 2 because we are now on sweep id 4 and 5.

imap_processing/tests/ialirt/data/l0/imap_swapi_esa-unit-conversion_20251201_v001.csv
greglucas, line 1: Rather than adding this, can we expand/add to the current ancillary file we have?

Here is what I did when trying to handle sweep id 3:
https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2484/files#diff-c75899ffbc3d739be233d0ea7cb956c8148cfdadd7e2443b0c66f708102d9401R329

Also, if you are hard-coding use of SWEEP2, shouldn't you be able to reference the lut that is in the main swapi tests folder?

imap_processing/tests/ialirt/data/l0/iois_1_packets_2025_343_00_00_17
greglucas, line 1: Do we need to add more packet files, shouldn't we already have swapi postlaunch packets in the other iois_* packet files we have in the repository?

laspsandoval: We do not currently have any iois packet files in the repo.

imap_processing/ialirt/l0/process_swapi.py
laspsandoval: You are right. I'm getting to know this code now. I made some changes.

imap_processing/tests/ialirt/data/l0/imap_swapi_esa-unit-conversion_20251201_v001.csv
laspsandoval: Yes. I made that change. And I am no longer hard-coding the sweep id.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -88,47 +88,32 @@ def optimize_pseudo_parameters(
...
+    sol = curve_fit(
+        f=count_rate,
+        xdata=energy_passbands.take(range(max_index - 3, max_index + 3), mode="clip"),
+        ydata=count_rates.take(range(max_index - 3, max_index + 3), mode="clip"),
+        sigma=count_rate_error.take(range(max_index - 3, max_index + 3), mode="clip"),
+        p0=initial_param_guess,
+    )
 
-    return solution_dict
+    return sol[0][0], sol[0][1], sol[0][2]
```
greglucas: Can you just return the first solution instead of unpacking everything individually?
```suggestion
    return sol[0]
```

imap_processing/tests/ialirt/data/l0/iois_1_packets_2025_343_00_00_17
greglucas: I see a lot that are git ignored. I assume these are marked in external test data somehow?

image

imap_processing/tests/ialirt/data/l0/swapi_ialirt_energy_steps.csv
greglucas, line 1: I still think this can be removed and tested with the utilities from the main repository can't it?
https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2503#discussion_r2609094824

Bring in the `esa_unit_conversion_table` fixture and change sweep#2 to 3 and then you verify that your lookup on 3 is valid without adding a new file and needing to update it for 4 and 5 which SWAPI is now at as well.

```
def test_solve_full_sweep_energy_id3(esa_unit_conversion_table, lut_notes_table):
    """Test the solve_full_sweep_energy function for sweep id 3 case"""
    # Modify the current conversion table we have to have entries for sweep id 3
    esa_unit_conversion_table = esa_unit_conversion_table.copy(deep=True)
    esa_unit_conversion_table.loc[
        esa_unit_conversion_table["Sweep #"] == 2, "Sweep #"
    ] = 3
```

imap_processing/tests/ialirt/data/l0/iois_1_packets_2025_343_00_00_17
laspsandoval: Yes. You are right. I had forgotten about those.

imap_processing/ialirt/l0/process_swapi.py
laspsandoval: Yes. Nice idea.

imap_processing/tests/ialirt/data/l0/swapi_ialirt_energy_steps.csv
laspsandoval: I can do it with everything except 
def test_optimize_parameters():

The energy values are different. To change this one I would need to request new validation values from the instrument team.

imap_processing/tests/ialirt/data/l0/iois_1_packets_2025_343_00_00_17
laspsandoval: Unfortunately those were before the swapi instrument was turned on. I need to go back and do some cleanup I think.

imap_processing/tests/ialirt/unit/test_process_swapi.py
```diff
@@ -32,6 +33,24 @@ def binary_packet_path():
     )
 
 
+@pytest.fixture(scope="session")
+def esa_unit_conversion_table() -> pd.DataFrame:
```
greglucas: I think you might be able to import the fixture from the other test module so you don't have to duplicate it here.

`from imap_processing.tests.swapi...`

greglucas (approved):
OK, this looks good now, just one minor suggestion if it works. Thank you, Laura!

PR #2509, opened 2025-12-12
CODICE l2 lo direct events

### LINKED ISSUE #2118: CoDICE: L2 Implement Lo direct events data products

lacoak21:
# Change Summary

closes #2118

## Overview
Direct Event data processing for CoDICE lo. I added new validation files as well. CoDICE hi direct events will be in a new PR

## Updated Files
- imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
   - Duplicate attribute
- imap_processing/cdf/config/imap_codice_l2-lo-direct-events_variable_attrs.yaml
   - attrs (Most of them were derived from the validation CDFs
    - add handling for lookup tables and de processing code 

## Testing
Test lookup utils and validate processing

### DIFF
--- a/imap_processing/codice/codice_l2.py
+++ b/imap_processing/codice/codice_l2.py
@@ -10,12 +10,13 @@ dataset = process_codice_l2(l1_filename)
 """
 
 import logging
+from pathlib import Path
 
 import numpy as np
 import pandas as pd
 import xarray as xr
 from imap_data_access import ProcessingInputCollection, ScienceFilePath
-from numpy._typing import NDArray
+from numpy.typing import NDArray
 
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.cdf.utils import load_cdf
@@ -983,7 +984,7 @@ def process_lo_direct_events(dependencies: ProcessingInputCollection) -> xr.Data
     energy_table, energy_bins = get_lo_de_energy_luts(dependencies)
     # Convert from position to elevation angle in degrees relative to the spacecraft
     # axis
-    l2_dataset = l1a_dataset.copy(deep=True)
+    l2_dataset = l1a_dataset
     # Create a new coordinate for elevation_angle based on inst_az
     pos_to_els = (
         LO_POSITION_TO_ELEVATION_ANGLE["sw"] | LO_POSITION_TO_ELEVATION_ANGLE["nsw"]
@@ -999,17 +1000,15 @@ def process_lo_direct_events(dependencies: ProcessingInputCollection) -> xr.Data
     # Convert spin_sector to spin_angle in degrees
     # Use equation from section 11.2.2 of algorithm document
     # Shift all spin sectors for all positions 13 - 24 adding 12 and mod 24
+    original_spin_sector = l2_dataset["spin_sector"].values
     l2_dataset["spin_sector"] = xr.where(
         (l2_dataset["position"] >= 13) & (l2_dataset["position"] <= 24),
         (l2_dataset["spin_sector"] + 12) % 24,
         l2_dataset["spin_sector"],
     )
-    l2_dataset["spin_angle"] = xr.DataArray(
-        data=l2_dataset["spin_sector"].data * 15.0 + 7.5,
-        dims=l2_dataset["spin_sector"].dims,
-    ).astype(np.float32)
+    l2_dataset["spin_angle"] = l2_dataset["spin_sector"].astype(np.float32) * 15.0 + 7.5
     l2_dataset["spin_angle"] = xr.where(
-        (l2_dataset["spin_sector"] > 23), np.nan, l2_dataset["spin_angle"]
+        (original_spin_sector > 23), np.nan, l2_dataset["spin_angle"]
     )
     # convert apd energy to physical units
     # Set the gain labels based on gain values
@@ -1085,18 +1084,6 @@ def process_lo_direct_events(dependencies: ProcessingInputCollection) -> xr.Data
         l2_dataset["epoch"].data,
         dims="epoch",
         attrs=cdf_attrs.get_variable_attributes("epoch", check_schema=False),
-    )
-    l2_dataset["epoch_delta_minus"] = xr.DataArray(
-        data=l2_dataset["epoch_delta_minus"].data.astype(np.int64),
-        dims="epoch",
-        attrs=cdf_attrs.get_variable_attributes(
-            "epoch_delta_minus", check_schema=False
-        ),
-    )
-    l2_dataset["epoch_delta_plus"] = xr.DataArray(
-        l2_dataset["epoch_delta_plus"].data.astype(np.int64),
-        dims="epoch",
-        attrs=cdf_attrs.get_variable_attributes("epoch_delta_plus", check_schema=False),
     )
     # Add labels
     l2_dataset["event_num_label"] = xr.DataArray(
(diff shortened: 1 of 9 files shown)


imap_processing/codice/codice_l2.py
```diff
@@ -10,12 +10,12 @@
 """
 
 import logging
-from pathlib import Path
 
 import numpy as np
 import pandas as pd
 import xarray as xr
 from imap_data_access import ProcessingInputCollection, ScienceFilePath
+from numpy._typing import NDArray
```
greglucas:
```suggestion
from numpy.typing import NDArray
```

imap_processing/codice/codice_l2.py
```diff
@@ -44,6 +44,97 @@
...
+    ]
+
+    return energy_lut, energy_bins_lut
+
+
+def get_mpq_calc_energy_conversion_vals(
+    dependencies: ProcessingInputCollection,
+) -> np.ndarray:
+    """
+    Get the MPQ calculation esa step to energy kev conversion lookup table values.
```
greglucas: Is MPQ defined anywhere? Could you define it somewhere so that if someone searches for those words they get to one of these files too.

imap_processing/codice/codice_l2.py
```diff
@@ -866,6 +957,162 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+    l1a_dataset = load_cdf(file_path)
+
+    # Update global CDF attributes
+    cdf_attrs = ImapCdfAttributes()
+    cdf_attrs.add_instrument_global_attrs("codice")
+    cdf_attrs.add_instrument_variable_attrs("codice", "l2-lo-direct-events")
+    energy_table, energy_bins = get_lo_de_energy_luts(dependencies)
+    # Convert from position to elevation angle in degrees relative to the spacecraft
+    # axis
+    l2_dataset = l1a_dataset.copy(deep=True)
```
greglucas: You never do anything with the l1a_dataset after this so you don't need to copy it at all. You could just load the l1a file directly into this variable above if you wanted too.
```suggestion
    l2_dataset = l1a_dataset
```

imap_processing/codice/codice_l2.py
```diff
@@ -866,6 +957,162 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+    l2_dataset["elevation_angle"] = (
+        l2_dataset["position"].dims,
+        elevation_angle.astype(np.float32),
+    )
+    # Convert spin_sector to spin_angle in degrees
+    # Use equation from section 11.2.2 of algorithm document
+    # Shift all spin sectors for all positions 13 - 24 adding 12 and mod 24
+    l2_dataset["spin_sector"] = xr.where(
+        (l2_dataset["position"] >= 13) & (l2_dataset["position"] <= 24),
+        (l2_dataset["spin_sector"] + 12) % 24,
```
greglucas, line 1006: Isn't this the same as subtracting 12?

```suggestion
        (l2_dataset["position"] >= 13) & (l2_dataset["position"] <= 24),
        l2_dataset["spin_sector"] - 12,
```

imap_processing/codice/codice_l2.py
```diff
@@ -866,6 +957,162 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+        l2_dataset["spin_sector"],
+    )
+    l2_dataset["spin_angle"] = xr.DataArray(
+        data=l2_dataset["spin_sector"].data * 15.0 + 7.5,
+        dims=l2_dataset["spin_sector"].dims,
+    ).astype(np.float32)
```
greglucas: I think you can do this directly without needing to create an DataArray yourself.

```suggestion
    l2_dataset["spin_angle"] = l2_dataset["spin_sector"].astype(np.float32) * 15.0 + 7.5
```

imap_processing/codice/codice_l2.py
```diff
@@ -866,6 +957,162 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+    l2_dataset["spin_angle"] = xr.where(
+        (l2_dataset["spin_sector"] > 23), np.nan, l2_dataset["spin_angle"]
+    )
```
greglucas, line 1012: Are there spin sectors greater than 24? Just before this you shift everything in bins 13-24 back around to the beginning values. Do you want to do this on the original spin_sector==24 instead of after the shift?

imap_processing/codice/codice_l2.py
```diff
@@ -866,6 +957,162 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+        dims="epoch",
+        attrs=cdf_attrs.get_variable_attributes(
+            "epoch_delta_minus", check_schema=False
+        ),
+    )
+    l2_dataset["epoch_delta_plus"] = xr.DataArray(
+        l2_dataset["epoch_delta_plus"].data.astype(np.int64),
+        dims="epoch",
+        attrs=cdf_attrs.get_variable_attributes("epoch_delta_plus", check_schema=False),
+    )
```
greglucas: For these, should we update the l1a data to be "correct" so you don't have to mess with anything here? I would have expected these to just pass through if you aren't doing anything to the values, but you are changing the datatype and changing the attributes.

imap_processing/codice/codice_l2.py
```diff
@@ -1022,7 +1269,7 @@ def process_codice_l2(
         # These converted variables are *in addition* to the existing L1 variables
         # The other data variables require no changes
         # See section 11.1.2 of algorithm document
-        pass
+        l2_dataset = process_lo_direct_events(dependencies)
```
greglucas, line 1259: Nice :)

greglucas (commented):
Overall looks good, just some minor nits.

imap_processing/codice/codice_l2.py
lacoak21: Yes! Good catch

imap_processing/codice/codice_l2.py
lacoak21: True these lines can be removed. I think they already were in good shape but I copied over some code.

imap_processing/codice/codice_l2.py
lacoak21: I think since I am checking position >= 13 & l2_dataset["position"] <= 24 it wouldnt be the same?

greglucas: Ahh, I see that you are doing position in the comparisons and setting spin sector now. I missed that distinction before. I think you're right here 👍

imap_processing/tests/ialirt/unit/test_process_codice.py
```diff
@@ -675,38 +675,38 @@ def test_process_codice_lo(
     c_over_o_abundance_ratio = np.divide(
         pseudo_density_dict[species[1]] + pseudo_density_dict[species[2]],
         o_abundance_ratio,
-        out=np.zeros_like(o_abundance_ratio, dtype=float),  # fill with 0s by default
+        out=np.full(o_abundance_ratio.shape, np.nan),  # fill with nans by default
```
lacoak21, line 678: @laspsandoval this test was failing because of the new validation files. I guess the zeros were switched to nans. Let me know if this looks alright.

greglucas (approved):

PR #2512, opened 2025-12-15
CoDICE l2 hi DE

### LINKED ISSUE #2059: CoDICE: L2 Implement hi direct events data products

lacoak21:
# Change Summary

## Overview
Processing code for CoDICE l2 DE

closes #2059 

## Updated Files
- imap_processing/cdf/config/imap_codice_l2-hi-direct-events_variable_attrs.yaml
   - CDF attrs 
-imap_processing/codice/codice_l2.py
   - Lut able helper function 
   - L2 hi de processing code (similar structure to lo but different enough to necessitate separate functions. Nothing really overlaps.
   - Constant dictionaries

## Testing
Validation test


imap_processing/codice/codice_l2.py
```diff
@@ -866,6 +907,133 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+        & (ssd_id_flat <= 15)
+        & (ssd_energy_flat != len(energy_table))
+    )
+    # The columns are organized in order of id and gains
+    # E.g. ssd 0 - LG, ssd 0 - MG, ssd 0 - HG, ssd 1 - LG, ssd 1 - MG, ssd 1 - HG, ...
+    cols = ssd_id_flat * 3 + (gain_flat - 1)
+    ssd_energy[valid_mask] = energy_table[ssd_energy_flat[valid_mask], cols[valid_mask]]
+    l2_dataset["ssd_energy"].data = ssd_energy.reshape(
+        l2_dataset["ssd_energy"].shape
+    ).astype(np.float32)
```
greglucas: This flattening then reshaping seems like we could use xarray dimension broadcasting better here somehow to have things work for us without needing to jump through these hoops. I'm not sure exactly what that would look like or how hard it would be to make that work, but with your use of `data.where()` in other PRs as well it seems doable to me. Maybe chalk this one up to a future refactor in Phase E though unless something jumps out to you immediately.

imap_processing/codice/codice_l2.py
```diff
@@ -866,6 +907,133 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+
+    # Convert spin_sector to spin_angle in degrees
+    theta_angles = np.array(
+        [SSD_ID_TO_SPIN_ANGLE.get(ssd_id, np.nan) for ssd_id in ssd_id_flat]
```
greglucas: Maybe related to the above, I wonder if you could make this a helper function instead of a dictionary.

```python
def ssd_id_to_spin_angle(ssd_id: xr.DataArray) -> xr.DataArray
    # flatten/reshape internally if needed, or possibly np.take()?
```

imap_processing/codice/constants.py
```diff
@@ -1140,3 +1140,38 @@
...
+    4: 90.0,
+    5: 64.3,
+    7: 41.4,
+    8: 30.0,
+    9: 41.4,
+    11: 64.3,
+    12: 90.0,
+    13: 115.7,
+    15: 138.6,
+}
```
greglucas: Another idea here is to turn this into a numpy index-array and insert the nan's directly in here in the indexes that are missing.
```python
SSD_ID_TO_ELEVATION = np.array([150.0, 138.6, np.nan, 115.7, 90.0, 64.3, np.nan, ...])
```

lacoak21: This worked really well!! Allowed me to get rid of all the reshaping. Actually some of the flattening and reshaping was totally unnecessary anyway. Not sure what my thought process was there 😄

greglucas (approved):

imap_processing/codice/codice_l2.py
```diff
@@ -866,6 +907,125 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+        cdf_attrs.get_variable_attributes("priority", check_schema=False)
+    )
+    l2_dataset["event_num"].attrs.update(
+        cdf_attrs.get_variable_attributes("event_num", check_schema=False)
+    )
+    l2_dataset["epoch"] = xr.DataArray(
+        l2_dataset["epoch"].data,
+        dims="epoch",
+        attrs=cdf_attrs.get_variable_attributes("epoch", check_schema=False),
+    )
```
tmplummer, line 1247: I'm not sure why epoch is getting re-instantiated using the same data. Otherwise can you follow the above pattern...
```suggestion
    for coord in l2_dataset.coords:
        l2_dataset[coord].attrs.update(cdf_attrs.get_variable_attributes(coord), check_schema=False)
```

tmplummer (approved):
This all looks fine to me. Just one suggestion.

PR #2519, opened 2025-12-17
Change CDF types of L2/L1D variables for MAG

mhairifin:
# Change Summary

## Overview

Some of the MAG variable types are unnecessarily large for the precision of the underlying data. This forces the types of those variables to be more appropriate to the underlying data. Specifically, the magnetic field vectors and magnitude variables will now be CDF_FLOAT rather than CDF_DOUBLE, and the range (which can only be 0,1,2,3) is CDF_UINT1.

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
N/A

## Updated Files

- imap_processing/cdf/config/imap_mag_l2_variable_attrs.yaml
   - Set CDF_DATA_TYPE for range, magnitude, and all vectors to CDF_UINT1, and CDF_FLOAT respectively
- imap_processing/mag/l2/mag_l2_data.py
   - Don't check schema for attributes so the CDF_DATA_TYPE attribute is carried through

## Testing
Updated tests to test the CDF_DATA_TYPE is set for the vectors

### DIFF
--- a/imap_processing/cdf/config/imap_mag_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_mag_l2_variable_attrs.yaml
@@ -50,6 +50,7 @@ vector_attrs: &vectors_default
     LABL_PTR_1: direction_label
     FILLVAL: 9223372036854775807
     FORMAT: F12.5
+    CDF_DATA_TYPE: CDF_FLOAT
     DICT_KEY: "SPASE>Field>FieldQuantity:Magnetic,Qualifier:Vector,CoordinateSystemName:DSRF,CoordinateRepresentation:Cartesian"
 
 # Frame-specific vector attributes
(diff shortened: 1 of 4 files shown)


greglucas:
Interesting, I didn't realize you could put `CDF_DATA_TYPE` in and it would write these differently to disk. Are you OK with the data originally being in-memory with the larger data type and only at the end writing to the lower precision or do you want this to be in-place initially when creating the arrays?

Is this ready for review, it still says draft but looks good to me.

mhairifin:
@greglucas Thanks for taking a look! We are fine with them being larger in memory - we are mostly concerned with generating unnecessarily large data files and our calibration/science data is not precise to the level of a CDF_DOUBLE. We were discussing whether other levels or data types would benefit from similar treatment, which is why this is a draft PR, but I will mark this as ready to review and can make those changes later if needed.

greglucas (approved):

greglucas:
Awesome, I like doing things incrementally to keep everything moving, so that works well for me!

PR #2521, opened 2025-12-17
ULTRA L1c deadtime correction fixes 

lacoak21:
# Change Summary

## Overview
Determine when the rates dataset was in sectored mode by looking at the spin numbers vs checking the params packet. This is more accurate / simpler/ and what the ULTRA IT does. They will be updating the algorithm document. 

## Updated Files
- imap_processing/tests/ultra/unit/conftest.py
   - Remove fixture - l1c does not need the params dataset anymore. I also discovered there was already a fixture for the rates dataset
- imap_processing/ultra/l1c/helio_pset.py and imap_processing/ultra/l1c/spacecraft_pset.py and imap_processing/ultra/l1c/ultra_l1c.py
   - remove params dataset from parameters
  - Update deadtime calculation function. Use the algorithm documents version instead of the powerpoint version. 
  - Determine sectored rates data by checking the spins 
  - Determine spin phase logically instead of with SPICE - we can assume each sector 0 starts at spin phase 0 

## Testing
Fix tests, remove params datasets

### DIFF
--- a/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
+++ b/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
@@ -324,6 +324,7 @@ def get_deadtime_ratios_by_spin_phase(
         # The center spin phase is the closest / most accurate spin phase.
         # There are 24 spin phases per sector so the nominal middle sector spin phases
         # would be: array([ 12., 36., ..., 300., 324.]) for 15 sectors.
+        # We can assume each sector 0 starts at spin phase 0
         spin_phases_centered = (sector_indices / num_spin_sectors) * 360.0 + 12.0
 
     # Create a dataset with spin phases and dead time ratios
--- a/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
+++ b/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
@@ -308,18 +308,18 @@ def get_deadtime_ratios_by_spin_phase(
     else:
         num_spin_sectors = 15
         sector_indices = np.arange(len(sectored_rates["epoch"])) % num_spin_sectors
-        # Get timestamps at the START of each spin (sector 0 of each spin)
+        # Get timestamps at the start of each spin (sector 0)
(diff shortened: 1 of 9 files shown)


imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -192,84 +192,80 @@ def get_deadtime_ratios(sectored_rates_ds: xr.Dataset) -> xr.DataArray:
     ----------
     sectored_rates_ds : xarray.Dataset
         Dataset containing sector mode image rates data.
+    spin_durations : numpy.ndarray
+        Array of spin durations corresponding to each sector.
```
greglucas: Would it make sense to put the `spin_durations` into the dataset container so you don't need a new argument passed in here?

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -192,84 +192,80 @@ def get_deadtime_ratios(sectored_rates_ds: xr.Dataset) -> xr.DataArray:
...
-    ]
 
-    sector_mode_mask = np.logical_or.reduce(conditions)
-    return rates_ds.isel(epoch=sector_mode_mask)
+    # Get the start indices of each sector mode spin
+    sector_starts = spin_change[spin_run_inds]
+    sectored_mode_mask = np.zeros(len(spins), dtype=bool)
+    for start in sector_starts:
+        # Set the group of 15 to True
+        sectored_mode_mask[start : start + 15] = True
```
greglucas: You can avoid the for loop with this broadcasting I think.

```python
starts = np.asarray(sector_starts)
# Create offsets 0..14 and broadcast
idx = starts[:, None] + np.arange(15)
sectored_mode_mask[idx] = True
```

greglucas (approved):
Looks good, feel free to take/leave the comments as you see fit.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
lacoak21: Nice 👍

PR #2522, opened 2025-12-17
ULTRA l1b use aux packet to get spins

lacoak21:
# Change Summary

## Overview
I found another place where the code is using the spin number (unit8) and doing gymnastics with the met time to match up to the universal spin table. We should be using the aux packet and coarse time to get the correct spin duration/ event times/ spin number. I missed that the old routine was still being used in l1b. 

## Updated Files
   - Remove outdated function! Merge it with the new function to keep everything in the same place. 
   - Create a new function that returns the spin and duration using the same logic as the event time calculation
 - imap_processing/ultra/l1c/helio_pset.py and imap_processing/ultra/l1c/spacecraft_pset.py and imap_processing/ultra/l1c/ultra_l1c.py
   - add aux packet as a dependency 

## Testing
Fix tests

### DIFF
--- a/imap_processing/ultra/l1b/ultra_l1b_extended.py
+++ b/imap_processing/ultra/l1b/ultra_l1b_extended.py
@@ -923,7 +923,7 @@ def get_spin_start_indices(aux_dataset: xr.Dataset, de_event_met: NDArray) -> ND
 
     Parameters
     ----------
-    aux_dataset : numpy.ndarray
+    aux_dataset : xarray.Dataset
         Auxiliary dataset containing spin information.
     de_event_met : numpy.ndarray
         Direct event MET.
@@ -968,7 +968,7 @@ def get_event_times(
 
     Parameters
     ----------
-    aux_dataset : numpy.ndarray
+    aux_dataset : xarray.Dataset
         Auxiliary dataset containing spin information.
     phase_angle : numpy.ndarray
         Phase angle.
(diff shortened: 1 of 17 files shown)


imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -968,9 +952,44 @@ def get_event_times(
     start_inds = np.searchsorted(spin_start_sec, de_event_met, side="right") - 1
     # Clip to valid range of indices
     start_inds = np.clip(start_inds, 0, len(spin_start_sec) - 1)
-    # Get the spin numbers for each event
-    spin_numbers = aux_dataset["spinnumber"].values[start_inds]
 
+    return start_inds
+
+
+def get_event_times(
```
greglucas, line 959: Don't you have this function in your other implementation area already? Could you import it from there instead?

greglucas (approved):
Looks good, just one question/comment about code re-use.

imap_processing/ultra/l1b/ultra_l1b_extended.py
lacoak21: That is actually the same function - the git diff is making it seem like I rewrote it. I just updated it a bit. Originally it was returning spin start times, event times, and spin numbers but I made a new function for spin durations and spin numbers.

greglucas: Ahh, OK. Perfect!

PR #2524, opened 2025-12-18
ULTRA l1b add backup method for getting spin information

lacoak21:
# Change Summary

I discovered that sometimes the de data may be associated with spins that are NOT in the aux dataset. I talked to Nick and Larry about this issue and getting the data from the universal spin table is the desired way to deal with this. 

## Updated Files
- updated file 1
- updated file 2

## Testing
Fix test and test new backup method

### DIFF
--- a/imap_processing/ultra/l1b/ultra_l1b_extended.py
+++ b/imap_processing/ultra/l1b/ultra_l1b_extended.py
@@ -13,6 +13,7 @@ from numpy.typing import NDArray
 from scipy.interpolate import LinearNDInterpolator, RegularGridInterpolator
 
 from imap_processing.quality_flags import ImapDEOutliersUltraFlags
+from imap_processing.spice.spin import get_spin_data
 from imap_processing.spice.time import met_to_ttj2000ns, ttj2000ns_to_et
 from imap_processing.ultra.constants import UltraConstants
 from imap_processing.ultra.l1b.lookup_utils import (
@@ -917,7 +918,9 @@ def get_phi_theta(
     return np.degrees(phi), np.degrees(theta)
 
 
-def get_spin_start_indices(aux_dataset: xr.Dataset, de_event_met: NDArray) -> NDArray:
+def get_spin_start_indices(
+    aux_dataset: xr.Dataset, de_event_met: NDArray
+) -> tuple[NDArray, NDArray]:
     """
     Get the spin start indices in the aux dataset for each event.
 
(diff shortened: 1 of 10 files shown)


greglucas (commented):
It seems like we would want to use one or the other, but not switch between two as that could confuse things later on. If we don't have enough aux packet data, do we need to increase the request period on batch_starter to include some earlier packets for Ultra? In the opposite direction, if we always have spin table data, could we always reference that table instead of mixing and matching the two different sources of data?

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -1023,19 +1064,40 @@ def get_spin_and_duration(
...
-    return spin_numbers, spin_durations
+    start_inds, missing_events = get_spin_start_indices(aux_dataset, de_event_met)
+    # Initialize spin info dataset
+    spin_info_per_event = xr.Dataset()
+    # Create dict of var name lookups
+    var_names = {
+        "spin_number": ("spinnumber", "spin_number"),
+        "spin_duration": ("duration", "spin_period_sec"),
+        "spin_starts": ("timespinstart", "spin_start_sec_sclk"),
+        "spin_start_subs": ("timespinstartsub", "spin_start_subsec_sclk"),
```
tmplummer, line 1046: I have a slight concern that the conversion to MET is different. It looked like in the aux source, `timespinstartsub` is milliseconds whereas in the universal spin-table, `spin_start_subsec_sclk` is microseconds. You can however use the already converted `spin_start_met` column in the dataframe which is floating point MET seconds.

imap_processing/ultra/l1b/ultra_l1b_extended.py
```diff
@@ -932,32 +935,76 @@ def get_spin_start_indices(aux_dataset: xr.Dataset, de_event_met: NDArray) -> ND
...
+            spin_start_sec, de_event_met[~missing_aux_data_mask], side="right"
+        )
+        - 1
+    )
 
-    return start_inds
+    return start_inds, missing_aux_data_mask
+
+
+def get_spin_data_from_universal_spin_table(de_event_met: NDArray) -> pandas.DataFrame:
```
tmplummer: Don't know if you saw the `spin.interpolate_spin_data` function. It has much the same functionality as this function but maybe you didn't want the extras in there (calculation of spin-phase)?

tmplummer (commented):
Looks good except for the one concern I have about the units of subseconds in the two different sources for spin info.

imap_processing/ultra/l1b/ultra_l1b_extended.py
lacoak21: Ah great catch! I missed that it was microseconds.

imap_processing/ultra/l1b/ultra_l1b_extended.py
lacoak21: Oh wow - I had assumed incorrectly that this was doing something different. Thanks!

greglucas (approved):

tmplummer (approved):

PR #2527, opened 2025-12-19
CoDICE lo counters single inst_az_label fix

lacoak21:
# Change Summary
Fix labels. Request from CoDICE team.

### DIFF
--- a/imap_processing/codice/codice_l1a_lo_counters_singles.py
+++ b/imap_processing/codice/codice_l1a_lo_counters_singles.py
@@ -164,7 +164,7 @@ def l1a_lo_counters_singles(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.
                 attrs=cdf_attrs.get_variable_attributes("inst_az", check_schema=False),
             ),
             "inst_az_label": xr.DataArray(
-                np.arange(inst_az, dtype=np.uint8).astype(str),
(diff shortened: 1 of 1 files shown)


imap_processing/codice/codice_l1a_lo_counters_singles.py
```diff
@@ -164,7 +164,7 @@ def l1a_lo_counters_singles(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.
                 attrs=cdf_attrs.get_variable_attributes("inst_az", check_schema=False),
             ),
             "inst_az_label": xr.DataArray(
-                np.arange(inst_az, dtype=np.uint8).astype(str),
+                (np.arange(inst_az, dtype=np.uint8) + 1).astype(str),
```
greglucas, line 167: I feel like this might get confusing why the label is different from the values? Do they want the inst_az values being 0-based indices but the labels being 1-based, or instead should they both be 1-based?

lacoak21: The positions themselves are zero based but they are referred to / known as 1 based.. I know that is confusing but I brought this up in the CoDICE tag up this morning and Micael confirmed that this was the desired label.

greglucas (approved):

PR #2529, opened 2025-12-19
2510 hi goodtimes   Add xarray accessor for handling goodtimes

### LINKED ISSUE #2528: Hi Goodtimes - create goodtimes dataset helper code
### Algorithm Description:

Write code to initialize and keep track of goodtimes through the goodtimes checks. Also include code to write out Goodtimes text file.

tmplummer:
# Change Summary

## Overview
This just contains some foundational code to be used throughout the Hi Goodtimes processing. Basically, I just added a custom xarray.Dataset accessor that has some functions useful for keeping track of the Goodtimes for a Pointing as various checks are performed to identify what times to remove from the goodtimes dataset.

The goodtimes dataset is initialized using a hi_l1a_de dataset and starts with all times in the Pointing considered good. Through the goodtimes processing, checks will be made that mark 8-spin sets as bad. Finally, the Goodtimes text file will be written out.

## New Files
- imap_processing/hi/hi_goodtimes.py
   - Add CullCode enum
   - Add function that initializes the goodtimes dataset from a l1a_de dataset.
   - Add goodtimes accessor that has functions for:
      - remove_times
      - get_good_intervals
      - get_cull_statistics (may not be used but Claude implemented it and it could be useful for logging)
      - write_txt
- imap_processing/tests/hi/test_hi_goodtimes.py
   - Full test coverage for new hi goodtimes code

Closes: #2528

### DIFF
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -17,9 +17,9 @@ INTERVAL_DTYPE = np.dtype(
     [
         ("met_start", np.float64),
         ("met_end", np.float64),
-        ("spin_bin_low", np.int32),
-        ("spin_bin_high", np.int32),
-        ("n_good_bins", np.int32),
+        ("spin_bin_low", np.uint8),
+        ("spin_bin_high", np.uint8),
+        ("n_good_bins", np.uint8),
         ("esa_step", np.uint8),
     ]
 )
@@ -423,8 +423,8 @@ class GoodtimesAccessor:
             - cull_code_counts: Dict mapping cull codes to counts
         """
         total_bins = self._obj["cull_flags"].size
-        good_bins = int(np.sum(self._obj["cull_flags"].values == 0))
-        culled_bins = total_bins - good_bins
+        culled_bins = np.count_nonzero(self._obj["cull_flags"])
+        good_bins = total_bins - culled_bins
 
         # Count occurrences of each cull code
         unique_codes, counts = np.unique(
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -251,6 +251,20 @@ class GoodtimesAccessor:
 
         met_values = self._obj.coords["met"].values
 
+        # check for met times out of range
+        met_array = np.atleast_1d(met)
+        # Add the difference between the last two MET values to the valid range
+        # to get the time of the last MET + 8_spins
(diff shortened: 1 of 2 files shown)


imap_processing/hi/hi_goodtimes.py
```diff
@@ -0,0 +1,488 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Structured dtype for good time intervals
+INTERVAL_DTYPE = np.dtype(
+    [
+        ("met_start", np.float64),
+        ("met_end", np.float64),
+        ("spin_bin_low", np.int32),
+        ("spin_bin_high", np.int32),
```
greglucas: uint16? I think we only have 3600 spin bins and they are all positive integers.

imap_processing/hi/hi_goodtimes.py
```diff
@@ -0,0 +1,488 @@
...
+        if bins is None:
+            # Flag all spin bins (0-89)
+            bins_array = np.arange(90)
+        else:
+            # Convert to array for consistent handling
+            bins_array = np.atleast_1d(bins)
+
+        # Validate bin indices
+        if np.any((bins_array < 0) | (bins_array >= 90)):
+            raise ValueError("Spin bins must be in range [0, 89]")
```
greglucas, line 250: Lo is using 60 spin bins (6 degree bins), does this need to be similar between the different instruments?

imap_processing/hi/hi_goodtimes.py
```diff
@@ -0,0 +1,488 @@
...
+        dict
+            Dictionary with cull code statistics:
+            - total_bins: Total number of MET × spin_bin combinations
+            - good_bins: Number of bins with cull_flags=0
+            - culled_bins: Number of bins with cull_flags>0
+            - fraction_good: Fraction of bins that are good
+            - cull_code_counts: Dict mapping cull codes to counts
+        """
+        total_bins = self._obj["cull_flags"].size
+        good_bins = int(np.sum(self._obj["cull_flags"].values == 0))
```
greglucas: You can invert this to calculate the culled_bins first and then get the good bins with a single function like `np.count_nonzero`

greglucas (approved):
This is a nice way to do this with the xarray accessor!

This makes me think that we could potentially use accessors within the map projection code area. Just noting this for potential future ideas in refactoring later on.

imap_processing/hi/hi_goodtimes.py
tmplummer: I don't think so. For Hi, this corresponds to the number of bins in the histogram packets and thus cannot be changed.

PR #2530, opened 2025-12-19
ULTRA L1b mask out invalid times

lacoak21:
# Change Summary

## Overview
Event times should not be calculated for events at "valid_mask". These will have phase_angle fill vals and the event time calculation will be incorrect. The old code used to do this so I basically reverted back to how we used to compute them. 

valid_mask = de_dataset["start_type"].data != FILLVAL_UINT8

## Updated Files
      - revert back code


greglucas (approved):
I don't think you need the initial copy since it is already creating a new array on that line.

The valid events of valid events is a bit confusing at first reading. Could use a comment about only updating where true or something like that.

PR #2531, opened 2025-12-22
MNT/FIX: Generalize Lo resweep handling

greglucas:
# Change Summary

## Overview
This is needed by other data products too, so we can pull out the ancillary file handling to get the resweep values into a helper function.

Change from for-loops and setting individual elements to numpy vectorization for resweeping.

Fix exposure time calculation that was incorrectly assigning to many ESA levels during a sweep. We should only have 7 total values in the exposure factor calculation of esa_sweeps.


tmplummer (approved):
LGTM

PR #2533, opened 2025-12-22
Update method used to find mean spin-axis for Pointing to just averag…

### LINKED ISSUE #2532: BUG - pointing attitude: mean spin-axis is wrong
### Description of the issue

Analysis by Marek from GLOWS shows that the quaternion averaging is not working as expected.

https://nam10.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdocs.google.com%2Fdocument%2Fd%2F1MpSZ_BtSXrfH_MmblLIzEsAtsL0x0tn2gm2q2Nvnwx0%2Fedit%3Fusp%3Dsharing&data=05%7C02%7CGreg.Lucas%40lasp.colorado.edu%7Cc9697f6050f44dd958b108de3f44c644%7Cb1ddd7831dbc446b878fa7b03478ea9e%7C1%7C0%7C639017761953180176%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=GCQAKRI9cN6tf%2FVvWFGFBWcRpAA4JQB4aa5QaXX6s8w%3D&reserved=0

_No response_

_No response_

### Code Snippet:

```python
Take the mean of the instantaneous spin-axes expressed in Cartesian
```

_No response_

tmplummer:
…e instantaneous z-axes across Pointing

# Change Summary

## Overview
Instead of averaging quaternions, just average the instantaneous z-axis expressed in HAE Cartesian coordinates.

## Updated Files
   - Change `_average_quaternions` to `_mean_spin_axis` and update functionality to do as the name change implies.
   - `_create_rotation_matrix` can now just use the mean z-axis as computed.
   - Update tests

Closes: #2532

### DIFF
--- a/imap_processing/spice/pointing_frame.py
+++ b/imap_processing/spice/pointing_frame.py
@@ -11,7 +11,7 @@ import spiceypy
 from imap_data_access import SPICEFilePath
 from numpy.typing import NDArray
 
-from imap_processing.spice.geometry import SpiceFrame
+from imap_processing.spice.geometry import SpiceFrame, frame_transform
 from imap_processing.spice.repoint import get_repoint_data
 from imap_processing.spice.time import (
     TICK_DURATION,
@@ -295,11 +295,11 @@ def calculate_pointing_attitude_segments(
             int(num_samples),
         )
 
-        # Get the average quaternions for the pointing
-        q_avg = _average_quaternions(et_times)
+        # Get the average spin-axis in HAE coordinates
+        z_avg = _mean_spin_axis(et_times)
 
         # Create a rotation matrix
-        rotation_matrix = _create_rotation_matrix(q_avg)
+        rotation_matrix = _create_rotation_matrix(z_avg)
 
         # Convert the rotation matrix to a quaternion.
         # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.m2q
@@ -317,9 +317,12 @@ def calculate_pointing_attitude_segments(
     return pointing_segments
 
 
-def _average_quaternions(et_times: np.ndarray) -> NDArray:
+def _mean_spin_axis(et_times: np.ndarray) -> NDArray:
     """
-    Average the quaternions.
+    Compute the mean spin axis for a given time range.
+
+    The mean spin-axis is computed by taking the mean of the instantaneous
+    spin-axes evaluated at each input ET time.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


imap_processing/tests/spice/test_pointing_frame.py
greglucas, line 1: Is there a reason the numbers in the test didn't get updated? I'm wondering if there is some way we can verify we are using the "new" method rather than the previous method and be certain we don't go back to the previous way.

imap_processing/spice/pointing_frame.py
```diff
@@ -317,9 +317,12 @@ def calculate_pointing_attitude_segments(
     return pointing_segments
 
 
-def _average_quaternions(et_times: np.ndarray) -> NDArray:
+def _mean_spin_axis(et_times: np.ndarray) -> NDArray:
     """
-    Average the quaternions.
+    Compute the mean spin axis for a given time range.
+
+    The mean spin-axis is computed by taking the mean of the instantaneous
```
lacoak21: Can you explain more about what the instantaneous spin axis are?

imap_processing/spice/pointing_frame.py
```diff
@@ -328,69 +331,40 @@ def _average_quaternions(et_times: np.ndarray) -> NDArray:
...
+    # data is built and sampled, there may or may not be aliasing with this
+    # approach.  If it turns out that we need to pull the quaternions
+    # directly from the CK there are several routines that exist to do this
+    # but it's not straight forward.  We'll revisit this if needed.
+    z_inertial_hae = frame_transform(
+        et_times, np.array([0, 0, 1]), SpiceFrame.IMAP_SPACECRAFT, SpiceFrame.ECLIPJ2000
+    )
+
+    # Compute the average spin axis by averaging each component across time
+    z_avg = np.mean(z_inertial_hae, axis=0)
```
greglucas, line 350: A couple of potential thoughts/questions here.

What is `et_times` being requested as? Would it make sense to sample this at a higher frequency under the hood here, to try and get closer to the actual quaternions that were input.

Should we compute the standard deviation and log/warn if it is outside of some bound for awareness purposes?

greglucas (approved):
Nice, it actually looks like a good simplification here too.

I added a few nitpicky comments, but the overall logic looks good to me!

imap_processing/spice/pointing_frame.py
tmplummer: As it is now, we are sampling ~10 times per spin. IMO, ideally, we should be pulling the quaternions straight from the attitude history files which I think are 1Hz and would give ~15/spin.

I could pull spin 70 data and add a test that checks that we get close to what GLOWS gets.

imap_processing/tests/spice/test_pointing_frame.py
tmplummer: I think that the test data may be degenerate in that Nick assumed that the spacecraft z-axis would be close to the true spin-axis.

imap_processing/spice/pointing_frame.py
greglucas: > IMO, ideally, we should be pulling the quaternions straight from the attitude history files which I think are 1Hz and would give ~15/spin

If you sample the data at 1Hz do we exactly get back the instantaneous value based on the quaternion, or is it somehow smoothed even on the exact data points so that we have a smooth curve but not match the "true" data? i.e. I'm wondering if we can just adjust our sampling period here to be what the quaternions are included at rather than this slight offset to avoid any interpolation/sampling.

imap_processing/spice/pointing_frame.py
```diff
@@ -328,69 +331,40 @@ def _average_quaternions(et_times: np.ndarray) -> NDArray:
...
     rotation_matrix : np.ndarray
         Rotation matrix.
     """
-    # Converts the averaged quaternion (q_avg) into a rotation matrix
-    # and get inertial z axis.
-    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.q2m
-    z_avg = spiceypy.q2m(list(q_avg))[:, 2]
-    # y_avg is perpendicular to both z_avg and the standard Z-axis.
+    # y_avg is perpendicular to both z_avg and the HAE Z-axis.
     y_avg = np.cross(z_avg, [0, 0, 1])
```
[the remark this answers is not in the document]
tmplummer: The spacecraft z-axis will never be close to the ecliptic north-pole, so this is not a concern.

imap_processing/tests/spice/test_pointing_frame.py
tmplummer: Test data updated to use flight data. Values match what I can read off of the GLOWS plots.

PR #2534, opened 2025-12-22
Hard code IMAP_SPACECRAFT ID and IMAP_SCLK ID

tmplummer:
# Change Summary

## Overview
I don't know why these errors started showing up after the PR to fix how the mean spin axis was merged. I also can't reproduce locally. The spacecraft ID and clock ID should never change, so I don't think that it is a problem to hard code these. They are used as hard coded values elsewhere anyway.


greglucas (approved):
Did one of our kernels change under the hood without us realizing it?

PR #2541, opened 2025-12-26
Lo: CG Map Normalizing 'energy_sc'

pleasant-menlo:
# Change Summary
We found an issue when trying to generate L3 HF maps from Lo HF maps. We found that the energy_sc values being computed were far larger than we would expect. All energy_sc values were being summed together and then not normalized after being projected into the map. We chose to do this with an exposure weighting to match Hi, but we are not sure if this is what the Lo team actually wants. It did yield a map that has reasonable looking values. 

You'll notice that there is no test coverage for this fix, as we don't intend for this PR to be merged as is but we didn't want it to get lost over the holiday break.

## Overview
In lo_l2.py, before projecting pset values to a map, multiply the energy_sc by exposure_factor. After projection, divide the exposure_factor back out.

### DIFF
--- a/imap_processing/lo/l2/lo_l2.py
+++ b/imap_processing/lo/l2/lo_l2.py
@@ -875,8 +875,8 @@ def calculate_all_rates_and_intensities(
         logger.info("Interpolating map intensities to helio-frame energies")
         dataset = interpolate_map_flux_to_helio_frame(
             dataset,
-            dataset["energy"],
-            dataset["energy"],
+            dataset["energy"] * 1000,
+            dataset["energy"] * 1000,
             ["ena_intensity", "bg_intensity"],
         )
 
--- a/imap_processing/lo/l2/lo_l2.py
+++ b/imap_processing/lo/l2/lo_l2.py
@@ -94,6 +94,10 @@ def lo_l2(
 
     logger.info("Step 3: Converting to dataset and adding geometric factors")
     dataset = sky_map.to_dataset()
+
+    if cg_correction:
+        dataset["energy_sc"] /= dataset["exposure_factor"]
+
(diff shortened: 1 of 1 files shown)


greglucas:
In the Lo code we have been projecting fields with different names and then after the fact removing/renormalizing those quantities, so it is clear whether something is efficiency_squared weighted vs efficiency weighted etc. I'm not sure if this should follow the same pattern here, rather than needing to remember if/when the `sc_energy` is exposure weighted or not. I guess this means that Hi is already exposure weighted when they do the projection? I agree though that we don't want a straight summation of these values when projected, so this idea seems reasonable to me.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/4c3c13510dcd2af77e120c48fb1ee727cc9659d4/imap_processing/lo/l2/lo_l2.py#L524-L534

imap_processing/lo/l2/lo_l2.py
```diff
@@ -869,8 +875,8 @@ def calculate_all_rates_and_intensities(
         logger.info("Interpolating map intensities to helio-frame energies")
         dataset = interpolate_map_flux_to_helio_frame(
             dataset,
-            dataset["energy"],
-            dataset["energy"],
+            dataset["energy"] * 1000,
```
tmplummer, line 878: I think that the documentation for `interpolate_map_flux_to_helio_frame` is wrong when it says that input energies must be in units of eV. It really just matters that the two inputs (`esa_energy` and `hf_energy`) are in the same units. This is because all the equations use ratios of energies so as long as the units are the same, it should produce the same results.

tmplummer:
@pleasant-menlo, this has been taken care of in two separate PRs. See: #2549 #2552

greglucas:
Closing this, since it has already been implemented separately.

Issue #2540, opened 2025-12-23
Hi Goodtimes - impliment goodtimes algs from culling.c


tmplummer:
### Algorithm Description:

Paul Janzen provided culling.c code for initial goodtimes algorithms. This ticket is to take that code, convert to python and make it production worthy.

PR #2553, opened 2026-01-06
2540 hi goodtimes   impliment goodtimes algs from cullingc

tmplummer:
# Change Summary

## Overview
Adds the following goodtimes check functions and test coverage:
- drop_incomplete_spin_sets
- drop_drf_times

## Updated Files
- imap_processing/hi/hi_goodtimes.py
   - create_goodtimes_dataset function updated to include all unique MET times. Times with missing packets are removed in the new `drop_incomplete_spin_sets` function.
   - Add `drop_incomplete_spin_sets` function.
   - Add `drop_drf_times function
- imap_processing/tests/hi/test_hi_goodtimes.py
   - Add test coverage for new functions

Part 1 of #2540


greglucas (commented):
The code looks good to me. One thing I was slightly confused by is the naming though. Generally "drop" means remove it from the dataset to me, so I was a bit confused how things were being updated and what was being returned vs in-place. I _think_ what is happening though is you are identifying/labeling bad times, not actually removing them from the dataset. So now those are "marked" with the cull_flags. In the tests you're comparing that the cull_flags were set, not that those items aren't present anymore.

What about `mark_incomplete_spins()` or something like that for naming rather than drop/filter?

tmplummer:
@greglucas 

> The code looks good to me. One thing I was slightly confused by is the naming though. Generally "drop" means remove it from the dataset to me, so I was a bit confused how things were being updated and what was being returned vs in-place. I _think_ what is happening though is you are identifying/labeling bad times, not actually removing them from the dataset. So now those are "marked" with the cull_flags. In the tests you're comparing that the cull_flags were set, not that those items aren't present anymore.
> 
> What about `mark_incomplete_spins()` or something like that for naming rather than drop/filter?

Yes, I see the confusion there. The goodtimes object is really just marking times that are not good, as you pointed out. Then when the text file gets written, only the remaining unmarked goodtimes are written out. I will update as you suggested.

greglucas (approved):

Issue #2518, opened 2025-12-17
BUG - MAG L1D jobs failing and duplicate runs


alastairtree:
### Description of the issue

The MAG L1D processing seems to have issues, for example Dec 11th there are no L1D files despite us having data. The job failure logs show a couple of issues:
- L1D Jobs seems to be running more than once, starting at the same time for same duration. Some jobs do the upload while the duplicate jobs shows skipped uploads
- errors relating to the spin data not being available that then run later - is L1D running too soon?
- Multiple versions for L1D files where there does not need to be so many versions

Sample of the errors:

```
/usr/local/lib/python3.12/site-packages/xarray/namedarray/core.py:261: UserWarning: Duplicate dimension names present: dimensions {'dim0'} appear more than once in dims=('dim0', 'dim0'). We do not yet support duplicate dimension names, but we do allow initial construction of the object. We recommend you rename the dims immediately to become distinct, as most xarray functionality is likely to fail silently if you do not. To rename the dimensions you will need to set the ``.dims`` attribute of each variable, ``e.g. var.dims=('x0', 'x1')``.
  self._dims = self._parse_dimensions(dims)
/usr/local/lib/python3.12/site-packages/imap_processing/spice/spin.py:210: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead
See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  out_df["sc_spin_phase"] = spin_phases
```

```
2025-12-12 20:49:49 - WARNING:cdflib.logging:ISTP Compliance Warning: range is listed as the DEPEND_0 for variable URFTOORFO, but the dimensions do not match.
2025-12-12 20:49:49 - WARNING:cdflib.logging:ISTP Compliance Warning: range is listed as the DEPEND_0 for variable URFTOORFI, but the dimensions do not match.
2025-12-12 20:49:49 - WARNING:cdflib.logging:Warning: Variable valid_start_datetime listed DEPEND_0 as , but no variable by that name was found.
2025-12-12 20:49:49 - WARNING:cdflib.logging:Warning: Variable gradiometer_factor listed DEPEND_0 as , but no variable by that name was found.
[...]

PR #2554, opened 2026-01-06
Optimize MAG L1D functions to reduce SPICE tranform overhead

maxinelasp:
# Change Summary
This addresses #2518 and includes some fixes described in that ticket. 

## Overview

1. Since the SPICE transform calls take so much time, the code now transforms to J2000 as a hub for data and then goes from there to each frame to reduce the number of dynamic rotations
2. Switching out apply_along_axis for more efficient numpy calls

This showed about a 20% improvement.

### DIFF
--- a/imap_processing/mag/l1d/mag_l1d.py
+++ b/imap_processing/mag/l1d/mag_l1d.py
@@ -1,5 +1,8 @@
 """Module for generating Level 1d magnetic field data."""
 
+from copy import deepcopy
+from itertools import product
+
 import numpy as np
 import xarray as xr
 
@@ -99,19 +102,15 @@ def mag_l1d(  # noqa: PLR0912
 
     # Nominally, this is expected to create MAGO data. However, if the configuration
     # setting for always_output_mago is set to False, it will create MAGI data.
+    l1d_norm.rotate_frame(ValidFrames.J2000)
 
-    l1d_norm.rotate_frame(ValidFrames.SRF)
-    norm_srf_dataset = l1d_norm.generate_dataset(attributes, day_to_process)
-    l1d_norm.rotate_frame(ValidFrames.DSRF)
-    norm_dsrf_dataset = l1d_norm.generate_dataset(attributes, day_to_process)
-    l1d_norm.rotate_frame(ValidFrames.GSE)
-    norm_gse_dataset = l1d_norm.generate_dataset(attributes, day_to_process)
-    l1d_norm.rotate_frame(ValidFrames.RTN)
-    norm_rtn_dataset = l1d_norm.generate_dataset(attributes, day_to_process)
-    output_datasets.append(norm_srf_dataset)
-    output_datasets.append(norm_dsrf_dataset)
-    output_datasets.append(norm_gse_dataset)
-    output_datasets.append(norm_rtn_dataset)
+    output_frames = [
+        ValidFrames.SRF,
+        ValidFrames.DSRF,
+        ValidFrames.GSE,
+        ValidFrames.RTN,
+    ]
+    input_datasets = [l1d_norm]
 
     if input_mago_burst is not None and input_magi_burst is not None:
         # If burst data is provided, use it to create the burst L1d dataset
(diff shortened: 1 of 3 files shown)


imap_processing/mag/l1d/mag_l1d.py
```diff
@@ -99,19 +102,15 @@ def mag_l1d(  # noqa: PLR0912
 
     # Nominally, this is expected to create MAGO data. However, if the configuration
     # setting for always_output_mago is set to False, it will create MAGI data.
+    l1d_norm.rotate_frame(ValidFrames.J2000)
```
tmplummer, line 105: I know that it isn't quite a clean, but I think you would see another incremental speed up by first rotating to the SRF frame, storing that data and then going to J2000 before doing the rest as you are.

As implemented, you are not taking advantage of the fixed rotation code you added for going direct from MAGI/MAGO to SRF.

imap_processing/mag/l1d/mag_l1d.py
```diff
@@ -173,3 +166,39 @@ def mag_l1d(  # noqa: PLR0912
...
+        Attributes instance with the correct MAG L1D from ImapCdfAttributes.
+    day_to_process : np.datetime64
+        The day to process - this will crop the data to the exact 24 hours provided.
+
+    Returns
+    -------
+    xr.Dataset
+        The resulting dataset.
+    """
+    l1d_copy = deepcopy(l1d_input)
```
tmplummer, line 199: Is this deep copy of the MagL1d object copying lots of data other than the vectors that get rotated? It might be less expensive to just store the J2000 cordinates rather than copying the entire object.

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -289,14 +289,25 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
...
+        single_mago_epoch = None
+        single_magi_epoch = None
+        if (
+            start_frame in (ValidFrames.MAGO, ValidFrames.MAGI)
+            and end_frame == ValidFrames.SRF
+        ):
+            # Since Instrument -> spacecraft rotations are static, we can optimize
+            # by only passing one time into frame_transform.
+            single_mago_epoch = self.epoch_et[0]
+            single_magi_epoch = self.magi_epoch_et[0]
```
tmplummer, line 307: This logic looks good to me.

tmplummer (commented):
Glad that you got some improvement. I suspect that the you could get another modest bump based on one of my comments.

I'd be interested to hear what you tried for multi-processing. I have some vague ideas about how we might be able to implement something that would get the spice multi-processing to work. I am thinking along the lines of running each of the transforms from J2000 -> DYNAMIC_FRAME in a separate process.

imap_processing/mag/l1d/mag_l1d.py
maxinelasp: That's because actually, at this point in the process, the frame is DSRF, not MAGO. Internally, the code spins from MAGO/MAGI -> SRF -> DSRF as part of processing, so that fixed rotation code is being used as part of l1d_norm generation.

imap_processing/mag/l1d/mag_l1d.py
maxinelasp: The vast majority of the data in the classes is the vectors/timestamps. It is true that there is some unnecessary copying going on here, but... since there is a fair bit of logic about MAGO/MAGI and selecting which one to use inside each class, I am hesitant to pull that out and duplicate it here.

maxinelasp:
@subagonsouth 

> I'd be interested to hear what you tried for multi-processing. I have some vague ideas about how we might be able to implement something that would get the spice multi-processing to work. I am thinking along the lines of running each of the transforms from J2000 -> DYNAMIC_FRAME in a separate process.

That's basically what I did. I used the concurrent.futures framework to basically run each of the instances inside the `for l1d, frame in product(input_datasets, output_frames):` in parallel. Unfortunately, it repeatedly crashed with `pxform`/SPICE errors before completing any of the runs. I wasn't able to get it working.

imap_processing/mag/l1d/mag_l1d.py
tmplummer: I see. Ok, so there is still some room for optimization.
- You could store the SRF values the first time that transform occurs.
- You could try to avoid doing the SRF -> DSRF. That transform traverses two dynamic frame transforms: SRF -> J2000  and then J2000 -> DSRF.

I will approve so you can get this optimization released but it might be worth capturing this additional possibility in a ticket.

tmplummer (approved):
Nice incremental speedup.

tmplummer:
> That's basically what I did. I used the concurrent.futures framework to basically run each of the instances inside the `for l1d, frame in product(input_datasets, output_frames):` in parallel. Unfortunately, it repeatedly crashed with `pxform`/SPICE errors before completing any of the runs. I wasn't able to get it working.

Interesting. Did you do any kernel furnishing as part of a setup for each process?

maxinelasp:
> Interesting. Did you do any kernel furnishing as part of a setup for each process?

No, I didn't specifically furnish kernels

PR #2559, opened 2026-01-07
2555 bug   spin phase calculation reports nan

### LINKED ISSUE #2555: BUG - Spin phase calculation reports NaN
### Description of the issue

For certain time queries, using either the `spice.spin.interpolate_spin_data`, `get_spacecraft_spin_phase`, or `get_instrument_spin_phase` functions, it is possible to get NaN results for the spin phase when it should be valid.

Notice in the plot that sometimes, the spin_period_sec for a given spin is less than the difference in spin_start_met times. This means that the `interpolate_spin_data` function is returning the correct row for the queried met time, but when computing the spin-phase it is possible to get a value > 1.

Spin phase equation:
`spin_phase = (query_met_time - spin_start_met) / spin_period_sec`

Image

```
from imap_processing.spice.spin import set_global_spin_table_paths, get_spin_data, get_spacecraft_spin_phase, interpolate_spin_data
set_global_spin_table_paths([spin_table_path / "imap_2025_355_2025_356_01.spin"])
spin_phase = get_spacecraft_spin_phase(504019261.9898615)
print_spin_phase
spin_data = interpolate_spin_data(504019261.9898615)
print(spin_data)
```

A valid spin phase should be computed.

NaN is returned.

### Code Snippet:

```python
Code
```

This issue is caused by the spin_period_sec values for spin number `i` not being exactly equal to spin_start_met[i + 1] - spin_start_met[i]. See plot. There are cases where the correct spin is interpolated, but the spin phase calculated using the equation in `interpolate_spin_data` results in a spin-phase that is slightly greater than 1.

Possible fixes:
- Clip the spin phase to a max value of 1.0 (currently spin-phase is strictly < 1)
- Mod the spin-phase with 1. The result would be sometimes a spin-phase very close to zero even though the time is > spin_start_met + spin_period_sec
- Detect these occurrences and use the difference between current spin_start_met and next spin_start_met. If we are in the last spin of the table, this would not work.

tmplummer:
# Change Summary

## Overview
This fixes the bug in computing spin-phase when query time is after spin_start_met + spin_period_sec but before the next spin_start_met time.
Bonus: It also adds some optimization and fixes the settingwithcopywarning that is clogging many logs.

## Updated Files
- imap_processing/spice/spin.py
   - Precomute actual_spin_period in cache function where spin # increments by one. Elsewhere, use spin_period_sec
   - Use copy when interpolating the spin dataframe to query times to avoid getting the SettingWithCopyWarning
   - Compute spin-phase using above computed actual_spin_period
- imap_processing/tests/spice/test_spin.py
   - Add test coverage that specifically tests the bug this ticket fixes
   - Update other tests as needed where spin_period_sec was being relied upon to compute spin-phase.

Closes: #2555

### DIFF
--- a/imap_processing/spice/spin.py
+++ b/imap_processing/spice/spin.py
@@ -184,7 +184,7 @@ def interpolate_spin_data(query_met_times: float | npt.NDArray) -> pd.DataFrame:
 
     # Make sure input times are within the bounds of spin data
     spin_df_start_time = spin_start_met[0]
-    spin_df_end_time = spin_start_met[-1] + spin_df["spin_period_sec"].values[-1]
+    spin_df_end_time = spin_start_met[-1] + actual_spin_periods[-1]
     input_start_time = query_met_times.min()
     input_end_time = query_met_times.max()
     if input_start_time < spin_df_start_time or input_end_time >= spin_df_end_time:
--- a/imap_processing/spice/spin.py
+++ b/imap_processing/spice/spin.py
@@ -54,7 +54,9 @@ def get_spin_data() -> pd.DataFrame:
             * `spin_start_subsec_sclk`: MET microseconds of spin start time.
             * `spin_start_met`: Floating point MET seconds of spin start.
             * `spin_start_utc`: UTC string of spin start time.
-            * `spin_period_sec`: Floating point spin period in seconds.
+            * `spin_period_sec`: Floating point spin period in seconds (estimated).
+            * `actual_spin_period`: Floating point actual spin period computed from
+              consecutive spin start times. More accurate than spin_period_sec.
             * `spin_period_valid`: Boolean indicating whether spin period is valid.
             * `spin_phase_valid`: Boolean indicating whether spin phase is valid.
             * `spin_period_source`: Source used for determining spin period.
@@ -106,6 +108,7 @@ def _load_spin_data_with_cache(csv_paths: tuple[Path]) -> pd.DataFrame:
                 "spin_start_utc": str,
                 "spin_period_sec": float,
                 "spin_period_valid": bool,
+                "spin_phase_valid": bool,
                 "spin_period_source": int,
                 "thruster_firing": bool,
             },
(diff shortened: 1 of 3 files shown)


imap_processing/tests/spice/test_spin.py
```diff
@@ -123,18 +127,24 @@ def test_get_spin_number(fake_spin_data, met_time, spin_number_expected):
...
             np.array([0.5, 0, np.nan, 0, np.nan, np.nan, 1 / 15]),
         ),
         # Test that this spin phase range [0, 1) is valid which
         # is same as [0, 360) degree angle. At 15 seconds the spacecraft
         # has completed a full spin
         (np.array([0, 15]), np.zeros(2)),
+        # Test gap between estimated spin end and actual next spin start
+        # Spin 9: start=135, spin_period_sec=14.5, estimated_end=149.5
+        # Spin 10: start=150 (actual start is 0.5s after estimated end)
+        # Query at 149.7 should be valid: (149.7-135)/15 = 0.98 < 1
```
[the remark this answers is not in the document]
tmplummer: Nope, I am happy with the comment as is.

imap_processing/spice/spin.py
```diff
@@ -159,11 +178,13 @@ def interpolate_spin_data(query_met_times: float | npt.NDArray) -> pd.DataFrame:
...
+    spin_start_met = spin_df["spin_start_met"].values
+    actual_spin_periods = spin_df["actual_spin_period"].values
+
     # Make sure input times are within the bounds of spin data
-    spin_df_start_time = spin_df["spin_start_met"].values[0]
-    spin_df_end_time = (
-        spin_df["spin_start_met"].values[-1] + spin_df["spin_period_sec"].values[-1]
-    )
+    spin_df_start_time = spin_start_met[0]
+    spin_df_end_time = spin_start_met[-1] + spin_df["spin_period_sec"].values[-1]
```
greglucas: Do you want to reference your actual_spin_periods value here to make that the canonical location of information?
```suggestion
    spin_df_end_time = spin_start_met[-1] + actual_spin_periods[-1]
```

greglucas (approved):

imap_processing/spice/spin.py
```diff
@@ -54,7 +54,9 @@ def get_spin_data() -> pd.DataFrame:
             * `spin_start_subsec_sclk`: MET microseconds of spin start time.
             * `spin_start_met`: Floating point MET seconds of spin start.
             * `spin_start_utc`: UTC string of spin start time.
-            * `spin_period_sec`: Floating point spin period in seconds.
+            * `spin_period_sec`: Floating point spin period in seconds (estimated).
+            * `actual_spin_period`: Floating point actual spin period computed from
```
tech3371, line 58: Just curious if I understood correctly. The spin period coming from ACS team is the estimated and then actual spin period is calculated by the code below to get more accurate one? Was that caused because of missing spin period?

tech3371 (approved):
code changes looks good to me. One curious question

imap_processing/spice/spin.py
tmplummer: My understanding of all this is that all of the spin-table data comes from a process onboard the s/c that uses star-tracker telemetry to estimate the attitude state (attitude and 1st derivative of the attitude). When that state passes through spin-phase zero, the time is recorded and the spin-period is computed using the estimated rate of rotation. Any small error in that rate will cause a small error in the spin-period. Then, as the s/c rotates through the next spin, it keeps updating the s/c attitude state and thus the next spin-start may not be exactly equal to the previous start plus the estimated spin-period.

imap_processing/spice/spin.py
tmplummer: Yes, great point.

PR #2562, opened 2026-01-08
CoDICE: Production preparation

### LINKED ISSUE #2135: Use codice_packet_definition.xml for housekeeping decom
Currently the CoDICE housekeeping data is being decommed against [`P_COD_NHK.xml`](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/codice/packet_definitions/P_COD_NHK.xml) in processing level 0. This is due to some (rather old) issues that I think may be resolved now. Ideally we should use [`codice_packet_definition.xml`](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/codice/packet_definitions/codice_packet_definition.xml) since all of the other products are using that. [Here](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/2a8dc5c6c7d91e9cdac7d520e53a2a5172d2267e/imap_processing/codice/codice_l0.py#L39) is the relevant code.

tech3371:
# Change Summary

## Overview

This code changes gets CoDICE non-DE products ready for production. DE preparation will come later.

NOTE:
I checked the CLI and they seem ready. I feel like this is all we need at the moment but dev test with latest LUT file will tell us. I am keeping this in draft PR until dev test is ok.


imap_processing/codice/packet_definitions/codice_packet_definition.xml
```diff
@@ -1465,7 +1528,8 @@
 			<xtce:IntegerParameterType name="COD_LO_IAL.CHKSUM" signed="false">
 				<xtce:IntegerDataEncoding sizeInBits="16" encoding="unsigned" />
 			</xtce:IntegerParameterType>
-			<xtce:IntegerParameterType name="COD_LO_PHA.SHCOARSE" signed="false">
+			<!-- TODO: undo this in coming DE segmented work -->
+			<!-- <xtce:IntegerParameterType name="COD_LO_PHA.SHCOARSE" signed="false">
```
greglucas: Why are you undoing this change? Can't we leave it in if you are skipping DEs.

I think we should have the XTCE be what it is and not comment things out.

tech3371: so keep it as we want in the upcoming work?

tech3371: I see what you mean. I will uncomment these

greglucas: Yes, I would prefer the XTCE to be exactly what I put in the other branch.

Although now I'm wondering if the tests will work with that update or not...

tech3371: True..the tests will fail but I need to skip them anyway. So your suggestion works for me still.

I haven't done the thorough tests yet because I am getting this draft PR up so that Luisa can test in dev if she gets chance today. Then I will resume dev testing my tomorrow.

greglucas (approved):

imap_processing/codice/codice_l1a_lo_angular.py
```diff
@@ -198,6 +198,41 @@ def l1a_lo_angular(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
...
+            (np.array(half_spin_per_esa_step), np.full(pad_size, np.nan))
+        )
+    # TODO: Handle epoch dependent acquisition time and half spin per esa step
+    #   For now, just tile the same array for all epochs.
+    #   Eventually we may have data from a day where the LUT changed. If this is the
+    #  case, we need to split the data by epoch and assign different acquisition times
+    half_spin_per_esa_step = np.tile(
+        np.asarray(half_spin_per_esa_step),
+        (len(unpacked_dataset["acq_start_seconds"]), 1),
+    )
```
tech3371, line 214: This is the case for when we have two SCI-LUT. When this happens, we need to use correct SCI-LUT for given table_id.

tech3371: why the name half_spin_per_esa_step with (epoch, esa_step) dims? CoDICE team had that. But this still time varying but the value of this variable is row_number from lo_stepping_tab.

how is this different from acquisition time? This is just recording the row_number. This is needed for L2.

use time to get correct SCI-LUT to get correct row_number data.

imap_processing/codice/utils.py
```diff
@@ -360,20 +362,34 @@ def get_codice_epoch_time(
     return center_times_seconds, delta_times
 
 
-def calculate_acq_time_per_step(low_stepping_tab: dict) -> np.ndarray:
+def calculate_acq_time_per_step(
```
tech3371, line 365: This will get effected from two lut in one day data.

imap_processing/codice/codice_l1a_lo_angular.py
```diff
@@ -198,6 +198,41 @@ def l1a_lo_angular(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
...
+    # For every energy after nso_half_spin, set data to fill values
+    nso_half_spin = unpacked_dataset["nso_half_spin"].values
+    nso_mask = half_spin_per_esa_step > nso_half_spin[:, np.newaxis]
+    species_mask = nso_mask[:, np.newaxis, :, np.newaxis, np.newaxis]
+    species_mask = np.broadcast_to(species_mask, species_data.shape)
+    species_data = species_data.astype(np.float64)
+    species_data[species_mask] = np.nan
+    # Set half_spin_per_esa_step to 63 which is the fill value
+    # half_spin_per_esa_step[nso_mask] = 63
+    # # Set acquisition time per esa step to nan where nso_mask is True
```
tech3371: remove these two lines

imap_processing/codice/codice_l1a_lo_angular.py
```diff
@@ -198,6 +198,41 @@ def l1a_lo_angular(unpacked_dataset: xr.Dataset, lut_file: Path) -> xr.Dataset:
...
+    )
+    # For every energy after nso_half_spin, set data to fill values
+    nso_half_spin = unpacked_dataset["nso_half_spin"].values
+    nso_mask = half_spin_per_esa_step > nso_half_spin[:, np.newaxis]
+    species_mask = nso_mask[:, np.newaxis, :, np.newaxis, np.newaxis]
+    species_mask = np.broadcast_to(species_mask, species_data.shape)
+    species_data = species_data.astype(np.float64)
+    species_data[species_mask] = np.nan
+    # Set half_spin_per_esa_step to 63 which is the fill value
+    # half_spin_per_esa_step[nso_mask] = 63
```
tech3371: can we make this nan?

tech3371 (commented):
some comments for ourselves.

imap_processing/codice/codice_l1a_lo_angular.py
lacoak21: Yes. An example date where we expect this to happen is dec 18th

imap_processing/codice/codice_l1a_lo_angular.py
```diff
@@ -58,7 +58,7 @@ def _despin_species_data(
     # 24 is derived by multiplying spin sector dim from collapse table by 2
     spin_sector_len = constants.LO_DESPIN_SPIN_SECTORS
     despun_shape = (num_packets, num_species, esa_steps, spin_sector_len, inst_az_dim)
-    despun_data = np.full(despun_shape, 0)
+    despun_data = np.full(despun_shape, np.nan)
```
lacoak21: I think this is causing unit _test_ failures. Are you sure this is supposed to be nan?

lacoak21: its also causing validation against the dec 10th data to break. I think you're right here to use np.nan so maybe we should bring this up in the tagup tomorrow.

PR #2569, opened 2026-01-09
Codice continuation packets

### LINKED ISSUE #2568: CoDICE Direct Event Failures
We are currently reading in many fields for each packet, but this is incorrect, the fields are only in the _first_ packet when the packets are segmented. This meant that we are extracting things with XTCE, then recombining those fields back into binary and then extracting again. The issue came about when the second packet's binary payload was not long enough to unpack all the individual fields and we get a reading off the end of the binary stream.

We should update the XTCE definition and handle the combining and parsing of the entire binary data payload, extracting the header fields within code rather than XTCE.

greglucas:
# Change Summary

## Overview

This changes the CoDICE packet definition for direct events. Previously, we were reading in many fields for each packet, but this is incorrect, the fields are only in the _first_ packet. This meant that we were extracting things with XTCE, then recombining those fields back into binary and then extracting again. The issue came about when the second packet's binary payload was not long enough to unpack all the individual fields.

This is a significant refactor as well. Rather than working with binary strings, we are working with bytes directly and doing bitshifts on numpy arrays. Thanks to good unit tests, this was able to be refactored with the help of AI.

closes #2568

### DIFF
--- a/imap_processing/codice/codice_l1a_de.py
+++ b/imap_processing/codice/codice_l1a_de.py
@@ -9,112 +9,107 @@ from imap_processing.codice.decompress import decompress
 from imap_processing.codice.utils import (
     CODICEAPID,
     CoDICECompression,
-    SegmentedPacketOrder,
     ViewTabInfo,
     apply_replacements_to_attrs,
     get_codice_epoch_time,
 )
 from imap_processing.spice.time import met_to_ttj2000ns
+from imap_processing.utils import combine_segmented_packets
 
 
-def get_de_metadata(packets: xr.Dataset, packet_index: int) -> bytes:
+def extract_initial_items_from_combined_packets(
+    packets: xr.Dataset,
+) -> xr.Dataset:
     """
-    Gather and return packet metadata (From packet_version through byte_count).
+    Extract metadata fields from the beginning of combined event_data packets.
 
-    Extract the metadata in the packet_indexed direct event packet, which is then
-    used to construct the full data of the group of packet_indexs.
+    Extracts bit fields from the first 20 bytes of each event_data array
+    and adds them as new variables to the dataset.
 
     Parameters
     ----------
-    packets : xarray.Dataset
-        The packet_indexed direct event packet data.
-    packet_index : int
-        The index of the packet_index of interest.
+    packets : xr.Dataset
+        Dataset containing combined packets with event_data.
 
     Returns
     -------
-    metadata : bytes
-        The compressed metadata for the packet_indexed packet.
+    xr.Dataset
+        Dataset with extracted metadata fields added.
     """
-    # String together the metadata fields and convert the data to a bytes obj
-    metadata_str = ""
-    for field, num_bits in constants.DE_METADATA_FIELDS.items():
-        metadata_str += f"{packets[field].data[packet_index]:0{num_bits}b}"
-    metadata_chunks = [metadata_str[i : i + 8] for i in range(0, len(metadata_str), 8)]
-    metadata_ints = [int(item, 2) for item in metadata_chunks]
-    metadata = bytes(metadata_ints)
+    # Initialize arrays for extracted fields
+    n_packets = len(packets.epoch)
 
-    return metadata
+    # Preallocate arrays
+    packet_version = np.zeros(n_packets, dtype=np.uint16)
+    spin_period = np.zeros(n_packets, dtype=np.uint16)
+    acq_start_seconds = np.zeros(n_packets, dtype=np.uint32)
+    acq_start_subseconds = np.zeros(n_packets, dtype=np.uint32)
+    spare_1 = np.zeros(n_packets, dtype=np.uint8)
+    st_bias_gain_mode = np.zeros(n_packets, dtype=np.uint8)
+    sw_bias_gain_mode = np.zeros(n_packets, dtype=np.uint8)
+    priority = np.zeros(n_packets, dtype=np.uint8)
+    suspect = np.zeros(n_packets, dtype=np.uint8)
+    compressed = np.zeros(n_packets, dtype=np.uint8)
+    num_events = np.zeros(n_packets, dtype=np.uint32)
+    byte_count = np.zeros(n_packets, dtype=np.uint32)
 
+    # Extract fields from each packet
+    for pkt_idx in range(n_packets):
+        event_data = packets.event_data.data[pkt_idx]
 
-def group_data(packets: xr.Dataset) -> list[bytes]:
-    """
-    Organize continuation packets into appropriate groups.
+        # Byte-aligned fields using int.from_bytes
+        packet_version[pkt_idx] = int.from_bytes(event_data[0:2], byteorder="big")
+        spin_period[pkt_idx] = int.from_bytes(event_data[2:4], byteorder="big")
+        acq_start_seconds[pkt_idx] = int.from_bytes(event_data[4:8], byteorder="big")
 
-    Some packets are continuation packets, as in, they are packets that are
-    part of a group of packets. These packets are marked by the `seq_flgs` field
-    in the CCSDS header of the packet. For CoDICE, the values are defined as
-    follows:
+        # Non-byte-aligned fields (bytes 8-12 contain mixed bit fields)
+        # Extract 4 bytes and unpack bit fields
+        mixed_bytes = int.from_bytes(event_data[8:12], byteorder="big")
 
-    3 = Packet is not part of a group
-    1 = Packet is the first packet of the group
-    0 = Packet is in the middle of the group
-    2 = Packet is the last packet of the group
+        # acq_start_subseconds: 20 bits (MSB)
+        acq_start_subseconds[pkt_idx] = (mixed_bytes >> 12) & 0xFFFFF
+        # spare_1: 2 bits
+        spare_1[pkt_idx] = (mixed_bytes >> 10) & 0x3
+        # st_bias_gain_mode: 2 bits
+        st_bias_gain_mode[pkt_idx] = (mixed_bytes >> 8) & 0x3
+        # sw_bias_gain_mode: 2 bits
+        sw_bias_gain_mode[pkt_idx] = (mixed_bytes >> 6) & 0x3
+        # priority: 4 bits
+        priority[pkt_idx] = (mixed_bytes >> 2) & 0xF
+        # suspect: 1 bit
+        suspect[pkt_idx] = (mixed_bytes >> 1) & 0x1
+        # compressed: 1 bit (LSB)
+        compressed[pkt_idx] = mixed_bytes & 0x1
 
-    For packets that are part of a group, the byte count associated with the
-    first packet of the group signifies the byte count for the entire group.
+        # Remaining byte-aligned fields
+        num_events[pkt_idx] = int.from_bytes(event_data[12:16], byteorder="big")
+        byte_count[pkt_idx] = int.from_bytes(event_data[16:20], byteorder="big")
 
-    Parameters
-    ----------
-    packets : xarray.Dataset
-        Dataset containing the packets to group.
+        # Remove the first 20 bytes from event_data (header fields from above)
+        # Then trim to the number of bytes indicated by byte_count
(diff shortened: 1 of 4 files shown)


imap_processing/codice/codice_l1a_de.py
```diff
@@ -17,104 +17,147 @@
...
+def combine_segmented_packets(packets: xr.Dataset) -> xr.Dataset:
     """
-    Gather and return packet metadata (From packet_version through byte_count).
+    Combine segmented packets into unsegmented packets.
 
-    Extract the metadata in the packet_indexed direct event packet, which is then
-    used to construct the full data of the group of packet_indexs.
+    All packets have (SHCOARSE, EVENT_DATA, CHKSUM) fields. To combine
+    the segmented packets, we only concatenate along the EVENT_DATA field
+    into the first packet of the group.
```
tech3371: This is for my future self and may be others. I don't want to lose context of these changes if someone looks at this later since those details are in our email. Can you break line where appropriate and make it more clear as needed?
```suggestion
Direct event data are segmented into packets. CoDICE assembles these segmented packets slightly differently than the standard. Onboard, CoDICE direct event data are segmented as follows:

Standalone / unsegmented packets:
Data is packed in the order defined in the telemetry definition in Excel. eg.
(SHCOARSE, many metadata fields, EVENT_DATA, CHECKSUM).

First segment:
Data is packed in the order defined in the telemetry definition in Excel. Eg.
(SHCOARSE, many metadata fields, EVENT_DATA, CHECKSUM).

As we see here, the first segment packet contains metadata defined in the telemetry definition. Those metadata are unpacked in later functions/steps because those metadata will always be fixed bit lengths and should exist.

Middle segment:
Data is packed in the following order: SHCOARSE, EVENT_DATA, CHECKSUM.
There can be multiple middle packets for a given packet group.

Last segment:
Data is packed in the following order: SHCOARSE, EVENT_DATA, CHECKSUM.
This last segment of event data can contain padded values.

Because of this behavior, we defined the XTCE to unpack the data using the field order
(SHCOARSE, EVENT_DATA, CHKSUM). This simplifies XTCE-based unpacking and allows
the remaining packet-specific details to be handled in code. In this function,
the segmented event data are combined by concatenating the EVENT_DATA field across
the packet group.
```

imap_processing/codice/codice_l1a_de.py
```diff
@@ -17,104 +17,147 @@
...
+        suspect[pkt_idx] = (mixed_bytes >> 1) & 0x1
+        # compressed: 1 bit (LSB)
+        compressed[pkt_idx] = mixed_bytes & 0x1
+
+        # Remaining byte-aligned fields
+        num_events[pkt_idx] = int.from_bytes(event_data[12:16], byteorder="big")
+        byte_count[pkt_idx] = int.from_bytes(event_data[16:20], byteorder="big")
+
+        # Remove the first 20 bytes from event_data
+        # Then trim to the byte_count value
```
tech3371:
```suggestion
        # Remove the first 20 bytes from event_data
        # Then trim to the byte_count value to remove padded values.
```

imap_processing/codice/codice_l1a_de.py
```diff
@@ -336,142 +510,14 @@ def l1a_direct_event(unpacked_dataset: xr.Dataset, apid: int) -> xr.Dataset:
...
-
-    # Decompress data shape is (epoch, priority * num_events)
-    decompressed_data = [
-        decompress(
-            group,
-            CoDICECompression.LOSSLESS,
-        )
-        for group in grouped_data
-    ]
+    # Combine segmented packets and extract header fields
```
tech3371, line 493:
```suggestion
    # Combine segmented packets and extract metadata fields
```

imap_processing/codice/codice_l1a_de.py
```diff
@@ -160,162 +203,293 @@ def unpack_bits(bit_structure: dict, de_data: np.ndarray) -> dict:
...
-        # num_priorities, then throw error. They should match.
-        if len(np.unique(unordered_priority)) != num_priorities:
-            raise ValueError(
-                f"Priority array for epoch {epoch_index} contains "
-                f"non-unique values: {unordered_priority}"
-            )
+    return dataset
+
+
+def _unpack_and_store_events(
```
tech3371, line 263: I followed all changes by tracing call from this `l1a_direct_event()` and everything made sense and really nice cleanup!

This function `_unpack_and_store_events()` is where I am little unsure. Did it work? Just curious.

tech3371: I tried locally with unittest data and then flight data. Looks like unittest data worked with these new changes but flight data fails with lzma error we have been seeing.

greglucas: What command were you testing with to see the error?

imap_processing/codice/codice_l1a_de.py
greglucas: I think this is too verbose. I made this a generic function in the ultra continuation packet PR so that multiple instruments can use it. Over there, it isn't the same packet definition as here, so this wouldn't make sense to add to the generic function.

imap_processing/codice/codice_l1a_de.py
greglucas: I don't like the word "metadata" since this isn't metadata, it is actual data fields. These are instrument specific data values, so I feel like header is a reasonable naming.

imap_processing/codice/codice_l1a_de.py
tech3371: can we add it somewhere in this `codice_l1a_de.py`? I don't mind where we put this in that file but would be nice to capture this for future reference.

imap_processing/codice/codice_l1a_de.py
tech3371: I think you already have the command now

tech3371:
Current code looks good to me. I think there are few changes coming from yesterday discussion. let me know when you want me to review this again.

imap_processing/codice/codice_l1a_de.py
greglucas: I added some of this to the codice_l1a_de file now, explaining what fields we have in XTCE and why we are doing it this way now.

greglucas:
OK, I think the last few commits contain the logic that we need for handling missing priority groups and dropping those. I still need to look into whether that is OK, or if we want to fill in some priority values and use FILL for the rest, or drop the entire sequence when that happens. But, this should be ready for review now and I did a cursory check and looks like it is producing valid events for the December test cases now.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -9,112 +11,114 @@
...
+    This was previously done in XTCE, but we can't do that because of
+    segmented packets that need to be combined. Each segmented packet
+    has its own (SHCOARSE, EVENTDATA, CHKSUM) fields, so we need to
+    only combine along the EVENTDATA field and extract data that way.
 
     Parameters
     ----------
-    packets : xarray.Dataset
-        Dataset containing the packets to group.
+    packets : xr.Dataset
```
lacoak21: Wont this cause documenation build errors? 

xarray.Dataset -> xr.Dataset

lacoak21 (commented):
OK ive only review part of this! Im going to do the rest first thing tomorrow.

imap_processing/utils.py
```diff
@@ -408,21 +413,70 @@ def combine_segmented_packets(
...
+    # Check for sequential source sequence counters
+    # CCSDS source sequence counter is a 14-bit field (0-16383)
+    counter_max = 16384
+    src_seq_ctr = ds["src_seq_ctr"].data
+
+    if len(src_seq_ctr) <= 1:
+        return
+
+    # Check if each counter equals (previous + 1) % counter_max
+    # This handles both normal increments and rollover (16383 -> 0)
```
tech3371, line 457: Helpful documentation!

imap_processing/utils.py
```diff
@@ -408,21 +413,70 @@ def combine_segmented_packets(
...
+    Parameters
+    ----------
+    ds : xarray.Dataset
+        Dataset containing the packets to check.
+    apid : int
+        APID of the packets.
+    """
+    # Check for sequential source sequence counters
+    # CCSDS source sequence counter is a 14-bit field (0-16383)
+    counter_max = 16384
```
tech3371, line 450: should we make this into a constant above? Otherwise, it looks good

imap_processing/utils.py
```diff
@@ -333,6 +333,9 @@ def packet_file_to_datasets(
             )
             ds = ds.isel(epoch=unique_indices)
 
+        # Log a warning if there are gaps in the source sequence counter
+        _check_source_sequence_counter(ds, apid)
```
tech3371, line 337: nice addition!

imap_processing/codice/codice_l1a_de.py
```diff
@@ -160,318 +164,342 @@ def unpack_bits(bit_structure: dict, de_data: np.ndarray) -> dict:
...
+    # Find incomplete groups (not exactly num_priorities packets)
+    incomplete_mask = counts != num_priorities
+    if np.any(incomplete_mask):
+        incomplete_times = unique_times[incomplete_mask]
+        incomplete_counts = counts[incomplete_mask]
+        logger.warning(
+            f"Found {len(incomplete_times)} incomplete priority group(s) "
+            f"for APID {apid}. Expected {num_priorities} packets per group. "
+            f"Incomplete groups at acq_start_seconds {incomplete_times.tolist()} "
+            f"with counts {incomplete_counts.tolist()}. Dropping these packets."
```
tech3371, line 382: nice and so much easier to follow!

tech3371 (approved):
This looks great to me. Thank you for taking this on! Sorry for delay review. I didn't see the comment in my filtered email.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -160,318 +164,342 @@ def unpack_bits(bit_structure: dict, de_data: np.ndarray) -> dict:
...
+        n_events = int(num_events_arr[pkt_idx])
+        if n_events == 0:
+            continue
+
+        # Extract and byte-reverse events for LSB unpacking
+        pkt_bytes = np.asarray(event_data_arr[pkt_idx], dtype=np.uint8)
+        pkt_bytes = pkt_bytes.reshape(n_events, 8)[:, ::-1]
+        all_event_bytes[offset : offset + n_events] = pkt_bytes
+
+        # Record destination indices for scattering
```
lacoak21: what do you mean by scattering here and on line 331?

imap_processing/codice/codice_l1a_de.py
greglucas: 🤔 This passed the documentation build... I think you're right though, so I went through and updated this here and elsewhere throughout this file. It looks like there was previously a mix of usage.

imap_processing/utils.py
greglucas: I don't think this is a critical constant and isn't used in other places, so I just kept it outside for now. We can re-evaluate later and move it out if needed/desired.

imap_processing/codice/codice_l1a_de.py
greglucas: Great question! My AI friend wrote that comment. But, what it is getting at is that we are "scattering" the data from one layout into another array-based layout in a sense. So an item in array 1 at location (1, 2, 3) will get "scattered" into array2 at location (2, 5, 7) and there is a relationship we are storing for how that mapping is working.

I don't think that was clear, so I tried to reword both locations. It is basically doing array-based index lookups and assignments.

lacoak21: Interesting! Got it. Ive only ever "scattered" data in reference to uncertainty and not something that we can "fix".

lacoak21 (approved):

imap_processing/codice/codice_l1a_de.py
greglucas: FYI: https://en.wikipedia.org/wiki/Gather/scatter_(vector_addressing)

PR #2570, opened 2026-01-09
I-ALiRT - Andriy cdf response

laspsandoval:
# Change Summary

## Overview
Complete change in cdf structure and updates to variable attributes based on Andriy's feedback. Also note that I intend to calculate each instrument's timestamp (Issue: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/2571 ).

## Updated Files: Major changes
- create_xarray.py
   - Basically a complete rework of this file. A lot has changed with the database. CoDICE data structure is now more clearly defined. And after speaking with Andriy I realized the need for an epoch for each instrument.
- imap_ialirt_l1_variable_attrs.yaml
   - A lot of feedback from Andriy for this.

## Updated Files: Minor changes
- imap_processing/ialirt/utils/constants.py
   - Organization of dimensions, energy bins, and data types.
- constants.py
   - Added constants based on CoDICE Algorithm Document page 64-65. These are not stored in DynamoDB (website documentation is still needed here), but they will be used in the cdf.
- imap_ialirt_global_cdf_attrs.yaml
   - Global description as recommended
- parse_mag.py
   - Find centerpoint of sc timestamp (instead of first timestamp)
- process_codice.py
   - Find centerpoint of sc timestamp (instead of first timestamp)
- process_hit.py
   - Find centerpoint of sc timestamp (instead of first timestamp)
- process_swapi.py
   - Find centerpoint of sc timestamp (instead of first timestamp)
- process_swe.py
   - Find centerpoint of sc timestamp (instead of first timestamp)

## Testing
test_create_xarray.py


laspsandoval:
@greglucas I think this is ready for your review when you have the chance. Really appreciate it.

imap_processing/cdf/config/imap_ialirt_global_cdf_attrs.yaml
```diff
@@ -9,10 +9,10 @@ instrument_base: &instrument_base
...
     See https://imap.princeton.edu for more details.
-  Instrument_type: "Multiple Instruments"
+  Instrument_type: Particles (space), Magnetic Fields (space), Plasma and Solar Wind (space), Ephemeris/Attitude/Ancillary
 
 imap_ialirt_l1_realtime:
     <<: *instrument_base
     Data_type: L1_REALTIME>Level-1 Realtime
     Logical_source: imap_ialirt_l1_realtime
-    Logical_source_description: IMAP-IALiRT Instrument Level-1 Realtime Data.
+    Logical_source_description: IMAP Active Link for Real-Time (I-ALiRT) Level-1 Realtime Data.
```
greglucas: Do you want to remove Realtime here because it is already in the name? I think it should be two words or hyphenated if you want to leave it though.
```suggestion
    Logical_source_description: IMAP Active Link for Real-Time (I-ALiRT) Level-1 Data.
```

imap_processing/cdf/config/imap_ialirt_l1_variable_attrs.yaml
```diff
@@ -1,6 +1,62 @@
...
+  TIME_SCALE: Terrestrial Time
+  REFERENCE_POSITION: Rotating Earth Geoid
+  RESOLUTION: ' '
+  DICT_KEY: "SPASE>Support>SupportQuantity:Temporal"
+  dtype: int64
+
+codice_hi_epoch:
+    <<: *instrument_epoch
+    CATDESC: CoDICE-Hi time, number of nanoseconds since J2000 with leap seconds included
+    FIELDNAM: CoDICE-Hi epoch
```
greglucas: I don't know that we should be changing these, but rather just re-use the same epoch description everywhere. It isn't "CoDICE time", it is just "time" that the codice instrument took the measurement. But, that isn't relevant it is just this is what time any value was taken at.

imap_processing/ialirt/l0/parse_mag.py
```diff
@@ -716,7 +716,9 @@ def process_packet(
         )
 
         met = grouped_data["met"][(grouped_data["group"] == group).values]
-        met_all.append(met.values[0])
+        # Calculate the center of the time variable
+        sc_met = int(np.mean([np.min(met), np.max(met)]))
```
greglucas:
```suggestion
        sc_met = (met[0] + met[-1]) // 2
```

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -471,12 +471,15 @@ def process_codice(
             )
             l2_lo = calculate_ratios(l1b_lo, l2_lut_path, l2_geometric_factor_path)
 
+            # Calculate the center of the time variable
+            sc_met = int(np.mean([np.min(met), np.max(met)]))
```
greglucas: Same equation as above to avoid unnecessary numpy calls.

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -471,12 +471,15 @@ def process_codice(
...
+
             codice_lo_data.append(
                 {
                     "apid": 478,
-                    "met": int(met[0]),
-                    "met_in_utc": met_to_utc(met[0]).split(".")[0],
-                    "ttj2000ns": int(met_to_ttj2000ns(met[0])),
+                    "met": int(sc_met),
```
greglucas: It is already an int because of the cast above.
```suggestion
                    "met": sc_met,
```

imap_processing/ialirt/l0/process_codice.py
```diff
@@ -513,12 +516,15 @@ def process_codice(
                 l2_hi
             ).tolist()
 
+            # Calculate the center of the time variable
+            sc_met = int(np.mean([np.min(met), np.max(met)]))
```
greglucas: Same equation as above. Maybe a helper function even? `_calculate_center_met_time(met) -> int`

imap_processing/ialirt/l0/process_hit.py
```diff
@@ -161,16 +161,18 @@ def process_hit(xarray_data: xr.Dataset) -> list[dict]:
...
 
         hit_data.append(
             {
                 "apid": 478,
-                "met": int(met),
-                "met_in_utc": met_to_utc(met).split(".")[0],
-                "ttj2000ns": int(met_to_ttj2000ns(met)),
+                "met": sc_met,
+                "met_in_utc": met_to_utc(sc_met).split(".")[0],
+                "ttj2000ns": int(met_to_ttj2000ns(sc_met)),
```
greglucas: All of these are duplicated throughout every instrument I think. Maybe something like a helper function so that all instrument populate this basic data in a consistent way?

```python
def _populate_instrument_header_items(met) -> dict
    return {"apid": 478, "met": sc_met, ...}

# Use it by combining the header time items and the instrument data separately.
hit_data.append(_populate_instrument_header_items(met) + {"instrument": "hit", ...})
```

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -25,147 +40,268 @@ def create_xarray_from_records(records: list[dict]) -> xr.Dataset:  # noqa: PLR0
...
+        attrs=cdf_manager.get_variable_attributes(
+            "codice_hi_energy_plus", check_schema=False
         ),
     )
 
     elevation = xr.DataArray(
-        data=np.arange(4, dtype=np.uint8),
-        name="codice_hi_h_elevation",
-        dims=["codice_hi_h_elevation"],
+        np.float32(HI_IALIRT_ELEVATION_ANGLE),
```
greglucas: Since this is a constant, why not set it initially in the constants array as this dtype?

Otherwise, this should also be written as:
```suggestion
        HI_IALIRT_ELEVATION_ANGLE.astype(np.float32),
```

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -25,147 +40,268 @@ def create_xarray_from_records(records: list[dict]) -> xr.Dataset:  # noqa: PLR0
...
+    # Formula:
+    #   θ_(g,n) = (θ_(g,0)+90°* n)  mod 360°
+    # where
+    #   n is number of sectored angles, 0 to 3,
+    #   g is size of the group (inst_az), 0 to 3,
+    spin_angles = (
+        HI_IALIRT_REF_SPIN_ANGLE[:, np.newaxis] + np.array([0, 1, 2, 3]) * 90
+    ) % 360.0
     spin_angle = xr.DataArray(
+        data=np.float32(spin_angles),
```
greglucas: use `.astype()`, but also think about whether this is a constant itself that you should add this calculation and note to the main location so you can just reference a single variable. i.e. these look constant to me themselves.

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -25,147 +40,268 @@ def create_xarray_from_records(records: list[dict]) -> xr.Dataset:  # noqa: PLR0
...
+        ),
+    )
+
+    spin_sector_labels = xr.DataArray(
+        [
+            "sector_0",
+            "sector_1",
+            "sector_2",
+            "sector_3",
+        ],  # or ["sector0","sector1",...] - whatever you prefer
```
greglucas: We should be consistent with the main instrument data product. Do you know what labels are used there? (with or without underscore or something else entirely)

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -25,147 +40,268 @@ def create_xarray_from_records(records: list[dict]) -> xr.Dataset:  # noqa: PLR0
...
     }
     dataset = xr.Dataset(
         coords=coords,
         attrs=cdf_manager.get_global_attributes("imap_ialirt_l1_realtime"),
     )
 
-    # Create empty dataset for each key.
-    for key in instrument_keys:
+    # Create variables with fill values.
+    for key in IALIRT_DIMS.keys():
```
greglucas: I don't think you need `.keys()` in these following calls.
```suggestion
    for key in IALIRT_DIMS:
```

imap_processing/ialirt/utils/create_xarray.py
```diff
@@ -25,147 +40,268 @@ def create_xarray_from_records(records: list[dict]) -> xr.Dataset:  # noqa: PLR0
...
+
+        shape = [dataset.dims[d] for d in dims]
+
+        data = np.full(shape, fill, dtype=dtype)
+        dataset[key] = xr.DataArray(data, dims=dims, attrs=attrs)
+
+    for i, record in enumerate(by_inst.get("mag", [])):
+        for key in IALIRT_DIMS.keys():
+            if key in ["mag_B_GSE", "mag_B_GSM", "mag_B_RTN"]:
+                dataset[key].data[i, :] = np.asarray(record[key], dtype=np.float32)
```
greglucas: Why are we iterating over the items and filling the array? It seems like we should convert the array all at once. `dataset[key].astype(np.float32)`

laspsandoval:
pre-commit.ci autofix

greglucas (approved):
Thank you for adding that helper function for the header!

PR #2573, opened 2026-01-13
Codice save half_spin_per_esa_step in l1a and use in l2 processing

lacoak21:
# Change Summary

## Overview
We should NOT be using the HALF_SPIN_LUT from the constants file. We should be getting those values from the sci lut. Joey requested that these be written to l1a products that use them in l2. 

## Updated Files
- imap_processing/codice/codice_l1a_lo_angular.py and imap_processing/codice/codice_l1a_lo_species.py
   - Write half_spin_per_esa step to CDF because those products at l2 will use that variable to compute intensities.
   - Update geometric calc function to use half spin values from the dataset carried through 
   - remove LUT from constants file. Those should NEVER be hardcoded. 

## Testing
Ensure geometric factor calculations are using the half spin values from the given dataset.

### DIFF
--- a/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_l1a_variable_attrs.yaml
@@ -185,6 +185,18 @@ acquisition_time_per_step:
     VALIDMAX: 625.000000
     VAR_TYPE: support_data
 
+half_spin_per_esa_step:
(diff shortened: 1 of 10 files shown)


tech3371 (approved):
This changes looks good.

tech3371:
I am going to merge this PR to be able to use this change in our `codice_prod` branch validation. Thank you for getting this in quick!

PR #2581, opened 2026-01-15
Ultra segmented packet support

greglucas:
# Change Summary

This adds a general helper to combine segmented packets together after the XTCE parsing. This required an updated to the XTCE xml so that the header fields aren't read in on the middle/last packets. This means we need to extract those fields manually after the XTCE parsing and combining of segmented packets. That is ultra specific, so I put the extraction function in the Ultra namespace.

### DIFF
--- a/imap_processing/utils.py
+++ b/imap_processing/utils.py
@@ -10,6 +10,7 @@ import pandas as pd
 import space_packet_parser as spp
 import xarray as xr
 from space_packet_parser.exceptions import UnrecognizedPacketTypeError
+from space_packet_parser.generators.ccsds import SequenceFlags
 from space_packet_parser.xtce import definitions, encodings, parameter_types
 
 from imap_processing.spice.time import met_to_ttj2000ns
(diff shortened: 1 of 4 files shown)


imap_processing/utils.py
```diff
@@ -349,6 +350,62 @@ def packet_file_to_datasets(
     return dataset_by_apid
 
 
+def combine_segmented_packets(
+    packets: xr.Dataset, binary_field_name: str = "packetdata"
+) -> xr.Dataset:
+    """
+    Combine segmented packets into unsegmented packets.
+
+    All packets have (SHCOARSE, EVENT_DATA, CHKSUM) fields. To combine
```
tech3371: should we update this to be generic now?
```suggestion
    All packets have (SHCOARSE, binary_field_name) fields. To combine
```

tech3371 (commented):
the changes looks good to me. It's coming nicely together! Nice work!

tech3371 (approved):

imap_processing/utils.py
greglucas: Good call! I reworded this to update it and remove CoDICE things that I copied over.

PR #2587, opened 2026-01-19
I-ALiRT - SWE add time interpolation instead of latest

laspsandoval:
# Change Summary

## Overview
Ruth is interpolating calibration factors based on time. Also, when we "reprocess" for a week to fill in gaps grabbing the latest calibration values will not be sufficient. This PR interpolates based on time rather than grabbing the latest.

## Updated Files
- process_swe.py
   - interpolates based on time rather than grabbing the latest.

### DIFF
--- a/imap_processing/ialirt/l0/process_swe.py
+++ b/imap_processing/ialirt/l0/process_swe.py
@@ -523,7 +523,7 @@ def process_swe(accumulated_data: xr.Dataset, in_flight_cal_files: list) -> list
         cal_met = in_flight_cal_df["met_time"].to_numpy()
 
         group_time_first_half = (
-            grouped["time_seconds"].where(grouped["swe_seq"] == 0, drop=True).values
+            grouped["time_seconds"].where(grouped["swe_seq"] < 30, drop=True).values
         )
         group_time_first_half_mid = (
             group_time_first_half[0] + group_time_first_half[-1]
@@ -533,7 +533,7 @@ def process_swe(accumulated_data: xr.Dataset, in_flight_cal_files: list) -> list
             [
                 float(
                     np.interp(
-                        int(group_time_first_half[0]),
+                        int(group_time_first_half_mid),
                         cal_met,
                         in_flight_cal_df[cem].to_numpy(),
                     )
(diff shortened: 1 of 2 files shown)


laspsandoval:
pre-commit.ci autofix

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -520,10 +517,56 @@ def process_swe(accumulated_data: xr.Dataset, in_flight_cal_files: list) -> list
...
+        group_time_first_half = (
+            grouped["time_seconds"].where(grouped["swe_seq"] < 30, drop=True).values
+        )
+        group_time_first_half_mid = (
+            group_time_first_half[0] + group_time_first_half[-1]
+        ) // 2
+
+        interp_cal_first_half = np.array(
+            [
+                float(
```
greglucas: I don't think you need to cast to "float" here since you are doing it in the dtype after this.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -520,10 +517,56 @@ def process_swe(accumulated_data: xr.Dataset, in_flight_cal_files: list) -> list
...
-        latest_cal = in_flight_cal_df.sort_values("met_time").iloc[-1][1::]
+        cal_cols = [f"cem{i}" for i in range(1, 8)]
+        cal_met = in_flight_cal_df["met_time"].to_numpy()
+
+        group_time_first_half = (
+            grouped["time_seconds"].where(grouped["swe_seq"] < 30, drop=True).values
+        )
+        group_time_first_half_mid = (
+            group_time_first_half[0] + group_time_first_half[-1]
+        ) // 2
```
greglucas, line 533: Could you add some high level comments about what this is doing. It wasn't immediately obvious to me at first glance.

imap_processing/ialirt/l0/process_swe.py
```diff
@@ -520,10 +517,56 @@ def process_swe(accumulated_data: xr.Dataset, in_flight_cal_files: list) -> list
...
+                    np.interp(
+                        int(group_time_second_half_mid),
+                        cal_met,
+                        in_flight_cal_df[cem].to_numpy(),
+                    )
+                )
+                for cem in cal_cols
+            ],
+            dtype=np.float64,
+        )
```
greglucas, line 565: Do we have this in the normal SWE code as a helper function already? This seems like something we should have if we don't... Can you open a ticket for looking into using a common CEM interpolator between science and I-ALiRT later?

greglucas (approved):
Just a few requests for comment, otherwise looks good to me.

imap_processing/ialirt/l0/process_swe.py
laspsandoval: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/2604

PR #2595, opened 2026-01-21
CoDICE l2 geometric factor: update mode calculation

lacoak21:
# Change Summary

## Overview
Update full / reduce mode calculation when getting the geometric factors at l2.

Summary from Tenzin 

updates:
Old ways:
Looks at rgfo and if value is 3, any value 4 or above should use reduced gain.
New ways:
Now, always triggering rgfo at step 0. Physically, we are not reducing anymore. So we don’t need to use reduced gain. We should use the geometric factor as it is for all of Lo intensity products. So any places that checks rgfo value to determine if we should use reduced gain, we don’t need to do that anymore.
Which data get affected?
Updated in all the data since Nov 24 SCI-LUT version was uploaded.

## Updated Files
   - Update mode calculation

## Testing
Fix tests.

### DIFF
--- a/imap_processing/codice/codice_l2.py
+++ b/imap_processing/codice/codice_l2.py
@@ -318,7 +318,7 @@ def compute_geometric_factors(
     # false (full mode)
     # TODO: The mode calculation will need to be revisited after FW changes in january
     #  2026.
-    # After February 24th 2025 we need to do this step a different way.
+    # After November 24th 2025 we need to do this step a different way.
     date_switch = datetime.datetime(2025, 11, 24, 13, 53, 59)
     start_date = dataset.attrs.get("Logical_file_id", None)
     if start_date is None:
--- a/imap_processing/codice/codice_l2.py
+++ b/imap_processing/codice/codice_l2.py
@@ -42,6 +42,7 @@ from imap_processing.codice.constants import (
     SW_POSITIONS,
 )
 from imap_processing.codice.utils import apply_replacements_to_attrs
+from imap_processing.spice.time import et_to_datetime64, ttj2000ns_to_et
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


laspsandoval:
I also saw this error:

Using the new science lut:
imap_codice_l1a-sci-lut_20251218_v001.json

I am getting an error on line 224 in codice_l1a_lo_species.py:
l1a_dataset["acquisition_time_per_step"] = xr.DataArray(
    calculate_acq_time_per_step(sci_lut_data["lo_stepping_tab"]),
    dims=("esa_step",),
    attrs=cdf_attrs.get_variable_attributes(
        "acquisition_time_per_step", check_schema=False
    ),
)

E               xarray.structure.alignment.AlignmentError: cannot reindex or align along dimension 'esa_step' because of conflicting dimension sizes: {104, 128} (note: an index is found along that dimension with size=128)

venv/lib/python3.12/site-packages/xarray/structure/alignment.py:451: AlignmentError

But maybe this is for another PR.

laspsandoval (approved):
Looks good to me. If you are uncertain then wait for Tenzin's review, though, since she knows the code better. But this is what Michael was talking about at the previous meeting so from my perspective it looks good.

lacoak21:
> Looks good to me. If you are uncertain then wait for Tenzin's review, though, since she knows the code better. But this is what Michael was talking about at the previous meeting so from my perspective it looks good.

Michael, Tenzin and Joey looked at it this morning so I feel pretty confident that this the correct way. The only issue im struggling to deal with is the failing ialirt test. I am using the Logical_file_id attribute to get the start date of the file so I can check which method to use but the ialirt dataset doesnt have it. Im hesitant to just add it to the test dataset in case that will not be there in production

laspsandoval:
> > Looks good to me. If you are uncertain then wait for Tenzin's review, though, since she knows the code better. But this is what Michael was talking about at the previous meeting so from my perspective it looks good.
> 
> Michael, Tenzin and Joey looked at it this morning so I feel pretty confident that this the correct way. The only issue im struggling to deal with is the failing ialirt test. I am using the Logical_file_id attribute to get the start date of the file so I can check which method to use but the ialirt dataset doesnt have it. Im hesitant to just add it to the test dataset in case that will not be there in production

You are correct. I-ALiRT will not have a logical_file_id. Would it be possible to use the epoch midpoint in that case?

lacoak21:
> You are correct. I-ALiRT will not have a logical_file_id. Would it be possible to use the epoch midpoint in that case?

Sure- ill just do that for all of them.

PR #2600, opened 2026-01-21
FIX: Lo DE handle partial ASCs

greglucas:
The spin_cycle variable is an average value over an ASC (28 spins). This means that when we have a partial ASC, the spin_cycle value could be in a "bad" spin which was reported as 0 in the ASC for spin_start_sec. To avoid this, we can use the average spin duration of the ASC and calculate the spin_start based on the start of the acquisition period of the ASC. This means we will always have a valid spin_start rather than 0 which caused SPICE lookup issues.

This gets l1b-de products being produced again and looks like it worked for a few quick old processing failures I tried locally.


tmplummer (approved):
LGTM

PR #2601, opened 2026-01-22
MNT: Separate out Lo l1b dataset logic into individual functions

greglucas:
The l1b dataset logic was within the if-blocks based on what dependency data was passed to the routine. This can be determined based on the descriptor requested and it can also be broken up into individual helper routines where the main code logic can be more easily identified and separated.


tmplummer (approved):
Excellent!

PR #2603, opened 2026-01-22
FIX: Add additional packet sequence checking

greglucas:
We can get a single FIRST / LAST / MIDDLE packet, which isn't the full segment, so we need to add a check for singular packets that are not UNSEGMENTED.


lacoak21 (approved):
Thanks for figuring this out Greg

PR #2605, opened 2026-01-22
2598 bug   lo l2 bootstrap correction

### LINKED ISSUE #2598: BUG - Lo L2 Bootstrap correction
### Description of the issue

Only apply bootstrap correction where boostrap intensity is valid.

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

tmplummer:
# Change Summary

## Overview
Fixes a few issues with the sputtering and bootstrap corrections.
Also updates the hydrogen geometric factors to a file from Fatemeh.

Fixes to sputtering:
- Only apply sputtering correction where valid values in sputtering are found.
- Set values in numpy arrays using numpy indexing rather than xarray.

Fixes to boostrap:
- Only set values where boostrap is valid
- Fix where that was converting NaNs to zeros

Closes: #2598


greglucas (approved):
Nice work tracking all of this down!

Minor comment that I discussed in-person about whether we should switch to working on numpy arrays right at the beginning of these corrections rather than calling all these `xr.where()` with multiple cases it would be something like `ena_intensity_np = data["ena_intensity"].values; ena_intensity_np[valid_mask] = ...`

PR #2611, opened 2026-01-26
MNT/FIX: Update Lo exposure time calculations for off-angle bins

greglucas:
# Change Summary

This updates the exposure time calculation for Lo pointing sets.

* Accounts for off-angle bins by sampling a nominal (5) set of spins during a pointing at 4096 (nticks/spin?) samples, seeing what bins those fall into. We then histogram the off-angle components to get an off-angle distribution that can be applied evenly to all spin-bins.
* Then, the total pointing time is divided equally into all spin_bins, then multiplied by the off-angle fraction above.
* Finally, the goodtimes bins are turned into a fractional mask. The fractional mask is applied to (esa_energy_step, spin_bin), to account for how much time a given energy step and spin bin were a part of the total exposure time.

### DIFF
--- a/imap_processing/lo/l1c/lo_l1c.py
+++ b/imap_processing/lo/l1c/lo_l1c.py
@@ -5,15 +5,19 @@ from dataclasses import Field
 from enum import Enum
 
 import numpy as np
+import pandas as pd
 import xarray as xr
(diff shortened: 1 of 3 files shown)


tmplummer (approved):
LGTM

PR #2612, opened 2026-01-26
MNT: Add pivot_angle to Lo direct event l1b and pass-through more

greglucas:
This moves the pivot_angle variable into l1b_de, which can then be referenced for derates and pset. Meaning the dependency tree will be updated to move the nhk dependency to the l1b-de instead of l1b-derates.


tmplummer (approved):
👍

PR #2615, opened 2026-01-26
2576 lo star sensor l1b

### LINKED ISSUE #2576: Lo star snesor L1B
### Algorithm Description:

Using algorithm document from Mitchel, implement the code to produce the star sensor L1B product.

tmplummer:
# Change Summary

## Overview

Add the Lo L2 star product.
Add epoch_to_fractional_doy function to time.py

Closes: #2576

### DIFF
--- a/imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
@@ -83,12 +83,6 @@ imap_lo_l1b_prostar:
   Logical_source: imap_lo_l1b_prostar
   Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1B Data
 
-imap_lo_l1b_star:
-  <<: *instrument_base
-  Data_type: L1B_star>Level-1B Star Sensor Spin-Averaged Profile
-  Logical_source: imap_lo_l1b_star
-  Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1B Star Sensor Spin-Averaged Profile Data
-
 imap_lo_l1b_nhk:
   <<: *instrument_base
   Data_type: L1B_star>Level-1B Nominal Housekeeping
--- a/imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
+++ b/imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
@@ -83,6 +83,12 @@ imap_lo_l1b_prostar:
   Logical_source: imap_lo_l1b_prostar
   Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1B Data
 
+imap_lo_l1b_star:
+  <<: *instrument_base
+  Data_type: L1B_star>Level-1B Star Sensor Spin-Averaged Profile
+  Logical_source: imap_lo_l1b_star
+  Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1B Star Sensor Spin-Averaged Profile Data
+
 imap_lo_l1b_nhk:
   <<: *instrument_base
   Data_type: L1B_star>Level-1B Nominal Housekeeping
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -2158,7 +2158,7 @@ def get_sampling_cadence_from_nhk(l1b_nhk: xr.Dataset) -> float:
     # Get mean value across all epochs (should be relatively constant)
     sampling_cadence = float(l1b_nhk["ifb_data_interval"].values.mean())
 
-    logger.info(f"Sampling cadence from NHK: {sampling_cadence:.3f} ms")
+    logger.info(f"Star sensor sampling cadence from NHK: {sampling_cadence:.3f} ms")
     return sampling_cadence
 
 
@@ -2189,7 +2189,7 @@ def l1b_star(
         L1B star sensor dataset with spin_angle, avg_amplitude, count_per_bin,
         and time range metadata. Each epoch corresponds to a group of records.
     """
-    logical_source = "imap_lo_l1b_star"
+    logical_source = "imap_lo_l1b_prostar"
     l1a_star = sci_dependencies["imap_lo_l1a_star"]
     l1b_nhk = sci_dependencies["imap_lo_l1b_nhk"]
     spin_data = sci_dependencies["imap_lo_l1a_spin"]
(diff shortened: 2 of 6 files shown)


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1889,3 +1900,403 @@ def _get_esa_level_indices(epochs: np.ndarray, anc_dependencies: list) -> np.nda
...
+            shcoarse_sec <= (t0 + time_window_offset + time_window_duration)
+        )
+        valid_mask = count_mask & time_mask
+    else:
+        valid_mask = count_mask
+
+    # Filter out repoint maneuvers
+    repoint_df = interpolate_repoint_data(shcoarse_sec)
+    # Exclude times where repoint_in_progress is True
+    repoint_mask = ~repoint_df["repoint_in_progress"].values
```
greglucas, line 1964: This is a cool use of this function! If I'm understanding right, the repoint_df will grow to the size of whatever shcoarse values you have and do linear interpolation between items and you are then taking advantage of the fact that anything other than 0 will evaluate to True and thus give you repoint_in_progress during transitions. Is that the right interpretation?

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1889,3 +1900,403 @@ def _get_esa_level_indices(epochs: np.ndarray, anc_dependencies: list) -> np.nda
...
+        Average sampling cadence in milliseconds.
+    """
+    if "ifb_data_interval" not in l1b_nhk:
+        raise KeyError(
+            "ifb_data_interval field not found in L1B NHK dataset. "
+            "Cannot calculate sampling cadence."
+        )
+
+    # Get mean value across all epochs (should be relatively constant)
+    sampling_cadence = float(l1b_nhk["ifb_data_interval"].values.mean())
```
greglucas, line 2158: In the other use of pivot_angle motion I decided to use "nearest" value to a given shcoarse. I'm wondering if you'd want to somehow subselect the time intervals during the star sensor "group" you have to average only over each individual grouping instead? This is definitely minor and looks good to me as-is because I agree it won't be changing significantly.

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1889,3 +1900,403 @@ def _get_esa_level_indices(epochs: np.ndarray, anc_dependencies: list) -> np.nda
...
+
+    logger.info(f"Sampling cadence from NHK: {sampling_cadence:.3f} ms")
```
greglucas: What is "sampling cadence" do you want to put IFB or something else descriptive in it to describe what this means?

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1889,3 +1900,403 @@ def _get_esa_level_indices(epochs: np.ndarray, anc_dependencies: list) -> np.nda
...
+    Returns
+    -------
+    l1b_star_ds : xr.Dataset
+        L1B star sensor dataset with spin_angle, avg_amplitude, count_per_bin,
+        and time range metadata. Each epoch corresponds to a group of records.
+    """
+    logical_source = "imap_lo_l1b_star"
+    l1a_star = sci_dependencies["imap_lo_l1a_star"]
+    l1b_nhk = sci_dependencies["imap_lo_l1b_nhk"]
+    spin_data = sci_dependencies["imap_lo_l1a_spin"]
```
greglucas, line 2194: I think you will need to add these sds-data-manager dependencies. Also, it looks like we call this dataset "prostar" there, so I'm not sure if this should be just star or prostar.

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -1396,3 +1403,663 @@ def test_calculate_de_rates(
...
+    sci_dependencies = {
+        "imap_lo_l1a_star": load_cdf(star_path),
+        "imap_lo_l1a_spin": load_cdf(spin_path),
+        "imap_lo_l1b_nhk": load_cdf(nhk_path),
+    }
+    anc_dependencies = []
+    descriptor = "star"
+    result = lo_l1b(sci_dependencies, anc_dependencies, descriptor)
+    assert len(result) == 1
+    print(write_cdf(result[0]))
```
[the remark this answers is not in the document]
greglucas: @subagonsouth just making sure you see this comment because it should be updated to remove the print and your directories before merge.

greglucas (approved):
Overall looks great to me! A few minor comments that can be ignored as you see fit.

imap_processing/tests/lo/test_lo_l1b.py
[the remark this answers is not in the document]
tmplummer: Hrmmm... I thought I removed this test. It was just a quick test I wrote to debug with flight data but shouldn't be in this final PR.

imap_processing/cdf/config/imap_lo_global_cdf_attrs.yaml
```diff
@@ -83,6 +83,12 @@ imap_lo_l1b_prostar:
   Logical_source: imap_lo_l1b_prostar
   Logical_source_description: IMAP Mission IMAP-Lo Instrument Level-1B Data
 
+imap_lo_l1b_star:
+  <<: *instrument_base
+  Data_type: L1B_star>Level-1B Star Sensor Spin-Averaged Profile
+  Logical_source: imap_lo_l1b_star
```
mitchellshen: This should be `imap_lo_l1b_prostar` for pipeline processing. ~by Greg and Mitchell

imap_processing/lo/l1b/lo_l1b.py
tmplummer: Mostly right. The "interpolation" method is to find the repoint row for each input MET time such that `repoint_start_time[i] <= MET < repoint_start_time[i+1]`. Maybe a misnomer, but it is sort of nearest neighbor interpolation? Then `repoint_in_progress` is computed and is already a boolean.

imap_processing/lo/l1b/lo_l1b.py
tmplummer: I don't think that this will change at all. The data I checked, had a constant value for the whole pointing, so I don't think that it matters. I believe it is some sort of instrument configuration that sets the sampling time.

imap_processing/lo/l1b/lo_l1b.py
tmplummer: Thanks for taking care of that for me!

PR #2616, opened 2026-01-27
MNT: Use pivot_angle from Lo when defining DE direction

### LINKED ISSUE #2614: Use pivot_angle for direct event calculation in lo l1b
Our current pointing direction calculation uses the IMAP_LO_BASE direction. That assumes a pivot_angle of 0-degrees. We need to update this direction to use the current pivot_angle in-use that has the proper rotation matrix applied until we get a real CK kernel for the pivot platform.

greglucas:
The default IMAP_LO_BASE is 0 degrees, but the nominal pivot angle is 90 degrees, which we need to use when defining look directions in the sky.

Testing locally, I get off_angles right in the middle at index 20 now rather than off the edge at 40 like we did previously. This brings up the question of what does off-angle mean if we have a pivot angle that is not sweeping out a perpendicular stripe. I _think_ it may all work out with how we've defined the transforms, but I don't know for sure. i.e. will the off-angle be 15 +/- 2 instead of 0 +/- 2 if we have a pivot-angle not directly at 90 degrees. I'm just writing this down so I don't forget about this later.

closes #2614

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -20,7 +20,7 @@ from imap_processing.spice.geometry import (
     SpiceFrame,
     cartesian_to_latitudinal,
     frame_transform,
-    instrument_pointing,
+    lo_instrument_pointing,
 )
 from imap_processing.spice.repoint import get_pointing_times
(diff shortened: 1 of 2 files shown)


tmplummer (approved):
👍

PR #2617, opened 2026-01-27
MNT: Change the definition of H/O that goes into a Lo pset

### LINKED ISSUE #2613: Hydrogen updates for IMAP Lo
Identify golden triples and only use these for maps.
* coincidence_type == 0
* mode_bit == 1

`tof0_s = tof0 + tof3 * 0.5`
`tof1_s = tof1 - tof3 * 0.5`

In L1c do a complete reselection, not using species like we are currently doing.

tof0_s and tof1_s and tof2

Peaks

```
H_PEAK_L = [20, 10, 10]
H_PEAK_H = [70, 50, 40]

CO_PEAK_L = [100, 60, 60]
CO_PEAK_H = [270, 150, 150]
```

greglucas:
Even though something may be identified as an H species, we want to be a little bit more strict in the pointing set H/O counts and require a golden triple with specific ranges for H and O TOFs to identify the species.

Refactor out the logic for filtering species into helper functions rather than inlining this more complex logic.

NOTE: I need to verify the coincidence types and range values with the Lo team still to verify those are what they want in the pointing set. The "species" identifier is still OK to be "H" in the l1b file, this is just more strict for the l1c pset values apparently.

closes #2613

### DIFF
--- a/imap_processing/lo/l1c/lo_l1c.py
+++ b/imap_processing/lo/l1c/lo_l1c.py
@@ -244,8 +244,173 @@ def filter_goodtimes(l1b_de: xr.Dataset, anc_dependencies: list) -> xr.Dataset:
     return filtered_epochs
 
 
+def get_triple_coincidences(de: xr.Dataset) -> xr.Dataset:
+    """
+    Get only the triple coincidence events from the L1B Direct Event dataset.
+
+    Parameters
+    ----------
+    de : xarray.Dataset
+        L1B Direct Event dataset.
+
+    Returns
+    -------
+    de_triples : xarray.Dataset
+        L1B Direct Event dataset with only triple coincidence events.
+    """
+    triple_types = ["111111", "111100", "111000"]
+    triple_idx = np.nonzero(np.isin(de["coincidence_type"], triple_types))[0]
+    de_triples = de.isel(epoch=triple_idx)
+
+    return de_triples
+
+
+def get_double_coincidences(de: xr.Dataset) -> xr.Dataset:
+    """
+    Get only the double coincidence events from the L1B Direct Event dataset.
+
+    Parameters
+    ----------
+    de : xarray.Dataset
+        L1B Direct Event dataset.
+
+    Returns
+    -------
+    de_doubles : xarray.Dataset
+        L1B Direct Event dataset with only double coincidence events.
+    """
+    double_types = [
+        "110100",
+        "110000",
+        "101101",
+        "101100",
+        "101000",
+        "100100",
+        "100101",
+        "100000",
+        "011100",
+        "011000",
+        "010100",
+        "010101",
+        "010000",
+        "001100",
+        "001101",
+        "001000",
+    ]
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -244,8 +244,172 @@ def filter_goodtimes(l1b_de: xr.Dataset, anc_dependencies: list) -> xr.Dataset:
...
+    de : xarray.Dataset
+        L1B Direct Event dataset.
+
+    Returns
+    -------
+    de_triples : xarray.Dataset
+        L1B Direct Event dataset with only triple coincidence events.
+    """
+    triple_types = ["111111", "111100", "111000"]
+    triple_idx = np.where(np.isin(de["coincidence_type"], triple_types))[0]
```
tmplummer: For a while now, I always try to comment that np.nonzero should be used instead of where... That's according to numpy documentation, but `where` is so ubiquitous that I'm not sure anymore.

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -244,8 +244,172 @@ def filter_goodtimes(l1b_de: xr.Dataset, anc_dependencies: list) -> xr.Dataset:
...
+        "011100",
+        "011000",
+        "010100",
+        "010101",
+        "010000",
+        "001100",
+        "001101",
+        "001000",
+    ]
+    double_idx = np.where(np.isin(de["coincidence_type"], double_types))[0]
```
tmplummer: Same comment about using `np.nonzero()`

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -244,8 +244,172 @@ def filter_goodtimes(l1b_de: xr.Dataset, anc_dependencies: list) -> xr.Dataset:
...
+    de : xarray.Dataset
+        L1B Direct Event dataset.
+
+    Returns
+    -------
+    de_h : xarray.Dataset
+        L1B Direct Event dataset with only hydrogen species.
+    """
+    h_peak_low = [20, 10, 10]
+    h_peak_high = [70, 50, 40]
```
tmplummer, line 376: It seems like these may change. Should we write a ticket to put them in an ancillary file?

tmplummer (approved):
Looks good other than minor nits

imap_processing/lo/l1c/lo_l1c.py
greglucas: Yeah, I agree with you here. I was mostly copying what was here before and moving it around. Surprisingly the `nonzero()` doesn't work with xarray arrays and I had to explicitly get the `.values()` out of it. Maybe there is a better way to use xarray.where or something like that instead even...

imap_processing/lo/l1c/lo_l1c.py
greglucas: Yes, good idea.

https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/2622

PR #2627, opened 2026-01-27
FIX: Update the rotation direction for the lo pivot angle

greglucas:
We need to rotate in the + pivot angle direction. There was an incorrect negative sign to rotate counter clockwise before.


greglucas:
OK, we just did some quick investigations and think the "-" rotation was actually correct because we want the X/Y values to be positive in the output when pivot platform is around 90 degrees.

tmplummer (approved):
Agree with rotating by `-pivot_angle`. This is because spiceypy.rotate provides the rotation matrix that is  "the rotation matrix which describes the rotation of the COORDINATE system through angle radians about the axis whose index is iaxis." Which I have convinced myself means that it is the inverse of the rotation matrix that we want.

PR #2628, opened 2026-01-27
I-ALiRT - Coordinated coverage schedule

laspsandoval:
# Change Summary

## Overview
I've been asked to start to plan for overlapping stations and creating schedules for them. 

## Updated Files
- generate_coverage.py
   - Handle stations so that they get priority. For example, we will always seek the most coverage from Kiel and inform other stations to not attempt to take data when Kiel is available. 
   - Add capability for stations to fill in gaps if there is ever a DSN outage.

## Testing
- test_generate_coverage.py

### DIFF
--- a/imap_processing/ialirt/generate_coverage.py
+++ b/imap_processing/ialirt/generate_coverage.py
@@ -4,7 +4,7 @@ import logging
 
 import numpy as np
 
-from imap_processing.ialirt.constants import STATIONS
+from imap_processing.ialirt.constants import STATIONS, StationProperties
 from imap_processing.ialirt.process_ephemeris import calculate_azimuth_and_elevation
 from imap_processing.spice.time import et_to_utc, str_to_et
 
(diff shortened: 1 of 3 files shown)


imap_processing/ialirt/generate_coverage.py
```diff
@@ -89,9 +103,13 @@ def generate_coverage(
...
 
-        visible[outage_mask] = False
-        # DSN contacts block other stations
-        visible[dsn_outage_mask] = False
+        # Block this station if DSN is active OR already-occupied by earlier stations
+        unavailable_mask = outage_mask | dsn_occupied_mask | non_dsn_occupied_mask
+        visible = visible_unblocked & ~unavailable_mask
+
+        # This station now occupies these times and will block later stations
+        non_dsn_occupied_mask |= visible
```
greglucas: Is this actually what you want? This means that the way this loop is defined will dictate the order and amount of time that the stations will cover. If there is 1-hour of overlap but the first station has 4 hours and the second station has 8 hours, that would set this to 4-hours for the first station every time and only 7-hours for the second station, but you might want that second station to take precedence because it has the longer block of time.

In general, I think this is OK right now but it should get some more thought to how we want to approach it. One other idea is to not block based on ground-stations but rather indicate how many stations could cover a specific area, so DSN can block, but the others just add to the coverage time.

greglucas (approved):

PR #2630, opened 2026-01-27
Updates to include missing ancillary file for glows

maxinelasp:
# Change Summary
During the ancillary file updates for GLOWS, I missed one. This PR will have a matching one to add the dependnecy in sds-data-manager. It primarily just wires up the file to the already-existing dataclass for passing the information around, so not a major change. 

## Overview

- Add CLI.py management of the l1b conversion table ancillary file
- Add a xr.Dataset->AncillaryParameters conversion step
- Pass around AncillaryParameters where needed.

## Testing

- Explicitly defined the values for the test from the variables

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -13,6 +13,7 @@ Examples
 from __future__ import annotations
 
 import argparse
+import json
 import logging
 import re
 import sys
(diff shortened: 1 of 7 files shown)


greglucas (approved):
Looks good. There are a lot of places in code you are adding this pass-through so I'm not sure if there would be a better way to bundle this up beforehand to avoid needing to add it in so many locations, say if you get yet another ancillary file or something along those lines.

maxinelasp:
@greglucas I agree, but I don't expect to get more files (this is not a change from GLOWS, rather just something I missed when I set up the rest of the inputs for whatever reason) so hopefully this is the last edit.

Issue #2507, opened 2025-12-11
BUG - exclusion of zeros in SWAPI I-ALiRT fitting


hafarooki:
### Description of the issue

For times when there is a zero count in one of the passbands near the proton peak in SWAPI, which is more common than expected because the I ALiRT counts are rounded down, the initial guess is returned as the fit by the algorithm. Need to exclude zeros from the fit. Extracted from #2485 to be fixed in a new PR because the PR scope had grown too broad.

### Code Snippet:

```python
Code
```

PR #2632, opened 2026-01-28
updates to swapi ialirt fitting algorithm

hafarooki:
# Change Summary

## Overview
updates to swapi ialirt fitting algorithm
1. adds a temporary correction factor based on WIND measurements to be used until the SWAPI L3 science data pipeline is finalized
2. report speed only when the fits fails 
3. use `raw count * 16 + 8` instead of `raw count * 16` to avoid truncating to zero counts, which can adversely affect the fits

## Updated Files
`process_swapi.py`: made the aforementioned changes

## Testing
updated `test_process_swapi.py` to account for the density correction, but did **not** yet create a test for reporting the speed only when the fits fail, nor for the `+8` shift in counts from the packet. For the latter, perhaps we could add a test using a real packet wherein there are only 3 nonzero counts near the peak.

### DIFF
--- a/imap_processing/ialirt/l0/process_swapi.py
+++ b/imap_processing/ialirt/l0/process_swapi.py
@@ -2,7 +2,6 @@
 
 import logging
 from decimal import Decimal
-from math import isfinite
 
 import numpy as np
 import pandas as pd
@@ -61,7 +60,7 @@ def count_rate(
 
     # see comment on Consts.temporary_density_factor
     density = density * Consts.temporary_density_factor
-    
+
     return (
         (density * Consts.eff_area * (beta / np.pi) ** (3 / 2))
         * (np.exp(-beta * (center_speed**2 + speed**2 - 2 * center_speed * speed)))
@@ -109,9 +108,9 @@ def optimize_pseudo_parameters(
             60000 * (initial_speed_guess / 400) ** 2,
         ]
     )
-    
+
     sol = None
-    
+
     try:
         five_point_range = range(max_index - 2, max_index + 2 + 1)
         xdata = energy_passbands.take(five_point_range, mode="clip")
@@ -124,26 +123,27 @@ def optimize_pseudo_parameters(
             sigma=sigma,
             p0=initial_param_guess,
         )
-        
-        # if covariance matrix is not finite, scipy failed to converge to a solution and could just be reporting the initial guess
+
+        # If covariance matrix is not finite, scipy failed to converge to a
+        # solution and could just be reporting the initial guess
         covariance_matrix_is_finite = np.all(np.isfinite(curve_fit_output[1]))
 
         # fit has failed if R^2 < 0.7
         yfit = count_rate(xdata, *curve_fit_output[0])
-        R2 = 1 - np.sum((ydata - yfit) ** 2) / np.sum((ydata - ydata.mean()) ** 2)
-        R2_is_acceptable = R2 >= 0.7
-        
-        if covariance_matrix_is_finite and R2_is_acceptable:
+        r2 = 1 - np.sum((ydata - yfit) ** 2) / np.sum((ydata - ydata.mean()) ** 2)
+        r2_is_acceptable = r2 >= 0.7
+
+        if covariance_matrix_is_finite and r2_is_acceptable:
             sol = curve_fit_output[0]
-    except RuntimeError as runtime_error:
-        logger.error(f"curve_fit failed", runtime_error)
+    except RuntimeError:
+        logger.error("curve_fit failed")
         sol = None
 
-    # report speed only if fit fails            
+    # report speed only if fit fails
     if sol is None:
         sol = initial_param_guess.copy()
-        sol[1:] = FILLVAL_FLOAT32 
-    
+        sol[1:] = FILLVAL_FLOAT32
+
     return sol
 
 
@@ -270,7 +270,8 @@ def process_swapi_ialirt(
 
         raw_coin_count = process_sweep_data(grouped_subset, "swapi_coin_cnt")
         # I-ALiRT packets have counts compressed by a factor of 16.
-        # Add 8 to avoid having counts truncated to 0 and to avoid counts being systematically too low
+        # Add 8 to avoid having counts truncated to 0 and to avoid
+        # counts being systematically too low
         raw_coin_count = raw_coin_count * 16 + 8
         # Subset to only the relevant I-ALiRT energy steps
         raw_coin_count = raw_coin_count[:, :NUM_IALIRT_ENERGY_STEPS]
--- a/imap_processing/ialirt/l0/process_swapi.py
+++ b/imap_processing/ialirt/l0/process_swapi.py
@@ -325,11 +325,18 @@ def process_swapi_ialirt(
             )
 
             # replace nans (resulting from geometric means that include fill values) with fill values
-            avg_pseudo_proton_speed, avg_pseudo_proton_density, avg_pseudo_proton_temperature = np.nan_to_num((
+            (
                 avg_pseudo_proton_speed,
                 avg_pseudo_proton_density,
-                avg_pseudo_proton_temperature
-            ), nan=FILLVAL_FLOAT32)
+                avg_pseudo_proton_temperature,
+            ) = np.nan_to_num(
+                (
+                    avg_pseudo_proton_speed,
+                    avg_pseudo_proton_density,
+                    avg_pseudo_proton_temperature,
+                ),
+                nan=FILLVAL_FLOAT32,
+            )
 
             swapi_data.append(
                 _populate_instrument_header_items(met)
--- a/imap_processing/ialirt/l0/process_swapi.py
+++ b/imap_processing/ialirt/l0/process_swapi.py
@@ -2,6 +2,7 @@
 
 import logging
 from decimal import Decimal
+from math import isfinite
 
 import numpy as np
 import pandas as pd
@@ -22,6 +23,7 @@ from imap_processing.swapi.l2.swapi_l2 import SWAPI_LIVETIME
 logger = logging.getLogger(__name__)
 
 NUM_IALIRT_ENERGY_STEPS = 63
+FILLVAL_FLOAT32 = -1.0e31
 
 
 def count_rate(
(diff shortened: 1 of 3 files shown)


hafarooki:
related: #2485 
closes #2507

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -57,6 +54,13 @@ def count_rate(
     speed = speed * 1000  # convert km/s to m/s
     density = density * 1e6  # convert 1/cm**3 to 1/m**3
 
+    # temporary correction factor based on WIND data available
+    # overlapping with the first ~month of SWAPI data.
+    # to be replaced once SWAPI's L3 processing pipeline is finalized
+    # this will increase the model count by a factor of e^1,
+    # changing the output density by a factor of e^-1.
+    density = density * np.exp(1)
```
laspsandoval: Move this up to the top so it is easy to change and not calculated every time.
DENSITY_MODEL_SCALE = np.exp(1)

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -91,82 +95,53 @@ def optimize_pseudo_parameters(
...
+                    range(max_index - 3, max_index + 3), mode="clip"
+                ),
+                sigma=current_sweep_count_rate_errors.take(
+                    range(max_index - 3, max_index + 3), mode="clip"
+                ),
+                p0=initial_param_guess,
+            )[0]
+        except Exception as _:
+            sol = initial_param_guess.copy()
+            sol[1:] = np.nan  # report the speed only if the fit fails
```
laspsandoval: We are using FILL values instead of nan since downstream we will create cdfs (and that is what SPDF has requested in the past). This file has all fill values listed:
imap_ialirt_l1_variable_attrs.yaml


swapi_pseudo_proton_density:
  <<: *default_float32
  CATDESC: Pseudo density of solar wind protons
  FIELDNAM: swapi_pseudo_proton_density
  LABLAXIS: swapi_pseudo_proton_density
  UNITS: 1/cm^3
  VALIDMIN: 0.0
  VALIDMAX: 1000.0

swapi_pseudo_proton_speed:
  <<: *default_float32
  CATDESC: Pseudo speed of solar wind protons in solar inertial frame
  FIELDNAM: swapi_pseudo_proton_speed
  LABLAXIS: swapi_pseudo_proton_speed
  UNITS: km/sec
  VALIDMIN: 0.0
  VALIDMAX: 10000.0

swapi_pseudo_proton_temperature:
  <<: *default_float32
  CATDESC: Pseudo temperature of solar wind protons in plasma frame
  FIELDNAM: swapi_pseudo_proton_temperature
  LABLAXIS: swapi_pseudo_proton_temperature
  UNITS: Kelvin
  VALIDMIN: 0.0
  VALIDMAX: 50000000.0

default_float32_attrs: &default_float32
  <<: *default
  FILLVAL: -1.0e31

So in other words add this to the top of your file and use it instead of np.nan.
FILLVAL_FLOAT32 = -1.0e31

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -91,82 +95,53 @@ def optimize_pseudo_parameters(
...
+                ),
+                ydata=current_sweep_count_rates.take(
+                    range(max_index - 3, max_index + 3), mode="clip"
+                ),
+                sigma=current_sweep_count_rate_errors.take(
+                    range(max_index - 3, max_index + 3), mode="clip"
+                ),
+                p0=initial_param_guess,
+            )[0]
+        except Exception as _:
```
laspsandoval: Add a log:

logger.info(f"curve_fit failed for <add info here>")

laspsandoval (commented):
Looks good. Just a few things to change. Looking forward to seeing the tests.

imap_processing/ialirt/l0/process_swapi.py
```diff
@@ -222,95 +188,77 @@ def process_swapi_ialirt(
...
         logger.info(
             f"The following swapi groups were skipped due to "
             f"missing or duplicate pkt_counter values: "
             f"{incomplete_groups}"
         )
 
+    raw_coin_count = process_sweep_data(grouped_dataset, "swapi_coin_cnt")
+    # I-ALiRT packets have counts compressed by a factor of 16.
+    # Add 8 to avoid having counts truncated to 0 and to avoid counts being systematically too low
+    raw_coin_count = raw_coin_count * 16 + 8
```
greglucas: do you want to only do this on the minimum side or you do want to do the systematic shift to higher counts too?

```python
raw_coin_count *= 16
raw_coin_count = min(raw_coin_count, 8)

File: imap_processing/ialirt/l0/process_swapi.py (hunk 3 above)
Comment (greglucas) [reply imap_processing/ialirt/l0/process_swapi.py]: Can you narrow the Exception down to something like ValueError, or similar so it isn't so broad?

File: imap_processing/ialirt/l0/process_swapi.py (hunk 1 above)
Comment (greglucas) [reply imap_processing/ialirt/l0/process_swapi.py]: I agree making this a constant would be good, and then that constant can be brought into the tests. Additionally, it can be changed to `1` later for a quick fix to get back to reality if needed.

Maybe call this `TEMPORARY_SCALE_FACTOR` or something to indicate it shouldn't be relied upon.

File: imap_processing/ialirt/l0/process_swapi.py (hunk 3 above)
Comment (hafarooki) [reply imap_processing/ialirt/l0/process_swapi.py]: Does this work? Would be nice to add information about what time in the data failed, but this function doesn't have that information
```python
    except RuntimeError as runtime_error:
        logger.error(f"curve_fit failed", runtime_error)
        sol = None
        ```

File: imap_processing/ialirt/l0/process_swapi.py (hunk 4 above)
Comment (hafarooki) [reply imap_processing/ialirt/l0/process_swapi.py]: We want to do this all the time because for a given count compressed through count = count // 16, we can decompress it anywhere between `count * 16` and `(count + 1) * 16 - 1`. To be correct on average (and, conveniently, avoid zero counts as well), we simply take the middle of the range (count * 16 + 8). Previously, we were taking the beginning of the range.

File: imap_processing/ialirt/l0/process_swapi.py (hunk 3 above)
Comment (greglucas) [reply imap_processing/ialirt/l0/process_swapi.py]: Is it a RuntimeError that gets thrown? I don't know what the failure mode of curve_fit is, but I figured it would have been something a little more granular that we could catch. But yes, that would be fine too.

Comment (hafarooki) [reply imap_processing/ialirt/l0/process_swapi.py]: Yep, unfortunately, they don't have a more specific exception:
image

Comment (hafarooki) [reply imap_processing/ialirt/l0/process_swapi.py]: @laspsandoval for the logging message, is this what you had in mind?

**Comment by hafarooki** [comment]
so far, i've added tests for everything except real packets with only 3 near the peak that havent been truncated to 0

**Comment by laspsandoval** [comment]
pre-commit.ci autofix

File: imap_processing/ialirt/l0/process_swapi.py [hunk 5]
```diff
@@ -324,22 +324,27 @@ def process_swapi_ialirt(
...
+                            f"{avg_pseudo_proton_speed:.3f}"
+                        ),
+                        "swapi_pseudo_proton_density": Decimal(
+                            f"{avg_pseudo_proton_density:.3f}"
+                        ),
+                        "swapi_pseudo_proton_temperature": Decimal(
+                            f"{avg_pseudo_proton_temperature:.3f}"
+                        ),
+                    }
+                )
```
hafarooki: @laspsandoval  I think this will result in times when _any_ of these parameters are fill values being excluded from the final output. If there are fill values, the geometric mean will be nan. This is not desired behavior because we want to include the speed even if the other two are fill values. I think it should be like this instead:
```python
avg_pseudo_proton_speed, avg_pseudo_proton_density, avg_pseudo_proton_temperature = np.nan_to_num((
    avg_pseudo_proton_speed,
    avg_pseudo_proton_density,
    avg_pseudo_proton_temperature
), FILLVAL_FLOAT32)

swapi_data.append(
    _populate_instrument_header_items(met)
    | {
        "instrument": "swapi",
        "swapi_epoch": int(met_to_ttj2000ns(avg_swapi_met)),
        "swapi_pseudo_proton_speed": Decimal(
            f"{avg_pseudo_proton_speed:.3f}"
        ),
        "swapi_pseudo_proton_density": Decimal(
            f"{avg_pseudo_proton_density:.3f}"
        ),
        "swapi_pseudo_proton_temperature": Decimal(
            f"{avg_pseudo_proton_temperature:.3f}"
        ),
    }
)
```

imap_processing/tests/ialirt/unit/test_process_swapi.py
```diff
@@ -180,12 +182,52 @@ def test_process_swapi_ialirt(
...
+        "swapi_coin_cnt4",
+        "swapi_coin_cnt5",
+    ]
+
+    for v in vars_to_zero:
+        xarray_data[v] = xr.zeros_like(xarray_data[v])
+
+    swapi_result = process_swapi_ialirt(xarray_data, esa_unit_conversion_table)
+
+    assert swapi_result == []
```
hafarooki: @laspsandoval I don't think this test is testing the expected behavior in the presence of zero counts correctly. What we need to test is that if the five points around the peak before decompression are `(0, 11, 50, 4, 0)` for example, this is decompressed as `(8, 11*16+8, 50*16+8, 4*16+8, 8)` instead of `(0, 11*16, 50*16, 4*16, 0)` , and that the fit now converges in such cases as a result.

hafarooki: A simple fix is to use a real packet that has zero counts near the peak, which is quite common

imap_processing/ialirt/l0/process_swapi.py
hafarooki: can confirm, current implementation has the wrong behavior; it skips reporting the average entirely when only the speed is not a fill value

imap_processing/tests/ialirt/unit/test_process_swapi.py
laspsandoval: Ok. I will let you update the test accordingly.

hafarooki:
The PR is ready to merge now, but we don't currently have any unit tests using real spacecraft packets. For now, I think it's OK to merge without it, but there should really be a unit test with real spacecraft packets, particularly one with zero counts.

hafarooki:
pre-commit.ci autofix

laspsandoval (approved):

PR #2633, opened 2026-01-28
CoDICE direct events: add fillvals for missing priories instead of dropping full groups

lacoak21:
# Change Summary

## Overview
Joey is padding incomplete groups with fillvals instead of dropping them like we are. This is causing a mismatch along the epoch dimension. This PR updates our code to do the same. 

***** This PR was created with the help of claude ***

## Updated Files
- imap_processing/codice/codice_l1a_de.py
   - instead of dropping incomplete groups, pad them with fillvals for the missing priorities

## Testing
Added a new test for incomplete priority group handling

### DIFF
--- a/imap_processing/codice/codice_l1a_de.py
+++ b/imap_processing/codice/codice_l1a_de.py
@@ -374,7 +374,8 @@ def process_de_data(
     # with the same acq_start_seconds value
     acq_start_seconds = packets["acq_start_seconds"].values
     unique_times, counts = np.unique(acq_start_seconds, return_counts=True)
-    # Find incomplete groups
+
+    # Find incomplete groups (not exactly num_priorities packets)
     incomplete_mask = counts != num_priorities
     if np.any(incomplete_mask):
         incomplete_times = unique_times[incomplete_mask]
@@ -388,9 +389,10 @@ def process_de_data(
 
         # Creat a list of groups with padding if any priorities are missing
         padded_groups = []
-        for time_val, count in zip(unique_times, counts, strict=False):
-            group_mask = acq_start_seconds == time_val
-            group = packets.isel(epoch=group_mask)
+        for time, count in zip(unique_times, counts, strict=False):
+            # Get the packets for this group
+            group_ids = np.where(acq_start_seconds == time)[0]
+            group = packets.isel(epoch=group_ids)
             if count < num_priorities:
                 # Find missing priorities
                 existing_priorities = set(group["priority"].values)
@@ -398,8 +400,8 @@ def process_de_data(
                     [p for p in range(num_priorities) if p not in existing_priorities]
                 )
                 num_missing_priorities = len(missing_priorities)
-                # Create a Dataset to hold the padded packets
-                # Use first packet as a template
+                # Use first packet as a template and expand along the epoch dimension
+                # for the number of missing priorities.
                 pad_packet = group.isel(epoch=[0] * num_missing_priorities).copy()
                 # Set padding values
                 pad_packet["num_events"].values = np.full(num_missing_priorities, 0)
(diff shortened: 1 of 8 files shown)


lacoak21:
@greglucas I ended up writing this because I wasnt sure where else to check when our data was mismatching. This helped me figure out what was going on with Joey's validation data. I guess its a win win because now we have this code.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -384,16 +384,51 @@ def process_de_data(
...
+                pad_packet["priority"].values = missing_priorities
+                # Set event_data to empty object arrays for padding packets
+                for i in range(num_missing_priorities):
+                    pad_packet["event_data"].data[i] = np.array([], dtype=np.uint8)
+                # Concatenate the existing priorities with the fillval priority groups
+                group = xr.concat([group, pad_packet], dim="epoch")
+                # Sort by priority
+                sort_idx = np.argsort(group["priority"].values)
+                group = group.isel(epoch=sort_idx)
+            elif counts > num_priorities:
```
laspsandoval: This should be "count" not "counts" I think.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -384,16 +384,51 @@ def process_de_data(
...
-    keep_mask = np.isin(acq_start_seconds, complete_times)
-    packets = packets.isel(epoch=keep_mask)
-
-    # Calculate number of epochs from complete groups
-    num_epochs = len(complete_times)
+        # Create a list of groups with padding if any priorities are missing
+        padded_groups = []
+        for time, count in zip(unique_times, counts, strict=False):
+            # Get the packets for this group
+            group_ids = np.where(acq_start_seconds == time)[0]
```
laspsandoval, line 394: Just checking. Will the acquisition time always be the same for each group?

imap_processing/codice/codice_l1a_de.py
```diff
@@ -384,16 +384,51 @@ def process_de_data(
...
+                existing_priorities = set(group["priority"].values)
+                missing_priorities = sorted(
+                    [p for p in range(num_priorities) if p not in existing_priorities]
+                )
+                num_missing_priorities = len(missing_priorities)
+                # Use first packet as a template and expand along the epoch dimension
+                # for the number of missing priorities.
+                pad_packet = group.isel(epoch=[0] * num_missing_priorities).copy()
+                # Set padding values
+                pad_packet["num_events"].values = np.full(num_missing_priorities, 0)
```
laspsandoval, line 407: Should the padding values be zeros or FILL vals? Just a question on this. I don't think any of this affects I-ALiRT so whatever makes sense for regular processing.

laspsandoval (approved):
There is one bug (I think) that you should correct. Other comments are just for you to double check.

Fix this one:
This should be "count" not "counts" I think.

imap_processing/codice/codice_l1a_de.py
greglucas: Yes, they are and that is the only way to identify a group from what we've found so far.

imap_processing/codice/codice_l1a_de.py
```diff
@@ -384,16 +384,51 @@ def process_de_data(
...
+        for time, count in zip(unique_times, counts, strict=False):
+            # Get the packets for this group
+            group_ids = np.where(acq_start_seconds == time)[0]
+            group = packets.isel(epoch=group_ids)
+            if count < num_priorities:
+                # Find missing priorities
+                existing_priorities = set(group["priority"].values)
+                missing_priorities = sorted(
+                    [p for p in range(num_priorities) if p not in existing_priorities]
+                )
```
greglucas, line 401: Something like this instead? `set(range(num_priorities)) - existing_priorities`

imap_processing/codice/codice_l1a_de.py
lacoak21: Good catch! Thanks

imap_processing/codice/codice_l1a_de.py
lacoak21: I actually think it should be zeros. Im going to update the comments for this.

greglucas (approved):
I think this looks reasonable to me. I haven't done a thorough review, but seems like it is right at first glance.

PR #2635, opened 2026-01-29
FIX: Lo update spin time calculation

greglucas:
- Use universal spin table to get true onboard timing. The Lo spin packet is not accurate enough or has different time information in it, so we need to reference DE events to the onboard timekeeping (spin table), which the Lo instrument references to reset its counter.

- Change formula for DE time tag counter. It is a fixed length pulse of 4.096ms, not varying with spin period. This means we need to multiply each DE tick by that length of time.

- Change spin bins from (-180, 180) to (0, 360)

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -28,7 +28,11 @@ from imap_processing.spice.repoint import (
     get_pointing_times,
     interpolate_repoint_data,
 )
-from imap_processing.spice.spin import get_spin_data, get_spin_number
+from imap_processing.spice.spin import (
+    get_spin_data,
+    get_spin_number,
+    interpolate_spin_data,
+)
 from imap_processing.spice.time import (
     epoch_to_fractional_doy,
     et_to_utc,
@@ -166,6 +170,7 @@ MONITOR_RATE_FIELDS = [
     "spin_cycle",
 ]
 # -------------------------------------------------------------------
+DE_CLOCK_TICK_S = 4.096e-3  # seconds per DE clock tick
 
 
 def lo_l1b(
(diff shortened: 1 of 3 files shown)


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -675,59 +673,51 @@ def get_spin_start_times(
...
-    avg_spin_durations: xr.DataArray,
 ) -> xr.Dataset:
     """
     Get the event MET for each direct event.
 
     Each direct event is converted from a data number to engineering unit in seconds.
-    de_eu_time de_dn_time / 4096 * avg_spin_duration
-    where de_time is the direct event time Data Number (DN) and avg_spin_duration
-    is the average spin duration for the ASC that the event was measured in.
+    time_from_start_of_spin = de_time * DE_CLOCK_TICK_S
```
greglucas, line 834: Should we be adding a half-tick here?
`(de_time + 0.5) * DE_CLOCK_TICK_S`

imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -675,59 +673,51 @@ def get_spin_start_times(
...
+    spin_df = get_spin_data()
+
+    # For each direct event, find the closest spin start time
+    # based on the DE's shcoarse time
+    de_shcoarse = l1a_de["shcoarse"].values
+    spin_start_sec = spin_df["spin_start_sec_sclk"].values
+
+    # Find the index of the closest spin start for each direct event
+    # We want the closest prior spin (spin that has already started)
+    closest_spin_indices = (
```
tmplummer: This basically duplicates the code in the `interpolate_spin_data` function. I assume that you just didn't know about that function... or is there something there that is undesirable for use here?

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spice/spin.py#L150

imap_processing/lo/l1b/lo_l1b.py
tmplummer: For Hi, each DE is tagged with the exact clock time (not adding 1/2 of a tick). When computing spin-phase or geometry, the 1/2 tick is added.

tmplummer (approved):
Looks good other than the recommendation to check the referenced spin function. I will approve and let you decide if you want to use that function or not.

imap_processing/lo/l1b/lo_l1b.py
greglucas: I need the exact start of the spin, not an interpolation. Is there a way to get that?

greglucas: OK, I don't like the name "interpolate_spin_data()" 😂 That confused me as I thought all values would be interpolated. I think the spin-phase may be interpolated, but no other quantities from the original dataframe.

I refactored the code to use that function now though and I added a test to `test_spin.py` to convince myself this is doing what I wanted it to.

Thanks for mentioning this!

PR #2636, opened 2026-01-29
Lo l1b off angle

greglucas:
# Change Summary

The second commit here is the new code. We need to account for the pivot angle in the "off_angle" calculations so that we are +/-2 degrees from that pivot_angle offset.

### DIFF
--- a/imap_processing/lo/l1b/lo_l1b.py
+++ b/imap_processing/lo/l1b/lo_l1b.py
@@ -1313,6 +1313,14 @@ def set_pointing_bin(l1b_de: xr.Dataset) -> xr.Dataset:
     lons = (lons + 360) % 360
     # third column: latitude
     lats = direction[:, 2]
+    # we want this relative to the pivot angle
+    # i.e. the off_angle is +/- 2 degrees from the pivot angle
+    lats = lats - (90 - l1b_de["pivot_angle"].values[0])
+    if np.any(lats < -2) or np.any(lats > 2):
+        logger.warning(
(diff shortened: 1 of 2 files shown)


imap_processing/lo/l1b/lo_l1b.py
```diff
@@ -1198,6 +1198,14 @@ def set_pointing_bin(l1b_de: xr.Dataset) -> xr.Dataset:
     lons = (lons + 360) % 360
     # third column: latitude
     lats = direction[:, 2]
+    # we want this relative to the pivot angle
+    # i.e. the off_angle is +/- 2 degrees from the pivot angle
+    lats = lats - (90 - l1b_de["pivot_angle"].values[0])
```
tmplummer, line 1318: I'm wondering if you really want to modify the `lats` here, or if it would make more sense to modify the `lat_bins`. The concern is that the off_angle values sort of lose the fact that they relative to the pivot angle. I would think that a user would want off_angle_bins to be in latitudinal degrees.

tmplummer (commented):
Just want to check that this is what is desired.

imap_processing/lo/l1b/lo_l1b.py
greglucas: I was also not sure, but I spoke with Lo yesterday and they want off_angle to be the range +/- 2, so relative to the pivot_angle no matter what the pivot_angle is. I agree this adds confusion though.

tmplummer (approved):

PR #2637, opened 2026-01-29
Lo l1c off angle

greglucas:
# Change Summary

In the DPS frame, we need to shift our off-angles to be relative to the current pivot angle which is contained within the pointing set. This is the same shift that happens in l1b. The 3rd commit is the new one here.


tmplummer (approved):
LGTM

PR #2645, opened 2026-01-30
MNT/FIX: Flag incomplete ASCs rather than removing them

greglucas:
# Change Summary

The Lo team was confused why there was a mismatch in data between L1a histogram and the L1b rates products. This update keeps the data and adds a flag for that timestamp instead of removing it.


greglucas:
I spoke with Nathan and this is prefereable and they are fine with this naming of the variable.

tmplummer (approved):
Approve

PR #2648, opened 2026-01-30
Codice l2 keep energy_table dim

lacoak21:
# Change Summary

## Overview
Keep energy table and fix fill val 

## Updated Files
- imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
   - Use the correct fillval
   - Do not drop energy_table. I was dropping it because "energy_table" wasnt in Joeys CDFs but it was under "energy_per_charge"

### DIFF
--- a/imap_processing/codice/codice_l2.py
+++ b/imap_processing/codice/codice_l2.py
@@ -649,7 +649,9 @@ def process_lo_angular_intensity(
     dataset["spin_sector"].attrs = cdf_attrs.get_variable_attributes(
         "spin_sector", check_schema=False
     )
-
+    print(dataset.energy_table)
+    if "energy_table" in dataset:
+        print("HIII")
     return dataset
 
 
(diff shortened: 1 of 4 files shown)


imap_processing/codice/codice_l2.py
```diff
@@ -1423,7 +1423,6 @@ def process_codice_l2(
                 "acquisition_time_per_esa_step",
                 "rgfo_half_spin",
                 "half_spin_per_esa_step",
-                "energy_table",
```
greglucas, line 1426: Do you need to rename this to match Joey then, or is there something you should be doing to test/verify this is there in the validation tests?

greglucas (approved):

imap_processing/codice/codice_l2.py
lacoak21: They want us to keep "energy_table". Good call I will update the test.

lacoak21: And ask them to rename it their validation data.

laspsandoval (approved):

PR #2652, opened 2026-01-30
Fix up L1C setting of background rates and bg_sys_err

tmplummer:
# Change Summary

## Overview
- For now assume that backgrounds ancillary product must not change mid-pointing when identifying bg rows to apply.
- Modify csv definition so that bin_end is and inclusive index rather than exclusive.


imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -1079,6 +1080,12 @@ def set_background_rates(
         for esa_step in range(0, 7):
             value = row[f"E-Step{esa_step + 1}"]
             if row["rate/sigma"] == "rate":
+                logging.debug(
+                    "Setting background rates: bg_rates.shape=%s, bin_start=%d, bin_end=%d",
+                    bg_rates.shape,
+                    bin_start,
+                    bin_end,
+                )
```
tmplummer:
```suggestion
```

ahotasu:
I approve (can't figure out how to self-assign as a reviewer). However, I will note that the met end times in the ancillary background file aren't large enough to cover the testing we're doing for the data on 2026/01/27. I have tested this using a modified backgrounds file with an end met one (second?) higher than what the algorithm is looking for for 2026/01/27. This successfully loads the backgrounds and applies them in the pset and L2 CG corrected fluxes (per a PR I'll put in momentarily).

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -1060,17 +1060,18 @@ def set_background_rates(
         )
 
     # find to the rows for the current pointing
+    # TODO: This assumes that the backgrounds will never change mid-pointing.
+    #    Is that a safe assumption?
     pointing_bg_df = background_df[
-        (background_df["GoodTime_start"] >= pointing_start_met)
-        & (background_df["GoodTime_end"] <= pointing_end_met)
+        (background_df["GoodTime_start"] <= pointing_start_met)
+        & (background_df["GoodTime_end"] >= pointing_end_met)
```
greglucas, line 1067: Do you want to change this to GoodTime start here? My thought is that right now you're requiring the full pointing period to be covered, so if you choose just the leading edge the we can just say we will use anything straddling that time period. Alternatively swap them and say (goodtimestart < pointing end AND goodtimeend > pointing start) which should cover the joint cases a little more robustly I think?

tmplummer: I'm not sure I am following this. I think that my hesitation with the suggestion to do:
```
(background_df["GoodTime_start"] < pointing_end_met)
& (background_df["GoodTime_end"] > pointing_start_met)
```
would be that it would make it likely that multiple sets of backgrounds would get selected. For example if pointing start/end times are every 10: `[10, 20, 30, 40]` and you want different backgrounds for each. If you aren't super precise with your start/end times, the goodtimes start/ends might look like:
```
[[9, 21], [19, 31], [29, 41]]
```
But then the logic would match multiple goodtime entries to each pointing.

tmplummer:
@ahotasu 

> However, I will note that the met end times in the ancillary background file aren't large enough to cover the testing we're doing for the data on 2026/01/27. I have tested this using a modified backgrounds file with an end met one (second?) higher than what the algorithm is looking for for 2026/01/27. This successfully loads the backgrounds and applies them in the pset and L2 CG corrected fluxes (per a PR I'll put in momentarily).

The files in this PR are just for testing. In production, the ancillary files in the SDC are used.

imap_processing/lo/l1c/lo_l1c.py
greglucas: Yeah, I agree with you. This whole way of doing it is brittle IMO. We should be doing it per pointing number or something like that if it has to be constant per pset and not dealing with SHCOARSE at all.

greglucas (approved):

Issue #2651, opened 2026-01-30
BUG - Lo L1c backgrounds not being selected


greglucas:
### Description of the issue

I tried to set background manually by setting a large inclusive range to cover the entire mission. The problem is that we only select rows where the goodtime_start is greater than the pointing_start_met. So we will only keep goodtimes within that range, which means we could miss an overlapping range.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/b5d6ea3cb83161c473061bd47deb8083678dc927/imap_processing/lo/l1c/lo_l1c.py#L1062-L1066

We should update this to allow for times that start before/after the valid time period to make sure we get some coverage of our backgrounds.

### Code Snippet:

```python
Code
```

ahotasu:
Can the GoodTime entries overlap? If not (and that is enforced?), then changing the & to an | on line 1065 should work...

tmplummer:
This was fixed: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2652

PR #2653, opened 2026-01-31
Lo pointing set and map background rates update

ahotasu:
Worked with Nathan to implement addition of background to intensities per the latest update to the mapping document.

# Change Summary

## Overview
Added scalar logarithmic stability factor and code to use that factor for adding the background intensities to the ENA intensities for all energies. This, combined with Greg and Tim's work on (BUG - Lo L1c backgrounds not being selected #2651) was used to generate new L1c Pset and L2 CG Corrected Flux products that Nathan validated late last night.

## New Dependencies
None

## Updated Files
- lo_l2.py
   - added background factor, summation, and uncertainties/systematic errors to the start of the flux calculation function, which operates on the intensities after the bootstrap and sputtering corrections, but before the logarithmic calculations in the PowerLawFluxCorrector

## Testing
Not versed enough yet on pytest to run. :-(


imap_processing/lo/l2/lo_l2.py
```diff
@@ -1307,6 +1307,25 @@ def calculate_flux_corrections(dataset: xr.Dataset, flux_factors: Path) -> xr.Da
     """
     logger.info("Applying flux corrections")
 
+    bg_logarithmic_stability_factor = 0.04
+
+    # Add in the background intensity to ensure that logarithms behave
+    # properly in the flux corrector when intensities are zero or very low.
```
tmplummer, line 1313: This is really confusing to me. I thought that a critical idea in the Lo algorithm is that the backgrounds do not get subtracted. So coming into this function, `ena_intensity = signal_intensity + bg_intensity`. Then this factor of the background is added in on top which means `ena_intensity = signal_intensity + bg_intensity + stability_factor * bg_intensity`. That doesn't make sense to me. I know that this is what Nathan wants but I am really struggling to understand why. To me this seems like it is purely an problem with the crappy intput data being used for validation and this fix actually degrades the output product.

imap_processing/lo/l2/lo_l2.py
```diff
@@ -1307,6 +1307,25 @@ def calculate_flux_corrections(dataset: xr.Dataset, flux_factors: Path) -> xr.Da
...
+
+    # Commensurately, adjust the uncertainties to account for this addition
+    dataset["ena_intensity_stat_uncert"] = np.sqrt(
+        (dataset["ena_intensity_stat_uncert"]) ** 2
+        + (dataset["bg_intensity"] * bg_logarithmic_stability_factor) ** 2
+    )
+    dataset["bg_intensity_sys_err"] = np.sqrt(
+        (dataset["bg_intensity_sys_err"]) ** 2
+        + (dataset["bg_intensity"] * bg_logarithmic_stability_factor) ** 2
+    )
```
greglucas, line 1327: There is no systematic error here it looks like:
https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/21547460903/job/62093470806?pr=2653#step:7:3631

https://github.com/IMAP-Science-Operations-Center/imap_processing/actions/runs/21547460903/job/62093470806?pr=2653#step:7:3375

imap_processing/lo/l2/lo_l2.py
greglucas: I completely agree with this comment, and that is why I'm hesitant to throw this in there at the last second.

I think this is actually really bad because it adds in 4% uncertainty, but then doesn't remove that later on, so could be very misleading in terms of what is going on, and this is only before flux corrections, but if someone is looking at a sputtering/bootstrap map, then those wouldn't be corrected. It seems like this is a recipe for confusion.

greglucas:
I think we should take some time and think about this more (not just us, but the instrument team as well). Are there other ways to approach this, like for instance setting the flux values to the `min(flux, bg)` so that flux is never lower than bg. Or to set `bg` to the flux value if you are measuring less than background. i.e. I am wondering if we can somehow make the algorithm more robust rather than changing the data itself that goes into the algorithm.

ahotasu:
Witthdrawing pull request.

PR #2659, opened 2026-02-02
Fix GLOWS processing DE L1A

maxinelasp:
# Change Summary
GLOWS L1A processing has been failing because DE processing is kind of fragile. This refactors the way DE processing for L1A works to make it more resilient to weird packets, and adds a try/catch for a complicated piece to try and recover more aggressively.

## Overview

* Update DE packet combining code to handle out of order packets, missing start packets, and missing end packets more reliably
* Update validation code to point to a new file (generated from the latest version of GLOWS example processing code)
* Removes old validation file, new one is part of downloaded data and is already in AWS
* also adds some more logging and error management


imap_processing/glows/l0/glows_l0_data.py
```diff
@@ -160,6 +160,27 @@ def __post_init__(self) -> None:
                 int(self.DE_DATA, 2).to_bytes(len(self.DE_DATA) // 8, "big")
             )
 
+    # Sort by SEQ
+    def __lt__(self, other: "DirectEventL0") -> bool:
+        """
+        Define sorting algorithm for DirectEventL0.
```
tmplummer: I find this description a little misleading. This is defining the less than operator, right? Then, the builtin `sort` and `sorted` methods use the less than operator to do the sorting.

imap_processing/glows/l1a/glows_l1a.py
```diff
@@ -93,18 +96,58 @@ def process_de_l0(
...
-            # add it to the last direct event in the list
-            de_list[-1].merge_de_packets(de)
+        de_tracker[de.SEC] = (
+            [de] if de.SEC not in de_tracker.keys() else de_tracker[de.SEC] + [de]
+        )
+
+    total_packets_found = 0
+    missing_end_packets = 0
+    missing_middle_packets = 0
+    for sec, de_list in de_tracker.items():
```
tmplummer: It seems like some of this could be simplified a bit by using `itertools.groupby()`. e.g.:
```
for sec, de_group in groupby(de_l0, lambda: x, x.SEC):
```

tmplummer (approved):
This looks fine as is. Just a couple of minor suggestions to consider.

imap_processing/glows/l0/glows_l0_data.py
maxinelasp: Yes, good call, I will update

imap_processing/glows/l1a/glows_l1a.py
maxinelasp: good idea!

PR #2663, opened 2026-02-03
I-ALiRT - select last from time

laspsandoval:
# Change Summary

## Overview
Instead of interpolation uses factors from the time closest to but less than the time of the data.

## Updated Files
- Instead of interpolation uses factors from the time closest to but less than the time of the data.

### DIFF
--- a/imap_processing/ialirt/l0/process_swe.py
+++ b/imap_processing/ialirt/l0/process_swe.py
@@ -520,6 +520,11 @@ def process_swe(accumulated_data: xr.Dataset, in_flight_cal_files: list) -> list
 
         # Interpolate to find the correct calibration factor
         in_flight_cal_df = read_in_flight_cal_data(in_flight_cal_files)
+        in_flight_cal_df = (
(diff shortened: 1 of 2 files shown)


imap_processing/ialirt/l0/process_swe.py
```diff
@@ -520,6 +520,11 @@ def process_swe(accumulated_data: xr.Dataset, in_flight_cal_files: list) -> list
 
         # Interpolate to find the correct calibration factor
         in_flight_cal_df = read_in_flight_cal_data(in_flight_cal_files)
+        in_flight_cal_df = (
+            in_flight_cal_df.dropna(subset=["met_time"])
+            .sort_values("met_time")
+            .reset_index(drop=True)
+        )
```
tech3371: can you move these lines into `read_in_flight_cal_data()`? it seems like it would be useful for science too.

tech3371 (approved):
one minor request. otherwise, LGTM

PR #2664, opened 2026-02-03
IDEX Update event time calculation

lacoak21:
# Change Summary

## Overview
The IDEX team realized they were using shcoarse for event time calculation. This was actually the time of the packet calculation and not the time when IDEX triggered. 

I think @tech3371 you had this concern in the science team meeting a long time ago but I thought they said we were doing it right! Anyway this is the correct way to do it. 

## Updated Files

-imap_processing/idex/idex_l1a.py
   - Fix function to handle epoch calculation for science times using event trigger time variables instead of shcoarse and sfine. Continue to use shcoarse and sfine for housekeeping because they do not contain the other variables.

### DIFF
--- a/imap_processing/idex/idex_l1a.py
+++ b/imap_processing/idex/idex_l1a.py
@@ -356,7 +356,6 @@ class RawDustEvent:
         """
         # Calculate the impact time in seconds since epoch
         self.impact_time = 0
-
         # The elapsed seconds are stored as a 32-bit unsigned integer that is split
         # across two 16-bit words for packetization. As a result, idx__txhdrtimesec1
         # represents multiples of 2^16 seconds, while idx_txhdrtimesec2 represents the
--- a/imap_processing/idex/idex_l1a.py
+++ b/imap_processing/idex/idex_l1a.py
@@ -96,7 +96,7 @@ class PacketParser:
                     f"imap_idex_{level}_evt"
                 )
                 data["epoch"] = calculate_idex_event_time(
-                    data["shfine"].data, data["shcoarse"].data
+                    data["shcoarse"].data, data["shfine"].data
                 )
                 data["epoch"].attrs = epoch_attrs
                 self.data.append(data)
@@ -108,7 +108,7 @@ class PacketParser:
                     f"imap_idex_{level}_catlst"
                 )
                 data["epoch"] = calculate_idex_event_time(
-                    data["shfine"].data, data["shcoarse"].data
+                    data["shcoarse"].data, data["shfine"].data
                 )
                 data["epoch"].attrs = epoch_attrs
                 self.data.append(data)
--- a/imap_processing/idex/idex_l1a.py
+++ b/imap_processing/idex/idex_l1a.py
@@ -250,35 +250,22 @@ def _read_waveform_bits(waveform_raw: str, high_sample: bool = True) -> list[int
 
 def calculate_idex_event_time(
     fine_time_subs: np.ndarray,
-    coarse_time_sec1: np.ndarray,
-    coarse_time_sec2: np.ndarray | None = None,
+    coarse_time_sec: np.ndarray,
 ) -> npt.NDArray[np.int64]:
     """
     Calculate the epoch time from the FPGA header time variables.
 
-    For science packets, we are given the MET seconds, we need to convert it to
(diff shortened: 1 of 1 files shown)


imap_processing/idex/idex_l1a.py
```diff
@@ -248,35 +248,55 @@ def _read_waveform_bits(waveform_raw: str, high_sample: bool = True) -> list[int
...
     numpy.ndarray[numpy.int64]
         The mission elapsed time converted to nanoseconds since the J2000 epoch
         in the terrestrial time (TT) timescale.
     """
-    # Get met time in seconds including shfine (number of 20 microsecond ticks)
-    met = shcoarse_time + shfine_time * 20e-6
-    return met_to_ttj2000ns(met)
+    if coarse_time_sec2 is not None:
+        # Reconstruct the total seconds from the two 16-bit words
+        coarse_event_time = 65536 * coarse_time_sec1 + coarse_time_sec2
```
tech3371: I am little curious why they want to multiply coarse second time with 65536. I have seen this number is CoDICE and they are doing this:
```
    center_times_seconds = (
        acq_start_seconds + acq_start_subseconds / 65536 + (delta_times / 1e9)
    )
```
Can you check with them if 65536 should be applied to `fine_time_subs` instead?

tech3371 (commented):
I had one minor question to IDEX team but what you had is good.

p.s. yes, I remember that discussion and I thought I was way off based on response I got. I should have push back more :)

imap_processing/idex/idex_l1a.py
```diff
@@ -248,35 +248,55 @@ def _read_waveform_bits(waveform_raw: str, high_sample: bool = True) -> list[int
...
+    For housekeeping packets, use the shcoarse and shfine variables instead.
 
     Parameters
     ----------
-    shcoarse_time : float, numpy.ndarray
-        The coarse time value from the FPGA header. Number of seconds since epoch.
-    shfine_time : float, numpy.ndarray
-        The fine time value from the FPGA header. Number of 20 microsecond "ticks" since
-         the last second.
+    fine_time_subs : numpy.ndarray, optional
```
greglucas: This isn't optional anymore since it is the first argument, the final coarse_time_sec2 is optional now though.

imap_processing/idex/idex_l1a.py
```diff
@@ -357,8 +377,10 @@ def __init__(self, header_packet: space_packet_parser.SpacePacket) -> None:
         """
         # Calculate the impact time in seconds since epoch
         self.impact_time = 0
-        self.impact_time = calculate_idex_epoch_time(
-            header_packet["SHCOARSE"], header_packet["SHFINE"]
+        self.impact_time = calculate_idex_event_time(
+            header_packet["IDX__TXHDRTIMESUBS"],
+            header_packet["IDX__TXHDRTIMESEC1"],
+            header_packet["IDX__TXHDRTIMESEC2"],
```
greglucas: Add a comment here about what these times are, but I think this might be clearer what is going on (to me at least). You are left-shifting the first timer to the first 16 bits to make a 32-bit integer, so I'd use the bitwise operators (you might need to do this on `.data` of the packets I'm not sure)


```suggestion
        self.impact_time = calculate_idex_event_time(
            (header_packet["IDX__TXHDRTIMESEC1"] << 16) + header_packet["IDX__TXHDRTIMESEC2"],
            header_packet["IDX__TXHDRTIMESUBS"],
```

greglucas (commented):
Personally, I think this is slightly confusing with fine first then two coarse counters. Rather than changing the calculate event time, can you just update the one spot of input to combine the two coarse ints into a single int that you pass into the function, rather than reworking the function itself?

imap_processing/idex/idex_l1a.py
lacoak21: Screenshot 2026-02-09 at 11 38 46 AM
Here are more details. It should be the second coarse time because it is split across 2 16 bit ints.

tech3371: Yeah, this is interesting. Thank you for that detail!

tech3371 (commented):
looks good

greglucas (approved):

Issue #2383, opened 2025-11-05
BUG - inaccuracy in L2 map obs_date values after projection


tmplummer:
### Description of the issue

While working through a Hi bug related to obs_date values I found that given a single dataset, the projected obs_date values deviated up to ~60 seconds away from the expected value of the PSET midpoint as defined by the epoch + epoch_delta / 2.

I had a suspicion that the error was due to using tt2000ns for the time that was being projected. Just to check, I converted the obs_date value in the PSET prior to projecting to tt2000s and the error was reduced to about 20-seconds.

### Steps to reproduce the issue

The relevant test that exhibits this inaccuracy is in test_hi_l2.py::test_genarate_hi_map

The check on the projected obs_date value is:
```
np.testing.assert_allclose(
        np.nanmax(sky_map.data_1d["obs_date"].data) / 1e9,
        pset_midpoint,
        atol=60,
    )
```

### Expected behavior (What should happen)

_No response_

### Actual behavior (What does happen)

_No response_

### Code Snippet:

```python
Code
```

### Additional notes, affected areas, and suggested fixes

From Claude... suggestions for how to handle this:

>   1. Offset approach (recommended)                                                                                                                                                                                                                              
>   Subtract a reference timestamp before accumulating, then add it back after computing the mean:
>                                                                                                                                                                                                                                                                 
>   ## Pick a reference time (e.g., first PSET's midpoint or epoch start)                                                                                                                                                                                        
>   reference_time = int(pset_ds["epoch"].values[0])
> 
>   ## Store offset during processing (small values, safe for float64)
>   obs_date_offset = mid_time - reference_time  # Will be ~1e16 or less for 3-month maps
> 
>   ## After weighted mean calculation, add reference back
>   final_obs_date = obs_date_weighted_mean + reference_time
> 
>   The offset values for a 3-month map would be at most ~8e15 ns, well within float64's exact integer range (2^53 ≈ 9e15).
> 
>   2. Scale to coarser units during accumulation
>   Convert to microseconds or milliseconds before accumulating, then convert back:
> 
>   ## Accumulate in microseconds (divide by 1e3)
>   obs_date_us = mid_time // 1000  # int64, values ~1e15
> 
>   ## After mean, convert back
>   final_obs_date = obs_date_mean_us * 1000
> 
>   3. Incremental weighted mean
>   Compute the weighted mean incrementally rather than accumulating sums, avoiding large intermediate values entirely.
> 
>   ---
[...]

PR #2689, opened 2026-02-10
2561 hi l2   update how full spin helio frame maps are produced

### LINKED ISSUE #2561: Hi L2 - Update how full spin helio frame maps are produced
### Algorithm Description:

From Dan's email:

> 
> Hi All,
> 
> I agree with Sungjun’s summary below, and I’ll note that it applies to L2 as well.  I went over the Lo CG algorithm and realized that it doesn’t explicitly discuss the case of combining data from ram and antiram parts of the sky, because this is not something that the Lo team plans to do (they only make ram maps).  I should have caught this when reviewing for Hi, so my apologies for not catching that.  We can discuss this at the meeting, but I believe we will need to make sure that we first create CG-corrected ram and antiram maps separately and then combine them.  I can think of at least two reasons:
> A key step in the CG correction is finding the mean spacecraft frame (SF) energy <Es> to associate with the PSET bins that contribute to a heliospheric frame (HF) map pixel for a given HF energy Eh. This makes sense as long as the PSET bins are generally close in angle wrt the SC velocity vector (and therefore close in Es).  However, if combining PSET bins from the ram and antiram directions in the same pixel, then the values Es will be quite different. In particular, for the ram bins Es  > Eh and for the antiram bins, Es  < Eh.  Thus, using the mean value for these violates the assumption that  <Es> is close in energy to the individual Es values.
> As I argued for the L3 combinations, in general the ram observations have better stats for viewing a given part of the sky than the antiram observations, even if the exposure time is the same, just because we collect more counts.  Thus, w should combine these after building ram- and antiram-only maps using uncertainty weighting.
>  
> Therefore, the Hi HF map construction will need to be modified some.  Sorry Tim!!
>  
> Regards,
> Dan

Sungjun's summary

> The survival probability is interpolated to the central energies of the ESA energy steps.
To create the full-spin map (ram + anti-ram), we use an uncertainty-weighted average.
To combine maps from different instruments, we use an exposure-time-weighted average.
For sky-map bins where both ram and anti-ram data, as well as H45 and H90 data, are combined (e.g., HiC, full, and 1-year maps), we first combine the ram and anti-ram maps using an uncertainty-weighted average, and then combine the H45 and H90 maps using an exposure-time-weighted average.
For the uncertainty-weighted average, we use only the statistical uncertainty, not the systematic error.

tmplummer:
# Change Summary

## Overview
Sorry for how big this got...
This PR updates Hi L2 processing to properly handle full-spin heliocentric frame maps by generating ram and anti-ram maps separately, applying Compton-Getting corrections to each independently, and then combining them.

## Updated Files
   - Added keyword input bool to keep systematic error from getting updated if not desired. Hi does not want systematic uncertainty to go through the flux correction.
- imap_processing/hi/hi_l2.py
   - create_sky_map_from_psets() now returns dict[str, RectangularSkyMap] instead of a single map
      - For helio-frame (hf) full-spin maps: returns {"ram": ..., "anti": ...}
      - For all other maps: returns {spin_phase: ...} (single entry)
   - Added new combine_maps() function to merge ram and anti-ram maps after CG correction
   - Added cleanup_intermediate_variables() to remove temporary calculation variables (bg_rate, energy_sc, ena_signal_rates, ena_signal_rate_stat_unc) from final output

Closes: #2561

### DIFF
--- a/imap_processing/hi/hi_l2.py
+++ b/imap_processing/hi/hi_l2.py
@@ -220,8 +220,7 @@ def create_sky_map_from_psets(
         # Finish the exposure time weighted mean calculation of backgrounds
         # Allow divide by zero to fill set pixels with zero exposure time to NaN
         with np.errstate(divide="ignore"):
-            for var in vars_to_exposure_time_average:
-                map.data_1d[var] /= map.data_1d["exposure_factor"]
+            map.data_1d[vars_to_exposure_time_average] /= map.data_1d["exposure_factor"]
 
         # Add ESA energy data to the map dataset for use in rate/intensity calculations
         energy_delta = esa_ds["bandpass_fwhm"] / 2
--- a/imap_processing/hi/hi_l2.py
+++ b/imap_processing/hi/hi_l2.py
@@ -29,12 +29,11 @@ SC_FRAME_VARS_TO_PROJECT = {
     "counts",
     "exposure_factor",
     "bg_rate",
-    "bg_rate_sys_err",
     "obs_date",
 }
 HELIO_FRAME_VARS_TO_PROJECT = SC_FRAME_VARS_TO_PROJECT | {"energy_sc"}
 # TODO: is an exposure time weighted average for obs_date appropriate?
-FULL_EXPOSURE_TIME_AVERAGE_SET = {"bg_rate", "bg_rate_sys_err", "obs_date", "energy_sc"}
+FULL_EXPOSURE_TIME_AVERAGE_SET = {"bg_rate", "obs_date", "energy_sc"}
 
 
 # =============================================================================
@@ -90,21 +89,25 @@ def hi_l2(
         )
 
     logger.info(f"Step 1: Creating sky map from {len(psets)} pointing sets")
-    sky_map = create_sky_map_from_psets(
+    sky_maps = create_sky_map_from_psets(
         psets,
         l2_ancillary_path_dict,
         map_descriptor,
     )
 
     logger.info("Step 2: Calculating rates and intensities")
-    sky_map.data_1d = calculate_all_rates_and_intensities(
-        sky_map.data_1d,
-        l2_ancillary_path_dict,
-        map_descriptor,
-    )
+    for sky_map in sky_maps.values():
+        sky_map.data_1d = calculate_all_rates_and_intensities(
+            sky_map.data_1d,
+            l2_ancillary_path_dict,
+            map_descriptor,
+        )
 
-    logger.info("Step 3: Finalizing dataset with attributes")
-    l2_ds = sky_map.build_cdf_dataset(
+    logger.info("Step 3: Combining maps (if needed)")
+    final_map = combine_maps(sky_maps)
+
+    logger.info("Step 4: Finalizing dataset with attributes")
+    l2_ds = final_map.build_cdf_dataset(
         "hi",
         "l2",
         descriptor,
@@ -124,7 +127,7 @@ def create_sky_map_from_psets(
     psets: list[str | Path],
     l2_ancillary_path_dict: dict[str, Path],
     descriptor: MapDescriptor,
-) -> RectangularSkyMap:
+) -> dict[str, RectangularSkyMap]:
     """
     Project Hi PSET data into a sky map.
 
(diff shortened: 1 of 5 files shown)


imap_processing/hi/hi_l2.py
```diff
@@ -152,93 +196,207 @@ def generate_hi_map(
...
+    """
+    # Step 1: Rename some PSET vars to match L2 variables
+    pset_processed = pset.rename(HiPointingSet.l1c_to_l2_var_mapping)
+
+    # Step 2: Add obs_date variable to be used in determining a map mean obs_date
+    mid_time = (
+        pset_processed["epoch"].values[0] + pset_processed["epoch_delta"].values[0] / 2
     )
+    pset_processed["obs_date"] = xr.full_like(
+        pset_processed["exposure_factor"], float(mid_time)
```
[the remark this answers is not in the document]
tmplummer: I have a ticket for improving handling of the obs_date. #2383

imap_processing/hi/hi_l2.py
```diff
@@ -187,31 +207,32 @@ def create_sky_map_from_psets(
...
+                hi_pset, list(vars_to_bin), pset_valid_mask=directional_mask
+            )
 
-    return output_map
+    for map in rect_maps.values():
+        # Finish the exposure time weighted mean calculation of backgrounds
+        # Allow divide by zero to fill set pixels with zero exposure time to NaN
+        with np.errstate(divide="ignore"):
+            for var in vars_to_exposure_time_average:
+                map.data_1d[var] /= map.data_1d["exposure_factor"]
```
greglucas: Can you do this with xarray's list indexing syntax instead?


```suggestion
            map.data_1d[vars_to_exposure_time_average] /= map.data_1d["exposure_factor"]
```

imap_processing/hi/hi_l2.py
```diff
@@ -141,18 +144,35 @@ def create_sky_map_from_psets(
...
+        matching the descriptor's spin_phase.
     """
     if len(psets) == 0:
         raise ValueError("No PSETs provided for map creation")
 
-    output_map = descriptor.to_empty_map()
-
-    if not isinstance(output_map, RectangularSkyMap):
+    # If we are making a full-spin, helio-frame map, we need to make one ram
+    # and one anti-ram map that get combined at the final L2 step.
```
greglucas, line 159: This reminds me of Lo needing to make an extra "Oxygen" map for sputtering calculations and needing to carry it through to some extent as well. I'm wondering if this dictionary-like mapping should be carried over to the Lo calculations as well to keep things consistent or if there is another way of organizing this multi-map projections work across all instruments.

Just noting this as a comment, nothing to do here.

imap_processing/hi/hi_l2.py
```diff
@@ -525,6 +550,116 @@ def combine_calibration_products(
...
+    if "ram" not in sky_maps or "anti" not in sky_maps:
+        raise ValueError(
+            f"Expected sky maps with 'ram' and 'anti' keys."
+            f"Instead got: {sky_maps.keys()}"
+        )
+
+    logger.info("Combining ram and anti-ram maps using inverse-variance weighting.")
+
+    ram_ds = sky_maps["ram"].data_1d
+    anti_ds = sky_maps["anti"].data_1d
```
greglucas, line 586: I'm not sure if this would be useful to have more general, but I think a lot of the math below could be generalized to something where you are working with a list of datasets input (not just ram/anti, but maybe combining multiple 7-day maps or something like that in the future...)

```python
datasets = [ds.data_1d for ds in sky_maps.values()]

weights = [xr.where(ds["ena_intensity_stat_uncert"] > 0, 1 / ds["ena_intensity_stat_uncert"] ** 2, 0) for ds in datasets]
...
```

Again, not needed here and I'm just writing it down for a quick note. It should be another ticket if this would even be useful.

greglucas (approved):

imap_processing/hi/hi_l2.py
tmplummer: Yeah, good point. I was toying with the idea of generating a dataset that had yet another dimension added for "spin_phase" which could later be combined. I think that some of these things would come easier if we did a refactor of the mapping code to leverage the xarray custom accessor functions.

imap_processing/hi/hi_l2.py
tmplummer: Oh, cool... I didn't know you could do that. I'll test it out.

PR #2699, opened 2026-02-13
Updates to pass pipeline settings thorugh GLOWS L2

maxinelasp:
# Change Summary
Small update which will support other L2 GLOWS  - passing ancillary files through to the processing. 

Similarly to the L1B changes this will have a corresponding change in SDS data manager. 

## Overview

Wires up ancillary file through L2 processing. 

Note: This PR was written primarily by Claude.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -742,11 +742,27 @@ class Glows(ProcessInstrument):
             science_files = dependencies.get_file_paths(source="glows")
             if len(science_files) != 1:
                 raise ValueError(
-                    f"GLOWS L1A requires exactly one input science file, received: "
-                    f"{science_files}."
+                    f"GLOWS L2 requires exactly one input science file, "
+                    f"received: {science_files}."
                 )
             input_dataset = load_cdf(science_files[0])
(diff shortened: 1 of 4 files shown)


imap_processing/cli.py
```diff
@@ -742,11 +742,27 @@ def do_processing(
...
+                    f"received: {science_files}."
                 )
             input_dataset = load_cdf(science_files[0])
-            datasets = glows_l2(input_dataset)
+
+            # Load pipeline settings for L2 processing
+            current_day = np.datetime64(
+                f"{self.start_date[:4]}-{self.start_date[4:6]}-{self.start_date[6:]}"
+            )
+            day_buffer = current_day + np.timedelta64(3, "D")
```
tech3371, line 754: so you ancillary files always has future days data?

tech3371 (approved):
looks good

imap_processing/cli.py
maxinelasp: Yes, this is to handle the difference between repoints and start date, to make sure I always have enough data available.

PR #2700, opened 2026-02-13
Removing templates so they will come from project level

maxinelasp:
Right now, all these templates are already set in the Organzation level, but they are overwritten here. This removal ensures the templates will all come from one place for all repos.


laspsandoval (approved):
Nice.

PR #2705, opened 2026-02-15
BUG - Lo L1a DE dimension mismatch

### LINKED ISSUE #2650: BUG - Lo L1a DE dimension mismatch
### Description of the issue

In the l1a-de file for Lo there is a mismatch between epoch and shcoarse that shouldn't be there. This causes the l1b-de job to fail because it calls `np.repeat(l1a["shcoarse"], l1a["num_events"])`, but `num_events` is of length epoch (204) and shcoarse is of length 612 (repeated 3 times?).

We likely need to go back and reprocess l1a data for this date and see if it produces the bad dates still, or if this is an artifact of previous processing dates before we did the large refactor of l1a event handling.

```
Dimensions:              (epoch: 204, direct_events: 535741, shcoarse: 612)
Coordinates:
  * shcoarse             (shcoarse) uint32 2kB 499860578 499860578 ... 499946476
  * epoch                (epoch) int64 2kB 815436644730736000 ... 81552254265...
```

```
imap_cli --instrument lo --data-level l1b --descriptor de --start-date 20251103 --version v001 --dependency imap_lo_l1b_de-58101a0f_20251103-repoint00037_v001.json  --repointing repoint00037
```

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

ahotasu:
Fixes #2650

# Change Summary
Change L1b DE references to use L1a DE "met" field instead of "shcoarse". The former is the same dimension as the "epoch" field, whereas the latter can be longer when segmented packets are present.

## Testing

Manual testing consisting of generation of DE products from 20251215 and 20260212, days which had previously failed and passed, respectfully, prior to this change.

### DIFF
--- a/imap_processing/tests/lo/test_lo_l1b.py
+++ b/imap_processing/tests/lo/test_lo_l1b.py
@@ -516,12 +516,9 @@ def test_get_spin_start_times(mock_interpolate_spin_data):
     )
     mock_interpolate_spin_data.return_value = mock_spin_df
 
-    # shcoarse is longer than met, simulating when segmented packets have
-    # occurred
     l1a_de = xr.Dataset(
         {
             "met": ("epoch", [15, 35]),
-            "shcoarse": ("shcoarse", [15, 16, 35]),
             "de_count": ("epoch", [2, 3]),
             "de_time": ("direct_event", [0, 1000, 2000, 3000, 4000]),
         },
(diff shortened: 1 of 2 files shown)


sdhoyt (approved):
Looks good. Just need to update the unit test input structures to have the met variable

imap_processing/tests/lo/test_lo_l1b.py
```diff
@@ -516,17 +516,20 @@ def test_get_spin_start_times(mock_interpolate_spin_data):
     )
     mock_interpolate_spin_data.return_value = mock_spin_df
 
+    # shcoarse is longer than met, simulating when segmented packets have
+    # occurred
     l1a_de = xr.Dataset(
         {
-            "shcoarse": ("epoch", [15, 35]),
+            "met": ("epoch", [15, 35]),
+            "shcoarse": ("shcoarse", [15, 16, 35]),
```
tech3371: do you need both or should we remove shcoarse?

tech3371 (approved):
looks good beside one minor comment

imap_processing/tests/lo/test_lo_l1b.py
ahotasu: No, both are not needed. Removing shcoarse in both tests.

PR #2712, opened 2026-02-17
Populate map principal data CATDESC from descriptor

jtniehof:
# Change Summary

## Overview

The CATDESC (and thus plot title in CAVA) for all ENA maps is "Mono-energetic ENA intensity". This doesn't provide much detail on what intensity is being shown, and the IMWG has requested that this be updated with a human-readable description of the map descriptor.

## File changes

- `ena_maps/utils/naming.py`: add function to parse a CATDESC from a MapDescriptor
- `tests/ena_maps/test_naming.py`: tests for above
- `ena_maps/ena_maps.py`: patch the CATDESC in `build_cdf_dataset`
- `tests/ena_maps/test_ena_maps.py`: update tests for above

## Testing

New test was added for `MapDescriptor.to_catdesc`; the test for `build_cdf_dataset` was updated to check the CATDESC was properly populated. `build_cdf_dataset` now requires a fairly well-constructed descriptor, so some tests were updated to provide that.

# Other notes

This PR is in draft while the IMWG reviews the actual contents of the CATDESC; there shouldn't be structural changes after that.

This does create a circular import between `ena_maps` and `ena_maps.utils.naming`. It doesn't seem to cause problems but I still don't like it.

There are several cases where a malformed map descriptor will cause a regex to fail and horrible exceptions to be thrown. I figure the implicit contract that deep in the code is that the descriptor is well-formed, and I usually avoid catching an exception just to throw another exception. Open to alternate approaches.

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -1422,20 +1422,10 @@ class RectangularSkyMap(AbstractSkyMap):
         )
 
         # And CATDESC for principal data
-        possible_principal_data = (
-            "dust_rate",
-            "ena_intensity",
-            "ena_spectral_index",
-            "glows_rate",
-            "isn_rate",
-            "isn_rate_bg_subtracted",
-        )
-        for principal_data in possible_principal_data:
-            if principal_data in cdf_ds:
-                catdesc = naming.MapDescriptor.from_string(descriptor).to_catdesc()
-                cdf_ds[principal_data].attrs["CATDESC"] = catdesc
-                # Only one principal variable
-                break
+        md = naming.MapDescriptor.from_string(descriptor)
+        principal_data = md.principal_data_var
+        if principal_data in cdf_ds:
+            cdf_ds[principal_data].attrs["CATDESC"] = md.to_catdesc()
 
         return cdf_ds
 
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -16,7 +16,7 @@ from numpy.typing import NDArray
 
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.cdf.utils import load_cdf
-from imap_processing.ena_maps.utils import map_utils, spatial_utils
+from imap_processing.ena_maps.utils import map_utils, naming, spatial_utils
 
 # The coordinate names can vary between L1C and L2 data (e.g. azimuth vs longitude),
 # so we define an enum to handle the coordinate names.
@@ -1420,6 +1420,22 @@ class RectangularSkyMap(AbstractSkyMap):
         cdf_ds["epoch"].attrs.update(
             {"DELTA_PLUS_VAR": "epoch_delta", "BIN_LOCATION": 0}
         )
+
+        # And CATDESC for principal data
+        possible_principal_data = (
+            "dust_rate",
+            "ena_intensity",
+            "ena_spectral_index",
+            "glows_rate",
+            "isn_rate",
+            "isn_rate_bg_subtracted",
+        )
+        for principal_data in possible_principal_data:
+            if principal_data in cdf_ds:
+                catdesc = naming.MapDescriptor.from_string(descriptor).to_catdesc()
+                cdf_ds[principal_data].attrs["CATDESC"] = catdesc
+                # Only one principal variable
+                break
 
         return cdf_ds
 
--- a/imap_processing/ena_maps/utils/naming.py
+++ b/imap_processing/ena_maps/utils/naming.py
@@ -199,7 +199,10 @@ class MapDescriptor:
                 if self.instrument_descriptor.startswith(desc)
             )
         )
-        sensor = "Comb" if self.sensor == "combined" else self.sensor
+        sensor = " Combined" if self.sensor == "combined" else self.sensor
+        species = self.species.title()
+        if species == "Uv":
+            species = "UV"
         m = re.match(
             r"^(drt|ena|int|isn|spx)(?:(?<=spx)\d+)?([^-_\s]*)$", self.principal_data
         )
@@ -207,13 +210,12 @@ class MapDescriptor:
             "drt": "Rate",
             "ena": "Inten",
             "int": "Inten",
-            "isn": "ISN Rate",
+            "isn": "Rate",
             "spx": "Spectral",
         }[m.group(1)]
+        if m.group(1) == "isn":
+            species = "ISN " + species
         extras = m.group(2)
-        species = self.species.title()
-        if species == "Uv":
-            species = "UV"
         coord = self.coordinate_system.upper()
         frame = {
             "hf": "Helio",
@@ -234,7 +236,7 @@ class MapDescriptor:
             if duration.endswith("Mo"):
                 duration += "n"
         catdesc = (
-            f"IMAP {instrument}{sensor} {quantity} {species}, {coord} "
+            f"IMAP {instrument}{sensor} {species} {quantity}, {coord} "
             f"{frame} Frame, {survival}, {spin_phase}, {resolution}, {duration}"
         )
         possible_extras = [
--- a/imap_processing/ena_maps/utils/naming.py
+++ b/imap_processing/ena_maps/utils/naming.py
@@ -173,6 +173,80 @@ class MapDescriptor:
             ]
         )
 
+    def to_catdesc(self) -> str:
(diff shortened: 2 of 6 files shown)


imap_processing/ena_maps/utils/naming.py
```diff
@@ -173,6 +173,82 @@ def to_string(self) -> str:
...
+        """
+        instrument_names = {
+            "l": "Lo",
+            "t": "Lo",
+            "ilo": "Lo",
+            "h": "Hi",
+            "u": "Ultra",
+            "idx": "IDEX",
+            "glx": "GLOWS",
+        }
```
tmplummer: There is some overlap here with the MappableInstrumentShortName enum class. I wonder of we could leverage that somehow. The two issues that would need to be sorted are:
- trimming the full Lo names
- convert Lo, Hi, Ultra to CamelCase but keep IDEX, GLOWS as is.

https://github.com/jtniehof/imap_processing/blob/0f6de5e32ad425b7c4fab46ccc86918bd578cc52/imap_processing/ena_maps/utils/naming.py#L18

One idea would be to modify the class such that the `value` attribute is a named tuple so that you could do something like:
```
instrument_long = self.instrument.long_name
instrument_short = self.instrument.short_name
```

imap_processing/ena_maps/utils/naming.py
```diff
@@ -173,6 +173,82 @@ def to_string(self) -> str:
...
+        instrument = next(
+            (
+                name
+                for desc, name in instrument_names.items()
+                if self.instrument_descriptor.startswith(desc)
+            )
+        )
```
tmplummer: I think this could be reversed to use the mapping and `instrument` attribute:
```suggestion
        instrument = instrument_names[self.instrument.value]
```

imap_processing/ena_maps/utils/naming.py
```diff
@@ -173,6 +173,82 @@ def to_string(self) -> str:
...
+        sensor = " Combined" if self.sensor == "combined" else self.sensor
+        species = self.species.title()
+        if species == "Uv":
+            species = "UV"
```
subagonsouth:
```suggestion
        species = self.species.upper()
```

tmplummer: Minor nit... a ternary operator should be used here.
```suggestion
        species = "UV" if self.species == "uv" else self.species.title()
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -1421,6 +1421,22 @@ def build_cdf_dataset(  # noqa: PLR0912
...
+            "glows_rate",
+            "isn_rate",
+            "isn_rate_bg_subtracted",
+        )
+        for principal_data in possible_principal_data:
+            if principal_data in cdf_ds:
+                catdesc = naming.MapDescriptor.from_string(descriptor).to_catdesc()
+                cdf_ds[principal_data].attrs["CATDESC"] = catdesc
+                # Only one principal variable
+                break
```
tmplummer: It seems like it would be cleaner to return a tuple form `to_catdesc()` consisting of the variable name and the catdesc or add a `principal_data()` property method so that you could just do:
```
principal_var, catdesc = naming.MapDescriptor.from_string(descriptor).to_catdesc()
cdf_ds[principal_var].attrs["CATDESC"] = catdesc
```
or
```
map_desc = naming.MapDescriptor.from_string(descriptor)
cdf_ds[map_desc.principal_data].attrs["CATDESC"] = map_desc.to_catdesc()
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -1421,6 +1421,22 @@ def build_cdf_dataset(  # noqa: PLR0912
             {"DELTA_PLUS_VAR": "epoch_delta", "BIN_LOCATION": 0}
         )
 
+        # And CATDESC for principal data
```
tmplummer, line 1424: Please note that Ultra does not use this `build_cdf_dataset()` method currently. I think there is an intent to move this method to the generic SkyMap class and transition Ultra to using that, but you will need to dig through the Ultra code to see where you can fit this in for updating the Ultra CATDESC.

imap_processing/tests/ena_maps/test_naming.py
```diff
@@ -304,3 +304,75 @@ def test_to_string(self):
...
+            ),
+            (
+                "h45-spx0305-h-hf-sp-ram-hae-4deg-3mo",
+                "IMAP Hi45 H Spectral, HAE Helio Frame, Surv Corr, Ram, 4 deg, 3 Mon",
+            ),
+            (
+                "hic-ena-h-hf-sp-ram-hae-4deg-3mo",
+                "IMAP Hi Combined H Inten, HAE Helio Frame, Surv Corr, Ram,"
+                " 4 deg, 3 Mon",
+            ),
```
tmplummer, line 323: This would be better suited as a parameterized test. For example:
```
@pytest.mark.parametrize(
    "map_desc, expected_catdesc",
    [
        (
            "h45-spx-h-hf-sp-ram-hae-4deg-3mo",
            "IMAP Hi45 H Spectral, HAE Helio Frame, Surv Corr, Ram, 4 deg, 3 Mon",
        ),
        (
            "h45-spx0305-h-hf-sp-ram-hae-4deg-3mo",
            "IMAP Hi45 H Spectral, HAE Helio Frame, Surv Corr, Ram, 4 deg, 3 Mon",
        ),
        (
            "hic-ena-h-hf-sp-ram-hae-4deg-3mo",
            "IMAP Hi Combined H Inten, HAE Helio Frame, Surv Corr, Ram,"
            " 4 deg, 3 Mon",
        ),
    ],
)
def test_to_catdesc(self, map_desc, expected_catdesc):
    md = MapDescriptor.from_string(descriptor_str)
    actual_catdesc = md.to_catdesc()
    assert actual_catdesc == expected_catdesc
```

tmplummer (commented):
Saw this draft PR and thought I would give some feedback.

imap_processing/ena_maps/utils/naming.py
jtniehof: I looked a bit at that class and was hesistant to do much changing because the CATDESC names are fairly specific to this task, and aren't new generic names. Your comment here and below encouraged me to take another stab at it and I think the results, while still hacky, are at least shorter and easier to understand.

imap_processing/ena_maps/utils/naming.py
jtniehof: Wrapping this up in using `self.instrument` instead of the descriptor made this whole mess go away.

imap_processing/ena_maps/utils/naming.py
jtniehof: I am _deeply_ ashamed that I missed a chance to use my favorite operator.

imap_processing/ena_maps/ena_maps.py
jtniehof: Yes, a property on MapDescriptor is a _lot_ better...I was already nervous about the fact I was copy-pasting this into the L3 code, and the Ultra situation made it even worse. See if the new approach makes sense.

imap_processing/ena_maps/ena_maps.py
jtniehof: Thanks for that catch! I was feeling a little nervous since it wasn't dealing with HEALPix.

I wound up making a complete copy of the main ultra_l2 test because it works quite differently between using the `output_map_structure` vs. the descriptor--the descriptor can't specify ECLIPJ2000 and it also casts the pixel size to a float somewhere along the way. That said, now I see there was another test using the descriptor so going to go move into that test now....done. I can rebase out that misstep if preferred.

imap_processing/tests/ena_maps/test_naming.py
jtniehof: Spot the guy who uses base unittest :) Thanks.

jtniehof:
> Saw this draft PR and thought I would give some feedback.

Thanks for many good catches, @subagonsouth. I think I've hit them all, but I left the conversations unresolved so you can check.

I have to update the L3 code and then I'm going to make one last check of how these display in CAVA before marking ready.

tmplummer (commented):
Great! Your changes cleaned this up nicely.

jtniehof:
Thanks! Jamey agrees that any concerns with display in CAVA are issues for CAVA, and we should proceed with this. So marking ready.

tmplummer (approved):
Thanks for implementing this!

PR #2714, opened 2026-02-17
Ultra l1b extended spin voltage cull

### LINKED ISSUE #2709: low voltage cull

lacoak21:
# Change Summary

## Overview

Add low voltage flags to l1b goodtimes / extendedspin. Use this in l1c psets to reject events at spins dependent on energy. 

This makes the current goodtimes configuration a little tricky because the code was only set up to really cull spins across all energies. To remedy this, I just added energy dependent flags to the goodtimes dataset without culling full spins. Then at l1c, we can toss out events at spins that are flagged if the flag is for the energy of the event. 

The low voltage cull is only dependent on spins and not energy but I added the frame work for culling dependent on energy bins for future algorithms. 

Bob specifically wants us to be able to cull bins of spins and energy configurable with constants. 

closes #2709

## FIle changes 

See copilot below 

## 
Tests for new functions including validation test that uses bobs code

### DIFF
--- a/imap_processing/ultra/l1b/ultra_l1b_culling.py
+++ b/imap_processing/ultra/l1b/ultra_l1b_culling.py
@@ -21,7 +21,10 @@ from imap_processing.ultra.l1b.lookup_utils import (
     get_scattering_coefficients,
     get_scattering_thresholds,
 )
-from imap_processing.ultra.l1b.quality_flag_filters import DE_QUALITY_FLAG_FILTERS
+from imap_processing.ultra.l1b.quality_flag_filters import (
+    DE_QUALITY_FLAG_FILTERS,
+    ENERGY_DEPENDENT_SPIN_QUALITY_FLAG_FILTERS,
+)
 from imap_processing.ultra.l1b.ultra_l1b_extended import get_spin_info
 from imap_processing.ultra.l1c.l1c_lookup_utils import build_energy_bins
 
@@ -549,10 +552,13 @@ def get_energy_and_spin_dependent_rejection_mask(
     """
     # Get the ebin flags for each energy bin from the goodtimes dataset.
     energy_range_edges = goodtimes_dataset["energy_range_edges"].values
+    # Get the quality flag arrays "turned on" for energy dependent culling from the
+    # goodtimes dataset.
+    flag_arrays = [
+        goodtimes_dataset[flag_name].values
+        for flag_name in ENERGY_DEPENDENT_SPIN_QUALITY_FLAG_FILTERS
+    ]
     ebin_flags = goodtimes_dataset["energy_bin_flags"].values
-    low_voltage = goodtimes_dataset["quality_low_voltage"].values
-    high_energy = goodtimes_dataset["quality_high_energy"].values
-    statistics = goodtimes_dataset["quality_statistics"].values
     # Create a dict of spin_number to index in the goodtimes dataset
     spin_to_idx = {
         spin: idx for idx, spin in enumerate(goodtimes_dataset["spin_number"].values)
@@ -570,10 +576,11 @@ def get_energy_and_spin_dependent_rejection_mask(
         # If the flag is set for any of the quality arrays, then reject
         # the event.
         flagged_at_spins = (
-            (low_voltage[goodtimes_inds] & energy_bin_flag)
-            | (high_energy[goodtimes_inds] & energy_bin_flag)
-            | (statistics[goodtimes_inds] & energy_bin_flag)
-        ) > 0
+            np.bitwise_or.reduce(
+                [qf[goodtimes_inds] & energy_bin_flag for qf in flag_arrays]
+            )
+            > 0
+        )
 
         # Mark flagged events as rejected
         mask_indices = np.where(mask)[0]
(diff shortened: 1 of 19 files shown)


imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -555,3 +616,146 @@ def count_rejected_events_per_spin(
...
+        return quality_flags
+
+    # Append the last start time plus the spin period to account for low voltage times
+    # that occur after the last spin start time
+    spin_start_times = np.append(
+        spin_start_times, spin_start_times[-1] + spin_periods[-1]
+    )
+
+    low_voltage_times = status_dataset["epoch"].data[low_voltage_inds]
+    # For each low voltage time, find the corresponding spin time
```
demajistre, line 665: I see here that the time bin that you are using is the 'epoch' -  I'd really recommend that you use the 'shcourse' (i.e., met) for time.  The spacecraft clock is used for both the spin definition and the status packet.  In this case it is more important to coordinate with 'what time the spacecraft thinks it is' instead of 'what time it actually is'.  The 'epoch' field depends on the clock kernel, which can change over the mission.

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -555,3 +616,146 @@ def count_rejected_events_per_spin(
...
+        # (spins at 0-4)
+        start_idx = (ind // spin_bin_size) * spin_bin_size
+        end_idx = start_idx + spin_bin_size
+        # Handle edge case where the last bin may not be full
+        # e.g. there are 12 spins, and spin_bin_size is 5, if ind is 10, then start_idx
+        # is 10 and end_idx is 15, but we only have spins up to index 11, so we need
+        # to cap the end_idx at 12
+        end_idx = min(end_idx, len(spins))
+        flag_inds = np.arange(start_idx, end_idx)
+        # Flag all the spins in the bin with all energy flags (voltage culling is
```
demajistre: I think that the way you are using the 'spin bins' here may not be correct.  The purpose of grouping the spins together is to coordinate the voltage, sep and statistical cull - so they can be done in a coordinated fashion. That is, all three culls should share the same spin groups  - and these groups should probably be fed in to this function.

demajistre (commented):
I'm hoping that you see the comments that I left referencing the code

imap_processing/ultra/l1b/ultra_l1b_culling.py
lacoak21: Good catch.

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -523,6 +525,63 @@ def get_de_rejection_mask(
...
+        goodtimes_inds = [spin_to_idx[spin] for spin in spin_number[mask]]
+        # Get the flag value for the current energy bin
+        energy_bin_flag = ebin_flags[i]
+        # If the flag is set for any of the quality arrays, then reject
+        # the event.
+        flagged_at_spins = (
+            (low_voltage[goodtimes_inds] & energy_bin_flag)
+            | (high_energy[goodtimes_inds] & energy_bin_flag)
+            | (statistics[goodtimes_inds] & energy_bin_flag)
+        ) > 0
```
tech3371: what is reason for `>0`?

imap_processing/ultra/l1b/ultra_l1b_culling.py
tmplummer: I think that all of this could be made a bit more efficient by filtering the dataframe first then extracting the values.
```
spin_df = spin_df[spin_df.spin_number.isin(spins)]
spin_period = spin_df["spin_period_sec"].values
spin_starttime = spin_df["spin_start_met"].values
spin_phase_valid = spin_df["spin_phase_valid"].values
spin_period_valid = spin_df["spin_period_valid"].values
```

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -555,3 +621,192 @@ def count_rejected_events_per_spin(
...
+    spin_tbin_edges = spin_start_times[::spin_bin_size][:n_spin_bins]
+    if spin_tbin_edges.size == 0:
+        # If there are no valid spin bins, return an array with a single edge at 0
+        raise ValueError(
+            f"No valid spin bins found for bin size: {spin_bin_size}"
+            f" and number of spins: {len(spins)}."
+        )
+    # Append the last start time plus the spin period to account for low times
+    # that occur after the last spin start time
+    spin_tbin_edges = np.append(spin_tbin_edges, spin_tbin_edges[-1] + spin_periods[-1])
```
tmplummer, line 773: I'm not sure if this matters here, but on Hi, I had an issue where there were small gaps or overlap of spins because I was using `spin_start_met + spin_period` to define a spin. That is never identically equal to the next spin_start_met. So, here, I think you could have a small gap in your final edge and what would be the next starting edge.

tmplummer (approved):
Overall, this looks good. There is lots going on here and I am missing some high level context but I don't see anything glaring.

lacoak21:
> Overall, this looks good. There is lots going on here and I am missing some high level context but I don't see anything glaring.

Yeah sorry Tim this grew out of hand. Thanks for reviewing.

imap_processing/ultra/l1b/ultra_l1b_culling.py
lacoak21: Hm I think its fine here because its only the very last bin.

Issue #2367, opened 2025-11-03
BUG - GLOWS L2 averaging calculations include bad times


maxinelasp:
## Description

Update L2 calculations to use only the good times calculated from L1B for averaging and standard devations.

## Location

`imap_processing/glows/l2/glows_l2.py:95-104`

## Current Implementation

```python
for field in all_variables:
    var_name = field.name
    if "average" in var_name:
        var_outputs[var_name] = (
            l1b_dataset[var_name].mean(dim="epoch", keepdims=True).data
        )
        var_outputs[var_name.replace("average", "std_dev")] = (
            l1b_dataset[var_name].std(dim="epoch", keepdims=True).data
        )
```

## Required Changes

- Change `l1b_dataset` to `good_data` in both the `.mean()` and `.std()` calls
- Verify that all average and std_dev calculations use only good times

## Related

Part of #1276 (GLOWS L2 follow up)

PR #2726, opened 2026-02-18
Add tests, rearrange GLOWS L2, fix exposure times

maxinelasp:
# Change Summary

## Overview

Moves GLOWS L2 Histogram calculation into the dataclass to match other levels, fixes the L2 exposure time calculation, and adds additional unit tests.

A lot of these changes are just for moving calculations from glows_l2.py to glows_l2_data.py. Look at the specific updates in the file changes section for actual changes. The main ones are around exposure_times.

## File changes

- `glows_l2_data.py`: Fixes exposure time calculation in `DailyLightcurve`
  — divides by number of bins (not hardcoded 3600) and accumulates uniformly  across good-time epochs rather than per-bin. Removes the now-incorrect  `calculate_exposure_times` static method. Fixes an in-place mutation bug in   `calculate_histogram_sums`. Completes `HistogramL2.__init__` with all L2  averages and std devs computed over good-time blocks only, with correct  shapes for scalar (`keepdims=True`) and vector (`[np.newaxis, :]`) fields.  Converts `return_good_times` to a `@staticmethod`.

- `glows_l2.py`: Removes logic that was moved into `HistogramL2.__init__`.

- `test_glows_l2_data.py`: New test file covering `DailyLightcurve` (photon  flux, flux uncertainty, zero-exposure bins, number of bins) and  `HistogramL2.return_good_times`.

- `test_glows_l2.py`: Updates `test_generate_l2` assertions to match  corrected output values; removes `test_exposure_times` which tested the  removed `calculate_exposure_times` method.

## Testing

- 6 new unit tests added in `test_glows_l2_data.py`
- Existing integration tests in `test_glows_l2.py` updated to reflect  corrected calculation behavior

Tests were generated with Claude but checked by hand. The averaging steps in __init__ were also done with AI. Nothing else was completed with AI, although it was checked against the validation scripts with Claude.

Completes tickets: #2367

Also fixes an un-ticketed bug with exposure time calculation

### DIFF
--- a/imap_processing/glows/l2/glows_l2_data.py
+++ b/imap_processing/glows/l2/glows_l2_data.py
@@ -6,6 +6,9 @@ import numpy as np
 import xarray as xr
 from numpy.typing import NDArray
 
+from imap_processing.glows import FLAG_LENGTH
+from imap_processing.glows.l1b.glows_l1b_data import PipelineSettings
+
 
 @dataclass
 class DailyLightcurve:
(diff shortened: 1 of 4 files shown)


imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -223,3 +209,173 @@ class HistogramL2:
...
+        numpy.ndarray
+            An array of indices for good times.
+        """
+        if len(active_flags) != flags.shape[1]:
+            print("Active flags don't matched expected length")
+
+        # A good time is where all the active flags are equal to one.
+        # Here, we mask the active indices using active_flags, and then return the times
+        # where all the active indices == 1.
+        good_times = np.where(np.all(flags[:, active_flags == 1] == 1, axis=1))[0]
```
tech3371, line 380: I don't know enough details here but feel like it could be improved. But I don't have suggestions yet.

tech3371 (approved):
looks good

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: Improved how? It should be pretty fast since it's all vectorized numpy, is the comment unclear?

PR #2727, opened 2026-02-18
ULTRA constants update

lacoak21:
# Change Summary

## Overview
Update an ULTRA constant "N_RE" which refers to the number of earth radii to use as the culling boundary.


tech3371 (approved):
btw, you may have set your PR to draft by default now.

PR #2731, opened 2026-02-19
GLOWS L2 - Updated Delta +/- uncertainties attribute for photon flux

### LINKED ISSUE #2315: Add GLOWS DELTA_* uncertainty pointers to metadata GLOWS L2
### Description of the issue

The GLOWS L2 data files are missing `DELTA_PLUS_VAR` and `DELTA_MINUS_VAR` pointers to the statistical uncertainty values in the file. Update the metadata configuration to add these attributes.

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/0d7eea6e4f2bcd0909510c53043a362192d9f791/imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml#L399

Should have both of those attributes that point to this variable name:
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/0d7eea6e4f2bcd0909510c53043a362192d9f791/imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml#L439

sdhoyt:
# Change Summary

Added the DELTA_PLUS_VAR, DELTA_MINUS_VAR attributes to the yaml. The flux uncertainties are symmetric so the same value can be used for both the plus and the minus

Ran CLI to test CDF creation

Closes #2315

### DIFF
--- a/imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
@@ -407,6 +407,9 @@ photon_flux:
   FORMAT: F8.2
   VALIDMIN: 0.0
   VALIDMAX: 30000.0 # max 30000 cps in FDIR / from 1 to 3.37 cps-per-R
+  # uncertainties are symmetrcial, so same value is used for +/-
+  DELTA_MINUS_VAR: flux_uncertainties
+  DELTA_PLUS_VAR: flux_uncertainties
   # SPASE data pulled from TWINS 1 LAD metadata
   DICT_KEY: SPASE>Wave>WaveType:Electromagnetic,WaveQuantity:Intensity,Qualifier:Scalar
 


imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
```diff
@@ -407,6 +407,9 @@ photon_flux:
   FORMAT: F8.2
   VALIDMIN: 0.0
   VALIDMAX: 30000.0 # max 30000 cps in FDIR / from 1 to 3.37 cps-per-R
+  # uncertainties are symmetrcial, so same value is used for +/-
+  DELTA_MINUS_VAR: flux_uncertainties
+  DELTA_PLUS_VAR: flux_uncertainties
```
tech3371, line 412: Do other variables in this file need delta plus/minus too? Eg. `flux_uncertainties` or `raw_uncertainties` or others. If we don't know yet, it's ok. When you run this pass by Andriy, he has been good with identifying those needs. @jtniehof is also a good person to receive inputs like this.

tech3371 (approved):
looks good besides one minor question.

imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
sdhoyt: Thanks, I'll look at that when I'm doing general metadata checks

PR #2748, opened 2026-02-20
ULTRA l1b add function to flag events that are within the earth keepout angle. 

lacoak21:
# Change Summary

## Overview
Add a function to flag events where the particle look direction is outside the Earth keepout angle. A lot of this PR is pulling out code from the function `compute_culling_mask` that @laspsandoval wrote and creating a new function `compute_unit_target_vectors`.

## File changes
    - pull out code from compute_culling_mask and make a new function that can be used generally for other instruments 
   - Compute the angle between the earth unit vector and each particle's look direction and flag events where the angle is less than the constant threshold given by the ULTRA IT team. 

## Testing
I wrote a test that uses the exact code that Bob uses in order to validate our functions. It can be found here https://github.com/demajistre/imap_processing/blob/dev/ultra_user/planets/ENA_planets.py


imap_processing/spice/geometry.py
```diff
@@ -762,3 +762,43 @@ def solar_longitude(
     lat_coords = cartesian_to_latitudinal(imap_pos, degrees=degrees)[..., 1]
 
     return float(lat_coords) if lat_coords.size == 1 else lat_coords
+
+
+def compute_unit_target_vectors(
```
lacoak21: The is the code that was pulled from compute_culling_mask that Laura already tested and validated but I wrote a test for it just to be sure.

imap_processing/ultra/l1c/ultra_l1c_culling.py
```diff
@@ -50,24 +50,13 @@ def compute_culling_mask(
...
-    state = imap_state(et, ref_frame=SpiceFrame.IMAP_DPS, observer=observer)
-    # Flip to get vector from IMAP to Earth
-    # position.shape = (len(et), 3)
-    position = -state[:, :3]
-
-    # Distance from IMAP to target (e.g. Earth) (km):
-    # distance.shape = (len(et),)
-    distance = np.linalg.norm(position, axis=1)  # shape (len(et),)
-
+    unit_target_vecs, distance = compute_unit_target_vectors(
```
laspsandoval: Nice way to organize.

imap_processing/ultra/constants.py
```diff
@@ -191,3 +193,5 @@ class UltraConstants:
     N_CULL_EBINS = 8
     # Bin to start culling at
     BASE_CULL_EBIN = 4
+    # Angle threshold in radians for 45 degree culling.
+    EARTH_ANGLE_45_THRESHOLD = np.radians(15)
```
laspsandoval, line 198: Will this be different for Ultra 90? Why 15 degrees?

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -675,6 +680,50 @@ def flag_low_voltage(
...
+        Earth keepout angle threshold (in radians) for ULTRA 45 instrument.
+
+    Returns
+    -------
+    valid_earth_angle_events : NDArray
+        A boolean array indicating which events have Earth angle greater than the
+        specified threshold.
+    """
+    de_dps_velocity = de_dataset["de_dps_velocity"].values
+    et = np.mean(de_dataset["event_times"].values)
```
laspsandoval: Why did you take the mean here and not just directly use the event time?

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -675,6 +680,50 @@ def flag_low_voltage(
...
+    )[0].squeeze()  # shape (3,)
+    # Calculate the magnitude of the velocity vector for each event
+    particle_mag = np.linalg.norm(de_dps_velocity, axis=1)
+    # Normalize and flip to get where each particle is looking.
+    unit_look_dirs = (
+        -de_dps_velocity / particle_mag[:, np.newaxis]
+    )  # shape (n_events, 3)
+    # Get cos(theta) between each particle look direction and Earth direction
+    cos_sep = np.dot(unit_look_dirs, earth_unit_vector)  # shape (n_events,)
+    # Clip cos_sep to the valid range of [-1, 1]
```
laspsandoval: Put reason for this in the comment (eliminate round off errors prior to taking the dot product; or something like that).

laspsandoval (commented):
A question about the mean calculation, but otherwise it looks good.

imap_processing/spice/geometry.py
tmplummer: This function seems quite specific. I might suggest making it more general by essentially making it a wrapper around spiceypy.spkezr. Something like:
```
def body_direction_and_distance(target_body: SpiceBody, et: float | NDarray, ref_rame: SpiceFrame, observer: SpiceBody, abcorr: bool = False):
    body_state, _ spiceypy.spkezr(target.name, et, ref_frame.name, abcorr, observer.name)
    position = body_state[:3]
    distance = np.linalg.norm(position, axis=1)
    return position/distance, distance
```
https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.spkezr

Alternatively, I think it would be appropriate to just use spiceypy.spkezr directly in your Ultra code. I just wonder if this very specific function is

tmplummer (commented):
Other than the one suggestion, looks good to me.

imap_processing/spice/geometry.py
lacoak21: OK that sounds good to me too. Yeah it probably is very specific - but I havent actually checked.

tmplummer (approved):
LGTM

PR #2749, opened 2026-02-20
Add variables to Hi L1B DE product

### LINKED ISSUE #2746: Hi L1B - add spin_set_met and packet_full variables
### Algorithm Description:

While working on Hi Goodtimes, it became apparent that I need to pass the 8-spin MET start time from the L1A DE file into the L1B DE file. It would also be useful to add a flag indicating which packets are full (664 events).

tmplummer:
# Change Summary

## Overview
                                                                                                                                                                                                                                                                
  Adds two new variables to the Hi L1B Direct Event product              

##  New Variables

  1. esa_step_met (float64, epoch dimension)
    - MET time when the ESA was stepped, computed from esa_step_seconds + esa_step_milliseconds / 1000
    - Provides a convenient floating-point representation of the ESA step timestamp
  2. ccsds_qf (uint8, epoch dimension)
    - Quality flag bitmask for each CCSDS packet
    - Bit 0 (PACKET_FULL): Set when packet contains 664 events (maximum capacity)

##  Implementation Details

  - Added de_esa_step_met() function to compute ESA step MET
  - Added de_ccsds_qf() function to compute CCSDS quality flags
  - Added ImapHiL1bDeFlags flag class to quality_flags.py following existing patterns
  - Added variable attribute definitions to imap_hi_variable_attrs.yaml
  - L1B DE output now contains 17 data variables (previously 15)

##  Files Changed

  - imap_processing/hi/hi_l1b.py - Added two new processing functions
  - imap_processing/quality_flags.py - Added ImapHiL1bDeFlags class
  - imap_processing/cdf/config/imap_hi_variable_attrs.yaml - Added variable definitions
  - imap_processing/tests/hi/test_hi_l1b.py - Added tests for new functions

Closes: #2746

### DIFF
--- a/imap_processing/hi/hi_l1b.py
+++ b/imap_processing/hi/hi_l1b.py
@@ -11,7 +11,7 @@ import xarray as xr
 from imap_processing import imap_module_directory
 from imap_processing.cdf.imap_cdf_manager import ImapCdfAttributes
 from imap_processing.cdf.utils import parse_filename_like
-from imap_processing.hi.hi_l1a import HALF_CLOCK_TICK_S
+from imap_processing.hi.hi_l1a import HALF_CLOCK_TICK_S, MILLISECOND_TO_S
 from imap_processing.hi.utils import (
     HIAPID,
     CoincidenceBitmap,
@@ -20,6 +20,7 @@ from imap_processing.hi.utils import (
     create_dataset_variables,
     parse_sensor_number,
 )
+from imap_processing.quality_flags import ImapHiL1bDeFlags
 from imap_processing.spice.geometry import (
     SpiceFrame,
     instrument_pointing,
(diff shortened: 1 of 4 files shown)


imap_processing/hi/hi_l1b.py
```diff
@@ -531,3 +534,86 @@ def get_esa_to_esa_energy_step_lut(
...
+
+    # Count events per CCSDS packet
+    # ccsds_index maps each event to its originating packet
+    ccsds_indices = dataset["ccsds_index"].values
+    n_packets = len(dataset.epoch)
+
+    # Filter out fill/out-of-range indices (e.g., uint16 FILLVAL 65535)
+    valid_mask = (ccsds_indices >= 0) & (ccsds_indices < n_packets)
+
+    # If there are no valid events, all packets keep default quality flag 0
```
lacoak21, line 606: I would think that if there were no valid events you would want everything flagged? Not with ImapHiL1bDeFlags.PACKET_FULL but another one?

lacoak21 (approved):

imap_processing/hi/hi_l1b.py
tmplummer: I could add another flag to indicate an empty packet but currently, the only flag I have defined indicates that the CCSDS packet contained the maximum number of events. You are correct that a packet having no packets probably indicates a problem but that may be caught elsewhere.

PR #2758, opened 2026-02-24
Lo L1c GoodTime Filter Fix

### LINKED ISSUE #2757: BUG - Lo L1C GoodTime filtering is not working properly
### Description of the issue

Lo PSETs created over the past few weeks have shown that GoodTimes are not being correctly filtered.

When GoodTimes are not properly filtered, the PSETs will also have H and O background rates that are zero.

_No response_

```python
Code
```

_No response_

ahotasu:
Updated the L1c GoodTime filter to be like the GoodTime fraction logic at line 684 as a response to bug #2757.

# Change Summary
Changed logic to ensure that the GoodTime is within the pointing time. 

Also added logic to return an empty dataset if there are no GoodTimes within the pointing. This seems reasonable to me--it prevents later issues with background rate determination, but I appreciate other thoughts on this solution.

## Testing
Reprocessed all L1c PSETs since Jan 17 locally and validated that the new logic is working by inspecting CDFs with MATLAB to ensure that the background rates are all non-zero, which must be the case when the GoodTime filtering is working properly.
Re-ran pytest for Lo L1c; all tests pass.


imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -87,7 +87,13 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
         logical_source = "imap_lo_l1c_pset"
         l1b_de = sci_dependencies["imap_lo_l1b_de"]
         l1b_goodtimes_only = filter_goodtimes(l1b_de, anc_dependencies)
-        # TODO: Need to handle case where no good times are found
+
+        # Handle case where no good times are found; if we don't bail
+        # here, then later background rates determination will fail
+        if l1b_goodtimes_only["epoch"].size == 0:
+            logging.warning("No good times found for L1B DE dataset.")
+            return []
```
tmplummer: This will cause L1C to succeed but produce no L1C file. Is that the desired behavior? Or should it raise an exception so that we see the job failed?

sdhoyt: Is there handling in place for returning an empty list?

sdhoyt (approved):

imap_processing/lo/l1c/lo_l1c.py
```diff
@@ -87,7 +87,13 @@ def lo_l1c(sci_dependencies: dict, anc_dependencies: list) -> list[xr.Dataset]:
         logical_source = "imap_lo_l1c_pset"
         l1b_de = sci_dependencies["imap_lo_l1b_de"]
         l1b_goodtimes_only = filter_goodtimes(l1b_de, anc_dependencies)
-        # TODO: Need to handle case where no good times are found
+
+        # Handle case where no good times are found; if we don't bail
+        # here, then later background rates determination will fail
+        if l1b_goodtimes_only["epoch"].size == 0:
```
tmplummer: Will you please add test coverage for this if block?

tmplummer (commented):
I have a couple of comments.

imap_processing/lo/l1c/lo_l1c.py
ahotasu: I've tested it and it works--no CDF product is created. See reply below for more info.

ahotasu: That's an excellent question that I was hoping someone would raise--it seems to me to be a philosophy/approach that the SDC needs to adopt for all instruments. I wasn't sure if this scenario had been seen or thought of before, so I was fishing for direction here... For now, this works as you indicate--I've tested with flight data that on certain days has no GoodTimes (so far?).

imap_processing/lo/l1c/lo_l1c.py
ahotasu: Will do.

ahotasu:
@subagonsouth : Updated logic per yesterday's mapping meeting to return an empty pset instead of simply not producing a pset product. Added test case and also verified using flight data for 20260209, which has no (currently defined) good times.

PR #2762, opened 2026-02-25
ULTRA l1b High Energy (SEP) culling

### LINKED ISSUE #2711: high energy cull

lacoak21:
# Change Summary

closes  #2711

## Overview

This PR adds a quality flag array to the ULTRA L1B extended spin and goodtimes CDF. The new quality flag array indicates spins and energy ranges that should be removed due to high energy events.
First, valid events are selected based on other flags and criteria at each energy range. Then, using the selected "culling energy channel" — the channel with the highest energy events, identified by the ULTRA IT team — entire groups of spins are flagged. The culling energy channel is shown to be correlated with SEPs in other channels, so it can be used to cull contaminated spins. Each energy channel has its own threshold, also provided by the ULTRA IT team. The counts at each spin bin for the culling channel are calculated, and then for each energy bin, if the counts exceed the energy threshold, that spin bin is flagged at that energy range.

These flags will be used to cull spins at l1c. 

## File changes
    - Add necessary culling constants 
- imap_processing/ultra/l1b/extendedspin.py 
   - Remove low voltage events from the DE dataset and then calculate the high energy quality flags
   - Add function to flag high energy events 
   - Add function to get valid events for each energy range 

## Testing
Add basic testing and a validation test. The validation files were computed using Bob Demajistre's code.

### DIFF
--- a/imap_processing/ultra/l1b/ultra_l1b_culling.py
+++ b/imap_processing/ultra/l1b/ultra_l1b_culling.py
@@ -712,6 +712,7 @@ def flag_high_energy(
     quality_flags : numpy.ndarray
         Quality flags.
     """
+    cull_channel = UltraConstants.HIGH_ENERGY_CULL_CHANNEL
     valid_events_per_energy = get_valid_events_per_energy_range(
         de_dataset, energy_ranges, UltraConstants.EARTH_ANGLE_45_THRESHOLD, sensor_id
     )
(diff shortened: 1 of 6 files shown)


imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -680,6 +680,136 @@ def flag_low_voltage(
...
+        Sensor ID (e.g., 45 or 90).
+
+    Returns
+    -------
+    valid_events_per_range : numpy.ndarray
+        A boolean array of shape (n_energy_ranges, n_events).
+    """
+    event_energies = de_dataset["energy_spacecraft"].values
+    valid_events_per_range = []
+    for i in range(len(energy_ranges) - 1):
```
tmplummer, line 785: I'll channel Greg and comment that this could probably be vectorized. It looks like several of the masks are not energy dependent like `valid_outliers`, `valid_scattering`, `valid_ebin`, and `valid_earth_angle`. One thought would be to add a variable to the dataset, or return a DataArray, that contains energy bin indices for all valid events and nan, or some specific value, for non-valid events. I won't hold approval based on this nit-pick.

tmplummer (approved):
I had a hard time finding anything to comment on. This all looks good to me!

PR #2763, opened 2026-02-25
2742 hi goodtimes   statistical filter 2

### LINKED ISSUE #2742: Hi Goodtimes - statistical filter 2
### Algorithm Description:

From the algorithm document:

> Statistical Filter 2: observations during commissioning show occasional short-lived “pulses” of qual-
ified counts, at least sometimes temporally correlated between H45 and H90, usually visible only at
the highest few energy steps. Usually, Statistical Filter 1 removes these.
To remove any remaining times:
– per SCI DE packet (correcting for any packet-shifted events as discussed previously), keep only
events qualifying as calibration product 1 or calibration product 2 (regardless of the definition of
these products).
– sort the events by timetag.
– locate all time ranges where n or more events were observed within m or fewer ticks.
– for the entire 8-spin science interval at this ESA step, mark the angle range defined by each time
range, plus one 4-degree “bin” on either side, as not good.
The specific values of m and n will likely need to be adjusted over the mission and if the calibration
products are redefined. Features of this sort generally appear to cover less than a full spin (maybe
180 or a bit more), so a time of 15s × 240
360 × 1 tick/2ms or 5000 ticks (10s) is reasonable. Summed
calibration products counts are observed to be generally less than 0.1/s, including background. Poisson
stats would then imply the odds of seeing 6 (7) or more counts in a 10 second interval are under 0.06%
(0.01%) and thus shouldn’t remove much actual signal. At the start of science operations, the values
to use are 6 and 5000.

tmplummer:
# Change Summary

## Summary                                                                                                                                                                                                                                                       
   
Implements Statistical Filter 2 from the IMAP-Hi Algorithm Document Section 2.3.2.3, which detects and removes short-lived "pulses" of qualified counts that may be temporally correlated between sensors.                                                    
                                                                             
## Changes

  hi_goodtimes.py (+250 lines)

  mark_statistical_filter_2: Main filter function that:
  - Groups events by 8-spin set (esa_sweep, esa_step)
  - Filters to qualified coincidence types (calibration products 1 or 2)
  - Sorts events by event_met
  - Finds clusters where ≥min_events occur within max_time_delta seconds
  - Marks affected spin bins (with padding and wraparound) for all METs in that 8-spin set

  _find_event_clusters: Vectorized helper that finds time clusters using:
  - Window span computation: event_times[n-1:] - event_times[:-n+1]
  - Transition detection via padded np.diff to identify contiguous regions

  _compute_bins_for_cluster: Computes spin bins to cull with:
  - np.unwrap to handle clusters spanning the 0/89 boundary
  - Configurable bin padding with modular wraparound

  utils.py (+29 lines)

  Added to HiConstants:
  - STAT_FILTER_2_MIN_EVENTS = 6
  - STAT_FILTER_2_MAX_TIME_DELTA = 10.0 (seconds)
  - STAT_FILTER_2_BIN_PADDING = 1
  - DE_CLOCK_TICK_DURATION_SECONDS and DE_CLOCK_TICK_DURATION_MICROSECONDS (moved from hi_l1a.py)

  hi_l1a.py, hi_l1b.py, hi_l1c.py, test_hi_l1c.py

  - Refactored to use HiConstants.DE_CLOCK_TICK_* instead of local constants

  test_hi_goodtimes.py (+465 lines)

  - TestFindEventClusters: 7 tests for cluster detection (empty, too few, spread out, single/multiple clusters, merging, exact threshold)
  - TestComputeBinsForCluster: 6 tests for bin computation (basic range, wrapping at 0/max, partial cluster, boundary spanning)
  - TestStatisticalFilter2: 6 integration tests (no qualified events, no clusters, cluster detected, multiple clusters, bin padding with wrapping, custom parameters)

Closes: #2742

### DIFF
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -1655,9 +1655,11 @@ def mark_statistical_filter_2(
     bins as bad, unlike Filters 0 and 1 which mark entire time intervals.
     Bin marking wraps around (e.g., bin 91 becomes bin 1).
 
-    The default parameters (min_events=6, max_time_delta=10.0) imply that
-    seeing 6+ qualified events in a 10-second window has probability
-    < 0.06% under normal conditions (background rate ~0.1/s).
+    The default parameters (min_events=6,
+    max_time_delta=HiConstants.STAT_FILTER_2_MAX_TIME_DELTA ≈ 9.995 s)
+    imply that seeing 6+ qualified events in an approximately 10-second
+    window has probability < 0.06% under normal conditions
+    (background rate ~0.1/s).
     """
     logger.info("Running mark_statistical_filter_2 culling")
 
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -1664,49 +1664,41 @@ def mark_statistical_filter_2(
     # Add esa_sweep coordinate to group packets into 8-spin sets
     l1b_de_with_sweep = _add_sweep_indices(l1b_de)
 
-    # Extract required variables
+    # Get packet-level arrays
     ccsds_index = l1b_de_with_sweep["ccsds_index"].values
-    coincidence_type = l1b_de_with_sweep["coincidence_type"].values
-    nominal_bin = l1b_de_with_sweep["nominal_bin"].values
-    l1b_event_met = l1b_de_with_sweep["event_met"].values
-
-    # Get esa_sweep and esa_step for each event via ccsds_index
     esa_sweep = l1b_de_with_sweep.coords["esa_sweep"].values
     esa_step = l1b_de_with_sweep["esa_step"].values
-    ccsds_met = l1b_de_with_sweep["ccsds_met"].values
 
-    # Map events to their packet's (esa_sweep, esa_step)
-    event_sweep = esa_sweep[ccsds_index]
-    event_step = esa_step[ccsds_index]
+    # Add event-level coordinates for grouping
+    l1b_de_with_sweep = l1b_de_with_sweep.assign_coords(
+        event_sweep=("event", esa_sweep[ccsds_index]),
+        event_step=("event", esa_step[ccsds_index]),
+    )
 
     # Filter to qualified events
+    coincidence_type = l1b_de_with_sweep["coincidence_type"].values
     is_qualified = np.isin(coincidence_type, list(qualified_coincidence_types))
 
     if not np.any(is_qualified):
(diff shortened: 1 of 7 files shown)


imap_processing/hi/hi_goodtimes.py
```diff
@@ -1495,3 +1495,249 @@ def mark_statistical_filter_1(
...
+    max_time_delta : float
+        Maximum time span in seconds for events to be considered clustered.
+
+    Returns
+    -------
+    list[tuple[int, int]]
+        List of (start_idx, end_idx) tuples marking cluster boundaries
+        in the input array. Indices are inclusive.
+    """
+    if len(event_times) < min_events:
```
lacoak21, line 1526: Isnt this a duplicate of lines 1697 and 1698?

imap_processing/hi/hi_goodtimes.py
```diff
@@ -1495,3 +1495,249 @@ def mark_statistical_filter_1(
...
+
+    # Find contiguous regions of True in the mask. Each contiguous region
+    # [i, j] in the mask corresponds to cluster [i, j + min_events - 1]
+
+    # Pad with False to handle edge cases
+    padded = np.concatenate(([False], cluster_mask, [False]))
+
+    # Find transitions: +1 = start of group, -1 = end of group
+    diff = np.diff(padded.astype(int))
+    starts = np.flatnonzero(diff == 1)
```
lacoak21, line 1547: why do you need flatnonzero? Isnt diff 1d?

imap_processing/tests/hi/test_hi_goodtimes.py
```diff
@@ -2484,3 +2487,465 @@ def test_current_index_out_of_range(self, goodtimes_for_filter1):
...
+        mark_statistical_filter_2(
+            goodtimes_for_filter2,
+            l1b_de,
+            qualified_types,
+            min_events=6,
+            max_time_delta=0.1,
+            bin_padding=2,
+        )
+
+        cull_flags = goodtimes_for_filter2["cull_flags"].sel(met=1000.0).values
```
lacoak21, line 2901: is it worth checking that the last bins arent flagged as well?

lacoak21 (approved):
LGTM!  💯

imap_processing/hi/hi_goodtimes.py
tmplummer: flatnonzero returns the array of indices directly instead of a tuple of array of indices. So, it avoids having to add the `[0]` on the end to index the first returned element.

lacoak21: Ohh - I see.

PR #2775, opened 2026-02-25
Skip L2 processing if no good times are found

### LINKED ISSUE #2372: ENH - Skip L2 processing when no good times are found
## Description

Currently, when no good times are found in L1B data, the L2 processing continues and creates an output file with potentially invalid or meaningless data. The code should exit early or skip processing to avoid creating invalid output files.

## Location

`imap_processing/glows/l2/glows_l2.py:80-91`

## Current Implementation

```python
if len(good_data["epoch"]) != 0:
    var_outputs["start_time"] = good_data["epoch"].data[0]
    var_outputs["end_time"] = good_data["epoch"].data[-1]
else:
    # No good times in the file
    var_outputs["start_time"] = l1b_dataset["imap_start_time"].data[0]
    var_outputs["end_time"] = (
        l1b_dataset["imap_start_time"].data[0]
        + l1b_dataset["imap_time_offset"].data[0]
    )
# Processing continues regardless...
```

## Required Changes

1. After detecting `len(good_data["epoch"]) == 0`, add early return or raise appropriate exception
2. Log warning message about no good times found
3. Decide appropriate behavior:
   - Option A: Return None and skip file creation
   - Option B: Raise exception with clear message
   - Option C: Create file with special flag indicating no valid data

## Related

Part of #1276 (GLOWS L2 follow up)

vmartinez-cu:
This PR updates L2 processing to exit processing early and return an empty list if no good times are found in the l1b input dataset.

## File changes
- glows_l2_data.py - added a factory class method that checks if good times exist before proceeding with creating the l2 histogram data class. If no good times are found, it returns None

- glows_l2.py - updated call to the histogram data class to use new class method to create the instance

## Testing

- test_glows_l2.py - updated existing unit tests to include coverage for when there are no good times in the data

Closes #2372

### DIFF
--- a/imap_processing/glows/l2/glows_l2_data.py
+++ b/imap_processing/glows/l2/glows_l2_data.py
@@ -1,7 +1,5 @@
 """Module containing the class definition for the HistogramL2 class."""
 
-from __future__ import annotations
-
 from dataclasses import InitVar, dataclass, field
 
 import numpy as np
@@ -125,8 +123,10 @@ class HistogramL2:
 
     Parameters
     ----------
-        good_l1b_data : xr.Dataset
-            GLOWS histogram L1B dataset filtered by good times.
+        l1b_dataset : xr.Dataset
+            GLOWS histogram L1B dataset, as produced by glows_l1b.py.
+        pipeline_settings : PipelineSettings
+            Pipeline settings object read from ancillary file.
 
     Attributes
     ----------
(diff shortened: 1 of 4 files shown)


imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -210,123 +210,142 @@ class HistogramL2:
...
         )
         self.spin_axis_orientation_std_dev = (
-            good_data["spin_axis_orientation_average"]
+            good_l1b_data["spin_axis_orientation_average"]
             .std(dim="epoch")
             .data[np.newaxis, :]
         )
 
+    @classmethod
+    def create(
```
maxinelasp: This duplicates the current __init__ function and makes it confusing IMO (this makes L2 behave differently from the other levels.) Instead of making this a class function, why not make it a static function that can then be used at the beginning of __init__? Or just leaving it inside init?

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -210,123 +210,142 @@ class HistogramL2:
     spin_axis_orientation_average: np.ndarray[np.double]
     bad_time_flag_occurrences: np.ndarray
 
-    def __init__(self, l1b_dataset: xr.Dataset, pipeline_settings: PipelineSettings):
+    def __init__(self, good_l1b_data: xr.Dataset):
```
maxinelasp: PipelineSettings has other data in it that needs to be used for future L2 functionality, so it still needs to be included in this init

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -210,123 +210,142 @@ class HistogramL2:
...
+            A HistogramL2 object if there are good times in the dataset,
+            or None if there are no good times.
+        """
+        # Check if dataset contains good times
+        good_idx = cls.return_good_times(
+            l1b_dataset["flags"],
+            np.array(pipeline_settings.active_bad_time_flags, dtype=float),
+        )
+        if len(good_idx) == 0:
+            return None
```
maxinelasp: It is confusing to me that a class creation step could return `None`. I would prefer that this check either occurs _before_ class creation (maybe with a static method so you could do `good_data = l1b_dataset.isel(HistogramL2.select_good_times(l1b_dataset["flags"], active_bad_time_flags])` and check if good_data has a length of zero before creating `HistogramL2`) or occur _after_ class creation, since all the function calls should work and be very fast with no data and you can check to see if `HistogramL2.number_of_good_l1b_inputs` is not zero or something

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -210,123 +210,142 @@ class HistogramL2:
...
         # TODO filter bad bins out. Needs to happen here while everything is still
-        # per-timestamp.
+        #  per-timestamp.
 
-        self.daily_lightcurve = DailyLightcurve(good_data)
+        self.daily_lightcurve = DailyLightcurve(good_l1b_data)
 
-        self.total_l1b_inputs = len(good_data["epoch"])
-        self.number_of_good_l1b_inputs = len(good_data["epoch"])
+        self.total_l1b_inputs = len(good_l1b_data["epoch"])
```
maxinelasp: This is not correct - this variable should be ALL the L1B inputs, including bad times.

maxinelasp (changes requested):
I think this implementation method is confusing - could you remove the class method and rework the logic slightly?

imap_processing/glows/l2/glows_l2_data.py
vmartinez-cu: I wasn't able to return None from init which is why I used a factory pattern here. To avoid returning dataset class objects that have bad data, I was trying to cut the processing early.  I could raise an exception instead and catch it in `glows_l2.py`?  

I was also trying to keep processing steps in the dataclass file to maintain the separation of concerns in the two L2 files. I interpreted `glows_l2.py` as creating an xarray dataset with cdf attributes while `glows_l2_data.py` handles filtering and processing of the data. If this isn't a strict separation of tasks, I can do this filtering before calling HistogramL2 in `glows_l2.py` if that's preferred

imap_processing/glows/l2/glows_l2_data.py
vmartinez-cu: Didn't know about future L2 functionality. Can you expand on that?

imap_processing/glows/l2/glows_l2_data.py
vmartinez-cu: The original variable passed in was "good_data" which only included the good times.  This change is only renaming that variable. I'm not sure I follow

imap_processing/glows/l2/glows_l2_data.py
vmartinez-cu: Filtering could certainly occur before class creation. I was trying to keep all processing steps out of the glows_l2.py file but if that's not important this is an easy change. I think filtering first then creating the object is better than creating a bad object since those steps aren't necessary even if they're fast

maxinelasp: Yeah this one is kind of weird because it is a processing step but also may short-circuit processing. I probably did a bad job explaining, but although the bulk of processing occurs in HistogramL2, there are some input/output steps that happen in glows_l2.py and I think this could qualify for one of those.

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: Ah, I see - that was a bug from me! total_l1b_inputs should have all inputs, good and bad. number_of_good_l1b_inputs should just have the good inputs. If you want to just pass the good data in like you do here (which I like) then total_l1b_inputs will need to be passed in to the init as well.

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: Yeah, like I mentioned in another comment, although there is generally a separation of concerns in the two files, this is a weird case that messes with that separation. I think the dataset with no good data isn't necessarily bad - even though we don't want to output it, it's not an incorrect thing to create. And if there aren't any good times then processing will basically return a bunch of empty arrays which isn't bad data, it's just... all been filtered away 😂 

I would also be fine with a caught exception, I think that is also a logical way to handle this issue!

vmartinez-cu:
Updates made and ready for review

tech3371 (approved):
LGTM

tech3371:
and nice work!!!

maxinelasp (approved):
Looks good, thanks!!

PR #2780, opened 2026-02-26
Update L1B to inherit the number of bins from L1A

maxinelasp:
### Background

Per ISTP/CDF convention, all variables must be allocated to the maximum expected size (3600 bins), with unused values filled with `FILLVAL`. In the GLOWS packet format, `HISTOGRAM_DATA` is a fixed 3600-element field. However, `NBINS` is a separate raw telemetry field propagated as `number_of_bins_per_histogram` — this encoded value (e.g. 225) is not the bin array length.

If we get sent fewer bins, we still want the output variables to be of len 3600. 

I found this issue when testing against validation data that is 255 bins long.

### Fix

Replace `self.number_of_bins_per_histogram` with `len(self.histogram)` wherever it sizes output arrays in `glows_l1b_data.py`. The field is retained in the dataclass output as a passthrough of the raw telemetry value.

### DIFF
--- a/imap_processing/glows/l1b/glows_l1b_data.py
+++ b/imap_processing/glows/l1b/glows_l1b_data.py
@@ -826,10 +826,9 @@ class HistogramL1B:
 
         # Add SPICE related variables
         self.update_spice_parameters()
-        # Calculate the spin angle bin center
-        phi = (
-            np.arange(self.number_of_bins_per_histogram, dtype=np.float64) + 0.5
-        ) / self.number_of_bins_per_histogram
+        # Calculate the spin angle bin center using actual histogram length from L1A
+        n_bins = len(self.histogram)
+        phi = (np.arange(n_bins, dtype=np.float64) + 0.5) / n_bins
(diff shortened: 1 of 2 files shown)


imap_processing/glows/l1b/glows_l1b_data.py
```diff
@@ -1076,7 +1075,7 @@ def _compute_histogram_flag_array(
             Array of shape (4, 3600) with bad-angle flags for each bin.
         """
         histogram_flags = np.full(
-            (4, self.number_of_bins_per_histogram),
```
laspsandoval, line 1079: Is number_of_bins_per_histogram being used anywhere? If not delete.

laspsandoval (approved):
Thanks Maxine.

imap_processing/glows/l1b/glows_l1b_data.py
maxinelasp: Good question, it does need to be output in the file and will be used for other calculations (since if half the bins are FILLVAL, it will effect averaging etc)

PR #2781, opened 2026-02-27
ULTRA l1b statistical outlier culling 

### LINKED ISSUE #2710: statistic cull

lacoak21:
# Change Summary

closes #2710 
## Overview

This PR implements the statistical culling algorithm provided by the ULTRA IT team. 

Algorithm explained: 
After removing the voltage drops and the high energy events, there still appear to be time some dependence in the signal during a pointing. 
•For each energy range in a pointing, we can calculate a mean and a standard deviation.  
•If the counts come from a source with no time dependence, then the calculated standard deviation should be very close to the square root of the mean (i.e., they should obey Poisson statistics). 
•If there is time variation in the signal, then the standard deviation will be larger than the square root of the mean.

This PR also changes the return qf arrays from the high energy and low voltage culling algorithms. Originally they were bit flag arrays but since I needed to use those masks in the stat culling algorithm i decided it would be far easier to keep them as bools and then convert them to flag arrays later in the code. 

*NOTE*
This code is basically the same as bobs culling code that he provided for reference [here. ](https://github.com/demajistre/imap_processing/blob/dev/ultra_user/culling/UltraCull0.py). I didnt find many areas to improve since it is an iterative process and can not be vectorized so I followed his code very closely. Again, there will be a future PR with the validation test that compares with his outputs directly.  

## Testing

Added tests for the stat cull. I will be adding the validation test in a separate PR because this one started getting pretty big.

### DIFF
--- a/imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
+++ b/imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
@@ -27,6 +27,7 @@ from imap_processing.ultra.l1b.ultra_l1b_culling import (
     flag_low_voltage,
     flag_rates,
     flag_scattering,
+    flag_statistical_outliers,
     get_binned_energy_ranges,
     get_binned_spins_edges,
     get_de_rejection_mask,
@@ -34,6 +35,7 @@ from imap_processing.ultra.l1b.ultra_l1b_culling import (
     get_energy_histogram,
     get_energy_range_flags,
     get_n_sigma,
+    get_poisson_stats,
     get_pulses_per_spin,
     get_spin_data,
     get_valid_earth_angle_events,
@@ -644,10 +646,10 @@ def test_flag_high_energy():
     energy_range_edges = np.array([3, 5, 7, 18, 25])  # Example energy bin edges
     # Spin bin 1 (events 0-3) 4 events fall within the culling energy bin
     #   - This is above all the energy thresholds except the second one (10),
-    #   so it should be flagged for all energy ranges except flag #2
+    #   so it should be flagged for all energy ranges except flag the 2nd
     # Spin bin 2 (events 4-7) only 1 event falls within the culling energy bin
     #   - This is above the lowest energy threshold (1) but below the rest, so
-    #   it should only be flagged with the last energy range flag (#3)
+    #   it should only be flagged at the last energy range
     # Spin bin 3 (events 8-11) No events fall within the culling energy bin,
     #   so it should not be flagged for any energy range
     energy = np.array([17, 16, 12, 15, 5, 8, 4, 6, 4, 1, 22, 20])
@@ -666,26 +668,31 @@ def test_flag_high_energy():
     spin_tbin_edges = np.arange(
         start=0, stop=len(energy) + 1, step=4
     )  # create spin bins of 4 seconds
-    energy_range_flags = get_energy_range_flags(energy_range_edges)
     quality_flags = flag_high_energy(
         de_dataset,
         spin_tbin_edges,
         energy_range_edges,
-        energy_range_flags,
+        None,
         cull_thresholds,
         90,
     )
 
     # check shape
-    assert len(quality_flags) == len(spin_tbin_edges) - 1
+    np.testing.assert_array_equal(
+        quality_flags.shape, (len(energy_range_edges) - 1, len(spin_tbin_edges) - 1)
+    )
     # Assert that the first spin bin is flagged for high energy for all energy ranges
     # except the second one
-    assert quality_flags[0] == (2**0 | 2**2 | 2**3)
-    # Assert that the second spin bin is only flagged for high energy for the last
-    # energy range
-    assert quality_flags[1] == 2**3
-    # Assert that the third spin bin is not flagged for any energy range
-    assert quality_flags[2] == 0
+    quality_flags[0, 0] = True
+    quality_flags[1, 0] = False
+    quality_flags[2, 0] = True
+    quality_flags[3, 0] = True
+    # # Assert that the second spin bin is only flagged for high energy for the last
+    # # energy range
+    quality_flags[3, 1] = True
+    assert not np.any(quality_flags[0:3, 1])
+    # # Assert that the third spin bin is not flagged for any energy range
+    assert not np.any(quality_flags[:, 2])
 
 
 @pytest.mark.external_test_data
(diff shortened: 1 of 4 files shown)


imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
```diff
@@ -644,10 +641,10 @@ def test_flag_high_energy():
     energy_range_edges = np.array([3, 5, 7, 18, 25])  # Example energy bin edges
     # Spin bin 1 (events 0-3) 4 events fall within the culling energy bin
     #   - This is above all the energy thresholds except the second one (10),
-    #   so it should be flagged for all energy ranges except flag #2
+    #   so it should be flagged for all energy ranges except flag the 2nd
```
tech3371: is this comment accurate?

imap_processing/ultra/l1b/extendedspin.py
```diff
@@ -132,8 +137,25 @@ def calculate_extendedspin(
...
+    stat_outliers_qf = np.bitwise_or.reduce(
+        stat_outliers_qf * energy_bin_flags[:, np.newaxis], axis=0
+    )
+    # Low voltage flag is shape (n_spin_bins,) but we want to convert from a boolean
+    # to a bitwise flag to be consistent with the other flags, where each spin that
+    # is flagged will have the bitflag of all the energy flags combined.
+    voltage_qf = voltage_qf * np.bitwise_or.reduce(energy_bin_flags)
     # Expand binned quality flags to individual spins.
     high_energy_qf = expand_bin_flags_to_spins(len(spin), high_energy_qf, spin_bin_size)
+    voltage_qf = expand_bin_flags_to_spins(len(spin), voltage_qf, spin_bin_size)
```
laspsandoval, line 155: Here:
voltage_qf (True = bad voltage)

Later you pass it to flag_high_energy as "mask":

if mask is not None:
      quality_flags[:, mask] = flagged[:, mask]

Should it be ~mask?

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -710,48 +708,250 @@ def flag_high_energy(
...
+    std_ratio : float
+        Ratio of the observed standard deviation to the expected Poisson
+        standard deviation.
+    sub_mask : numpy.ndarray
+        Boolean array of the same length as counts. True where a bin is
+        a statistical outlier (more than 3 sigma from the mean).
+    """
+    std = np.std(counts)
+    std_ratio = std / np.sqrt(np.mean(counts)) - 1
+    sub_mask = np.abs((counts - np.mean(counts)) / std) > 3
```
laspsandoval, line 918: Will std ever be zero here?

imap_processing/tests/ultra/unit/test_ultra_l1b_culling.py
```diff
@@ -666,26 +663,31 @@ def test_flag_high_energy():
...
-    # Assert that the second spin bin is only flagged for high energy for the last
-    # energy range
-    assert quality_flags[1] == 2**3
-    # Assert that the third spin bin is not flagged for any energy range
-    assert quality_flags[2] == 0
+    assert quality_flags[0, 0]
+    assert not quality_flags[1, 0]
+    assert quality_flags[2, 0]
+    assert quality_flags[3, 0]
+    # # Assert that the second spin bin is only flagged for high energy for the last
```
laspsandoval: Double comments here.

imap_processing/ultra/l1b/extendedspin.py
```diff
@@ -75,27 +75,32 @@ def calculate_extendedspin(
...
+        de_dataset,
         spin_tbin_edges,
         energy_ranges,
-        energy_bin_flags,
+        voltage_qf,
         energy_thresholds,
         instrument_id,
     )
+    # Combine high energy and voltage flags to use for statistical outlier flagging.
+    mask = (
```
laspsandoval, line 94: I think this should be voltage_qf[np.newaxis, :] | high_energy_qf 

Flag for voltage or high energy

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -710,48 +708,250 @@ def flag_high_energy(
...
+    convergence = np.full(n_energy_bins, False)
+    # Keep track of how many iterations we have done of flagging outliers and
+    # recalculating stats per energy bin
+    iterations = np.zeros(n_energy_bins)
+    # keep track of the standard deviation difference from poisson stats per energy bin
+    std_diff = np.zeros(n_energy_bins, dtype=float)
+    count_summary = get_valid_de_count_summary(
+        de_dataset, energy_ranges, spin_tbin_edges, sensor_id
+    )  # shape (n_energy_bins, n_spin_bins)
+    for e_idx in np.arange(n_energy_bins):  # shape: (n_energy_bins,)
```
laspsandoval: range(n_energy_bins) instead.

imap_processing/ultra/l1b/ultra_l1b_culling.py
```diff
@@ -710,48 +708,250 @@ def flag_high_energy(
...
+        for it in range(n_iterations):
+            # only consider bins that are currently unflagged for this energy bin
+            counts = count_summary[e_idx, ~iter_mask[e_idx]]
+            # Step 1. check if any energy bins have less than 3 spin bins with counts.
+            # If so, flag all spins for that energy bin and skip to the next iteration
```
laspsandoval, line 835: Should you put this check before the other call to get_poisson_states(counts) as well as keeping this one?

laspsandoval (commented):
Some comments and questions. Nicely done! It looks good.

imap_processing/ultra/l1b/extendedspin.py
lacoak21: Ah nice good catch

imap_processing/ultra/l1b/ultra_l1b_culling.py
lacoak21: I asked about that too but its not necessary because if they didnt have enough counts then they would converged would be marked true and there is a check below that only recalculates it if it is unconverged.

imap_processing/ultra/l1b/ultra_l1b_culling.py
lacoak21: Good point. It might

laspsandoval (approved):
Thanks Luisa! Nice work.

PR #2790, opened 2026-03-02
IDEX l2b epoch job failure 

lacoak21:
# Change Summary

## Overview

I noticed that IDEX l2b job failed due to epoch not increasing monotonically. This was due to the input files not being sorted by date and also using np.unique() to determine the unique list of DOYs. Np.unique also sorts the array which I did not know. 

## File changes

I made sure to pass in the housekeeping and science datasets sorted by date as well as keeping those lists preserved throughout the code. I could have just sorted the arrays at the end by epoch but this helps with debugging to have everything sorted nicely.


laspsandoval (approved):
Looks good.

tmplummer (approved):
LGTM.

PR #2791, opened 2026-03-03
ULTRA l1b stat cull validation test

lacoak21:
# Change Summary

## Overview

Add a validation test for the statistical outlier culling algorithm. 

## File changes

During the validation testing I realized I was modifying the std_diff array when recalculating the convergence after combining the flags. So that is the only part of the actual code that I had to update.


tmplummer (approved):
LGTM

PR #2817, opened 2026-03-06
ULTRA add code for ultra90 efficiencies

lacoak21:
# Change Summary

## Overview

Very simple PR to use efficiencies for ULTRA90. Before this, the ULTRA team instructed us to use 45 efficiencies for both because they hadn't gotten around to making them for 90 yet. Nick sent me new efficiency ancillary files for both 90 and 45 yesterday. I will be uploading those as soon as the goodtimes/culling is set to get pushed to prod. 

## Testing
Fix tests.


tech3371 (approved):
LGTM

PR #2818, opened 2026-03-06
2815 bug   hi l1c goodtimes

### LINKED ISSUE #2815: BUG - Hi L1C Goodtimes
### Description of the issue

Forgot to apply TOF window filters to find qualified events in statistical filters.

Use code in Hi PSET to provide filter, relocating the code if needed such that it is common between the two.

_No response_

_No response_

```python
Code
```

_No response_

tmplummer:
## Summary                                                                                                                                                                                                                                                       
                                                          
  Fixes a bug where statistical filters 1 and 2 in hi_goodtimes.py only checked coincidence_type when identifying "qualified" events. They now also check TOF window values, matching the logic in hi_l1c.py:pset_counts().                                     
                                                                                                                                                                                                                                                                
## Changes                                                                                                                                                                                                                                                       
                                                                             
  - Bug fix: Events must pass BOTH coincidence_type AND TOF window checks to be "qualified"
  - Bug fix: Changed dimension name from "event" to "event_met" in mark_statistical_filter_2()
  - Refactor: Added shared utility functions (iter_qualified_events_by_config, compute_qualified_event_mask, etc.) to utils.py, removing duplicate code between hi_l1c.py and hi_goodtimes.py
  - Tests: Added test_only_qualified_events_contribute_to_clusters to verify filtering behavior


Closes: #2815

### DIFF
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -232,13 +232,24 @@ def _apply_goodtimes_filters(
 
     # Pre-compute qualified event masks for each dataset
     # These masks check BOTH coincidence_type AND TOF windows
-    qualified_masks: dict[int, np.ndarray] = {}
-    for i, l1b_de in enumerate(l1b_de_datasets):
+    for l1b_de in l1b_de_datasets:
         ccsds_index = l1b_de["ccsds_index"].values
-        esa_energy_steps = l1b_de["esa_energy_step"].values[ccsds_index]
 
-        qualified_masks[i] = compute_qualified_event_mask(
-            l1b_de, cal_product_config, esa_energy_steps
+        # Handle invalid events (FILLVAL trigger_id) to avoid IndexError
+        # For pointings with no valid events, trigger_id will be at FILLVAL
+        trigger_id_fillval = l1b_de["trigger_id"].attrs.get("FILLVAL", 65535)
+        valid_events = l1b_de["trigger_id"].values != trigger_id_fillval
+
+        # Initialize with -1 (won't match any config row since ESA energy steps > 0)
+        esa_energy_steps = np.full(len(ccsds_index), -1, dtype=np.int32)
+        if np.any(valid_events):
+            esa_energy_steps[valid_events] = l1b_de["esa_energy_step"].values[
+                ccsds_index[valid_events]
+            ]
+
+        l1b_de["qualified_mask"] = xr.DataArray(
+            compute_qualified_event_mask(l1b_de, cal_product_config, esa_energy_steps),
+            dims=["event_met"],
         )
     logger.info("Pre-computed qualified event masks for all datasets")
 
@@ -266,7 +277,6 @@ def _apply_goodtimes_filters(
         goodtimes_ds,
         l1b_de_datasets,
         current_index,
-        qualified_masks,
     )
 
     # 6. Statistical Filter 2 - short-lived event pulses
@@ -274,7 +284,6 @@ def _apply_goodtimes_filters(
     mark_statistical_filter_2(
         goodtimes_ds,
         current_l1b_de,
-        qualified_masks[current_index],
     )
 
 
@@ -1467,7 +1476,6 @@ def _compute_qualified_counts_per_sweep(
 
 def _build_per_sweep_datasets(
     l1b_de_datasets: list[xr.Dataset],
-    qualified_masks: dict[int, np.ndarray],
 ) -> dict[int, xr.Dataset]:
     """
     Build per-sweep datasets with qualified counts for each Pointing.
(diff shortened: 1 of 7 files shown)


imap_processing/hi/hi_goodtimes.py
```diff
@@ -1469,8 +1476,9 @@ def _build_per_sweep_datasets(
     ----------
     l1b_de_datasets : list[xarray.Dataset]
         List of L1B DE datasets for multiple Pointings.
-    qualified_coincidence_types : set[int]
-        Set of coincidence type integers that qualify for calibration products.
+    qualified_masks : dict[int, np.ndarray]
+        Dictionary mapping dataset index to boolean mask indicating which events
+        qualify for calibration products (checking both coincidence_type AND TOF).
```
tech3371: the `qualified_masks : dict[int, np.ndarray]` and `qualified_mask: np.ndarray` looks same but looks like different data type. Can you rename one of them?

imap_processing/hi/hi_l1c.py
```diff
@@ -369,106 +368,37 @@ def pset_counts(
...
-        # esa_energy_step is recorded for each packet rather than for each DE,
-        # so we use ccsds_index to get the esa_energy_step for each DE
-        esa_mask = (
-            l1b_de_dataset["esa_energy_step"].data[de_ds["ccsds_index"].data]
-            == esa_energy
+    # esa energy step combination. Use the shared generator to iterate over all
+    # config combinations and get qualified event masks.
+    for esa_energy, config_row, qualified_mask in iter_qualified_events_by_config(
+        de_ds, config_df, esa_energy_steps
+    ):
```
tech3371, line 380: Can you take this out `iter_qualified_events_by_config(
        de_ds, config_df, esa_energy_steps
    )` and assign to a variable and then use in the for loop? It may help with performance.

imap_processing/hi/utils.py
```diff
@@ -571,3 +573,251 @@ def calibration_product_numbers(self) -> npt.NDArray[np.int_]:
...
+        Fill values for each TOF field - events with fill values pass the check.
+        If not provided, fill value handling is disabled.
+
+    Returns
+    -------
+    mask : numpy.ndarray
+        Boolean mask where True = event passes all specified TOF window checks.
+    """
+    if tof_fill_vals is None:
+        tof_fill_vals = {}
```
tech3371: I don't know enough here. So sorry if I get this wrong. You are not returning at here because you can get mask with (low, high) bit operation?

imap_processing/hi/utils.py
```diff
@@ -571,3 +573,251 @@ def calibration_product_numbers(self) -> npt.NDArray[np.int_]:
...
+    Returns
+    -------
+    dict[str, float]
+        Dictionary mapping TOF variable names to their fill values.
+    """
+    tof_fill_vals = {}
+    for pair in CalibrationProductConfig.tof_detector_pairs:
+        tof_var = f"tof_{pair}"
+        if tof_var in de_ds:
+            tof_fill_vals[tof_var] = de_ds[tof_var].attrs.get("FILLVAL", np.nan)
```
tech3371: Just curious. Why is there case when tof_var is not in the DE? is that because you can not detect some of those pairs sometimes in a day?

imap_processing/hi/utils.py
```diff
@@ -571,3 +573,251 @@ def calibration_product_numbers(self) -> npt.NDArray[np.int_]:
...
+    Yields
+    ------
+    esa_energy : Any
+        The ESA energy step value.
+    config_row : namedtuple
+        The config row from itertuples() containing calibration product settings.
+    qualified_mask : np.ndarray
+        Boolean mask where True = event qualifies for this (esa, cal_prod).
+    """
+    n_events = len(de_ds["event_met"]) if "event_met" in de_ds.dims else 0
```
tech3371, line 742: Is there reason to continue if no event is found?

imap_processing/hi/utils.py
```diff
@@ -571,3 +573,251 @@ def calibration_product_numbers(self) -> npt.NDArray[np.int_]:
...
+    # Build TOF fill values from dataset attributes
+    tof_fill_vals = _build_tof_fill_vals(de_ds)
+
+    for esa_energy, esa_df in cal_product_config.groupby(level="esa_energy_step"):
+        # Mask for events at this ESA energy step
+        esa_mask = esa_energy_steps == esa_energy if n_events > 0 else np.array([])
+
+        for config_row in esa_df.itertuples():
+            if n_events == 0 or not np.any(esa_mask):
+                yield esa_energy, config_row, np.zeros(n_events, dtype=bool)
```
tech3371, line 753: I haven't used yield enough. I read that it's returning to the caller function if no event_was found or esa_mask. Is that true?

tech3371: Now I am wondering if you called `iter_qualified_events_by_config()` function in the for loop for this reason.

tech3371 (approved):
I haven't found anything major. LGTM!

imap_processing/hi/hi_goodtimes.py
tmplummer: Yep, I just added the qualified_mask directly to the l1b_de datasets instead of carrying them around separately.

imap_processing/hi/utils.py
tmplummer: Good catch. It should always be present and if it were missing, this should raise an error.

imap_processing/hi/utils.py
tmplummer: Yes, it is intended to be used in a for loop.

PR #2820, opened 2026-03-09
ULTRA l1c fix exposure time calculation

lacoak21:
# Change Summary

## Overview

When I produced the first l2 maps that included the new culling update, the ULTRA team noticed that the exposure looked too high and I realized that I had not been excluding spins that were flagged from the exposure time. I reproduced maps using this branch and they confirmed that this was the necessary fix. 

## File changes

   - vectorized version of energy dependent event culling 
   - Calculate exposure time using the correct number of valid spins in the pointing. 

## Testing
Add a test to check that the correct spins were used in calculating the exposure factor.

### DIFF
--- a/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
+++ b/imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
@@ -491,14 +491,14 @@ def get_spacecraft_exposure_times(
     valid_spins = (
         np.bitwise_or.reduce(flag_arrays)[np.newaxis, :] & energy_flags[:, np.newaxis]
     ) == 0
-    # pad valid_spins with arrays of all true at either end if energy bins are outside
-    # the range of the goodtimes dataset energy edges
-    # If the energy bin is below or above that the ranges, that just means those
-    # ranges were not included in the quality flag filtering in the goodtimes dataset.
+    # Pad valid_spins with arrays of all true at either end to account for energy bins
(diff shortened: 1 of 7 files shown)


imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
```diff
@@ -464,35 +469,60 @@ def get_spacecraft_exposure_times(
...
     ]
-    # Get only valid spin data
-    valid_mask = (spin_data["spin_phase_valid"].values == 1) & (
-        spin_data["spin_period_valid"].values == 1
+    bin_to_range = np.digitize(energy_bins, goodtimes_dataset.energy_range_edges)
+    valid_spins = (
+        np.bitwise_or.reduce(flag_arrays)[np.newaxis, :] & energy_flags[:, np.newaxis]
+    ) == 0
+    # pad valid_spins with arrays of all true at either end if energy bins are outside
+    # the range of the goodtimes dataset energy edges
```
tmplummer: I'm confused by this statement indicating that padding only occurs "if energy bins are outside the range of the goodtimes dataset energy edges" but then it looks like padding always occurs.

tmplummer (approved):
Just the one question about the comment vs. implementation.

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
lacoak21: Ah yes that is confusing. Ill clear that up.

PR #2827, opened 2026-03-09
Change CDF output variables to UTC string, matching algorithm document

maxinelasp:
# Change Summary

## Overview

Converts `start_time` and `end_time` in the GLOWS L2 CDF output from raw J2000
nanoseconds to UTC strings, matching the format specified in the algorithm
document. Fixes a CDF writing error caused by mismatched attribute types.

## File changes

- **`imap_processing/glows/l2/glows_l2.py`**: Added `utc_time_variables` list
  (`start_time`, `end_time`) and a conversion step in `create_l2_dataset` that
  calls `met_to_utc(ttj2000ns_to_met(value))` before writing these fields to
  the output dataset.

- **`imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml`**: Updated
  `time_data_defaults` to use string-compatible CDF attributes (`FILLVAL: ' '`,
  `FORMAT: A30`, `VALIDMIN/VALIDMAX: ' '`). The previous `FILLVAL: 0` and
  `FORMAT: I1000` caused a cdflib `ValueError` when writing string data because
  cdflib infers CDF_CHAR type from the string data but then fails to convert an
  integer FILLVAL to that type.

### DIFF
--- a/imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_glows_l2_variable_attrs.yaml
@@ -40,11 +40,11 @@ support_data_defaults: &support_data_defaults
 
 time_data_defaults: &time_data_defaults
   <<: *support_data_defaults
-  FILLVAL: 0
-  FORMAT: I1000
+  FILLVAL: ' '
+  FORMAT: A30
   UNITS: N/A
-  VALIDMIN: 0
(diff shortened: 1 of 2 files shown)


sdhoyt (approved):

maxinelasp:
.pre-commit autofix

PR #2831, opened 2026-03-10
Quicklooks doc

anamanica:
# Change Summary

Adding quicklooks documentation in preparation for instrument teams needing to give an outline of quicklook information in addition to me leaving at the beginning of May.

## Overview

Added quicklook.rst in the docs folder within the algorithm code documentation sub directory. 

## File changes

quicklooks.rst
index.rst (In same file path)

## Testing

None Needed


docs/source/algorithm-code-documentation/index.rst
```diff
@@ -25,6 +25,7 @@ Instruments
    swapi
    swe
    ultra
+   quicklooks
```
vmartinez-cu, line 28: I wonder if this should have it's own heading or be listed under `Utilities` instead of under the `Instruments` section since it's not an instrument. @tech3371 what do you think?

vmartinez-cu (approved):
Nice job with this! The documentation is clearly written and nicely formatted.

docs/source/algorithm-code-documentation/index.rst
tech3371: Agree. Under `Utilities` section is better.

tech3371:
@anamanica in case you are, you don't need to wait for my approval. Veronica approval is enough!

PR #2835, opened 2026-03-11
2822 hi l1b goodtimes   implement bad tdc cal culling algorithm

### LINKED ISSUE #2822: Hi L1B Goodtimes - implement bad TDC cal culling algorithm
### Algorithm Description:

This should be the last goodtime filter to implement. Use the example C code from Paul to identify when the TDC cal fails using the diag_fee packet?

tmplummer:
## Summary
                                                                                                                                                                                                                                                                
  - Implement mark_bad_tdc_cal culling algorithm - Add new goodtimes filter that removes times with failed TDC calibration based on DIAG_FEE packet data
  - Integrate L1A DIAG_FEE dependency - Add l1a_diagfee parameter to hi_goodtimes() and _apply_goodtimes_filters() functions
  - Update CLI to load DIAG_FEE data - Add file path retrieval and CDF loading for L1A DIAG_FEE in the Hi goodtimes processing path
  - Reorder hi_goodtimes parameters - Move current_repointing to first positional argument for consistency

##  Details

  mark_bad_tdc_cal Algorithm

  Based on C reference (drop_bad_tdc_diagfee in culling.c):
  - Scans DIAG_FEE packets chronologically and checks TDC1/TDC2/TDC3 calibration status (bit 1 = success)
  - If any checked TDC has failed calibration, marks all times from that DIAG_FEE packet until the next as bad
  - Skips duplicate DIAG_FEE packets that appear within 10 seconds (quirk when entering HVSCI mode)
  - Requires at least 2 DIAG_FEE packets to perform checking

  Filter Order Update

  Goodtimes filters now apply in order:
  1. mark_incomplete_spin_sets
  2. mark_drf_times
  3. mark_bad_tdc_cal (new)
  4. mark_overflow_packets
  5. mark_statistical_filter_0
  6. mark_statistical_filter_1
  7. mark_statistical_filter_2


Closes: #2822

### DIFF
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -1212,7 +1212,10 @@ def mark_bad_tdc_cal(
         # Remove times from this DIAG_FEE packet until next. We are skipping the
         # first packet of a duplicate pair, so determining the window based on the
         # current packet met and next packet met covers the time window between
-        # non-duplicate DIAG_FEE packets.
+        # non-duplicate DIAG_FEE packets. We can ignore the ~10 seconds of slop
+        # around duplicate packets because these packets should only be produced
+        # when IMAP-Hi is transitioning to HVSCI mode which means that there will
+        # be no DE packets being produced.
         df_time = diagfee_met[i]
         next_df_time = diagfee_met[i + 1] if i < len(diagfee_met) - 1 else np.inf
 
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -1208,7 +1208,7 @@ def mark_bad_tdc_cal(
     tdc_failed_indices = np.nonzero(~is_duplicate & tdc_failed)[0]
 
     n_times_removed = 0
-    for i in failed_indices:
+    for i in tdc_failed_indices:
         # Remove times from this DIAG_FEE packet until next. We are skipping the
         # first packet of a duplicate pair, so determining the window based on the
         # current packet met and next packet met covers the time window between
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -211,7 +211,7 @@ def _apply_goodtimes_filters(
     """
     Apply all goodtimes culling filters to the dataset.
 
-    Modifies goodtimes_ds in place by applying filters 1-6.
+    Modifies goodtimes_ds in place by applying filters 1-7.
 
     Parameters
     ----------
@@ -1191,7 +1191,10 @@ def mark_bad_tdc_cal(
     # Based on sample code in culling_v2.c, skip this check if we have fewer
     # than two diag_fee packets.
     if len(diagfee.epoch) < 2:
-        logger.warning("No DIAG_FEE data to use for selecting good times")
+        logger.warning(
+            f"Insufficient DIAG_FEE packets to select good times "
+            f"(found {len(diagfee.epoch)}, need at least 2)"
+        )
         return
 
     df_met = diagfee["shcoarse"].values
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -1145,20 +1145,18 @@ def mark_overflow_packets(
 def mark_bad_tdc_cal(
     goodtimes_ds: xr.Dataset,
     diagfee: xr.Dataset,
-    check_tdc1: bool = True,
-    check_tdc2: bool = True,
-    check_tdc3: bool = True,
     cull_code: int = CullCode.LOOSE,
 ) -> None:
     """
     Remove times with failed TDC calibration (DIAG_FEE method).
 
-    Based on C reference: drop_bad_tdc_diagfee in culling.c
+    Based on C reference: drop_bad_tdc_diagfee in culling_v2.c provided by
+    IMAP-Hi team.
 
     This function scans DIAG_FEE packets chronologically and checks the TDC
-    calibration status for each packet. If any checked TDC has failed
-    calibration, all times from that DIAG_FEE packet until the next DIAG_FEE
-    packet are marked as bad.
+    calibration status for each packet. If any TDC has failed calibration,
+    all times from that DIAG_FEE packet until the next DIAG_FEE packet are
+    marked as bad.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


imap_processing/hi/hi_goodtimes.py
```diff
@@ -1130,6 +1142,95 @@ def mark_overflow_packets(
...
+    diagfee: xr.Dataset,
+    check_tdc1: bool = True,
+    check_tdc2: bool = True,
+    check_tdc3: bool = True,
+    cull_code: int = CullCode.LOOSE,
+) -> None:
+    """
+    Remove times with failed TDC calibration (DIAG_FEE method).
+
+    Based on C reference: drop_bad_tdc_diagfee in culling.c
```
tech3371: I am guessing it's referencing c code of Paul's. Is there link or reference or can you add context that it's from Paul may be helpful for who to ask in the future.

imap_processing/hi/hi_goodtimes.py
```diff
@@ -1130,6 +1142,95 @@ def mark_overflow_packets(
...
+        - shcoarse: Mission Elapsed Time (MET)
+        - tdc1_cal_ctrl_stat: TDC1 calibration status (bit 1 = success)
+        - tdc2_cal_ctrl_stat: TDC2 calibration status (bit 1 = success)
+        - tdc3_cal_ctrl_stat: TDC3 calibration status (bit 1 = success)
+    check_tdc1 : bool, optional
+        Whether to check TDC1 calibration status. Default is True.
+    check_tdc2 : bool, optional
+        Whether to check TDC2 calibration status. Default is True.
+    check_tdc3 : bool, optional
+        Whether to check TDC3 calibration status. Default is True.
```
tech3371: I am guessing these will be disabled in the future as needed? What will cause these to set to false? Just curious.

imap_processing/hi/hi_goodtimes.py
```diff
@@ -1130,6 +1142,95 @@ def mark_overflow_packets(
...
+    cull_code : int, optional
+        Cull code to use for marking bad times. Default is CullCode.LOOSE.
+
+    Notes
+    -----
+    This function modifies goodtimes_ds in place.
+
+    Quirk: Two DIAG_FEE packets are generated when entering HVSCI mode.
+    The first packet is skipped if two packets appear within 10 seconds.
```
tech3371, line 1179: what 'two packets'? science/DE packets or DIAG_FEE packets?

tech3371: Nvm, looks like two DIAG_FEE packets.

imap_processing/hi/hi_goodtimes.py
```diff
@@ -1130,6 +1142,95 @@ def mark_overflow_packets(
...
+    # Based on sample code in culling_v2.c, skip this check if we have fewer
+    # than two diag_fee packets.
+    if len(diagfee.epoch) < 2:
+        logger.warning(
+            f"Insufficient DIAG_FEE packets to select good times "
+            f"(found {len(diagfee.epoch)}, need at least 2)"
+        )
+        return
+
+    df_met = diagfee["shcoarse"].values
```
tech3371: could you get more than two diag_fee packets? Just curious based on above check and comments.

imap_processing/hi/hi_goodtimes.py
```diff
@@ -1130,6 +1142,95 @@ def mark_overflow_packets(
...
+    met_values = goodtimes_ds.coords["met"].values
+    n_times_removed = 0
+
+    for i in range(len(df_met)):
+        # Skip duplicate DIAG_FEE packets (within 10 seconds of next)
+        if i < len(df_met) - 1 and df_met[i] + 10 > df_met[i + 1]:
+            continue
```
tech3371: Is there way to look for duplicate through different way, eg. looking at shcoarse and something? I am concerned that what if next ten are always not duplicate in the future.

imap_processing/hi/hi_goodtimes.py
```diff
@@ -1130,6 +1142,95 @@ def mark_overflow_packets(
...
+
+        # Check TDC calibration status (bit 1: 1=good, 0=bad)
+        any_tdc_failed = False
+
+        if check_tdc1 and (diagfee["tdc1_cal_ctrl_stat"].values[i] & 2) == 0:
```
tech3371: Curious why you are `&` with 2. Can you add comment about it?

imap_processing/hi/hi_goodtimes.py
```diff
@@ -1130,6 +1142,95 @@ def mark_overflow_packets(
...
+        if check_tdc3 and (diagfee["tdc3_cal_ctrl_stat"].values[i] & 2) == 0:
+            any_tdc_failed = True
+
+        if any_tdc_failed:
+            # Remove times from this DIAG_FEE until next
+            df_time = df_met[i]
+            next_df_time = df_met[i + 1] if i < len(df_met) - 1 else np.inf
+
+            in_window = (met_values >= df_time) & (met_values < next_df_time)
+            mets_to_cull = met_values[in_window]
```
tech3371: Can you add comment about why you are culling the window this way and not using the +10 duplicate packets? or algorithm document reference is ok too.

imap_processing/hi/hi_goodtimes.py
tmplummer: These were options provided in the example C code. I suspect that they will not ever get used and will always default to True.

imap_processing/hi/hi_goodtimes.py
tmplummer: Yep, we should normally get exactly two. diagfee is an xr.Dataset so this is getting all `shcoarse` values which will normally be an array of length 2.

imap_processing/hi/hi_goodtimes.py
tmplummer: I refactored to vectorize this function as much as possible. AFAIK, this is the only/best way to identify "duplicates" which are not true duplicate packets because they have unique shcoarse values and src_seq_cntr.

imap_processing/hi/hi_goodtimes.py
tmplummer: I removed these options.

imap_processing/hi/hi_goodtimes.py
tmplummer: This is a binary and operation. I think that the comment "(bit 1: 1=good, 0=bad)" explains it?

imap_processing/hi/hi_goodtimes.py
```diff
@@ -1130,6 +1142,93 @@ def mark_overflow_packets(
...
+    # Identify any packets where any of the three TDC calibrations failed.
+    # TDC failure check (bit 1: 1=good, 0=bad)
+    tdc_failed = (
+        ((diagfee["tdc1_cal_ctrl_stat"].values & 2) == 0)
+        | ((diagfee["tdc2_cal_ctrl_stat"].values & 2) == 0)
+        | ((diagfee["tdc3_cal_ctrl_stat"].values & 2) == 0)
+    )
+
+    # Only loop over non-duplicate packets with TDC failures
+    failed_indices = np.nonzero(~is_duplicate & tdc_failed)[0]
```
tech3371: I do like this approach much better. Sorry, one minor rename of variable to match the comment.
```suggestion
    tdc_failed_indices = np.nonzero(~is_duplicate & tdc_failed)[0]
```

tech3371 (approved):
Thank you for refactoring. It's easier to follow!

PR #2850, opened 2026-03-17
MNT: update numpy version, fix mypy errors

maxinelasp:
# Change Summary
These changes address mypy errors from numpy version updates, which is required because Python 3.14 which needs Numpy 2.

** This is almost entirely Claude generated! ** I don't have a good understanding of instruments so please review the code you feel confident in understanding. This PR only affects type-hints, so it should have a minimal effect.

I have carefully looked through the changes and don't see anything concerning.


imap_processing/ena_maps/utils/map_utils.py
```diff
@@ -77,7 +77,7 @@ def vectorized_bincount(
     # dimension by an integer multiple of the number of bins. Doing so gives
     # each element in the additional dimensions its own set of 1D bins: index 0
     # uses bins [0, minlength), index 1 uses bins [minlength, 2*minlength), etc.
-    offsets = np.arange(n_binsets).reshape(*non_spatial_shape, 1) * minlength
+    offsets: NDArray = np.arange(n_binsets).reshape(*non_spatial_shape, 1) * minlength
```
tech3371, line 80: I am curious why it picked NDArray vs np.ndarray. Should we be consistent?

tech3371 (approved):
looks good. Thank you for doing this for everyone!

imap_processing/ena_maps/utils/map_utils.py
tech3371: I merged the PR since we need to bypass codecov for this PR...but if you like to make change for this small one, feel free to open small PR.

Issue #2720, opened 2026-02-18
BUG - Spin angle should not be averaged across L1B blocks in GLOWS L2


maxinelasp:
Parent issue: #1276

In `DailyLightcurve.__post_init__` (`glows_l2_data.py:86`):

```python
self.spin_angle = np.average(l1b_data["imap_spin_angle_bin_cntr"].data, axis=0)
```

The bin center angles are fixed for the observation day (same 3600 bins every block). Averaging them is unnecessary and potentially misleading. Use the first (or any single) timestamp's values directly instead of averaging across all timestamps.

Additionally, per the algorithm document (Section 10.6), `position_angle_offset_average` should be set to `ψ_G,eff = 360° - ψ_GLOWS` (a single computed value, not an average over the day), and `position_angle_offset_std_dev` should always be 0. The current L1B code (`glows_l1b_data.py:855-857`) correctly sets std_dev to 0, but the L2 code (`glows_l2.py:121-127`) re-averages all `*_average` fields including `position_angle_offset_average`, which is incorrect. The position angle offset should be passed through from L1B unchanged, not re-averaged.

maxinelasp:
This is closely related to #2719 and proably should be completed at the same time. I'm already working on #2719, so I can take this ticket.

Issue #2719, opened 2026-02-18
BUG - Bad angle formatting in GLOWS L2


maxinelasp:
Parent issue: #1276

The `spin_angle` variable in `DailyLightcurve` (`glows_l2_data.py:86`) is computed by averaging `imap_spin_angle_bin_cntr` across all L1B timestamps. This produces IMAP spin angles, but L2 should use the **GLOWS position angle** (ψ_PA), which is obtained by converting from the IMAP spin angle using the position angle offset (Eq. 29/30 in the algorithm doc, Section 10.6).

Fix `spin_angle` in `DailyLightcurve` to store GLOWS position angles rather than raw IMAP spin angles. The conversion subtracts the GLOWS effective angle and applies modulo 360.

PR #2852, opened 2026-03-18
Glows l2 spin angle

maxinelasp:
## Overview

Implements spin angle and position angle offset calculations in the GLOWS L2 histogram processing pipeline. The IMAP spin angle (ψ) stored at L1B is converted to the GLOWS position angle (ψPA, Eq. 29 of the algorithm document), which measures rotation from the northernmost point of the GLOWS scanning circle. All bin-dimensioned lightcurve arrays are rolled so that bin 0 corresponds to the northernmost point.

This also includes some fixes for bin number management. 

## File changes

- glows/l2/glows_l2_data.py: Adds compute_position_angle() to HistogramL2, which implements Eq. 30 (ψG,eff = (360 − ψGLOWS) % 360) using the GLOWS boresight azimuth from the SPICE frames kernel. Updates DailyLightcurve to accept position_angle as an InitVar, apply the ψ → ψPA conversion (Eq. 29), and roll all bin arrays to north-first order. Corrects position_angle_offset_average and position_angle_offset_std_dev to scalar values (one per L2 file, not per L1B block), consistent with the algorithm document's statement that std_dev is always zero and the offset is a fixed instrument property.

- glows/l2/glows_l2.py: Minor fixes carried in from prior work — uses GlowsConstants.STANDARD_BIN_COUNT for the bins coordinate and pads bin arrays to standard length with CDF FILLVALs before writing to the output dataset.

Some comments and unit tests were updated with Claude. The unit tests were generated against the validation code provided by GLOWS.

Testing
3 new unit tests added for spin angle and position angle behavior: test_spin_angle_offset_formula, test_spin_angle_starts_at_minimum, and test_compute_position_angle. 1 existing test (test_position_angle_offset_average) updated to reflect the corrected scalar type. Other tests updated to fix SPICE issues.

Completes #2720 and #2719


imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -107,6 +127,28 @@ def __post_init__(self, l1b_data: xr.Dataset) -> None:
...
+                : self.number_of_bins
+            ]
+            # Convert ψ → ψPA (Eq. 29): position angle measured from the
+            # northernmost point of the scanning circle.
+            self.spin_angle = (spin_angle_bin_cntr - position_angle + 360.0) % 360.0
+
+            # Roll all bin arrays so bin 0 corresponds to the northernmost
+            # point (minimum ψPA).
+            roll = -np.argsort(self.spin_angle)[0]
+            self.spin_angle = np.roll(self.spin_angle, roll)
```
vmartinez-cu, line 141: Interesting, I thought spin_angle would already correspond to the northernmost point when it gets defined above, but I guess rolling the bins here ensures that the smallest angle is the start

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -213,8 +256,8 @@ class HistogramL2:
     pulse_length_std_dev: np.ndarray[np.double]
     spin_period_ground_average: np.ndarray[np.double]
     spin_period_ground_std_dev: np.ndarray[np.double]
-    position_angle_offset_average: np.ndarray[np.double]
-    position_angle_offset_std_dev: np.ndarray[np.double]
+    position_angle_offset_average: np.double
+    position_angle_offset_std_dev: np.double
```
vmartinez-cu, line 258: I was wondering if this should be a single value. This is helpful. Thanks for the update!

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -392,3 +431,23 @@ def return_good_times(flags: xr.DataArray, active_flags: NDArray) -> NDArray:
         # where all the active indices == 1.
         good_times = np.where(np.all(flags[:, active_flags == 1] == 1, axis=1))[0]
         return good_times
+
+    def compute_position_angle(self) -> float:
```
vmartinez-cu, line 433: Nice addition of this method - the result will be helpful for computing ecliptic coords

imap_processing/tests/glows/test_glows_l2_data.py
```diff
@@ -228,3 +207,108 @@ def test_filter_good_times():
...
+    target_module = (
+        "imap_processing.glows.l2.glows_l2_data.get_instrument_mounting_az_el"
+    )
+    with patch(target_module, return_value=(270.0, 0.0)):
+        result = HistogramL2.compute_position_angle(None)
+    assert result == pytest.approx(90.0)
+
+
+@pytest.fixture
+def l1b_dataset_full():
```
vmartinez-cu, line 253: could this fixture replace the other minimal l1b_dataset fixture?

vmartinez-cu (approved):
Nice job! Some of this will be helpful for the ticket I'm working on

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -107,6 +127,28 @@ def __post_init__(self, l1b_data: xr.Dataset) -> None:
...
+            spin_angle_bin_cntr = l1b_data["imap_spin_angle_bin_cntr"].data[0][
+                : self.number_of_bins
+            ]
+            # Convert ψ → ψPA (Eq. 29): position angle measured from the
+            # northernmost point of the scanning circle.
+            self.spin_angle = (spin_angle_bin_cntr - position_angle + 360.0) % 360.0
+
+            # Roll all bin arrays so bin 0 corresponds to the northernmost
+            # point (minimum ψPA).
+            roll = -np.argsort(self.spin_angle)[0]
```
tmplummer: This would avoid doing a full sort:
```suggestion
            roll = -np.argmin(self.spin_angle)
```

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -64,10 +66,29 @@ def __post_init__(self, l1b_data: xr.Dataset) -> None:
         l1b_data : xarray.Dataset
             L1B data filtered by good times, good angles, and good bins for one
             observation day.
+        position_angle : float
+            The offset angle of the GLOWS instrument from the spin plane - this is used
+            in spin angle calculations. This number may change with different SPICE
+            kernels, but does not vary by time inside the code, so we can just use a
+            fixed value.
```
tmplummer: This description does not make sense for me when I look at the equation where position_angle is used. Based on this comment, I think you are referring to the mounting elevation angle but in the equation it is being used as if it is the mounting azimuth angle.

I don't think that this value will ever change with SPICE kernel.

tmplummer (approved):
A couple of suggestions, but looks good to me.

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: I'll take a look at the comment. I found the document pretty confusing but the code should be correct (based off of the GLOWS Python code) so I think your point about the mounting azimuth angle makes sense. 

The algorithm document mentioned that the value may change - but I don't want to be confusing so I'll take out that note.

imap_processing/tests/glows/test_glows_l2_data.py
maxinelasp: This fixture is only used in one test, and it's much more than is needed for most of the tests and will be annoying to update, so I think I'll leave it. Good thing to call out though, I appreciate it!

PR #2857, opened 2026-03-23
ULTRA l1c counts nside update

### LINKED ISSUE #2696: ULTRa L1c write counts at a higher nside to combat aliasing
Apparently counts are required to be integral values at l2- this causes weird aliasing counts across pixels in l2 maps when we push counts. We have to push them to the map because they need to be integral values. To fix this, Nick proposed writing counts out at a higher nside and then downsampling them at l2. Im not 100% following this but I think it generally makes sense to me. 

IF we are going from more multiple pixels to one pixels at l2 we will have less edge issues that are caused by these integral counts.

lacoak21:
# Change Summary

closes #2696

## Overview

Counts at l1c need to be binned at a higher resolution and then downsampled at l2. 

## Testing
Add tests to check downsampling.

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -614,17 +614,17 @@ class UltraPointingSet(HealpixPointingSet):
         """
         pset_data = self.data
         counts_n_pix = pset_data.sizes["counts_pixel_index"]
-        if counts_n_pix != hp.nside2npix(self.nside):
+        pset_n_pix = hp.nside2npix(self.nside)
+        if counts_n_pix != pset_n_pix:
             # Raise an error if the nside the counts were sampled at is lower than the
             # nside of the output map. We never want counts to be upsampled.
-            if counts_n_pix < hp.nside2npix(self.nside):
+            if counts_n_pix < pset_n_pix:
                 raise ValueError(
                     f"Counts in the input PSET are sampled at nside "
                     f"{hp.npix2nside(counts_n_pix)}, and the pset is {self.nside}. "
                     f"This would require upsampling the counts, which we do not want."
                 )
             counts_nside = hp.npix2nside(counts_n_pix)
-            pset_n_pix = hp.nside2npix(self.nside)
             n_energy_bins = pset_data.sizes["energy_bin_geometric_mean"]
             order_diff = int(np.log2(counts_nside // self.nside))
             counts = pset_data["counts"].values[
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -663,6 +663,10 @@ class UltraPointingSet(HealpixPointingSet):
                     CoordNames.HEALPIX_INDEX.value,
                 ),
             )
+            logger.info(
+                f"Counts variable with nside = {counts_nside} downsampled to "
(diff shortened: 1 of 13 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -600,6 +602,69 @@ def __repr__(self) -> str:
...
+            # Raise an error if the nside the counts were sampled at is lower than the
+            # nside of the output map. We never want counts to be upsampled.
+            if counts_n_pix < hp.nside2npix(self.nside):
+                raise ValueError(
+                    f"Counts in the input PSET are sampled at nside "
+                    f"{hp.npix2nside(counts_n_pix)}, and the pset is {self.nside}. "
+                    f"This would require upsampling the counts, which we do not want."
+                )
+            counts_nside = hp.npix2nside(counts_n_pix)
+            pset_n_pix = hp.nside2npix(self.nside)
```
tmplummer: This is the third time that `hp.nside2npix(self.nside)` is called. Assign the variable outside of the `if`. Then I think it is cleaner to do:
```
if counts_n_pix < pset_n_pix:
    raise ...
elif counts_n_pix > pset_n_pix:
   # downsample code
```

imap_processing/ena_maps/ena_maps.py
```diff
@@ -600,6 +602,69 @@ def __repr__(self) -> str:
...
+            n_energy_bins = pset_data.sizes["energy_bin_geometric_mean"]
+            downsampled_counts = np.zeros((n_energy_bins, pset_n_pix))
+            order_diff = int(np.log2(counts_nside // self.nside))
+            for i in range(n_energy_bins):
```
tmplummer: I think that this could be done in a vectorized fashion using xarray since the same operation is being done on each energy bin slice of the counts data. Might be worth seeing what Claude comes up with? Just a thought, not critical.

imap_processing/ultra/l1c/helio_pset.py
```diff
@@ -160,12 +161,14 @@ def calculate_helio_pset(
...
-        nside=nside,
+        nside=UltraConstants.L1C_COUNTS_NSIDE,
     )
     helio_pset_quality_flags = np.full(
         n_pix, ImapPSETUltraFlags.NONE.value, dtype=np.uint16
     )
-    healpix = np.arange(n_pix)
+    counts_healpix = np.arange(n_pix)
+    # Determine nside for non "counts" variables from the lookup table
+    healpix = hp.nside2npix(nside)
```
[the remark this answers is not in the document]
tmplummer: +1

tmplummer (commented):
Wow, I am surprised that Ultra is changing something this fundamental this late in the game! Overall, this looks good. I think that Copilot has some valuable feedback in places. I just have a couple of suggestions.

lacoak21:
> Wow, I am surprised that Ultra is changing something this fundamental this late in the game! Overall, this looks good. I think that Copilot has some valuable feedback in places. I just have a couple of suggestions.

Agreed. I met with Tenzin this morning to chat about all the last minute requests coming in 😨. Oops also I realized I had stashed code - sorry for some unfinished stuff

tmplummer (approved):
This looks good to me.

PR #2859, opened 2026-03-23
IDEX l1b cli logic

lacoak21:
# Change Summary

## Overview

Now that IDEX l1b dependencies may have multiple science files pass in from batch starter, we want to make sure that we are only passing in the latest l1a file. 

## File changes
imap_processing/cli.py
- only pass in the latest idex l1a file to l1b. 

## Testing
Add a test to make sure the sorting is working.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -1037,9 +1037,13 @@ class Idex(ProcessInstrument):
                 )
             # get CDF file
             science_files = dependencies.get_file_paths(source="idex")
+            # Load all the science files. There should only be one, but in the case of
+            # multiple files, we want to make sure to load them all and sort them by
+            # time and pass in the latest one.
+            dependencies = [load_cdf(f) for f in science_files]
+            latest_file = sorted(dependencies, key=lambda ds: ds["epoch"].data[0])[-1]
             # process data
-            dependency = load_cdf(science_files[0])
-            datasets = [idex_l1b(dependency)]
+            datasets = [idex_l1b(latest_file)]
         elif self.data_level == "l2a":
             if len(dependency_list) != 3:
                 raise ValueError(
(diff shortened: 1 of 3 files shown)


imap_processing/cli.py
```diff
@@ -1037,9 +1023,13 @@ def do_processing(
                 )
             # get CDF file
             science_files = dependencies.get_file_paths(source="idex")
+            # Load all the science files. There should only be one, but in the case of
+            # multiple files, we want to make sure to load them all and sort them by
+            # time and pass in the latest one.
+            dependencies = [load_cdf(f) for f in science_files]
+            latest_file = sorted(dependencies, key=lambda ds: ds["epoch"].data[0])[-1]
```
tech3371: So we are only using the latest file still?

tech3371 (approved):
I already forgot why we are using latest. Can you add comment about why that is the case?

imap_processing/cli.py
lacoak21: Yes we are only using the latest. This is because I extended the query range for idex l1b jobs to get the correct spice coverage. Unfortunately this means that we now may get multiple idex l1a science files. We still only want the latest one.

imap_processing/hi/hi_l1c.py
```diff
@@ -703,7 +703,7 @@ def good_time_and_phase_mask(
     met_indices = np.clip(met_indices, 0, len(gt_mets) - 1)
 
     # Convert nominal_bins to int32 for indexing
-    spin_bins = nominal_bins.astype(np.int32)
+    spin_bins: npt.NDArray[np.int32] = nominal_bins.astype(np.int32)
```
lacoak21, line 706: @subagonsouth I was dealing with a mypy failure here on this line.

tmplummer (approved):

PR #2862, opened 2026-03-24
Use pset midpoint in map projection at l2

### LINKED ISSUE #2582: BUG - Failure to produce maps due to Pointing Sets epoch misaligning with SPICE data
### Description of the issue

For some repointings, the epoch associated with a pointing set falls before the SPICE coverage window begins for the DPS frame. For example, `imap_ultra_l1c_90sensor-heliopset_20251128-repoint00062_v005.cdf` has an epoch of 2025-11-28T10:02:11.600076, but the coverage window in `imap_dps_2025_332_2025_334_001.ah.bc` only starts at 2025-11-28T10:11:48.599.

Frame transformations for pointing sets are currently calculated at the exact epoch from the pointing set, which can cause the transformation to fail due to not having SPICE coverage. This prevents making maps that include this day of data.

Some options to address this:

1. Make the L1c pointing set epoch be the middle of the data collection window
3. Leave the L1c as is but use the middle of the pointing set time (epoch + epoch_delta/2) when computing transformations

To demonstrate the pointing set epoch falling outside of the SPICE coverage range, the following command can be run with the provided dependency file:

```
imap-cli --instrument ultra --data-level l2 --descriptor u90-ena-h-hf-nsp-full-hae-4deg-3mo --start-date 20251128 --version v000 --dependency imap_ultra_l3_u90-no-spice_20251128_v000.json
```

[imap_ultra_l3_u90-no-spice_20251128_v000.json](https://github.com/user-attachments/files/24647528/imap_ultra_l3_u90-no-spice_20251128_v000.json)

Running the command should result in generating a map containing a single pointing set's worth of data.

Running the command results in failure to generate a map due to the following error:
```
spiceypy.utils.exceptions.SpiceNOFRAMECONNECT: 
================================================================================

Toolkit version: CSPICE_N0067

SPICE(NOFRAMECONNECT) --

At epoch 8.1759620078309E+08 TDB (2025 NOV 28 10:03:20.783 TDB), there is insufficient information available to transform from reference frame -43901 (IMAP_DPS) to reference frame -43924 (IMAP_HAE). IMAP_DPS is a CK frame; a CK file containing data for instrument or structure -43901 at the epoch shown above, as well as a corresponding SCLK kernel, must be loaded in order to use this frame. Frame IMAP_HAE could be transformed to frame 1 (J2000). Failure to find required CK data could be due to one or more CK files not having been loaded, or to the epoch shown above lying within a coverage gap or beyond the coverage bounds of the loaded CK files. It is also possible that no loaded CK file has required angular velocity data for the input epoch, even if a loaded CK does have attitude data for that epoch. You can use CKBRIEF with the -dump option to display coverage intervals of a CK file.

pxform_c --> PXFORM --> REFCHG

================================================================================
```

lacoak21:
# Change Summary
closes #2582

## Overview

Menlo found a bug in which some pset epoch times did not have pointing attitude coverage. This was due to a small conversion error. See #2582 for more info. 

Tim/ menlo suggested using the midpoint ET of the pointing instead of the start. 

## File changes

imap_processing/ena_maps/ena_maps.py
- Use the midpoint time instead of the start of the pointing (epoch) 

## Testing

Add a small test to make sure epoch_delta returns the correct value

### DIFF
--- a/imap_processing/ena_maps/ena_maps.py
+++ b/imap_processing/ena_maps/ena_maps.py
@@ -22,7 +22,7 @@ from imap_processing.ena_maps.utils import map_utils, naming, spatial_utils
 # so we define an enum to handle the coordinate names.
 from imap_processing.ena_maps.utils.coordinates import CoordNames
 from imap_processing.spice import geometry
-from imap_processing.spice.time import ttj2000ns_to_et
+from imap_processing.spice.time import met_to_ttj2000ns, ttj2000ns_to_et
 
 logger = logging.getLogger(__name__)
 
@@ -142,13 +142,9 @@ def match_coords_to_indices(
     # which must be converted to ephemeris time (ET) for SPICE.
     if event_et is None:
         if isinstance(input_object, PointingSet):
-            event_et = ttj2000ns_to_et(
-                input_object.epoch + input_object.epoch_delta / 2
-            )
+            event_et = input_object.midpoint_j2000_et
         elif isinstance(output_object, PointingSet):
-            event_et = ttj2000ns_to_et(
-                output_object.epoch + output_object.epoch_delta / 2
-            )
+            event_et = output_object.midpoint_j2000_et
         else:
             raise ValueError(
                 "Event time must be specified if both objects are SkyMaps."
@@ -307,16 +303,17 @@ class PointingSet(ABC):
         return self.data["epoch"].values[0]
 
     @property
-    def epoch_delta(self) -> int:
+    def midpoint_j2000_et(self) -> float:
         """
-        The singular epoch delta value from the xarray.Dataset.
+        The midpoint of the pointing in ET.
 
         Returns
         -------
-        epoch_delta: int
-            The epoch delta value [J2000 TT ns] of the pointing set.
+        midpoint: int
+            The midpoint value [J2000 ET] of the pointing set.
         """
-        return self.data["epoch_delta"].values[0] if "epoch_delta" in self.data else 0
+        epoch_delta = self.data["epoch_delta"].values[0]
+        return float(ttj2000ns_to_et(self.epoch + epoch_delta / 2))
 
     @property
     def unwrapped_dims_dict(self) -> dict[str, tuple[str, ...]]:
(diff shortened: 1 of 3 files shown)


imap_processing/ena_maps/ena_maps.py
```diff
@@ -301,6 +306,18 @@ def epoch(self) -> int:
...
+    def epoch_delta(self) -> int:
+        """
+        The singular epoch delta value from the xarray.Dataset.
+
+        Returns
+        -------
+        epoch_delta: int
+            The epoch delta value [J2000 TT ns] of the pointing set.
+        """
+        return self.data["epoch_delta"].values[0] if "epoch_delta" in self.data else 0
```
lacoak21: Hm on second thought, all psets should have epoch_delta right? Maybe I shouldn't return 0 otherwise. Any thoughts here?

tmplummer: It looks to me like Lo doesn't have an epoch_delta. I think the options for Lo are to add epoch delta, or use the variables `pointing_start_met` and `pointing_end_met` to compute the epoch delta in a custom method `LoPointingSet.epoch_delta`.

tmplummer: I agree that this method should not return 0 if `epoch_delta` is not in `self.data`.

imap_processing/ena_maps/ena_maps.py
```diff
@@ -301,6 +306,18 @@ def epoch(self) -> int:
         """
         return self.data["epoch"].values[0]
 
+    @property
+    def epoch_delta(self) -> int:
```
tmplummer: Just a thought, but if you also added, or only added a property like:
```
@property
def midpoint_j2000(self) -> float:
    epoch_delta = self.data["epoch_delta"].values[0]
    return ttj2000ns_to_et(self.epoch + epoch_delta / 2)
```
it would save a little bit of duplicated code.

imap_processing/ena_maps/ena_maps.py
lacoak21: Oh perfect. Ill fix this and add that method for lo.

imap_processing/ena_maps/ena_maps.py
lacoak21: Nice thats much better.

tmplummer (approved):
LGTM

PR #2871, opened 2026-03-25
GLOWS L2 - Add ecliptic coordinates for histogram bin centers

### LINKED ISSUE #2722: ENH - Compute ecliptic location of bin centers in GLOWS L2 (post-SPICE)
Parent issue: #1276

`DailyLightcurve.ecliptic_lon` and `ecliptic_lat` are currently set to `np.zeros(number_of_bins)` (`glows_l2_data.py:92-93`).

These should be computed using SPICE: given the spin-axis ecliptic longitude/latitude (available from `spin_axis_orientation_average` in L1B) and the GLOWS scanning circle radius, compute the J2000 ecliptic longitude and latitude of each histogram bin center. See Section 12.7.6 of the algorithm document for context — these coordinates are also needed for the bad-angle masking in #2721.

vmartinez-cu:
This PR adds a new method to compute the ecliptic longitude and latitude for each histogram bin center in the GLOWS L2 product using SPICE transformations. It also adds a unit test and mocking of these computations in existing tests where appropriate to remain independent of SPICE/time conversions.

**Code Updates**
* Added the `compute_ecliptic_coords_of_bin_centers` static method to `DailyLightcurve`, which transforms instrument pointing directions from the IMAP spacecraft frame to the ECLIPJ2000 frame using SPICE, providing ecliptic longitude and latitude for each histogram bin center.
* Updated the `__post_init__` method of `DailyLightcurve` to compute and assign `ecliptic_lon` and `ecliptic_lat` using the new method, replacing the previous placeholder arrays.

**Testing**
* Added a `mock_ecliptic_bin_centers` test fixture to mock ecliptic coordinate computations, ensuring unit tests for `DailyLightcurve` are independent of SPICE/time conversions.
* Updated all affected tests in `test_glows_l2_data.py` and `test_glows_l2.py to use the new fixture, ensuring their reliability and isolation from 
* Added a unit test to test the output of the new method

Closes #2722

### DIFF
--- a/imap_processing/glows/l2/glows_l2_data.py
+++ b/imap_processing/glows/l2/glows_l2_data.py
@@ -147,8 +147,12 @@ class DailyLightcurve:
             self.exposure_times = np.roll(self.exposure_times, roll)
             self.flux_uncertainties = np.roll(self.flux_uncertainties, roll)
             self.histogram_flag_array = np.roll(self.histogram_flag_array, roll)
+
+            # Get the midpoint start time covered by repointing kernels
+            # needed to compute ecliptic coordinates
+            mid_idx = len(l1b_data["imap_start_time"]) // 2
             et_imap_start_time = sct_to_et(
-                met_to_sclkticks(l1b_data["imap_start_time"][0].data)
+                met_to_sclkticks(l1b_data["imap_start_time"][mid_idx].data)
             )
             self.ecliptic_lon, self.ecliptic_lat = (
                 self.compute_ecliptic_coords_of_bin_centers(
--- a/imap_processing/glows/l2/glows_l2_data.py
+++ b/imap_processing/glows/l2/glows_l2_data.py
@@ -183,10 +183,8 @@ class DailyLightcurve:
         """
         Compute the ecliptic coordinates of the histogram bin centers.
 
-        Histogram bin centers represent the center spin angle for each bin in the imap
-        frame, which corresponds to a specific pointing direction in space. This method
-        transforms the instrument pointing direction for each bin center from the IMAP
-        spacecraft frame to the ECLIPJ2000 frame.
+        This method transforms the instrument pointing direction for each bin
+        center from the IMAP Pointing frame (IMAP_DPS) to the ECLIPJ2000 frame.
 
         Parameters
         ----------
--- a/imap_processing/glows/l2/glows_l2_data.py
+++ b/imap_processing/glows/l2/glows_l2_data.py
@@ -151,12 +151,12 @@ class DailyLightcurve:
             # Get the midpoint start time covered by repointing kernels
             # needed to compute ecliptic coordinates
             mid_idx = len(l1b_data["imap_start_time"]) // 2
-            et_imap_start_time = sct_to_et(
+            pointing_midpoint_time_et = sct_to_et(
                 met_to_sclkticks(l1b_data["imap_start_time"][mid_idx].data)
             )
             self.ecliptic_lon, self.ecliptic_lat = (
                 self.compute_ecliptic_coords_of_bin_centers(
-                    et_imap_start_time, self.spin_angle
+                    pointing_midpoint_time_et, self.spin_angle
                 )
             )
 
@@ -182,7 +182,7 @@ class DailyLightcurve:
 
     @staticmethod
     def compute_ecliptic_coords_of_bin_centers(
-        data_start_time_et: float, spin_angle_bin_centers: NDArray
+        data_time_et: float, spin_angle_bin_centers: NDArray
     ) -> tuple[np.ndarray, np.ndarray]:
         """
         Compute the ecliptic coordinates of the histogram bin centers.
@@ -192,8 +192,8 @@ class DailyLightcurve:
 
         Parameters
         ----------
-        data_start_time_et : float
-            Ephemeris time corresponding to the start of the histogram accumulation.
+        data_time_et : float
+            Ephemeris time corresponding to the midpoint of the histogram accumulation.
 
         spin_angle_bin_centers : numpy.ndarray
             Spin angle bin centers for the histogram bins, measured in the IMAP frame,
@@ -218,7 +218,7 @@ class DailyLightcurve:
 
         # Transform coordinates to ECLIPJ2000 frame using SPICE transformations.
         ecliptic_coords = frame_transform_az_el(
-            data_start_time_et,
+            data_time_et,
             az_el,
             SpiceFrame.IMAP_DPS,
             SpiceFrame.ECLIPJ2000,
--- a/imap_processing/glows/l2/glows_l2_data.py
+++ b/imap_processing/glows/l2/glows_l2_data.py
@@ -8,6 +8,7 @@ from numpy.typing import NDArray
 
 from imap_processing.glows import FLAG_LENGTH
 from imap_processing.glows.l1b.glows_l1b_data import PipelineSettings
+from imap_processing.glows.utils.constants import GlowsConstants
 from imap_processing.spice.geometry import (
     SpiceFrame,
     frame_transform_az_el,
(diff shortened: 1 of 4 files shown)


imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -144,8 +147,14 @@ def __post_init__(self, l1b_data: xr.Dataset, position_angle: float) -> None:
             self.exposure_times = np.roll(self.exposure_times, roll)
             self.flux_uncertainties = np.roll(self.flux_uncertainties, roll)
             self.histogram_flag_array = np.roll(self.histogram_flag_array, roll)
-            self.ecliptic_lon = np.roll(self.ecliptic_lon, roll)
-            self.ecliptic_lat = np.roll(self.ecliptic_lat, roll)
+            et_imap_start_time = sct_to_et(
+                met_to_sclkticks(l1b_data["imap_start_time"][0].data)
```
tmplummer: Something that recently came up to be aware of: If this start time is identically equal to the start time of the Pointing, it can sometimes have a type conversion error that causes the queried value to be outside the coverage window of the DPS kernel. If "imap_start_time" is a MET time that has been checked against the repoint table MET time to ensure it is in the Pointing, this implementation should be fine. If there is a chance of a rounding error, I suggest using the midpoint MET time instead of start time here.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -167,6 +176,55 @@ def calculate_histogram_sums(histograms: NDArray) -> NDArray:
...
+        data_start_time_et: float, spin_angle_bin_centers: NDArray
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """
+        Compute the ecliptic coordinates of the histogram bin centers.
+
+        Histogram bin centers represent the center spin angle for each bin in the imap
+        frame, which corresponds to a specific pointing direction in space. This method
+        transforms the instrument pointing direction for each bin center from the IMAP
+        spacecraft frame to the ECLIPJ2000 frame.
+
```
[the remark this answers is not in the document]
tmplummer: Agree. Perhaps this should read "This method transforms the instrument pointing direction for each bin center from the IMAP Pointing frame (IMAP_DPS) to the ECLIPJ2000 frame."

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -167,6 +176,55 @@ def calculate_histogram_sums(histograms: NDArray) -> NDArray:
...
+        # Create array of azimuth, elevation coordinates in the DPS frame (n_bins, 2)
+        az_el = np.stack((azimuth, np.full_like(azimuth, elevation)), axis=-1)
+
+        # Transform coordinates to ECLIPJ2000 frame using SPICE transformations.
+        ecliptic_coords = frame_transform_az_el(
+            data_start_time_et,
+            az_el,
+            SpiceFrame.IMAP_DPS,
+            SpiceFrame.ECLIPJ2000,
```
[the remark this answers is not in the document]
tmplummer: Copilot is wrong here. We want the exception to get raised if the time is outside of the coverage window.

imap_processing/tests/glows/test_glows_l2.py
```diff
@@ -93,6 +96,8 @@ def test_generate_l2(
     mock_ancillary_exclusions,
     mock_pipeline_settings,
     mock_conversion_table_dict,
+    furnish_kernels,
```
tmplummer: I'm not sure I see why the `furnish_kernels` fixture was added here. Maybe an artifact from before the `mock_ecliptic_bin_centers` was added?

tmplummer (commented):
Great! Everything looks good to me except for the one gotcha I mentioned about the time used for the transform_az_el call. Probably worth checking in with @maxinelasp about where that start time comes from.

imap_processing/glows/l2/glows_l2_data.py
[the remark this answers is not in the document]
vmartinez-cu: Thanks for your comment here!

imap_processing/glows/l2/glows_l2_data.py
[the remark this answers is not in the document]
vmartinez-cu: Your suggestions is much clearer, thanks! Are the IMAP spacecraft and DPS frames different?

imap_processing/tests/glows/test_glows_l2.py
vmartinez-cu: Good catch - I'll remove this

imap_processing/glows/l2/glows_l2_data.py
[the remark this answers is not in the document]
tmplummer: Yes. The IMAP_SPACECRAFT frame is attached to the spacecraft and rotates with it. The IMAP_DPS frame is coincident with the s/c but does not rotate with it.

tech3371:
Looks like Tim and Maxine is reviewing this PR. I will standby for now :)

imap_processing/glows/l2/glows_l2_data.py
maxinelasp: That is not a safe assumption we can make here, so I think Tim's suggestion of midpoint timestamp makes the most sense. @subagonsouth, is this only for the `sct_to_et` conversion? Or are there other cases where this might be an issue?

maxinelasp (commented):
Nice work! Only thing is the change suggested by Tim - everything else looks good :) Once that is addressed, consider this approved by me

imap_processing/glows/l2/glows_l2_data.py
tmplummer: The problem isn't really with the time conversion functions. Those functions will work for whatever time you give them. The problem arises when a certain series of time conversions happen to the Pointing start time that gets passed into a `frame_transform` call that uses the DPS kernel. I don't know how much detail you want but, here goes.

The DPS kernel has times represented in spacecraft clock format which is integer seconds and integer ticks (1 tick is 20us on IMAP). The repoint table also has times represented in sclk format. But, spice functions use double precision floating point TDB seconds (AKA ET). To complicate this even more, often in IMAP, we are storing the epoch as integer nanoseconds. So, a condition can arise where the conversion from sclk (MET) to epoch then back to TDB to pass into the frame_transform function which SPICE then converts back to sclk can result in an sclk value that is slightly before the sclk start time of the coverage provided by the DPS kernel. Since the DPS kernel is static during a pointing, the safest way to avoid this situation is to use the midpoint time of the Pointing.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -144,8 +147,18 @@ def __post_init__(self, l1b_data: xr.Dataset, position_angle: float) -> None:
             self.exposure_times = np.roll(self.exposure_times, roll)
             self.flux_uncertainties = np.roll(self.flux_uncertainties, roll)
             self.histogram_flag_array = np.roll(self.histogram_flag_array, roll)
-            self.ecliptic_lon = np.roll(self.ecliptic_lon, roll)
-            self.ecliptic_lat = np.roll(self.ecliptic_lat, roll)
+
+            # Get the midpoint start time covered by repointing kernels
+            # needed to compute ecliptic coordinates
+            mid_idx = len(l1b_data["imap_start_time"]) // 2
+            et_imap_start_time = sct_to_et(
```
tmplummer: Nit pick: You might want to update this variable name to indicate that it is the Pointing midpoint time.

tmplummer (approved):
Other than the one nit-pick... Looks great!

maxinelasp (approved):

imap_processing/glows/l2/glows_l2_data.py
vmartinez-cu: Good suggestion - renaming this variable will make this step more clear

PR #2873, opened 2026-03-26
Support  for --dependency when its a path to an arbitrarily-located local .json file

vineetbansal:
# Change Summary

`imap_cli ... --dependency dependency.json` now works correctly.

I'm still fairly new to this code so my apologies in advance if I'm misunderstanding something here!

## Overview

Support for `--dependency <local_json_file.json>` didn't exactly work when I tried it. It's my understanding that `.json` files for dependencies are not meant to be downloaded from the server (which is what the code was trying to do) but generated locally.

## File changes

`cli.py` now doesn't try to download the `.json` file but simply reads it for its contents.

## Testing

The associated test (modified now) originally only worked because it bypassed the logic of `download` by placing it at the correct place before testing the CLI command. This modified test will fail in the current code without the associated changes.

### DIFF
--- a/imap_processing/cli.py
+++ b/imap_processing/cli.py
@@ -27,7 +27,7 @@ import spiceypy
 import xarray as xr
 from cdflib.xarray import xarray_to_cdf
 from cdflib.xarray.xarray_to_cdf import ISTPError
-from imap_data_access.io import IMAPDataAccessError, download
+from imap_data_access.io import IMAPDataAccessError
 from imap_data_access.processing_input import (
     ProcessingInputCollection,
     ProcessingInputType,
@@ -267,8 +267,7 @@ def _parse_args() -> argparse.Namespace:
         logger.info(
             f"Interpreting dependency argument as a JSON file: {args.dependency}"
         )
-        dependency_filepath = download(args.dependency)
-        with open(dependency_filepath) as f:
+        with open(args.dependency) as f:
             args.dependency = f.read()
 
     return args
--- a/imap_processing/tests/test_cli.py
+++ b/imap_processing/tests/test_cli.py
@@ -130,11 +130,8 @@ def test_parse_args_dependency_json_file(caplog, tmp_path):
         },
     ]
     test_json_filename = "imap_ultra_l2_test-dependency-json_20250520_v999.json"
-    test_json_dir = tmp_path / "imap/dependency/ultra/l2/2025/05/"
-    test_json_dir.mkdir(parents=True, exist_ok=True)
-    test_json_dst = test_json_dir / test_json_filename
 
-    with open(test_json_dst, "w") as f:
+    with open(test_json_filename, "w") as f:
         f.write(json.dumps(test_json_content))
 
     test_args = [
@@ -142,7 +139,7 @@ def test_parse_args_dependency_json_file(caplog, tmp_path):
         "--instrument",
         "mag",
         "--dependency",
-        str(test_json_dst),
+        str(test_json_filename),
         "--data-level",
         "l1a",
         "--start-date",


imap_processing/cli.py
```diff
@@ -267,8 +267,7 @@ def _parse_args() -> argparse.Namespace:
         logger.info(
             f"Interpreting dependency argument as a JSON file: {args.dependency}"
         )
-        dependency_filepath = download(args.dependency)
-        with open(dependency_filepath) as f:
+        with open(args.dependency) as f:
```
tmplummer, line 270: The current implementation should already work if your dependency named appropriately and located in the correct directory. The `download` function looks for the file in the appropriate location and will not attempt to download it if that file exists. I could see the utility in not validating the filename or path if an absolute path to the dependency is give.

This change would break a very useful function for developers. When a processing job fails in production, we can grab the dependency name from the SDC website and run that exact job locally. Processing will automatically download the dependency file for us. So, we definitely want to keep that functionality.

tmplummer (commented):
See my comment in the code. This change breaks a very useful download functionality. I suggest that it isn't too onerous to name your dependency json appropriately and locate it in the correct directory. If that is done, the current functionality will not try to download the file.

vineetbansal:
@subagonsouth - thanks! So my assumption was wrong - dependency files *are* meant to be downloaded when using the `--dependency` flag. Makes sense. However, that means that running against pre-generated `.json` files (for scripting purposes, or in my case, running against custom dependencies to investigate the `imap_processing` codebase itself) will never work.

I'd rather not manually place files in `IMAP_DATA_DIR` since in a way that's "reserved" for officially released files.

It's another small tweak to support both use cases - simply not doing anything if the path exists before calling the `download` function. If this would be acceptable, I can add another commit to this PR. If you think my use-case is too esoteric, then I can work around it and you can close this.

tech3371 (changes requested):
Vineet, can you create a ticket before making changes to main code? These changes will need to be reviewed before anyone can make changes. Especially, this `cli.py` file is used by all instrument pipeline and by various users. So we need to be careful when making updates like this.

Please try Tim's suggestions on creating directly structure.

vineetbansal:
Thanks Tenzin - perhaps I misunderstand - this is just a PR originating from my fork. I can't (and don't want to!) make any changes to the main code. If you want me to open a ticket independent of this PR so you can keep track of "user stories" etc. let me know where I can do that. Or perhaps you meant open an "issue" ?

tech3371:
> Thanks Tenzin - perhaps I misunderstand - this is just a PR originating from my fork. I can't (and don't want to!) make any changes to the main code. If you want me to open a ticket independent of this PR so you can keep track of "user stories" etc. let me know where I can do that. Or perhaps you meant open an "issue" ?

Yes, if you can create a ticket meaning GitHub issue to capture this use case, it would be helpful. If we merge this PR, it would effect main code and have ripple effect.

tech3371:
I am going to close this PR for now and if you can transition this finding to a ticket, that would be great. Thank you for understanding our workflow and supporting it!

tmplummer:
@vineetbansal 

> I'd rather not manually place files in IMAP_DATA_DIR since in a way that's "reserved" for officially released files.
>
> It's another small tweak to support both use cases - simply not doing anything if the path exists before calling the download function.

I'm not opposed to a change that does as you suggest.

> Thanks Tenzin - perhaps I misunderstand - this is just a PR originating from my fork. I can't (and don't want to!) make any changes to the main code.

But this PR that you opened is to merge the changes on your forked repo into the main imap_processing repo. No PR is needed to make any changes that you want to your repo, but then you have to deal with keeping your forked repo up to date with changes in the main repo.

vineetbansal:
Sure - I was "proposing changes" through this PR, not "making changes". It's perfectly reasonable to me that certain PRs get closed without merge. I'll open an issue on this first and then a PR if the conversation suggests so, if that's the preferred workflow. Thanks!

PR #2877, opened 2026-03-27
MNT: update github access role based on project document

tech3371:
# Change Summary

## Overview

Update based on project document and yesterday findings.

### DIFF
--- a/docs/source/development/git-access-roles.rst
+++ b/docs/source/development/git-access-roles.rst
@@ -3,15 +3,14 @@
 GitHub Access & Permissions Guide
 ======================================
 
-This document outlines the different permission levels available in IMAP
-repositories and what each role can do.
+This document outlines the permission levels available in IMAP SDC
+repositories.
 
 Overview
 ========
 
-GitHub provides five repository roles. Below is a detailed breakdown of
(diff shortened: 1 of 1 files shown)


tech3371:
Added all people who will need to be informed of what access we need to give if new contributor comes. Right now, no action needed from SDC.

jtniehof (approved):
I like that this is a lot more focused on the specifics of what we care about. Only suggestion would be to consider adding a link to https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role so that it's easy to get at the full details if someone needs to.

tech3371:
> I like that this is a lot more focused on the specifics of what we care about. Only suggestion would be to consider adding a link to https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role so that it's easy to get at the full details if someone needs to.

Totally! I will link that in that page!

tmplummer (approved):
I don't really understand the removal roles, but LGTM.

PR #2881, opened 2026-03-30
Allow customization of MAG L2 frames

alastairtree:
# Change Summary

## Overview

Adds option to customise which frames are generated when calling  mag l2 programatically. Used (externally) by MAG team to generate l2 files without needing all CDF files

## Testing

I have updated the unite tests to cover this new use case.
The main API has not actually changed as this is just a new param with defaults that match the existing behaviou so should have zero consequences other than making it more flexible

### DIFF
--- a/imap_processing/mag/l2/mag_l2.py
+++ b/imap_processing/mag/l2/mag_l2.py
@@ -12,6 +12,13 @@ from imap_processing.mag.l2.mag_l2_data import MagL2, ValidFrames
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_L2_FRAMES = [
+        ValidFrames.SRF,
+        ValidFrames.GSE,
+        ValidFrames.GSM,
+        ValidFrames.RTN,
+        ValidFrames.DSRF,  # should be last as some vectors may become NaN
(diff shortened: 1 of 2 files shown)


alastairtree:
Paging @maxinelasp - please could you take a look at this, a small change to MAG L2 to enable us to generate frames programmatically without having to generate all of them. We (intend) to call this code in our internal pipeline over at https://github.com/ImperialCollegeLondon/imap-pipeline-core/blob/feat/set-quality-and-nan/src/mag_toolkit/calibration/CalibrationApplicator.py#L207

maxinelasp (approved):

alastairtree:
Thanks @maxinelasp. What is the cadence of your releases & when can i expect to see this change on pypy? No pressure, just want to know what to expect.

maxinelasp:
@alastairtree We usually do a few a month - if you would like this out ASAP, I can cut a release this week

alastairtree:
Next routine release is fine thanks. I cant merge it so over to you to release it please and thank you!

laspsandoval:
@alastairtree we just released imap_processing.

PR #2888, opened 2026-03-30
Fix hi goodtimes CDF to txt convertor

### LINKED ISSUE #2879: BUG - Hi Goodtimes text convertor write out goodtimes only
### Description of the issue

Currently the CDF to text convertor writes out each contiguous segment indicating both goodtimes and badtimes. It should be updated to only indicate goodtimes.

_No response_

_No response_

```python
Code
```

_No response_

tmplummer:
# Change Summary

This PR only changes logic used to write CDF files to txt. This is not used in producing goodtimes products. It is only used by the IT to convert the CDF to a txt format used on IBEX so that Paul's tools will work with the goodtimes txt file.

The previous logic was not what Paul wanted. He wants the list to indicate the goodtimes only. I was previously writing out the culled times.

Closes: #2879

### DIFF
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -604,8 +604,8 @@ class GoodtimesAccessor:
             cull_next = grouped[i + 1][1]
 
             # Compare shapes first (different lengths = different pattern)
-            # In a nominal Pointing, the final ESA sweep will get cut short by the repoint
-            # maneuver. This is the typical case where the shape would be different.
+            # In a nominal Pointing, the final ESA sweep will get cut short by
+            # the repoint maneuver. This causes a difference in shape.
             if cull_curr.shape != cull_next.shape:
                 pattern_changes.append(True)
             else:
--- a/imap_processing/hi/hi_goodtimes.py
+++ b/imap_processing/hi/hi_goodtimes.py
@@ -687,8 +687,10 @@ class GoodtimesAccessor:
             else:
                 bad_vals = cull_pattern[cull_pattern > 0]
                 if len(bad_vals) > 0:
-                    bad_cull_value |= int(np.bitwise_or.reduce(bad_vals))
-                    region_cull = int(bad_vals[0])
+                    # Aggregate all non-zero cull codes for this ESA step so that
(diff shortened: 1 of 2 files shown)


imap_processing/hi/hi_goodtimes.py
```diff
@@ -552,141 +552,213 @@ def mark_bad_times(
...
+            # The grouped list contains tuples (sweep_idx, cull_flags_ds).
+            # Grab just the cull_flags_ds values for comparison.
+            cull_curr = grouped[i][1]
+            cull_next = grouped[i + 1][1]
 
-        for met_idx in range(1, len(met_values)):
-            if np.array_equal(cull_flags[met_idx], current_pattern):
-                # Same pattern - extend current group
-                esa_step_mask |= 1 << int(esa_steps[met_idx] - 1)
+            # Compare shapes first (different lengths = different pattern)
```
lacoak21, line 606: I think im getting lost here - why would the shapes be different?

imap_processing/hi/hi_goodtimes.py
```diff
@@ -2004,6 +2083,11 @@ def _find_event_clusters(
     # Find transitions: +1 = start of group, -1 = end of group
     diff = np.diff(padded.astype(int))
     starts = np.flatnonzero(diff == 1)
+    # We need to adjust ends for the shortening from diffs performed.
```
lacoak21, line 2088: Nice comment addition

lacoak21 (approved):
Looks good to me! I just had one clarifying question.

imap_processing/hi/hi_goodtimes.py
tmplummer: Yeah, this is not obvious. Only the final sweep should have a different shape because the repoint maneuver will cut it off wherever it is. This causes the final sweep to have a different number of ESA steps.

tmplummer, line 606:
```suggestion
            # Compare shapes first (different lengths = different pattern)
            # In a nominal Pointing, the final ESA sweep will get cut short by the repoint
            # maneuver. This is the typical case where the shape would be different.
```

PR #2891, opened 2026-03-31
CoDICE: Fixing attrs bugs for CAVA

### LINKED ISSUE #2765: CoDICE: CAVA metadata support
```
Subject: Re: CoDICE in CAVA

I grabbed imap_codice_l2_hi-omni_20260104_v001.cdf off the website.

There are a lot of invalid FILLVALs...should be -1e31, and are 1e31

Despite 20260104 as the name, there are several records from 2026-01-03

The epoch_delta variables are doubles not ints.

A lot of the energy uncertainties have values under the VALIDMIN (maybe
the values for the energy center channels were applied to the
uncertainties?)

Most format strings are (incorrectly) C-style instead of Fortran-style.

There's some SI_CONVERSION attrs that make no sense (" > ", i.e.
"multiply by blank to convert this to SI units of blank")

I'm not sure of all the "N/A". There aren't LABLAXIS attributes on those
variables, but technically it's not required because they're using
LABL_PTR_1. That might be a CAVA bug--if we're normally displaying
LABLAXIS below there, we should probably use FIELDNAM, or put nothing
there, if there is a LABLAXIS. (This is a part of the spec that's really
unclear IMO). I can't find the code in CAVA that populates that field,
and there is no "N/A" actually in any of the attributes.

The epoch is compressed and shouldn't be.

So it's got a lot of the required metadata but needs some work to be
compliant.

--
Jonathan Niehof, Ph.D.
UNH Space Science Center
106 Morse Hall, 8 College Rd., Durham, NH 03824
(603) 862-0649
```

tech3371:
# Change Summary
closes #2765

## Overview

Fixes for three products that needed fixes for CAVA, L2 omni, sectored and lo-sw-species.

## File changes

Updated CDF attr YAML and code to not carry l1b attrs which caused few issues. L2 needs to read from L2 attrs file.

### DIFF
--- a/imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
+++ b/imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
@@ -10,22 +10,23 @@ max_int: &max_int 9223372036854775807
 energy_attrs: &energy_default
   VAR_TYPE: support_data
   CATDESC: Geometric mean energy per nucleon
+  DISPLAY_TYPE: time_series
   FIELDNAM: Energy Table
   LABLAXIS: Energy
   SCALETYP: log
   UNITS: MeV/nuc
-  FORMAT: '%f'
+  FORMAT: F32.1
   FILLVAL: *real_fillval
   VALIDMAX: 200.0
-  VALIDMIN: 0.05000000074505806
+  VALIDMIN: 0.0
   DICT_KEY: SPASE>Particle>ParticleType:Ion,ParticleQuantity:EnergyPerCharge,Qualifier:Characteristic
 
 # ------------------------------- Coordinates -------------------------------
 epoch_delta_minus:
   CATDESC: Time from acquisition start to acquisition center
   FIELDNAM: epoch delta minus
-  FILLVAL: -9223372036854775808
-  FORMAT: I18
+  FILLVAL: *min_int
+  FORMAT: I19
   LABLAXIS: Epoch Delta Minus
   SCALETYP: linear
   UNITS: ns
@@ -37,8 +38,8 @@ epoch_delta_minus:
 epoch_delta_plus:
   CATDESC: Time from acquisition center to acquisition end
   FIELDNAM: epoch delta plus
-  FILLVAL: -9223372036854775808
-  FORMAT: I18
+  FILLVAL: *min_int
+  FORMAT: I19
   LABLAXIS: Epoch Delta Plus
   SCALETYP: linear
   UNITS: ns
@@ -334,11 +335,11 @@ unc_c:
   VALIDMIN: 0.0
   VALIDMAX: 4096.0
   UNITS: '# / cm2 s sr MeV/nuc'
-  FORMAT: '%f'
+  FORMAT: F32.1
   CATDESC: Uncertainties for c (Root 2 spacing)
   VAR_TYPE: support_data
   SI_CONVERSION: ' > '
-  SCALETYP: linear
+  SCALETYP: log
   FILLVAL: *real_fillval
   DISPLAY_TYPE: time_series
   DEPEND_0: epoch
@@ -351,11 +352,11 @@ unc_fe:
   VALIDMIN: 0.0
   VALIDMAX: 4096.0
   UNITS: '# / cm2 s sr MeV/nuc'
-  FORMAT: '%f'
+  FORMAT: F32.1
   CATDESC: Uncertainties for fe (Root 2 spacing)
   VAR_TYPE: support_data
   SI_CONVERSION: ' > '
-  SCALETYP: linear
+  SCALETYP: log
   FILLVAL: *real_fillval
   DISPLAY_TYPE: time_series
   DEPEND_0: epoch
@@ -368,11 +369,11 @@ unc_h:
   VALIDMIN: 0.0
   VALIDMAX: 4096.0
   UNITS: '# / cm2 s sr MeV/nuc'
-  FORMAT: '%f'
+  FORMAT: F32.1
   CATDESC: Uncertainties for h (Root 2 spacing)
   VAR_TYPE: support_data
   SI_CONVERSION: ' > '
-  SCALETYP: linear
+  SCALETYP: log
   FILLVAL: *real_fillval
   DISPLAY_TYPE: time_series
   DEPEND_0: epoch
@@ -385,11 +386,11 @@ unc_he3:
   VALIDMIN: 0.0
   VALIDMAX: 4096.0
   UNITS: '# / cm2 s sr MeV/nuc'
-  FORMAT: '%f'
+  FORMAT: F32.1
   CATDESC: Uncertainties for he3 (Root 2 spacing)
   VAR_TYPE: support_data
   SI_CONVERSION: ' > '
-  SCALETYP: linear
+  SCALETYP: log
   FILLVAL: *real_fillval
   DISPLAY_TYPE: time_series
   DEPEND_0: epoch
@@ -402,11 +403,11 @@ unc_he4:
   VALIDMIN: 0.0
   VALIDMAX: 4096.0
   UNITS: '# / cm2 s sr MeV/nuc'
-  FORMAT: '%f'
+  FORMAT: F32.1
   CATDESC: Uncertainties for he4 (Root 2 spacing)
   VAR_TYPE: support_data
   SI_CONVERSION: ' > '
-  SCALETYP: linear
+  SCALETYP: log
   FILLVAL: *real_fillval
   DISPLAY_TYPE: time_series
   DEPEND_0: epoch
@@ -419,11 +420,11 @@ unc_junk:
   VALIDMIN: 0.0
   VALIDMAX: 4096.0
   UNITS: '# / cm2 s sr MeV/nuc'
-  FORMAT: '%f'
+  FORMAT: F32.1
   CATDESC: Uncertainties for junk (Root 2 spacing)
   VAR_TYPE: support_data
   SI_CONVERSION: ' > '
-  SCALETYP: linear
+  SCALETYP: log
   FILLVAL: *real_fillval
   DISPLAY_TYPE: time_series
   DEPEND_0: epoch
@@ -436,11 +437,11 @@ unc_ne_mg_si:
   VALIDMIN: 0.0
   VALIDMAX: 4096.0
   UNITS: '# / cm2 s sr MeV/nuc'
-  FORMAT: '%f'
+  FORMAT: F32.1
   CATDESC: Uncertainties for ne-mg-si (Root 2 spacing)
   VAR_TYPE: support_data
   SI_CONVERSION: ' > '
-  SCALETYP: linear
+  SCALETYP: log
   FILLVAL: *real_fillval
   DISPLAY_TYPE: time_series
   DEPEND_0: epoch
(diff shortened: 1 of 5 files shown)


tech3371:
@jtniehof pining you for awareness.

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
```diff
@@ -10,22 +10,23 @@ max_int: &max_int 9223372036854775807
...
+  VALIDMIN: 0.0
   DICT_KEY: SPASE>Particle>ParticleType:Ion,ParticleQuantity:EnergyPerCharge,Qualifier:Characteristic
 
 # ------------------------------- Coordinates -------------------------------
 epoch_delta_minus:
   CATDESC: Time from acquisition start to acquisition center
   FIELDNAM: epoch delta minus
-  FILLVAL: -9223372036854775808
-  FORMAT: I18
+  FILLVAL: *real_fillval
```
tmplummer: I'm not sure that I understand this change. I would expect epoch_delta_minus/plus to be int64 to match the epoch datatype and contain integer nanoseconds.

imap_processing/codice/codice_l2.py
```diff
@@ -892,7 +930,29 @@ def process_hi_omni(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
+            ),
+        ),
     }
+
+    # TODO: remove this addition once SAMMI or CDF attr manager handles this.
+    # Update epoch attrs to add epoch_delta_plus and epoch_delta_minus
+    epoch_attrs = l1b_dataset["epoch"].attrs.copy()
+    epoch_attrs["DELTA_MINUS_VAR"] = "epoch_delta_minus"
+    epoch_attrs["DELTA_PLUS_VAR"] = "epoch_delta_plus"
+    l1b_dataset["epoch"].attrs = epoch_attrs
```
tmplummer: Is the copy() needed here? Is it not possible to just do:
```
l1b_dataset["epoch"].attrs["DELTA_MINUS_VAR"] = "epoch_delta_minus"
l1b_dataset["epoch"].attrs["DELTA_PLUS_VAR"] = "epoch_delta_plus"
```

imap_processing/codice/codice_l2.py
```diff
@@ -993,6 +1077,13 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
         attrs=cdf_attrs.get_global_attributes("imap_codice_l2_hi-sectored"),
     )
 
+    # TODO: remove this addition once SAMMI or CDF attr manager handles this
+    # Update epoch attrs to add epoch_delta_plus and epoch_delta_minus
+    epoch_attrs = l1b_dataset["epoch"].attrs.copy()
```
tmplummer: Same question about needing to copy and reassign the epoch attrs here.

imap_processing/codice/codice_l2.py
```diff
@@ -815,71 +815,109 @@ def process_hi_omni(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
             attrs=cdf_attrs.get_variable_attributes(
                 "energy_h_label", check_schema=False
             ),
         ),
-        "energy_he3": l1b_dataset["energy_he3"],
+        "energy_he3": xr.DataArray(
+            l1b_dataset["energy_he3"].values,
+            dims=("energy_he3",),
+            attrs=cdf_attrs.get_variable_attributes("energy_he3", check_schema=False),
+        ),
```
tmplummer, line 834: Lots of repeated code here. I wonder if it would be possible to do something like:
```
for coord in LIST_OF_COORDS:
    new_coords[coord] = xr.DataArray(
        l1b_dataset[coord].values,
        dims=(coord,),
        attrs=cdf_attrs.get_variable_attributes(coord, check_schema=False),
    )
```

tmplummer (approved):
Nothing blocking a merge, but a couple questions and an idea to reduce duplicated code.

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
vmartinez-cu, line 497: Just a comment but not needed for this PR - It might be worth considering a more descriptive CATDESC for the particle variables. HIT L2 attrs have some examples of more descriptive strings for this field.

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
```diff
@@ -518,12 +517,11 @@ fe:
 h:
   VAR_TYPE: data
   DEPEND_0: epoch
-  DISPLAY_TYPE: spectrogram
-  FIELDNAM: h
+  DISPLAY_TYPE: time_series
+  FIELDNAM: Diff. Intensity - h
```
vmartinez-cu, line 521: For HIT, these variables have a spectrogram display type. Just curious if the same is needed here or if this is personal preference by the instrument teams

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
tech3371: I was getting this warning when running ISTP checks. I remember now. Epoch delta were double because we had to divide by something and so on. That's why deltas were double type.

```
epoch_delta_plus
		FILLVAL value of '9.223372036854776E18' is non-standard.
		    The recommended value is '-1.0E31'.
		FORMAT 'I32' is invalid for a CDF_DOUBLE type variable
```

imap_processing/cdf/config/imap_codice_l2-hi-sectored_variable_attrs.yaml
```diff
@@ -409,12 +410,12 @@ spin_angle:
   VAR_TYPE: support_data
   CATDESC: Spin Angle
   FIELDNAM: Spin Angle
-  SCALETYP: linear
+  SCALETYP: log
```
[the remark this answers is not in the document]
tech3371: no. log is prefered.

imap_processing/cdf/config/imap_codice_l2-lo-species_variable_attrs.yaml
```diff
@@ -17,13 +19,13 @@ elevation_angle:
 spin_sector:
   CATDESC: Spin Sector Index
   FIELDNAM: Spin Sector Index
-  FILLVAL: -1
-  FORMAT: I2
+  FILLVAL: *uint8_fillval
```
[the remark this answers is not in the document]
tech3371: not at this time. Edit: Nice catch! I totally missed it.

imap_processing/codice/codice_l2.py
tech3371: true!

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
jtniehof: I think the fix there is to change the type of the variable, as Tim suggested, not the fill (and the format).

imap_processing/codice/codice_l2.py
```diff
@@ -815,71 +815,109 @@ def process_hi_omni(dependencies: ProcessingInputCollection) -> xr.Dataset:
...
             dims=("energy_he4",),
             attrs=cdf_attrs.get_variable_attributes(
                 "energy_he4_label", check_schema=False
             ),
         ),
-        "energy_c": l1b_dataset["energy_c"],
+        "energy_c": xr.DataArray(
+            l1b_dataset["energy_c"].values,
+            dims=("energy_c",),
+            attrs=cdf_attrs.get_variable_attributes("energy_c", check_schema=False),
```
vmartinez-cu, line 857: Should `check_schema=False`? I thought this was only needed for dimensions and label variables to avoid DEPEND_0 being added unnecessarily

imap_processing/codice/codice_l2.py
```diff
@@ -1046,23 +1137,33 @@ def process_hi_sectored(dependencies: ProcessingInputCollection) -> xr.Dataset:
         )
 
         # Replace existing species data with omni-directional intensities
+        species_attrs = cdf_attrs.get_variable_attributes(species, check_schema=False)
+        # Replace {species} in attributes with actual species name
+        species_attrs = apply_replacements_to_attrs(species_attrs, {"species": species})
```
vmartinez-cu, line 1122: I didn't know about this option to replace names in the attributes. This is helpful!

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
tech3371: so if delta's data is float, convert to int? I haven't looked closely but it's ok to round up for delta if they are float type? If that is ok, I can totally revert and change data type of delta.

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
jtniehof: If it's higher dimensionality (i.e. not a single scalar quantity per time record), plotting in CAVA needs the DISPLAY_TYPE to be `spectrogram`. I know the CoDICE team prefers stacked line plots for some of these (that'll be `stack_plot`) but it's not supported yet. This is true for the uncertainties as well.

Issue tracking this on the L3 side: IMAP-Science-Operations-Center/imap_L3_processing#9

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
jtniehof: This is a mildly uninformed opinion since I haven't tracked the time handling through the l1 code.

Since the deltas are in ns, rounding shouldn't cause an issue in terms of scientific interpretation or anything like that. There might be some slight weirdness with certain tools if adjacent times get rounded differently, which might imply a small gap between successive measurements or overlap between them. I did a chunk check of an existing omni hi l2 and it looks like the fractional part of all the deltas is zero, so that feels like something not worth worrying about right now.

tech3371: Perfect! Thank you for that information! It's scientific user experience that I don't enough and didn't want to assume :)

I will make changes Tim suggested!

imap_processing/codice/codice_l2.py
tech3371: These are dimension variable. Each species need it's own `energy_{species}`. That's why.

imap_processing/codice/codice_l2.py
tech3371: yeah, Nat started it and I took it for CoDICE. been so helpful, especially for CoDICE's unique cases.

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
tech3371: I hoping to get minimal changes needed for CAVA today to unblock scientist. But I plan to go back and do more thorough updates for CoDICE L2. I will keep this in mind!

imap_processing/codice/codice_l2.py
tech3371: I am going to make this change in upcoming PR. Really good suggestion!

imap_processing/codice/codice_l2.py
tech3371: I tried but something going on. Can I implement this change in the upcoming PR?

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
tech3371: I am going to revert these epoch updates and update l1a to convert to int in upcoming PR. Want to get this CAVA support for Leng Yeng.

imap_processing/cdf/config/imap_codice_l2-hi-omni_variable_attrs.yaml
jtniehof: > I hoping to get minimal changes needed for CAVA today to unblock scientist. But I plan to go back and do more thorough updates for CoDICE L2. I will keep this in mind!

:+1: Once the CoDICE team can look at the data in CAVA I think they can give much better feedback on the metadata...otherwise it's hard to tell the CATDESC from the FIELDNAM from the LABLAXIS.

PR #2893, opened 2026-03-31
GLOWS fix L1B error

maxinelasp:
# Change Summary
This change fixes the L1B error we saw with new ancillary files. 

Basically, the new file is empty, and the code did not gracefully handle this case.


tech3371 (approved):
LGTM

PR #2901, opened 2026-04-01
A rotated NaN is still a NaN

alastairtree:
Fix rotations outside of spice so they maintain FILL_VAL during rotation.
Before this we were introducing nearly but not quite NaN/FILL_VAL meaning very bad field data.

After this is merged and released please can the SDC regenerate new versions of  all MAG L2 BURST science (not normal - as it has no NaNs).

## Testing

Added failing unit test to repro the issue. We see this when publishing L2 science data.

### DIFF
--- a/imap_processing/spice/geometry.py
+++ b/imap_processing/spice/geometry.py
@@ -18,6 +18,8 @@ import numpy.typing as npt
 import spiceypy
 from numpy.typing import NDArray
 
+from imap_processing.mag import constants
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


alastairtree:
@maxinelasp please can you also review and release this ASAP please? And then regen all L2 BM. This issue means all our L2 BM science has very dodgy values like -0.999999999999e30 and not the proper FILL_VAL. I am away for a few days so please push this forward in my absence if possible. Thank you.

maxinelasp (approved):
Great catch! I'll get this released ASAP and re-process.

imap_processing/spice/geometry.py
```diff
@@ -317,6 +319,14 @@ def frame_transform(
...
     result = np.squeeze(rotate @ position[..., np.newaxis])
 
+    # For every FILLVAL in the input position, ensure the output is also NaN or FILLVAL
+    if np.isnan(position).any() or (position == constants.FILLVAL).any():
+        result = np.where(
+            np.isnan(position) | (position == constants.FILLVAL),
+            constants.FILLVAL,
+            result,
+        )
+
```
tmplummer, line 329: Oh, no! This is MAG specific code in the common geometry function. If the input is a NaN, this function already behaves as expected. If the input is a MAG specific FILLVAL, then that case should be handled explicitly in the MAG code, not the general geometry module. @maxinelasp please revert this.

maxinelasp: Shoot, good catch Tim! I assumed this was in the rotate frames function in the mag code and totally missed the file. Sorry! I will revert and put up a proper fix.

PR #2903, opened 2026-04-01
Mag L2 rotation - propogate FILLVALS

maxinelasp:
# Change Summary
This moves the change from #2901 into MAG-specific code.

### DIFF
--- a/imap_processing/mag/l1d/mag_l1d_data.py
+++ b/imap_processing/mag/l1d/mag_l1d_data.py
@@ -298,26 +298,40 @@ class MagL1d(MagL2L1dBase):  # type: ignore[misc]
             self.epoch_et: np.ndarray = ttj2000ns_to_et(self.epoch)
             self.magi_epoch_et: np.ndarray = ttj2000ns_to_et(self.magi_epoch)
 
-        self.vectors = frame_transform(
+        new_vectors = frame_transform(
             self.epoch_et,
             self.vectors,
             from_frame=start_frame.spice_frame,
             to_frame=end_frame.spice_frame,
             allow_spice_noframeconnect=True,
         )
+        if np.isnan(self.vectors).any() or (self.vectors == FILLVAL).any():
+            new_vectors = np.where(
+                np.isnan(self.vectors) | (self.vectors == FILLVAL),
+                FILLVAL,
+                new_vectors,
+            )
+        self.vectors = new_vectors
 
         # If we were in MAGO frame, we need to rotate MAGI vectors from MAGI to
         # end_frame
         if start_frame == ValidFrames.MAGO:
             start_frame = ValidFrames.MAGI
 
-        self.magi_vectors = frame_transform(
+        new_magi_vectors = frame_transform(
             self.magi_epoch_et,
             self.magi_vectors,
             from_frame=start_frame.spice_frame,
             to_frame=end_frame.spice_frame,
             allow_spice_noframeconnect=True,
         )
+        if np.isnan(self.magi_vectors).any() or (self.magi_vectors == FILLVAL).any():
+            new_magi_vectors = np.where(
+                np.isnan(self.magi_vectors) | (self.magi_vectors == FILLVAL),
+                FILLVAL,
+                new_magi_vectors,
+            )
+        self.magi_vectors = new_magi_vectors
 
         self.frame = end_frame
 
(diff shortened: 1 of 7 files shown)


imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -298,26 +298,40 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
...
             self.epoch_et,
             self.vectors,
             from_frame=start_frame.spice_frame,
             to_frame=end_frame.spice_frame,
             allow_spice_noframeconnect=True,
         )
+        if np.isnan(self.vectors).any() or (self.vectors == FILLVAL).any():
+            new_vectors = np.where(
+                np.isnan(self.vectors) | (self.vectors == FILLVAL),
+                FILLVAL,
```
tmplummer, line 311: I don't know if you care, but this will convert existing NaNs to FILLVAL.

I would have thought that the FILLVAL got converted to NaN when reading in the CDF, making this code unnecessary, but that appears to not be the case.

imap_processing/mag/l1d/mag_l1d_data.py
```diff
@@ -298,26 +298,40 @@ def rotate_frame(self, end_frame: ValidFrames) -> None:
...
 
-        self.vectors = frame_transform(
+        new_vectors = frame_transform(
             self.epoch_et,
             self.vectors,
             from_frame=start_frame.spice_frame,
             to_frame=end_frame.spice_frame,
             allow_spice_noframeconnect=True,
         )
+        if np.isnan(self.vectors).any() or (self.vectors == FILLVAL).any():
```
tmplummer, line 308: Although there is not test coverage proving it, I believe that any input `self.vector` that has a single NaN component will result in frame_transform outputting a NaN vector for that transformed position vector. So, this check really only need be done for the vectors that contain FILLVAL.

imap_processing/tests/mag/test_mag_l1d.py
```diff
@@ -571,6 +571,57 @@ def test_enhanced_gradiometry_with_quality_flags_detailed():
...
+    vectors[2] = [np.nan, np.nan, np.nan]
+    vectors[4] = [1.0, np.nan, 3.0]
+    mag_l1d_test_class.vectors = vectors
+
+    # Set some MAGI vectors to FILLVAL and NaN
+    magi_vectors[1] = [FILLVAL, FILLVAL, FILLVAL]
+    magi_vectors[3] = [np.nan, np.nan, np.nan]
+    mag_l1d_test_class.magi_vectors = magi_vectors
+
+    def mock_frame_transform(
```
tmplummer, line 592: If you did want to test that position vector NaN results in NaN without doing the `where` on NaNs, you could mock the `geometry.get_rotation_matrix` function to just return a 3x3 identity matrix.

tmplummer (commented):
A couple of thoughts to consider.

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: No, I don't think the FILLVAL is converted to NaN. I also don't think the NaNs are converted to FILLVAL by the writer, but they should be FILLVAL.

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: That makes sense to me - I think I will leave it as-is just in case, I don't think it makes that much of a difference - unless you disagree

imap_processing/tests/mag/test_mag_l1d.py
maxinelasp: Smart, I like that change

imap_processing/mag/l1d/mag_l1d_data.py
tmplummer: Hmmm, I don't know why it isn't getting converted. It looks like it should be to me: https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/cdf/utils.py#L54

Also, by default in the CDF xarray_to_cdf function, the `nan_to_fillval` keyword defaults to True. 🤷

tmplummer (approved):
Good to go from my perspective.

imap_processing/mag/l1d/mag_l1d_data.py
maxinelasp: Let me track this down tomorrow and see if I can remove the NaN checks!

maxinelasp: I believe the reader does convert correctly to NaN when reading CDFs, but the output does not automatically convert. There are other spots in the code which fill in FILLVALs, so I'm going to leave this check in because it is possible to have a mix of NaN and FILLVALs.

PR #2909, opened 2026-04-02
I-ALiRT - Hit data access

laspsandoval:
This pull request updates the `process_hit` function in `imap_processing/ialirt/l0/process_hit.py` to improve how hit groups with zero status values are handled and logged, and adjusts the corresponding unit test. The main changes group all hit groups with zero status values and log them together as a warning, rather than logging each occurrence individually and skipping processing. The test is updated to reflect the new behavior and expected output.

**Improvements to hit group status handling:**

* Modified `process_hit` to collect all hit groups with zero status values in a `status_groups` list and log them together as a warning at the end, instead of logging and skipping each group immediately. This change improves logging clarity and ensures all groups are processed. [[1]](diffhunk://#diff-2c68c0931ae91c5fb9d37d67976f98dab0f93ed5ac01ca869790f1a3754a0c51R135) [[2]](diffhunk://#diff-2c68c0931ae91c5fb9d37d67976f98dab0f93ed5ac01ca869790f1a3754a0c51L154-R155) [[3]](diffhunk://#diff-2c68c0931ae91c5fb9d37d67976f98dab0f93ed5ac01ca869790f1a3754a0c51R218-R221)
* Removed redundant status checks and logging for each group, simplifying the code and reducing unnecessary log output.
* Updated import statements to remove unused `met_to_utc` import, as it is no longer needed for per-group logging.

**Unit test updates:**

* Updated the `test_process_hit` unit test to expect 15 processed hits (instead of 1), and to match the new expected values for the hit product, reflecting the changes in how hit groups with zero status values are handled.

### DIFF
--- a/imap_processing/ialirt/l0/process_hit.py
+++ b/imap_processing/ialirt/l0/process_hit.py
@@ -11,7 +11,7 @@ from imap_processing.ialirt.utils.grouping import (
     find_groups,
 )
 from imap_processing.ialirt.utils.time import calculate_time
-from imap_processing.spice.time import met_to_ttj2000ns, met_to_utc
+from imap_processing.spice.time import met_to_ttj2000ns
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


imap_processing/ialirt/l0/process_hit.py
```diff
@@ -172,14 +168,6 @@ def process_hit(xarray_data: xr.Dataset) -> list[dict]:
             grouped_data["hit_met"][(grouped_data["group"] == group).values].values[0]
         )
 
-        status_values = grouped_data["hit_status"][
```
lacoak21, line 175: Interesting was this a duplicate check here?

lacoak21 (approved):
Nice looks good and makes sense to me.

imap_processing/ialirt/l0/process_hit.py
laspsandoval: Yes. But also they want to produce data even if the status is not nominal. So that's what this ticket is about.

lacoak21: Cool yeah that part makes sense.

PR #2930, opened 2026-04-07
GLOWS L2 night flags

### LINKED ISSUE #2369: ENH - GLOWS L2 Apply sunrise/sunset offsets to is_night flag transitions
## Description
The infrastructure to read `sunrise_offset` and `sunset_offset` from pipeline settings already exists in `PipelineSettings` dataclass, but these offsets are not being applied anywhere in the processing code. Need to implement the logic to shift is_night flag transitions according to these offsets during L2 processing.

## Background
Per Marek (algorithm doc v4.4.7 Sec. 3.9.1, item 2):
- When is_night transitions from 0->1 (sunset): shift by `sunset_offset` histograms
- When is_night transitions from 1->0 (sunrise): shift by `sunrise_offset` histograms
- Offsets can be negative, zero, or positive values

## Required Changes
1. Identify where is_night flag transitions occur in L1B histogram sequences
2. Apply `sunset_offset` when detecting is_night 0->1 transition
3. Apply `sunrise_offset` when detecting is_night 1->0 transition
4. Ensure offsets are applied during L2 processing when computing from L1B histograms

Note: offsets are already read from pipeline settings and available for use.

## Reference
- Algorithm doc v4.4.7, Sec. 3.9.1, item 2
- Pipeline settings file: `imap_glows_lxx-pipeline-settings_20250923_v000.json`

## Related
Closes #1991

## Laura Notes
Pointing maneuver - is night
Sunrise - histograms on either side
Sunset - histograms on either side
Epoch = sampling time in l1b_data

Use offset from is_night timestamp and flag using the same flag

Remove is_night from good times histograms

Double check flags in L2

laspsandoval:
This pull request introduces logic to adjust the "is_night" flag in the GLows L2 processing pipeline, allowing sunrise and sunset offsets to be applied when determining valid (good) data blocks. This change ensures that the night region can be flexibly extended or shortened around sunrise and sunset transitions, as specified in the algorithm documentation. Comprehensive tests are also added to validate the new offset functionality.

**Night flag offset logic:**

* Added a new static method `apply_is_night_offsets` to `HistogramL2` that adjusts the `is_night` flag in the flags array based on configurable `sunrise_offset` and `sunset_offset` parameters, allowing for flexible extension or shortening of the night region at both transitions.
* Integrated the new offset logic into the L2 processing pipeline by applying offsets before selecting good data blocks, ensuring the selection honors the adjusted night/day boundaries.

**Testing enhancements:**

* Added a parameterized test `test_apply_is_night_offsets` that verifies the offset logic for various combinations of sunrise and sunset offsets, ensuring the `is_night` flag is correctly adjusted in all cases.
* Updated the test fixture `l1b_dataset_full` and related comments to clarify that all flags are good and both epochs are daytime, aligning with the new logic and improving test clarity. [[1]](diffhunk://#diff-85922de4b06d3b8b853330bbcaf5d45c1cac4405f6663481d8738eb2c19bd958L395-R428) [[2]](diffhunk://#diff-85922de4b06d3b8b853330bbcaf5d45c1cac4405f6663481d8738eb2c19bd958R438-R440) [[3]](diffhunk://#diff-85922de4b06d3b8b853330bbcaf5d45c1cac4405f6663481d8738eb2c19bd958L416-R452)

### DIFF
--- a/imap_processing/glows/l2/glows_l2_data.py
+++ b/imap_processing/glows/l2/glows_l2_data.py
@@ -370,10 +370,20 @@ class HistogramL2:
         """
         active_flags = np.array(pipeline_settings.active_bad_time_flags, dtype=float)
 
+        # Apply sunrise/sunset offsets to extend the night region around
+        # is_night transitions before selecting good blocks.
+        flags = self.apply_is_night_offsets(
+            l1b_dataset["flags"].data,
+            is_night_idx=6,  # is_night is the 7th bad-time flag (0-indexed)
+            sunrise_offset=pipeline_settings.sunrise_offset,
+            sunset_offset=pipeline_settings.sunset_offset,
+        )
+        flags_da = xr.DataArray(flags, dims=l1b_dataset["flags"].dims)
+
         # Select the good blocks (i.e. epoch values) according to the flags. Drop any
         # bad blocks before processing.
         good_data = l1b_dataset.isel(
-            epoch=self.return_good_times(l1b_dataset["flags"], active_flags)
+            epoch=self.return_good_times(flags_da, active_flags)
         )
         # TODO: bad angle filter
         # TODO: filter bad bins out. Needs to happen here while everything is still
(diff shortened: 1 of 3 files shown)


imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -533,6 +543,97 @@ def return_good_times(flags: xr.DataArray, active_flags: NDArray) -> NDArray:
...
+        """
+        flags_with_offsets = flags.copy()
+
+        # If sunrise_offset=0 and sunset_offset=0 then no corrections are needed
+        # relative to is_night transition set onboard.
+        if sunrise_offset == 0 and sunset_offset == 0:
+            return flags_with_offsets
+
+        is_night_col = flags[:, is_night_idx]
+        n = flags.shape[0]
```
tech3371: can you use different variable name for `n`? I feel like it will loose it's meaning after a while.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -370,10 +370,20 @@ def __init__(
         """
         active_flags = np.array(pipeline_settings.active_bad_time_flags, dtype=float)
 
+        # Apply sunrise/sunset offsets to extend the night region around
+        # is_night transitions before selecting good blocks.
+        flags = self.apply_is_night_offsets(
+            l1b_dataset["flags"].data,
+            is_night_idx=6,  # is_night is the 7th bad-time flag (0-indexed)
```
vmartinez-cu: Will this always be a value of 6? It might be worth adding this as a constant to the GlowsConstants class

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -533,6 +543,97 @@ def return_good_times(flags: xr.DataArray, active_flags: NDArray) -> NDArray:
...
+        -----
+        Algorithm doc v4.4.7, Sec. 3.9.1, item 2
+        is_night: 1 = daytime (good), 0 = night (bad)
+        """
+        flags_with_offsets = flags.copy()
+
+        # If sunrise_offset=0 and sunset_offset=0 then no corrections are needed
+        # relative to is_night transition set onboard.
+        if sunrise_offset == 0 and sunset_offset == 0:
+            return flags_with_offsets
```
vmartinez-cu: Since offsets don't get applied to the flags in this case, you could do this check on the `flags` array before copying it to `flags_with_offsets`. This way you can avoid an unnecessary copy of the data and make it more clear that the original `flags` is returned without corrections.

imap_processing/tests/glows/test_glows_l2_data.py
```diff
@@ -342,6 +342,38 @@ def test_filter_good_times():
...
+        (1, 1, [1, 1, 1, 1, 0, 0, 0, 1]),
+        # sunset>0 shortens at sunset; sunrise<0 shortens at sunrise
+        (-1, 1, [1, 1, 1, 1, 0, 1, 1, 1]),
+        # sunset<0 extends at sunset; sunrise>0 extends at sunrise
+        (1, -1, [1, 1, 0, 0, 0, 0, 0, 1]),
+        # sunset<0 extends at sunset; sunrise<0 shortens at sunrise
+        (-1, -1, [1, 1, 0, 0, 0, 1, 1, 1]),
+        # zero offsets: no change
+        (0, 0, [1, 1, 1, 0, 0, 0, 1, 1]),
+    ],
```
vmartinez-cu, line 362: Nitpick: since the parameters are in order of `sunrise_offset` then `sunset_offset`, consider updating your comments to follow this order as well for clarity

vmartinez-cu (approved):
Nice job on this!

imap_processing/glows/l2/glows_l2_data.py
medley56: I agree on avoiding magic constants, especially when it's a kwarg.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -533,6 +543,97 @@ def return_good_times(flags: xr.DataArray, active_flags: NDArray) -> NDArray:
...
+        -------
+        numpy.ndarray
+            Copy of flags with the is_night column adjusted per the offsets.
+
+        Notes
+        -----
+        Algorithm doc v4.4.7, Sec. 3.9.1, item 2
+        is_night: 1 = daytime (good), 0 = night (bad)
+        """
+        flags_with_offsets = flags.copy()
```
medley56: I'm not sure how big this array is but copy will have memory implications. If you can avoid it without creating ugly side effects to the parent data set, consider modifying it in place.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -533,6 +543,97 @@ def return_good_times(flags: xr.DataArray, active_flags: NDArray) -> NDArray:
...
+
+        Parameters
+        ----------
+        flags : numpy.ndarray
+            Flags array with shape (n_epochs, FLAG_LENGTH), 0=bad, 1=good.
+        is_night_idx : int
+            Column index of the is_night flag in the flags array.
+        sunrise_offset : float
+            Additional histogram shift at the sunrise (is_night 1->0) transition.
+        sunset_offset : float
```
medley56: These offsets are being passed in as floats but seem to be used as ints. If they need to be ints, I would make the method signature require ints. I would be confused if I passed in 3.1, 3.2, and 3.4 and got the same result each time.

imap_processing/glows/l2/glows_l2_data.py
medley56: If you do this, be careful when you create the flags_da DataArray in init. I'm not sure how Xarray handles that. It might use the reference to the original flags array (i.e. no copy). So IFF the flags are unchanged, if you subsequently change the flags in the l1b_dataset, you'd also be changing the flags in flags_da. However, if the flags _do_ change here, then you'd get a copy rather than a reference, which could lead to weird bugs.

imap_processing/glows/l2/glows_l2_data.py
```diff
@@ -533,6 +543,97 @@ def return_good_times(flags: xr.DataArray, active_flags: NDArray) -> NDArray:
...
+
+        if sunrise_offset > 0:
+            # Night (flag = 0) extends by sunrise_offset relative
+            # to is_night 0 -> 1 transition.
+            for i in sunrise_index:
+                flags_with_offsets[
+                    i + 1 : min(n, i + 1 + int(sunrise_offset)), is_night_idx
+                ] = 0
+
+        if sunrise_offset < 0:
```
medley56: This should be an elif, otherwise both mutually exclusive if statements get hit every time. Also, I would add a comment that the fall through case of sunrise_offset == 0 just returns the array as-is.

imap_processing/tests/glows/test_glows_l2_data.py
```diff
@@ -342,6 +342,38 @@ def test_filter_good_times():
...
+    flags[3:6, 6] = 0  # epochs 3-5 are night
+
+    result = HistogramL2.apply_is_night_offsets(
+        flags,
+        is_night_idx=6,
+        sunrise_offset=sunrise_offset,
+        sunset_offset=sunset_offset,
+    )
+
+    assert np.array_equal(result[:, 6], np.array(expected_is_night, dtype=float))
```
medley56, line 379: Consider adding a check that the original `flags` array is unchanged and that result is not the same object as flags, if that's the intended behavior.

medley56 (approved):

PR #2939, opened 2026-04-08
BUG: ULTRA l2 downsample counts

### LINKED ISSUE #2938: ULTRA L2 downsample counts BUG
There was a bug in ultra l2 code that used the wrong indices to order counts from ring - nested

lacoak21:
# Change Summary
closes #2938

## Overview

Bug fix. 

To convert the counts array from ring to nested, I should use nest2ring because this returns what each ring pixels is at each nested index. Using the reversal was scrambling counts. 

ring2nest gives, for each ring pixel, what is the nested pixel ind. But we want for each nested pixel what is the ring pixel ind.


tmplummer (approved):
👍

PR #2941, opened 2026-04-08
ULTRA l1c theta limit check

### LINKED ISSUE #2856: ULTRA limit fov for 3month maps
The ULTRA team requested that we build psets with a more limited FOV. This will effect l1c code.

lacoak21:
# Change Summary

closes #2856

## Overview

The ultra team wants to limit the FOV, This pr calculates pixels that are not in the limited FOV and those pixels are excluded from counts, exposure factor, and sensitivity (eff, gf). 

I sent maps to the ULTRA team and they are looking good. I think they still may need to check some things. 

## File changes
imap_processing/ultra/l1c/helio_pset.py
- Add boolean flag for restricting the FOV
- Make sure to filter counts that are not in the restricted fov

imap_processing/ultra/l1c/l1c_lookup_utils.py
- Create function to check if theta/phi pairs are in the restricted fov 
- When creating the valid pixel map, skip pixels not in restricted fov - downstream this is used by exposure factor, efficiencies and geometric function calculations. 

imap_processing/ultra/l1c/ultra_l1c_pset_bins.py
- I also found a small bug where dims were reversed. That code is currently not in use so it never errored out. 

## Testing

Update test to work with new rejection logic. Add new tests for the fov restriction code.

### DIFF
--- a/imap_processing/tests/ultra/unit/test_helio_pset.py
+++ b/imap_processing/tests/ultra/unit/test_helio_pset.py
@@ -39,29 +39,31 @@ def test_validate_exposure_time_and_sensitivities(
         .squeeze()
     )
     npix = 12288  # nside 32
-    # Create a minimal dataset to pass to the function
+    # Create a minimal goodtimes dataset to pass to the function
+    # Create mock spin data that has 5525 nominal spins
+    # Create DataFrame
+    nspins = 5522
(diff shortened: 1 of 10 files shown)


tmplummer (approved):
This all looks good to me.

